PR #17075, opened 2023-04-18
[config] Add regression message for Conan v2 builds

uilianries:
When a package which is listed in https://github.com/conan-io/conan-center-index/blob/master/.c3i/conan_v2_ready_references.yml fails to build on a new PR, it will be considered a regression error, which means, for that specific package, Conan v2 will be mandatory.

However, the current information on the PR, informs that Conan v2 is only informative and is not mandatory, creating a misinformation.

We are preparing a new feature which will update the message according the situation, so users will not be confused by the build result message.

---

### DIFF
--- a/.c3i/config_v2.yml
+++ b/.c3i/config_v2.yml
@@ -36,7 +36,7 @@ tasks:
     update_labels: false
     user_feedback:
       title: "Conan v2 pipeline"
-      description: "> **Note**: Conan v2 builds are informative and they are not required for the PR to be merged."
+      description: "> **Note**: Conan v2 builds may be required once they are on the [v2 ready](https://github.com/conan-io/conan-center-index/blob/master/.c3i/conan_v2_ready_references.yml) list"
       regression: "> **Regression**: Conan v2 builds are mandatory and they are required for the PR to be merged, because this recipe worked with Conan v2 previously."
       text_on_failure: "The v2 pipeline failed. Please, review the errors and note this will be required for pull requests to be merged in the near future."
       collapse_on_success: false
--- a/.c3i/config_v2.yml
+++ b/.c3i/config_v2.yml
@@ -37,6 +37,7 @@ tasks:
     user_feedback:
       title: "Conan v2 pipeline (informative, not required for merge)"
       description: "> **Note**: Conan v2 builds are informative and they are not required for the PR to be merged."
+      regression: "> **Regression**: Conan v2 builds are mandatory and they are required for the PR to be merged, because this recipe worked with Conan v2 previously."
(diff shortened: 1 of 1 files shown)


jwillikers (dismissed):

.c3i/config_v2.yml
```diff
@@ -37,6 +37,7 @@ tasks:
     user_feedback:
       title: "Conan v2 pipeline (informative, not required for merge)"
```
prince-chrismc, line 38:
```suggestion
      title: "Conan v2 pipeline"
```

.c3i/config_v2.yml
```diff
@@ -37,6 +37,7 @@ tasks:
...
       description: "> **Note**: Conan v2 builds are informative and they are not required for the PR to be merged."
```
prince-chrismc, line 39:
```suggestion
      description: "> **Note**: Conan v2 builds may be required once they are on the [v2 required reference]() list"
```

prince-chrismc (commented):
We should update the comment from the bot too

.c3i/config_v2.yml
uilianries: It's optional by default, only for those which are listed in the conan_v2_ready_references.yml they mandatory. It's better keeping it explicit, removing the informative section may cause confusion.

uilianries: Sorry, yes, the `title` field needs to be updated.

prince-chrismc (approved):

.c3i/config_v2.yml
prince-chrismc: Yes I put it in the description with a link :)

I think it's the right time to be more pushy about supporting v2

uilianries: > I think it's the right time to be more pushy about supporting v2

Agreed. :)

.c3i/config_v2.yml
```diff
@@ -35,8 +35,9 @@ tasks:
     write_comments: false
     update_labels: false
     user_feedback:
-      title: "Conan v2 pipeline (informative, not required for merge)"
-      description: "> **Note**: Conan v2 builds are informative and they are not required for the PR to be merged."
+      title: "Conan v2 pipeline"
+      description: "> **Note**: Conan v2 builds may be required once they are on the [v2 ready](https://github.com/conan-io/conan-center-index/blob/master/.c3i/conan_v2_ready_references.yml) list"
```
SSE4, line 39: why `may be`? not always?

SSE4 (approved):

.c3i/config_v2.yml
prince-chrismc: There's a missing comma but I should be fine, the intention is clear about the condition

PR #17077, opened 2023-04-18
civetweb: bump openssl

SpaceIm:

---

### DIFF
--- a/recipes/civetweb/all/conanfile.py
+++ b/recipes/civetweb/all/conanfile.py
@@ -83,9 +83,9 @@ class CivetwebConan(ConanFile):
     def requirements(self):
         if self.options.with_ssl:
             if Version(self.version) < "1.16":
-                self.requires("openssl/1.1.1t")
+                self.requires("openssl/[^1]")
             else:
-                self.requires("openssl/3.1.0")
+                self.requires("openssl/[>=1 <4]")
         if self.options.get_safe("with_zlib"):
             self.requires("zlib/1.2.13")
 
--- a/recipes/civetweb/all/conanfile.py
+++ b/recipes/civetweb/all/conanfile.py
@@ -1,4 +1,4 @@
-from conan import ConanFile
+from conan import ConanFile, conan_version
 from conan.errors import ConanInvalidConfiguration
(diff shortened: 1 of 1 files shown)


jwillikers (approved):

SpaceIm:
```
ERROR: Version range '^1' from requirement 'openssl/[^1]' required by 'civetweb/1.13' could not be resolved in remote 'conan-center'
```

@jcar87 @memsharded Is it not supposed to be a valid version range, equivalent to `>=1 < 2`?

SpaceIm:
```
ERROR: Version range '>=1 <2' from requirement 'openssl/[>=1 <2]' required by 'civetweb/1.15' could not be resolved in remote 'conan-center'
```

?

jwillikers (dismissed):

recipes/civetweb/all/conanfile.py
```diff
@@ -82,7 +82,10 @@ def layout(self):
 
     def requirements(self):
         if self.options.with_ssl:
-            self.requires("openssl/1.1.1t")
+            if Version(self.version) < "1.16":
```
jcar87, line 85:
```
            if Version(self.version) < "1.16":
                self.requires("openssl/1.1.1t")
            else:
                self.requires("openssl/[>=1 <4]")
```

I would suggest something like this - indeed Conan 1.x will not be able to resolve a version range that does not include OpenSSL 3.x - I wouldn't complicate things, if the older versions of civetweb are _not_ compatible with OpenSSL 3.x, we can keep it to the most recent OpenSSL 1.1.1.

OpenSSL 1.1.1 will be on its end of life later this year - https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/ - I think there won't be may more version bumps anyway in this period of a few months until then.

jcar87 (changes requested):
See comment re: requirement on OpenSSL

recipes/civetweb/all/conanfile.py
SpaceIm: done

jwillikers (approved):

jcar87 (approved):

PR #17093, opened 2023-04-18
uni-algo: fix broken logic in validate() + cleanup dead code

SpaceIm:
- fix logic in `validate()`:
  - it was broken for legacy compiler=Visual Studio because it was testing version of this flavor against version of compiler=msvc flavor...
  - check compiler.cppstd if set
  - consistent min versions between legacy Visual Studio & new msvc
- remove dead code (export_conandata_patches useless here)
- no need to call CMakeToolchain in case of header-only
- simplify definition of cmake_file_name & cmake_target_name. Relying on self.name to explicitly define these properties is not only unnecessarily confusing, but also not future proof in case of recipe renaming. Rule of thumb: never rely on `self.name` to set libs intrinsic values. It's fine for stuff like ConanInvalidConfiguration or creating a custom file because they are related to recipe, not lib underlying logic.

---

### DIFF
--- a/recipes/uni-algo/all/conanfile.py
+++ b/recipes/uni-algo/all/conanfile.py
@@ -83,9 +83,10 @@ class UniAlgoConan(ConanFile):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
-        tc = CMakeToolchain(self)
-        tc.variables["UNI_ALGO_HEADER_ONLY"] = self.options.header_only
-        tc.generate()
+        if not self.options.header_only:
+            tc = CMakeToolchain(self)
+            tc.variables["UNI_ALGO_HEADER_ONLY"] = self.options.header_only
+            tc.generate()
 
     def build(self):
         if not self.options.header_only:
--- a/recipes/uni-algo/all/conanfile.py
+++ b/recipes/uni-algo/all/conanfile.py
@@ -2,7 +2,7 @@ from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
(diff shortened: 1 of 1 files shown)


recipes/uni-algo/all/conanfile.py
```diff
@@ -69,24 +67,26 @@ def package_id(self):
...
+
+        if is_msvc(self) and self.options.get_safe("shared"):
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared with msvc")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
-        tc = CMakeToolchain(self)
-        tc.variables["UNI_ALGO_HEADER_ONLY"] = self.options.header_only
```
RubenRBS, line 88: I prefer when this is set for all cases and the value is changed, but it's in no way a blocker for me

recipes/uni-algo/all/conanfile.py
```diff
@@ -32,21 +33,18 @@ class UniAlgoConan(ConanFile):
...
     def _min_cppstd(self):
-        return 17
+        return "17"
 
     @property
     def _compilers_minimum_version(self):
         return {
             "Visual Studio": "16",
-            "msvc": "191",
+            "msvc": "192",
```
RubenRBS, line 42: Nice catch!

RubenRBS (approved):
Thanks a lot for your effort in fixing the recipe up, and with the detailed PR message, it really helps to understand the changes and makes the review much easier :)

jwillikers (approved):

recipes/uni-algo/all/conanfile.py
SpaceIm: I prefer to not keep a logic which doesn't make sense. Here conan_toolchain.cmake is not used at all if header only in current recipe logic. So keeping this logic would confuse maintainers, since it may mean that it's mandatory when header only is True.

PR #17109, opened 2023-04-18
aws-c-compression: add missing interface definition if shared + modernize more for conan v2

SpaceIm:
see https://github.com/awslabs/aws-c-compression/blob/v0.2.16/include/aws/compression/exports.h

---

### DIFF
--- a/recipes/aws-c-compression/all/test_v1_package/CMakeLists.txt
+++ b/recipes/aws-c-compression/all/test_v1_package/CMakeLists.txt
@@ -1,10 +1,8 @@
 cmake_minimum_required(VERSION 3.1)
-project(test_package LANGUAGES C)
+project(test_package)
 
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
 conan_basic_setup(TARGETS)
 
-find_package(aws-c-compression REQUIRED CONFIG)
-
-add_executable(${PROJECT_NAME} ../test_package/test_package.c)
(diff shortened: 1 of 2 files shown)


jwillikers (approved):

recipes/aws-c-compression/all/test_v1_package/CMakeLists.txt
```diff
@@ -1,10 +1,8 @@
...
-project(test_package LANGUAGES C)
+project(test_package)
 
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
 conan_basic_setup(TARGETS)
 
-find_package(aws-c-compression REQUIRED CONFIG)
-
-add_executable(${PROJECT_NAME} ../test_package/test_package.c)
-target_link_libraries(${PROJECT_NAME} PRIVATE AWS::aws-c-compression)
```
danimtb, line 10: Why removing this find_package()? I mean, if we are still generating the alias targets to maintain compatibility, shouldn't we still keep this?

SpaceIm: I don't understand, it's a common pattern we have in all modern recipes. It avoids code duplication and mistakes by ensuring a common CMakeLists in test package & test_v1_package, so find_package is not removed at all, but delegated to CMakeLists of test_package.

danimtb: oh sorry, I didn't read it well. It makes sense

PR #17113, opened 2023-04-19
[libelf/0.8.13]: Fix windows clang build

calebkiage:
Specify library name and version:  **libelf/0.8.13**

---


jcar87 (commented):
Hi @calebkiage - thank you for your contribution.

Is there an issue or otherwise details about the issue this is trying to fix? In order for us to proceed with the review, we would need a PR description that describes the problem and ways to reproduce it, so that we can match it up with the proposed solution.

Thanks so much!

Issue #17131, opened 2023-04-19
[package] glu/system: "Package opengl was not found in the pkg-config search path" on Ubuntu


SpaceIm:
### Description

On WSL2 Ubuntu 22.04, `glu/system` fails in `package_info()` while calling `pkg_config.fill_cpp_info`.

It can't find pc file of its dependency: `opengl.pc`. And indeed this file is missing in `/usr/lib/x86_64-linux-gnu/pkgconfig` unless I install `libopengl-dev` with apt.
Here is the problem:
- system installed `glu.pc` expects to find `opengl.pc`
- `opengl/system` recipe doesn't install `libopengl-dev` but `libgl-dev` (or fallback to `libgl1-mesa-dev`) (these legacy packages provide `gl.pc`, not `opengl.pc`)

There is an issue here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1009731

But I think that `libopengl-dev` should be added to apt in opengl recipe.  `libgl-dev` is the legacy package actually:

```
$ more /usr/lib/x86_64-linux-gnu/pkgconfig/gl.pc
prefix=/usr
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: GL
Description: Legacy OpenGL and GLX library and headers.
Version: 1.2
Libs: -L${libdir} -lGL
Cflags: -I${includedir}
```

```
$ more /usr/lib/x86_64-linux-gnu/pkgconfig/opengl.pc
prefix=/usr
libdir=${prefix}/lib/x86_64-linux-gnu
includedir=${prefix}/include

Name: OpenGL
Description: OpenGL (without GLX) library and headers.
Version: 4.5
Libs: -L${libdir} -lOpenGL
Cflags: -I${includedir}
```

### Package and Environment Details

* Package Name/Version: **glu/system**
* Operating System+version: **Linux Ubuntu 22.04 (WSL2)**
* Compiler+version: **GCC 12.1.0**
* Conan version: **conan 2.0.4**
* Python version: **Python 3.10.6**

### Conan profile

```
{% set cc = "/usr/bin/gcc-12" %}
{% set cxx = "/usr/bin/g++-12" %}

[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=12
compiler.libcxx=libstdc++11
build_type=Release
[options]
[tool_requires]
[conf]
[...]

PR #17144, opened 2023-04-20
opengl: prefer libopengl-dev on Debian

SpaceIm:
closes https://github.com/conan-io/conan-center-index/issues/17131

---

### DIFF
--- a/recipes/opengl/all/conanfile.py
+++ b/recipes/opengl/all/conanfile.py
@@ -1,6 +1,7 @@
 from conan import ConanFile
-from conan.tools.system import package_manager
+from conan.errors import ConanException
 from conan.tools.gnu import PkgConfig
+from conan.tools.system import package_manager
 
 required_conan_version = ">=1.50.0"
 
@@ -30,7 +31,7 @@ class SysConfigOpenGLConan(ConanFile):
         yum.install(["mesa-libGL-devel"], update=True, check=True)
 
         apt = package_manager.Apt(self)
-        apt.install_substitutes(["libgl-dev"], ["libgl1-mesa-dev"], update=True, check=True)
+        apt.install_substitutes(["libopengl-dev"], ["libgl-dev"], ["libgl1-mesa-dev"], update=True, check=True)
 
         pacman = package_manager.PacMan(self)
         pacman.install(["libglvnd"], update=True, check=True)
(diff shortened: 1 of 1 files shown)


ericLemanissier:
I'm not sure about this change. libpengl-dev does not seem to be a drop-in replacement of libgl-dev, but more of a complement. 
[libopengl-dev](https://debian.pkgs.org/11/debian-main-amd64/libopengl-dev_1.3.2-1_amd64.deb.html) provides only opengl.pc, without any headers, contrary to [libgl-dev](https://debian.pkgs.org/11/debian-main-amd64/libgl-dev_1.3.2-1_amd64.deb.html) which provides gl.pc (notice the different name) and header files.
I think we need to install both packages and publish them as different components.

jwillikers (approved):

SpaceIm:
Indeed headers are missing, I don't know which package to install. But gl.pc refers to the legacy implementation, it's not complementary.

ericLemanissier:
The point is that some package call `pkg-config opengl ...` or link to `-lOpenGL` directly, and other packages call `pkg-config gl ...` or link to `-lGL` directly. There is no way `opengl/system` can decide which of these should be installed. it depends on the consumer

SpaceIm:
So what would you do? package_info() has to be deterministic somehow.

ericLemanissier:
install both, and publish both a different components (the same way xorg publishes several components https://github.com/conan-io/conan-center-index/blob/master/recipes/xorg/all/conanfile.py#L83). We should always publish gl component, but publish opengl component only if it is actually installed (some distros do not provide it yet)

Alternatively, we could just make another recipe for libopengl-dev, but it's unfortunate that the name opengl/system is already taken :/ It would have been clearer to have opengl/system install libopengl-dev and gl/system install libgl-dev

SpaceIm:
When both libopengl-dev & libgl-dev are installed on my distro, find_package(OpenGL) displays a warning that it has found two implementations of OpenGL and it chooses the deprecated one.

ericLemanissier:
that's handled by https://cmake.org/cmake/help/latest/policy/CMP0072.html

artem-ogre:
Will this PR be worked on? Because of this problem qt5 package can't be built with gcc-11.

valgur (approved):

PR #17147, opened 2023-04-21
jsonnet: add version 0.20.0, support conan v2

toge:
Specify library name and version:  **jsonnet/***

---


toge:
waiting for https://github.com/conan-io/conan-center-index/pull/17146

recipes/jsonnet/all/patches/0.20.0/0003-use-cpp17.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index c14c93c..b1cb1b3 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -43,7 +43,7 @@ else()
+     message(WARNING "Compiler ${CMAKE_CXX_COMPILER_ID} not supported")
+ endif()
+ 
+-set(CMAKE_CXX_STANDARD 11)
```
memsharded, line 9: Not sure that just changing (and hardcoding the C++ standard) is the best way. I guess that if users define ``compiler.cppstd=20`` it will be mostly ignored and overwritten because of this being hardcoded here?

(not saying this is a problem of this PR, but it seems the problem was there in the past too)

recipes/jsonnet/all/conandata.yml
```diff
@@ -9,18 +12,37 @@ sources:
...
   "0.17.0":
     - patch_file: "patches/0.17.0/0001-fix-nlohmann-include.patch"
-      base_path: "source_subfolder"
+      patch_description: "fix include path to use cci package"
+      patch_type: "conan"
     - patch_file: "patches/0.17.0/0002-cmake-fixes.patch"
-      base_path: "source_subfolder"
+      patch_description: "fix rapidyaml name, add installation"
+      patch_type: "conan"
+    - patch_file: "patches/0.17.0/0003-std-uint32_t.patch"
```
memsharded, line 46: Patches to C++ source code are a raising concern, and we are moving towards considering them a red flag in general.
If this is to force the 0.17.0 source code to be C++17 compatible, then the approach should be more like:
- Do not add the patch
- Raise a ``ConanInvalidConfiguration()`` for <=0.17.0 if cppstd>=17. 

That is a perfectly valid approach, the original 0.17.0 library didn't support 17, unless they patch it in a new 0.17.1, we should probably try to be as aligned as possible.

memsharded (commented):
This is looking good in general, thanks very much for your contribution.
Please have a look at the comments and let me know if you have any question.

uilianries:
I started an internal CI build to generate the missing `rapidyaml/0.5.0` 🤞

memsharded:
The build is failing due to a missing binary in the dependencies. We will try to fill that binary, could you please check the comments above? Thanks!

memsharded:
We made the PR green, building some missing dependencies binaries.
Please have a look at the comments and let us know. Thanks!

recipes/jsonnet/all/patches/0.20.0/0003-use-cpp17.patch
toge: @memsharded 
Thank you for your advise.
You are right. It is not mandatory.
I will remove it.

It is an upstream patch which have been merged already.
https://github.com/google/jsonnet/pull/1076

recipes/jsonnet/all/conandata.yml
toge: @memsharded 
I agree with you.
It is very reasonable.
I will fix another PR with same issues.
Thanks a lot!

toge:
@memsharded  @uilianries 
Thanks a lot for all the help.
I have improved this PR.
Let's see the result of CI.

recipes/jsonnet/all/patches/0.20.0/0003-use-cpp17.patch
toge: @memsharded 
I met a compilation error caused by `set(CMAKE_CXX_STANDARD 11)` in msvc.
May I apply `0003-use-cpp17.patch` again?


```
error C2429: language feature 'nested-namespace-definition' requires compiler flag '/std:c++17' 
(compiling source file C:\J\w\prod\BuildSingleReference@2\.conan\data\jsonnet\0.20.0\_\_\build
```

memsharded: My concern here is that it passes CI, but it is causing a mis-alignment of the resulting binaries. No matter what users pass ``compiler.cppstd=20`` or ``compiler.cppstd=14``, that apparently should be fine in some compilers, but failing in others. And the change affect all compilers (but not changing it is also not great, because it is still misaligned with the provided ``compiler.cppstd``

The ideal (for every package, not only this) would be:
- Remove it from CMakeLists.txt
- The ``conan_toolchain.cmake`` will define it, and it is mandatory in Conan 2.0
- If Conan 2.0-only support was possible we would be done. But if we need to maintain the 1.X pipeline and behavior, we could explicitly define it when ``compiler.cppstd=None`` in the ``toolchain.cache_variables["CMAKE_CXX_STANDARD"]`` to the value that it needs.

I am approving this PR, this is a net improvement, it only affects the new version, this can be addressed later.

memsharded (dismissed):

recipes/jsonnet/all/patches/0.20.0/0003-use-cpp17.patch
```diff
@@ -0,0 +1,13 @@
+iff --git a/CMakeLists.txt b/CMakeLists.txt
+index c14c93c..b1cb1b3 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -43,7 +43,7 @@ else()
+     message(WARNING "Compiler ${CMAKE_CXX_COMPILER_ID} not supported")
+ endif()
+
+-set(CMAKE_CXX_STANDARD 11)
++set(CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
memsharded: Just making sure to add something in the recipe files, that can make our life easier later and help to improve this in the future.

memsharded (approved):

RubenRBS (approved):

PR #17153, opened 2023-04-21
ccache: add version 4.8

gjasny:
Specify library name and version:  **ccache/4.8**

---

### DIFF
--- a/recipes/ccache/all/conanfile.py
+++ b/recipes/ccache/all/conanfile.py
@@ -63,9 +63,6 @@ class CcacheConan(ConanFile):
                     f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
                 )
 
-    def package_id(self):
-        del self.info.settings.compiler
-
     def build_requirements(self):
         self.tool_requires("cmake/3.25.3")
 
--- a/recipes/ccache/all/conanfile.py
+++ b/recipes/ccache/all/conanfile.py
@@ -1,4 +1,4 @@
-from conan import ConanFile, conan_version
+from conan import ConanFile
 from conan.tools.cmake import cmake_layout, CMake, CMakeToolchain, CMakeDeps
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.files import copy, get, apply_conandata_patches, export_conandata_patches
@@ -62,10 +62,9 @@ class CcacheConan(ConanFile):
                 raise ConanInvalidConfiguration(
                     f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
                 )
-        if Version(conan_version).major < 2 and \
-            self.settings.compiler== "clang" and Version(self.settings.compiler.version).major == "11" and \
+        if self.settings.compiler== "clang" and Version(self.settings.compiler.version).major == "11" and \
             self.settings.compiler.libcxx == "libstdc++":
-            raise ConanInvalidConfiguration(f"{self.ref} requires C++ filesystem library, which CCI doesn't support.")
+            raise ConanInvalidConfiguration(f"{self.ref} requires C++ filesystem library, that is not supported by Clang 11 + libstdc++.")
 
     def build_requirements(self):
         self.tool_requires("cmake/3.25.3")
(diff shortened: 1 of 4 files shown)


gjasny:
@prince-chrismc do you have an idea why the conan1 build on Linux fails in the test stage with:
```
----Running------
> ccache --version
-----------------
ccache: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by ccache)
ccache: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' not found (required by ccache)
ccache: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ccache)
```

From the logs it looks like no build happened and the binary package has been re-used from an incompatible build.

prince-chrismc:
The clang 11 image for 1.x has a bad stdlibc++ installed 😀 

Just block that exact config (there's examples in cci) and you should be fine to continue.

We are evaluating removing it :)

gjasny:
> Just block that exact config (there's examples in cci) and you should be fine to continue.
@prince-chrismc I tried, I really tried. First in the build conanfile, later in the test conanfile. But it looks like the CI is running the package tests in an (to me) unexpected way. Could you please try it yourself and show me where I messed up?

Thanks,
Gregor

jcar87:
For executable tools (like `ccache`, `cmake`, etc)  where the recipe removes the `compiler` setting (such as this one), the CI service will then make a decision to build the binary for the `_oldest_` possible configuration (for Conan Center this would be gcc7) that means that for each `os`, `arch` combination there is only one binary, as the version of the compiler is not usually relevant to being able to run the executable - this is a way of reflecting the backward compatibility when running the executable on system with newer runtime (`glibc` and `libstdc++`).

In this case, _because_ the newer versions of `ccache` have a requirement to use gcc 8 or newer, a newer version of gcc is used to compile it, and because gcc8 users a _newer_ version of libstdc++, it will fail and not run when running on a system that has an older version of the runtime - in this case, the docker image for `clang11` has a libstdc++ that matches that of gcc7, and thus it will not run there. There is nothing wrong per se with the clang11 image in this instance.

My suggestion would one of the following:
* Remove the `package_id()` method from the recipe altogether, and let Conan Center CI build a different package with each supported compiler, or
* tweak the `package_id()` in such a way that it performs partial information erasure as specified in the [documentation](https://docs.conan.io/2/reference/conanfile/methods/package_id.html#partial-information-erasure). For example,

```
def package_id(self):
    v = Version(str(self.info.settings.compiler.version))
    if self.info.settings.compiler == "gcc" and v >= "8":
        self.info.settings.compiler.version = "GCC 8 or greater"
```

Either of the above will trigger a build with clang11 with its own package_id, rather than skipping it due to compatibility.

@uilianries  please advise for the preferred solution

uilianries:
I would remove package id and keep one package per configuration. It will result in more packages and probably some users will have missing package depending their local profiles, but it will guarantee that we build and test with the same compiler version.

gjasny:
> it will guarantee that we build and test with the same compiler version

I thought the same and pushed the change.

gjasny:
@uilianries Could you add yourself as a reviewer?

recipes/ccache/all/conanfile.py
```diff
@@ -67,23 +53,19 @@ def requirements(self):
...
-    def package_id(self):
-        del self.info.settings.compiler
+        check_min_vs(self, 192)
+        if not is_msvc(self):
+            minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+            if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+                raise ConanInvalidConfiguration(
+                    f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+                )
+        if Version(conan_version).major < 2 and \
```
uilianries, line 65: This is not related to the Conan version. Those docker images are available for any Conan client version, including 1.x, so in case someone tries to use `docker.io/conanio/clang11-ubuntu16.04:1.59.0` will suffer the same error. I need to investigate it further.

recipes/ccache/all/conanfile.py
```diff
@@ -1,10 +1,10 @@
-from conan import ConanFile
+from conan import ConanFile, conan_version
```
uilianries, line 1:
```suggestion
from conan import ConanFile
```
No longer needed, please remove it.

uilianries (approved):
LGTM

jcar87 (approved):
thanks @gjasny, @uilianries !

PR #17158, opened 2023-04-21
libmad: conan v2 support

SpaceIm:
I've replaced build system by a simple CMakeList.txt for 2 reasons:
- msvc build was relying on `.dsp` files which are not supported anymore by `devenv` since Visual Studio 2022.
- these old autotools files didn't scale (no support for macOS M1)

---


jwillikers (approved):

czoido (approved):
After checking this PR, I think that an exceptional situation like changing the build system it's justified for this case

PR #17168, opened 2023-04-22
Revert "nuraft: bump openssl & asio"

szmyd:
Reverts conan-io/conan-center-index#17081


szmyd:
If someone wants updates, they need to override, no blast everyone with the package right?

Issue #12705, opened 2022-08-31
[OpenBLAS] Target CPU autodetection makes binaries undeployable


nenomius:
https://github.com/xianyi/OpenBLAS#normal-compile says this:

> Simply invoking `make` (or `gmake` on BSD) will detect the CPU automatically. To set a specific target CPU, use `make TARGET=xxx`, e.g. `make TARGET=NEHALEM`.

Current recipe does not set that: https://github.com/conan-io/conan-center-index/blob/69fa4d8643d434ab7529e84c70a7d4321c95f9dd/recipes/openblas/all/conanfile.py#L65

This results in OpenBLAS autodetecting CPU features to build for using build-server CPU info, which makes resulting binaries broken for any CPU that don't have some of instruction sets that happened to be available at build time (crashes on illegal instructions and similar fun things).

uilianries:
@nenomius Any specific target that you are using and failed?

nenomius:
@uilianries
Ubuntu 20.04 x86_64.
We have Intel CPU with AVX512 supported on build-server, and some production servers don't have AVX512 supported. Those servers got crashes with SIGILL on instructions from AVX512 set.
We got OpenBLAS as dependency of dependency and i don't (yet) have minimal repro.

uilianries:
@nenomius okay, it explains better. PLease, for future issues, when something is not working, you can open using the `bug` template, which contains a nice form which help us to reproduce your scenario. 

Anyway, we will need to update that recipe. The thing is, we use cmake to build, so we will need to find some way to pass targets name to cmake and work just like for make command.

Issue #5910, opened 2021-06-15
[question] [OpenBLAS]  static library implicitly requeries libgfortran.so


udovenkoav1981:
Hello, i use openblas/0.3.13 from conan ceneter is static mode (with lapack), i add it as depency to my app (i use CMAKE build system).  
When i try to link my app i get error: undefined reference to `_gfortran_concat_string' (this is part of gfortran runtime). 

Problem is: that conan don't have any info that libopenblas.a  must be linked with libgfortran.so

If i use openblas in dymanic mode - i don't have any problem, because .so have explicitly dependency to libgfortran.so.

Can you fix OpenBLAS recipe?

Issue #5338, opened 2021-04-26
[question] why does the openblas recipe add gfortran to the system_libs?


lexknuther:
Why does the openblas recipe add gfortran to it's system libs?  Removing this seems to have no impact on a variety of Linux OSes (Ubuntu 20.04, CentOS 7, RHEL 8, etc.).  I thought that libgfortran was the libc equivalent for fortran and was automatically linked to all fortran executables.

samuel-emrys:
I'm pretty sure because to build openblas with lapack support you need to use a fortran compiler. Adding this as a system lib I assume makes the compiler discoverable by openblas, and so when it goes to build it can build with lapack support. Removing it appears to have no impact because:
(a) you haven't enabled a lapack build
(b) you don't have gfortran installed
(c) it does have an impact because you're no longer able to build it with lapack support

Note that openblas will fallback to compiling without lapack support if it can't find a fortran compiler - this doesn't trigger a build fail.

Issue #1867, opened 2020-06-08
[OpenBLAS] conanfile.py assumes gfortran compiler


leezu:
https://github.com/conan-io/conan-center-index/blob/5e30b91bd37f87337a98eb159ea529ae9091ab13/recipes/openblas/all/conanfile.py#L87-L88

gfortran is not the only Fortran compiler supported to compile OpenBLAS. Sometimes gfortran is a no-go as it introduces a runtime dependency on the GPL `libgfortran.so`. When redistributing artifacts with runtime dependency on `libgfortran.so`, `libgfortran.so` would need to be redistributed as well. However, many organizations (eg. [Apache Software Foundation](https://www.apache.org/legal/resolved.html#category-x)) have policies against redistributing GPL artifacts.

Flang Fortran compiler can be used instead of gfortran and does not introduce runtime dependencies on GPL libraries. Flang toolchain can be obtained via https://github.com/flang-compiler/flang/releases and for example OpenBLAS compiled via `CC=gcc make DYNAMIC_ARCH=1 USE_OPENMP=1 FC=flang`

What's the best way to structure a conanfile.py to support different fortran toolchains? Should I open an issue about fortran compile support at https://github.com/conan-io/conan/ as well?

Croydon:
This is a hidden system requirement anyway which we should avoid in the mid-term

I believe once we have LLVM packages we should package Flang and add it as a require

leezu:
@Croydon your suggestion is a good mid-term solution. Do you have any recommendation how to extend the existing approach until then? Would the following be appropriate:

- Add `flang` option to conanfile (based on https://github.com/conan-community/conan-openblas cf https://github.com/conan-io/conan-center-index/issues/1769)
- If `flang` is specified, export `FC=flang` to the cmake file if `FC` is not defined in users environment
- Change the linking line from `-lgfortran` to `-lflang -lflangrti -lpgmath`

Croydon:
I don't have a strong opinion either way, but I think such an option could be fine

Croydon:
We now have a gfortan package which re-packages official binaries for Windows, Linux and macOS. Better than nothing

https://github.com/conan-io/conan-center-index/pull/4249

PR #17173, opened 2023-04-23
openblas: Improve recipe plus conan v2

joakimono:
Specify library name and version:  **openblas/***

- Add latest OpenBLAS release version 0.3.23
- Migrate recipe to support Conan v2
- Add new option 'use_openmp', which defaults to False
- Refactor replace_in_file to patch files instead
- Fix intel fortran compiler identification in OpenBLAS cmake
- Allow option 'build_lapack' without fortran using C_LAPACK since 0.3.23
  - Should be the default option since 0.3.23? It is the default for OpenBLAS native CMakeLists
- Allow disabling AVX512 via environment variable, a workaround for #12705
- Only make gfortran a dependency if actually used, related to #5910, #1867, #5338

---


recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,221 @@ class OpenblasConan(ConanFile):
...
+    def _fortran_runtime(self, fortran_id):
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            if fortran_id == "GNU":
+                if self.settings.compiler == "gcc":
+                    if Version(self.settings.compiler.version).major in ["5", "6"]:
+                        return "gfortran"  # Runtime version gfortran3
+                    if Version(self.settings.compiler.version).major == "7":
+                        return "gfortran"  # Runtime version gfortran4
+                    if Version(self.settings.compiler.version).major > "7":
+                        return "gfortran"  # Runtime version gfortran5
```
samuel-emrys, line 49: This is some verbose logic to ultimately return `"gfortran"` - is there a reason it needs to be segmented like this? Would something like the following be sufficient here?
```suggestion
                    if Version(self.settings.compiler.version).major > "5":
                        return "gfortran"
```

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,221 @@ class OpenblasConan(ConanFile):
         "fPIC": [True, False],
         "build_lapack": [True, False],
         "use_thread": [True, False],
+        "use_openmp": [True, False],
         "dynamic_arch": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "build_lapack": False,
```
samuel-emrys, line 32: I'd suggest if it's possible to build lapack without fortran (I thought this was just a C api that ultimately still needed to link to the fortran libs?), then we should build it by default.  

Because of the requirement that all CCI packages must build with the defaults of their dependencies, this opens the door to a number of scientific packages that choke without lapack support.

```suggestion
        "build_lapack": True,
```

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,221 @@ class OpenblasConan(ConanFile):
...
-        cmake.definitions["NOFORTRAN"] = not self.options.build_lapack
-        cmake.definitions["BUILD_WITHOUT_LAPACK"] = not self.options.build_lapack
-        cmake.definitions["DYNAMIC_ARCH"] = self.options.dynamic_arch
-        cmake.definitions["USE_THREAD"] = self.options.use_thread
+            # This checks explicit user-specified fortran compiler
+            if not self._fortran_compiler:
+                if Version(self.version) < "0.3.21":
+                    self.output.warning(
+                        "Building with LAPACK support requires a Fortran compiler.")
+                else:
```
samuel-emrys, line 139: I don't think it's important to guard for different versions here. This modifications in this PR are handled by introducing a new recipe revision. If it can be applied to all versions then it should.
```suggestion
```

samuel-emrys (changes requested):
A few small changes, but looking pretty good!

recipes/openblas/all/conanfile.py
joakimono: No reason, other than compiler versions require different runtime versions of gfortran. I applied your suggested change.

recipes/openblas/all/conanfile.py
joakimono: Since openblas v0.3.21, it is possible to build with LAPACK support without Fortran. A f2c (fortran to c) converted copy of LAPACK 3.9.0 is bundled, which completely avoids the need for a Fortran runtime, see https://github.com/xianyi/OpenBLAS/releases/tag/v0.3.21 why 3.9.0 and not newer. This allows us to provide LAPACK on CCI without a Fortran compiler. One related issues is #4509.

recipes/openblas/all/conanfile.py
joakimono: Since 0.3.21 is the oldest version that allows building LAPACK without Fortran, the recipe needs this guard.

recipes/openblas/all/conanfile.py
joakimono: I have added a check in `config_options()` that changes the default `build_lapack` to `True` if the version is new enough. Note that the recipe then will have different default options depending on package version. I do not know if this against CCI policy.

samuel-emrys: If that's the case, I'm not sure it's desirable to introduce different defaults across versions. My intuition would be to keep it stable across versions (so keep `"build_lapack": False`, but this has to be balanced with the benefits of having lapack available to support other conan center packages. @uilianries @prince-chrismc can either of you weigh in here?

samuel-emrys:
@joakimono I think force pushing impacts review state and makes it difficult to see the changes you're making, so i think it's preferable to just fix with commits over the top. It's all squashed when merged anyway

recipes/openblas/all/conanfile.py
joakimono: Reminder to possibly weigh in regarding this suggested change. @uilianries @prince-chrismc.

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,222 @@ class OpenblasConan(ConanFile):
...
+
+        if self.settings.os == "Windows" and self.options.build_lapack and self.options.use_openmp:
+            # In OpenBLAS cmake/system.cmake: Disable -fopenmp for LAPACK Fortran codes on Windows
+            self.output.warning("OpenMP is disabled on LAPACK targets on Windows")
+
+    def build_requirements(self):
+        if self.options.build_lapack and self.settings.os == "Windows":
+            self.tool_requires("ninja/1.11.1")
+
+    def system_requirements(self):
```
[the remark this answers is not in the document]
samuel-emrys: I think the answer here is yes you should model the system library components when for the configurations they apply to. For example, if the os is linux and the version is less than 0.3.21 and `build_lapack == True`, you could add `gfortran` as a system requirement. It's not good to have implicit dependencies so yeah, model them here.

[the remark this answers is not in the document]
joakimono: The latest commit adds system requirements for gfortran and openmp, if applicable. It is a bit difficult to make an exhaustive package list for all various distributions, but they are verified on at least one recent distribution for each package manager. FreeBSD /pkg and Macos /brew (and possibly some others) are missing.

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +24,276 @@ class OpenblasConan(ConanFile):
...
+        apt = package_manager.Apt(self)
+        dnf = package_manager.Dnf(self)
+        pacman = package_manager.PacMan(self)
+        # pkg = package_manager.Pkg(self) # TODO
+        yum = package_manager.Yum(self)
+        zypper = package_manager.Zypper(self)
+
+        if self.info.options.build_lapack == "gfortran":
```
samuel-emrys, line 141: This isn't a valid `build_lapack` option in the current configuration

joakimono: That is true, but I have attempted to make use of `self.info`, for which is `options.build_lapack` is possibly extended in the [package_id](https://docs.conan.io/2/reference/conanfile/methods/package_id.html#package-id) function. It is a (sneaky) way to allow more binary package configurations without adding any option items. The main benefit is to avoid an exhaustive list of fortran compilers to the `build_lapack` option, but retain distinct binary packages depending on fortran compiler used (including runtime system packages). The correct way would be to extend `settings.yaml` with fortran compilers and distribute this, but this is not practical. The drawback is that a consumer would need to set either the environment variable `FC` or populating 'fortran' of `tools.build:compiler_executables` for this to be picked up correctly.

joakimono: I have removed system_requirements in the latest commit. If compiled with Fortran, the user is responsible for installing/deploying runtime dependencies. I agree that my suggested approach was incomplete and could cause troubles such as those you pointed out.

I have not yet added any new option such as `with_fortran` and have kept the `build_lapack` manipulation in package_id. I did not fully understand your suggestion regarding this. My intention with this is to "expose" different binary packages, at least to differentiate packages built with `build_lapack=True`,  but without Fortran (since 0.3.23) from those actually using Fortran.

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,214 @@ class OpenblasConan(ConanFile):
...
+        tc.variables["MSVC_STATIC_CRT"] = False
 
-        # This is a workaround to add the libm dependency on linux,
-        # which is required to successfully compile on older gcc versions.
-        cmake.definitions["ANDROID"] = self.settings.os in ["Linux", "Android"]
+        # Env variable escape hatch for enabling AVX512
+        no_avx512 = environ.get("NO_AVX512")
+        if no_avx512 is None:
+            no_avx512 = True
+        tc.variables["NO_AVX512"] = no_avx512
```
samuel-emrys, line 153: Is there a reason this one isn't an option on the package? Strikes me that this would modify the binary that's being produced, thus we would want an option to make sure this is reflected in the package id and control when the package id changes or doesn't?

recipes/openblas/all/conanfile.py
samuel-emrys: I guess the point I was making was that a `with_fortran` option might be more transparent as to what's causing a change in the package id. What you've got here is a clever solution but is a deviation from what I've normally seen in cci recipes so wasn't quite as intuitive. Given that this is to modify the package id based on compiler that hasn't been modelled though, I think that perhaps what you've got is a better solution than I proposed with `with_fortran`. I also hadn't considered that you'd already put in place a mechanism to propagate the system lib dependency downstream by reading from the `FORTRAN_COMPILER` file you create. Fine to disregard my suggestion here.

recipes/openblas/all/conanfile.py
joakimono: This was a deliberate attempt of a "simple" fix to create functional binary packages for consumer CPUs that do not have AVX-512, without having to create a new option. The build server might have AVX-512, which causes unusable binary packages, as reported in #12705.

We could/should add this as an option to be more transparent. However, there are also several other options that we might expose, e.g. `NO_AVX`, `NO_AVX2`, and probably others as well. These instruction sets are more commonly available, though, and is less important to expose.

Do you want me to add a new option, e.g. `with_avx512`, which defaults to `False`?

samuel-emrys: Yeah I think an option makes the most sense here. This is just how this is modelled in Conan. Especially if this is causing unusable binaries

joakimono: I've added a new option as suggested.

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,212 @@ class OpenblasConan(ConanFile):
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
-    @functools.lru_cache(1)
-    def _configure_cmake(self):
-        cmake = CMake(self)
+    def generate(self):
 
+        ninja_generator = "Ninja" if self.settings.os == "Windows" and \
+            self.options.build_lapack else None
+        tc = CMakeToolchain(self, generator=ninja_generator)
```
samuel-emrys, line 125: Is there a particular requirement to use ninja here? it's not common in cci recipes to be so explicit - ideally the user can use whatever their preference is.

samuel-emrys: @joakimono that doesn't seem strictly true - [this section of the wiki](https://github.com/xianyi/OpenBLAS/wiki/How-to-use-OpenBLAS-in-Microsoft-Visual-Studio#2-use-cmake-to-generate-visual-studio-solution-files) seems to suggest that it works. Perhaps it's just a matter of fleshing out the `validate()` section to make sure it's built with the right combination of configuration when the generator is visual studio? You should be able to check this with something like

```python
if "Visual Studio" in self.conf.get("tools.cmake.cmaketoolchain:generator") and self.options.get_safe("dynamic_arch"):
    # raise exception
```
Alternatively, I think something preferable to specifying to use Ninja all the time would be to override this in `configure` - though that's still questionable... (I'm not sure if the `conf` values are considered read only) something like
```python
def configure(self):
    generator = self.conf.get("tools.cmake.cmaketoolchain:generator")
    if "Visual Studio" in generator:
        self.output.warning(f"{generator} is not compatible with {self.name}. Updating tools.cmake.cmaketoolchain:generator to use 'Ninja' instead.")
        self.conf.update("tools.cmake.cmaketoolchain:generator", "Ninja")
```

Regardless, I don't like that this forces the use of ninja - especially where any of the other generators may be equally valid except for the visual studio one.

joakimono: My current proposal only force to use the Ninja generator in cases I have confirmed by trial that Visual Studio's standard generator fails to successfully build the project. It should be noted that the default value of the `generator` argument of `CMakeToolchain` is `None`, for which `ninja_generator` in the snippet above also evaluates when Ninja is not needed. To be clear; Ninja is only needed when building on Windows with option `build_lapack=True`. The only exception, if i recall correctly, is if it is built without a Fortran compiler. I could add another check to not force Ninja if no Fortran compiler is used.

I will make some tests regarding your suggestions above, especially regarding get/set of `tools.cmake.cmaketoolchain:generator`. However, I'm not sure if this variable is defined by default, which makes checking for 'Visual Studio' pointless unless the user explicitly has specified a generator. On second thought, it might be useful to inform the user that their requested generator has been overridden.

samuel-emrys: yes, you're right - it's not defined by default, and it will just choose the platform default. On second look, I think what you've got is probably fine for most cases, but we should try to adhere to the law of least surprise. In that vein, I think you should do the following:
1. Add a warning output to indicate that the `Ninja` generator is being used when the selection logic is satisfied
2. Check that `tools.cmake.cmaketoolchain:generator` hasn't been set to something other than `Visual Studio 20xx` - if it has, you should prefer the user specified value instead. This could be something like

```python
conf_generator = self.conf.get("tools.cmake.cmaketoolchain:generator")
ninja_conditions = self.settings.os == "Windows" and self.options.build_lapack and (not conf_generator or "Visual Studio" in conf_generator):
generator = "Ninja" if ninja_conditions else None
if ninja_conditions:
    self.output.warning("The Visual Studio generator is not compatible with OpenBLAS when building LAPACK. Overwriting generator to 'Ninja'")
tc = CMakeToolchain(self, generator=generator)
```

joakimono: I have applied these suggested changes in the latest revision.

samuel-emrys (approved):

samuel-emrys:
@uilianries @RubenRBS could either of you review please? Would be good to get this in to unblock other math packages requiring a conan 2.0 uplift.

jiaoew1991:
Any progress on this PR that I've been following for a few weeks and hasn't been merged yet?

recipes/openblas/all/conanfile.py
RubenRBS: This has been scheduled for discussion with the team, thanks a lot for your patience

RubenRBS (commented):
Hi! Thanks a lot for your contribution and patience, we really appreciate both :)
As @samuel-emrys mentioned, important changes to big recipes usually take a bit more time because their review process is inherently slower, we need to check more things! We have now started to move forward with this, and for me most changes look ok.
I've also scheduled the discussion about default options to our next review team meeting, I'll let you know once we come to a conclusion

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,219 @@ class OpenblasConan(ConanFile):
...
+        return None
 
     @property
-    def _build_subfolder(self):
-        return "build_subfolder"
+    def _fortran_compiler(self):
+        comp_exe = self.conf.get("tools.build:compiler_executables")
+        if comp_exe and 'fortran' in comp_exe:
+            return comp_exe["fortran"]
+        return environ.get('FC')
```
uilianries, line 74:
```suggestion
        return None
```
I would keep only conan conf support for safety. So when people report some error, they need to share their profile too. Plus, env vars are easily manipulated and may cause a mess, and that's the influence of adopting conf in Conan 2.x

uilianries (commented):
Thank you for your effort, I see a lot of work here and this new release seems be offering some new support.

recipes/openblas/all/conanfile.py
```diff
@@ -19,134 +23,219 @@ class OpenblasConan(ConanFile):
...
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def package_id(self):
+        if self.info.options.build_lapack:
+            # Capture fortran compiler in package id (if found)
+            f_compiler = self._fortran_compiler
+            if not f_compiler:
```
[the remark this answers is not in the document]
memsharded: Being overly strict with the fortran compiler version added fully to the ``package_id`` will easily result in "missing binary" for all users, all the time, because of minimal variations in their locally installed ``gfortran`` version, that might not even be used for building. So probably we don't want to auto detect the version here with ``compiler --version``, and if that is the case, it should be taken from the declarative Conan ``settings`` or ``conf``

recipes/openblas/all/conanfile.py
joakimono: Added suggestion in commit

recipes/openblas/all/conanfile.py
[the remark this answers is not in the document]
joakimono: With my latest suggestion, the `package_id` option `build_lapack` is changed to include Fortran compiler id, as well as major and minor version. This is done using a CMake helper script, which is run in package_id(), with its limitations at version 3.15 (not all Fortran compiler ids are detected correctly, e.g. Classic Flang or Intel LLVM). Generally, it is laborious to properly map Fortran ABI compatibility into the `package_id`, but at least removing `patch` might reduce the occurrence of "missing binary". Ideally CMake >=3.24 should be used, but I believe build_requirements are evaluated later and not available in `package_id()`.

Since OpenBLAS 0.3.21, Fortran is not required. In this case, if the user do not specify `fortran` in `tools.build:compiler_executable`, it is built without Fortran and therefore the `package_id` is not updated with any Fortran information.

Note that this build option is not enabled on CCI (except perhaps since 0.3.23).

I admit that this way of doing it is rather cumbersome and are open for better suggestions. I have tried to avoid using a custom `settings.yml`, because of the simplicity of not needing to distribute custom `settings.yml` files.

RubenRBS:
Let me first re-trigger the build, as the unexpected errors in the v1 pipeline should now be resolved, and let's check from there :)

jiaoew1991:
I am very happy to see that all the checks have passed. I have been following for a long time, and I am really looking forward to this PR being merged as soon as possible.

samuel-emrys (approved):

samuel-emrys:
@uilianries @RubenRBS  looks like the ci on this one has been down for a fair while - any chance of a nudge? Looks like it's ready for re-review as well. Would be good to get this one in.

recipes/openblas/all/conanfile.py
joakimono: Any update on this topic? I'm willing to forego setting `"build_lapack": True` for >=0.3.21, alternatively, build both `[True, False]` (if possible) if this could expedite the merging of this pull request.

danimtb:
Hi!
Thank you for your patience and contributions to improve the recipe.

@RubenRBS and I have been taking a closer look at this and came up with the following ideas:
- Changing the default option "build_lapack" is something that should be avoided, as this is changing the expectations for consumers.
- Adding the Fortran dependency, even the Fortran runtime, adds a lot of complexity and the compiler is something that is not covered on CCI so it will be difficult to test and maintain.
- Changes in the recipe are so big and affect the old versions of the library with too many patches that make it difficult to review and maintain.

Our suggestions go more along the line:
- Creating a different conanfile.py for the new changes to avoid dealing with the different quirks of older versions.
- As commented above, since version 0.3.21 the requirement for Fortran is not needed, so this new conanfile should support 0.3.21 onwards only to make things easier.

We hope that the suggestions make sense and thanks again for your patience.

danimtb (commented):
Hi!
Thank you for your patience and contributions to improve the recipe.

@RubenRBS and I have been taking a closer look at this and came up with the following ideas:
- Changing the default option "build_lapack" is something that should be avoided, as this is changing the expectations for consumers.
- Adding the Fortran dependency, even the Fortran runtime, adds a lot of complexity and the compiler is something that is not covered on CCI so it will be difficult to test and maintain.
- Changes in the recipe are so big and affect the old versions of the library with too many patches that make it difficult to review and maintain.

Our suggestions go more along the line:
- Creating a different conanfile.py for the new changes to avoid dealing with the different quirks of older versions.
- As commented above, since version 0.3.21 the requirement for Fortran is not needed, so this new conanfile should support 0.3.21 onwards only to make things easier.

We hope that the suggestions make sense and thanks again for your patience.

RubenRBS:
Again, thanks for your patience, I'd like to mention that in the future, it's usually faster to stagger PR changes into multiple PRs instead of a big one like this, as it takes us less time to understand them if they their size is smaller :)

joakimono:
@danimtb, @RubenRBS : Should I apply the suggested simplifications here, or create a separate PR?

samuel-emrys:
> @danimtb, @RubenRBS : Should I apply the suggested simplifications here, or create a separate PR?

@joakimono given the response times for the conan team recently, my suggestion would be:

* Keep this PR with the conan 2 migration and version bump, but remove the fortran improvements you've made
* Open a new PR with the fortran improvements for a new recipe

For what it's worth this decision doesn't make a lot of sense to me - this isn't a fork of openblas, it just enables an alternate configuration. The only real change as far as the user is concerned is the change in default parameter (which doesn't _have_ to be the case). I think that creating a new recipe will be more confusing to users rather than less, especially given the inability to document what recipes are for or how/when they should be used.

My personal opinion is that the benefit provided by changing the default parameter in the latest version far outweighs the downside of changing a default parameter. Still, I think it's pretty important to get a conan 2 compatible version of openblas in ASAP, so my suggestion would be to proceed as above and we can defer a more detailed conversation about where fortran compatibility belongs to a separate PR.

danimtb:
@samuel-emrys thanks for your suggestions above and the general intention to move this PR forward.

Just for clarification, I am not suggesting creating a new **reference** with the fortran improvements, but instead, keep them under the same `openblas` references but using a different conanfile.py.

I know that could lead to having to maintain 2 conanfiles.py, but I think the chanches since 0.3.21 are significantly important to start with a new recipe and it will ease the maintenance and review of it. Just how I see it.

samuel-emrys:
> Just for clarification, I am not suggesting creating a new **reference** with the fortran improvements, but instead, keep them under the same `openblas` references but using a different conanfile.py.

Ah, understood. I misinterpreted your original statement to mean the creation of a new `openblas-fortran` or similar.

PR #17177, opened 2023-04-23
ktx: conan v2 support

SpaceIm:

---

### DIFF
--- a/recipes/ktx/all/patches/0004-unvendor-dependencies.patch
+++ b/recipes/ktx/all/patches/0004-unvendor-dependencies.patch
@@ -1,0 +1,48 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -121,7 +121,6 @@ set(KTX_MAIN_SRC
+     lib/basisu/transcoder/basisu_transcoder.cpp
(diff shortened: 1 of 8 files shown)


jwillikers (approved):

recipes/ktx/all/patches/0004-unvendor-dependencies.patch
RubenRBS, line 1: I see this patch is new. This the dependencies just not work in the past?

RubenRBS (commented):
Thanks a lot!

recipes/ktx/all/patches/0004-unvendor-dependencies.patch
SpaceIm: It worked but it was relying on `link_libraries(${CONAN_LIBS})` in CMakeLists wrapper, which is specific to legacy `cmake` generator, so it must be addressed differently when you use `CMakeToolchain`. I've also moved some replace_in_file to this patch.

RubenRBS: Nice! Thanks for the explanation

RubenRBS (approved):

PR #17186, opened 2023-04-24
rocksdb: conan v2 support

SpaceIm:

---


mpimenov (approved):

jwillikers (approved):

recipes/rocksdb/all/test_package/conanfile.py
```diff
@@ -1,22 +1,29 @@
...
     def test(self):
-        if not tools.cross_building(self):
-            if not self.options["rocksdb"].shared:
-                bin_path = os.path.join("bin", "test_package_cpp")
-                self.run(bin_path, run_environment=True)
-
-            bin_path = os.path.join("bin", "test_package_stable_abi")
-            self.run(bin_path, run_environment=True)
+        if can_run(self):
+            self.run(f"ctest --output-on-failure -C {self.settings.build_type}", env="conanrun")
```
RubenRBS, line 29: - This is misleading for external contributors that might think that test_package should run library tests instead of package tests
- Is this change motivated just to avoid two calls to self.run? Usually logic in-recipe is better than in the CMakeLists.
- This has a downside of not printing the output if it passes, so information about the test run might be lost in our CI systems

All in all I'd say that it's better to keep the old logic of tests, they are standard Conan that everyone is used to it

RubenRBS (changes requested):
Thanks! The recipe migration looks good but the ctest thingy we've come to the conclusion that it's not wanted - could you please go back to the usual way there? Thanks!

recipes/rocksdb/all/test_package/conanfile.py
SpaceIm: > Is this change motivated just to avoid two calls to self.run? Usually logic in-recipe is better than in the CMakeLists.

The motivation is always the same when you see ctest in test package, and it's not new: conan v2 doesn't offer access to dependencies options in test(), this is why we have now several ctest in test package when availability of an executable we want to run depends on dependency option.

It has already been reported to conan maintainers at the beginning of v2 migration, and ctest trick has been introduced by @jcar87, so I'm surprised to now see these reviews about ctest.

jwillikers (approved):

RubenRBS:
Let us know if you need any help @SpaceIm :)

SpaceIm:
> Let us know if you need any help @SpaceIm :)

Yep, any help would be appreciated. Cross build works fine on my mac with rocksdb/6.29.5, I've not tested rocksdb/6.8.1 but I don't understand why it fails in c3i.

Bronek:
> > Let us know if you need any help @SpaceIm :)
> 
> Yep, any help would be appreciated. Cross build works fine on my mac with rocksdb/6.29.5, I've not tested rocksdb/6.8.1 but I don't understand why it fails in c3i.

I think this is caused by linker errors, right ?

```
rocksdb/6.8.1 (test package): Running CMake.build()
rocksdb/6.8.1 (test package): RUN: cmake --build "/Users/jenkins/w/prod-v2/bsr/cci-44d8b8dd/recipes/rocksdb/all/test_package/build/apple-clang-13-armv8-17-release" -- -j4
[ 25%] Building CXX object CMakeFiles/test_package_cpp.dir/test_package.cpp.o
[ 50%] Building CXX object CMakeFiles/test_package_stable_abi.dir/test_package_stable_abi.cpp.o
[ 75%] Linking CXX executable test_package_stable_abi
  "_rocksdb_close", referenced from:
      _main in test_package_stable_abi.cpp.o
  "_rocksdb_free", referenced from:
      _main in test_package_stable_abi.cpp.o
  "_rocksdb_open", referenced from:
      _main in test_package_stable_abi.cpp.o
  "_rocksdb_options_create", referenced from:
      _main in test_package_stable_abi.cpp.o
  "_rocksdb_options_set_create_if_missing", referenced from:
      _main in test_package_stable_abi.cpp.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [test_package_stable_abi] Error 1
make[1]: *** [CMakeFiles/test_package_stable_abi.dir/all] Error 2
[100%] Linking CXX executable test_package_cpp
[100%] Built target test_package_cpp
make: *** [all] Error 2

Recipe 'rocksdb/6.8.1 (test package)' cannot build its binary
```

SpaceIm:
Yes sure, the problem is what is the root cause and how to fix it?

Bronek:
Thank you for porting https://github.com/conan-io/conan-center-index/pull/21049 . Note, the same fix also applies to version 6.27.3 (but not 6.20.3)

RubenRBS (approved):

jwillikers (approved):

Issue #15370, opened 2023-01-19
[package] libcurl/7.87.0: libcurl will build against gsasl but fails to link with it later


lakinwecker:
### Description

I am trying to use proj/9.1.1, but it fails to link against libcurl, because libcurl has compiled against my locally install sasl. 

What follows is just conjecture on my part because I stopped looking into it after I found out that I can set the `with_curl` option to `False` for proj/9.1.1 and this will work around it.

I believe that neither of the proj nor libcurl conan packages mention or use sasl explicitly, so probably libcurl is finding it automagically and using it. However, the conan cmake files that are generated also do not know that they should be linking against sasl?

I am more than happy to run some test commands if that helps and include better output.

### Package and Environment Details

* Package Name/Version: **libcurl/7.87.0**
* Operating System+version: **Linux Arch**
* Compiler+version: **GCC 12**
* Conan version: **conan 1.57.0**
* Python version: **Python 3.10**

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=12
compiler.libcxx=libstdc++11
build_type=Release
[options]
[build_requires]
[env]

### Steps to reproduce

 conan install -r conancenter proj/9.1.1@ -pr:b=GCC -pr:h=GCC -b outdated

### Logs

<details><summary>Click to expand log</summary>

```
!-> conan install -r conancenter proj/9.1.1@ -pr:b=GCC -pr:h=GCC -b outdated
Configuration (profile_host):
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

Configuration (profile_build):
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
os_build=Linux
[options]
[build_requires]
[env]

jbig/20160605: Package is up to date
libdeflate/1.15: Package is up to date
libjpeg/9e: Package is up to date
libwebp/1.2.4: Package is up to date
nlohmann_json/3.11.2: Package is up to date
openssl/1.1.1s: Package is up to date
sqlite3/3.40.0: Package is up to date
xz_utils/5.2.5: Package is up to date
zlib/1.2.13: Package is up to date
zstd/1.5.2: Package is up to date
[...]

PR #17189, opened 2023-04-24
libbigwig: conan v2 support; bump deps

robomics:
Specify library name and version:  **libbigwig/all**

- Modernize recipe
- Bump dependencies
- Change defaults to build with CURL to be consistent with upstream

---

### DIFF
--- a/recipes/libbigwig/all/conanfile.py
+++ b/recipes/libbigwig/all/conanfile.py
@@ -88,13 +88,6 @@ class LibBigWigConan(ConanFile):
         self.cpp_info.libs = ["BigWig"]
         self.cpp_info.system_libs = ["m"]
 
-        self.cpp_info.requires = []
-        if self.options.get_safe("with_curl"):
-            self.cpp_info.requires.append("libcurl::libcurl")
-        if self.options.get_safe("with_zlibng"):
-            self.cpp_info.requires.append("zlib-ng::zlib-ng")
-        else:
-            self.cpp_info.requires.append("zlib::zlib")
 
         if not self.options.get_safe("with_curl"):
             self.cpp_info.defines = ["NOCURL"]
--- a/recipes/libbigwig/all/conanfile.py
+++ b/recipes/libbigwig/all/conanfile.py
@@ -63,11 +63,6 @@ class LibBigWigConan(ConanFile):
             if not zlib_ng.options.zlib_compat:
                 raise ConanInvalidConfiguration(f"{self.ref} requires the dependency option zlib-ng:zlib_compat=True")
 
-        if self.options.with_curl:
-            libcurl = self.dependencies["libcurl"]
-            if libcurl.options.with_imap or libcurl.options.with_pop3 or libcurl.options.with_smtp:
-                raise ConanInvalidConfiguration(f"{self.ref} requires libcurl using the follow options: -o libcurl:with_imap=False -o libcurl:with_pop3=False -o libcurl:with_smtp=False")
-
     def source(self):
         get(self, **self.conan_data["sources"][self.version], destination=self.source_folder, strip_root=True)
 
--- a/recipes/libbigwig/all/conanfile.py
+++ b/recipes/libbigwig/all/conanfile.py
@@ -42,11 +42,11 @@ class LibBigWigConan(ConanFile):
 
     def requirements(self):
         if self.options.with_curl:
-            self.requires("libcurl/7.85.0")
+            self.requires("libcurl/8.0.1")
         if self.options.with_zlibng:
-            self.requires("zlib-ng/2.0.6")
+            self.requires("zlib-ng/2.0.7")
         else:
-            self.requires("zlib/1.2.12")
+            self.requires("zlib/1.2.13")
 
     def validate(self):
         if self.info.settings.os == "Windows":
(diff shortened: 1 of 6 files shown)


robomics:
ready for review :)

recipes/libbigwig/all/conanfile.py
```diff
@@ -91,12 +83,20 @@ def package(self):
...
+        self.cpp_info.libs = ["BigWig"]
+        self.cpp_info.system_libs = ["m"]
+
+        self.cpp_info.requires = []
+        if self.options.get_safe("with_curl"):
+            self.cpp_info.requires.append("libcurl::libcurl")
+        if self.options.get_safe("with_zlibng"):
+            self.cpp_info.requires.append("zlib-ng::zlib-ng")
+        else:
+            self.cpp_info.requires.append("zlib::zlib")
```
memsharded, line 97: The ``requires`` information is only necessary when dealing with components. This package has no components, and is not requiring specific components either, which should be the default behavior already.
```suggestion
```

memsharded (approved):

recipes/libbigwig/all/conanfile.py
```diff
@@ -26,32 +28,26 @@ class LibBigWigConan(ConanFile):
...
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
 
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
         if self.options.with_curl:
-            self.requires("libcurl/7.85.0")
+            self.requires("libcurl/8.0.1", transitive_headers=True)
```
RubenRBS, line 46: Sometimes it's hard to see where the need for this trait arises, so for posterity: Needed because of https://github.com/dpryan79/libBigWig/blob/master/bigWigIO.h#L5

robomics: You're right, should've documented this with a comment in the recipe.

Thanks!

RubenRBS: That's a great way to do it yes :)

robomics: Done!

robomics:
@prince-chrismc could you please have a look at this PR?

recipes/libbigwig/all/conanfile.py
```diff
@@ -62,13 +60,9 @@ def validate(self):
             if not zlib_ng.options.zlib_compat:
                 raise ConanInvalidConfiguration(f"{self.ref} requires the dependency option zlib-ng:zlib_compat=True")
 
-        if self.options.with_curl:
-            libcurl = self.dependencies["libcurl"]
-            if libcurl.options.with_imap or libcurl.options.with_pop3 or libcurl.options.with_smtp:
-                raise ConanInvalidConfiguration(f"{self.ref} requires libcurl using the follow options: -o libcurl:with_imap=False -o libcurl:with_pop3=False -o libcurl:with_smtp=False")
```
prince-chrismc, line 68: Why is this validation no longer required?

recipes/libbigwig/all/conanfile.py
```diff
@@ -91,12 +85,13 @@ def package(self):
...
-        self.cpp_info.libs = ["BigWig"]
-        self.cpp_info.system_libs = ["m"]
         self.cpp_info.set_property("cmake_file_name", "BigWig")
         self.cpp_info.set_property("cmake_target_name", "BigWig::BigWig")
+        self.cpp_info.libs = ["BigWig"]
+        self.cpp_info.system_libs = ["m"]
+
 
-        if not self.options.with_curl:
+        if not self.options.get_safe("with_curl"):
```
prince-chrismc, line 94: Why the `get_safe`? the option is never removed 😕

recipes/libbigwig/all/test_package/CMakeLists.txt
```diff
@@ -1,8 +1,10 @@
...
+project(test_package LANGUAGES C)
+
+set(CMAKE_C_STANDARD 11)
 
 find_package(BigWig REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.c)
 target_link_libraries(${PROJECT_NAME} PRIVATE BigWig::BigWig)
-target_compile_features(${PROJECT_NAME} PRIVATE c_std_11)
+target_compile_features(${PROJECT_NAME} PRIVATE c_std_${CMAKE_C_STANDARD})
```
prince-chrismc, line 10: I would put this is the "less is more" bucket, i'd refrain from style changes like this 🙏

recipes/libbigwig/all/conanfile.py
robomics: The check you're quoting was needed to avoid a bunch of weird undefined references to `gsasl_*` when building `with_curl=True` (see [here](https://github.com/conan-io/conan-center-index/pull/12195#issuecomment-1214193211) for more details).

Because we couldn't make sense of this error, we decided to disable CURL support by default (this is in contrast to the defaults used by upstream), and to add the validation step in question (the validation step was useful because libcurl doesn't link to gsasl when those options are turned off).

Few months back a user reported the same issue [here](https://github.com/conan-io/conan-center-index/issues/15370).
The issue was promptly fixed by the almighty @SpaceIm https://github.com/conan-io/conan-center-index/pull/15373.

With the latest recipe for libcurl I am no longer able to reproduce the link error, so I believe this check is no longer needed.

prince-chrismc: Ahhh thank you so much for the links and detailed description!

prince-chrismc (approved):

SSE4 (approved):

RubenRBS (approved):

PR #17190, opened 2023-04-24
grpc: use newer openssl

jcar87:
Specify library name and version:  **openssl/all**

(DRAFT)

* grpc recipe: bump version of OpenSSL to use version range that covers OpenSSL 1.1.x and OpenSSL 3.x. This should resolve to the highest available unless overridden elsewhere. This should fix conflict issues in Conan 2.x, make it easier to continue maintaining Conan 1.x without version bumps, and users can control the exact version with lockfiles (encouraged) or overrides (discouraged)

---

### DIFF
--- a/recipes/grpc/all/conanfile.py
+++ b/recipes/grpc/all/conanfile.py
@@ -92,7 +92,7 @@ class grpcConan(ConanFile):
         else:
             self.requires("abseil/20220623.0", transitive_headers=True)
         self.requires("c-ares/1.18.1")
-        self.requires("openssl/1.1.1t")
+        self.requires("openssl/[>=1.1 <4]")
         self.requires("re2/20220601")
         self.requires("zlib/1.2.13")
         self.requires("protobuf/3.21.9", transitive_headers=True, transitive_libs=True)


recipes/grpc/all/conanfile.py
```diff
@@ -92,7 +92,7 @@ def requirements(self):
         else:
             self.requires("abseil/20220623.0", transitive_headers=True)
         self.requires("c-ares/1.18.1")
-        self.requires("openssl/1.1.1t")
+        self.requires("openssl/[>=1.1 <4]")
```
SpaceIm, line 95: I believe that conancenter should first clarify policies regarding version range:
- for conan v1 & v2?
- upper bound or not (if unknown like here)?

jcar87: I'm currently working on updating the documentation for this.

For the time being, we will only allow OpenSSL and CMake (as a `tool_requires`), as these have very well known compatibility paths and version strategies (unlikely to break on a minor or patch). We will monitor how well this works for Conan Center (both users and our CI services) before making any further decisions or allowing it more generally.

SpaceIm: Ok, it's a good one to start with.

Once stabilized, other widely used libraries with quite stable API may benefit from version range in downstream recipes:

- zlib (140 recipes)
- libjpeg (42 recipes)
- libpng (40 recipes)
- libcurl (42 recipes)
- libiconv (34 recipes)
- zstd (29 recipes)
- sqlite3 (26 recipes)
- libxml2 (26 recipes)
- libtiff (25 recipes)
- freetype (23 recipes)
- xz_utils (20 recipes)
- bzip2 (19 recipes)
- lz4 (16 recipes)
- icu (15 recipes)
- libwebp (15 recipes)

SpaceIm (approved):

RubenRBS (approved):

PR #17191, opened 2023-04-24
restinio: update expected-lite version

jwidauer:
Specify library name and version:  **restinio/>0.6.17**

This PR introduces the following changes:
* Makes it easier to modify the version of requirements based off of the restinio version
* Updates the `expected-lite` dependency version to `0.6.1`, which is the same as is used in the restinio repo

---

### DIFF
--- a/recipes/restinio/all/conanfile.py
+++ b/recipes/restinio/all/conanfile.py
@@ -39,17 +39,25 @@ class RestinioConan(ConanFile):
     def requirements(self):
         self.requires("http_parser/2.9.4")
 
-        if Version(self.version) >= "0.6.17":
-            self.requires("fmt/9.1.0")
-        elif Version(self.version) >= "0.6.16":
-            self.requires("fmt/9.0.0")
-        else:
-            self.requires("fmt/8.1.1")
+        # Set up default versions for requirements
+        requirements = {
+            "fmt": "8.1.1",
+            "expected-lite": "0.5.0",
+            "optional-lite": "3.5.0",
+            "string-view-lite": "1.6.0",
+            "variant-lite": "2.0.0"
+        }
 
-        self.requires("expected-lite/0.5.0")
-        self.requires("optional-lite/3.5.0")
-        self.requires("string-view-lite/1.6.0")
-        self.requires("variant-lite/2.0.0")
+        # Override defaults, if necessary
+        if Version(self.version) >= "0.6.16":
+            requirements["fmt"] = "9.0.0"
+            requirements["expected-lite"] = "0.6.1"
+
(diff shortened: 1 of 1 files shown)


recipes/restinio/all/conanfile.py
```diff
@@ -39,17 +39,25 @@ def layout(self):
...
+            "expected-lite": "0.5.0",
+            "optional-lite": "3.5.0",
+            "string-view-lite": "1.6.0",
+            "variant-lite": "2.0.0"
+        }
+
         if Version(self.version) >= "0.6.17":
-            self.requires("fmt/9.1.0")
+            requirements["fmt"] = "9.1.0"
+            requirements["expected-lite"] = "0.6.1"
```
prince-chrismc, line 53: Why is this conditional?

prince-chrismc (commented):
If you only change the 1 line then it will be automatically merged, we generally take the less is more approach to things and I am not convinced twice as much code is "easier"

jwidauer:
@prince-chrismc I, unfortunately, couldn't have only changed one line, since the `expected-lite` version had to be set differently for both restinio versions `>=0.6.16` and `0.6.17`, meaning I would have had to change, at least, three lines.
So I'd have to change it to something like this:

```python
if Version(self.version) >= "0.6.17":
    self.requires("fmt/9.1.0")
    self.requires("expected-lite/0.6.1")
elif Version(self.version) >= "0.6.16":
    self.requires("fmt/9.0.0")
    self.requires("expected-lite/0.6.1")
else:
    self.requires("fmt/8.1.1")
    self.requires("expected-lite/0.5.0")

self.requires("optional-lite/3.5.0")
self.requires("string-view-lite/1.6.0")
self.requires("variant-lite/2.0.0")
```

With the proposed changes, it should be possible to have only single line changes, if there is any overriding of versions required.
I would prefer changing it to this proposal, to make it easier for others in the future, but I'm happy to just duplicate the requires calls in the if statements, if that is the preferred option.
Let me know what you think.

prince-chrismc:
😕 Looking that the changes in expected-lite https://github.com/martinmoene/expected-lite/compare/v0.5.0...v0.6.3#diff-787af7e01d746e9a95f277518a787ab3dd7a34498560b55317fc787d896adbf5 from 0.5 to latest, there's no breaking changes.

Did you try just updating it and not adding any conditional logic?

prince-chrismc:
https://github.com/conan-io/conan-center-index/pull/17265 too 
https://github.com/conan-io/conan-center-index/pull/17266

prince-chrismc:
done in https://github.com/conan-io/conan-center-index/pull/17266

PR #17192, opened 2023-04-24
Docs: update policy on version ranges

jcar87:
As we continue the migration to Conan 2.0, we are going to start allowing version ranges in a limited way, starting with OpenSSL and CMake.

This should have a few advantages:
* No need to update downstream consumers when OpenSSL or CMake are updated. Note that this may still require new binaries being generated for those downstream consumers. 
* Once enough recipes are updated, there should be no version conflicts related to these
* For users that have a profile with settings for which there are no binaries provided by Conan Center, if `--build=missing` is passed, those recipes that have CMake as a tool-requires will now resolve to a single version, rather than multiple versions. This has been problematic in the past if multiple versions of CMake have to be built from source.

### DIFF
--- a/docs/adding_packages/dependencies.md
+++ b/docs/adding_packages/dependencies.md
@@ -175,7 +175,7 @@ for consumer, we do impose some limits on Conan features to provide a smoother f
 
 ### Version Ranges
  
-Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/1/versioning/version_ranges.html). With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful dependencies. Currently, these are:
+Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/1/versioning/version_ranges.html). With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful of dependencies. Currently, these are:
 * OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
 * CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts.
 
--- a/docs/adding_packages/dependencies.md
+++ b/docs/adding_packages/dependencies.md
@@ -174,18 +174,14 @@ for consumer, we do impose some limits on Conan features to provide a smoother f
 * [`python_requires`](https://docs.conan.io/1/reference/conanfile/other.html#python-requires) are not allowed.
 
 ### Version Ranges
+ 
+Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/1/versioning/version_ranges.html). With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful dependencies. Currently, these are:
+* OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
+* CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts.
 
-Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/1/versioning/version_ranges.html). However,
-in the context of ConanCenter they pose a few key challenges when being used generally to consume packages, most notably:
+Conan maintainers may introduce this for other dependencies over time.
 
-* Non-Deterministic `package-id`: With version ranges the newest compatible package may yield a different `package_id` than the one built
-  and published by ConanCenter resulting in frustrating error "no binaries found". For more context
-  see [this excellent explanation](https://github.com/conan-io/conan-center-index/pull/8831#issuecomment-1024526780).
-
-* Build Reproducibility: If consumers try to download and build the recipe at a later time, it may resolve to a different package version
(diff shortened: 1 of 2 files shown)


SSE4:
OpenSSL doesn't follow SemVer, so version ranges won't work on it. am I missing the point?

RubenRBS:
While it's true that OpenSSL 1 does not follow semver ordering, Conan version parsing falls back to lexicographical ordering when mixing numbers and letters. Which just happens to work with openssl versioning:
```py
        v1 = Version("1.1.1q")
        v2 = Version("1.1.1r")
        v3 = Version("1.1.1s")
        assert v1 < v2
        assert v1 < v3
        assert v2 < v3
```

This is one of the reasons we're implementing this step by step, to ensure there are no unexpected sideeffects :)

jcar87:
> OpenSSL doesn't follow SemVer, so version ranges won't work on it. am I missing the point?
> 
> ```
> There are some packages that do not follow semver. A popular one would be the OpenSSL package with versions as 1.0.2n. They cannot be used with version-ranges. To require such packages you always have to use explicit versions (without brackets).
> ```


This document specifies the versioning policy for OpenSSL 3.x, as well as a small summary of OpenSSL 1.x: https://www.openssl.org/policies/general/versioning-policy.html

The example you are quoting is valid for OpenSSL 1.x - something built with OpenSSL 1.1.x may not be compatible with OpenSSL 1.0 because there can be breaking changes (ABI and API). OpenSSL 1.1.1 is already near its end of life (see this announcement https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/), and this lower bound forces at least 1.1.x, with the assumption, based on the announcement, that there will never be a 1.2 version.

For OpenSSL 3.x, all new minor releases are expected to be ABI and API compatible with earlier ones, hence we believe it's safe to keep the version range as greater than 3. In fact I believe Conan's own defaults are a bit more stricter than the OpenSSL 3.x policy, and may cause more "missing" binaries than actually necessary, but we will adjust over time if necessary.

RubenRBS (approved):

uilianries (approved):

docs/adding_packages/dependencies.md
```diff
@@ -174,18 +174,14 @@ for consumer, we do impose some limits on Conan features to provide a smoother f
 * [`python_requires`](https://docs.conan.io/1/reference/conanfile/other.html#python-requires) are not allowed.
 
 ### Version Ranges
+ 
+Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/1/versioning/version_ranges.html). With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful of dependencies. Currently, these are:
+* OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
+* CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts.
```
prince-chrismc, line 180: should it say `>=` given the rest of the sentence?

SpaceIm, line 180: `cmake/[^3.x.x]` is more elegant than `cmake/[>3.XX <4]`

docs/adding_packages/dependencies.md
```diff
@@ -174,18 +174,14 @@ for consumer, we do impose some limits on Conan features to provide a smoother f
 * [`python_requires`](https://docs.conan.io/1/reference/conanfile/other.html#python-requires) are not allowed.
 
 ### Version Ranges
+ 
+Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/1/versioning/version_ranges.html). With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful of dependencies. Currently, these are:
+* OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
```
SpaceIm, line 179: conan v1 client can't resolve 1.1.x versions. See https://github.com/conan-io/conan-center-index/pull/17077#issuecomment-1520644706

docs/adding_packages/dependencies.md
RubenRBS: Good catch

SpaceIm: I also think that we should not constrain upper version of cmake. CMake is backward compatible.

jcar87: CMake itself is generally backwards compatible - but when CMake 3.0.0 was released, it broke compatibility with code that worked for CMake 2.4 or earlier - so it was a breaking release - it just so happens that it retained compatibility with 2.8 which was a long standing use.

I appreciate the version range we are proposing is perhaps too explicit and too careful, but if we were in CMake 2/3 times, a requirement on `cmake/[>=2.3]` would have caused failures when Conan started resolving to `cmake/3.0.0` upon release.  Hence we think, in absence of a explicit policy from CMake maintainers, stop at the boundary of the `major` component changing.

As for `cmake/[^3.x.x]`, we prefer to be explicit as @RubenRBS (and as @prince-chrismc  points out, it should be ` cmake/[>=3.x.x <4]`, which will be ammended).

Please note that PRs that do not adhere to these policies may not be considered for merge.

PR #17224, opened 2023-04-25
libzmq: Support for no encryption in conan v2

gummif:
Specify library name and version:  **libzmq/all**

fixes https://github.com/conan-io/conan-center-index/issues/17289

---

### DIFF
--- a/recipes/zeromq/all/conanfile.py
+++ b/recipes/zeromq/all/conanfile.py
@@ -22,7 +22,7 @@ class ZeroMQConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
-        "encryption": [None, "none", "libsodium", "tweetnacl"], # None for conan v1 compat, use "none" in conan v2 for no encryption
+        "encryption": [False, "libsodium", "tweetnacl"],
         "with_norm": [True, False],
         "poller": [None, "kqueue", "epoll", "devpoll", "pollset", "poll", "select"],
         "with_draft_api": [True, False],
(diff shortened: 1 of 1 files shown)


recipes/zeromq/all/conanfile.py
```diff
@@ -22,7 +22,7 @@ class ZeroMQConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
-        "encryption": [None, "libsodium", "tweetnacl"],
+        "encryption": [None, "none", "libsodium", "tweetnacl"], # None for conan v1 compat, use "none" in conan v2 for no encryption
```
prince-chrismc, line 25: See https://github.com/conan-io/conan-center-index/issues/17289#issuecomment-1526500000 for my rational :)

```suggestion
        "encryption": [False, "libsodium", "tweetnacl"],
```

recipes/zeromq/all/conanfile.py
```diff
@@ -71,7 +71,7 @@ def source(self):
 
     def generate(self):
         tc = CMakeToolchain(self)
-        tc.variables["ENABLE_CURVE"] = bool(self.options.encryption)
+        tc.variables["ENABLE_CURVE"] = self.options.encryption == "libsodium" or self.options.encryption == "tweetnacl"
```
prince-chrismc, line 74:
```suggestion
        tc.variables["ENABLE_CURVE"] = bool(self.options.encryption)
```

prince-chrismc (commented):
I think `False` is the best solution here, the `None` was a bit erroneous and I considered leaving the old value and having a warning saying it was deprecated but I would rather prioritize it being correct for 2.0 so this will force people to update and not have a problem migrating 🤞

recipes/zeromq/all/conanfile.py
gummif: This would break v1 I guess, why not `"encryption": [None, False, "libsodium", "tweetnacl"],` ?

uilianries: Break how? Boolean of None is False. Please, let's keep it simpler, only False is enough for this case.

uilianries (approved):
LGTM

PR #16281, opened 2023-02-24
tesseract: bump dependencies + modernize more for conan v2

SpaceIm:
This recipe is not fully v2 compatible due to usage of legacy `self.deps_cpp_info`.

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17226, opened 2023-04-25
tesseract: add 5.3.1, add package_type, update dependencies

toge:
Specify library name and version:  **tesseract/***

---


recipes/tesseract/all/conanfile.py
```diff
@@ -59,13 +59,13 @@ def layout(self):
...
     def requirements(self):
-        self.requires("leptonica/1.82.0")
+        self.requires("leptonica/1.83.0", transitive_headers=True)
         # libarchive is required for 4.x so default value is true
         if self.options.get_safe("with_libarchive", default=True):
             self.requires("libarchive/3.6.2")
         # libcurl is not required for 4.x
         if self.options.get_safe("with_libcurl", default=False):
-            self.requires("libcurl/7.86.0")
+            self.requires("libcurl/8.0.1")
```
RubenRBS, line 68: Did you test that the major version bump works properly? They are always scary :$

recipes/tesseract/all/conanfile.py
```diff
@@ -59,13 +59,13 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("leptonica/1.82.0")
+        self.requires("leptonica/1.83.0", transitive_headers=True)
```
RubenRBS, line 62: It's sometimes hard to know the reasoning behind adding this trait, so for posterity: This is needed because of https://github.com/tesseract-ocr/tesseract/blob/9422915eb777f1df063838f9a2b4f72ce3aa6e46/src/tesseract.cpp#L34 (among possibly others)

RubenRBS (commented):
Thanks. I'll try to get yours and @SpaceIm's PRs for this recipe merged together

recipes/tesseract/all/conanfile.py
SpaceIm: A .cpp file can't be responsible of public visibility of an external header. Instead, you have to look at includes inside public headers of tesseract.

SpaceIm:
Duplicate of https://github.com/conan-io/conan-center-index/pull/16281, where I've experimented some issues related to this new version of leptonica (as you can see in this PR as well).

recipes/tesseract/all/conanfile.py
SpaceIm: leptonica is not a public dependency, why is there this `transitive_headers=True`?

toge:
@SpaceIm 
Oh.
Sorry, I missed your PR.
I will close this.

recipes/tesseract/all/conanfile.py
toge: It's my fault. 
leptonica transitive_headers is required in test_package.
It is not necessary.
I removed it.

PR #17228, opened 2023-04-25
libelf: conditionally remove shared option and set package type

jcar87:
Specify library name and version:  **libelf/all**

When `libelf` is in the graph, and `*:shared=True`, the graph will not properly resolve because this check raises an `InvalidConfiguration` - this would force a consumer to explicitly set `shared=False` for `libelf` on some platforms.

If on certain platforms _only_ the static version can exist, then it's not an option - so we can better reflect this in the `configure()` method: remove the option in those contexts. 

This is currently what is blocking the generation of `glib` with `shared=True` on macOS - it  depends on `libelf`. We should make it depend on the only variant that can exist! 

---

### DIFF
--- a/recipes/libelf/all/conanfile.py
+++ b/recipes/libelf/all/conanfile.py
@@ -18,7 +18,6 @@ class LibelfConan(ConanFile):
     homepage = "https://directory.fsf.org/wiki/Libelf"
     license = "LGPL-2.0"
     topics = ("elf", "fsf", "libelf", "object-file")
-
     package_type = "library"
     settings = "os", "arch", "compiler", "build_type"
     options = {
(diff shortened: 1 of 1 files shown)


recipes/libelf/all/conanfile.py
```diff
@@ -41,7 +40,10 @@ def config_options(self):
             del self.options.fPIC
 
     def configure(self):
-        if self.options.shared:
+        if self.settings.os not in ["Linux", "FreeBSD", "Windows"]:
+            self.options.rm_safe("shared")
+            self.package_type = "static-library"
```
SpaceIm, line 45: shouldn't it be done in config_options()?

jcar87: absolutely!

RubenRBS (dismissed):

czoido (approved):

PR #17244, opened 2023-04-26
Z3: Revise the recipe to support Conan v2 and add the latest four minor releases.

0xFireWolf:
Specify library name and version:  **z3/4.9.1, 4.10.2, 4.11.2, 4.12.1**

**This PR contains the following changes:**
- Revise the recipe to make it compatible with both Conan v1 and v2.
- Fix an issue that the option `multithreaded` is passed to Z3's build system improperly.
    - The correct CMake variable is `Z3_SINGLE_THREADED` instead of `SINGLE_THREADED`.
- Remove deprecated old versions, 4.8.x and earlier.
- Remove the option `multiprecision`.
- Remove the support for MPIR and its related patches that modify the original `CMakeLists.txt` for each release.
    - Note that Z3 officially supports GMP only. Hacking the build system for each Z3 release to use MPIR instead of GMP makes it hard to maintain the recipe. Users who strongly prefer MPIR to GMP may consider to submit a proposal to Z3's development team.
- Add a new boolean option `use_gmp` whose default value, `False`, is consistent with the default value of Z3's CMake option `Z3_USE_LIB_GMP`.
- Patch `CMakeLists.txt` for each supported version to use the GMP library provided by Conan Center.
    - Note that this patch is implemented in Python and will be applied if and only if users have specified to use the GMP library (i.e., by setting the option `use_gmp=True`)
- Raise the minimum GCC version to 8.
    - Compiling Z3 using GCC 7 leads to an internal compiler error.
- Set the flag `-stdlib` explicitly when compiling Z3 using Clang on Linux to fix the linker errors related to the standard library.
- Disable the test suite "DeMorgan" because it triggers a page fault if Z3 v4.11.2 is compiled by Clang in release mode.

 **Known Issues:**
When testing the recipe with the option `use_gmp=True`, Z3 cannot find the header `gmp.h` even though Conan has generated the target `gmp::gmp` successfully. **A little help here would be much appreciated.** For your convenience, you can test the recipe using the following commands. Note that the CI pipeline won't be able to capture this issue, because by default `use_gmp` is false.

```bash
# Use Z3's internal implementation
conan create all/conanfile.py --version 4.12.1 --build missing -o "z3*:use_gmp=False"
conan create all/conanfile.py --version 4.11.2 --build missing -o "z3*:use_gmp=False"
conan create all/conanfile.py --version 4.10.2 --build missing -o "z3*:use_gmp=False"
conan create all/conanfile.py --version 4.9.1 --build missing -o "z3*:use_gmp=False"

# Use GMP's implementation: gmp.h header not found error
conan create all/conanfile.py --version 4.12.1 --build missing -o "z3*:use_gmp=True"
conan create all/conanfile.py --version 4.11.2 --build missing -o "z3*:use_gmp=True"
conan create all/conanfile.py --version 4.10.2 --build missing -o "z3*:use_gmp=True"
conan create all/conanfile.py --version 4.9.1 --build missing -o "z3*:use_gmp=True"
```

Please let me know if we can improve this recipe further. Thanks for your time.

---


recipes/z3/all/conanfile.py
```diff
@@ -21,109 +23,104 @@ class Z3Conan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "multithreaded": [True, False],
-        "multiprecision": ["internal", "gmp", "mpir"]
+        "use_gmp": [True, False]
```
SpaceIm, line 26: I've read your rational about this change, but do you know that there is set_property in CMakeDeps to avoid patching upstream CMakeLists? I think we can keep this option without any maintenance burden with some smart set_property.

0xFireWolf: I see what you means. We can pretend that the target name of the MPIR library is `GMP::GMP`. I will give a try. Thanks for the suggestions.

recipes/z3/all/conandata.yml
```diff
@@ -1,30 +1,13 @@
...
-    sha256: "6962facdcdea287c5eeb1583debe33ee23043144d0e5308344e6a8ee4503bcff"
+  "4.12.1":
+    url: "https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.12.1.tar.gz"
+    sha256: "a3735fabf00e1341adcc70394993c05fd3e2ae167a3e9bb46045e33084eb64a3"
+  "4.11.2":
+    url: "https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.11.2.tar.gz"
+    sha256: "e3a82431b95412408a9c994466fad7252135c8ed3f719c986cd75c8c5f234c7e"
   "4.10.2":
-    url: "https://github.com/Z3Prover/z3/archive/z3-4.10.2.tar.gz"
+    url: "https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.10.2.tar.gz"
```
uilianries, line 9: Please, prefer the project's release instead of Github generated tars. We had an incident in the past with checksums, and the recommendation was always keep using the project release package when possible.

SpaceIm: There is no project release of the source code. I think. `https://github.com/Z3Prover/z3/archive/refs/tags/z3-4.10.2.tar.gz` is slightly better than `https://github.com/Z3Prover/z3/archive/z3-4.10.2.tar.gz`, but both are auto generated by github anyway.

danimtb (approved):

SSE4 (approved):

PR #17279, opened 2023-04-27
cppserver: use version range for openssl & bump asio

SpaceIm:

---

### DIFF
--- a/recipes/cppserver/all/conanfile.py
+++ b/recipes/cppserver/all/conanfile.py
@@ -57,8 +57,8 @@ class CppServer(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("asio/1.24.0")
-        self.requires("openssl/1.1.1s")
+        self.requires("asio/1.27.0")
+        self.requires("openssl/[>=1.1 <4]")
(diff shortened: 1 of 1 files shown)


recipes/cppserver/all/conanfile.py
```diff
@@ -57,8 +57,8 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("asio/1.24.0")
-        self.requires("openssl/1.1.1s")
+        self.requires("asio/1.27.0")
+        self.requires("openssl/[>=1.1 <4]")
```
mzukovec, line 61: In the conan [documentation](https://docs.conan.io/1/versioning/version_ranges.html), it is stated that the version ranges for packages, which do not follow semver notation, should use explicit versions. Is it no longer true?


> There are some packages that do not follow semver. A popular one would be the OpenSSL package with versions as 1.0.2n. They cannot be used with version-ranges. To require such packages you always have to use explicit versions (without brackets).

SpaceIm: see https://github.com/conan-io/conan-center-index/pull/17192

PR #17291, opened 2023-04-28
[parallel-hashmap] Update version to 1.3.11 

greg7mdp:
Specify library name and version:  **parallel-hashmap/1.3/11**


SpaceIm:
@greg7mdp parallel-hashmap recipe is already compatible with conan v2, I don't understand title of the PR.

recipes/parallel-hashmap/all/conanfile.py
```diff
@@ -10,6 +10,7 @@
 
 class ParallelHashmapConan(ConanFile):
     name = "parallel-hashmap"
+    version = "1.3.11"
```
SpaceIm, line 13:
```suggestion
```

greg7mdp:
> @greg7mdp parallel-hashmap recipe is already compatible with conan v2, I don't understand title of the PR.

My mistake, I changed the title.

recipes/parallel-hashmap/all/conanfile.py
greg7mdp: I removed it, but when I run `conan create .` locally using conan v2, it fails if the version is not present.

PR #16736, opened 2023-03-27
arg_router: Add recipe

cmannett85:
**arg_router/1.3.0**

I'm the author of this library, it's already a part of vcpkg so it makes sense to make it available on Conan too.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).

Can't do this one because [it doesn't support Conan v2 yet](https://github.com/conan-io/hooks/issues/465).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17308, opened 2023-04-29
arg_router: add 1.4.0 + use version range for cmake + fix recipe logic regarding cppstd

SpaceIm:
follow up of https://github.com/conan-io/conan-center-index/pull/16736#discussion_r1169232368

- check min cppstd only if compiler.cppstd set in profile
- use version range for cmake
- define CMake variables in CMakeToolchain
- propagate boost headers only
- add required_conan_version
- fix test v1 package

---


SpaceIm:
@cmannett85 https://github.com/cmannett85/arg_router/blob/v1.3.0/include/arg_router/utility/terminal.hpp#L11-L13 should be fixed by explicitly define NOMINMAX and eventually WIN32_LEAN_AND_MEAN. But in general it's very bad idea to expose windows.h in public headers of a library.

cmannett85:
> @cmannett85 https://github.com/cmannett85/arg_router/blob/v1.3.0/include/arg_router/utility/terminal.hpp#L11-L13 should be fixed by explicitly define NOMINMAX and eventually WIN32_LEAN_AND_MEAN.

I'll create an issue in the upstream project for this, I can model it on how Boost manages it.

> But in general it's very bad idea to expose windows.h in public headers of a library.

Understandable but unavoidable as it's (currently) a header-only library.  The TMP compiling is very hard on the compiler so I make it [explicit in the docs](https://github.com/cmannett85/arg_router#do-not-make-the-parse-tree-type-accessible) to only define the parse tree (and therefore include `arg_router.hpp`) in a single source file.

Thanks for sorting out this recipe - I'm not a Conan user nor have I ever worked professionally on a team that uses it, so I've no experience with it.  Also the differences between the Conan documentation, v1 ConanCenter documentation, and v2 ConanCenter documentation doesn't exactly make it easy.

recipes/arg_router/all/conanfile.py
```diff
@@ -1,57 +1,88 @@
...
-    requires = "boost/1.81.0", "span-lite/0.10.3"
-
     settings = "os", "arch", "compiler", "build_type"
     package_type = "header-library"
-    generators = "CMakeDeps", "CMakeToolchain"
-    no_copy_source = True
+
+    @property
+    def _min_cppstd(self):
+        return "17"
```
cmannett85, line 25: It needs to be 20 for MSVC/VS.

recipes/arg_router/all/conanfile.py
```diff
@@ -1,57 +1,88 @@
...
+
+    @property
+    def _min_compilers_version(self):
+        return {
+            "apple-clang": "10",
+            "clang": "5",
+            "gcc": "10",
+            "msvc": "191",
+            "Visual Studio": "15",
```
cmannett85, line 34: I've only ever tested it on MSVC 19.3x and VS 17.x - I've found MSVC to have poor TMP support.

recipes/arg_router/all/test_package/CMakeLists.txt
```diff
@@ -1,9 +1,8 @@
...
-        LANGUAGES CXX)
+find_package(arg_router REQUIRED CONFIG)
 
-find_package(arg_router REQUIRED)
-
-add_executable(conan_test_project "main.cpp")
-target_link_libraries(conan_test_project PUBLIC arg_router::arg_router)
+add_executable(${PROJECT_NAME} test_package.cpp)
+target_link_libraries(${PROJECT_NAME} PRIVATE arg_router::arg_router)
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
```
cmannett85, line 8: Isn't this set by the toolchain/profile?

recipes/arg_router/all/conanfile.py
cmannett85: If you're wondering why, it's due to this MSVC bug: https://developercommunity.visualstudio.com/t/1395099
C++20 allows me to use NTTP-based static strings which works around this.

recipes/arg_router/all/test_package/CMakeLists.txt
prince-chrismc: The value is absent in Conan 1.x by defautl so it need to be set explicit for things to work in CCI :)

cmannett85:
I don't have permission to push to `SpaceIm:arg_router-fix-logic`, so I'm attaching a patch instead: [arg_router-conan-msvc-fix.patch](https://github.com/conan-io/conan-center-index/files/12330419/arg_router-conan-msvc-fix.patch)

This includes:
* Update the package version to v1.4.0
* Use a minimum of C++20 when building with MSVC, otherwise it's C++17
* Set the minimum gcc version and Boost version to that tested against the upstream CI
* Added a missing `/Zc:__cplusplus` MSVC flag

SpaceIm:
@cmannett85 you can submit a PR to my branch.

cmannett85:
I build on g++-9 as a part of [my CI pipeline](https://github.com/cmannett85/arg_router/blob/develop/.github/workflows/pr_checker.yml#L127), I suspect the failure is [down to this](https://c3i.jfrog.io/c3i/misc/logs/pr/17308/5-linux-gcc/arg_router/1.3.0//5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9-build.txt):
```
************************* WARNING: GCC OLD ABI COMPATIBILITY ***********************
 
Conan detected a GCC version > 5 but has adjusted the 'compiler.libcxx' setting to
'libstdc++' for backwards compatibility.
Your compiler is likely using the new CXX11 ABI by default (libstdc++11).
```

cmannett85:
It's working for me locally:
```
$ conan create all/conanfile.py --version 1.4.0 -pr:b=default -pr:h=default --build missing
...
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=9
os=Linux
[conf]
...
-- Build files have been written to: /home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package/build/gcc-9-x86_64-20-release

arg_router/1.4.0 (test package): Running CMake.build()
arg_router/1.4.0 (test package): RUN: cmake --build "/home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package/build/gcc-9-x86_64-20-release" -- -j8
[2/2] Linking CXX executable test_package


arg_router/1.4.0 (test package): Running test()
arg_router/1.4.0 (test package): RUN: ./test_package --help
just-cats

Prints cats!

    --help,-h    Display this help and exit
    --cat        English cat
    -猫          日本語の猫
    -🐱          Emoji cat
    --แมว        แมวไทย
    --кіт        український кіт

An example program for arg_router.
```
I haven't worked out yet what the difference between this and what the CI is running.

cmannett85:
I removed this from `conanfile.py`:
```
    self.cpp_info.defines.append("NOMINMAX")
```
Because it's not needed in v1.4.0 - but it still is in v1.3.0...

cmannett85:
I think I've worked out why it's building locally with gcc-9:
```
arg_router/1.4.0 (test package): Running CMake.configure()
arg_router/1.4.0 (test package): RUN: cmake -G "Ninja" -DCMAKE_TOOLCHAIN_FILE="/home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package/build/gcc-9-x86_64-20-release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="/home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "/home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package"
-- Using Conan toolchain: /home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package/build/gcc-9-x86_64-20-release/generators/conan_toolchain.cmake
-- Conan toolchain: C++ Standard 20 with extensions OFF
-- The CXX compiler identification is GNU 12.3.0
...
```
It's still using the latest gcc due to this in the generated `rules.ninja` file:
```
#############################################
# Rule for compiling CXX files.

rule CXX_COMPILER__test_package_Release
  depfile = $DEP_FILE
  deps = gcc
  command = /usr/bin/c++ $DEFINES $INCLUDES $FLAGS -MD -MT $out -MF $DEP_FILE -o $out -c $in
  description = Building CXX object $out
```
I switched from using Ninja to Make to see if it was specific to that, but Make is the same:
```
CMakeFiles/test_package.dir/test_package.cpp.o: CMakeFiles/test_package.dir/compiler_depend.ts
	@$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --green --progress-dir=/home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package/build/gcc-9-x86_64-20-release/CMakeFiles --progress-num=$(CMAKE_PROGRESS_1) "Building CXX object CMakeFiles/test_package.dir/test_package.cpp.o"
	/usr/bin/c++ $(CXX_DEFINES) $(CXX_INCLUDES) $(CXX_FLAGS) -MD -MT CMakeFiles/test_package.dir/test_package.cpp.o -MF CMakeFiles/test_package.dir/test_package.cpp.o.d -o CMakeFiles/test_package.dir/test_package.cpp.o -c /home/cmannett85/workspace/conan-center-index/recipes/arg_router/all/test_package/test_package.cpp
```

cmannett85:
The Windows failures are down to an MSVC bug introduced in 19.35, there's a [workaround for it](https://github.com/cmannett85/arg_router/blob/develop/include/arg_router/policy/validator.hpp#L126) in v1.4.0, but that doesn't really help us here.  The other MSVC version used (19.29) is also buggy resulting in an ICE.

Personally I'd just remove v1.3.0 from `conandata.yml`, but I suspect there's rules in place here that won't allow it.

danimtb:
@cmannett85 thanks a lot for the investigation!! That would help to move this forward.

Instead of dropping the 1.3.0, I think just raising an invalid config for msvc 193 and activating the define for conan 1.4.0 should be enough.

Issue #16051, opened 2023-02-15
[package] boost/all: can't build with Linux Cross Compiler with spaces


dornbirndevelops:
### Description

I have a Yocto SDK installed on my system which I intend to use for a cross compilation.

```bash
export CC="aarch64-poky-linux-gcc  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/usr/local/poky-x86_64/sysroots/cortexa53-crypto-poky-linux"
export CXX="aarch64-poky-linux-g++  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/usr/local/poky-x86_64/sysroots/cortexa53-crypto-poky-linux"
export CPP="aarch64-poky-linux-gcc -E  -mcpu=cortex-a53 -march=armv8-a+crc+crypto -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/usr/local/poky-x86_64/sysroots/cortexa53-crypto-poky-linux"
```

I propagate the relevant information to conan via a profile buildenv/tool_requires.

I managed to fix it locally by changing a single line in [boost/all/conanfile.py](https://github.com/conan-io/conan-center-index/blob/master/recipes/boost/all/conanfile.py#L1250):
```diff
-            contents += f" \"{cxx_fwd_slahes}\""
+            contents += f" {cxx_fwd_slahes}"
```

see logs below.

### Package and Environment Details

* Package Name/Version: **boost/1.79.0**
* Operating System+version: **Linux Ubuntu 22.04**
* Compiler+version: **GCC 11**
* Conan version: **conan 1.58.0**

### Conan profile

buildprofile
```ini
[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=11
compiler.libcxx=libstdc++11
build_type=Release
```

hostprofile
```ini
[settings]
os=Linux
arch=armv8
compiler=gcc
compiler.version=11
compiler.libcxx=libstdc++11
build_type=Release

[buildenv]
[...]

PR #17315, opened 2023-04-29
boost: Use the sysroot conf variable

jwillikers:
This allows cross-compiling the Boost recipe.
Related to issue #16051.

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -1277,9 +1277,9 @@ class BoostConan(ConanFile):
         asflags = buildenv_vars.get("ASFLAGS", "") + " "
 
         sysroot = self.conf.get("tools.build:sysroot")
-        sysroot = sysroot.replace("\\", "/") if sysroot is not None else None
-        sysroot = f'"{sysroot}"' if ' ' in sysroot else sysroot
         if sysroot and not is_msvc(self):
+            sysroot = sysroot.replace("\\", "/")
+            sysroot = f'"{sysroot}"' if ' ' in sysroot else sysroot
             cppflags += f"--sysroot={sysroot} "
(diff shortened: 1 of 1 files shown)


grafikrobot (changes requested):
Don't you need to guard the options with compiler as not all accept that option or syntax?

jwillikers:
> Don't you need to guard the options with compiler as not all accept that option or syntax?

I thought about guarding it in case the compiler is `msvc`, but then it occurred to me that no one is going to actually use the `sysroot` option with `msvc`, right? As for other compilers, I think GCC and all `clang` variants will support it just fine. I'm not sure for the rest, but I would almost assume that if someone is using the `sysroot` option, their compiler supports `--sysroot`.

jwillikers:
> > Don't you need to guard the options with compiler as not all accept that option or syntax?
> 
> I thought about guarding it in case the compiler is `msvc`, but then it occurred to me that no one is going to actually use the `sysroot` option with `msvc`, right? As for other compilers, I think GCC and all `clang` variants will support it just fine. I'm not sure for the rest, but I would almost assume that if someone is using the `sysroot` option, their compiler supports `--sysroot`.

I found the documentation for it: `tools.build:sysroot: Pass the --sysroot=<tools.build:sysroot> flag if available. (None by default)`. Hmm let me see what Conan proper does to check for that.

jwillikers:
> ### This branch is out-of-date with the base branch

There's only one check against `Visual Studio`, not even `msvc` in one place. Otherwise, the compiler is ignored in most other places, possible handled by the generator like in the case of CMake. I do need to escape the string, so I will do that.

grafikrobot (approved):

SSE4 (approved):

prince-chrismc (approved):

PR #17317, opened 2023-04-29
amqp-cpp: add 4.3.24

theirix:
Specify library name and version:  **amqp-cpp/4.3.24**

Raised requirements to C++17 for latest version.

---

### DIFF
--- a/recipes/amqp-cpp/all/conanfile.py
+++ b/recipes/amqp-cpp/all/conanfile.py
@@ -50,27 +50,36 @@ class AmqpcppConan(ConanFile):
             self.requires("openssl/[>=1.1 <4]")
 
     @property
+    def _min_cppstd(self):
+        return "11" if Version(self.version) < "4.3.20" else "17"
+
+    @property
     def _compilers_minimum_version(self):
         return {
-            "gcc": "7.4",
-            "Visual Studio": "15.7",
-            "clang": "6",
-            "apple-clang": "10",
-        }
+            "17": {
+                "gcc": "7.4",
+                "Visual Studio": "15.7",
+                "msvc": "191",
+                "clang": "6",
+                "apple-clang": "10",
+            },
+        }.get(self._min_cppstd, {})
 
     def validate(self):
-        if Version(self.version) >= "4.3.20":
-            if self.settings.compiler.get_safe("cppstd"):
-                check_min_cppstd(self, "17")
-            compiler = self.settings.compiler
-            min_version = self._compilers_minimum_version.get(str(compiler), False)
-            if min_version:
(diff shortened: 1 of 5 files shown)


recipes/amqp-cpp/all/conanfile.py
```diff
@@ -46,6 +49,29 @@ def requirements(self):
...
+                check_min_cppstd(self, "17")
+            compiler = self.settings.compiler
+            min_version = self._compilers_minimum_version.get(str(compiler), False)
+            if min_version:
+                if Version(compiler.version) < min_version:
+                    raise ConanInvalidConfiguration(
+                        f"{self.name} requires C++17, which {compiler} {compiler.version} does not support.")
+            else:
+                self.output.info(
+                    f"{self.name} requires C++17. Your compiler is unknown. Assuming it supports C++17.")
```
SpaceIm, line 73:
```suggestion
    @property
    def _min_cppstd(self):
        return "11" if Version(self.version) < "4.3.20" else "17"

    @property
    def _compilers_minimum_version(self):
        return {
            "17": {
                "gcc": "7.4",
                "Visual Studio": "15.7",
                "msvc": "191",
                "clang": "6",
                "apple-clang": "10",
            },
        }.get(self._min_cppstd, {})

    def validate(self):
        if self.settings.compiler.get_safe("cppstd"):
            check_min_cppstd(self, self._min_cppstd)

        def loose_lt_semver(v1, v2):
            lv1 = [int(v) for v in v1.split(".")]
            lv2 = [int(v) for v in v2.split(".")]
            min_length = min(len(lv1), len(lv2))
            return lv1[:min_length] < lv2[:min_length]

        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
        if minimum_version and loose_lt_semver(str(self.settings.compiler.version), minimum_version):
            raise ConanInvalidConfiguration(
                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
            )
```

I think it's neater like this, future proof in terms of maintenance, and avoid to raise for Visual Studio 2019 or gcc7.

theirix:
Great, thank you! I am wondering why such generic code hasn't yet landed in tools in 2.0. Remember an old ticket with discussion.

prince-chrismc:
> Great, thank you! I am wondering why such generic code hasn't yet landed in tools in 2.0. Remember an old ticket with discussion.

Its a limitation of c3i that we are working around, when 1,x gets dropped we can just use the value from the settings :)

SpaceIm (approved):

prince-chrismc (approved):

PR #17320, opened 2023-04-30
[armadillo] Add version 12.2.0

samuel-emrys:
Specify library name and version:  **armadillo/12.2.0**

12.2.0 saw the removal of HDF5 from the armadillo wrapper as per https://gitlab.com/conradsnicta/armadillo-code/-/issues/227. I've added it back in here for consistency with the interface of previous versions and remove the need for consumers to concern themselves with dependency management of armadillo dependencies. This represents a deviation from the way this is being managed upstream though, so happy to take guidance on whether hdf5 should still be included as a dependency in the armadillo recipe for 12.2.0.

---

### DIFF
--- a/recipes/armadillo/all/conandata.yml
+++ b/recipes/armadillo/all/conandata.yml
@@ -8,6 +8,9 @@ sources:
   "11.4.3":
     url: "https://sourceforge.net/projects/arma/files/armadillo-11.4.3.tar.xz"
     sha256: "87603263664988af41da2ca4f36205e36ea47a9281fa6cfd463115f3797a1da2"
+  "12.2.0":
+    url: "https://sourceforge.net/projects/arma/files/armadillo-12.2.0.tar.xz"
+    sha256: "b0dce042297e865add3351dad77f78c2c7638d6632f58357b015e50edcbd2186"
 
 patches:
   "10.7.0":
@@ -20,5 +23,9 @@ patches:
       patch_type: "conan"
   "11.4.3":
     - patch_file: "patches/0002-Guard-dependency-discovery-11.4.x.patch"
+      patch_description: "Add find_package statements to inject conan dependencies"
+      patch_type: "conan"
+  "12.2.0":
+    - patch_file: "patches/0003-Guard-dependency-discovery-12.2.x.patch"
       patch_description: "Add find_package statements to inject conan dependencies"
       patch_type: "conan"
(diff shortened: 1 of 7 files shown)


memsharded:
Hi @samuel-emrys 

Thanks for your contribution

> 12.2.0 saw the removal of HDF5 from the armadillo wrapper as per https://gitlab.com/conradsnicta/armadillo-code/-/issues/227. I've added it back in here for consistency with the interface of previous versions and remove the need for consumers to concern themselves with dependency management of armadillo dependencies. This represents a deviation from the way this is being managed upstream though, so happy to take guidance on whether hdf5 should still be included as a dependency in the armadillo recipe for 12.2.0.

I have read https://gitlab.com/conradsnicta/armadillo-code/-/issues/227, and I would say it would be better to try to keep aligned with the ``armadillo`` upstream. If it "breaks" the old patterns in new 12.2.0 version, it would be better that Conan users also break and are aware about this change than hiding this detail to them just for convenience. Sometimes updating to newer versions is hard and breaking, but is is not the responsibility of the package manager to solve that, but just try to bring the package in the most convenient way, and respecting as much as possible the upstream decissions.

So probably some ``if Version(self.version) >= 12.2.0:  .... hdf5...`` would be the right approach here.

samuel-emrys:
@memsharded I've implemented the changes as requested, but as you can see we run into some [funny crossbuild behaviour](https://github.com/conan-io/conan-center-index/pull/17320#issuecomment-1617849441) because hdf5 doesn't support it. I was considering adding a `if not cross_building(self)` statement into the test_package, but I realised this doesn't really solve the problem because then the package will just fail with missing symbols rather than an invalid configuration message from hdf5.

The fix then is that I need to invalidate cross build packages for armadillo to capture the invalid hdf5 package upstream - which is essentially manually recreating the dependency tree, which is in some ways redundant because this logic is already a function of conan's core behaviour. See https://github.com/conan-io/conan-center-index/pull/17320/files#diff-1f1fd087beb9f0772179e8b77fdc33f5e2c8fc3c1ea0d708fc08707bac5d2657R126-R130

Are you sure this is the method you want to use here? It seems to me that this just increases the complexity from a package management perspective as well as from a consumer perspective. It would be much simpler to just add the necessary `find_package(HDF5) target_link_libraries(armadillo HDF5::HDF5)` statements back to the CMakeLists.txt file.

memsharded (approved):

jcar87 (approved):

PR #17331, opened 2023-05-01
sdbus-cpp/all: Bump dependencies

bobrofon:
Specify library name and version:  **sdbus-cpp/all**

---

### DIFF
--- a/recipes/sdbus-cpp/all/conanfile.py
+++ b/recipes/sdbus-cpp/all/conanfile.py
@@ -55,7 +55,7 @@ class SdbusCppConan(ConanFile):
             del self.options.fPIC
 
     def requirements(self):
-        self.requires("libsystemd/251.4")
+        self.requires("libsystemd/253.3")
 
     def validate(self):
         if self.info.settings.compiler.get_safe("cppstd"):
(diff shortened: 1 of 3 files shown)


bobrofon:
That's strange. This linter error shouldn't be here
```
Error: recipes/sdbus-cpp/all/test_pkgconf/conanfile.py:9: [E9005(conan-missing-name), SdbusCppTestConan] Missing name attribute
```
because `SdbusCppTestConan` is a test package recipe, and it shouldn't contain name attribute.

prince-chrismc:
> That's strange. This linter error shouldn't be here
> 
> ```
> Error: recipes/sdbus-cpp/all/test_pkgconf/conanfile.py:9: [E9005(conan-missing-name), SdbusCppTestConan] Missing name attribute
> ```
> 
> because `SdbusCppTestConan` is a test package recipe, and it shouldn't contain name attribute.

Its because the regex for the paths do not line up with the convention used for this recipe. https://github.com/conan-io/conan-center-index/actions/runs/4849884302/jobs/8642285388#step:8:19

We are re-evaluating the linters so I will slap a ticket for this

PR #17344, opened 2023-05-01
cyclonedds-cxx: Add new recipe

wadehunk:
Specify library name and version:  **cyclonedds-cxx/0.10.4**

This recipe adds the C++ bindings to CycloneDDS. 

---


wadehunk:
Waiting for PR to be approved that was opened on 26 Apr.

recipes/cyclonedds-cxx/all/conanfile.py
```diff
@@ -0,0 +1,163 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        # Use the corresponding version of cyclonedds
+        self.requires("cyclonedds/{}".format(self.version), transitive_headers=True)
```
uilianries, line 67: Which header in cyclonedds-cxx uses cyclonedds header as public? It would be great a comment like:

`#INFO <cyclonedds-cxx/foobar.h>:95 uses <cyclonedds/bar.h>` 

So we can track it easier, as transitive_headers should be used only when a public header includes a dependency.

uilianries (commented):
Thank you for your contribution. I have question related to this recipe.

recipes/cyclonedds-cxx/all/conanfile.py
wadehunk: Documented some of the dependencies in a comment.

recipes/cyclonedds-cxx/all/conanfile.py
```diff
@@ -0,0 +1,163 @@
...
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+    short_paths = True
```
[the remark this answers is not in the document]
wadehunk: Removed

recipes/cyclonedds-cxx/all/conanfile.py
```diff
@@ -0,0 +1,163 @@
...
+        cmakelists = os.path.join(self.source_folder, "CMakeLists.txt")
+        replace_in_file(self, cmakelists,
+                        "get_target_property(cyclonedds_has_shm CycloneDDS::ddsc SHM_SUPPORT_IS_AVAILABLE)",
+                        "set(cyclonedds_has_shm {})".format(self.dependencies["cyclonedds"].options.with_shm))
+        replace_in_file(self, cmakelists,
+                        "get_target_property(cyclonedds_has_type_discovery CycloneDDS::ddsc TYPE_DISCOVERY_IS_AVAILABLE)",
+                        "set(cyclonedds_has_type_discovery {})".format(self.dependencies["cyclonedds"].options.enable_discovery))
+        replace_in_file(self, cmakelists,
+                        "get_target_property(cyclonedds_has_topic_discovery CycloneDDS::ddsc TOPIC_DISCOVERY_IS_AVAILABLE)",
+                        "set(cyclonedds_has_topic_discovery {})".format(self.dependencies["cyclonedds"].options.enable_discovery))
```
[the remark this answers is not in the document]
wadehunk: CMake options are available for these items in the upstream, but these lines in particular are used to ensure the provided options match the capabilities compiled in the `cyclonedds` c libraries.

recipes/cyclonedds-cxx/all/cmake/CycloneDDS-CXX_idlcxx.cmake
uilianries, line 1: Could you please explain why this file is really needed? Is it not possible to solve it viz CMakeDeps.

wadehunk: As I mentioned here: https://github.com/conan-io/conan-center-index/pull/17344#discussion_r1406641806, this library isn't meant for linking, but for use as a loadable plug-in by the upstream cyclonedds idlc application. CMakeDeps generates INTERFACE targets to support linking, so getting the path to the idlcxx library via generator expression `$<TARGET_FILE:CycloneDDS-CXX::idlcxx>` is not possible. The use of this generator expression is an underlying assumption in the upstream implementation.

I have made a change to package a modified version of the Generate.cmake helper function that will find the path to the library instead of using the generator expression if that is more acceptable. It is effectively the same but it requires modifying and packaging the upstream implementation.

toge (approved):

recipes/cyclonedds-cxx/all/cmake/CycloneDDS-CXX_idlcxx.cmake
ErikTempelaarVO: @wadehunk 

Apologies for reviving this old thread, but I have an issue with this part of the recipe. Maybe you can offer some guidance.
My CMake deletes the generated header, with this message:

```
Deleting primary custom command output "/workspaces/devcontainer/build-coverage/PLCnextSequenceOfBytes.hpp" because another output "/workspaces/devcontainer/build-coverage/PLCnextSequenceOfBytesOUTPUT_DIR" does not exist.
```

I think this is caused by the custom Generate.cmake that provides an OUTPUT_DIR for the idlc-generator, but this option was not yet available in 0.10.4. Maybe a version bump to 0.10.5 is enough to fix it?

wadehunk: It looks like you are correct that the cyclonedds 'idlc_generate_generic` cmake function does not support the OUTPUT_DIR option as implied by cyclonedds-cxx. 

Once a new release of cyclonedds is created, I can update the recipe to add the new version. If you need a fix before then I see two options:

1) Update the recipe using a version based on master. You could do this locally until a new version is released.
2) Comment out [this line](https://github.com/conan-io/conan-center-index/blob/59cab50aa90f922e61a746eef7c488dd788e4989/recipes/cyclonedds-cxx/all/cmake/Generate.cmake#L56) in the conan recipe. It won't fix the underlying issue with OUTPUT_DIR support for cyclonedds, but it should allow you to get past the error assuming you don't need to use the OUTPUT_DIR option.

I was able to do 2 above on the cyclonedds-cxx HelloWorld example like you described above and it seemed to work out. I'm not sure why this issue is not present in the recipe test_package as it looks very similar to the HelloWorld example.

franramirez688: Solved by https://github.com/conan-io/conan-center-index/pull/29363

PR #17369, opened 2023-05-02
[sentry-crashpad] Add version 0.6.2

MartinDelille:

---


toge:
@MartinDelille 
```
WARN: libcurl/8.1.2: requirement openssl/[>=1.1 <4] overridden by sentry-crashpad/0.5.4 to openssl/1.1.1t 
ERROR: Missing binary: libcurl/8.1.2:aad9c5b4e3219a071c23322be53a09e3621c5a07
```
It may be caused by `self.requires("openssl/1.1.1t")`.

Please use `self.requires("openssl/[>=1.1 <4]")`.

MartinDelille:
@uilianries @prince-chrismc Any idea why the dependencies are missing here ?

uilianries:
@MartinDelille I started a new internal CI build to generate libcurl again. Could you please update the openssl version to range as described by @toge ?

MartinDelille:
Oups sorry @toge I missed your comment!

uilianries:
@MartinDelille It seems like the test_package is not prepared to work with Conan 2.x. Could you please update it too?

MartinDelille:
Yes. What is the correct way to obtain an executable path as done here ? https://github.com/conan-io/conan-center-index/blob/master/recipes/sentry-crashpad/all/test_package/conanfile.py#L30

uilianries:
@MartinDelille The correct way would be by `self.dependencies` but is not accessible by Conan 1.x in `test()`. So you will need to save crashpad's path in a file, as conan test executes `build()`, `generate()` and `test()` in separate stages and doesn't preserve the context. You can use:

```python
from conan import ConanFile
from conan.tools.build import can_run
from conan.tools.files import mkdir, save, load
from conan.tools.cmake import cmake_layout, CMake
import os


    test_type = "explicit"

    def requirements(self):

    def layout(self):
        cmake_layout(self)

    def generate(self):
        handler_exe = "crashpad_handler.exe" if self.settings.os == "Windows" else "crashpad_handler"
        handler_bin_path = os.path.join(self.dependencies[self.tested_reference_str].package_folder, "bin", handler_exe)
        save(self, os.path.join(self.build_folder, "handler_bin_path"), handler_bin_path)

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build()

    def test(self):
        if can_run(self):
            test_env_dir = "test_env"
            mkdir(self, test_env_dir)
            bin_path = os.path.join(self.cpp.build.bindirs[0], "test_package")
            handler_bin_path = load(self, os.path.join(self.build_folder, "handler_bin_path"))
            self.run(f"{bin_path} {test_env_dir} {handler_bin_path}", env="conanrun")
```

uilianries (approved):
LGTM

SSE4 (approved):

PR #17372, opened 2023-05-02
gdbm: Support Conan V2

jwillikers:
Required for PR #17355.

Also, added version 1.23.

For some reason, these changes result in the preprocessor macro `YYDEBUG` not being defined to `1` in the currently packaged versions of `gdbm` when the `gdbmtool_debug` option is enabled.
I've mitigated this by defining the macro appropriately when `gdbmtool_debug` is enabled.
Note that this definition does not need to be specified manually for version 1.23, only for the previous versions.

Specify library name and version:  **gdbm/***

---


jwillikers:
@RubenRBS It looks like this is hitting the same issue as #13400. Considering the v2 test package worked earlier, the `test_v1_package` must be where the problem is.

jwillikers:
@RubenRBS Looks like I was able to sneak past the test package timeout with the latest rebuild. If you could review, it would be much appreciated. I'm trying to get this updated so that I can get V2 builds for #17355 passing. Thanks!

RubenRBS:
Sure thing! Sorry about the timeout issue, we have planned a deep investigation into its root cause this week to see if we can come up with something, as such a test should not be timing out

RubenRBS (approved):
Thanks!

SSE4 (approved):

PR #16760, opened 2023-03-29
paho-mqtt-cpp: update for conan 2.0

kuroneko:
Specify library name and version:  **paho-mqtt-cpp/1.2** (also 1.0.1 and 1.1 should their recipes be updated).

Existing recipe is not conan 2.0 compatible.  Blocker for internal projects.

This does raise an error with the hook for (KB-H055) (Both 'requires' attribute and 'requirements()' method should not be declared at same recipe) but I've not been able to identify why this assert is being triggered as there is no requires attribute set.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17387, opened 2023-05-04
paho-mqtt-cpp: deprecate ssl option & ensure to use config file generated by CMakeDeps

SpaceIm:
follow up of https://github.com/conan-io/conan-center-index/pull/16760#discussion_r1151557984

- Remove openssl from requirements. OpenSSL is not a direct dependency of paho-mqtt-cpp.
- Deprecate ssl option. See https://github.com/conan-io/conan-center-index/pull/17387#discussion_r1185862546
- Override cmake names of paho-mqtt-c to ensure that config file generated by CMakeDeps is used by CML of paho-mqtt-cpp. See https://github.com/conan-io/conan-center-index/pull/17387#discussion_r1185869481
- Remove useless patches for 1.0.1 (2 were useless before this PR, due to migration to CMakeToolchain, and the last one is not required anymore with fix just above).

---


jwillikers (dismissed):

recipes/paho-mqtt-cpp/all/conanfile.py
```diff
@@ -27,7 +26,7 @@ class PahoMqttCppConan(ConanFile):
     default_options = {
         "shared": False,
         "fPIC": True,
-        "ssl": True
+        "ssl": True,
```
prince-chrismc, line 29: I feel like it deserves a comment that this is an option but has no requirement since it only changes what is being used from the C version of the library

Its a bit unusual IMO? Thoughts?

SpaceIm: This option should be deprecated. If you look closely at source code and upstream CML, this option it's no-op for paho-mqtt-cpp library itself. It's only used to enable few tests & samples, or for their custom FindPahoMqttC.cmake to inject an interface definition of `paho-mqtt-c`

SpaceIm: I also see that paho-mqtt-cpp uses its own custom `FindPahoMqttC.cmake` defining `PahoMqttC::PahoMqttC`, so file generated CMakeDeps is ignored. Let's fix that.

jwillikers (approved):

prince-chrismc (approved):
I like this packaging 👏

PR #17399, opened 2023-05-04
termcolor: conan v2 support

SpaceIm:

---

### DIFF
--- a/recipes/termcolor/all/conanfile.py
+++ b/recipes/termcolor/all/conanfile.py
@@ -1,36 +1,49 @@
 import os
-from conans import ConanFile, CMake, tools
 
-required_conan_version = ">=1.33.0"
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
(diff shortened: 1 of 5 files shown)


recipes/termcolor/all/conanfile.py
```diff
@@ -1,36 +1,49 @@
...
+from conan.tools.files import copy, get, rmdir
+
+required_conan_version = ">=1.50.0"
+
 
 class TermcolorConan(ConanFile):
     name = "termcolor"
     description = "Termcolor is a header-only C++ library for printing colored messages to the terminal."
-    topics = ("termcolor", "terminal", "color")
+    topics = ("terminal", "color")
```
RubenRBS, line 13:
```suggestion
    topics = ("terminal", "color", "header-only")
```

RubenRBS (approved):

SpaceIm:
It doesn't really matter here since it's header only, but there might be a custom target doing stuff at build time, so I prefer to always call build and don't try to be too smart.

PR #17401, opened 2023-05-05
[config] Add values for cppstds for conan v2 pipeline

jcar87:
Service configuration:

When building packages with Conan 2.0, it will build packages for the first compatible `cppstd` in the list, that is, the first one that does not throw a `ConanInvalidConfiguration`.

This should enable building packages in Conan Center with Conan 2.0 for recipes that require a higher value of `cppstd` .

### DIFF
--- a/.c3i/config_v2.yml
+++ b/.c3i/config_v2.yml
@@ -107,11 +107,11 @@ configurations:
 
 cppstd:
   apple-clang:
-    "13": ["gnu17"]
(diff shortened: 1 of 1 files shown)


SpaceIm:
Quick question: why is there a different logic for apple-clang? I mean that for other compilers, c3i tries to build for their default C++ standard first, but not for apple-clang (default C++ standard for apple-clang is still C++98).

RubenRBS (approved):

PR #17408, opened 2023-05-05
sonic-cpp: add recipe

toge:
Specify library name and version:  **sonic-cpp/***

---

### DIFF
--- a/recipes/sonic-cpp/all/conanfile.py
+++ b/recipes/sonic-cpp/all/conanfile.py
@@ -17,14 +17,6 @@ class SonicCppConan(ConanFile):
     topics = ("json", "parser", "writer", "serializer", "deserializer", "header-only")
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
-    options = {
-        "shared": [True, False],
-        "fPIC": [True, False],
-    }
-    default_options = {
-        "shared": False,
-        "fPIC": True,
-    }
 
     @property
     def _min_cppstd(self):
--- a/recipes/sonic-cpp/all/conanfile.py
+++ b/recipes/sonic-cpp/all/conanfile.py
@@ -73,5 +73,5 @@ class SonicCppConan(ConanFile):
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
 
-        if self.settings.compiler in ["gcc", "clang"]:
+        if self.settings.compiler in ["gcc", "clang", "apple-clang"]:
             self.cpp_info.cxxflags.extend(["-mavx2", "-mpclmul"])
--- a/recipes/sonic-cpp/all/conanfile.py
+++ b/recipes/sonic-cpp/all/conanfile.py
@@ -38,10 +38,8 @@ class SonicCppConan(ConanFile):
 		}
 
     def requirements(self):
-        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
-        if (self.settings.compiler.cppstd and str(self.settings.compiler.cppstd) < "17") or \
-           (minimum_version and Version(self.settings.compiler.version) < minimum_version):
-            self.requires("string-view-lite/1.7.0", transitive_headers=True)
+        # TODO: required string-view_lite only on __cplusplus < 201703L
+        self.requires("string-view-lite/1.7.0", transitive_headers=True)
 
     def package_id(self):
         self.info.clear()
--- a/recipes/sonic-cpp/all/conanfile.py
+++ b/recipes/sonic-cpp/all/conanfile.py
@@ -1,0 +1,76 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, copy
+from conan.tools.layout import basic_layout
+from conan.tools.microsoft import is_msvc
+import os
+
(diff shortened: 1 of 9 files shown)


recipes/sonic-cpp/all/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.8)
+project(test_package)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_basic_setup(TARGETS)
+
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/../test_package/
+                 ${CMAKE_CURRENT_BINARY_DIR}/test_package/)
```
memsharded, line 8: test_v1_package can be dropped for new recipes (not necessary in this PR, but in other new ones, this can be left as-is)

memsharded (approved):
Thanks for your contribution!

recipes/sonic-cpp/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+			"gcc": "8",
+			"clang": "7",
+			"apple-clang": "12",
+		}
```
uilianries, line 38:
```suggestion
        return {
            "gcc": "8",
            "clang": "7",
            "apple-clang": "12",
        }
```

avoid mixing tab and spaces. python interpreter will not break, but will warn you because is not recommended by PEP8.

uilianries (dismissed):
Arigatou!

recipes/sonic-cpp/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "8",
+            "clang": "7",
+            "apple-clang": "12",
+        }
+
+    def requirements(self):
+        # TODO: required string-view_lite only on __cplusplus < 201703L
+        self.requires("string-view-lite/1.7.0", transitive_headers=True)
```
RubenRBS, line 42: Having this check for cppstd would mean that the library would always be tested with this dependency in CI, as it only builds the first valid version it finds, and in this case it would be 11 so the dependency would always be pulled, but I _think_ it's ok? In the general user case when using cpp >= 17 there just won't be any dependency pulled and if that fails then it's _just_ an upstream error

```suggestion
        if valid_min_cppstd(self, 17):
            self.requires("string-view-lite/1.7.0")
```

(No need to define transitive headers for `header-library` either, [see the relevant docs here])(https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-type-trait-inferring)

recipes/sonic-cpp/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, copy
+from conan.tools.layout import basic_layout
+from conan.tools.microsoft import is_msvc
+from conan.tools.scm import Version
```
RubenRBS, line 7:
```suggestion
```

RubenRBS (commented):
I'm proposing some changes to address the TODO you left in the code, as I think it'd nice to have from the begining

recipes/sonic-cpp/all/conanfile.py
```diff
@@ -0,0 +1,76 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
```
RubenRBS, line 3:
```suggestion
from conan.tools.build import check_min_cppstd, valid_min_cppstd
```

recipes/sonic-cpp/all/conanfile.py
```diff
@@ -0,0 +1,76 @@
...
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "8",
+            "clang": "7",
+            "apple-clang": "12",
+        }
+
+    def requirements(self):
+        if valid_min_cppstd(self, 17):
```
RubenRBS, line 40:
```suggestion
        if valid_max_cppstd(self, 17):
```

recipes/sonic-cpp/all/conanfile.py
```diff
@@ -0,0 +1,76 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd, valid_min_cppstd
```
RubenRBS, line 3:
```suggestion
from conan.tools.build import check_min_cppstd, valid_max_cppstd
```

RubenRBS (commented):
I'm a dummy

toge:
@RubenRBS 
There seems to be an error caused by `valid_max_cppstd`.
Should I wait for conan version up in CI environment?

RubenRBS:
Yeah, I saw that, it's weird, that function has been there since forever, maybe I misspellled the import path, will check tomorrow :)

toge:
@RubenRBS 
Thanks a lot!

RubenRBS (approved):

jcar87 (approved):

PR #17413, opened 2023-05-05
eabase: add Conan v2 support, add v2.09.12

valgur:
Specify library name and version:  **eabase/2.09.12**

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17414, opened 2023-05-05
eastl: add Conan v2 support

valgur:
Specify library name and version:  **eastl/3.18.00**

Depends on #17413 (eabase).

---


recipes/eastl/all/conanfile.py
```diff
@@ -25,92 +40,107 @@ class EastlConan(ConanFile):
...
         if self.options.shared:
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("eabase/2.09.06")
+        self.requires("eabase/2.09.12", transitive_headers=True, transitive_libs=True)
```
memsharded, line 71:
```suggestion
        self.requires("eabase/2.09.12")
```
Not necessary for ``package_type = "header-library"``

memsharded (approved):

recipes/eastl/all/conanfile.py
```diff
@@ -25,92 +40,107 @@ class EastlConan(ConanFile):
...
+            self,
+            **self.conan_data["sources"][self.version],
+            destination=self.source_folder,
+            strip_root=True,
+        )
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["EASTL_BUILD_BENCHMARK"] = False
+        tc.variables["EASTL_BUILD_TESTS"] = False
```
uilianries, line 96:
```suggestion
        tc.variables["EASTL_BUILD_TESTS"] = False
        tc.variables["CMAKE_CXX_STANDARD"] = self._min_cppstd
```

taking a look in the CMakeLists.txt from eastl project, it does not declare a required C++ standard, but headers consumes C++11 and C++14 features. A minimal standard should be declared.

recipes/eastl/all/conanfile.py
```diff
@@ -25,92 +40,107 @@ class EastlConan(ConanFile):
...
 
     @property
-    def _minimum_compilers_version(self):
+    def _compilers_minimum_version(self):
         return {
             "Visual Studio": "14",
-            "gcc": "5",
-            "clang": "3.2",
-            "apple-clang": "4.3",
+            "gcc": "6",
```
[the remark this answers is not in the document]
uilianries: @valgur Thank you for sharing the error. Please, check my next comment, the EASTL does not declare a required C++ standard, that's why is failing with GCC5. Customizing variables solves your problem, so you can add GCC 5 again.

uilianries (approved):
LGTM

recipes/eastl/all/conanfile.py
```diff
@@ -1,20 +1,35 @@
...
+        "EASTL stands for Electronic Arts Standard Template Library. "
+        "It is an extensive and robust implementation that has an "
+        "emphasis on high performance."
+    )
     topics = ("eastl", "stl", "high-performance")
     license = "BSD-3-Clause"
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/electronicarts/EASTL"
 
+    package_type = "header-library"
```
memsharded, line 32: But it doesn't seem that this library is header-only, but a regular library. Am I missing something? https://github.com/electronicarts/EASTL/blob/c6141b94d763d2f536bddf40d6abaf878f7ab7ac/CMakeLists.txt#L25

recipes/eastl/all/conanfile.py
```diff
@@ -50,67 +54,94 @@ def _minimum_compilers_version(self):
...
         if self.settings.os == "Windows":
             del self.options.fPIC
 
     def configure(self):
         if self.options.shared:
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
```
memsharded, line 68: Might require a cmake_layout to build with CMake?

recipes/eastl/all/conanfile.py
```diff
@@ -50,67 +54,94 @@ def _minimum_compilers_version(self):
...
 
     def requirements(self):
-        self.requires("eabase/2.09.06")
+        self.requires("eabase/2.09.12")
+
+    def package_id(self):
+        self.info.clear()
```
memsharded, line 74: Not a header-only, shouldn't be cleared

memsharded (changes requested):
Thanks very much for your contribution. However, I think there might be some misunderstading regarding this lib being header-only, I think it is not. I am not sure why this passes, maybe because the ``test_package`` contains ``eastl::max_element`` only, and that is implemented in the headers and doesn't require any linkage?

uilianries:
@memsharded good catch! thanks for reviewing it!

SSE4:
> max_element

@memsharded `max_element` is template function, thus it's implemented in [header](https://github.com/electronicarts/EASTL/blob/c6141b94d763d2f536bddf40d6abaf878f7ab7ac/include/EASTL/algorithm.h#L407) file (it obviously cannot be implemented in .cpp, because of template parameter). therefore, it only needs a header file and doesn't require any linkage.
now with your question answered, could you possibly unblock review?

SSE4:
@memsharded ping

recipes/eastl/all/conanfile.py
```diff
@@ -50,67 +53,90 @@ def _minimum_compilers_version(self):
...
         if self.options.shared:
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("eabase/2.09.06")
+        self.requires("eabase/2.09.12", transitive_headers=True)
```
uilianries, line 70: Okay, we will have the follow error in case is not exposed:

```
p/include/EASTL/internal/config.h:61:11: fatal error: 'EABase/eabase.h' file not found
        #include <EABase/eabase.h>
```

uilianries (approved):
LGTM

uilianries:
I just started a new CI job for Conan v2, as it failed but was an internal error.

toge (approved):

SSE4 (approved):

PR #17423, opened 2023-05-06
[onnxruntime] Set minimum supported compiler version for GCC to 7

Linux13524:
Specify library name and version:  **onnxruntime/1.14.1**

The changelog here states GCC >= 7 is currently supported: https://github.com/microsoft/onnxruntime/releases/tag/v1.13.1

I just build a package with GCC 7 and that worked fine. Also the error message was a bit missleading as I set cppstd to 17:

```
Invalid ID: onnxruntime/1.14.1 requires C++17, which your compiler does not support.
``` 

---


Linux13524:
I did my test built with a Docker image based on the conan-docker-tools modern images. There the GCC version is 7.5.0, in the pipeline however it is 7.2.0. AFAIK the CCI pipelines do not use this Docker images right? But then there would still be the question why it is not compiled with the latest GCC 7 version?

Hopobcn (approved):

czoido (approved):

jcar87 (approved):

PR #17429, opened 2023-05-06
ada: add version 2.3.1, enable shared build

toge:
Specify library name and version:  **ada/***

* Add `shared` option, as some versions support it
* Change package type to `library` (since it can now be shared or static)
* In the `configure()` method, conditionally remove the `shared` option for older versions, and set the `package_type` to `static_library` instead - note that there is a limitation in Conan Center CI, but this should be done in `config_options()` instead.
* Add version `2.3.1`

---


jcar87 (approved):
Hi @toge  thanks for your contribution.

I've amended the logic in the `configure()` method to fix the CI issue, and updated the PR description to reflect the changes

RubenRBS (approved):

PR #17439, opened 2023-05-07
bitserializer: fix conan v2 regression introduced by previous PR & new logic in 0.50

SpaceIm:
- fix regression introduced by https://github.com/conan-io/conan-center-index/pull/14933 regarding conan v2 (it's self.info.clear(), and we must rely on self.info in package_id())
- add package_type
- add fPIC option when with_csv option is enabled since it generates a static lib
- use cmake_layout in case of CMake build
- fix transitive traits of rapidyaml (it's a dependency of header-only part of this lib, like other dependencies) 
- remove with_csv & with_rapidyaml options if version < 0.50, instead of raising
- avoid replace_in_file() with strict=False, it's a bad practice

---

### DIFF
--- a/recipes/bitserializer/all/conanfile.py
+++ b/recipes/bitserializer/all/conanfile.py
@@ -1,13 +1,13 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import copy, get, rmdir, replace_in_file
 from conan.tools.layout import basic_layout
 from conan.tools.scm import Version
-from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain
 import os
 
-required_conan_version = ">=1.50.0"
+required_conan_version = ">=1.51.1"
 
 
 class BitserializerConan(ConanFile):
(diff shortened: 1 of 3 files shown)


recipes/bitserializer/all/conanfile.py
```diff
@@ -50,17 +52,29 @@ def _compilers_minimum_version(self):
...
         # All components of library are header-only except csv-archive
-        return not self.options.with_csv
+        options = self.info.options if info else self.options
+        return not options.with_csv
 
-    def _patch_sources(self):
-        # Remove 'ryml' subdirectory from #include
-        replace_in_file(self, os.path.join(self.source_folder, "include/bitserializer/rapidyaml_archive.h"), "#include <ryml/", "#include <", strict=False)
+    def config_options(self):
+        if self.settings.os == "Windows" or Version(self.version) < "0.50":
```
PavelKisliak, line 63: I think that there is need to check if current build is "header-only" like below:
```suggestion
        if self.settings.os == "Windows" or self._is_header_only():
```
BitSerializer doesn't produce any static libraries without option `with_csv`.

SpaceIm: It doesn't really matter, patch is valid regardless of with_csv option value.

SpaceIm: Good point, but you can't rely on options values in config_options, so it must be handled in configure(), I'll push a fix.

prince-chrismc (approved):

SSE4 (approved):

PR #17088, opened 2023-04-18
paho-mqtt-c: bump openssl + drop maintenance of 1.3.0 & 1.3.1

SpaceIm:
Specify library name and version:  **lib/1.0**

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17442, opened 2023-05-07
paho-mqtt-c: remove anl library from system libs

mtudan:
Remove `anl` library from default system libraries if on Linux because it is not available on non-glibc distros like Alpine. This will fix building your apps on e.g. Alpine that link with `paho-mqtt-c` library. We might use some logic to determine if `anl` is available like the upstream project does.

Specify library name and version:  **paho-mqtt-c/1.3.12**

---


prince-chrismc:
@SpaceIm  You added anl in https://github.com/conan-io/conan-center-index/pull/17088/files#diff-cefdb38dd5cec800413f548cb236f4f8cb7fd28165ef036164d59b456553cbd4R121 do you happen to recall why?

SpaceIm:
> @SpaceIm You added anl in https://github.com/conan-io/conan-center-index/pull/17088/files#diff-cefdb38dd5cec800413f548cb236f4f8cb7fd28165ef036164d59b456553cbd4R121 do you happen to recall why?

Because it was reported by hooks: https://github.com/conan-io/conan-center-index/pull/17088#issuecomment-1513505200

See also https://github.com/eclipse/paho.mqtt.c/blob/7db21329301b1f527c925dff789442db3ca3c1e7/src/CMakeLists.txt#L67-L70, but yes it's optional. There is not enough granularity in conan profiles to predict if anl is required as a system libs or not.

prince-chrismc:
Cheers! Thanks for helping :)

mtudan:
@prince-chrismc @franramirez688 @jcar87

Any news with this PR? That's quite old right now.

uilianries:
It's possible to customize the settings, adding distro: https://docs.conan.io/2/reference/config_files/settings.html#adding-new-settings

It could be used for this case, having a condition to not add `anl` in case distro Alpine, for instance.

PR #16392, opened 2023-03-05
Wg21 linear algebra conan 2

Twon:
Specify library name and version:  **wg21-linear_algebra/0.7.2**

Updating this package for Conan 2.0 support.  I am a maintainer of this library.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17456, opened 2023-05-08
Remove output.warn, no longer supported in Conan 2.0

Twon:
Specify library name and version:  **wg21-linear_algebra/0.7.3**

As a maintainer of this library, I am submitting this fix to address the final comment of this review, which was received post-merge: https://github.com/conan-io/conan-center-index/pull/16392
---

### DIFF
--- a/recipes/wg21-linear_algebra/all/conanfile.py
+++ b/recipes/wg21-linear_algebra/all/conanfile.py
@@ -42,8 +42,8 @@ class LAConan(ConanFile):
             check_min_cppstd(self, self._minimum_cpp_standard)
         min_version = self._minimum_compilers_version.get(str(compiler))
         if not min_version:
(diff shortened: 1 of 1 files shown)


RubenRBS (commented):
This moves from assuming that the compiler supports the version to _not_. Was this intended?

If not, after checking what other recipes do, I think collapsing the if to `if min_version and Version(self.settings....` to keep the previous behaviour, which is what most other recipes do :)

Twon:
> This moves from assuming that the compiler supports the version to _not_. Was this intended?
> 
> If not, after checking what other recipes do, I think collapsing the if to `if min_version and Version(self.settings....` to keep the previous behaviour, which is what most other recipes do :)

Thanks for the suggestion, that sounds more sensible.

RubenRBS (approved):

jwillikers (approved):

PR #17457, opened 2023-05-08
libtool: Fix copying config_guess and config_sub

### LINKED ISSUE #16740: [libtool] libtool/2.4.7 2 newest revision does not build on Debian Buster gcc8
### Description

Hi,

I tried to build the latest revision of libtool and it failed during the build step.

Here is my profile:

Configuration for profile default:

[settings]
os=Linux
os.version=debian.buster
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=8
compiler.libcxx=libstdc++11
build_type=Release
[options]
[conf]
[build_requires]
[env]

I tried the 2 latest revision without any success but it worked with the first revision of the recipe: **0df56d1f5ae0b5c5c1bbb818c88247a8**

* Package Name/Version: **libtool/2.4.7**, revision: **d55a3c2435c19e830fa77aa5c6c19350** and **a2ac067b5916a1fa808bd6f94f69d95c**
* Operating System+version: **Linux Debian Buster 10.13**
* Compiler+version: **GCC 8**
* Conan version: **conan 1.58.0**
* Python version: **Python 3.11.1**

### Conan profile

[settings]
os=Macos
os_build=Macos
arch=armv8
arch_build=armv8
compiler=apple-clang
compiler.version=14
compiler.libcxx=libc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install libtool/2.4.7@ --build=missing -r conancenter

### Logs

```
ERROR: libtool/2.4.7: Error in build() method, line 126
        self._patch_sources()
while calling '_patch_sources', line 122
        shutil.copy(config_sub, os.path.join(self.source_folder, "build-aux", "config.sub"))
        TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

</details>

jwillikers:
Specify library name and version:  **libtool/***

~~The `libtool` recipe currently  fails to build due to an error when copying the config_guess and config_sub files from gnu-config. This fixes that error.~~

~~Fixes #16740.~~

This is a slight improvement to how the `libtool` recipe copies the `config_guess` and `config_sub` files, though it is not necessary.
The issue I thought this fixed was actually corrected by using a newer revision of `libtool`.

---


SpaceIm:
I'm not against this pattern, but previous implementation is also correct. I think https://github.com/conan-io/conan-center-index/issues/16740 comes from a deprecated RREV of gnu-config.

jcar87:
Hi @jwillikers, can you confirm if this is still a problem even after updating `gnu-config` to a more recent recipe revision?

jwillikers:
> Hi @jwillikers, can you confirm if this is still a problem even after updating `gnu-config` to a more recent recipe revision?

@jcar87 Updating the `libtool` recipe revision was sufficient to fix my problem. I've left this PR open in case it is a desired improvement, since it uses the internal Conan `copy` method instead of `shutil`'s `copy` function.

RubenRBS (approved):
Approving so your work into fixinf the issue does not go _completely_ wasted, we really appreciate the time you took to debug and fix the original issue :)

jcar87:
Thanks @jwillikers  - 

Before going ahead with the merge I'm currently investigating a few things.
* By silently doing nothing if `gnu_config` is null, that is, if the `self.conf.get()` returns none, it falls back to using `config.sub` and `config.guess` as provided by libtool - I'm assuming (and this could be wrong as well), that if they were added to the recipe it was for a valid reason. I'd like to avoid a situation where silently doing nothing could result in different behaviours for different users. The error when doing the shutil.copy() is unfortunate but at least it is predictable - we could consider making it an error if the conf is not provided directing the users to update the gnu-config recipe revision to address the problem.

* I have noticed that the original sources of `libtool/2.4.7` contain copies of `config.sub` and `config.guess` that are _newer_ than the version of `gnu-config` this recipe - so for the most recent libtool we are actuall copying an older pair of scripts, perhaps unnecessarily.

jwillikers:
@jcar87 Thanks for looking into that! Just let me know if I should go ahead and close this PR or require any changes.

SSE4 (approved):

PR #17465, opened 2023-05-09
Add vir-simd library

AlvaroFS:
**vir-simd/0.2.0**

I'd like to add the vir-simd library which provides a fallback implementation for std::experimental::simd, since I think it will be useful.

---

### DIFF
--- a/recipes/vir-simd/all/conanfile.py
+++ b/recipes/vir-simd/all/conanfile.py
@@ -12,8 +12,8 @@ required_conan_version = ">=1.52.0"
 
 class VirSIMDConan(ConanFile):
     name = "vir-simd"
-    description = "This project aims to provide a simple fallback for users of std::experimental::simd (Parallelism TS 2)"
-    license = "LGPL-3.0"
+    description = "A fallback std::experimental::simd (Parallelism TS 2) implementation with additional features"
+    license = "LGPL-3.0-or-later"
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/mattkretz/vir-simd"
     topics = ("simd", "parallelism-ts", "cpp17", "header-only")
@@ -50,6 +50,9 @@ class VirSIMDConan(ConanFile):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
+
+        if self.settings.os == "Windows":
+            raise ConanInvalidConfiguration(f"{self.ref} Windows support is not provided so far.")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
--- a/recipes/vir-simd/all/conanfile.py
+++ b/recipes/vir-simd/all/conanfile.py
@@ -1,0 +1,86 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, copy
+from conan.tools.layout import basic_layout
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.52.0"
+
+
+class VirSIMDConan(ConanFile):
+    name = "vir-simd"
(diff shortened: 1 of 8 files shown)


recipes/vir-simd/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+
+
+required_conan_version = ">=1.52.0"
+
+
+class VirSIMDConan(ConanFile):
+    name = "vir-simd"
+    description = "This project aims to provide a simple fallback for users of std::experimental::simd (Parallelism TS 2)"
+    license = "LGPL-3.0"
```
mattkretz, line 16: `LGPL-3.0-or-later` to be precise

recipes/vir-simd/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.52.0"
+
+
+class VirSIMDConan(ConanFile):
+    name = "vir-simd"
+    description = "This project aims to provide a simple fallback for users of std::experimental::simd (Parallelism TS 2)"
```
mattkretz, line 15: How about "A fallback std::experimental::simd (Parallelism TS 2) implementation with additional features"

mattkretz (commented):
I have no experience with packaging for Conan so I can't review that part. But if I understand correctly, then no compilation tests of the vir-simd package are run, right? The complete test-suite is copied from the libstdc++ implementation and not suitable to run in general. It will thrash the average machine. But the simple `vir/test.cpp` could be used to do some sanity checks. Though the `Makefile` `check-extensions` target isn't ready for non-GNU compilers...

recipes/vir-simd/all/conanfile.py
AlvaroFS: Yes, much better. I just copied the first line from the repository.

recipes/vir-simd/all/conanfile.py
```diff
@@ -0,0 +1,89 @@
...
+        # examples: bzip2, freetype, gdal, icu, libcurl, libjpeg, libpng, libtiff, openssl, sqlite3, zlib...
+        self.cpp_info.set_property("cmake_module_file_name", "vir-simd")
+        self.cpp_info.set_property("cmake_module_target_name", "vir-simd::vir-simd")
+        # Set these to the appropriate values if package provides a CMake config file
+        # (package-config.cmake or packageConfig.cmake, with package::package target, usually installed in <prefix>/lib/cmake/<package>/)
+        self.cpp_info.set_property("cmake_file_name", "vir-simd")
+        self.cpp_info.set_property("cmake_target_name", "vir-simd::vir-simd")
+        # Set this to the appropriate value if the package provides a pkgconfig file
+        # (package.pc, usually installed in <prefix>/lib/pkgconfig/)
+        self.cpp_info.set_property("pkg_config_name", "vir-simd")
```
memsharded, line 89:
```suggestion
```
These are not really necessary, only necessary when the names deviate from the package name. But if it is the same, better avoid the noise

recipes/vir-simd/all/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+project(test_package LANGUAGES CXX)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_basic_setup(TARGETS)
+
```
memsharded, line 6: This can be removed, conan-center-index no longer require ``test_v1_package`` (it doesn't harm to have them there, you can leave it, just know it is no longer a requirement)

memsharded (approved):
Looking good, thanks very much for your contribution!

SSE4 (approved):

recipes/vir-simd/all/test_v1_package/CMakeLists.txt
AlvaroFS: Sorry I've just seen your comments and it's already merged. I'll keep them in mind for the next update.

PR #16678, opened 2023-03-23
openvc: 4.x modernize more for conan v2

prince-chrismc:
Specify library name and version:  **lib/1.0**

Building this locally I ran into a version conflict

```
Graph root
    cli
Requirements
    imath/3.1.6#7f6b123c01e44d29f5170f2141083e44 - Downloaded (conancenter)
    jasper/4.0.0#478d874438e50163be99725c8603281a - Downloaded (conancenter)
    jbig/20160605#2d29fa02aacd76902e0d2cbbc24631ef - Downloaded (conancenter)
    libdeflate/1.17#71f492ef993b45d9b961b36fb85e801f - Downloaded (conancenter)
    libjpeg/9e#68269859e4325ddc3f995d1fd3fc9187 - Downloaded (conancenter)
    libpng/1.6.39#deca555b0890e584fc895ba756e83c06 - Downloaded (conancenter)
    libtiff/4.4.0#e6d8942022c89a22dfbe1a9b77289c7e - Downloaded (conancenter)
    opencv/4.5.5#107c7183c36c63faeb1512659ccd6c70 - Cache
    openexr/3.1.5#88ba2322e127ebe5635fcbfdff4928d2 - Downloaded (conancenter)
    xz_utils/5.4.0#186e94237b66f785e4ea095f2f621a56 - Downloaded (conancenter)
    zlib/1.2.13#13c96f538b52e1600c40b88994de240f - Downloaded (conancenter)
    zstd/1.5.4#15ac0bc440f89247276d78ea4fdf8b14 - Downloaded (conancenter)
ERROR: Version conflict: libtiff/4.4.0->libwebp/1.3.0, opencv/4.5.5->libwebp/1.2.4.
```

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17475, opened 2023-05-09
Bump eigen in opencv

czoido:
Bump eigen to 3.4.0 to avoid doing overrides

### DIFF
--- a/recipes/opencv/4.x/conanfile.py
+++ b/recipes/opencv/4.x/conanfile.py
@@ -198,7 +198,7 @@ class OpenCVConan(ConanFile):
         if self.options.get_safe("with_tiff"):
             self.requires("libtiff/4.4.0")
         if self.options.with_eigen:
-            self.requires("eigen/3.3.9")
+            self.requires("eigen/3.4.0")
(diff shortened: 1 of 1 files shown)


czoido:
> AFAIK, OpenCV 3.x & 4.x fail to compile against eigen 3.4.0 with several Visual Studio versions. I've already tried last year. See #8868.

Thanks for the reference, let's see what's the build result and if any 4.x version is affected

SpaceIm:
Unexpected error in v2 pipeline. It has been merged before c3i had the time to add opencv to v2 ready references after https://github.com/conan-io/conan-center-index/pull/16678.

czoido:
> Unexpected error in v2 pipeline. It has been merged before c3i had the time to add opencv to v2 ready references after #16678.

Yes, we noticed, thanks for reporting. We have been regenerating packages for OpenCV and should be available soon.

db4:
Hi @czoido,
In the light of https://github.com/conan-io/conan-center-index/issues/18077 don't you think that this PR should be reverted?

Issue #17384, opened 2023-05-04
[package] dbus/1.15.2: options system_socket and system_pid_file are not passed to the build system


Knapock:
### Description

The options system_socket and system_pid_file are defined in the recipe but they are ignored.
Adding the following lines in the generate() method fixes it (for meson build):
            tc.project_options["system_socket"] = str(self.options.system_socket)
            tc.project_options["system_pid_file"] = str(self.options.system_pid_file)

### Package and Environment Details

* Package Name/Version: **dbus/1.15.2**
* Conan version: **conan 1.59.0**

### Conan profile

any

### Steps to reproduce

Set the default values for system_socket and system_pid_file.

Build the recipe.

The set values are ignored, leading to the error "Failed to initialize connection to D-Bus: Socket name too long" when using the built package

### Logs

Failed to initialize connection to D-Bus: Socket name too long

Knapock:
The default values for these options could also be updated.
Currently they include the install prefix, which is very long and leads to the "Socket name too long" error.
Using the following default would also mean the conna-installed dbus could communicate with the system dbus:
        "system_socket": "/run/dbus/system_bus_socket",
        "system_pid_file": "/run/dbus/pid",

PR #17491, opened 2023-05-10
Dbus/add options socket and pid file

Knapock:
**dbus/1.15.2**

This PR aims to fix https://github.com/conan-io/conan-center-index/issues/17384

---

### DIFF
--- a/recipes/dbus/1.x.x/conanfile.py
+++ b/recipes/dbus/1.x.x/conanfile.py
@@ -25,8 +25,8 @@ class DbusConan(ConanFile):
     settings = "os", "arch", "compiler", "build_type"
     short_paths = True
     options = {
-        "system_socket": ["ANY"],
-        "system_pid_file": ["ANY"],
+        "system_socket": [None, "ANY"],
+        "system_pid_file": [None, "ANY"],
         "with_x11": [True, False],
         "with_glib": [True, False],
         "with_systemd": [True, False],
@@ -34,8 +34,8 @@ class DbusConan(ConanFile):
         "session_socket_dir": ["ANY"],
     }
     default_options = {
-        "system_socket": "",
-        "system_pid_file": "",
+        "system_socket": None,
+        "system_pid_file": None,
         "with_x11": False,
         "with_glib": False,
         "with_systemd": False,
@@ -104,7 +104,9 @@ class DbusConan(ConanFile):
             tc.project_options["checks"] = False
             tc.project_options["doxygen_docs"] = "disabled"
             tc.project_options["modular_tests"] = "disabled"
-            tc.project_options["session_socket_dir"] = str(self.options.session_socket_dir)
+            tc.project_options["system_socket"] = str(self.options.get_safe("system_socket", ""))
+            tc.project_options["system_pid_file"] = str(self.options.get_safe("system_pid_file", ""))
+            tc.project_options["session_socket_dir"] = str(self.options.get_safe("session_socket_dir", ""))
             tc.project_options["selinux"] = "enabled" if self.options.get_safe("with_selinux", False) else "disabled"
             tc.project_options["systemd"] = "enabled" if self.options.get_safe("with_systemd", False) else "disabled"
             if self.options.get_safe("with_systemd", False):
@@ -127,13 +129,14 @@ class DbusConan(ConanFile):
             tc.variables["DBUS_WITH_GLIB"] = bool(self.options.get_safe("with_glib", False))
             tc.variables["DBUS_DISABLE_ASSERT"] = is_apple_os(self)
             tc.variables["DBUS_DISABLE_CHECKS"] = False
+            tc.variables["DBUS_SYSTEM_BUS_DEFAULT_ADDRESS"] = str(self.options.get_safe("system_socket", ""))
 
             # Conan does not provide an EXPAT_LIBRARIES CMake variable for the Expat library.
             # Define EXPAT_LIBRARIES to be the expat::expat target provided by Conan to fix linking.
             tc.variables["EXPAT_LIBRARIES"] = "expat::expat"
 
             # https://github.com/freedesktop/dbus/commit/e827309976cab94c806fda20013915f1db2d4f5a
-            tc.variables["DBUS_SESSION_SOCKET_DIR"] = str(self.options.session_socket_dir)
+            tc.variables["DBUS_SESSION_SOCKET_DIR"] = str(self.options.get_safe("session_socket_dir", ""))
 
             tc.cache_variables["CMAKE_FIND_PACKAGE_PREFER_CONFIG"] = False
             tc.generate()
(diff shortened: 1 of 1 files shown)


recipes/dbus/1.x.x/conanfile.py
```diff
@@ -104,6 +104,8 @@ def generate(self):
             tc.project_options["checks"] = False
             tc.project_options["doxygen_docs"] = "disabled"
             tc.project_options["modular_tests"] = "disabled"
+            tc.project_options["system_socket"] = self.options.get_safe("system_socket", "")
+            tc.project_options["system_pid_file"] = self.options.get_safe("system_pid_file", "")
```
jwillikers, line 108: Can you add the equivalent options for CMake below?

jwillikers (changes requested):
Thanks for the PR here. It fixes an important oversight where those options aren't taken into consideration.

Try bumping the Meson version to `meson/1.1.0`. That might fix the build failure. Also, bump glib to `glib/2.76.2`, while your at it so that we've got everything up-to-date.

These options should probably be configured to allow a `None` value and use that as the default.

```python
 options = {
        "system_socket": [None, "ANY"],
        "system_pid_file": [None, "ANY"],
        "with_x11": [True, False],
        "with_glib": [True, False],
        "with_systemd": [True, False],
        "with_selinux": [True, False],
        "session_socket_dir": ["ANY"],
    }
    default_options = {
        "system_socket": None,
        "system_pid_file": None,
        "with_x11": False,
        "with_glib": False,
        "with_systemd": False,
        "with_selinux": False,
        "session_socket_dir": "/tmp",
}
```

This should be clearer to consumers on what is going on.

If you can add the equivalent options for when CMake is in use, that would be great.

Finally, I read your issue report. Thanks for the details. I looked into it a little bit, and I think we will need to end up doing something about the default path to the socket being to long when it is in the cache. I'm not sure the best way to do that yet as it seems mostly related to deployment and there are potentially security concerns as well as the potential for multiple versions of `dbus` in the cache to conflict. At least, I think. I need to look a bit more into the exact behavior. Anyways, using these options is great for now, then at least consumers can override the values as needed. I also wonder if it would be possible to increase the allowed size of the path to the socket in `dbus` upstream.

recipes/dbus/1.x.x/conanfile.py
Knapock: Done for "system_socket", I couldn't find the cmake option for "system_pid_file" in https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/README.cmake.

Knapock: It seems Dbus can manage an empty string for these options but not "None":
https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/meson.build#L869

jwillikers: @Knapock Yeah, there isn't a CMake option for `system_pid_file` since CMake is mostly meant for Windows.

I probably wasn't clear on using `None`. I recommend using that in `options` and `default_options` in the `conanfile.py`. Where you pass it into Meson, pass in `""` as the default value so that it gets passed in when it is `None`.

Knapock: @jwillikers Yes, I gave that a shot yesterday but the builds still failed, complaining about the "None". I pushed that change again.

recipes/dbus/1.x.x/conanfile.py
```diff
@@ -104,7 +104,9 @@ def generate(self):
             tc.project_options["checks"] = False
             tc.project_options["doxygen_docs"] = "disabled"
             tc.project_options["modular_tests"] = "disabled"
-            tc.project_options["session_socket_dir"] = str(self.options.session_socket_dir)
+            tc.project_options["system_socket"] = self.options.get_safe("system_socket", "")
+            tc.project_options["system_pid_file"] = self.options.get_safe("system_pid_file", "")
+            tc.project_options["session_socket_dir"] = self.options.get_safe("session_socket_dir", "")
```
jwillikers, line 109: @Knapock Just wrap these with `str`.

```suggestion
            tc.project_options["system_socket"] = str(self.options.get_safe("system_socket", ""))
            tc.project_options["system_pid_file"] = str(self.options.get_safe("system_pid_file", ""))
            tc.project_options["session_socket_dir"] = str(self.options.get_safe("session_socket_dir", ""))
```

recipes/dbus/1.x.x/conanfile.py
```diff
@@ -127,13 +129,14 @@ def generate(self):
             tc.variables["DBUS_WITH_GLIB"] = bool(self.options.get_safe("with_glib", False))
             tc.variables["DBUS_DISABLE_ASSERT"] = is_apple_os(self)
             tc.variables["DBUS_DISABLE_CHECKS"] = False
+            tc.variables["DBUS_SYSTEM_BUS_DEFAULT_ADDRESS"] = self.options.get_safe("system_socket", "")
```
jwillikers, line 132:
```suggestion
            tc.variables["DBUS_SYSTEM_BUS_DEFAULT_ADDRESS"] = str(self.options.get_safe("system_socket", ""))
```

recipes/dbus/1.x.x/conanfile.py
```diff
@@ -127,13 +129,14 @@ def generate(self):
...
 
             # Conan does not provide an EXPAT_LIBRARIES CMake variable for the Expat library.
             # Define EXPAT_LIBRARIES to be the expat::expat target provided by Conan to fix linking.
             tc.variables["EXPAT_LIBRARIES"] = "expat::expat"
 
             # https://github.com/freedesktop/dbus/commit/e827309976cab94c806fda20013915f1db2d4f5a
-            tc.variables["DBUS_SESSION_SOCKET_DIR"] = str(self.options.session_socket_dir)
+            tc.variables["DBUS_SESSION_SOCKET_DIR"] = self.options.get_safe("session_socket_dir", "")
```
jwillikers, line 139:
```suggestion
            tc.variables["DBUS_SESSION_SOCKET_DIR"] = str(self.options.get_safe("session_socket_dir", ""))
```

jwillikers (changes requested):

recipes/dbus/1.x.x/conanfile.py
Knapock: done

recipes/dbus/1.x.x/conanfile.py
Knapock: done

jwillikers (approved):

jwillikers:
@Knapock Can you sign the CLA and move this PR out of draft? It may be helpful to get more eyes on the V2 errors.

uilianries:
I just started a new CI build for Conan v2 with this PR, because the current error is weird. I think it deserve a retry, as this PR would be nice for Conan v2 too.

PR #17499, opened 2023-05-10
at-spi2-core: deprecate old recipe

ericLemanissier:
Specify library name and version:  **at-spi2-core/***

---

### DIFF
--- a/recipes/at-spi2-core/all/conanfile.py
+++ b/recipes/at-spi2-core/all/conanfile.py
@@ -1,5 +1,9 @@
-from conans import ConanFile, Meson, tools
-from conans.errors import ConanInvalidConfiguration
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, replace_in_file,rmdir
+from conan.tools.scm import Version
+
+from conans import Meson
 import os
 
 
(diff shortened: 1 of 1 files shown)


jwillikers (approved):

Croydon (approved):

ericLemanissier:
The motivation behind this PR is that consumers should migrate to a newer version of at-spi2-core, which includes atk and at-spi2-atk since 2.45.1 cf https://gitlab.gnome.org/GNOME/at-spi2-core/-/blob/main/NEWS#L70

recipes/at-spi2-core/all/conanfile.py
```diff
@@ -11,6 +15,7 @@ class AtSpi2CoreConan(ConanFile):
     homepage = "https://gitlab.gnome.org/GNOME/at-spi2-core/"
     license = "LGPL-2.1-or-later"
     generators = "pkg_config"
+    deprecated = True
```
RubenRBS, line 18: As seen in the linked PR https://github.com/conan-io/conan/pull/13991, the deprecated field can contain custom text to show to the user, which is a great opportunity to show them _why_ they should use the newer version. So something like your comment would be great!

```suggestion
    deprecated = "Consumers should migrate to at-spi2-core/[>=2.45.1], which includes atk and at-spi2-atk"
```

czoido (approved):

jwillikers (approved):

PR #7613, opened 2021-10-08
LLVM 13 and 14

rdeterre:
Specify library name and version:  **llvm/13.0.0**

This is a new PR for the changes from @Manu343726 and @tcanabrava to add an LLVM recipe.

PR #17509, opened 2023-05-10
package: llvm/13 - llvm/16

jusito:
Specify library name and version:  llvm/13.0.1, llvm/14.0.6, llvm/15.0.7, llvm/16.0.6

[Current State 07.2023 waiting for reviews ](https://github.com/conan-io/conan-center-index/pull/17509#issuecomment-1620066348)

follow up on previous pull request: https://github.com/conan-io/conan-center-index/pull/7613

My _main_ focus in this pr is:
- llvm/14.0.6 without any project/runtime should work on linux/mac
- llvm/14.0.6 + clang / dynlib should be usable if supported by the project itself
- keeping certain binaries, e.g. keeping clang-14 as binary to compile with this specific clang version to LLVM IR but using a different compiler for llvm/clang itself

What I added:
- Conan2 migration
- job pool calculation according to available memory, helps a lot in a shared ci environment as a trade of, could fix the previous timeout [explained here](https://github.com/conan-io/conan-center-index/pull/7613#issuecomment-1112075014)
- Linux/Ubuntu tested with clang-15 in many different configurations
  - beside lldb/libc/compiler-rt everything should work, this three are always a bit special
    - with libc enabled clang can crash in sin/cos/tan for llvm/13-14 in debug build
    - compiler-rt always requires x86 headers for compilation
    - lldb has circular deps which I removed by hand, probably thats not always a good idea
- Windows & Mac ongoing development
- lots of testing and tweaking

Tasks:
- Windows testing
- Mac testing, currently lack of hardware
- compatibility settings, why rebuild llvm without clang if there is a package which contains clang?

Feedback and help welcome.

---

### DIFF
--- a/recipes/llvm/all/conanfile.py
+++ b/recipes/llvm/all/conanfile.py
@@ -247,9 +247,10 @@ class Llvm(ConanFile):
 
         if self.options.conan_center_index_limits:
             # XXX conandata.yml is reduced in cci ci to exactly one version so we can't look it up
-            if not str(self.version) in ['13.0.1', '14.0.6', '15.0.7', '16.0.6', '17.0.2']:
+            # 14 needed for a follow up pr, 16 and 17 are identical in requirements
+            if not str(self.version) in ['14.0.6', '17.0.2']:
                 raise ConanInvalidConfiguration(
-                    "llvm version is disabled for conan center index ci because its not the latest patch level in the configuration without the ci would run for multiple days. You can enable it with option conan_center_index_limits=False.")
+                    "llvm version is disabled for conan center index ci. We have a tight ci budget for such a huge recipe. You can enable it with option conan_center_index_limits=False.")
             if self.settings.build_type == "Debug":
                 raise ConanInvalidConfiguration(
                     "LLVM Debug builds are disabled as a workaround of conan center index ci memory limits. You can enable it with option conan_center_index_limits=False.")


jusito:
Currently waiting for feeback or review, or if there is a result for #17999 . The ci worked on the last [commit](https://github.com/conan-io/conan-center-index/pull/17509#issuecomment-1596778268) so Iam fine with the current state for now. @jcar87 because this recipe is quite large, if you have a question or if I could help you with some information to speed things up please don't hesitate to ask or point out weird stuff I have done which doesn't make sense to you.

If you want to help in this recipe:
- Windows builds are disabled because I have no idea what I need to do there. Version check needs adaptation regarding [this](https://github.com/conan-io/tribe/blob/main/design/032-msvc_support.md) and there is a mapping to MD(d)/MT(d) needed. @sykhro still interested in windows builds?
- compatibility settings, e.g. why rebuild llvm without clang if there is a package which contains clang?
Imho even without this features this recipe contains enough changes for now.

whaliendev:
> > * Windows builds are disabled because I have no idea what I need to do there. Version check needs adaptation regarding [this](https://github.com/conan-io/tribe/blob/main/design/032-msvc_support.md) and there is a mapping to MD(d)/MT(d) needed. @sykhro still interested in windows builds?
> 
> Yes, I will try to test this but I don't have access to strong compiling machines over the summer.

I have a strong compiling machine and I'd also like to help. But I'm not familiar with windows and MSVC, what should I do?

sykhro:
> > > * Windows builds are disabled because I have no idea what I need to do there. Version check needs adaptation regarding [this](https://github.com/conan-io/tribe/blob/main/design/032-msvc_support.md) and there is a mapping to MD(d)/MT(d) needed. @sykhro still interested in windows builds?
> > 
> > 
> > Yes, I will try to test this but I don't have access to strong compiling machines over the summer.
> 
> I have a strong compiling machine and I'd also like to help. But I'm not familiar with windows and MSVC, what should I do?

If you don't want the whole IDE, you can install the "Build Tools for Visual Studio Community". Be sure to install the C++ components. You can't get it wrong if you invoke the installer via powershell:
```pwsh
$InstallArgs = @(
    '--add Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools'
    '--add Microsoft.VisualStudio.Workload.VCTools'
    '--add Microsoft.VisualStudio.Component.VC.Tools.x86.x64'
    '--add Microsoft.VisualStudio.Component.Windows11SDK.22621'
    '--add Microsoft.Component.VC.Runtime.UCRTSDK'
    '--add Microsoft.VisualStudio.Component.VC.ATL'
    '--add Microsoft.VisualStudio.Component.VC.ATLMFC'
    '--wait'
    '--quiet'
    '--nocache'
)
$Result = Start-Process VS_INSTALLER.EXE -ArgumentList "$InstallArgs" -Wait -PassThru
```
 Once this is done, you can use Conan as usual after you have created a profile with `msvc` (default should work too).

whaliendev:
> > > > * Windows builds are disabled because I have no idea what I need to do there. Version check needs adaptation regarding [this](https://github.com/conan-io/tribe/blob/main/design/032-msvc_support.md) and there is a mapping to MD(d)/MT(d) needed. @sykhro still interested in windows builds?
> > > 
> > > 
> > > Yes, I will try to test this but I don't have access to strong compiling machines over the summer.
> > 
> > 
> > I have a strong compiling machine and I'd also like to help. But I'm not familiar with windows and MSVC, what should I do?
> 
> If you don't want the whole IDE, you can install the "Build Tools for Visual Studio Community". Be sure to install the C++ components. You can't get it wrong if you invoke the installer via powershell:
> 
> ```powershell
> $InstallArgs = @(
>     '--add Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools'
>     '--add Microsoft.VisualStudio.Workload.VCTools'
>     '--add Microsoft.VisualStudio.Component.VC.Tools.x86.x64'
>     '--add Microsoft.VisualStudio.Component.Windows11SDK.22621'
>     '--add Microsoft.Component.VC.Runtime.UCRTSDK'
>     '--add Microsoft.VisualStudio.Component.VC.ATL'
>     '--add Microsoft.VisualStudio.Component.VC.ATLMFC'
>     '--wait'
>     '--quiet'
>     '--nocache'
> )
> $Result = Start-Process VS_INSTALLER.EXE -ArgumentList "$InstallArgs" -Wait -PassThru
> ```
> 
> Once this is done, you can use Conan as usual after you have created a profile with `msvc` (default should work too).

many thanks. I'll help to test this recipe later.

whaliendev:
@jusito it built on Windows successfully with default options on msvc 17, which ships with visual studio build tools 2022. Nice work! 
However, the test_package recipe fails, the build log is:
```shell
llvm/16.0.6 (test package): Running test()
'.' is not recognized as an internal or external command, operable program, 
or batch file.
ERROR: llvm/16.0.6 (test package): Error in test() method, line 55
        self.run("./test_package")
        ConanException: Error 1 while executing ./test_package
```
just a minor tweak is enough. 

I'll test the recipe with other projects and runtimes enabled later. 

Do you have any other key options you would like to test on Windows? I can help.

jusito:
@whaliendev  Thanks for testing, currently I would suggest only testing the different llvm versions but if you yourself don't need specific configurations beside default its probably just very time consuming.

Regarding Windows todos.
To enable windows builds for Conan v1 and v2 there are two issues:
1. Correctly checking for msvc / visual studio version. Currently the recipe has this line which is migrated from a previous contributor:
image
If we have a look into conan settings, ...
... Conan v1 settings:
image
... Conan v2 settings:
image
We can see the version can be 16 or 17, doesn't look like it can be 16.3 or 16.4.
Also for Conan v2 Visual studio is removed and we can see this version check for 16 is only valid for visual studio but msvc versions are always higher. You can see it here: https://github.com/conan-io/tribe/blob/main/design/032-msvc_support.md
So the version check for 16.4 needs a migration to [check_min_vs](https://docs.conan.io/1/reference/conanfile/tools/microsoft.html#check-min-vs) which checks for msvc version which starts at 170. No idea which msvc version is matching visual studio 16.4. If this is done, version check should be ready for Conan v1 and v2.

2. From the settings above we can also see the runtime value for msvc and Visual Studio is different. Currently the recipe is fine with MD, MDd, MT, MTd because this values are accepted by llvm directly. But for static and dynamic value there is a mapping needed. Maybe dynamic + release = MD, dynamic + debug = MDd, static + release MT, static + debug = MTd but thats just a guess. Because of this I added this raise here:
image
Which disables msvc but is fine with Visual Studio, so Conan v2 is disabled and only Conan v1 Visual Studio builds should work. I tried to dig a bit deeper into the topic but according to some discussions here its probably not that simple.
The mapping is needed where we are setting LLVM_USE_CRT_... .
image

whaliendev:
> @whaliendev Thanks for testing, currently I would suggest only testing the different llvm versions but if you yourself don't need specific configurations beside default its probably just very time consuming.
> 
> Regarding Windows todos. To enable windows builds for Conan v1 and v2 there are two issues:
> 
> 1. Correctly checking for msvc / visual studio version. Currently the recipe has this line which is migrated from a previous contributor:
>    image
>    If we have a look into conan settings, ...
>    ... Conan v1 settings:
>    image
>    ... Conan v2 settings:
>    image
>    We can see the version can be 16 or 17, doesn't look like it can be 16.3 or 16.4.
>    Also for Conan v2 Visual studio is removed and we can see this version check for 16 is only valid for visual studio but msvc versions are always higher. You can see it here: https://github.com/conan-io/tribe/blob/main/design/032-msvc_support.md
>    So the version check for 16.4 needs a migration to [check_min_vs](https://docs.conan.io/1/reference/conanfile/tools/microsoft.html#check-min-vs) which checks for msvc version which starts at 170. No idea which msvc version is matching visual studio 16.4. If this is done, version check should be ready for Conan v1 and v2.
> 2. From the settings above we can also see the runtime value for msvc and Visual Studio is different. Currently the recipe is fine with MD, MDd, MT, MTd because this values are accepted by llvm directly. But for static and dynamic value there is a mapping needed. Maybe dynamic + release = MD, dynamic + debug = MDd, static + release MT, static + debug = MTd but thats just a guess. Because of this I added this raise here:
>    image
>    Which disables msvc but is fine with Visual Studio, so Conan v2 is disabled and only Conan v1 Visual Studio builds should work. I tried to dig a bit deeper into the topic but according to some discussions here its probably not that simple.
>    The mapping is needed where we are setting LLVM_USE_CRT_... .
>    image

Yes, it's indeed very time-consuming. Next, I'd like to enable different subprojects and runtimes to verify that this recipe can be built successfully, at least on relatively newer toolchains. After that, I'll consider checking the version check for v1 and v2, but this may take a long time to accompalnish. I'll update the test cases status later if there is enough time. 

On a side note, the current recipe works very well on Linux, and I'm grateful for that. For you guys who want to manage LLVM using Conan on Linux, you can set up a Conan local repository. Thanks again.

sykhro:
One thing that would be really useful is the support for custom allocators on windows.
It can only be used with the static crt, but it provides an insane performance boost, especially during linking (see `LLVM_INTEGRATED_CRT_ALLOC`)

FeignClaims:
@jusito Hi, thanks for your excellent work. It truly helped a lot for me.

However, I noticed there's an unhandled option [`LLVM_ENABLE_ZSTD`](https://github.com/llvm/llvm-project/blob/561d3076aa6a3de5ae3afcc1f96630df3943c921/llvm/CMakeLists.txt#L518C5-L518C21) which brings a dependency on zstd since maybe LLVM 15.

Current

image

FeignClaims:
Hi, thanks for your excellent work. It truly helped a lot for me.

However, I noticed there's an unhandled option [LLVM_ENABLE_ZSTD](https://github.com/llvm/llvm-project/blob/561d3076aa6a3de5ae3afcc1f96630df3943c921/llvm/CMakeLists.txt#L518C5-L518C21) which brings a dependency on zstd since maybe LLVM 15.

My current workaround is to `self.requires("zstd/1.5.5")` and link it myself.

Currently the [Building LLVM with CMake](https://llvm.org/docs/CMake.html#options-and-variables) says nothing about this option, so I'm wondering whether this was a mistake because of the doc missing or actually intentional.

image

FeignClaims:
> Currently the [Building LLVM with CMake](https://llvm.org/docs/CMake.html#options-and-variables) says nothing about this option, so I'm wondering whether this was a mistake because of the doc missing or actually intentional.

llvm/llvm-project#65201 may be related to this.

jusito:
thank you, there are also some other options which adding deps.
image

and for 16 they introduce LLVM_USE_STATIC_ZSTD there, probably we also need to handle it I guess.
I have a look into that.

sykhro:
> thank you, there are also some other options which adding deps. image
> 
> and for 16 they introduce LLVM_USE_STATIC_ZSTD there, probably we also need to handle it I guess. I have a look into that.
> 
> Edit: LLVM_Z3_INSTALL_DIR also needed because LLVM_ENABLE_Z3 is already there. Not sure how to handle with conan.

Saw this just now. We are doing it like this in our recipe:
```python
def generate(self):
        [...]
        if self.options.with_z3:
            tc.cache_variables["LLVM_ENABLE_Z3_SOLVER"] = True
             tc.cache_variables["LLVM_Z3_INSTALL_DIR"] = (
                Path(self.dependencies["z3"].package_folder).resolve().as_posix()
            )
       [...]
```
(don't mind the dictionary - we do a stage2 clang build so the

jacobfriedman:
Why introduce the ninja build dependency? For that reason I've decided not to use this.

sykhro:
> Why introduce the ninja build dependency? For that reason I've decided not to use this.

By using ninja you have finer control over parallel jobs and memory consumption (try linking lld in debug mode!). I do agree that it shouldn't be mandatory, but I can see CI struggling to build this for CCI otherwise

jacobfriedman:
> > Why introduce the ninja build dependency? For that reason I've decided not to use this.
> 
> By using ninja you have finer control over parallel jobs and memory consumption (try linking lld in debug mode!). I do agree that it shouldn't be mandatory, but I can see CI struggling to build this for CCI otherwise

Thanks for the quick reply- I was looking deeper... ninja's not actually a requirement to build- it should probably be updated as an option if so. I'm not sure how to express that as an optional requirement in the conanfile... which means I'll have to fork this and remove the cmake declaration. It's just not *technically* a requirement to build the main library, but that's just my opinion - if anyone else wants to weigh in that would be appreciated.

With that all said... thank you for the great work! I'll pull it down, snip ninja out and be on my, now, merry way.

sykhro:
> > > Why introduce the ninja build dependency? For that reason I've decided not to use this.
> > 
> > 
> > By using ninja you have finer control over parallel jobs and memory consumption (try linking lld in debug mode!). I do agree that it shouldn't be mandatory, but I can see CI struggling to build this for CCI otherwise
> 
> Thanks for the quick reply- I was looking deeper... ninja's not actually a requirement to build- it should probably be updated as an option if so. I'm not sure how to express that as an optional requirement in the conanfile... which means I'll have to fork this and remove the cmake declaration. It's just not _technically_ a requirement to build the main library, but that's just my opinion - if anyone else wants to weigh in that would be appreciated.
> 
> With that all said... thank you for the great work! I'll pull it down, snip ninja out and be on my, now, merry way.

You should thank @jusito :) I'm merely relaying suggestions since we have a very similar (but not identical) recipe for our usecase.
It's absolutely doable to have a recipe option for this, I'm just not sure if conan-center allows options that are actually meta-options (no matter if you use ninja or other generators, the expected outcome is always the same).

I would suggest to remove ninja from this recipe. Users could opt in specifying it as a tool-requires from `conan install ...`

jusito:
@jacobfriedman I would be interested in your reasoning for stripping Ninja out?

jacobfriedman:
> @jacobfriedman I would be interested in your reasoning for stripping Ninja out?

Speaking for myself; it's not really needed, so it's not a requirement.

The better question is: do you absolutely need to put it in?

jusito:
@jacobfriedman 
We need Ninja here for CI because even with it, we are pretty close to max CI time, but it's integrated in a way that it will calculate job pools according to your available resources for CPU and memory. One of the patches is adding it and this functionality as part of a llvm pr.

jacobfriedman:
Compiling LLVM is probably one of the edge-cases that the CI should be able to handle- and conan, or any sysadmin, would probably allow. 

Otherwise, it's simply not a requirement! One does not need ninja to compile this.

jacobfriedman:
See above @jusito

jusito:
@jacobfriedman 
It is an edge case, same as qt and boost which are added to exceed max ci time a bit. https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L46-L50
This is a following pr which exists since llvm 14 and I don't see help from conan devs for this currently, it was very time consuming passing CI with this recipe. I highly doubt you would like to test v1/v2 in 4 versions with different configurations with a single job at a time, I would call that painfully slow. Ninja is well supported by llvm, ninja build provided via Conan so you don't need to install it yourself. Ninja itself isn't a random niche tool out there, so what is behind your "not mandatory -> will remove it"?

1. So CI needs it for sure in the current state.
2. For testing purposes speed it up a lot.
3. Could you please explain the reason behind "not mandatory -> I will spend additional time to remove it"

jusito:
@jacobfriedman 
[I really hope that there is feedback on this, maybe I should have asked more precisely for it, but I created an Issue and collected several topic around it.](https://github.com/conan-io/conan-center-index/issues/17999)

There is a discussion about CI time; every build should be tested, and if you exceed it, remove old versions.

sykhro:
> @jacobfriedman [I really hope that there is feedback on this, maybe I should have asked more precisely for it, but I created an Issue and collected several topic around it.](https://github.com/conan-io/conan-center-index/issues/17999)
> 
> There is a discussion about CI time; every build should be tested, and if you exceed it, remove old versions.

I would advise also shooting a message in the #conan channel on the cpp slack, it's more likely that someone would respond there

jacobfriedman:
Thanks. If the CI can accommodate the edge case, then ninja is not
mandatory.

First - for security. Omit needless dependencies where possible as a best
practice — period. Depedencies become attack vectors.

Second ‐ dependencies can be problematic to manage and update. Who wants to
manage this in the future if versions or APIs change? It becomes a
maintenance cost.

Third - implementation. What other platforms does this support at the
moment that haven't been tested? This is a stretch but shouldn't really
matter given 1 and 2.

jusito:
> ```python
> def generate(self):
> [...]
> if self.options.with_z3:
>      tc.cache_variables["LLVM_ENABLE_Z3_SOLVER"] = True
>      tc.cache_variables["LLVM_Z3_INSTALL_DIR"] = (
>         Path(self.dependencies["z3"].package_folder).resolve().as_posix()
>     )
> [...]
> ```
> 
> the use of `pathlib` as_posix() on all platforms is deliberate (avoids bugs with backslashes on Windows)

Is your change tested with Conan 2? Because from my testing this will not work with conan 1+llvm 13 and conan 2.
This issue is similar to libxml2 :

> ERROR: llvm/16.0.6: Required package 'z3' not in component 'requires'

sykhro:
> > ```python
> > def generate(self):
> > [...]
> > if self.options.with_z3:
> >      tc.cache_variables["LLVM_ENABLE_Z3_SOLVER"] = True
> >      tc.cache_variables["LLVM_Z3_INSTALL_DIR"] = (
> >         Path(self.dependencies["z3"].package_folder).resolve().as_posix()
> >     )
> > [...]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the use of `pathlib` as_posix() on all platforms is deliberate (avoids bugs with backslashes on Windows)
> 
> Is your change tested with Conan 2? Because from my testing this will not work with conan 1+llvm 13 and conan 2. This issue is similar to libxml2 :
> 
> > ERROR: llvm/16.0.6: Required package 'z3' not in component 'requires'



> > ```python
> > def generate(self):
> > [...]
> > if self.options.with_z3:
> >      tc.cache_variables["LLVM_ENABLE_Z3_SOLVER"] = True
> >      tc.cache_variables["LLVM_Z3_INSTALL_DIR"] = (
> >         Path(self.dependencies["z3"].package_folder).resolve().as_posix()
> >     )
> > [...]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the use of `pathlib` as_posix() on all platforms is deliberate (avoids bugs with backslashes on Windows)
> 
> Is your change tested with Conan 2? Because from my testing this will not work with conan 1+llvm 13 and conan 2. This issue is similar to libxml2 :
> 
> > ERROR: llvm/16.0.6: Required package 'z3' not in component 'requires'

I'm not sure that's a v1 vs v2 issue. We map cmake library names to conan package names and add those ourselves with `self.cpp_info.components[component].set_property("cmake_target_name", component)`

jacobfriedman:
> > @jacobfriedman [I really hope that there is feedback on this, maybe I should have asked more precisely for it, but I created an Issue and collected several topic around it.](https://github.com/conan-io/conan-center-index/issues/17999)
> > There is a discussion about CI time; every build should be tested, and if you exceed it, remove old versions.
> 
> I would advise also shooting a message in the #conan channel on the cpp slack, it's more likely that someone would respond there

I don't have a "workspace" email there. Could you give us a hand, or if anyone has access to that slack, could they bump this?

jacobfriedman:
@jusito is this the change you would need to remove Ninja?

jacobfriedman:
Also- would a system_tool / build requirement also require gcc? If using clang, the test_package won't work.

jusito:
@jacobfriedman 

1. regarding pr, short no, does it help this pr in general yes. I would be fine and for me less effort to add an option which excludes Ninja and only using Make Generator but default using Ninja because of the compile / link job pool calculation to optimise CI time / not be killed. This would need migration to python and a lot of testing which I just don't have the time for.
2. You just want to exchange Ninja with another generator Make, so you just switching dependencies its not an additional one. Your third is the strongest argument for me, but the recipe for Conan doesn't restrict it to certain platforms, therefore I don't see an issue.
3. The current state doesn't require gcc if clang is used, why do you think so?

jacobfriedman:
#2: I'm not switching dependencies, this removes the actual stated requirement - which is not required. If you want to specify the build_requirement, then put ninja there. I just built without ninja. Ninja is not a requirement. It's akin to saying gasoline is required for a vehicle to go down a hill because you want it to get to the bottom as fast as possible.

#3: I just built on an M2 with GCC and everything worked well. self.info.options.enable_unsafe_mode concerns test_package and possible linking issues, but I didn't test with clang.

jusito:
3. I added this message because I had issues with it on Linux but apple-clang with libc++ is fine with it, as you can see in the conan center index ci.
2. You need to use a generator, so if you want to see Make there I would add Make as requirement because it would be an requirement so you would exchange it. Not adding any generator as dependency would be a flaw imho because it would violate 3.

jacobfriedman:
Most packages requiring make or cmake do not explicitly say so. It is good
practice. I am now figuring out myself how to call a conan-installed cmake
(llvm) build meta-requirement.

Whatever it is, Ninja or make (should be make imo), it should be explicit.
Especially if it's a compiler/toolchain. I can add a PR if you'd like. Let
me know.

uilianries:
I Added llvm to a larger timeout list: https://github.com/conan-io/conan-center-index/pull/20389

I strongly suggest only including LLVM 16 for now and including older versions later, only if strict necessary. LLVM takes hours to be built and consumes as much resources as any other regular library, so I would prefer being conservative about huge packages like this one.

jwillikers:
> I Added llvm to a larger timeout list: #20389
> 
> I strongly suggest only including LLVM 16 for now and including older versions later, only if strict necessary. LLVM takes hours to be built and consumes as much resources as any other regular library, so I would prefer being conservative about huge packages like this one.

I agree, though I'd also be okay if you bumped to 17.

jwillikers (changes requested):

recipes/llvm/all/patches/16/calculate_job_pools.patch
```diff
@@ -0,0 +1,53 @@
+diff --git a/llvm/cmake/modules/HandleLLVMOptions.cmake b/llvm/cmake/modules/HandleLLVMOptions.cmake
```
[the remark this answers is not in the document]
jusito: At least one reviewer is fine after some cleanup: https://github.com/llvm/llvm-project/pull/65274

jusito:
@jwillikers 
Thanks, I will fix it next week.

> I agree, though I'd also be okay if you bumped to 17.

I need llvm 14 for one of our projects; that's my most important point in pushing this recipe. Therefore, I would prefer to keep at least llvm 14 but I will also check again how far we are in current code migration. I also looked around and saw the whole discussion about removing outdated versions. Imho there should be a distinction between "available in Conan" and "available in Conan and tested in cci CI," but probably there is a better issue to discuss this.

recipes/llvm/all/patches/16/calculate_job_pools.patch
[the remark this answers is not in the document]
jwillikers: Awesome! That's fantastic. Thanks for doing that. I get the feeling that's likely to make my life easier when I find myself compiling LLVM at some point in the future.

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,791 @@
...
+            'use_llvm_cmake_files': False,
+            # creating job pools with current free memory
+            'ram_per_compile_job': '2000',
+            'ram_per_link_job': '14000',
+            'conan_center_index_limits': True,
+            'enable_unsafe_mode': False,
+        }
+    }
+
+    exports_sources = 'patches/**/*', 'conandata.yml'
```
jwillikers, line 166: Use `export_conandata_patches(self)` in the `export_sources()` method here instead.
```suggestion
    def export_sources(self):
        export_conandata_patches(self)
```

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,791 @@
...
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
```
jwillikers, line 173: Is this due to the `no_copy_source` option being set? You should apply patches in the `build` method instead of the source method.

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,791 @@
...
+
+    def is_windows(self):
+        return self.settings.os == "Windows"
+
+    def is_macos(self):
+        return self.settings.os == "MacOS"
+
+    def is_linux(self):
+        return self.settings.os == "Linux"
```
jwillikers, line 182: To be safe, you should prefix any custom methods or attributes with an `_` to avoid conflicts.

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,791 @@
...
+        if release < 15:
+            unsupported_options.append('with_zstd')
+            unsupported_options.append('with_httplib')
+        # TODO fix with_zstd and llvm 17
+        if release >= 17:
+            self.output.warning(f"with_zstd will fail in llvm 17 with conan 2")
+            unsupported_options.append('with_zstd')
+        for opt in unsupported_options:
+            self.output.warning(f"{opt} is unsupported in llvm {release}")
+            self.options.rm_safe(opt)
```
jwillikers, line 207: It's probably better to delete options based off of the version and any `settings` in the `config_options` method instead. This way, the options won't be available when consumers try to set them in the first place. You'll want to put that above the `configure` method to reflect the evaluation order.

```python
    def config_options(self):
            self.options.rm_safe("fPIC")  # FPIC MANAGEMENT (KB-H007)
        release = Version(self.version).major
        unsupported_options = []
        if release < 14:
            unsupported_options.append('with_curl')
        if release < 15:
            unsupported_options.append('with_zstd')
            unsupported_options.append('with_httplib')
        # TODO fix with_zstd and llvm 17
        if release >= 17:
            unsupported_options.append('with_zstd')
        for opt in unsupported_options:
            self.options.rm_safe(opt)
```

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,791 @@
...
+
+    def generate(self):
+        tc = CMakeToolchain(self, "Ninja")
+        tc.generate()
+
+    def build(self):
+        cmake = self._cmake_configure()
+        cmake.build()
+
+    def _cmake_configure(self):
```
jwillikers, line 331: This should be migrated to use the toolchain in the `generate` method instead of the CMake `configure` method.

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,791 @@
...
+            # no version number in llvm 15-17
+            self.requires('cpp-httplib/[>=0.5.4 <1.0.0]')
+
+    def build_requirements(self):
+        # Older cmake versions may have issues generating the graphviz output used
+        # to model the components
+        self.tool_requires("cmake/[>=3.21.3 <4.0.0]")
+        self.tool_requires("ninja/[>=1.10.0 <2.0.0]")
+
+    def generate(self):
```
jwillikers, line 323: I think you need VirtualBuildEnv here.

jwillikers (changes requested):
Thanks for adding LLVM 17!

I recommend looking over the [CMake package template](https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates/cmake_package), as you'll want to align with it as much as possible.

recipes/llvm/all/conanfile.py
jwillikers: Oops, nevermind. I see this is exactly what we do for the Qt package.

recipes/llvm/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,32 @@
+from conans import ConanFile, CMake, tools
```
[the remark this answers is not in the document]
jusito: I think I didn't have one in the first place but ci complained about.

jusito:
@jwillikers I will test the configs locally first before adding your suggestions, some parts are really what I was looking for, thanks a lot!

recipes/llvm/all/test_v1_package/conanfile.py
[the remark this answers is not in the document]
jwillikers: I believe the policy has changed since you opened your PR.

recipes/llvm/all/conanfile.py
```diff
@@ -0,0 +1,792 @@
...
+            'with_xml2': False,
+            'keep_binaries_regex': '^$',
+
+            # options removed in package id
+            # XXX Should these files used by conan at all?
+            'use_llvm_cmake_files': False,
+            # creating job pools with current free memory
+            'ram_per_compile_job': '2000',
+            'ram_per_link_job': '14000',
+            'conan_center_index_limits': True,
```
jwillikers, line 161: Instead of an option, you can use a `conf` variable to control this, like `user.c3i.skip:llvm`.
You can get the value of this variable via `self.conf.get`.

```
self.conf.get('user.c3i.skip:llvm', default=False, check_type=bool)
```

Then in `global.conf` or on the CLI, it can be defined.

```
conan ... --conf user.c3i.skip:llvm=True
```

jwillikers (changes requested):

jwillikers:
Why is `LLVM_TARGETS_TO_BUILD` set to `all` by default? Could this default instead to just the target architecture specified in the host profile?

Also, this recipe would be helpful for #20528, which requires components from `llvm`, one of which is `libclc`. I'm attempting to use this recipe as a dependency, but `libclc` isn't produced by the recipe when I enable the `with_project_libclc` option.
Something like the following should help with packaging `libclc` for reuse.

```python
    def package(self):
        cmake = self._cmake_configure()
        cmake.install()
        if self.options.get_safe("with_project_libclc"):
            copy(self, "*.bc", os.path.join(self.package_folder, "share", "clc"), os.path.join(self.package_folder, "lib", "clc"))
```

```python
    def package_info(self):
        self.cpp_info.components["libclc"].set_property("pkg_config_name", "libclc")
        self.cpp_info.components["libclc"].libdirs = [os.path.join(self.package_folder, "lib", "clc")]
```

jwillikers:
Ah, I see now. It appears that `libclc` should probably be placed in a separate package. It depends on [SPIRV-LLVM-Translator](https://github.com/KhronosGroup/SPIRV-LLVM-Translator) which depends on both SPIRV-tools and LLVM. Also, I had to build it directly from its separate source directory locally to build it at all.

Also, with LLVM 17, I noticed that the `flang` and `mlir` projects are absent from the list of projects.

jusito:
> Why is `LLVM_TARGETS_TO_BUILD` set to `all` by default? Could this default instead to just the target architecture specified in the host profile?

I stumbled upon this too, but didn't thought about taking host profile which sounds reasonable.

> Also, with LLVM 17, I noticed that the flang and mlir projects are absent from the list of projects.

Ok, also a good idea for your config_options suggestion, to remove it there. For this recipe and changed config I have already a workflow to check for definition/requirement changes, didn't expect to miss such change, thanks!

> [...] but `libclc` isn't produced by the recipe when I enable the `with_project_libclc` option. [...] Ah, I see now. It appears that libclc should probably be placed in a separate package. [...]

So this option shouldn't be part of this recipe.

regarding:
> self.conf.get('user.c3i.skip:llvm', default=False, check_type=bool)
Where did you find this? During my search even looked through related issues and never found a note on this, really handy if it works.

jwillikers:
That was recommended [here](https://github.com/conan-io/conan-center-index/pull/20290#discussion_r1361715509).

jacobfriedman:
Happy to see this moving along. Thank you for the hard work @jusito

jacobfriedman:
Building LLVM 15 with a direct_deploy, clang requires libatomic. I wonder if COMPILER_RT_BUILD_STANDALONE_LIBATOMIC will help if COMPILER_RT is needed.

jusito:
@jacobfriedman 
Just to clarify for replication, for a llvm 15 build:
- LLVM_ENABLE_PROJECTS should contain clang and compiler-rt
- What do mean by direct deploy, conan create ?
- there is a dependency missing which may be fixed with COMPILER_RT_BUILD_STANDALONE_LIBATOMIC, which adds clang_rt.atomic as dependency to compiler_rt
correct?

sykhro:
> @jacobfriedman Just to clarify for replication, for a llvm 15 build:
> 
> * LLVM_ENABLE_PROJECTS should contain clang and compiler-rt

`compiler-rt` should be in `LLVM_ENABLE_RUNTIMES`

> * there is a dependency missing which may be fixed with COMPILER_RT_BUILD_STANDALONE_LIBATOMIC, which adds clang_rt.atomic as dependency to compiler_rt

`compiler-rt` already includes `libatomic`. the `COMPILER_RT_BUILD_STANDALONE_LIBATOMIC` flag builds `clang_rt.atomic` separately, so that you can link to it with `-latomic` like you would do with gcc

jacobfriedman:
It wasn't built natively on mine with llvm 15 using this codebase, unless I
need to update. If you're sure it's included, then all the better.

jwillikers:
Please don't close `github-actions` bot.

jcar87:
We are currently not prioritising packages that add toolchains (rather than libraries or pure executables), as we are prioritising other issues in our backlog.

LLVM itself has a myriad of components, and I suspect different users would have different needs. We are opening this issue to track this: https://github.com/conan-io/conan-center-index/issues/23834  - actual uses cases with examples will help the team gain better insights at how a potential llvm recipe may be used, whether or not it is worth packaging different LLVM components in separate recipes, and so on. Any information of this would be appreciated in that issue so that we can move forward, thanks!

jacobfriedman:
Why would you close this and open another issue to justify this project?

Issue #16059, opened 2023-02-16
[package] <LIBRARY-NAME>/<LIBRARY-VERSION>: SHORT DESCRIPTION


olivia76:
### Description

It is not possible to build the arrow package with parquet support and with lz4 / brotli encoders/decoders activated on conanio/gcc9 docker image.

### Package and Environment Details

conanio/gcc9 docker image.

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=9
compiler.libcxx=libstdc++11
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

`mkdir -p build; cd build; conan install .. --build=missing --profile:build=default`

conanfile.py:

```
from conans import ConanFile, tools
from conan.tools.cmake import CMakeDeps, CMakeToolchain, cmake_layout, CMake
import os

class CppParquet(ConanFile):
    name = "cppparquet"
    version = "0.0.0"
    settings = "os", "compiler", "build_type", "arch"
    options = {"shared": [True, False], "fPIC": [True, False]}
    default_options = {"shared": True, "fPIC": True}
    build_policy = "missing"

    requires = (
        "arrow/10.0.1",
        )
    exports_sources = "CMakeLists.txt", "src/*", "include/*", "tests/*", "apps/*", "conanfile.py", "cmake/*"
    
    def config_options(self):
        if self.settings.os == "Windows":
            del self.options.fPIC

    def layout(self):
        cmake_layout(self)

    def generate(self):
        tc = CMakeToolchain(self)
        tc.cache_variables["ENABLE_COVERAGE"] = self.options.ENABLE_COVERAGE
        tc.generate()
        deps = CMakeDeps(self)
        deps.generate()
        
    def build_requirements(self):
        #self.requires("lz4/1.9.4")
        self.test_requires("catch2/2.13.6")

    def build(self):
        cmake = CMake(self)
        cmake.verbose = True
        cmake.configure()
        cmake.build()
        cmake.test()
        
    def package_info(self):
        # These are default values and doesn't need to be adjusted
        self.cpp_info.includedirs = ["include"]
        self.cpp_info.libdirs = ["lib"]
        self.cpp_info.bindirs = ["bin"]
        self.cpp_info.libs = ["cppdataset"]
        self.env_info.LD_LIBRARY_PATH.append(os.path.join(self.package_folder, "lib"))
        
    def configure(self):
[...]

PR #17510, opened 2023-05-10
arrow: use CMakeDeps set property to match the expected Brotli names

jcar87:
Specify library name and version:  **arrow/all**

Instead of patching the CMakeLists to match the target names generated by CMakeDeps, use CMakeDep's ability to change the target names so that they match the ones in the CMakeLists instead - this should be more robust as patches tend to get lost across versions.

Close: https://github.com/conan-io/conan-center-index/issues/16059


RubenRBS (approved):

SpaceIm:
Probably more future proof.

FYI, there is https://github.com/apache/arrow/pull/14671 since 11.0.0

PR #17513, opened 2023-05-10
poco: use version range openssl + bump dependencies

SpaceIm:
- use version range for openssl
- bump sqlite3 & libpq
- drop maintenance of poco < 1.11.0. They are not compatible with openssl 3.x, and c3i can't build these versions anymore because packages of latest RREV of libmysqlclient (dependency of poco) are built with openssl 3.x, so it leads to missing packages.

---


uilianries (approved):
LGTM

RubenRBS (approved):
Thanks a lot!

PR #17523, opened 2023-05-11
vulkan-memory-allocator: Add package_type and update vulkan-headers dependency

RubenRBS:
Specify library name and version:  **vulkan-memory-allocator/all**

Found while looking into https://github.com/conan-io/conan/issues/13875. This fixes a bug where the vulkan-headers should have been a transitive headers dependency, as it uses Vulkan headers


SpaceIm (approved):

SpaceIm:
> This fixes a bug where the vulkan-headers should have been a transitive headers dependency, as it uses Vulkan headers

Jut to confirm: when `package_type = "header-library"`, can we safely assume that transitive_headers & transitive_libs of all requirements are True by default?

RubenRBS:
Yep! [It was not properly documented until a few days ago](https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-type-trait-inferring)

PR #17525, opened 2023-05-11
Enabled source copy on PhysX library to fix local patch application

zillemarco:
Specify library name and version:  **physx**

After updating the PhysX recipe, I tried to use it locally and saw that the first time it compiled locally it run without issues.
The second run, after simply changing the configuration from `debug` to `release`, the recipe gave me the following error:

```
ERROR: physx/4.1.1: Error in build() method, line 146
        self._patch_sources()
while calling '_patch_sources', line 169
        replace_in_file(self, os.path.join(physx_source_cmake_dir, "CMakeLists.txt"),
        ConanException: replace_in_file didn't find pattern 'SET(CMAKE_POSITION_INDEPENDENT_CODE ON)' in 'C:\Users\Marco\.conan2\p\physx3a9a3bf0a00f2\s\src\physx\source\compiler\cmake\CMakeLists.txt' file.
```

After some troubleshooting (thanks a lot `sam` on Slack), it turned out that the patches were being applied on the "main" copy of the source, which worked the first time but would error out after because `replace_in_file` is used in strict mode.

---


prince-chrismc:
Can you please hit the linter comments too get some extra value from a pr :)

prince-chrismc (approved):

SSE4 (approved):

RubenRBS (approved):

SpaceIm:
Why is it not auto merged?

RubenRBS:
The bot thinks that some checks did not pass, but it's puzzling me _which_ ones, as it seems they all are ok 🤔

PR #17528, opened 2023-05-11
Add arm-gnu-toolchain

kammce:
Specify library name and version:

    - arm-gnu-toolchain/11.3.0
    - arm-gnu-toolchain/12.2.0

New addition to ConanCenter.

The author of the toolchain binaries is ARM inc. Original files can be found here:
https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads

Makes the bare metal ARM GCC toolchain accessible via "tool_requires" for bare metal projects.

---


jwillikers (changes requested):
Good work. We use a similar package at my work. I wasn't sure if it would be worth waiting for a general GCC package built from source, because then we could use that to build such a package. I don't think that recipe is configurable enough for this yet, and who knows how long that might take. Having a package for the Arm GNU toolchain has been really helpful and having it on Conan Center should make things even easier for embedded development.

recipes/gnu-arm-embedded-toolchain/all/conandata.yml
```diff
@@ -0,0 +1,39 @@
+sources:
+  "11.3.0":
+    "Linux":
+      "x86_64":
+        url: "https://github.com/libhal/gnu-arm-embedded-toolchain/releases/download/v11.3/arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi.tar.xz"
```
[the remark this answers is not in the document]
kammce: Oh you are right. Yes, this is supposed to be from their website not mine. This looks like an older version of what I was using. I'll change this ASAP.

recipes/gnu-arm-embedded-toolchain/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
+from conan import ConanFile
+from conan.tools.files import get, copy
+import os
+
+
+required_conan_version = ">=1.50.0"
+
+
+class GnuArmEmbeddedToolchain(ConanFile):
+    name = "gnu-arm-embedded-toolchain"
```
jwillikers, line 10: I believe the name was changed to the Arm GNU Toolchain.

```suggestion
    name = "arm-gnu-toolchain"
```

jwillikers (changes requested):

recipes/gnu-arm-embedded-toolchain/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+            sha256=download_info["sha256"])
+
+        '''
+        NOTE: I'm not sure how exactly I should go about handling this:
+        The link is:
+        https://developer.arm.com/GetEula?Id=2821586b-44d0-4e75-a06d-4279cd97eaae
+
+        But it returns a number of licenses concatenated together as HTML.
+        '''
+        # get(self, self.conan_data["sources"]
```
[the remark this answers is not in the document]
kammce: They all look to be exactly the same. But I'll try to make the license key work in the conandata.yml file.

recipes/gnu-arm-embedded-toolchain/all/toolchain.cmake
```diff
@@ -0,0 +1,177 @@
...
+# ------------------------------------------------------------------------------
+set(CORTEX_M4F_FLAGS -mcpu=cortex-m4 -mthumb -mtpcs-frame -mtpcs-leaf-frame
+  -mfloat-abi=softfp -mfpu=fpv4-sp-d16)
+
+set(CORTEX_M4_FLAGS -mcpu=cortex-m4 -mthumb -mtpcs-frame -mtpcs-leaf-frame
+  -mfloat-abi=soft)
+
+# ------------------------------------------------------------------------------
+# Define Sub Routines
+# ------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
kammce: A lot of this is copied from my own project, but I think making this minimalist would be best.

recipes/gnu-arm-embedded-toolchain/all/toolchain.cmake
```diff
@@ -0,0 +1,177 @@
...
+# -Og   Enables optimizations that do not interfere with debugging.
+# -g    Produce debugging information in the operating system’s native format.
+set(CMAKE_C_FLAGS_DEBUG "-O0 -g"
+  CACHE INTERNAL "C Compiler options for debug build type")
+set(CMAKE_CXX_FLAGS_DEBUG "-O0 -g"
+  CACHE INTERNAL "C++ Compiler options for debug build type")
+set(CMAKE_ASM_FLAGS_DEBUG "-g"
+  CACHE INTERNAL "ASM Compiler options for debug build type")
+set(CMAKE_EXE_LINKER_FLAGS_DEBUG ""
+  CACHE INTERNAL "Linker options for debug build type")
```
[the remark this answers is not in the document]
kammce: Removed.

recipes/gnu-arm-embedded-toolchain/all/toolchain.cmake
```diff
@@ -0,0 +1,177 @@
...
+# Options for RELEASE build
+# -Os   Optimize for size. -Os enables all -O2 optimizations.
+set(CMAKE_C_FLAGS_RELEASE "-Os -g -DNDEBUG"
+  CACHE INTERNAL "C Compiler options for release build type")
+set(CMAKE_CXX_FLAGS_RELEASE "-Os -g -DNDEBUG"
+  CACHE INTERNAL "C++ Compiler options for release build type")
+set(CMAKE_ASM_FLAGS_RELEASE ""
+  CACHE INTERNAL "ASM Compiler options for release build type")
+set(CMAKE_EXE_LINKER_FLAGS_RELEASE ""
+  CACHE INTERNAL "Linker options for release build type")
```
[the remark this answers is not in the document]
kammce: Removed.

recipes/arm-gnu-toolchain/all/conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def build(self):
+        url = self.download_info["url"]
+        url = url + "?rev=" + self.download_info["rev"]
+        url = url + "&hash=" + self.download_info["hash"]
+        filename = os.path.basename(urlparse(url).path)
+
+        # using urlretrieve over conan.tools.files.get because get results in
+        # a 405 forbidden error whereas urlretrieve does not
+        urllib.request.urlretrieve(url, filename)
```
memsharded, line 55: This is problematic, because it will not be cached by the (work in progress) new cache for sources. It will work only for ``conan.tools.files import download/get``. So it is important to check why, and try to solve it, instead of workarounding that in the recipe.

recipes/arm-gnu-toolchain/all/test_package/conanfile.py
```diff
@@ -0,0 +1,23 @@
+from conan import ConanFile
+from conan.tools.cmake import CMake, cmake_layout
+
+
+class TestPackageConan(ConanFile):
+    settings = "os", "arch", "compiler", "build_type"
+    generators = "CMakeToolchain", "CMakeDeps", "VirtualBuildEnv"
+
+    def requirements(self):
```
memsharded, line 9: This needs to be ``build_requirements(self)`` method
```suggestion
    def build_requirements(self):
```

kammce:
> ### Conan v1 pipeline ❌
> Failure in build 13 (`c686510fe1667fddebb29027a28c0599aebbeac1`):
> 
> * **arm-gnu-toolchain/12.2.1@:**
>   CI failed to create some packages ([All logs](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/17528/13-macos-m1-clang/arm-gnu-toolchain/12.2.1//summary.json))
>   Logs for packageID `a5ad5696abf650a25eea8f377806b3d5fe234e6e`:
>   ```
>   [settings]
>   arch=armv8
>   build_type=Release
>   compiler=apple-clang
>   compiler.libcxx=libc++
>   compiler.version=13.0
>   os=Macos
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   ```
>   [...]
>   [env]
>   [conf]
>   tools.system.package_manager:mode=install
>   tools.system.package_manager:sudo=True
>   tools.apple:sdk_path=/Applications/conan/xcode/13.0/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk
>   
>   arm-gnu-toolchain/12.2.1 (test package): Installing package
>   Requirements
>       arm-gnu-toolchain/12.2.1 from local cache - Cache
>   Packages
>       arm-gnu-toolchain/12.2.1:a5ad5696abf650a25eea8f377806b3d5fe234e6e - Cache
>   
>   Cross-build from 'Macos:x86_64' to 'Macos:armv8'
>   Installing (downloading, building) binaries...
>   arm-gnu-toolchain/12.2.1: Already installed!
>   arm-gnu-toolchain/12.2.1 (test package): Generator 'VirtualBuildEnv' calling 'generate()'
>   arm-gnu-toolchain/12.2.1 (test package): Generator 'CMakeToolchain' calling 'generate()'
>   arm-gnu-toolchain/12.2.1 (test package): Preset 'release' added to CMakePresets.json. Invoke it manually using 'cmake --preset release'
>   arm-gnu-toolchain/12.2.1 (test package): If your CMake version is not compatible with CMakePresets (<3.19) call cmake like: 'cmake <path> -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/conan_toolchain.cmake -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=Release'
>   arm-gnu-toolchain/12.2.1 (test package): Generator txt created conanbuildinfo.txt
>   arm-gnu-toolchain/12.2.1 (test package): Generator 'CMakeDeps' calling 'generate()'
>   arm-gnu-toolchain/12.2.1 (test package): Aggregating env generators
>   arm-gnu-toolchain/12.2.1 (test package): Generated conaninfo.txt
>   arm-gnu-toolchain/12.2.1 (test package): Generated graphinfo
>   Using lockfile: '/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/conan.lock'
>   Using cached profile from lockfile
>   [HOOK - conan-center.py] pre_build(): [FPIC MANAGEMENT (KB-H007)] 'fPIC' option not found
>   [HOOK - conan-center.py] pre_build(): [FPIC MANAGEMENT (KB-H007)] OK
>   arm-gnu-toolchain/12.2.1 (test package): Calling build()
>   
>   ----Running------
>   > . "/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/conanbuild.sh" && cd '/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7' && cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE="Release" -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_SYSTEM_NAME="Darwin" -DCONAN_IN_LOCAL_CACHE="OFF" -DCONAN_COMPILER="apple-clang" -DCONAN_COMPILER_VERSION="13.0" -DCONAN_LIBCXX="libc++" -DCMAKE_INSTALL_PREFIX="/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/package" -DCMAKE_INSTALL_BINDIR="bin" -DCMAKE_INSTALL_SBINDIR="bin" -DCMAKE_INSTALL_LIBEXECDIR="bin" -DCMAKE_INSTALL_LIBDIR="lib" -DCMAKE_INSTALL_INCLUDEDIR="include" -DCMAKE_INSTALL_OLDINCLUDEDIR="include" -DCMAKE_INSTALL_DATAROOTDIR="share" -DCMAKE_PREFIX_PATH="/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7" -DCMAKE_MODULE_PATH="/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7" -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY="ON" -DCONAN_EXPORTED="1" -Wno-dev '/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package'
>   -----------------
>   -- The CXX compiler identification is unknown
>   -- Detecting CXX compiler ABI info
>   -- Detecting CXX compiler ABI info - failed
>   -- Configuring incomplete, errors occurred!
>   See also "/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/CMakeFiles/CMakeOutput.log".
>   See also "/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/CMakeFiles/CMakeError.log".
>   CMake Error at CMakeLists.txt:18 (target_compile_features):
>     target_compile_features no known features for CXX compiler
>   
>     ""
>   
>     version .
>   
>   
>   ERROR: arm-gnu-toolchain/12.2.1 (test package): Error in build() method, line 14
>   	cmake.configure()
>   	ConanException: Error 1 while executing cd '/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7' && cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE="Release" -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_SYSTEM_NAME="Darwin" -DCONAN_IN_LOCAL_CACHE="OFF" -DCONAN_COMPILER="apple-clang" -DCONAN_COMPILER_VERSION="13.0" -DCONAN_LIBCXX="libc++" -DCMAKE_INSTALL_PREFIX="/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7/package" -DCMAKE_INSTALL_BINDIR="bin" -DCMAKE_INSTALL_SBINDIR="bin" -DCMAKE_INSTALL_LIBEXECDIR="bin" -DCMAKE_INSTALL_LIBDIR="lib" -DCMAKE_INSTALL_INCLUDEDIR="include" -DCMAKE_INSTALL_OLDINCLUDEDIR="include" -DCMAKE_INSTALL_DATAROOTDIR="share" -DCMAKE_PREFIX_PATH="/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7" -DCMAKE_MODULE_PATH="/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package/build/952c770f7b0a1c0ec0ac98ce7d64e32207bd0ce7" -DCMAKE_EXPORT_NO_PACKAGE_REGISTRY="ON" -DCONAN_EXPORTED="1" -Wno-dev '/Users/jenkins/w/prod/BuildSingleReference@3/cci-aeda1cc0/recipes/arm-gnu-toolchain/all/test_v1_package'
>   ```
> * **arm-gnu-toolchain/11.3.0@:**
>   Didn't run or was cancelled before finishing
> 
> **Note:** To save resources, CI tries to finish as soon as an error is found. For this reason you might find that not all the references have been launched or not all the configurations for a given reference. Also, take into account that we cannot guarantee the order of execution as it depends on CI workload and workers availability.
> 
> ### Conan v2 pipeline ❌
> > Note: Conan v2 builds may be required once they are on the [v2 ready](https://github.com/conan-io/conan-center-index/blob/master/.c3i/conan_v2_ready_references.yml) list
> 
> The v2 pipeline failed. Please, review the errors and note this will be required for pull requests to be merged in the near future.
> 
> See details:
> Failure in build 13 (`c686510fe1667fddebb29027a28c0599aebbeac1`):
> 
> * **arm-gnu-toolchain/11.3.0@:**
>   Didn't run or was cancelled before finishing
> * **arm-gnu-toolchain/12.2.1@:**
>   CI failed to create some packages ([All logs](https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/17528/13-windows-msvc/arm-gnu-toolchain/12.2.1//summary.json))
>   Logs for packageID `522dcea5982a3f8a5b624c16477e47195da2f84f`:
>   ```
>   [settings]
>   arch=x86_64
>   build_type=Release
>   compiler=msvc
>   compiler.cppstd=14
>   compiler.runtime=dynamic
>   compiler.runtime_type=Release
>   compiler.version=192
>   os=Windows
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   ```
>   [...]
>   build_type=Release
>   compiler=msvc
>   compiler.cppstd=14
>   compiler.runtime=dynamic
>   compiler.runtime_type=Release
>   compiler.version=192
>   os=Windows
>   [conf]
>   tools.system.package_manager:mode=install
>   tools.system.package_manager:sudo=True
>   
>   
>   ======== Computing dependency graph ========
>   Graph root
>       cli
>   Requirements
>       arm-gnu-toolchain/12.2.1#55447aecef04629865c5c9726784d5ac - Cache
>   
>   ======== Computing necessary packages ========
>   arm-gnu-toolchain/12.2.1: Forced build from source
>   Requirements
>       arm-gnu-toolchain/12.2.1#55447aecef04629865c5c9726784d5ac:522dcea5982a3f8a5b624c16477e47195da2f84f - Build
>   
>   ======== Installing packages ========
>   arm-gnu-toolchain/12.2.1: Calling source() in C:\J\w\prod-v2\BuildSingleReference@2\p\arm-gb753c0d6403c0\s
>   
>   -------- Installing package arm-gnu-toolchain/12.2.1 (1 of 1) --------
>   arm-gnu-toolchain/12.2.1: Building from source
>   arm-gnu-toolchain/12.2.1: Package arm-gnu-toolchain/12.2.1:522dcea5982a3f8a5b624c16477e47195da2f84f
>   arm-gnu-toolchain/12.2.1: Copying sources to build folder
>   arm-gnu-toolchain/12.2.1: Building your package in C:\J\w\prod-v2\BuildSingleReference@2\p\t\arm-g62233d1b2e36b\b
>   arm-gnu-toolchain/12.2.1: Generating aggregated env files
>   arm-gnu-toolchain/12.2.1: Generated aggregated env files: ['conanbuild.bat', 'conanrun.bat']
>   arm-gnu-toolchain/12.2.1: Calling build()
>   Downloading arm-gnu-toolchain-12.2.rel1-mingw-w64-i686-arm-none-eabi.zip
>   arm-gnu-toolchain/12.2.1: Unzipping 1.25GB, this can take a while
>   
>   arm-gnu-toolchain/12.2.1: ERROR: 
>   Package '522dcea5982a3f8a5b624c16477e47195da2f84f' build failed
>   arm-gnu-toolchain/12.2.1: WARN: Build folder C:\J\w\prod-v2\BuildSingleReference@2\p\t\arm-g62233d1b2e36b\b
>   *********************************************************
>   Recipe 'arm-gnu-toolchain/12.2.1' cannot build its binary
>   It is possible that this recipe is not Conan 2.0 ready
>   If the recipe comes from ConanCenter check: https://conan.io/cci-v2.html
>   If it is your recipe, check if it is updated to 2.0
>   *********************************************************
>   
>   ERROR: arm-gnu-toolchain/12.2.1: Error in build() method, line 57
>   	urllib.request.urlretrieve(self.license_url, "LICENSE")
>   	URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1091)>
>   ```
> 
> **Note:** To save resources, CI tries to finish as soon as an error is found. For this reason you might find that not all the references have been launched or not all the configurations for a given reference. Also, take into account that we cannot guarantee the order of execution as it depends on CI workload and workers availability.

The error is 

```
ERROR: arm-gnu-toolchain/12.2.1: Error in build() method, line 57
	urllib.request.urlretrieve(self.license_url, "LICENSE")
	URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1091)>
```

Hmmm not sure why this is happening. @memsharded @jwillikers, may I simply download the license to the recipe folder and copy it from source from the `all` directory.

recipes/arm-gnu-toolchain/all/conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def package_info(self):
+        self.cpp_info.includedirs = []
+
+        bin_folder = os.path.join(self.package_folder, "bin/bin")
+        self.cpp_info.bindirs = [bin_folder]
+        self.buildenv_info.append_path("PATH", bin_folder)
+
+        self.conf_info.define(
+            "tools.cmake.cmaketoolchain:system_name", "GENERIC")
```
jwillikers, line 86:
```suggestion
            "tools.cmake.cmaketoolchain:system_name", "Generic")
```

Just to follow the case used for the platform file: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Generic.cmake

jwillikers:
> 

Looks like putting the license in the recipe makes the recipe too large. That SSL error might have to do with how Conan manages SSL certificates. Is it possible you could use [conan.tools.files.download()](https://docs.conan.io/2/reference/tools/files/downloads.html#conan-tools-files-download) instead to download the license?

kammce:
> > 
> 
> Looks like putting the license in the recipe makes the recipe too large. That SSL error might have to do with how Conan manages SSL certificates. Is it possible you could use [conan.tools.files.download()](https://docs.conan.io/2/reference/tools/files/downloads.html#conan-tools-files-download) instead to download the license?

Using this before results in a 403 Forbidden error. Its why I didn't use it originally. 

```
ERROR: arm-gnu-toolchain/12.2.1: Error in build() method, line 53
        download(self, self.license_url, "LICENSE")
        AuthenticationException: 403: Forbidden
```

Do you know how to get around this? @jwillikers

jwillikers:
> > > 
> > 
> > 
> > Looks like putting the license in the recipe makes the recipe too large. That SSL error might have to do with how Conan manages SSL certificates. Is it possible you could use [conan.tools.files.download()](https://docs.conan.io/2/reference/tools/files/downloads.html#conan-tools-files-download) instead to download the license?
> 
> Using this before results in a 403 Forbidden error. Its why I didn't use it originally.
> 
> ```
> ERROR: arm-gnu-toolchain/12.2.1: Error in build() method, line 53
>         download(self, self.license_url, "LICENSE")
>         AuthenticationException: 403: Forbidden
> ```
> 
> Do you know how to get around this? @jwillikers

No, but I have a bad feeling about it. It might be forbidding requests coming from the Conan Center CI servers or it may be an issue in the Conan tools download method. I wish they could just include the license in the tarballs. We're probably going to need input from the Conan team on what to do about this. For now, maybe see if you can get things working without the license file, i.e. use a dummy license file, or perhaps just mirror the license somewhere else for now?

kammce:
🤔 Seems that attempting to write an application that can test cross building is problematic. Its hard to convince CMake/Conan to compile using a different compiler when the host/build targets make it very clear you aren't cross compiling. Do we test against baremetal in CI? That would be best and all other targets should simply be skipped. What do you think? (I'll start working on that in the next commit)

jwillikers:
> thinking Seems that attempting to write an application that can test cross building is problematic. Its hard to convince CMake/Conan to compile using a different compiler when the host/build targets make it very clear you aren't cross compiling. Do we test against baremetal in CI? That would be best and all other targets should simply be skipped. What do you think? (I'll start working on that in the next commit)

Yes, that's exactly what needs to happen. It's just that it is a bit of a can of worms since then we actually need sensible architecture settings to represent the different instruction sets and hardware configurations for embedded Arm. I've had to implement this in house. Here's a snippet from the `settings.yaml`.


```yaml
arch:
  thumbv7em:
    float-abi: ["soft", "softfp", "hard"]
    fpu:
      vfpv4:
      fpv5:
        precision: ["single", "double"]
    processor: [None, "cortex-m0", "cortex-m4", "cortex-m7"]
```

I think this is pretty self-explanatory, in that it defines things for the Arm embedded ISA, version 7, using the `thumb` format of course, since for embedded no one actually seems to use the larger Arm format. I also defined things for the floating-point unit, floating-point ABI, and the specific processor. These are all translatable to architecture compiler flags for GCC. I used these sub-settings as a way to generate the architecture compiler flags as part of the CMakeToolchain before `tools.build.*flags` were available to inject arbitrary compiler flags into the build and also keep binary hashes distinct. So, we might just need something like `thumbv7em` added to Conan's architecture list in the default `settings.yaml`. Re-using `armv7*` architectures for embedded MCUs is out of the question for me, it's a completely different ISA.

Once that is available, it should be possible to add a profile to Conan Center for that architecture with compiler flags and probably a linker script which targets a specific MCU. I'd recommend the Raspberry Pi Pico due to its popularity, affordability, and the fact that it has a nice open-source C++ SDK. That would make a good Conan package someday. ;)

Once that profile is available, then your recipe will actually be able to build. You'll be able to add checks in the `validate()` method that verify both the _build_ and _host_ architectures for the build. This would look something like the following.

```python
def validate(self):
    if not hasattr(self, "settings_build"):
        raise ConanInvalidConfiguration(f"{self.name} must be used with a build profile.")

    supported_build_operating_systems = ["Linux", "Macos", "Windows"]
    if not self.settings_build.os in supported_build_operating_systems:
            f"The build os '{self.settings_build.os}' is not supported. "
            f"Pre-compiled binaries are only available for {supported_build_operating_systems}."
        )

    supported_build_architectures = {
        "Linux": ["armv8", "x86_64"],
        "Macos": ["armv8", "x86_64"],
        "Windows": ["x86", "x86_64"],
    }
    if (
        not self.settings_build.arch
        in supported_build_architectures[str(self.settings_build.os)]
    ):
            f"The build architecture '{self.settings_build.arch}' is not supported for {self.settings_build.os}. "
            f"Pre-compiled binaries are only available for {supported_build_architectures[str(self.settings_build.os)]}."
        )

    supported_host_operating_systems = ["baremetal"]
    if (
        not self.settings_target.os
        in supported_host_operating_systems
    ):
            f"The target host architecture '{self.settings_target.os}' is not supported for {self.settings_build.os}. "
            f"Pre-compiled binaries are only available for {supported_host_operating_systems[self.settings_build.os]}."
        )

    supported_host_architectures = ["thumbv7em"]
    if not self.settings_target.arch in supported_host_architectures:
            f"The build os '{self.settings_target.arch}' is not supported. "
            f"Pre-compiled binaries are only available for {supported_host_architectures}."
        )
```

That should be on the right track to only build for embedded targets on supported build operating systems and architectures.

Now, there's also the big problem that introducing an MCU profile to all of the recipe builds on Conan Center is likely to break recipes which don't inclusively define the supported OS or architecture or properly configure things for a baremetal target. Maybe adding the new profile will just require recipes to account for that when they are being updated or maybe some infrastructure changes would be done to address this. It would be great to have embedded builds of some kind for libraries like `fmt` in CI.

If we manage to get past the necessary hurdles in Conan and Conan Center for this recipe, it should be much easier to start adding recipes in Conan Center targeting MCU's, i.e. FreeRTOS, the Raspberry Pi Pico SDK, etc.

kammce:
> I think this is pretty self-explanatory, in that it defines things for the Arm embedded ISA, version 7, using the thumb format of course, since for embedded no one actually seems to use the larger Arm format. I also defined things for the floating-point unit, floating-point ABI, and the specific processor. These are all translatable to architecture compiler flags for GCC. I used these sub-settings as a way to generate the architecture compiler flags as part of the CMakeToolchain before tools.build.*flags were available to inject arbitrary compiler flags into the build and also keep binary hashes distinct. So, we might just need something like thumbv7em added to Conan's architecture list in the default settings.yaml. Re-using armv7* architectures for embedded MCUs is out of the question for me, it's a completely different ISA.

I'm still learning conan and figuring out the right ways to do things. I'm guessing that we would construct this tool package to ensure that changing the architecture settings doesn't affect the package id which would be annoying to have to download the same toolchain for each and every possible MCU configuration. Something like

```python
    def package_id(self):
        del self.info.settings.compiler
        del self.info.settings.build_type
        del self.info.settings.arch
```

> Once that is available, it should be possible to add a profile to Conan Center for that architecture with compiler flags and probably a linker script which targets a specific MCU. I'd recommend the Raspberry Pi Pico due to its popularity, affordability, and the fact that it has a nice open-source C++ SDK. That would make a good Conan package someday. ;)

We can actually just omit the linker script. At least for GCC, the toolchain will provide an arbitrary default linker script for you.  I haven't gotten a chance to get into board bring up levels of work with the Pico, but with its dual core and other IP blocks, I could see this being overly complicated when we can omit it. 

> Once that profile is available, then your recipe will actually be able to build. You'll be able to add checks in the validate() method that verify both the build and host architectures for the build. This would look something like the following.

That sounds great! Currently I'm using options for a lot of this, but I believe using settings would be more clear. 

> Now, there's also the big problem that introducing an MCU profile to all of the recipe builds on Conan Center is likely to break recipes which don't inclusively define the supported OS or architecture or properly configure things for a baremetal target. Maybe adding the new profile will just require recipes to account for that when they are being updated or maybe some infrastructure changes would be done to address this. It would be great to have embedded builds of some kind for libraries like fmt in CI.
> 
> If we manage to get past the necessary hurdles in Conan and Conan Center for this recipe, it should be much easier to start adding recipes in Conan Center targeting MCU's, i.e. FreeRTOS, the Raspberry Pi Pico SDK, etc.

Totally agree! Which is why I'm trying to get all of this infrastructure in so I can bring up my project and kick off the conversation around baremetal targets 😄

kammce:
Actually to clarify, I use options for other things, specifically getting particular libraries to emit their linker scripts to the application linker. I would use mcu specific profiles to pass the linker script but it causes issues with CMake when it tries to validate that the compiler works and the linker script gets injected there. I can technically make it work but it gets a bit complicated because I have another library that provides linker script templates that also need to be put into the command line path. All happens automatically when I build an actual application with all of the linked libraries, but does not for that initial "test that the compiler works" stage.

jwillikers:
The `package_id` is a bit of a quandary, since when the package is actually used properly as a `tool_requires`, the settings will evaluate correctly to those of the build profile. So I don't think `arch` and `os` need to be set. Now that I look at the recipe I made, I overrode things in the `generate` method of the `conanfile.py` of the test package to get it to test building for an MCU and not something for the build settings. Here is what that would probably look like for the test package's `conanfile.py`.

```python
    def generate(self):
        tc = CMakeToolchain(self)
        tc.variables["CMAKE_TRY_COMPILE_TARGET_TYPE"] = "STATIC_LIBRARY"
        tc.blocks["arch_flags"].values = {
            "arch_flag": "-mthumb -mfloat-abi=hard -march=armv7e-m+fp -mtune=cortex-m4"
        }
        tc.variables[
            "CMAKE_EXE_LINKER_FLAGS_INIT"
        ] = "${{CMAKE_EXE_LINKER_FLAGS_INIT}} -specs=nano.specs -specs=nosys.specs"
        tc.generate()
```

You could probably override the linker flags via the `conf` `tools.build.linkerflags` variable instead of overriding the CMake variable directly, but I'm pretty sure that the `arch_flag` will need to overridden like this to avoid using an incorrect flag for the build architecture.

So I guess I didn't actually use a profile for an MCU for the `test_package` to build properly, so maybe we can get away without an MCU profile in Conan Center to be able to create this particular package.

kammce:
I'll do some tests with this package and see how it works out :)

kammce:
> The `package_id` is a bit of a quandary, since when the package is actually used properly as a `tool_requires`, the settings will evaluate correctly to those of the build profile. So I don't think `arch` and `os` need to be deleted. Now that I look at the recipe I made, I overrode things in the `generate` method of the `conanfile.py` of the test package to get it to test building for an MCU and not something for the build settings. Here is what that would probably look like for the test package's `conanfile.py`.
> 
> ```python
>     def generate(self):
>         tc = CMakeToolchain(self)
>         tc.variables["CMAKE_TRY_COMPILE_TARGET_TYPE"] = "STATIC_LIBRARY"
>         tc.blocks["arch_flags"].values = {
>             "arch_flag": "-mthumb -mfloat-abi=hard -march=armv7e-m+fp -mtune=cortex-m4"
>         }
>         tc.variables[
>             "CMAKE_EXE_LINKER_FLAGS_INIT"
>         ] = "${{CMAKE_EXE_LINKER_FLAGS_INIT}} -specs=nano.specs -specs=nosys.specs"
>         tc.generate()
> ```
> 
> You could probably override the linker flags via the `conf` `tools.build.linkerflags` variable instead of overriding the CMake variable directly, but I'm pretty sure that the `arch_flag` will need to overridden like this to avoid using an incorrect flag for the build architecture.
> 
> So I guess I didn't actually use a profile for an MCU for the `test_package` to build properly, so maybe we can get away without an MCU profile in Conan Center to be able to create this particular package.

So to be clear, is this something that you would like this PR to do for now to get the test_package building somewhat for now? Would that be enough to get this package towards approval or would we still need to wait for the new settings to be added to conan?

jwillikers:
I think you are going to need to change the test package to work that since the toolchain package will be used as a tool_requires, which will result in the `build` and `host` profiles always being the same, i.e. the build profile. That's how `tool_requires` work. That will result in the test package using the build profile and not the profile for a target.... which is why this would be necessary.

kammce:
I'm playing around with this and it seems that `validate()` is having trouble since the `test_package` has the profile the same as the build. Meaning we will fail because the `os != "baremetal"`.

jwillikers:
> The `validate()` function you provided, that I'm playing with has the issue that it stop if the profile host doesn't match baremetal. Is there a way in a recipe to force the profile to be a specific thing? Because we cannot even get to the CMakeToolchain step until we make that work.

I should have told you to remove the portions using `self.settings_target` since that won't work when Conan Center CI runs `conan create`. The validate method should look like the following instead.

```python
def validate(self):
    if not hasattr(self, "settings_build"):
        raise ConanInvalidConfiguration(f"{self.name} must be used with a build profile.")

    supported_build_operating_systems = ["Linux", "Macos", "Windows"]
    if not self.settings_build.os in supported_build_operating_systems:
            f"The build os '{self.settings_build.os}' is not supported. "
            f"Pre-compiled binaries are only available for {supported_build_operating_systems}."
        )

    supported_build_architectures = {
        "Linux": ["armv8", "x86_64"],
        "Macos": ["armv8", "x86_64"],
        "Windows": ["x86", "x86_64"],
    }
    if (
        not self.settings_build.arch
        in supported_build_architectures[str(self.settings_build.os)]
    ):
            f"The build architecture '{self.settings_build.arch}' is not supported for {self.settings_build.os}. "
            f"Pre-compiled binaries are only available for {supported_build_architectures[str(self.settings_build.os)]}."
        )
```

kammce:
Yup, those are correct. That part works just fine. I'll try again.

jwillikers:
Okay, so use this to get things to work with a single profile:

```sh
@property
def _settings_build(self):
    return getattr(self, "settings_build", self.settings)

def validate(self):
    supported_build_operating_systems = ["Linux", "Macos", "Windows"]
    if not self._settings_build.os in supported_build_operating_systems:
            f"The build os '{self._settings_build.os}' is not supported. "
            f"Pre-compiled binaries are only available for {supported_build_operating_systems}."
        )

    supported_build_architectures = {
        "Linux": ["armv8", "x86_64"],
        "Macos": ["armv8", "x86_64"],
        "Windows": ["x86", "x86_64"],
    }
    if (
        not self._settings_build.arch
        in supported_build_architectures[str(self._settings_build.os)]
    ):
            f"The build architecture '{self._settings_build.arch}' is not supported for {self._settings_build.os}. "
            f"Pre-compiled binaries are only available for {supported_build_architectures[str(self._settings_build.os)]}."
        )
```

recipes/arm-gnu-toolchain/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,31 @@
+from conans import ConanFile, CMake, CMakeToolchain
```
jwillikers, line 1: There is no `CMakeToolchain` in `conans`. That might be trickier to get this working with a V1 test package. I think you can now drop the `test_v1_package` since it is after May 15th, see https://github.com/conan-io/conan-center-index/discussions/17406

jwillikers:
> @jwillikers any suggestions on how to force MSVC + CMake to use this compiler rather than the MSVS compiler?

I bet it is sourcing the generated build environment script which sets up everything to use the MSVC compiler tool. You'd need to probably avoid using the build environment in the test package, maybe just for MSVC.

kammce:
This is strange. Now the issue is Conan. It says, after building and installing the package that it cannot find the package prebuilt and isn't looking at the cache correctly. I do see this issue randomly pop up when cross building other packages, but wasn't sure exactly what causes it. My guess was the wrong profile ID is picked for an older version of the packages that were cross built. But this only has 1 package. So I'm not sure whats happening. 

My test device is an M1/arm64 macbook and it works fine for me. I'm not exactly sure what the delta is between this package  build and whats on my machine.

kammce:
Oh! I see the issue. I need to forbid this:

```

Configuration (profile_host):
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.libcxx=libc++
compiler.version=13.0
os=Macos
[options]
[build_requires]
[env]
[conf]
tools.system.package_manager:sudo=True
tools.apple:sdk_path=/Applications/conan/xcode/13.0/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk

Configuration (profile_build):
[settings]
arch=x86_64
build_type=Release
compiler=apple-clang
compiler.libcxx=libc++
compiler.version=13.0
os=Macos
[options]
[build_requires]
[env]
[conf]
tools.system.package_manager:sudo=True
tools.apple:sdk_path=/Applications/conan/xcode/13.0/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk
```

We don't need to cross build from x86 to arm64. Not sure whats the best way to prevent this case from executing.

kammce:
I apologize for the number of commits I pushing out. It works just fine on my machine but for some reason fails on CI so I'm trying to adjust but I'm not exactly sure what to do to make it happy.

kammce:
@jwillikers @memsharded I think this is the right way going forward. We don't need to consider profiles that set the compiler to anything else than GCC. Unfortunately that doesn't include Windows because the profile only uses Visual Studio. But I've tested this same package on windows machines in the past and it worked just fine. It may be good to add more entries for clang & gcc for Windows as well. Can use https://chocolatey.org/ to do the installation of the compilers.

kammce:
I would like to clarify that you cannot add architecture flags to the `cflags`, `cxxflags`, or `linker flags`. The only flags that can be added to a package are flags that are intended for every ARM build, such as `-mthumb`. I have never heard of a reason to use standard 32-bit ARM instructions, as they would increase the binary size for minimal benefit. We cannot allow the host's architecture to define the compiler flags, as convenient as that might be, because tool requirements can only see the build system settings, not the host settings. This limitation is beneficial, and we should not try to modify Conan's `tool_requires` to adjust its `package_id` based on the host settings. This would result in a package for each possible configuration of the ARM Cortex M processor and floating point unit. If we cached all these packages in Artifactory, we would have hundreds of 2GB binary packages, most of which would contain identical binaries. This would result in a waste of space.

To provide compiler flags and the appropriate architectural settings to differentiate packages, we can distribute Conan profile fragments as follows: https://github.com/libhal/libhal-armcortex/commit/973d91479c9016c2e0f87b3bdab590e7ecc3c21a.

## `cortex-m` fragment
```
[settings]
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++
compiler.version=12.2
os=baremetal
arch=thumbv7em
arch.float_abi={{ float_abi }}
arch.processor={{ cpu }}

[tool_requires]
arm-gnu-toolchain/12.2.1

[conf]
tools.build:cflags=["-mfloat-abi={{ float_abi }}", "-mcpu={{ cpu }}"]
tools.build:cxxflags=["-mfloat-abi={{ float_abi }}", "-mcpu={{ cpu }}"]
tools.build:exelinkflags=["--specs=nano.specs", "--specs=nosys.specs", "-mfloat-abi={{ float_abi }}", "-mcpu={{ cpu }}"]
```

## `cortex-m0plus` profile
```
{% set cpu = "cortex-m0plus" %}
{% set float_abi = "soft" %}
{% include "cortex-m" %}
```

Then, an MCU can provide its full profile like this: https://github.com/libhal/libhal-lpc40/commit/1542b3f290e8c0f1a451101f7a88641abfa79a9b. Users can then execute the following commands:

```bash
conan config install -sf conan/profiles/ -tf profiles https://github.com/libhal/libhal-armcortex.git
conan config install -sf conan/profiles/ -tf profiles https://github.com/libhal/libhal-lpc40.git
conan build app -pr lpc4078 -s build_type=Debug
```

Users can also add or create their own profiles from scratch or modify the fragments in different ways, if needed. I believe this is an elegant solution for providing global settings, configuration options, and compiler flags outside of the package. I've been working to implement this solution for my platform, and it's been working quite well.

This method does have limitations. For instance, I would have to consider a version of these fragments where the user can choose the compiler version (11.3 or 12.2 or whatever the future holds). Currently, this does not support Clang, but we can add additional profiles for different compilers, as they will likely require different compiler flags.

kammce:
@memsharded @jwillikers Sorry to bother you too again, but could this get another look? And also see my post https://github.com/conan-io/conan-center-index/pull/17528#issuecomment-1577796846 regarding compiler flags.

jwillikers (approved):

kammce:
I was able to talk to someone who works at ARM to get their 12.3.0 toolchain to have the license information embedded in the zip file. I'll remove version 11.3.0 and 12.2.1 since they do not contain a license file.

danimtb (dismissed):

kammce:
Sorry to delay this PR any further. I know this is PR has been around for quite a while. But I'd like to do the followings things:

1. Add the `picolibc` project to this package. Picolibc has helped my application binaries reduce in size by 71% compared to the default newlib and newlib-nano stdlibc implementations. Those other implementations will still be supported.
2. Keep versions 11.3.0 and 12.2.1 for those that wish to use it. I know getting the URL from a github gist is a janky, but it works around the issue with ARM's EULA rendering system.
3. Make the `build_policy = "missing"`. I do this on my repository so I can simply upload the recipe and reduce on storage space. I'm not sure if conan index already does something like this.

I've already tested picolibc with my own version of this package I use for my package index, so it should work here as well.
I'm currently asking the picolibc maintainer if they can add the 12.3.0 version of their prebuilts to their github releases to make it downloadable via this package recipe.

jwillikers:
@kammce I'm all for adding picolibc. You'll need to update the license attribute to reflect that. It looks like everything the project ships is probably under variations of the BSD license.

kammce:
Good catch. I'll work on that now.

kammce:
I dug through the picolibc COPYING files to find all of the licenses. Many of them are custom bespoke licenses from the newlib project. All of the custom ones seem to be quite permissive. I add the following to the list:

```
 "GPL-2.0-only", "BSD-3-Clause", "BSD-2-Clause-FreeBSD", "AGPL-3.0-only", "BSD-2-Clause"
```

I don't know if that is too much and confusing to the license reader.

jwillikers (approved):

jwillikers:
> I dug through the picolibc COPYING files to find all of the licenses. Many of them are custom bespoke licenses from the newlib project. All of the custom ones seem to be quite permissive. I add the following to the list:
> 
> ```
>  "GPL-2.0-only", "BSD-3-Clause", "BSD-2-Clause-FreeBSD", "AGPL-3.0-only", "BSD-2-Clause"
> ```
> 
> I don't know if that is too much and confusing to the license reader.

Better to be accurate than succinct when it comes to licenses. They get messy sometimes , but consumers need to know about them.

kammce:
Alright, this looks good to me. It has everything that I wanted to put into it that passes CI. Ready for any needed changes for this PR.

jwillikers (approved):
LGTM!

kammce:
@RubenRBS anything else that needs to be updated on the PR?

RubenRBS:
Hi! Thanks for your patience :) I have scheduled a pair review with @danimtb for this PR this week as its topic is a bit outside my wheelhouse

recipes/arm-gnu-toolchain/all/test_package/conanfile.py
```diff
@@ -0,0 +1,24 @@
...
+    generators = "CMakeDeps", "CMakeToolchain", "VirtualBuildEnv"
+
+    def build_requirements(self):
+        self.tool_requires(self.tested_reference_str)
+
+    def layout(self):
+        cmake_layout(self)
+
+    def build(self):
+        if str(self.settings.compiler) == "gcc":
```
RubenRBS, line 16: This if smells iffy.
We need more insight into why this was put here, or at least logs of it working on Windows this is not tested by the CI (No mingw configurations are currently built, although it's a planned feature)

Also (and this is just me maybe missing something else) why is this check here and not on the validate() method of the recipe? Any usecase?

recipes/arm-gnu-toolchain/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+cmake_minimum_required(VERSION 3.15)
+
+# Required because Conan does not emit `set(CMAKE_SYSTEM_NAME "GENERIC")` or
+# `set(CMAKE_SYSTEM_PROCESSOR "ARM")` causing host flags to be leacked to CMake
+# in the conan_toolchain.cmake
+set(CMAKE_TOOLCHAIN_FILE ../toolchain.cmake)
```
RubenRBS, line 6: Why is the whole toolchain needed? This does not test the package on a standalone way because it includes the file from the CCI tree itself.

Also, is there no way to define the missing variables using the CMake helpers in Conan? I would have expected it to be doable

recipes/arm-gnu-toolchain/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+project(test_package LANGUAGES CXX)
+
+set(test_package_compiler_flags -mcpu=cortex-m0plus
+  -mfloat-abi=soft
+  --specs=nosys.specs
+  --specs=nano.specs
+  --specs=picolibc.specs)
+
+add_executable(${PROJECT_NAME} main.cpp)
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_20)
```
RubenRBS, line 17: Totally noob question here, but is this C++20 requirement only because of internal test package implementation, or does the toolchain itself expect it (I doubt it but better to confirm!)?

recipes/arm-gnu-toolchain/all/conanfile.py
```diff
@@ -0,0 +1,161 @@
...
+    @property
+    def license_url(self):
+        # All versions of arm-gnu-toolchain (previously called gnu-arm-embedded)
+        # after 12.3.1 will include their licenses in their package
+        # releases. This is a work around for versions of the compiler without
+        # licenses.
+        if (self.version == "11.3.0" or
+            self.version == "12.2.1" or
+            self.version == "12.3.1"):
+            return "https://gist.githubusercontent.com/kammce/dc566a05f6ab2787ceef5b706012e7a2/raw/4cb8ab752d7c0f87cc074afa4e548a2be8766210/EULA.html"
```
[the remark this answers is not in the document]
RubenRBS: Rereading the review comments, https://github.com/conan-io/conan-center-index/pull/17528#issuecomment-1680462750 you mention that you will remove the older versions. This seems to me the better approach, as this being a new recipe it's ok if it's missing old versions unless someone specifically has a good reason to request older one!

RubenRBS (changes requested):
Thanks a lot for your contribution! @danimtb and I have reviewed the recipe and have some concerns over it

kammce:
I'm not sure if the compiler should be validated in the toolchain recipe or in the test package. But for now, I placed them in the toolchain recipe.

recipes/arm-gnu-toolchain/all/conanfile.py
```diff
@@ -81,6 +81,18 @@ def validate(self):
                 f"{supported_build_architectures[build_os]}."
             )
 
+        compiler_name = str(self.settings.compiler)
+        if compiler_name != "gcc":
+            raise ConanInvalidConfiguration(
+                "The compiler must be set to gcc to use this toolchain.")
+
+        compiler_version = str(self.settings.compiler.version)
+        supported_compiler_versions = ["11.3", "12.2", "12.3"]
```
jwillikers, line 90: The consumers can also specify a major version of the compiler, i.e. `11` or `12`, to allow broader binary compatibility. It would be good to allow those cases as well.

```suggestion
        supported_compiler_versions = ["11", "11.3", "12", "12.2", "12.3"]
```

jwillikers (changes requested):

jwillikers (approved):

kammce:
Hmmm, back up against the build issue, where CMake cannot figure out which compiler to use. 🤔

kammce:
For now, I'll table working on this PR until I get some help from the conan team for how to make this toolchain work with CCI. I'm probably doing something incorrectly, but it looks like the "correct" way of doing things does not work, and I'm just throwing stuff at the wall hoping it works. 

Conan 1.x's CMake invocation seems to miss the user toolchain file and simply selects the wrong compiler.
Conan 2.x's CMake invocation seems to use flags based on the host's machine. We may have to wait until the arm baremetal settings are added to conan and CCI uses them for their builds. I forget the issue number, but I do recall that @jwillikers made the issue.

kammce:
Poke, still working on this, but have been busy with life.

jwillikers (approved):

jwillikers:
Not stale.

kammce:
Hmmm, maybe its a good idea to drop this PR. I believe Diego wanted to go a different way when it came to compiler tool packages in the index. I've been doing well using my own JFrog artifactory for my arm gnu toolchain. Not sure if there is still a desire to have such a thing in the conan center. If there is, I'll try working on this again. If not, we can close this.

jcar87:
Thanks @kammce kammce - thanks for your PR and happy to hear your needs are currently covered with your own Artifactory instance.
Indeed, as we discussed at CppCon and to provide further context - we currently have concerns about how the Conan 2 package types are suited to full toolchains rather than libraries or standalone executables. The issues that we've had with the existing gcc recipe (as well as the python one, for similar reasons) are enough  for us to consider a more cautious approach - rather than provide things that are difficult for us to maintain.

On the other hand, we'd happily accept a contribution in the `conan-recipes-contrib` repository - we have plans to start working on recipes isolated toolchains - including, if feasible, msvc / Xcode - as well as aiming to provide and document toolchains to be able to cross-compile conan center recipes (Linux being the main goal here) - and we will start there, so it may be that our development benefits from having similar things grouped together.

Thanks your patience on this!

PR #16404, opened 2023-03-06
qt5: conan V2 migration

ericLemanissier:
fixes conan-io/conan-center-index#16249
fixes conan-io/conan-center-index#16084
fixes conan-io/conan-center-index#16113

based on @rainman110's work

Specify library name and version:  **qt/5.***

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17539, opened 2023-05-12
[qt/6.x.x] V2 migration

MartinDelille:
- Remove qt/6.2.4
- [qt/6.x.x] Conan v2

---


jwillikers (approved):

jwillikers (approved):

ericLemanissier (approved):

jwillikers (approved):

MartinDelille:
@uilianries Would you have time to review this PR ?

recipes/qt/6.x.x/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+project(test_package)
+
+include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
+conan_set_vs_runtime()
+conan_set_libcxx()
+conan_output_dirs_setup()
+
+find_package(Qt6 COMPONENTS Core Network Sql Concurrent Xml REQUIRED CONFIG)
+set(CMAKE_AUTOMOC ON)
+set(CMAKE_AUTORCC ON)
```
nrbnlulu, line 12: duplicated

recipes/qt/6.x.x/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,145 @@
...
+Sysconf = bin/sysconfdir
+LibraryExecutables = bin/archdatadir/bin
+Plugins = bin/archdatadir/plugins
+Imports = bin/archdatadir/imports
+Qml2Imports = bin/archdatadir/qml
+Translations = bin/datadir/translations
+Documentation = bin/datadir/doc
+Examples = bin/datadir/examples""".format(self.dependencies["qt"].package_folder.replace('\\', '/')))
+
+    def _is_mingw(self):
```
nrbnlulu, line 38: All of these can be `@cached_property`

recipes/qt/6.x.x/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,145 @@
...
+        return False and self.options["qt"].shared and\
+            not cross_building(self) and\
+            not self.settings.os == "Macos" and\
+            not self._is_mingw()
+
+    def _qmake_supported(self):
+        return self.options["qt"].shared
+
+    def _build_with_qmake(self):
+        if not self._qmake_supported():
```
nrbnlulu, line 51: I'm not sure why you do this check here and not in `build()`?

recipes/qt/6.x.x/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,145 @@
...
+            cmake.configure()
+            cmake.build()
+
+    def build(self):
+        self._build_with_qmake()
+        self._build_with_meson()
+        self._build_with_cmake_find_package_multi()
+
+    def _test_with_qmake(self):
+        if not self._qmake_supported():
```
nrbnlulu, line 121: same here, can be done in `test()`...

nrbnlulu (changes requested):
Thanks for the PR hopefully will be merged soon

recipes/qt/6.x.x/test_package/conanfile.py
```diff
@@ -13,13 +13,32 @@
...
-    generators = "qt", "cmake", "cmake_find_package_multi", "cmake_find_package", "pkg_config", "qmake"
+    generators = "cmake", "cmake_find_package_multi", "cmake_find_package", "pkg_config", "qmake"
 
     def build_requirements(self):
         self.tool_requires("cmake/3.25.3")
         if self._meson_supported():
             self.tool_requires("meson/1.0.1")
 
+    def generate(self):
+        path = self.deps_cpp_info["qt"].rootpath.replace("\\", "/")
```
uilianries, line 24:
```suggestion
        path = self.dependencies["qt"].package_folder.replace("\\", "/")
```
deps_cpp_info is Conan 1.x

recipes/qt/6.x.x/test_package/conanfile.py
```diff
@@ -13,13 +13,32 @@
...
+        folder = os.path.join(path, "bin")
+        bin_folder = "bin" if self.settings.os == "Windows" else "libexec"
+        save(self, "qt.conf", f"""[Paths]
+Prefix = {path}
```
uilianries, line 28: why do you need to add this configuration file only now?

uilianries (commented):
Thank you for your contribution, but I see as a big losing by not having an embedded generator as before. People needing to manage .conf file with all attributes now, will break current project using Qt package for sure.

MartinDelille:
> Thank you for your contribution, but I see as a big losing by not having an embedded generator as before. People needing to manage .conf file with all attributes now, will break current project using Qt package for sure.

@uilianries From what I understand it was not possible to add a generator in a conanfile.py with conan v2. I based my work on the qt5 migration here: https://github.com/conan-io/conan-center-index/pull/16404

uilianries:
@MartinDelille Is it qmake generator or another custom generator? Asking because qmake is not included in Conan 2.x yet, but there is a feature asking for, maybe, that's the best opportunity to implement it.

jwillikers:
@uilianries Qt 6 migrated to CMake, fyi.

uilianries (approved):
LGTM

uilianries:
> @uilianries Qt 6 migrated to CMake, fyi.

Thank you, now it's clear.

rob-avalon:
```
ERROR: qt/6.5.0: Error in source() method, line 598
	ConanException: 'self.settings' access in 'source()' method is forbidden
```

Found a [discussion](https://cpplang.slack.com/archives/C41CWV9HA/p1683721172937099) on why self.setting is forbidden in sources. 

> because once the sources are downloaded, they are kept in local cache, so they are re-used as-is when building other 
> configurations on the same computer (debug, release, compiler etc.)

> the best way is to download sources in the build method in this case. Like this your are safe, you always download sources
> matching your current settings

Sounds like the work around may be to move [these few lines](https://github.com/conan-io/conan-center-index/blob/b77038bdaee9ac6a67a6585032767d7aeb1b74ca/recipes/qt/6.x.x/conanfile.py#LL598C9-L598C9) into `build()`? Does that make sense to anyone?

Looking in conan center index [it looks like most recipes](https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+apply_conandata_patches&type=code&p=1) call `apply_conandata_patches` in the `build()` method, which also seems to line up with that suggestion.

I don't have any experience writing packages, but maybe that could help someone figure out why 6.5.0 failed in v2

sabelka:
I think "settings_build" should be used here instead. When cross-compiling eg. on a Linux build system for a Windows target one doesn't need this hack for long paths.

ericLemanissier:
I added the `settings_build` change in https://github.com/conan-io/conan-center-index/pull/17942

PR #17542, opened 2023-05-12
stdgpu: add recipe

### LINKED ISSUE #17324: [request] stdgpu
### Webpage

https://stotko.github.io/stdgpu/

### Source code

https://github.com/stotko/stdgpu

stdgpu is an open-source library providing several generic GPU data structures for fast and reliable data management. Multiple platforms such as CUDA, OpenMP, and HIP are supported allowing you to rapidly write highly complex agnostic and native algorithms that look like sequential CPU code but are executed in parallel on the GPU.

The CUDA and HIP portions of the code are almost, but not entirely, header-only. I suppose ConanCenter could limit the builds to OpenMP-only due to lack of CI support and packaging conventions for the CUDA/HIP portions.

valgur:
Specify library name and version:  **stdgpu/1.3.0**

Resolves #17324.

Depends on #17411.

---


jwillikers (approved):

RubenRBS (dismissed):

recipes/stdgpu/all/cmake/stdgpu-dependencies-cuda.cmake
```diff
@@ -0,0 +1,6 @@
+# Mimics the behavior of stdgpu-dependencies.cmake exported by stdgpu for the CUDA backend
```
uilianries, line 1: The project looks active, is it possible to send it to the upstream?

valgur: Possibly. I think it already does something similar already, though. I'll have to look into it.
Either way, these CMake modules are unavoidable in this recipe due to CUDA and HIP not being available as packages, so these are needed to correctly configure them as dependencies.

recipes/stdgpu/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_min_versions.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
```
uilianries, line 118:
```suggestion
        if self.options.backend == "cuda" and self.dependencies["thrust"].options.device_system != "cuda":
            raise ConanInvalidConfiguration(f"{self.ref} option backend=cuda should use '-o thrust/*:device_system=cuda' as well.")
        if self.options.backend == "openmp" and self.dependencies["thrust"].options.device_system != "omp":
            raise ConanInvalidConfiguration(f"{self.ref} option backend=openmp should use '-o thrust/*:device_system=omp' as well.")

```

Forcing an option does not impose user's options, so you still need to validate it to avoid override.

valgur: Good to know, thanks!

recipes/stdgpu/all/conanfile.py
```diff
@@ -0,0 +1,211 @@
...
+            }
+
+    def export_sources(self):
+        copy(self, "cmake/*", dst=self.export_sources_folder, src=self.recipe_folder)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+        if self.options.backend != "openmp":
+            del self.options.openmp
```
uilianries, line 80:
```suggestion
```

It will not work because options values are not evaluated yet in config_options, they will use default value always. You need to move it to configure instead. Only self.settings is evaluated here. More information here: https://docs.conan.io/2/reference/conanfile/methods/config_options.html

recipes/stdgpu/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+
+find_package(stdgpu CONFIG REQUIRED)
+
+if (stdgpu_VERSION_STRING EQUAL "1.3.0")
+    set(CMAKE_CXX_STANDARD 14)
+else()
+    set(CMAKE_CXX_STANDARD 17)
+endif()
+
+if (STDGPU_BACKEND STREQUAL "cuda")
```
uilianries, line 12: nice, testing those options.

uilianries (approved):
LGTM

jwillikers (approved):

PR #15190, opened 2023-01-09
Add LEAF to Conan Center

kammce:
Specify library name and version: leaf/1.81.0

I work with and have consent by the author to put LEAF on the Conan Center Index. This library is a dependency of my library libhal.

Original github library can be found here: https://github.com/zajo/leaf

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17568, opened 2023-05-14
leaf-1.82.0

zajo:
boost-leaf/1.82.0

I'm the author of LEAF. This PR makes the latest GitHub release available in Conan.

Note to maintainers: this is a lightweight standalone library that does not depend on Boost or anything else, even though it is included in Boost releases. This Conan recipe exists so that LEAF can be used without dragging the massive Boost release as a dependency.

---


RubenRBS:
Hi @zajo thanks a lot for your contribution. As discussed in https://github.com/conan-io/conan-center-index/pull/15190, supporting standalone libraries also present in Boost is a no-go for now. I have to double check if adding new versions of `deprecated` libraries is still ok.

Note that there's active work to support modularized Boost currently underway, which might solve all the issues with not wanting to drag the whole of Boost for libraries that don't depend on it.

zajo:
@RubenRBS FYI there are bug fixes in this update.

As for Boost "modularization", I think that's a not going to work in general. If you look at the [Boost Dependency Report](https://pdimov.github.io/boostdep-report/) you'll see that most Boost libraries are dependent on other Boost libraries and there is no practical way to reliably untangle a library in general. However, [Level 0 Boost libraries](https://pdimov.github.io/boostdep-report/boost-1.82.0/module-levels.html) can be distributed separately because they don't depend on Boost (LEAF is a Level 0 library).

RubenRBS (approved):
That makes sense

kammce:
Bump

PR #17573, opened 2023-05-15
metis: add support for conan v2

joakimono:
Specify library name and version:  **metis/5.1.1**

- Add support for conan v2
- Add new option `with_64bit_types` to allow choosing int and float types to be 32 or 64 bit. 32-bit types are required for some downstream packages
- Depends on gklib, with PR #17570 

---


recipes/metis/all/conanfile.py
```diff
@@ -14,88 +16,88 @@ class METISConan(ConanFile):
...
         cmake.install()
-        self.copy("LICENSE", src=self._source_subfolder, dst="licenses")
 
     def package_info(self):
         self.cpp_info.libs = ["metis"]
         self.cpp_info.requires.append("gklib::gklib")
+
+        self.cpp_info.set_property("cmake_file_name", "metis")
+        self.cpp_info.set_property("cmake_target_name", "metis::metis")
+
```
SpaceIm, line 91:
```suggestion
```

recipes/metis/all/conanfile.py
```diff
@@ -14,88 +16,88 @@ class METISConan(ConanFile):
...
             self.cpp_info.defines.append("USE_GKREGEX")
-        if self._is_msvc:
+        if is_msvc(self):
             self.cpp_info.defines.append("__thread=__declspec(thread)")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.filenames["cmake_find_package"] = "METIS"
+        self.cpp_info.filenames["cmake_find_package_multi"] = "metis"
+        self.cpp_info.names["cmake_find_package"] = "METIS"
+        self.cpp_info.names["cmake_find_package_multi"] = "metis"
```
SpaceIm, line 103:
```suggestion
```

SpaceIm (changes requested):

SpaceIm:
"Add support for conan v2" is not very informative given that there are 1500 recipes in conancenter, could you change this title please?

recipes/metis/all/conanfile.py
```diff
@@ -14,88 +16,78 @@ class METISConan(ConanFile):
...
-            self._cmake.definitions["ASSERT2"] = self.settings.build_type == "Debug"
-            self._cmake.definitions["METIS_IDX64"] = True
-            self._cmake.definitions["METIS_REAL64"] = True
-            self._cmake.configure(build_folder=self._build_subfolder)
-        return self._cmake
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
```
RubenRBS, line 60: No need to set this, already done by the CMakeToolchain integration :)

RubenRBS (changes requested):
Thanks a lot! Requesting changes in this PR so it does not get merged by accident while we wait for the other PR to get merged first, but otherwise the changes look good!

recipes/metis/all/conanfile.py
SpaceIm: It's required when cmake_minimum_required is less than 3.0, if you want to set install_name of shared lib to `@rpath` on Apple (relocatable shared lib). CMakeToolchain does not manage this policy.

SpaceIm: It would be enough, didn't see it. What is the rational behind this bump of cmake min version? It's not explained.

recipes/metis/all/conandata.yml
```diff
@@ -5,12 +5,28 @@ sources:
...
+      patch_description: "Remove building of programs"
     - patch_file: "patches/004-use-conan-gklib.patch"
-      base_path: "source_subfolder"
+      patch_type: "conan"
+      patch_description: "Use conan gklib"
     - patch_file: "patches/005-fix-install.patch"
-      base_path: "source_subfolder"
+      patch_type: "conan"
+      patch_description: "Fix runtime install"
+    - patch_file: "patches/006-require-newer-cmakelists.patch"
```
franramirez688, line 22: @joakimono why is this needed? I saw upstream is requiring 2.8, is it related to using the `CMAKE_POLICY_DEFAULT_CMP0042` policy?

franramirez688 (changes requested):
Thanks for this contribution! Only one clarification before moving forward with the PR.

recipes/metis/all/conandata.yml
joakimono: I changed it due to remove another policy warning CMP0054 regarding quoted variables. I tried to suppress this like is done  with CMP0042 above, but for some reason it had no effect. The warning is triggered by the generated `cmakedeps_macros.cmake:33`. I assumed that changing the minimum required cmake in the upstream did not matter much, since the recipe requires 3.15 anyway. It should be noted that the metis CMakeLists.txt files seem pretty standard to me and do not use any deprecated 2.8 functionality.

franramirez688:
It looked like it got stuck at some point. Updating the branch to see if the CI is passing for Conan v2 and v1.

joakimono:
> It looked like it got stuck at some point. Updating the branch to see if the CI is passing for Conan v2 and v1.

The recently merged dependency gklib seems not to be built with the same build matrix configuration as this PR. Even though gklib was all green, it seems e.g. only apple-clang 13 was built (conan v1). #17570.

SSE4:
@RubenRBS @franramirez688 can you unblock review?

SSE4:
@franramirez688 @RubenRBS can you unblock? it seems like your questions were answered long time ago.

RubenRBS (approved):

franramirez688 (approved):

PR #17589, opened 2023-05-16
7bitdi : add new recipe

7bitcoder:
Specify library name and version:  **7bitdi/1.0.0**

This is my c++17 dependency injection library
https://github.com/7bitcoder/7bitDI
I will use this library in my other projects

---


recipes/7bitdi/all/test_package/test_package.cpp
RubenRBS, line 1: This test package code is fine, but note that they are meant to test that the package was correctly built, so something as simple as including the header file and printing `_7BIT_DI_VERSION` would also be ok as a test package and consume "less" resources in the CI. (Not that it matters much for this case as it's pretty small code, but every bit counts when you have thousands of different test packages!) :)

RubenRBS (commented):
Thanks a lot for your contribution, the changes look good, but I need to double check the PR later to approve it, as I've ran out of time right now :)

recipes/7bitdi/all/test_package/test_package.cpp
7bitcoder: Hi, thanks for the review, no problem I simplified the code in test_package so now it uses less CI resources

recipes/7bitdi/all/conanfile.py
```diff
@@ -0,0 +1,138 @@
...
+    def generate(self):
+        if not self.options.header_only:
+            tc = CMakeToolchain(self)
+            tc.variables["_7BIT_DI_BUILD_EXAMPLES"] = False
+            tc.variables["_7BIT_DI_BUILD_TESTS"] = False
+            tc.variables["_7BIT_DI_BUILD_DOC"] = False
+            tc.variables["_7BIT_DI_BUILD_SINGLE_HEADER"] = False
+            tc.variables["_7BIT_DI_INSTALL"] = True
+            tc.variables["_7BIT_DI_LIBRARY_TYPE"] = self.getSevenBitDILibraryType()
+            tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
```
czoido, line 84: This is added by CMakeToolchain by default, so I think you can remove that line without problems.
```suggestion
```

7bitcoder: Thanks, I just removed that line.

recipes/7bitdi/all/conanfile.py
```diff
@@ -0,0 +1,137 @@
...
+            tc = CMakeToolchain(self)
+            tc.variables["_7BIT_DI_BUILD_EXAMPLES"] = False
+            tc.variables["_7BIT_DI_BUILD_TESTS"] = False
+            tc.variables["_7BIT_DI_BUILD_DOC"] = False
+            tc.variables["_7BIT_DI_BUILD_SINGLE_HEADER"] = False
+            tc.variables["_7BIT_DI_INSTALL"] = True
+            tc.variables["_7BIT_DI_LIBRARY_TYPE"] = self.getSevenBitDILibraryType()
+            tc.generate()
+        cmake_deps = CMakeDeps(self)
+        cmake_deps.generate()
```
jwillikers, line 86: Do you only require the dependencies for CMake when it is not header-only?

recipes/7bitdi/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.8)
```
jwillikers, line 1: I think we are defaulting to `3.15` as the minimum version now.

```suggestion
cmake_minimum_required(VERSION 3.15)
```

jwillikers (dismissed):
Only a couple of negligible suggestions. LGTM.

recipes/7bitdi/all/conanfile.py
7bitcoder: Hi, actually I realized that these lines of code can be removed

recipes/7bitdi/all/test_package/CMakeLists.txt
7bitcoder: Changed version to 3.15

jwillikers (approved):

danimtb (approved):

PR #17590, opened 2023-05-16
opus: Add option to disable stack protector.

simoncent:
Add an option to allow the stack protector to be disabled to make it easier to build with Emscripten. This is already exposed through CMake, so the Conan recipe change is minimal; just add an option & pass it through.

The `OPUS_STACK_PROTECTOR` option defaults to `ON` in the CMake build, so I have defaulted the Conan option to `True` to mirror that.

---

### DIFF
--- a/recipes/opus/all/conanfile.py
+++ b/recipes/opus/all/conanfile.py
@@ -20,11 +20,13 @@ class OpusConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "fixed_point": [True, False],
+        "stack_protector": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "fixed_point": False,
+        "stack_protector": True,
     }
 
     def export_sources(self):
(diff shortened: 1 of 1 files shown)


SpaceIm:
I would even disable `OPUS_FORTIFY_SOURCE`, `OPUS_FAST_MATH` and `OPUS_STACK_PROTECTOR` since they just seem to be babysitting of compiler options.

simoncent:
I agree with your sentiment, but I'm a bit reluctant to change the default behaviour unexpectedly on people. Do you think that's a non-issue? Exposing the option feels like the safer approach, although it is a bit irritating. 

I don't think `OPUS_FORTIFY_SOURCE` and `OPUS_FAST_MATH` are exposed as CMake options in opus 1.3.1, and the recipe hasn't been updated to support 1.4.0 yet AFAIK so those should probably go in that eventual PR; since they were only added in the 1.4.0 release, defaulting them to `False` feels less controversial because no one will be expecting the flags to be passed, although when using MinGW it looks like they do require an extra library linking, so maybe that complicates it slightly.

Anyway, for this PR I think we only need be concerned with `OPUS_STACK_PROTECTOR`, and I don't mind too much either way whether it's an option of just disabled and not exposed, the only issue is whether it's OK to surprise people with a change to the default behaviour.

uilianries (approved):
LGTM as it's already exposed by cmake and the default value is correct. Plus, making easier to build another package makes sense.

jwillikers (approved):

PR #17592, opened 2023-05-16
mailio: add version 0.22.0

toge:
Specify library name and version:  **mailio/0.22.0**

---


RubenRBS (commented):
Thanks!

As for the error, I think it might be due to missing the `transitive_headers` trait for openssl?

toge:
@RubenRBS 
Thanks your advise.
I fixed this problem.

RubenRBS:
The error seems to be a persistent one, will check tomorrow what's up with it

toge:
@RubenRBS 
Thanks a lot!

RubenRBS:
Fixed!

This happened because the first run uploaded a corrupted package. This is a problem for the current CI implementation, as before building any package, we check if it's already there and skip it if so (`UPSTREAM` status in the logs). The problem is that there's nothing in place to catch cases where a corrupted package is found. I've talked with @uilianries for ways to improve that.

The solution was to remove the corrupted package and let it rebuild, hopefully this time it does not get corrupted

recipes/mailio/all/conanfile.py
```diff
@@ -58,7 +58,7 @@ def layout(self):
 
     def requirements(self):
         self.requires("boost/1.81.0", transitive_headers=True)
-        self.requires("openssl/[>=1.1 <4]")
+        self.requires("openssl/[>=1.1 <4]", transitive_headers=True)
```
RubenRBS, line 61: Seems like `transitive_libs=True` for openssl will be needed :)
```suggestion
        self.requires("openssl/[>=1.1 <4]", transitive_headers=True, transitive_libs=True)
```

RubenRBS (changes requested):

recipes/mailio/all/conanfile.py
toge: Thank you for your help!

toge (commented):
make openssl transitive_libs

recipes/mailio/all/conanfile.py
```diff
@@ -58,7 +58,7 @@ def layout(self):
 
     def requirements(self):
         self.requires("boost/1.81.0", transitive_headers=True)
```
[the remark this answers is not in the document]
RubenRBS: Seems like transitive_libs is also needed here

RubenRBS (changes requested):

toge:
@RubenRBS 
Thanks a lot for your help with this PR.

cmake requirement in build_requirements has been modified as you wrote.
https://github.com/conan-io/conan-center-index/blob/d85c212e945b3272d7ec32b8c9ecf231e6cbd28f/recipes/mailio/all/conanfile.py#L73-L74

Is there anything left to do?

RubenRBS:
> Is there anything left to do?

Did you push the cmake changes? I don't see it in the diff. Other than that, after that fixing the cmake, we would only need to figure out what's happening with the v2 pipeline 🤔

SSE4 (approved):

SSE4:
@RubenRBS can you unblock? thanks

RubenRBS:
I want to see if we can get the v2 pipeline working before merging :)

SSE4:
> I want to see if we can get the v2 pipeline working before merging :)

@RubenRBS can you unblock? I think v2 pipeline is not mandatory to pass here. if you need to investigate, I think you can make a copy of that branch in staging CI. it's already blocked for 3 weeks...

RubenRBS (approved):

PR #17593, opened 2023-05-16
[google-cloud-cpp] update to v2.10.1

coryan:
Specify library name and version:  **google-cloud-cpp/2.10.1**

Updates `google-cloud-cpp` to the latest release (v2.10.1).  This is the last in a series of PRs to complete the update.

The patches are substantially smaller in v2.10.1, as compared to v2.5.0, because I changed the package upstream to require fewer patches.  The conanfile required some changes because I never tested supporting more than one version.  I think it will require fewer changes in the future.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #16438, opened 2023-03-07
[google-cloud-cpp]: support for v2.x.y series

coryan:
This change splits packaging for the v2.x vs. v1.x series.  The 2.x series requires C++14, and include dozens of new components. I have updated the package to use the `googleapis` package in Conan.

Specify library name and version:  **google-cloud-cpp/2.5.x**

The current recipe does not work. It tries to "unvendor" googleapis, but did not complete the work.  It also does not compile all the components.  I am one of the maintainers for `google-cloud-cpp` upstream.  FWIW, I think the next release will have fewer (or at least smaller) patches.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17598, opened 2023-05-17
[grpc-proto] relax dependency on googleapis

coryan:
Specify library name and version:  **grpc-proto/cci.20220627**

Another PR in the series to upgrade `google-cloud-cpp` to v2.10.1.  We need to relax the version matches for `googleapis` or things will not work (see the failures in #17593).  `grpc-proto` depends on very few protos from `googleapis`, and those protos are extremely stable.  I chose `>= googleapis/cci.20221108` because that is the first version that correctly preserves the dependencies between the different googleapis libraries.

---

### DIFF
--- a/recipes/grpc-proto/all/conanfile.py
+++ b/recipes/grpc-proto/all/conanfile.py
@@ -57,7 +57,7 @@ class GRPCProto(ConanFile):
        # protobuf symbols are exposed from generated structures
        # https://github.com/conan-io/conan-center-index/pull/16185#issuecomment-1501174215
         self.requires("protobuf/3.21.9", transitive_headers=True, transitive_libs=True, run=can_run(self))
-        self.requires("googleapis/cci.20230501")
+        self.requires("googleapis/[>=cci.20221108]")
 
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):


coryan:
@RubenRBS any thoughts on how to workaround the Conan v1 limitations? It rejects `>=cci.20221108` as a valid version range.

RubenRBS:
Yup, this is going to be somewhat of a problem, let me double check with the team how to properly handle this

RubenRBS:
It's in the planned team discussion for tomorrow, I'll keep you posted :)

RubenRBS:
It seems that ranges like this in v1 just can't be supported. The only way forward suggested by @memsharded is for this to branch the code based on the Conan version. This can be acomplished with `from conan import conan_version`. If `conan_version` is less than 2.0, pin the recipe to a manual version, and for v2, use the range.

Try implementing it and I'll talk with @jcar87 to know what his opinions are on this, at it might have some unintended consecuences :)

coryan:
> It seems that ranges like this in v1 just can't be supported. The only way forward suggested by @memsharded is for this to branch the code based on the Conan version. This can be acomplished with `from conan import conan_version`. If `conan_version` is less than 2.0, pin the recipe to a manual version,

I already tried pinning to a particular version.  That does not work.  Each version of `google-cloud-cpp` works with one, and only one, version of `googleapis`.  Older versions of `google-cloud-cpp` cannot compile if `grpc-proto` and `grpc` are pinned to newer versions of `googleapis`.  Likewise newer versions of `google-cloud-cpp` cannot compile if `grpc-proto` and `grpc` are pinned to older versions.

> and for v2, use the range.

Ack.

> 
> Try implementing it and I'll talk with @jcar87 to know what his opinions are on this, at it might have some unintended consecuences :)

I have a suggestion:

Drop the idea of unvendoring `googleapis`.  As I explained in [#16438](https://github.com/conan-io/conan-center-index/pull/16438#issuecomment-1520507603), `googleapis` is not used by `grpc` except in its tests.  And the tests are not compiled by Conan.  It makes no sense to create an extra package and a dependency for something that is not really used.

coryan:
Any further thoughts on this?

jcar87:
Hi @coryan - indeed based on this discussion I believe the best course of action would be to revert back to the behaviour of having google-cloud-cpp build the protos from googleapis it needs, rather than rely on them as a separate library.  What we know today:
* `google-cloud-cpp` only works with one and only one revision of `googleapis`
* there is no overlap or possibility of symbol collision between the protos from `googleapis` that are vendored inside `google-cloud-cpp` and `grpc`/`grpc-proto`

I don't think the extents of the details were known when `googleapis` was originally introduced as a Conan library. On the other hand, the original intent was for the `googleapis` recipe to _not_ contain compiled code - just the `.proto` definitions - I suspect this could have potentially allowed making this a private/invisible dependency such that `google-cloud-cpp` could depend on an exact version without causing conflicts, however support for that is limited in Conan 1.x and it caused issues with the Conan Center CI pipeline that are unrelated to this. 

So as far as I'm concerned and given the problems we've experienced, going back to vendoring in the protos inside `google-cloud-cpp` seems like the best course of action.

Thanks for providing insights and for championing the efforts to have up-to-date versions of `google-cloud-cpp`. I'm at your disposal to move things forward!

PR #17046, opened 2023-04-17
harfbuzz: bump dependencies

db4:
**harfbuzz/6.0.0**

PR #7734, opened 2021-10-17
- [gstreamer] use full_package_mode for GLib

SSE4:
long story, please read carefully:
https://github.com/conan-io/conan-center-index/pull/7142#discussion_r706835445
https://github.com/conan-io/conan/issues/9712

TLDR:
1. GLib must exist in address space only once, otherwise UB (may crash, hang, as we see in #7634)
2. it means the only valid combinations of options are:
2.1. `glib:shared=False`, `gstreamer:shared=False`, `gst-plugins-*:shared=False`
2.2. `glib:shared=True`, `gstreamer:shared=True`, `gst-plugins-*:shared=True`
2.3. everything else is unsupported and should be discarded in `validate`
3. that complicates an issue more `glib:shared=False` is not supported on Windows (read, if you're interested: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1655 https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1316)
4. that CCI currently builds are:
4.1. all static (default) - `*:shared=False`
4.2. `<current recipe>:shared=True` + all the dependencies static (default)
5. *CONSEQUENCE 1*: there will be no shared GStreamer builds on CCI, they must be built from sources (shared will be rejected because it's a mixture of static + shared)
6. *CONSEQUENCE 2*: there will be no Windows GStreamer builds on CCI, they must be built from sources (static will be rejected because static Windows GLib is not supported, shared will be rejected for the reason specified in the previous point)

being pragmatic, it's probably too much to reject these builds and require to always build from sources.
OTOH these builds are technically invalid anyway, and shouldn't be used by anyone, since they contain UB.
I am very tentative on these changes, so it's a draft for now, let's discuss and exchange opinions/arguments.

Specify library name and version:  **gstreamer/all**

This is also a good place to share with all of us **why you are submitting this PR** (specially if it is a new addition to ConanCenter): is it a dependency of other libraries you want to package? Are you the author of the library? Thanks!

---

- [ ] I've read the [guidelines](https://github.com/conan-io/conan-center-index/blob/master/docs/how_to_add_packages.md) for contributing.
- [ ] I've followed the [PEP8](https://www.python.org/dev/peps/pep-0008/) style guides for Python code in the recipes.
- [ ] I've used the [latest](https://github.com/conan-io/conan/releases/latest) Conan client version.
- [ ] I've tried at least one configuration locally with the
      [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17607, opened 2023-05-18
glib: regenerate binaries

ericLemanissier:
for conan-io/conan-center-index#17046

Specify library name and version:  **glib/***

---

### DIFF
--- a/recipes/glib/all/conanfile.py
+++ b/recipes/glib/all/conanfile.py
@@ -163,6 +163,7 @@ class GLibConan(ConanFile):
             src=os.path.join(self.package_folder, "share"),
             dst=os.path.join(self.package_folder, "res"),
         )
+        rmdir(self, os.path.join(self.package_folder, "share"))
         rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
         fix_apple_shared_install_name(self)
         fix_msvc_libname(self)
--- a/recipes/glib/all/conanfile.py
+++ b/recipes/glib/all/conanfile.py
@@ -8,7 +8,6 @@ from conan.tools.meson import Meson, MesonToolchain
 from conan.tools.microsoft import is_msvc
 from conan.tools.scm import Version
 import os
-import shutil
 
 
 required_conan_version = ">=1.53.0"
@@ -160,9 +159,9 @@ class GLibConan(ConanFile):
         meson.install()
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
         rmdir(self, os.path.join(self.package_folder, "libexec"))
-        shutil.move(
-            os.path.join(self.package_folder, "share"),
-            os.path.join(self.package_folder, "res"),
+        copy(self, pattern="*",
+            src=os.path.join(self.package_folder, "share"),
+            dst=os.path.join(self.package_folder, "res"),
         )
         rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
         fix_apple_shared_install_name(self)


RubenRBS:
Hi @ericLemanissier thanks for the pro-active PR to try and fix the issue.

Unfortunately even if this runs and gets merged, the other PR is going to still fail, as the glib error is not on missing packages that were meant to exists, but on a specific issue with its recipe messing up the package ids of the requirements. @jcar87 has been investigating them for a few days now and hopefully can give some more insight in a bit/when we manage to fix it

jcar87:
Hi @ericLemanissier  - thanks for this! Have been investigating this for a while and it is a tricky situation - although I believe we can fix this. In short: I'm afraid this PR will not end up generating the missing package IDs.

The Conan v2 pipeline will buy libraries with the following options:

- `*:shared=True`
- `*:shared=True`

This will apply to the entire dependency tree - in an aim to avoid legacy issues with the Conan 1.x approach, where `libname:shared=True` but everything else with the default of `shared=False` can result in a situation where we have multiple copies of the same static library across multiple shared libraries. If libraries have global state, this could be an issue (and glib is a primer example of this). Even if they don't have global state, this could result in UB even if things appear to run correctly.


The issue at hand has to do with `tool_requires`, intended to express dependencies on packages that contain executables used during build time. When the 1.x pipeline builds executables (that is, they don't have the `shared` option), we don't pass  any `*:shared` option at all - it will build with the defaults, which in most cases is `shared=False`.

When you try to build a recipe that _has_ a tool requires on something else, and you do it like this:
```
conan create . --version=1.0 -o "*:shared=True"
```

the shared option only applies to the _host_ dependencies, not the _build_ one - and this is the same when we build things in Conan center - So for `tool_requires` it will always pick them up with their default options (in the overwhelming majority of cases, this is `shared=False` as per Conan Cente rpolicies).

if the user _wanted_ the `shared=True` option for tool requirements, they'd have to do so explicitly:
```
conan create . --version=1.0 -o:h "*:shared=True"  -o:b "*:shared=True"
```

we opted to _not_ do this in Conan Center - for a number of reasons:
1) on the one hand, we assume the user is overwhelmingly more likely to only specify `-o "*:shared=True"` 
2) if Conan Center has `shared=False` by default, and that's what used for `tool_requires`, it simplifies the runtime story (to an extent), since executables will have statically linked dependencies 
3) Conan Center CI only applies "*:shared=True` and "*:shared=False` when building _libraries_, but not executables - so for a `package_type="application"`, if it has dependencies, they will come in the default option. This means that if Conan Center applied `*:shared=True` for the _build_ profile, there would be missing binaries as we just don't build that configuration at all.


now the problem with `glib`, is that it is also a `tool_requires` because it contains executables used at build time for downstream consumers. However, unlike most other libraries in Conan Center, its default is `shared=True` - what's happening is that for downstream consumers that have:
`self.tool_requires("glib/xxx")` - in the build profile conan will not apply either `*:shared=True` or `*:shared=False`, so it will go with the defaults for all the libraries involved. That is, `glib:shared=True`, and everything else is `shared=False` - this package ID is not something we build (or even want to build), so it will be missing - and that's why this PR will not actually result in building the missing package ID. 

When working locally for this particular case, `-o:b "*:shared=False"` should help when using a configuration for which we provide binaries, or `--build=missing` if not.


I think the solution in this case is rather simple, and we should revert the change here: https://github.com/conan-io/conan-center-index/pull/12171 and glib should be static by default. Changes like this one: https://github.com/conan-io/conan-center-index/pull/7734 already protect from the invalid configuration, and making it `shared=True` does _not_ avoid the missing binaries that could arise when glib is updated in some of the dependencies - so I'm not sure having `shared=True` actually resolved anything in that particular instance.

ericLemanissier:
OK. I thought the situation of shared libraries for build profile had been solved because the harfbuzz PR build somehow actually started on conan v2. Let's continue discussion on the other glib PR then.

Issue #11684, opened 2022-07-13
[question] Using `full_package_mode` for glib with gtk dependencies causes brittle builds


hesham-essam:
My PR for building gtk on windows was successfully built #11308, but after #11545 it can't be built anymore, since harfbuzz updated its glib version. And since gtk and its dependencies use `full_package_mode` for glib (which is necessary to prevent multiple instances of glib), I get `MISSING_DEPENDENCIES` error for `harfbuzz` (as harfbuzz with the old glib is not in the cache anymore).

This hard dependency makes it hard to update glib, since now if glib's version is bumped in any of gtk's dependencies, gtk will fail to build in the CI until all other dependencies bump glib's version as well.

How do you think this should be handled?

planetmarshall:
I'm not sure what the best solution for this is - I'm inclined to suggest static glib should be an unsupported configuration (I can't debug a GTK app in Linux built with static glib - even with everything built from scratch), but I'm not sure that would solve the problem.

This is really an issue with the cached binaries, as there's nothing really wrong with the recipe - it can be built using `--build missing` by the end user with no problems.

morevnaproject:
Hi! Any news on this topic? It would be really nice to have a solution, as we need Gtk. ^__^

planetmarshall:
I think this is going to continue to be a problem for any recipes where `full_package_mode` is specified. Is it possible for CCI packages to be built with `--build missing`? Or would that just consume too many resources?

uilianries:
CCI will not use `--build missing`.  Will need to continue to update recipes.

hesham-essam:
Now that glib is shared, I think we should be removing`full_package_mode` (or at least use it only when linking against static glib). These patches are really causing problems.

uilianries:
@hesham-essam Yes, removing full_package_mode sounds correct. We will need new PRs with those fixes.

hesham-essam:
This is solved now

PR #17610, opened 2023-05-18
glib: revert shared option to be False by default

jcar87:
Specify library name and version:  **glib/all**

Revert the `shared` option to have a `False` value by default, in line with the Conan Center defaults.

The PR that introduced this change: https://github.com/conan-io/conan-center-index/pull/12171 attempted to do so in order to be able to _remove_ the `full_package_mode` on the `glib` dependency found in recipes where `glib` is a dependency. However, `full_package_mode` has not been removed, and most crucially, having `shared=True` by default does _not_ solve the problem mentioned in this issue: https://github.com/conan-io/conan-center-index/issues/11684 - there will still be missing binaries if the exact version of glib changes across the dependency graph. 

Having a default on `shared=True` causes "missing binary" problems in the Conan 2.0 CI pipeline, when `glib` is `tool_requires` - it will require a configuration where `glib` is shared, but all its dependencies are static - we no longer build this sort of binaries in Conan Center (static libraries embedded in shared ones), as it can cause symbol duplication across multiple libraries loaded in the same process.

The problems described in https://github.com/conan-io/conan-center-index/pull/7734 are already solved in Conan 2.0 - so in due time (sooner rather than later), we can due away with the `full_package_mode` as Conan 2.0 can handle this without issue.

### Changes in test package
On macOS, when `*:shared=True`, the `VirtualRunEnv` will generate an entry in `DYLD_LIBRARY_PATH` that contains `libiconv.2.dylib` from Conan. 
When running the `gdbus-codegen` (a Python script) tool in the test_package, the Python interpreter itself was built against a different libiconv, and fails with the following error:
```
   glib/2.75.2 (test package): RUN: /Users/jenkins/w/prod-v2/BuildSingleReference@3/p/glibbe449c3b10b54/p/bin/gdbus-codegen -h
   dyld[22955]: Symbol not found: (_iconv)
     Referenced from: '/usr/local/Cellar/gettext/0.21/lib/libintl.8.dylib'
     Expected in: '/Users/jenkins/w/prod-v2/BuildSingleReference@3/p/libic50650dc30c43d/p/lib/libiconv.2.dylib'
   /bin/sh: line 1: 22955 Abort trap: 6           /Users/jenkins/w/prod-v2/BuildSingleReference@3/p/glibbe449c3b10b54/p/bin/gdbus-codegen -h
```

This PR fixes this by unsetting `DYLD_LIBRARY_PATH` and setting `DYLD_FALLBACK_LIBRARY_PATH` instead - this allows the runtime linker to load the libraries first by their install location, and only if not found would they be loaded from paths in the environment variable. I don't expect this to be something we need to do generally in more recipes, as it is specific to this tool invoking a system managed Python interpreter.

### DIFF
--- a/recipes/glib/all/conanfile.py
+++ b/recipes/glib/all/conanfile.py
@@ -35,7 +35,7 @@ class GLibConan(ConanFile):
         "with_mount": [True, False],
     }
     default_options = {
-        "shared": True,
+        "shared": False,
         "fPIC": True,
         "with_elf": True,
         "with_mount": True,
--- a/recipes/glib/all/test_package/conanfile.py
+++ b/recipes/glib/all/test_package/conanfile.py
@@ -1,7 +1,7 @@
 from conan import ConanFile
 from conan.tools.build import can_run
 from conan.tools.cmake import cmake_layout, CMake, CMakeDeps, CMakeToolchain
-from conan.tools.env import VirtualBuildEnv, VirtualRunEnv
+from conan.tools.env import Environment, VirtualBuildEnv, VirtualRunEnv
 from conan.tools.gnu import PkgConfig, PkgConfigDeps
 import os
 
@@ -25,6 +25,15 @@ class TestPackageConan(ConanFile):
         tc.generate()
         virtual_run_env = VirtualRunEnv(self)
         virtual_run_env.generate()
+
+        if self.settings.os == "Macos":
+            env = Environment()
+            # Avoid conflicts with system libiconv
+            # see: https://github.com/conan-io/conan-center-index/pull/17610#issuecomment-1552921286
+            env.define_path("DYLD_FALLBACK_LIBRARY_PATH", "$DYLD_LIBRARY_PATH")
+            env.define_path("DYLD_LIBRARY_PATH", "")
+            env.vars(self, scope="run").save_script("conanrun_macos_runtimepath")
+
         if self.settings.os == "Windows":
             deps = CMakeDeps(self)
             deps.generate()


jcar87:
2.0 pipeline is failing on MacOS with `*:shared=True` with this error:

```
   glib/2.75.2 (test package): RUN: /Users/jenkins/w/prod-v2/BuildSingleReference@3/p/glibbe449c3b10b54/p/bin/gdbus-codegen -h
   dyld[22955]: Symbol not found: (_iconv)
     Referenced from: '/usr/local/Cellar/gettext/0.21/lib/libintl.8.dylib'
     Expected in: '/Users/jenkins/w/prod-v2/BuildSingleReference@3/p/libic50650dc30c43d/p/lib/libiconv.2.dylib'
   /bin/sh: line 1: 22955 Abort trap: 6           /Users/jenkins/w/prod-v2/BuildSingleReference@3/p/glibbe449c3b10b54/p/bin/gdbus-codegen -h
   
   ERROR: glib/2.75.2 (test package): Error in test() method, line 50
   	self.run(f"{gdbus_codegen} -h", env="conanrun")
   	ConanException: Error 134 while executing
```

This is caused because `gdbus-codegen` is a python script, and the Python on the CI agents (from homebrew) has a dependency on `/usr/local/opt/gettext/lib/libintl.8.dylib`, which depends on Apple's macOS `libiconv` instead of Conan's libiconv, which is being loaded because it's in the runenv.

SpaceIm:
Could you regenerate v1 packages of all dependent please? Otherwise it breaks many recipes in CI. See https://github.com/conan-io/conan-center-index/pull/16613#issuecomment-1616113162

PR #17614, opened 2023-05-18
gtest: posibility to disable pthread from options 

mralexlupu:
Specify library name and version:  **gtest**

---


RubenRBS:
If there's an unexpected error again in this new build I'll check the logs to see what went wrong :)

recipes/gtest/all/conanfile.py
```diff
@@ -113,7 +115,7 @@ def generate(self):
 
         if self.settings.compiler.get_safe("runtime"):
             tc.variables["gtest_force_shared_crt"] = "MD" in msvc_runtime_flag(self)
-        if self.settings.os == "Windows" and self.settings.compiler == "gcc":
+        if self.settings.os == "Windows" and self.settings.compiler == "gcc" or self.options.disable_pthreads:
             tc.variables["gtest_disable_pthreads"] = True
```
uilianries, line 119:
```suggestion
        tc.variables["gtest_disable_pthreads"] = self.options.disable_pthreads
```

Keep it managed by the users, otherwise, the option will not obey the wished behavior.

It should be safe, as default value is False:https://github.com/google/googletest/blob/cb455a71fb23303e37ce8ee5b1cde6a2c18f66a5/googletest/CMakeLists.txt#L22

And, for mingw is not used anyway: https://github.com/google/googletest/blob/cb455a71fb23303e37ce8ee5b1cde6a2c18f66a5/googletest/cmake/internal_utils.cmake#L72

mralexlupu: totally agree, did keep that because of uncertainty of it's need

uilianries (approved):

PR #17615, opened 2023-05-18
wt: add transitive_headers = True for boost dependency

pskoko:
Specify library name and version:  **wt/all**

Boost headers are exposed in public headers of Wt (for example Wt/Http/Client.h -> Wt/AsioWrapper/system_error.hpp -> boost/system/system_error.hpp) so it is necessary to add transitive_header = True for boost dependency. 

---


RubenRBS:
Hi @pskoko thanks for your contribution. I have submitted https://github.com/pskoko/conan-center-index/pull/1 to fix the missing binary problem for libmysqlclient :)

recipes/wt/all/conanfile.py
```diff
@@ -92,15 +92,15 @@ def _required_boost_components(self):
...
-            self.requires("boost/1.76.0")
+            self.requires("boost/1.76.0", transitive_headers = True)
         elif Version(self.version) < "4.9.0":
-            self.requires("boost/1.80.0")
+            self.requires("boost/1.80.0", transitive_headers = True)
         else:
-            self.requires("boost/1.81.0")
+            self.requires("boost/1.81.0", transitive_headers = True)
         if self.options.connector_http:
             self.requires("zlib/1.2.13")
```
RubenRBS, line 101: This might need `transitive_libs` or something?

pskoko: @RubenRBS I am not sure for zlib, but errors in failed apple build point out that maybe transitive_libs=True should be added for openssl.  On Apple build openssl version range is resolved to 3.1.0 which, unlike 1.1.1t, has zlib as its dependency (unless no_zlib=True is set up), and error messages show that all missing symbols are from libcrypto.a which is from openssl. 

Also, by experimenting with using wt as shared library on Linux, I concluded that libunwind also needs transitive libs = True. Without that, linking an executable which uses wt as shared library fails, it prints that some symbols from xz (which is dependency of libunwind) are undefined.

pskoko: On Windows I managed to get the errors equivalent to the apple build from CI (by setting openssl dependency explictitly to 3.1.0 version, it would always resolve to 1.1.1t): missing zlib from libcrypto.lib (for example: libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol deflate referenced in function zlib_stateful_compress_block).  Adding transitive_libs = True on openssl didn't resolve it.

jcar87:
Closing this via https://github.com/conan-io/conan-center-index/pull/18401 - commits were cherry-picked from this PR branch to preserve correct authorship.

PR #17618, opened 2023-05-18
libspng: add version 0.7.4

toge:
Specify library name and version:  **libspng/0.7.4**

- add version 0.7.4
    - since 0.7.4, libspng provides cmake config files.

---


recipes/libspng/all/patches/0.7.4-0001-allow-miniz.patch
```diff
@@ -0,0 +1,39 @@
...
+ endif()
+ 
+-find_package(ZLIB REQUIRED)
++if(SPNG_USE_MINIZ)
++    find_package(miniz REQUIRED)
++    set(compress_LIBS miniz::miniz)
++else()
++    find_package(ZLIB REQUIRED)
++    set(compress_LIBS ZLIB::ZLIB)
++endif()
```
memsharded, line 24: I might be missing some historical reasons in this recipe or library, but why if the library uses just zlib, why injecting this option to use miniz? I'd say drop the ``with_miniz`` option and go with the library ``zlib`` implemented one.

toge: @RubenRBS @jcar87 
Sorry for the late response.
The reason is as @jcar87 wrote about miniz support. (Thanks @jcar87)
I will try to create a PATCH with libspng.

If that is not acceptable, close this PR and create another PR to build with meson.
This is because I want to provide the recipe with as many official options as possible.

jcar87: Hi @toge ! THanks for your reply - I believe the PR is fine since it overcomes a buildsystem limitation, we just didn't have enough context. Perhaps you could elaborate a little more in the `patch_description` to say that miniz is supported, but is currently missing support in the cmake scripts - and link either my comment above: https://github.com/conan-io/conan-center-index/pull/17618#discussion_r1242166287, or even better, an issue raised upstream (doesn't need to be a full fledged patch, but we greatly encourage raising issues first in the upstream libraries rather than fixing them here directly).

This traceability can help future maintainers remove the patches once they are no longer needed, and also encourages improvements all around :D Thanks for your contribution!

recipes/libspng/all/conandata.yml
```diff
@@ -1,11 +1,18 @@
...
   "0.7.3":
     url: "https://github.com/randy408/libspng/archive/refs/tags/v0.7.3.tar.gz"
     sha256: "a50cadbe808ffda1a7fab17d145f52a23b163f34b3eb3696c7ecb5a52340fc1d"
   "0.7.2":
     url: "https://github.com/randy408/libspng/archive/refs/tags/v0.7.2.tar.gz"
     sha256: "4acf25571d31f540d0b7ee004f5461d68158e0a13182505376805da99f4ccc4e"
 patches:
+  "0.7.4":
+    - patch_file: "patches/0.7.4-0001-allow-miniz.patch"
+      patch_description: "add miniz option which is written in docs/BUILD.md"
```
jcar87, line 14:
```suggestion
      # Miniz is supported by the project, but does not currently expose a CMake
      # option to toggle it. This can be removed once the project supports it
      # or when the recipe moves to use Meson instead
      patch_description: "add miniz option which is written in docs/BUILD.md"
```

jcar87 (approved):

czoido (approved):

Issue #17390, opened 2023-05-04
[question] ninja package still broken?


tbsuht:
### What is your question?

Hi,

this regeneration of the ninja package did not fix the problem, at least I experience that: https://github.com/conan-io/conan-center-index/pull/17135

I need to rebuild the ninja package locally to have a usable package. (for conan 2.x)

Can someone confirm this?

ericLemanissier:
On my end the problem is fixed. Can you try with a clean conan cache (`rm -r ~/.conan2/p`) or add `-u` argument to your conan command so that it updates your local cache.

tbsuht:
@RubenRBS Pinging as suggested in the last PR: https://github.com/conan-io/conan-center-index/pull/17135#issuecomment-1521334379

RubenRBS:
Taking a look now, thanks for the heads up :)

RubenRBS:
Sorry for the radio silence, work is ongoing on addressing this issue, but we have had to prioritize other things first

jcar87:
Hi @tbsuht  - thanks for reporting this, and @ericLemanissier  for providing additional details.

If the error you were getting when running `ninja` had to do with `GLIBCXX` symbols, I believe this should now be fixed. If you could try again and let us know, that would be great! 


For context - `ninja` is built from C++ code, and on Linux for Conan 2.0 we are building with gcc 11 - this imposes a requirement on the running system, where a `libstdc++` is required that contains symbols with for `GLIBCXX_3.4.29` (matches gcc11 which is what the ninja binaries are currently generated with for Conan 2.0) - as per GCC's ABI guidelines: https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html. I suspect on some systems it is possible to install a newer runtime without issues, but this is not a burden we would like to impose on our users.

This should now be resolved since https://github.com/conan-io/conan-center-index/pull/17620 by statically linking the C++ runtime

PR #17620, opened 2023-05-18
ninja: use static runtime on Linux

jcar87:
Specify library name and version:  **ninja/all**

This PR ensures that ninja is built with a static C++ runtime on Linux, taking advantage of the fact that Ninja has no dependency on other libraries. 

Additional background:
This recipe erases the `compiler` setting, however it is still relevant at runtime since it requires that the system this executable is run on, has a version of libstdc++ that is _at least_ the same or greater than the one it was built with, which can cause problems on other systems.

More information:
**Before this change**:
```
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-aarch64.so.1]
```
Executable size: 305 KB

**After this PR**:
```
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-aarch64.so.1]
```

Executable size: 1.7 MB

I think the increase in executable size is negligible while still broadening the Linux distros this can run on (the other axis being `glibc`, but our Docker images ensure that is an older version for broader compatibility).

Close: https://github.com/conan-io/conan-center-index/issues/17390

### DIFF
--- a/recipes/ninja/all/conanfile.py
+++ b/recipes/ninja/all/conanfile.py
@@ -30,6 +30,8 @@ class NinjaConan(ConanFile):
         tc = CMakeToolchain(self)
         tc.variables["BUILD_TESTING"] = False
         if self.settings.os == "Linux" and "libstdc++" in self.settings.compiler.libcxx:
+            # Link C++ library statically on Linux so that it can run on systems
+            # with an older C++ runtime
             tc.cache_variables["CMAKE_EXE_LINKER_FLAGS"] = "-static-libstdc++ -static-libgcc"
         tc.generate()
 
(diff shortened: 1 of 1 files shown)


uilianries (approved):
LGTM. It should solve problems related to GLIBCXX version required when loading the application. As we don't consider compiler in settings, the same application can be used as tool requirements for any compiler. Plus, the size is not big as I thought, we have much bigger applications in CCI that don't have system libraries statically linked.

PR #17629, opened 2023-05-19
7zip: Add version 22.01 and Conan 2 compatibility

gjasny:
Specify library name and version:  **7zip/22.01**

---

### DIFF
--- a/recipes/7zip/19.00/conanfile.py
+++ b/recipes/7zip/19.00/conanfile.py
@@ -1,6 +1,5 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
-from conan.tools.env import Environment, VirtualBuildEnv, VirtualRunEnv
 from conan.tools.files import copy, chdir, download, get, replace_in_file
 from conan.tools.gnu import Autotools, AutotoolsDeps, AutotoolsToolchain
 from conan.tools.layout import basic_layout


recipes/7zip/19.00/conanfile.py
```diff
@@ -14,7 +18,8 @@ class SevenZipConan(ConanFile):
     license = ("LGPL-2.1", "BSD-3-Clause", "Unrar")
```
jwillikers, line 18:
```suggestion
    license = ("LGPL-2.1-or-later", "BSD-3-Clause", "Unrar")
```

recipes/7zip/19.00/conanfile.py
```diff
@@ -14,7 +18,8 @@ class SevenZipConan(ConanFile):
...
     homepage = "https://www.7-zip.org"
     topics = ("7zip", "zip", "compression", "decompression")
```
jwillikers, line 20:
```suggestion
    topics = ("archive", "compression", "decompression", "zip")
```

recipes/7zip/19.00/conanfile.py
```diff
@@ -55,47 +77,40 @@ def _msvc_platform(self):
...
+        if is_msvc(self):
             self._build_msvc()
         else:
             self._build_autotools()
 
     def package(self):
-        self.copy("DOC/License.txt", src="", dst="licenses")
-        self.copy("DOC/unRarLicense.txt", src="", dst="licenses")
+        copy(self, "License.txt", dst=os.path.join(self.package_folder, "licenses"), src=os.path.join(self.source_folder, "DOC"))
+        copy(self, "unRarLicense.txt", dst=os.path.join(self.package_folder, "licenses"), src=os.path.join(self.source_folder, "DOC"))
```
jwillikers, line 104: Does this license only apply to the DLL? If we don't package the DLL, then maybe we don't need it?

jwillikers (changes requested):
Just a question about the license really, otherwise it looks good.

recipes/7zip/19.00/conanfile.py
gjasny: No, it is explicitly mentioned in the `License.txt` and also relates to the source code:
```
  The licenses for files are:

    1) CPP/7zip/Compress/Rar* files: the "GNU LGPL" with "unRAR license restriction"
    2) CPP/7zip/Compress/LzfseDecoder.cpp: the "BSD 3-clause License"
    3) Some files are "public domain" files, if "public domain" status is stated in source file.
    4) the "GNU LGPL" for all other files. If there is no license information in 
       some source file, that file is under the "GNU LGPL".

  The "GNU LGPL" with "unRAR license restriction" means that you must follow both 
  "GNU LGPL" rules and "unRAR license restriction" rules.
```

And the generated binary contains rar support:
```
7z i

...
 0  ...F..................  Rar      rar r00       R a r ! 1A 07 00
 0  ...F..................  Rar5     rar r00       R a r ! 1A 07 01 00
...
```

jwillikers: Thanks, I wasn't sure since their license web page says things differently: https://www.7-zip.org/license.txt

jwillikers (dismissed):

recipes/7zip/config.yml
```diff
@@ -1,3 +1,3 @@
 versions:
-  "19.00":
+  "22.01":
```
danimtb, line 2: Why are you removing the 19.00 version? Normally we keep the versions that are already included in the repo unless we have many of them (in that case we keep only the latest patch).

gjasny: The `19.00` version dependency `"lzma_sdk/9.20"` requires a Conan 2 port. I can port it as well if you think that the changes I did to 7zip look sane.

gjasny: Let's wait for #17667

gjasny: I re-added v19

jwillikers (approved):

danimtb (approved):

PR #17637, opened 2023-05-20
expat: add `large_size` option`

### LINKED ISSUE #17541: [package] expat/all: EXPAT_LARGE_SIZE not defined / no option
### Description

Some Expat features do not work when xml files are >2GB on some systems (Windows or any system with 32bit `long`). It would be useful if the EXPAT_LARGE_SIZE configuration option was exposed within the conan configuration. According to expat "It may not be supported by all compilers, and is turned off by default." (imho I think it should be on by default)

The test code within conan-centre is actually already designed around having the flag as configuring with EXPAT_LARGE_SIZE should set the XML_LARGE_SIZE define. This would need to be taken into account if adding this option.

* Package Name/Version: **expat/2.x**
* Compiler+version: **MSVC 2017**
* Conan version: **conan 1.56**
* Python version: **Python 3.8**

### Conan profile

[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.toolset=v142
compiler.version=16
compiler.cppstd=17
os=Windows
os_build=Windows
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

`conan install -r expat/2.5.0`

Open a very large xml file. Use `XML_GetCurrentByteIndex` on an element that is beyond the 2GB line. Return value will be invalid.

### Logs

N/A

toge:
Specify library name and version:  **expat/***

Try to fix #17541.

Because EXPAT_LARGE_SIZE has been introduced since 2.2.8, all versions in conan recipe support this feature.

https://github.com/libexpat/libexpat/blob/R_2_2_8/expat/Changes

---

### DIFF
--- a/recipes/expat/all/conanfile.py
+++ b/recipes/expat/all/conanfile.py
@@ -20,11 +20,13 @@ class ExpatConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "char_type": ["char", "wchar_t", "ushort"],
+        "large_size": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "char_type": "char",
+        "large_size": False,
     }
 
     def export_sources(self):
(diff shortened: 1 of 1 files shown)


duncanspumpkin (approved):
Solves my issues

memsharded (approved):

jwillikers (approved):

PR #17663, opened 2023-05-23
xlsxio: add new recipe

hiiizxf:
Specify library name and version:  xlsxio/*

lib xlsxio is usefull to reading values from and writing values to .xlsx files. 
I would like to see it in conan center.

---


recipes/xlsxio/all/conanfile.py
```diff
@@ -0,0 +1,144 @@
...
+        "with_minizip_ng": [True, False],
+        "with_wide": [True, False],
+        "build_documentation": [True, False],
+    }
+    default_options = {
+        "fPIC": True,
+        "shared": False,
+        "build_pc_files": True,
+        "build_tools": True,
+        "build_examples": True,
```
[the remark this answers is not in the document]
danimtb: Maybe even delete this option in the recipe. CMake options != recipe options. If the options do not make sense for the user input view, then it is fine to nop add it here

recipes/xlsxio/all/conanfile.py
```diff
@@ -0,0 +1,144 @@
...
+        "with_libzip": [True, False],
+        "with_minizip_ng": [True, False],
+        "with_wide": [True, False],
+        "build_documentation": [True, False],
+    }
+    default_options = {
+        "fPIC": True,
+        "shared": False,
+        "build_pc_files": True,
+        "build_tools": True,
```
danimtb, line 36: Do you know why these options are enabled? Does it make sense when using Conan  (as conan generates pc files if you chose the generator for so)?

Are build tools required here as well? What do they do?

recipes/xlsxio/all/conanfile.py
```diff
@@ -0,0 +1,144 @@
...
+        "build_examples": True,
+        "with_libzip": False,
+        "with_minizip_ng": False,
+        "with_wide": False,
+        "build_documentation": False,
```
danimtb, line 41: same here with the documentation option. I think it is not needed here

danimtb (commented):
Great job in the recipe! I see some issues with the options though. Please take a look at the comments. I think we can simplify them 😄

recipes/xlsxio/all/conanfile.py
hiiizxf: 1.The pc file is generated from template.pc.in when build with cmake, when using Conan it do not make sense, i will remove it.
2.Here "build_tools" is not tools used to build project,  but is runtime target which can be deployed to a device.

recipes/xlsxio/all/conanfile.py
hiiizxf: OK

danimtb (approved):

PR #17427, opened 2023-05-06
[tk] Migrate tk to conan 2.0

samuel-emrys:
* Move to modern conan implementation
* Transition most file modifications to patches

Specify library name and version:  **tk/8.6.10**

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17664, opened 2023-05-23
[util-linux-libuuid] Add util-linux-libuuid package

### LINKED ISSUE #17337: [request] util-linux-libuuid/2.38.1
### Package Name/Version

util-linux-libuuid/2.38.1

### Webpage

https://github.com/util-linux/util-linux

### Source code

https://github.com/util-linux/util-linux

The `libuuid` library from `util-linux` is a BSD-licensed UUID implementation that is commonly used. It is part of the `util-linux` project like the existing `libmount` Conan package.

samuel-emrys:
* This is the most actively maintained fork of libuuid
* This package should supersede the existing libuuid recipe, which should be deprecated as an inactive fork of the original libuuid. The existing libuuid recipe does not match what gets installed on linux systems, which causes a symbol conflict when a `system` package is used with `libuuid` in the dependency tree, such as `xorg/system`. See https://github.com/conan-io/conan-center-index/pull/17427#issuecomment-1545476420 for a concrete example of this.

Closes #17337

Specify library name and version:  **libuuid/2.39**

---

### DIFF
--- a/recipes/util-linux-libuuid/all/conandata.yml
+++ b/recipes/util-linux-libuuid/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "1.3.0":
+    url: "https://github.com/util-linux/util-linux/archive/refs/tags/v2.39.tar.gz"
+    sha256: "186cb427cd6b4654f381357b60af7ffb0ae9bb50d7af7d87e0723858f7318b80"
--- a/recipes/util-linux-libuuid/all/conandata.yml
+++ b/recipes/util-linux-libuuid/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
-  "1.3.0":
+  "2.39":
     url: "https://github.com/util-linux/util-linux/archive/refs/tags/v2.39.tar.gz"
     sha256: "186cb427cd6b4654f381357b60af7ffb0ae9bb50d7af7d87e0723858f7318b80"
--- a/recipes/util-linux-libuuid/all/conanfile.py
+++ b/recipes/util-linux-libuuid/all/conanfile.py
@@ -3,7 +3,7 @@ from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import fix_apple_shared_install_name
 from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import copy, get, rm, rmdir, chdir
-from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.gnu import Autotools, AutotoolsToolchain, AutotoolsDeps
 from conan.tools.layout import basic_layout
 from conan.tools.scm import Version
 import os
@@ -71,6 +71,11 @@ class UtilLinuxLibuuidConan(ConanFile):
         if self.settings.os == "Windows":
             raise ConanInvalidConfiguration(f"{self.ref} is not supported on Windows")
 
+    def requirements(self):
+        if self.settings.os == "Macos":
+            # Required because libintl.{a,dylib} is not distributed via libc on Macos
+            self.requires("gettext/0.21")
+
     def build_requirements(self):
         self.tool_requires("libtool/2.4.7")
         self.tool_requires("m4/1.4.19")
@@ -85,6 +90,7 @@ class UtilLinuxLibuuidConan(ConanFile):
     def generate(self):
         env = VirtualBuildEnv(self)
         env.generate()
+
         tc = AutotoolsToolchain(self)
         tc.configure_args.append("--disable-all-programs")
         tc.configure_args.append("--enable-libuuid")
@@ -93,6 +99,9 @@ class UtilLinuxLibuuidConan(ConanFile):
         if "x86" in self.settings.arch:
             tc.extra_cflags.append("-mstackrealign")
         tc.generate()
+
+        deps = AutotoolsDeps(self)
+        deps.generate()
 
     def build(self):
         with chdir(self, self.source_folder):
--- a/recipes/util-linux-libuuid/all/conanfile.py
+++ b/recipes/util-linux-libuuid/all/conanfile.py
@@ -1,0 +1,99 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import copy, get, rm, rmdir, chdir
+from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.layout import basic_layout
+import os
+
+required_conan_version = ">=1.53.0"
+
+
+class UtilLinuxLibuuidConan(ConanFile):
+    name = "util-linux-libuuid"
+    description = "Universally unique id library"
+    url = "https://github.com/conan-io/conan-center-index"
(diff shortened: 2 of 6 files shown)


samuel-emrys:
I haven't been able to replicate the error message on my mac locally:

> ERROR: You must have libtool-2 installed to generate the util-linux build system.

If anybody has any suggestions on how to replicate this it would be appreciated.

recipes/util-linux-libuuid/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "1.3.0":
```
ericLemanissier, line 2: Is this version number really meaningfull ? It did not change in two years https://github.com/util-linux/util-linux/blame/master/meson.build#L11 in spite of several tags on the repo

samuel-emrys: Yeah, fair enough. I saw this in the meson.build file and assumed that it was well managed, and liked that it referred to the libuuid version directly rather than the bundle. It doesn't seem tenable to keep 1.3.0 though, so I'll update it to use the util-linux version

recipes/util-linux-libuuid/all/conanfile.py
```diff
@@ -0,0 +1,134 @@
...
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "bin"))
+        rmdir(self, os.path.join(self.package_folder, "sbin"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        fix_apple_shared_install_name(self)
+
+    def package_info(self):
+        self.cpp_info.set_property("pkg_config_name", "uuid")
+        self.cpp_info.set_property("cmake_target_name", "libuuid::libuuid")
+        self.cpp_info.set_property("cmake_file_name", "libuuid")
```
ericLemanissier, line 132:
```suggestion
        self.cpp_info.set_property("cmake_target_name", "LibUUID::LibUUID")
        self.cpp_info.set_property("cmake_file_name", "LibUUID")
```
cf https://gitlab.kitware.com/cmake/cmake/blob/HEAD/Source/Modules/FindLibUUID.cmake

recipes/util-linux-libuuid/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.8)
+project(test_package LANGUAGES C)
+
+find_package(libuuid REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} test_package.c)
+target_link_libraries(${PROJECT_NAME} PRIVATE libuuid::libuuid)
```
ericLemanissier, line 7:
```suggestion
target_link_libraries(${PROJECT_NAME} PRIVATE LibUUID::LibUUID)
```

gegles (approved):
Thanks @samuel-emrys!

SSE4 (approved):

ericLemanissier:
@RubenRBS @memsharded @czoido @uilianries can one of you review this PR ? thanks!

danimtb (approved):

uilianries (approved):
LGTM

jwillikers:
@samuel-emrys Thanks  for adding this!

recipes/util-linux-libuuid/all/conanfile.py
```diff
@@ -0,0 +1,134 @@
...
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "bin"))
+        rmdir(self, os.path.join(self.package_folder, "sbin"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        fix_apple_shared_install_name(self)
+
+    def package_info(self):
+        self.cpp_info.set_property("pkg_config_name", "uuid")
+        self.cpp_info.set_property("cmake_target_name", "LibUUID::LibUUID")
+        self.cpp_info.set_property("cmake_file_name", "LibUUID")
```
SpaceIm, line 132: Where does it come from?

samuel-emrys: See https://github.com/conan-io/conan-center-index/pull/17664#discussion_r1223051519

recipes/util-linux-libuuid/all/conanfile.py
```diff
@@ -0,0 +1,134 @@
...
+            # Required because libintl.{a,dylib} is not distributed via libc on Macos
+            self.requires("libgettext/0.21")
+
+    def build_requirements(self):
+        self.tool_requires("libtool/2.4.7")
+        self.tool_requires("m4/1.4.19")
+        self.tool_requires("pkgconf/1.9.3")
+        self.tool_requires("bison/3.8.2")
+        self.tool_requires("autoconf/2.71")
+        self.tool_requires("automake/1.16.5")
```
SpaceIm, line 90: Very bad idea to explicitly reference automake, autoconf and m4, it doesn't scale in build requirements, it's well known. Why this recipe doesn't follow other autoconf recipes guidelines?

samuel-emrys: I'm not experienced with automake. I used a few other recipes as guidelines but they obviously haven't incorporated the practices you're talking about and it didn't come up in review. It would be useful if those were documented somewhere easily discoverable. Feel free to submit a new PR with the necessary changes if that's a better way of managing these.

recipes/util-linux-libuuid/all/conanfile.py
SpaceIm: It's an internal file of CMake source code, not an official FindUUID.cmake file shipped in CMake program, so it shouldn't be used as a reference. This recipe shouldn't override cmake names.

recipes/util-linux-libuuid/all/conanfile.py
SpaceIm: A good reference is autotools recipes template.

(sorry, I won't fix this, I don't have the time. It will break when someone will bump automake version in libtool recipe)

recipes/util-linux-libuuid/all/conanfile.py
samuel-emrys: Ah, I'd missed that detail. Knowing that a convention has been used by CMake though (even though it's internal), wouldn't that still be the sensible choice for Conan? Since util-linux provides no config file, I would imagine people will take what already exists. Seems like this might be a sensible convention to standardise on..

recipes/util-linux-libuuid/all/conanfile.py
samuel-emrys: Good to know, but it wouldn't have addressed this - it has an explicit example requiring automake, and no discussion of when it shouldn't be required.

SpaceIm: > it has an explicit example requiring automake, and no discussion of when it shouldn't be required.

There is a comment explaining when it's required: https://github.com/conan-io/conan-center-index/blob/60fa77f44eafacf99b7f1d43650f13dd8db5d971/docs/package_templates/autotools_package/all/conanfile.py#L93C1-L96

PR #17671, opened 2023-05-23
rmm: new recipe

valgur:
Specify library name and version:  **rmm/23.06.00**

A recipe for RAPIDS Memory Manager (RMM): https://github.com/rapidsai/rmm

Depends on #17411.

---


recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+    def configure(self):
+        # Thrust needs to use CUDA as the backend for RMM.
+        # Disabled for ConanCenter CI.
+        # self.options["thrust"].backend = "cuda"
+        pass
+
+    def requirements(self):
+        self.requires("thrust/1.17.2", transitive_headers=True, transitive_libs=True)
+        self.requires("spdlog/1.11.0", transitive_headers=True, transitive_libs=True)
+        self.requires("fmt/9.1.0", transitive_headers=True, transitive_libs=True)
```
memsharded, line 42:
```suggestion
        self.requires("thrust/1.17.2")
        self.requires("spdlog/1.11.0")
        self.requires("fmt/9.1.0")
```
If ``package_type`` is a header-library, then it is not necessary to define the transitive_xxx, they are automatic.

recipes/rmm/all/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.15)
+project(RMM CXX)
+
+if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/RMM_RAPIDS.cmake)
+  file(DOWNLOAD https://raw.githubusercontent.com/rapidsai/rapids-cmake/branch-23.04/cmake/write_version_file.cmake
+       ${CMAKE_CURRENT_BINARY_DIR}/write_version_file.cmake)
+endif()
+include(${CMAKE_CURRENT_BINARY_DIR}/write_version_file.cmake)
```
memsharded, line 8: Why is this necessary? It is a header-only library, not being built. It seems this whole file is not used?

recipes/rmm/all/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
```
memsharded, line 1: The ``test_v1_package`` is no longer necessary, you can completely drop it.

memsharded (commented):
Thanks for your contribution!

The CI seems to be failing because ``thrust/1.17.0`` is the latest existing version, ``1.17.2`` is still not there.

recipes/rmm/all/CMakeLists.txt
valgur: It's not. Sorry about that. Forgot to delete this after some initial experiments.

recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+        )
+
+    def _patch_sources(self):
+        # Add missing include in logger.hpp
+        replace_in_file(
+            self,
+            os.path.join(self.source_folder, "include", "rmm", "logger.hpp"),
+            "#include <string>",
+            "#include <string>\n#include <array>",
```
uilianries, line 90: Is it possible to send it to the upstream as a patch? It looks be a bug.

recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,118 @@
...
+        export_conandata_patches(self)
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def configure(self):
+        # Thrust needs to use CUDA as the backend for RMM.
+        # Disabled for ConanCenter CI.
+        # self.options["thrust"].backend = "cuda"
+        pass
```
uilianries, line 50:
```suggestion
    def configure(self):
        # Thrust needs to use CUDA as the backend for RMM.
        # Disabled for ConanCenter CI.
        # self.options["thrust"].backend = "cuda"
        pass
```
Please, move it to validate, but as a warning message. `self.output.warning("...")`

recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,118 @@
...
+        header_path.write_text(
+            textwrap.dedent(
+                f"""\
+            #pragma once
+            #define RMM_VERSION_MAJOR {int(major)}
+            #define RMM_VERSION_MINOR {int(minor)}
+            #define RMM_VERSION_PATCH {int(patch)}
+            """
+            )
+        )
```
uilianries, line 82:
```suggestion
        major, minor, patch = self.version.split(".")[:3]
        header_path = Path(self.source_folder) / "include" / "rmm" / "version_config.hpp"
        header_path.parent.mkdir(parents=True, exist_ok=True)
        header_path.write_text(
            textwrap.dedent(
                f"""\
            #pragma once
            #define RMM_VERSION_MAJOR {int(major)}
            #define RMM_VERSION_MINOR {int(minor)}
            #define RMM_VERSION_PATCH {int(patch)}
            """
            )
        )
```
Does version_config.hpp exist in the upstream? If don't, please, remove it. You could send a PR to the upstream instead.

recipes/rmm/all/conanfile.py
valgur: Already did and will be included in the next release: https://github.com/rapidsai/rmm/pull/1295

recipes/rmm/all/conanfile.py
valgur: Added a comment with the reasoning for this.

recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+        if "backend" not in self.options["thrust"] or self.options["thrust"].backend != "cuda":
```
uilianries, line 63:
```suggestion
        if self.dependencies["thrust"].options.get_safe("backend") != "cuda":
```
Accessing dependencies' options by `self.options` is no longer supported in Conan 2.x. `self.dependencies` is the new form and backward compatible with Conan 1.x

recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+        header_path.write_text(
+            textwrap.dedent(
+                f"""\
+            #pragma once
+            #define RMM_VERSION_MAJOR {int(major)}
+            #define RMM_VERSION_MINOR {int(minor)}
+            #define RMM_VERSION_PATCH {int(patch)}
+            """
+            )
+        )
```
uilianries, line 84:
```suggestion
        save(self, os.path.join(self.source_folder, "include", "rmm", "version_config.hpp")
            textwrap.dedent(
                f"""\
            #pragma once
            #define RMM_VERSION_MAJOR {Version(self.version).major}
            #define RMM_VERSION_MINOR {Version(self.version).minor}
            #define RMM_VERSION_PATCH {Version(self.version).patch}
            """
            )
        )
```

Conan tools offers [save](https://docs.conan.io/2/reference/tools/files/basic.html#conan.tools.files.files.save) that safer than using regular methods, because it also considers write permission in the folder (rare but we have problems in Windows).

uilianries (commented):
looking good, only few comments now. Using conan tools there are some advantages, like file permission and encoding.

recipes/rmm/all/conanfile.py
```diff
@@ -0,0 +1,117 @@
...
+
+class RmmConan(ConanFile):
+    name = "rmm"
+    description = "RAPIDS Memory Manager"
+    license = "Apache-2.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/rapidsai/rmm"
+    topics = ("cuda", "memory-management", "memory-allocation", "rapids", "header-only")
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
```
uilianries, line 22:
```suggestion
    settings = "os", "arch", "compiler", "build_type"
    no_copy_source = True
```
Avoid copying source files to the build folder, as you are not building something, only copying to the package folder . 

https://docs.conan.io/2/reference/conanfile/attributes.html#no-copy-source

uilianries (approved):
LGTM

SSE4 (approved):

PR #17672, opened 2023-05-24
jemalloc: compile on Macos M1

jiaoew1991:
Specify library name and version:  **jemalloc/all**

---

### DIFF
--- a/recipes/jemalloc/all/conanfile.py
+++ b/recipes/jemalloc/all/conanfile.py
@@ -84,7 +84,8 @@ class JemallocConan(ConanFile):
         if self.settings.compiler == "clang" and Version(self.settings.compiler.version) <= "3.9":
             raise ConanInvalidConfiguration("Unsupported compiler version")
         if self.settings.os == "Macos" and self.settings.arch not in ("x86_64", "x86"):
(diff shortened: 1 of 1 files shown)


recipes/jemalloc/all/conanfile.py
```diff
@@ -84,7 +84,8 @@ def validate(self):
         if self.settings.compiler == "clang" and Version(self.settings.compiler.version) <= "3.9":
             raise ConanInvalidConfiguration("Unsupported compiler version")
         if self.settings.os == "Macos" and self.settings.arch not in ("x86_64", "x86"):
-            raise ConanInvalidConfiguration("Unsupported arch")
+            if Version(self.settings.compiler.version) < "5.3.0":
```
jcar87, line 87: This is testing against the compiler version, rather than the library version. Maybe you mean 
```Version(self.version) < "5.3.0"```

On the other hand, I can't find anything in the changelog that may indicate that this is resolved however:
https://github.com/jemalloc/jemalloc/blob/dev/ChangeLog

but then again the recipe left no comments as to what the reason was for being compatible with arm64 on macOS

PR #17540, opened 2023-05-12
arrow: add version 12.0.0

raulcd:
Specify library name and version:  **arrow/12.0.0**

On the Arrow 12.0.0 release [PLASMA was removed](https://github.com/apache/arrow/issues/33243) and [ACERO was moved to its own shared library](https://github.com/apache/arrow/issues/15280).

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17691, opened 2023-05-24
arrow: fix the library name for the libacero component

mpimenov:
I think we got this part wrong in https://github.com/conan-io/conan-center-index/pull/17540
It's not clear if this name stays (see https://github.com/apache/arrow/issues/15280) but currently
the file being created is libarrow_acero.a

I'm not certain what to do with names of other components, such as pkg_config: for arrow_substrait
we prepend "arrow_" everywhere, for "arrow_flight_sql" we do not. I've chosen to make
the smallest possible diff that I am sure about.


mpimenov:
@raulcd PTAL

raulcd:
I am currently working on the release for 12.0.1 and I noticed other things that I didn't add to Conan. I will have to perform a small refactoring around the Acero / Plasma deprecation. There were changes on the Arrow repo that I missed when doing the change, so I'll have to reuse the changes that were performed there: https://github.com/apache/arrow/blob/9888ac36c1424840a5d95d1bf6a88cbb7671f0a3/ci/conan/all/conanfile.py
Thanks for this PR. I am ok with it being merged but there will be more changes required.

uilianries (approved):

SSE4 (approved):

PR #17692, opened 2023-05-24
arrow: add a missing dependency

mpimenov:
In correspondence of what is done here:
https://github.com/apache/arrow/blob/apache-arrow-12.0.0/cpp/src/arrow/dataset/CMakeLists.txt#L56-L63


RubenRBS (approved):
Good catch, thanks a lot!

PR #17575, opened 2023-05-15
embag: add new recipe

valgur:
Specify library name and version:  **embag/0.0.42**

Resolves #17574.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17698, opened 2023-05-25
bazel: update to Conan v2, add v6.2.0

valgur:
Specify library name and version:  **bazel/6.2.0**

For #17575.

---


recipes/bazel/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,12 @@
+from conan.tools.build import can_run
```
memsharded, line 1: ``test_v1_package`` can be dropped already, no longer a requirement

memsharded (commented):
Looking good, but it seems there is a mistake in the Windows URLs in ``conandata.yml``, that is causing it to fail?

memsharded (approved):

PR #17699, opened 2023-05-25
Extra cmake modules v2 migration

weatherhead99:
Specify library name and version:  **extra-cmake-modules/5.106.0**

This is updating the extra-cmake-modules library to 5.106.0 and also making the recipe conan 2.0 compatible.
---

- [ X ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ X ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ X ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.


weatherhead99:
I do not understand these build fails as [ca51391](https://github.com/conan-io/conan-center-index/pull/17699/commits/ca51391c4a739623af30f33cbfe085b6f9ecf48c) should have fixed that stray tab...

recipes/extra-cmake-modules/all/conanfile.py
```diff
@@ -8,49 +11,50 @@ class ExtracmakemodulesConan(ConanFile):
...
-        self.cpp_info.builddirs = ["res/ECM/cmake", "res/ECM/kde-modules", "res/ECM/modules", "res/ECM/test-modules", "res/ECM/toolchain"]
+        lic_folder_inst = os.path.join(self.package_folder, "licenses")
+        copy(self, "*", src=lic_folder, dst=lic_folder_inst)
+
+        cmake = CMake(self)
+        cmake.install()
 
     def package_id(self):
-        self.info.header_only()
+        self.info.settings.clear()
```
jwillikers, line 55: Should this just be `self.info.clear()`?

```suggestion
        self.info.clear()
```

jwillikers (approved):

recipes/extra-cmake-modules/all/conanfile.py
weatherhead99: I don't know, should it be? Certainly there are no current options either (cf settings) and I don't see what would want to be added in future... so if you prefer this then fine!

jwillikers: It would probably be good to stick with it for consistency with what's documented [here](https://docs.conan.io/2/tutorial/creating_packages/other_types_of_packages/header_only_packages.html#header-only-library-with-tests) and in the header-only package template.

weatherhead99: ok, changed in 16f04614

jwillikers (approved):
Thanks, LGTM!

jwillikers (approved):

recipes/extra-cmake-modules/all/conanfile.py
```diff
@@ -8,49 +12,65 @@ class ExtracmakemodulesConan(ConanFile):
...
 
     def package(self):
-        cmake = self._configure_cmake()
-        cmake.install()
-        self.copy("testhelper.h", src=os.path.join(self.source_folder, self._source_subfolder, "tests/ECMAddTests"), dst="res/tests")
-        self.copy("*", src=os.path.join(self.source_folder, self._source_subfolder, "LICENSES"), dst="licenses")
+        cm_folder = "{}-{}".format(self.name, self.version)
+        lic_folder_st = os.path.join(cm_folder, "LICENSES")
+        lic_folder = os.path.join(self.source_folder, lic_folder_st)
+        self.output.info("license folder: {}".format(lic_folder))
```
uilianries, line 58:
```suggestion
```

Not needed as when running package step, all copied files are listed, including license. Plus, there is will be an warning message from the client in case licenses folder is empty.

recipes/extra-cmake-modules/all/conanfile.py
```diff
@@ -8,49 +12,65 @@ class ExtracmakemodulesConan(ConanFile):
...
+
+        cmake = CMake(self)
+        cmake.install()
 
     def package_id(self):
-        self.info.header_only()
+        self.info.clear()
+
+    def package_info(self):
+        # this package is CMake files, it doesn't need an extra one generating
```
uilianries, line 70:
```suggestion
        self.cpp_info.bindirs = []
        self.cpp_info.libdirs = []
```

Both library and executable folders are not used here, right?

recipes/extra-cmake-modules/all/test_package/conanfile.py
```diff
@@ -1,17 +1,34 @@
...
-        if not tools.cross_building(self):
-            bin_path = os.path.join("bin", "example")
-            self.run(bin_path, run_environment=True)
+        if can_run(self):
+            progname = "example"
+            if is_msvc(self):
+                progname += ".exe"
 
+            runpath = os.path.join(self.cpp.build.bindirs[0], progname)
+            self.run(runpath, env="conanrun")
```
uilianries, line 34:
```suggestion
        if can_run(self):
            bin_path = os.path.join(self.cpp.build.bindir, "example")
            self.run(bin_path, env="conanrun")
```


You could simplify even more: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/test_package/conanfile.py

recipes/extra-cmake-modules/all/conandata.yml
```diff
@@ -11,3 +11,9 @@ sources:
   "5.93.0":
     url: "https://download.kde.org/stable/frameworks/5.93/extra-cmake-modules-5.93.0.tar.xz"
     sha256: "093dea7b11647bc5f74e6971d47ef15b5c410cba2b4620acae00f008d5480b21"
+  "5.106.0":
```
uilianries, line 14: Do you need the version 5.106.0? Otherwise, I would ask to keep only 5.108.0 to build extra not used packages. In case someone need that specific version, we add later.

uilianries (commented):
Thank you for your contribution!

recipes/extra-cmake-modules/all/conanfile.py
weatherhead99: ok, will remove this (wasn't my code, was left over from conan 1 era)

recipes/extra-cmake-modules/all/conandata.yml
weatherhead99: removed

recipes/extra-cmake-modules/all/test_package/conanfile.py
weatherhead99: I had a difficult time getting this to compile on Windows under MSBuild. But I've checked on my own windows machine and this now does seem to work, so have addid this change to commits below

uilianries: @weatherhead99 on Windows, is not needed to add `.exe` suffix, the cli should solve it automatically. It's not a Conan stuff, but default Windows behavior.

weatherhead99: fine. I only added this in the first place because it wouldn't run in my particular powershell configuration but I'm no windows expert so I assume my system is weird somehow

recipes/extra-cmake-modules/all/test_package/conanfile.py
```diff
@@ -1,17 +1,34 @@
...
     def test(self):
-        if not tools.cross_building(self):
-            bin_path = os.path.join("bin", "example")
-            self.run(bin_path, run_environment=True)
+        if can_run(self):
+            progname = "example"
+            if is_msvc(self):
+                progname += ".exe"
 
+            runpath = os.path.join(self.cpp.build.bindir, progname)
```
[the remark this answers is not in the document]
weatherhead99: I did not write most of this, why was it accepted in the first place? But ok, will simplify again

uilianries (approved):
LGTM!

SSE4 (approved):

PR #17702, opened 2023-05-25
benchmark: add pfm option

andioz:
Specify library name and version:  **benchmark/1.8.0**

Benchmark has a build option for enabling performance counters: https://github.com/google/benchmark/blob/v1.5.4/docs/perf_counters.md

We want to use this feature, therefore I try to add an option which is disabled by default. As far as I have seen this is currently only for Linux supported, and requires the "libpfm4" library, which is available via Conan. This feature was added with version 1.5.4, therefore I include a version check too (even if our supported versions starts with a higher version number). 

Some time past by since I made my last contribution, and this is my very first in the time of Conan 2.x - I hope I did everything in the right way, please correct me if not.

---

### DIFF
--- a/recipes/benchmark/all/conanfile.py
+++ b/recipes/benchmark/all/conanfile.py
@@ -38,7 +38,7 @@ class BenchmarkConan(ConanFile):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
-        if self.settings.os != "Linux":
+        if self.settings.os != "Linux" or Version(self.version) < "1.5.4":
             del self.options.enable_libpfm
 
     def configure(self):
--- a/recipes/benchmark/all/conanfile.py
+++ b/recipes/benchmark/all/conanfile.py
@@ -55,7 +55,7 @@ class BenchmarkConan(ConanFile):
 
     def requirements(self):
         if self.options.get_safe("enable_libpfm"):
-            self.requires("libpfm4/4.12.0")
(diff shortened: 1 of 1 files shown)


Hopobcn (approved):

RubenRBS (changes requested):
Requesting changes until the linked PR is merged so this gets the latest revision :)

andioz:
> Requesting changes until the linked PR is merged so this gets the latest revision :)

Do you plan to update the libpfm version too? (4.13.0 is available)

RubenRBS:
Ah I missed that, yes! Will do

recipes/benchmark/all/conanfile.py
```diff
@@ -49,6 +53,10 @@ def validate(self):
         if Version(self.version) < "1.7.0" and is_msvc(self) and self.options.shared:
             raise ConanInvalidConfiguration(f"{self.ref} doesn't support msvc shared builds")
 
+    def requirements(self):
+        if self.options.get_safe("enable_libpfm"):
+            self.requires("libpfm4/4.12.0")
```
RubenRBS, line 58:
```suggestion
            self.requires("libpfm4/4.13.0")
```

RubenRBS (changes requested):

RubenRBS (approved):

jwillikers (approved):

PR #17707, opened 2023-05-25
Create recipe for foxglove-schemas-protobuf package

kevswims:
Specify library name and version:  **foxglove-schemas-protobuf/0.1.0**

I am submitting this library to make it easier to consume the protobuf definitions and generated code from other applications, specifically ones that use the foxglove-websocket package which is already published.

For more discussion on this see https://github.com/foxglove/schemas/pull/110

---


RubenRBS:
Thanks a lot for your contribution and your patience while we reviewed the rest of the PR backlog. At first glance the changes look good, but will let you know once I can properly review them tomorrow, I just wanted to say thanks for now :)

recipes/foxglove-schemas-protobuf/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
```
RubenRBS, line 59:
```suggestion
        check_min_vs(self, 191)
        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
```

recipes/foxglove-schemas-protobuf/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import get, copy, apply_conandata_patches, export_conandata_patches
```
RubenRBS, line 5:
```suggestion
from conan.tools.files import get, copy, apply_conandata_patches, export_conandata_patches
from conan.tools.microsoft import check_min_vs
```

recipes/foxglove-schemas-protobuf/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def build(self):
+        self._patch_sources()
```
RubenRBS:
```suggestion
        apply_conandata_patches(self)
```

RubenRBS:
Hi @kevswims thanks a lot for your contribution and patience. I've pushed some minor changes, and the rest looks good to me :)

recipes/foxglove-schemas-protobuf/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    package_type = "library"
+
+    @property
+    def _min_cppstd(self):
+        return 17
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "Visual Studio": "16.9",
```
jcar87, line 42: I think this is not a valid version in `settings.yml`, unless I'm missing something

recipes/foxglove-schemas-protobuf/all/patches/0001-Add-cmakelists.txt-for-building-with-Conan.patch
```diff
@@ -0,0 +1,45 @@
...
+---
+ CMakeLists.txt | 25 +++++++++++++++++++++++++
+ 1 file changed, 25 insertions(+)
+ create mode 100644 CMakeLists.txt
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+new file mode 100644
+index 0000000..c5ba3e8
+--- /dev/null
++++ b/CMakeLists.txt
```
jcar87, line 15: if this is adding a new file, I think it would make it a lot easier to maintain moving forwards if this was in the recipe folder and then copied to the src folder as part of the recipe creation process, _rather_ than a patch that gets applied to add a file. This is done in other recipes as well for similar cases

recipes/foxglove-schemas-protobuf/all/patches/0001-Add-cmakelists.txt-for-building-with-Conan.patch
```diff
@@ -0,0 +1,45 @@
...
++    OUTPUT ${proto_sources}
++    COMMAND ${CMAKE_COMMAND} -E make_directory autogenerated
++    COMMAND ${Protobuf_PROTOC_EXECUTABLE} --proto_path=${CMAKE_CURRENT_SOURCE_DIR}/schemas/proto --cpp_out=autogenerated ${all_protos}
++)
++
++add_library(foxglove_schemas_protobuf ${proto_sources})
++target_include_directories(foxglove_schemas_protobuf PRIVATE ${CMAKE_CURRENT_BINARY_DIR}/autogenerated)
++target_link_libraries(foxglove_schemas_protobuf PUBLIC protobuf::libprotobuf)
++set_property(TARGET foxglove_schemas_protobuf PROPERTY CXX_STANDARD 17)
++install(TARGETS foxglove_schemas_protobuf)
```
jcar87, line 41: For easier maintainability I would suggest having `cmake.install()` also handle the logic of copying the relevant .pb.h files, or just disregard it altogether and copy also the library in the `package()` method - doing both seems a bit split when the things are meant to "travel" together

recipes/foxglove-schemas-protobuf/all/conanfile.py
kevswims: I think this came over when I copied a different recipe. I'll take it out since I don't even know if it applies.

recipes/foxglove-schemas-protobuf/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+        apply_conandata_patches(self)
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.install()
+        copy(self, "*.proto", src=self.source_folder, dst=os.path.join(self.package_folder, "include/foxglove"), keep_path=False)
```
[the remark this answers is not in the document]
kevswims: This is how the protobuf package from google does it.

The reasoning is that you can include proto files inside each other. An example would be if I wanted to create a gRPC server with these message types I would need these proto files to include in the server proto file.

recipes/foxglove-schemas-protobuf/all/patches/0001-Add-cmakelists.txt-for-building-with-Conan.patch
kevswims: This is fixed so headers and the proto files are installed in the cmake now.

kevswims:
> Hi @kevswims thanks a lot for your contribution and patience. I've pushed some minor changes, and the rest looks good to me :)

Thanks @RubenRBS I got your changes squashed in along with the changes suggested by @jcar87 . I think this is ready for another review pending the builds passing.

recipes/foxglove-schemas-protobuf/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+        if self.settings.os == "Windows" and self.options.shared:
+            raise ConanInvalidConfiguration("Windows shared builds are not supported yet.")
+
+    def build_requirements(self):
+        self.tool_requires("protobuf/3.21.9")
+
+    def requirements(self):
+        self.requires("protobuf/3.21.9", transitive_headers=True, transitive_libs=True)
```
uilianries, line 71: Protobuf is very strict when generating header, pushing the specific version was used to create .hh and .cc files. I should validate if this recipe needs or not to depends on full_package_mode for protobuf. Otherwise, consumer could have a compiler error when using a different Protobuf version.

uilianries: @kevswims correct, like I said, protobuf is very strict when generating headers and sources, so this package is not a regular library, but a product from proto files, it's another thing, the regular rules of compatibility may not fit in case a consumer uses another version of protobuf, but the headers are not compatible due the explicit define made by protobuf.

kevswims: @uilianries I understand that part. Is there a specific thing that needs to be changed in the recipe to enforce that? I am not familiar with the full_package_mode settings and it does not even appear to be an option in conan 2.0.

uilianries: @kevswims let me confirm it today. But in case you would need to do it:

```python
def package_id(self):
    self.info.requires["protobuf"].patch_mode()
```

https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-id-mode

https://docs.conan.io/2/reference/conanfile/attributes.html#package-id-embed-non-embed-unknown-mode

uilianries: All fine, protobuf is minor compatible. I checked locally right, by building different applications, using different patch version of Protobuf and adding them a very same application. No errors.

Plus, it's documented about its backward compatibility with minor versions: https://github.com/protocolbuffers/protobuf/blob/532d2a24800890890d764e2de8fc2716eb3213e9/java/README.md?plain=1#L142

uilianries (commented):
LGTM

jcar87 (approved):

SSE4 (approved):

PR #17708, opened 2023-05-26
bshoshany-thread-pool: add version 3.5.0

bshoshany:
Specify library name and version:  **bshoshany-thread-pool/3.5.0**

---

### DIFF
--- a/recipes/bshoshany-thread-pool/all/conanfile.py
+++ b/recipes/bshoshany-thread-pool/all/conanfile.py
@@ -38,6 +38,7 @@ class BShoshanyThreadPoolConan(ConanFile):
         return "source_subfolder"
 
     def package(self):
+        copy(self, "*.hpp", self.source_folder, os.path.join(self.package_folder, "include"))
         copy(self, "*.hpp", os.path.join(self.source_folder, "include"), os.path.join(self.package_folder, "include"))
         copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
 
--- a/recipes/bshoshany-thread-pool/all/conanfile.py
+++ b/recipes/bshoshany-thread-pool/all/conanfile.py
@@ -38,7 +38,7 @@ class BShoshanyThreadPoolConan(ConanFile):
         return "source_subfolder"
 
     def package(self):
-        copy(self, "*.hpp", self.source_folder, os.path.join(self.package_folder, "include"))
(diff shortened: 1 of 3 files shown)


bshoshany:
I'm not really sure why all the checks are failing. Also, it seems to be re-checking all the previous releases for some reason. Can someone please take a look at this?

bshoshany:
I did all the same things I did in previous PRs, but it's failing checks. I think it may be due to the fact that the header files have been moved to the `include` subfolder, but it does work locally on my system. Again, I would appreciate if someone who understand conan better than I do could take a look.

recipes/bshoshany-thread-pool/all/conanfile.py
```diff
@@ -38,7 +38,7 @@ def _source_subfolder(self):
         return "source_subfolder"
 
     def package(self):
-        copy(self, "*.hpp", self.source_folder, os.path.join(self.package_folder, "include"))
+        copy(self, "include/*.hpp", self.source_folder, self.package_folder)
```
uilianries, line 41:
```suggestion
        copy(self, "*.hpp", os.path.join(self.source_folder, "include"), self.package_folder)
```

The second parameter is `pattern` only, not path. That's why you are having the error about empty packages.

bshoshany:
Thanks so much @uilianries! I think I understand what I did wrong now, `conanfile.py` is supposed to be used for all versions (since it's in the `all` folder) but my version only worked for v3.5.0 due to the path change in this version. So with your change, the same file now works for all versions. Just writing this here to remind this to myself in case I make any similar changes in future versions.

Looks like all the checks are passing now except 2, and I cannot see the details for these 2 since it says "403 Forbidden" when I click on them. Any ideas?

bshoshany:
@uilianries great, thanks for your help.

bshoshany:
I think I figured out the problem, the change you made still didn't fix compatibility with older releases where the `.hpp` files were not in the `include` subfolder, so now I added 2 separate lines to account for `.hpp` files either in the root folder or the `include` subfolder. If this can be done in 1 line, that would be more elegant, but I could not find a way to make `conan.tools.files.copy()` look for files recursively in all subfolders.

I checked and this now works locally for all previous versions, so hopefully all the CI checks will pass this time.

uilianries (approved):
LGTM

bshoshany:
Great, thank you so much for your help! :)

ericLemanissier (approved):

PR #17351, opened 2023-05-01
imgui: add version 1.89.5

dBagrat:
Generated and committed by [Conan Center Bot](https://github.com/qchateau/conan-center-bot) Find more updatable recipes in the [GitHub Pages](https://qchateau.github.io/conan-center-bot/)

Specify library name and version: imgui/1.89.5

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #17717, opened 2023-05-26
imgui: Fix docking branch version detection, add package_type, fix & simplify test_package, more

RubenRBS:
Specify library name and version:  **imgui/all**

While debugging https://github.com/conan-io/conan-center-index/pull/17351, found a few things that could be improved

* Throw an invalid configuration for gcc 5 so that the linked PR passes once this PR is merged.
* Fix docking branch version detection regex
* Fix docking branch testing in test_package. The define was never being triggered
* Simplify test_package. We really don't need to render 50 frames to see that we packaged it correctly


uilianries (approved):
The executable also requires C++11 support for 1.89.5. Plus, Conan 1.x only access self.dependencies on generate() method when testing.

danimtb (approved):

danimtb (approved):

uilianries (approved):
LGTM

PR #17719, opened 2023-05-26
Added libqasm recipe

rturrado:
Specify library name and version:  **libqasm/0.6.5**

libqasm is a library for handling cQASM files using C++ or Python (https://github.com/QuTech-Delft/libqasm, https://libqasm.readthedocs.io/en/latest/).
It is used by OpenQL, a framework for high-level quantum programming in C++/Python (https://github.com/QuTech-Delft/OpenQL, https://openql.readthedocs.io/en/latest/).
Both libqasm and OpenQL are developed by QuTech-Delft (TU Delft), for which I work as a software engineer.
It is a new addition to ConanCenter.
I basically want OpenQL users to be able to use libqasm in a simple way (i.e. without needing to install libqasm's dependencies manually).

---

      I've used conan 2.0.4.
      I've tried building both in Linux and Windows in release and debug modes. These are the commands I use in each platform:
      conan create all/conanfile.py --version 0.5.0
      conan create all/conanfile.py --version 0.5.0 -o libqasm/*:build_type=Debug


recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,113 @@
...
+        "fPIC": True,
+        "build_python": False,
+        "build_tests": False,
+        "compat": False,
+        "tree_gen_build_tests": False
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 23
```
RubenRBS, line 41:
```suggestion
        return 11
```

Seems like the min cppstd is 11 https://github.com/QuTech-Delft/libqasm/blob/develop/CMakeLists.txt#L8?

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,113 @@
...
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "13",
+            "clang": "13",
+            "gcc": "11",
+            "msvc": "19"
+        }
+
+    def build_requirements(self):
+        self.build_requires("cmake/3.20.1")
```
RubenRBS, line 53:
```suggestion
```

Unless there's a specific reason for such version, the CMakeList upstream seems to support pretty old ones?

Note that removing this line also fixes the v2 pipeline errors (tested locally)

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,113 @@
...
+        self.folders.source = "."
+        self.folders.build = os.path.join("build", str(self.settings.build_type))
+        self.folders.generators = os.path.join(self.folders.build, "generators")
+
+        self.cpp.package.libs = ["cqasm"]
+        self.cpp.package.includedirs = ["include"]
+        self.cpp.package.libdirs = ["lib"]
+
+        self.cpp.source.includedirs = ["include"]
+        self.cpp.build.libdirs = ["."]
```
RubenRBS, line 75: Isn't this mostly what `cmake_layout()` does? Why is it explicitly written here?

RubenRBS (changes requested):
Hi! Thanks a lot for your contribution. I have some minor changes needed before merging, but overall looks good, thanks a lot!

rturrado:
> Hi! Thanks a lot for your contribution. I have some minor changes needed before merging, but overall looks good, thanks a lot!

Many thanks, Rubén! And thanks for having a look and helping me with your review. I'm now going to reply to you comments ;-)

recipes/libqasm/all/conanfile.py
rturrado: `libqasm` builds with C++11 at the moment, but I'm going to be working on a evolution of the library, in order to support the version 3.0 of the `cqasm` language and, for that evolution, I had thought of using C++23 features. However, since Conan Center seems to build binaries only up to C++20, I'm finally going to use C++20, at least for the near future.

recipes/libqasm/all/conanfile.py
rturrado: I've removed this line in the last commit. C++20 only needs CMake 3.12, and Conan pipelines use CMake 3.15.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"]["0.5.2"], strip_root=True)
```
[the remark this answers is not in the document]
rturrado: I'm hardcoding the version because I got a Linter error telling me that I cannot set `version` as a setting. Where is this Conan reading this `self.version` from? Thanks!

recipes/libqasm/all/conanfile.py
rturrado: `libqasm` has a root folder with a `CMakeLists.txt` file and a `src` folder. I need `folders.source` to point to `.`, in order to read that top `CMakeLists.txt`, which is setting a variable that I need in `src/CMakeLists.txt`. I think I tried with `cmake_layout` and it didn't let me set `src_folder = "."`, but it was forcing me to set `src_folder = "src"`, hence this more verbose code.

SpaceIm: Looking at upstream CMakeLists, I don't think you need anything special, you can just use cmake_layout.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+    description = "Library to parse cQASM files"
+    topics = ("code generation", "parser", "compiler", "quantum compilation", "quantum simulation")
+
+    # Binary configuration
+    settings = "os", "compiler", "build_type", "arch"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "build_python": [True, False],
+        "build_tests": [True, False],
```
SpaceIm, line 25: I don't think that test like option can be accepted

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+
+        self.cpp.package.libs = ["cqasm"]
+        self.cpp.package.includedirs = ["include"]
+        self.cpp.package.libdirs = ["lib"]
+
+        self.cpp.source.includedirs = ["include"]
+        self.cpp.build.libdirs = ["."]
+
+    def source(self):
+        get(self, **self.conan_data["sources"]["0.5.4"], strip_root=True)
```
SpaceIm, line 63:
```suggestion
        get(self, **self.conan_data["sources"][self.version], strip_root=True)
```

recipes/libqasm/all/test_package/conanfile.py
```diff
@@ -0,0 +1,30 @@
...
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.build import cross_building
+
+
+class LibqasmTestConan(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    # VirtualBuildEnv and VirtualRunEnv can be avoided if "tools.env.virtualenv:auto_use" is defined
+    # (it will be defined in Conan 2.0)
+    generators = "CMakeDeps", "CMakeToolchain", "VirtualBuildEnv", "VirtualRunEnv"
+    apply_env = False
```
SpaceIm, line 13:
```suggestion
    # VirtualBuildEnv and VirtualRunEnv can be avoided if "tools.env.virtualenv:auto_use" is defined
    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv"
```

recipes/libqasm/all/test_package/conanfile.py
```diff
@@ -0,0 +1,30 @@
+import os
+
+from conan import ConanFile
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.build import cross_building
```
SpaceIm, line 5:
```suggestion
from conan.tools.build import can_run
```

recipes/libqasm/all/test_package/conanfile.py
```diff
@@ -0,0 +1,30 @@
...
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def layout(self):
+        cmake_layout(self)
+
+    def test(self):
+        if not cross_building(self):
```
SpaceIm, line 28:
```suggestion
        if can_run(self):
```

SpaceIm (changes requested):

recipes/libqasm/all/conanfile.py
RubenRBS: Let me talk with the team on when you can expect C++23 configurations to be available and I'll get back to you in a few days, please feel free to use it for your project!

recipes/libqasm/all/conanfile.py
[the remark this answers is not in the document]
RubenRBS: It's built in into Conan. When calling `conan create`, just add a `--version=0.5.2` argument to your invocation and Conan will understand that it's targeting that version

recipes/libqasm/all/test_package/conanfile.py
rturrado: Done!

recipes/libqasm/all/test_package/conanfile.py
rturrado: Done!

recipes/libqasm/all/test_package/conanfile.py
rturrado: Done!

recipes/libqasm/all/conanfile.py
rturrado: Hi @SpaceIm, many thanks for your review!

I think I had previously spoken with you through Slack, and you had commented to me the idea of building the tests always by default and then using something like `skip_tests` to avoid that build. I have two comments regarding that:

1. I don't like the idea of building tests by default. I would like the uninformed user that is building my package to get a quick and clean, without tests and test binaries, build.

2. In case the `skip_tests` had other benefits that I cannot yet see, how could I connect it with the `CMakeLists.txt`? I understand it would be with a line such as `tc.variables["LIBQASM_BUILD_TESTS"] = not self.options.skipt_tests`?

Thanks!

recipes/libqasm/all/conanfile.py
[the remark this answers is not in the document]
rturrado: Done!

recipes/libqasm/all/conanfile.py
rturrado: Yes, this is what I had seen before (see my explanation above).

I have tried using `cmake_layout` again, and it has failed a [Conan v2 Lint](https://github.com/conan-io/conan-center-index/actions/runs/5119660631/jobs/9205202841?pr=17719).

It basically complains saying:
```
Error: recipes/libqasm/all/conanfile.py:53: [E9013(conan-layout-src-folder-is-src), LibqasmConan.layout] layout should set `src_folder` to `src`
```

But I cannot do that. I need to set my `src_folder` to `.`.

rturrado:
> Hi all,
> 
> Any idea why this hook is reporting an [error](https://c3i.jfrog.io/c3i/misc/logs/pr/17719/12-linux-gcc/libqasm/0.5.5//b91d4c850085ce839f1cd3308ec904a1a58da7ea-build.txt) for `fPIC`?
> 
> ```
> [HOOK - conan-center.py] pre_build(): ERROR: [FPIC MANAGEMENT (KB-H007)] 'fPIC' option not managed correctly. Please remove it for shared option: del self.options.fPIC (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H007-FPIC-MANAGEMENT) 
> ```
> 
> I have already updated the `conanfile.py` to remove the `fPIC` option for shared libraries (apart from for Windows):
> 
> ```
> def config_options(self):
>     if self.options.shared:
>         del self.options.fPIC
>     if self.settings.os == "Windows":
>         del self.options.fPIC
> ```
> 
> Is it because I still have `fPIC` as an `options` and `default_options`?
> 
> ```
>     options = {
>         "shared": [True, False],
>         "fPIC": [True, False],
>         "build_python": [True, False],
>         "build_tests": [True, False],
>         "compat": [True, False],
>         "tree_gen_build_tests": [True, False]
>     }
>     default_options = {
>         "shared": False,
>         "fPIC": True,
>         "build_python": False,
>         "build_tests": False,
>         "compat": False,
>         "tree_gen_build_tests": False
>     }
> ```
> 
> Thanks.

I've already fixed this by changing the way `fPIC` option is removed:

```
def configure(self):
    if self.options.shared:
        self.options.rm_safe("fPIC")
```

rturrado:
Hi all,

My latest commit is failing in the installation phase because no `licenses` folder is found (see [here](https://c3i.jfrog.io/c3i/misc/logs/pr/17719/13-linux-gcc/libqasm/0.5.6//a25d6c83542b56b72fdaa05a85db5d46f5f0f71c-build.txt)). This is the actual error message:

```
[HOOK - conan-center.py] post_package(): ERROR: [PACKAGE LICENSE (KB-H012)] No 'licenses' folder found in package: /home/conan/w/prod/BuildSingleReference/.conan/data/libqasm/0.5.6/_/_/package/a25d6c83542b56b72fdaa05a85db5d46f5f0f71c  (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H012-PACKAGE-LICENSE)
```

I have a couple of questions:
1. Is this a CMake/installation issue? I have a `LICENSE.md` file in my project. Maybe I only need to copy it to the installation folder.
2.

rturrado:
> Hi all,
> 
> My latest commit is failing in the installation phase because no `licenses` folder is found (see [here](https://c3i.jfrog.io/c3i/misc/logs/pr/17719/13-linux-gcc/libqasm/0.5.6//a25d6c83542b56b72fdaa05a85db5d46f5f0f71c-build.txt)). This is the actual error message:
> 
> ```
> [HOOK - conan-center.py] post_package(): ERROR: [PACKAGE LICENSE (KB-H012)] No 'licenses' folder found in package: /home/conan/w/prod/BuildSingleReference/.conan/data/libqasm/0.5.6/_/_/package/a25d6c83542b56b72fdaa05a85db5d46f5f0f71c  (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H012-PACKAGE-LICENSE)
> ```
> 
> I have a couple of questions:
> 
> 1. Is this a CMake/installation issue? I have a `LICENSE.md` file in my project. Maybe I only need to copy it to the installation folder. Will
> 2. Is there a way to run these hooks locally? I have `conan 2.0.4` installed in my machine. I remember to have run hooks locally a few months ago with `conan 1.x`, but I haven't seen the way to run them with `conan 2`. I've just been able to run some `lints`.
> 
> Thanks!

I've already fixed the first point by copying the license file at the `package` method:

```
def package(self):
    copy(self, "LICENSE.md", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
```

recipes/libqasm/all/conanfile.py
SpaceIm: > But I cannot do that. I need to set my src_folder to ..

Why?

rturrado:
TODO: check why apple-clang 13 fails (see logs [here](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/17719/15-macos-m1-clang/libqasm/0.5.6//summary.json)).

I've changed the recipe now to support only from apple-clang 14 onwards, but that leaves us with zero macos/clang binaries built.

RubenRBS:
Ping me if you need extra help figuring out the apple-clang/msvc issues :)

rturrado:
> > Ping me if you need extra help figuring out the apple-clang/msvc issues :)
> 
> Hi @RubenRBS, many thanks for your help!
> 
> I see `macos/clang13/x86_64` builds are OK, but `macos/clang13/M1-arm64` builds fail. If you have any idea where the error may be coming from, it would be great to have those builds running. Otherwise, I would just be happy not supporting `M1` builds at the moment.
> 
> To summarise the error:
> 
> * A couple of targets are generated, `tree-gen` and `func-gen`.
> * These targets are then used in a parent `CMakeLists.txt`.
> * All builds except for `M1` find those targets correctly. But `M1` builds don't:
> 
> ```
> [ 17%] Built target tree-gen
> ...
> /bin/sh: tree-gen: command not found
> ```

I think my last commit fixes this error. However, a new error is stopping the `macos/clang/M1-arm64` build:

```
/bin/sh: ../_deps/tree-gen-build/tree-gen: Bad CPU type in executable
```

After googling a bit, this may indicate an `x86_64` binary is trying to be run on an `M1-arm64` architecture. Could this be possible if we have a look at the profiles of the failed builds? For example, from the [Release](https://c3i.jfrog.io/c3i/misc/logs/pr/17719/20-macos-m1-clang/libqasm/0.5.7//2b2201f54b77f855edf1189196431cb2e3f87839-build.txt) build:

```
Default settings
	os=Macos
	os_build=Macos
	arch=x86_64
	arch_build=x86_64
	compiler=apple-clang
	compiler.version=13
	compiler.libcxx=libc++
	build_type=Release
*** You can change them in /Users/jenkins/w/prod/BuildSingleReference@3/.conan/profiles/default ***
*** Or override with -s compiler='other' -s ...s***


Configuration (profile_host):
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.libcxx=libc++
compiler.version=13.0
os=Macos
[options]
libqasm:shared=True
[build_requires]
[env]
[conf]
```

rturrado:
> > > Ping me if you need extra help figuring out the apple-clang/msvc issues :)
> > 
> > 
> > Hi @RubenRBS, many thanks for your help!
> > I see `macos/clang13/x86_64` builds are OK, but `macos/clang13/M1-arm64` builds fail. If you have any idea where the error may be coming from, it would be great to have those builds running. Otherwise, I would just be happy not supporting `M1` builds at the moment.
> > To summarise the error:
> > 
> > * A couple of targets are generated, `tree-gen` and `func-gen`.
> > * These targets are then used in a parent `CMakeLists.txt`.
> > * All builds except for `M1` find those targets correctly. But `M1` builds don't:
> > 
> > ```
> > [ 17%] Built target tree-gen
> > ...
> > /bin/sh: tree-gen: command not found
> > ```
> 
> I think my last commit fixes this error. However, a new error is stopping the `macos/clang/M1-arm64` build:
> 
> ```
> /bin/sh: ../_deps/tree-gen-build/tree-gen: Bad CPU type in executable
> ```
> 
> After googling a bit, this may indicate an `x86_64` binary is trying to be run on an `M1-arm64` architecture. Could this be possible if we have a look at the profiles of the failed builds? For example, from the [Release](https://c3i.jfrog.io/c3i/misc/logs/pr/17719/20-macos-m1-clang/libqasm/0.5.7//2b2201f54b77f855edf1189196431cb2e3f87839-build.txt) build:
> 
> ```
> Default settings
> 	os=Macos
> 	os_build=Macos
> 	arch=x86_64
> 	arch_build=x86_64
> 	compiler=apple-clang
> 	compiler.version=13
> 	compiler.libcxx=libc++
> 	build_type=Release
> *** You can change them in /Users/jenkins/w/prod/BuildSingleReference@3/.conan/profiles/default ***
> *** Or override with -s compiler='other' -s ...s***
> 
> 
> Configuration (profile_host):
> [settings]
> arch=armv8
> build_type=Release
> compiler=apple-clang
> compiler.libcxx=libc++
> compiler.version=13.0
> os=Macos
> [options]
> libqasm:shared=True
> [build_requires]
> [env]
> [conf]
> ```

I've got some more insight into this error from a CMake discussion group (see [here](https://discourse.cmake.org/t/executable-not-found-for-macos-clang-m1-arm64-but-yes-for-macos-clang-x86-64/8235/6?u=rturrado)). I'm quoting Ben Boeckel's answer in that thread:

> Rosetta 2 is likely getting in the way here. The insight is that when a process is running in x86_64 mode under Rosetta, any process it launches will also see x86_64 only and arm64-only binaries will refuse to launch. It would seem that /bin/sh (or something that launches it) is x86_64-only and that is why it fails to launch.

May it be that the Conan Center runners don't use Rosetta 2 so cannot run `x86_64` binaries on `M1-arm64` systems? @RubenRBS  @SpaceIm

recipes/libqasm/all/conanfile.py
SSE4: @RubenRBS  I think having C++23 in 2023 should be expected

SSE4:
> Hi! Thanks a lot for your contribution. I have some minor changes needed before merging, but overall looks good, thanks a lot!

@RubenRBS can you unblock? it's blocked for 3 weeks, and review comments seems to be addressed already.

SSE4 (approved):

SSE4:
@RubenRBS ping :)

recipes/libqasm/all/conanfile.py
rturrado: Because my main `CMakeLists.txt` is at `.` level. Then I have another `CMakeLists.txt` at `src` level.

rturrado:
Update:
- I'm now on holidays, but retake work on this PR next week.
- We are also working on a `libqasm` branch without `tree-gen`, `flex/bison/m4` dependencies. That should avoid the issue we are seeing when building `macos/clang13/M1-arm64`.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+        self.folders.source = "."
+        self.folders.build = os.path.join("build", str(self.settings.build_type))
+        self.folders.generators = os.path.join(self.folders.build, "generators")
+
+        self.cpp.package.libs = ["cqasm"]
+        self.cpp.package.includedirs = ["include"]
+        self.cpp.package.libdirs = ["lib"]
+
+        self.cpp.source.includedirs = ["include"]
+        self.cpp.build.libdirs = ["."]
```
valgur, line 80:
```suggestion
        cmake_layout(self, src_folder="src")
```
The recipe works fine with the standard layout. A more detailed layout is only useful for in-source recipes for editable mode support, imo. A more detailed layout does not really add anything here.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.variables["LIBQASM_BUILD_PYTHON"] = self.options.build_python
+        tc.variables["LIBQASM_BUILD_TESTS"] = self.options.build_tests
+        tc.variables["LIBQASM_COMPAT"] = self.options.compat
+        tc.variables["TREE_GEN_BUILD_TESTS"] = self.options.tree_gen_build_tests
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
```
valgur, line 96:
```suggestion
        # Fix broken hash for https://github.com/QuTech-Delft/tree-gen/tree/9ee414a7ade4b0ff293afeb095e6da0b2f44bb7a
        replace_in_file(self, os.path.join(self.source_folder, "src", "CMakeLists.txt"),
                        "c3d2d1a041a5826d3e377e4722774cbb8264d7c0",
                        "d708ea629bad014011d3f1efc352077414dd86f7")
        cmake = CMake(self)
```
The `tree-gen` git hash in the project is broken. Replaced it with the closest one in time to this version.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+
+        self.cpp.source.includedirs = ["include"]
+        self.cpp.build.libdirs = ["."]
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
```
valgur, line 87:
```suggestion
        venv = VirtualBuildEnv(self)
        venv.generate()

```
Needed for the build_requirements.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+            "clang": "13",
+            "gcc": "10",
+            "msvc": "192",
+            "Visual Studio": "16"
+        }
+
+    def build_requirements(self):
+        self.tool_requires("m4/1.4.19")
+        if self.settings.os == "Windows":
+            self.tool_requires("winflexbison/2.5.24")
```
valgur, line 57:
```suggestion
            self.tool_requires("winflexbison/2.5.25")
```

valgur (changes requested):
The recipe is very close to complete. Just a few minor tweaks needed.

rturrado:
> The recipe is very close to complete. Just a few minor tweaks needed.

Hi @valgur! Many thanks for your comments and your review. I expect to resume work on `libqasm`'s recipe during April.

recipes/libqasm/all/conanfile.py
rturrado: Cool, thanks!

recipes/libqasm/all/conanfile.py
rturrado: Thanks!

recipes/libqasm/all/conanfile.py
rturrado: We have now a `tree-gen/1.06` Conan package. We should use that instead.

recipes/libqasm/all/conanfile.py
rturrado: Great, thanks!

rturrado: Done!

recipes/libqasm/all/conanfile.py
rturrado: Don't I need the `CMakeDeps` lines? I have them in my `tree-gen` recipe.

recipes/libqasm/all/conanfile.py
rturrado: We don't need this dependency anymore.

recipes/libqasm/all/conanfile.py
rturrado: Using `cmake_layout` now.

recipes/libqasm/all/conanfile.py
rturrado: Using `self._should_build_test` now.

recipes/libqasm/all/conanfile.py
rturrado: I've left this set to `20` in the end. Does Conan Center already support `23`?

rturrado:
Hi people!

I'm resuming the work on this recipe.

`libqasm` now requires (and build_requires) the `tree-gen` Conan package so that should fix the cross-compilation issues we were having some time ago.

We are already using `libqasm` Python package (e.g., in [OpenSquirrel](https://github.com/QuTech-Delft/OpenSquirrel)), and it would be great to use it as a Conan package as well (e.g., in [QX simulator](https://github.com/QuTech-Delft/qx-simulator)).

Thanks for your help!

rturrado:
Any quick fix for `PDB FILES NOT ALLOWED (KB-H017)`?

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,108 @@
...
+
+class LibqasmConan(ConanFile):
+    name = "libqasm"
+    license = "Apache-2.0"
+    homepage = "https://github.com/QuTech-Delft/libqasm"
+    url = "https://github.com/conan-io/conan-center-index"
+    description = "Library to parse cQASM files"
+    topics = ("code generation", "parser", "compiler", "quantum compilation", "quantum simulation")
+    settings = "os", "compiler", "build_type", "arch"
+    options = {
```
RubenRBS, line 22:
```suggestion
    package_type = "library"
    options = {
```

rturrado: Done!

rturrado:
> Any quick fix for `PDB FILES NOT ALLOWED (KB-H017)`?

I spoke with @uilianries, and he gave me a fix for this:

```
    def package(self):
        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
```

rturrado:
I have a fix for the failing build. It needs to pick `antlr4_shared` instead of `antlr4_static` when building with `*/*:shared=True`. Once I put that fix in a `libqasm/0.6.2` release, I'll update this recipe to point to 0.6.2. And hopefully everything's fine by then.

RubenRBS:
Hi @rturrado, don't worry about the failing CI, the PR will be picked up automatically once we leave maintenance mode, thanks a lot for your patience!

rturrado:
> Hi @rturrado, don't worry about the failing CI, the PR will be picked up automatically once we leave maintenance mode, thanks a lot for your patience!

Ah, cool, @RubenRBS, I wasn't sure of how to force a recompilation whenever the build system was back on again.

Thanks a lot to you for your work and help!

uilianries:
@rturrado Thank you for your PR! Please, take a look in my patch: https://github.com/rturrado/conan-center-index/pull/6

recipes/libqasm/all/conanfile.py
valgur: Yes, it's required for `find_package()` to work with the `self.requires()` dependencies.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,124 @@
...
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.variables["LIBQASM_BUILD_PYTHON"] = self.options.build_python
+        tc.variables["LIBQASM_BUILD_TESTS"] = self._should_build_test
+        tc.generate()
+        env = VirtualBuildEnv(self)
+        env.generate()
+
+    def build(self):
+        cmake = CMake(self)
```
valgur, line 106: I would recommend removing `-Werror` from https://github.com/QuTech-Delft/libqasm/blob/v3-0.5.7/src/CMakeLists.txt#L296 using `replace_in_file`. It's not a good idea to include it in public builds, since it's prone to fail with new warnings and checks added with new compiler versions.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,124 @@
...
+        cmake.install()
+        fix_apple_shared_install_name(self)
+        rm(self, "*.interp", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "*.pdb", os.path.join(self.package_folder), recursive=True)
+        rm(self, "*.tokens", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "cmake_install.cmake", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "Makefile", os.path.join(self.package_folder, "include"), recursive=True)
+        rmdir(self, os.path.join(os.path.join(self.package_folder, "include", "CMakeFiles")))
+        rmdir(self, os.path.join(os.path.join(self.package_folder, "include", "v3x", "CMakeFiles")))
```
valgur, line 121: Are the CMake install() commands in the project buggy? These look out of place, especially in the `include` dir.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,124 @@
...
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE.md", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        fix_apple_shared_install_name(self)
+        rm(self, "*.interp", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "*.pdb", os.path.join(self.package_folder), recursive=True)
```
valgur, line 116:
```suggestion
        rm(self, "*.pdb", self.package_folder, recursive=True)
```

valgur (approved):
LGTM. Just some minor suggestions.

recipes/libqasm/all/conanfile.py
rturrado: Hi @valgur, thanks for your comment. I will think about this, but prefer leaving them on at the moment.

recipes/libqasm/all/conanfile.py
rturrado: Yes, that's something I definitely have to have a look at, because I've seen my IDE, locally, with these CMake files and folders all around. Thanks!

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,124 @@
...
+            self.options.rm_safe("fPIC")
+        # https://github.com/QuTech-Delft/libqasm/blob/0.6.3/src/CMakeLists.txt#L234
+        self.options["antlr4-cppruntime"].shared = self.options.shared
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def build_requirements(self):
+        self.tool_requires("tree-gen/1.0.7")
+        self.tool_requires("zulu-openjdk/11.0.19")
```
valgur, line 68:
```suggestion
        self.tool_requires("zulu-openjdk/21.0.1")
```
I assume the latest JDK version is not supported? It builds fine with it, though.

recipes/libqasm/all/conanfile.py
valgur: Looked into it. The CMake install command needs some tweaking to exclude irrelevant files from the build dir.
```patch
--- src/CMakeLists.txt
+++ src/CMakeLists.txt
@@ -315,9 +315,11 @@
     LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
 )
-get_target_property(INCLUDE_DIRS cqasm-lib-obj INTERFACE_INCLUDE_DIRECTORIES)
 install(
-    DIRECTORY ${INCLUDE_DIRS}
+    DIRECTORY
+        "${CMAKE_CURRENT_SOURCE_DIR}/../include/"
+        "${CMAKE_CURRENT_BINARY_DIR}/../include/v3x"
+        "${CMAKE_CURRENT_BINARY_DIR}/v3x"
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
-    PATTERN "*.hpp" PATTERN "*.inc"
+    FILES_MATCHING PATTERN "*.hpp" PATTERN "*.h" PATTERN "*.inc"
 )
```
I would also recommend moving the generated headers from `${CMAKE_CURRENT_BINARY_DIR}/v3x` to `${CMAKE_CURRENT_BINARY_DIR}/include/v3x`, if possible, since it's still creating empty `CMakeFiles` in the installed `include/` dir.

valgur (changes requested):
Please apply the patch for the CMake install() command either here or upstream.

recipes/libqasm/all/conanfile.py
rturrado: Many thanks for the review, @valgur! I think I will  put all of these changes into a libqasm/0.6.4.

recipes/libqasm/all/conanfile.py
rturrado: Hm... not sure now why I'm sticking to 11.0.19. I'll give 21.0.1 a go in libqasm/0.6.4. Thanks!

rturrado:
> Please apply the patch for the CMake install() command either here or upstream.

I think I'll apply it in libqasm itself, yes. Many thanks!!

recipes/libqasm/all/conanfile.py
rturrado: I've changed my mind. I'm using `replace_in_file` to remove it.

recipes/libqasm/all/conanfile.py
rturrado: Done!

recipes/libqasm/all/conanfile.py
rturrado: I have the changes ready to be tested next week in libqasm. I have updated the script that is invoking ANTLR to generate the lexer and parser files so that it moves the `.h` files to `include/v3x`. With that change, we wouldn't need that `${CMAKE_CURRENT_BINARY_DIR}/v3x` line anymore.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(f"{self.ref} requires C++{self._min_cppstd},"
+                                            f"which your compiler does not support.")
+
+        if self.dependencies["antlr4-cppruntime"].options.shared != self.options.shared:
```
uilianries, line 83:
```suggestion
        if self.settings.arch != "wasm" and self.dependencies["antlr4-cppruntime"].options.shared != self.options.shared:
```
Ops, antlr4 is only installed when not using wasm.

uilianries (commented):
Almost go to go!

recipes/libqasm/all/conanfile.py
valgur: Hmm, you are now removing `/WX` only, but not `-Werror`. Is that intentional?

rturrado: Thanks for noticing,  @valgur! That was unwanted. I'll do it next Tuesday once I'm back to work.

recipes/libqasm/all/conanfile.py
rturrado: Noticed. I'll change this on Tuesday once I'm back at work.  Many thanks @uilianries !

rturrado:
> Please apply the patch for the CMake install() command either here or upstream.

I've created [this PR](https://github.com/QuTech-Delft/libqasm/pull/220) to address these changes. Once I have a libqasm 0.6.4 I'll use it as base for this recipe.

The PR basically:
- fixes the CMake installation as suggested in @valgur's patch, and
- moves ANTLR's generated include files from `src` to `include/v3x`.

recipes/libqasm/all/conanfile.py
rturrado: Done!

recipes/libqasm/all/conanfile.py
rturrado: Done in libqasm/0.6.4!

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,134 @@
...
+    def package(self):
+        copy(self, "LICENSE.md", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        fix_apple_shared_install_name(self)
+        rm(self, "*.interp", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "*.pdb", self.package_folder, recursive=True)
+        rm(self, "*.tokens", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "cmake_install.cmake", os.path.join(self.package_folder, "include"), recursive=True)
+        rm(self, "Makefile", os.path.join(self.package_folder, "include"), recursive=True)
```
valgur, line 129:
```suggestion
        rm(self, "*.pdb", self.package_folder, recursive=True)
```
These should no longer be necessary if the CMake install() commands are working correctly.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,134 @@
...
+        rmdir(self, os.path.join(os.path.join(self.package_folder, "include", "CMakeFiles")))
+        rmdir(self, os.path.join(os.path.join(self.package_folder, "include", "v3x", "CMakeFiles")))
```
valgur, line 131:
```suggestion
```

valgur (changes requested):
LGTM. The package() step can be cleaned up further probably, though.

recipes/libqasm/all/conanfile.py
rturrado: On it! Thanks!

recipes/libqasm/all/conanfile.py
rturrado: Trying it! Thanks!

valgur (approved):

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,128 @@
...
+        env = VirtualBuildEnv(self)
+        env.generate()
+
+    def _patch_sources(self):
+        if is_msvc(self):
+            replace_in_file(self, os.path.join(self.source_folder, "src", "CMakeLists.txt"), "/WX", "")
+            replace_in_file(self, os.path.join(self.source_folder, "test", "CMakeLists.txt"), "/WX", "")
+        else:
+            replace_in_file(self, os.path.join(self.source_folder, "src", "CMakeLists.txt"), "-Werror", "")
+            replace_in_file(self, os.path.join(self.source_folder, "test", "CMakeLists.txt"), "-Werror", "")
```
valgur, line 112:
```suggestion
        werror = "/WX" if is_msvc(self) else "-Werror"
        replace_in_file(self, os.path.join(self.source_folder, "src", "CMakeLists.txt"), werror, "")
        replace_in_file(self, os.path.join(self.source_folder, "test", "CMakeLists.txt"), werror, "")
```
A purely cosmetic suggestion.

valgur (approved):

valgur (approved):

rturrado:
@SpaceIm @SSE4  @RubenRBS @uilianries anyone please do a second code review? Thanks!

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if self.settings.arch != "wasm" and self.dependencies["antlr4-cppruntime"].options.shared != self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} requires antlr4-cppruntime to be built with the same shared option value.")
+
+    def requirements(self):
+        self.requires("fmt/10.2.1", transitive_headers=True)
+        self.requires("range-v3/0.12.0", transitive_headers=True)
+        self.requires("tree-gen/1.0.7", transitive_headers=True, transitive_libs=True)
+        if not self.settings.arch == "wasm":
+            self.requires("antlr4-cppruntime/4.13.1", transitive_headers=True)
```
uilianries, line 91: - fmt transitive header: [include/cqasm-version.hpp](https://github.com/QuTech-Delft/libqasm/blob/dffe9a42563dc0f41ab098b47a699188d59ba48f/include/cqasm-version.hpp#L7)
- range-v3 transitive header: https://github.com/QuTech-Delft/libqasm/blob/dffe9a42563dc0f41ab098b47a699188d59ba48f/include/cqasm-result.hpp#L8
- tree-gen transitive header: https://github.com/QuTech-Delft/libqasm/blob/dffe9a42563dc0f41ab098b47a699188d59ba48f/include/cqasm-tree.hpp#L8
- antlr4 transitive header: https://github.com/QuTech-Delft/libqasm/blob/dffe9a42563dc0f41ab098b47a699188d59ba48f/include/v3x/BuildTreeGenAstVisitor.hpp#L3

uilianries (commented):
Looking good, only small details.

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,125 @@
...
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        # https://github.com/QuTech-Delft/libqasm/blob/0.6.3/src/CMakeLists.txt#L234
+        self.options["antlr4-cppruntime"].shared = self.options.shared
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def build_requirements(self):
+        self.tool_requires("tree-gen/1.0.7")
```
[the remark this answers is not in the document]
rturrado: Done!

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,125 @@
...
+        self.tool_requires("zulu-openjdk/21.0.1")
+        if self.settings.arch == "wasm":
+            self.tool_requires("emsdk/3.1.50")
+        if self._should_build_test:
+            self.test_requires("gtest/1.14.0")
```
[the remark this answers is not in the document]
rturrado: Done!

recipes/libqasm/all/conanfile.py
rturrado: Beautiful!

rturrado:
> Looking good, only small details.

Many thanks @uilianries!

uilianries (dismissed):
LGTM

valgur (approved):

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,127 @@
...
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMakeToolchain, CMakeDeps, CMake, cmake_layout
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import copy, get, rm, replace_in_file
+from conan.tools.microsoft import is_msvc
+from conan.tools.scm import Version
+
+required_conan_version = ">=1.53.0"
```
RubenRBS, line 13:
```suggestion
required_conan_version = ">=1.60.0 <2 || >=2.0.6"
```

recipes/libqasm/all/conanfile.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        # https://github.com/QuTech-Delft/libqasm/blob/0.6.3/src/CMakeLists.txt#L234
+        self.options["antlr4-cppruntime"].shared = self.options.shared
```
RubenRBS, line 62: In https://github.com/conan-io/conan-center-index/pull/17719/files#diff-522fd2c5d58b9eb8acbb577bd09633d36fce79b4f24346b1cb08478c372dc499R85 the check is only done if the `arch` is not `wasm`. Shouldn't this check also be done here?

RubenRBS (commented):
Hi! Thanks a lot, I have a few last minute change requests, but other than that it looks finally ready :)

I have also a request. The library seems aware of `wasm`. Have you locally tested it in such arch? If so, could you please provide compilation logs of the default build (So just `conan create ...`) and shared? (`conan create ... -o="*:shared=True"`)? Thanks a lot! It helps tremendously with future maintenance and traceability

recipes/libqasm/all/conanfile.py
rturrado: Thanks! Done!

recipes/libqasm/all/conanfile.py
rturrado: Good point @RubenRBS. At `def requirements()` we only require `antlr4-cppruntime` `if not self.settings.arch == "wasm"`. Adding the same check here.

rturrado:
> Hi! Thanks a lot, I have a few last minute change requests, but other than that it looks finally ready :)
> 
> I have also a request. The library seems aware of `wasm`. Have you locally tested it in such arch? If so, could you please provide compilation logs of the default build (So just `conan create ...`) and shared? (`conan create ... -o="*:shared=True"`)? Thanks a lot! It helps tremendously with future maintenance and traceability


Hi @RubenRBS. I am not sure if we have what you ask for because we only cross-compile for emscripten/WASM.

libqasm WASM binaries are already being tested/used in the future Quantum Inspire 2 website.

You can find the binaries for the latest libqasm release [here](https://github.com/QuTech-Delft/libqasm/releases).

Should you want to build them and run them:

```
$ curl -fsSL https://deno.land/install.sh | sh  # install deno
$ git clone https://github.com/QuTech-Delft/libqasm.git  # clone libqasm
$ cd libqasm
libqasm$ conan build . -pr=conan/profiles/emscripten -pr:b=conan/profiles/release -b missing  # cross-compile for emscripten
libqasm$ mv build/Release/emscripten/cqasm_emscripten.{,m}js  # rename js file to mjs (because it's an ES6 module)
libqasm$ cd emscripten
libqasm/emscripten$ deno run -A test_cqasm.ts  # test the WASM binaries
```

You also have GitHub logs for the `conan build` above [here](https://github.com/QuTech-Delft/libqasm/actions/runs/8830447787/job/24243607543).

And, for the shared build, please find attached the logs: [conan__build__pr_emscripten__pr_b_release__o_shared_True__b_missing.txt](https://github.com/conan-io/conan-center-index/files/15212709/conan__build__pr_emscripten__pr_b_release__o_shared_True__b_missing.txt)

I  hope this helps!

uilianries (approved):

rturrado:
@RubenRBS sorry for the stalking haha but I'm just so much willing this to go forward. Thanks!

RubenRBS (approved):
Thanks a lot for taking the time to create the recipe, we really appreciate it :)

rturrado:
> Thanks a lot for taking the time to create the recipe, we really appreciate it :)

You won't believe me but I've been dreaming about this moment since more than one year ago. 😄

Long live the Conan team!

PR #17740, opened 2023-05-29
libwebm: Add recipe for building WebM container library.

simoncent:
Specify library name and version:  **libwebm/1.0.0.29**

This adds a recipe for the WebM container library. I have strayed from the CMake build option names with `with_new_parser_api`, but I think this is much clearer than the name that the CMake build uses for the equivalent option, which is `ENABLE_WEBM_PARSER` (which a user inexperienced with the library would assume is the entire purpose of the library IMO).

I needed to include a patch to fix the install target, which is otherwise only available for the nested webm parser. I plan to open a PR with the Google folks in the hope that future versions won't need the patch.

---

### DIFF
--- a/recipes/libwebm/all/conandata.yml
+++ b/recipes/libwebm/all/conandata.yml
@@ -1,0 +1,8 @@
+sources:
+  "1.0.0.29":
+    url: "https://chromium.googlesource.com/webm/libwebm/+archive/2f9fc054ab9547ca06071ec68dab9d54960abb2e.tar.gz"
+patches:
+  "1.0.0.29":
+    - patch_file: "patches/1.0.0.29-cmake-install-targets.patch"
+      patch_description: "Always declare the CMake install target for libwebm.a/so"
+      patch_type: "conan"
--- a/recipes/libwebm/all/conanfile.py
+++ b/recipes/libwebm/all/conanfile.py
@@ -1,0 +1,81 @@
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
+import os
+
+required_conan_version = ">=1.53.0"
+
+
+class LibwebmConan(ConanFile):
+    name = "libwebm"
+    description = "Library for muxing and demuxing WebM media container files"
+    topics = ("conan", "libwebm", "webm", "container", "demuxing", "muxing", "media", "audio", "video")
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://chromium.googlesource.com/webm/libwebm/"
+    license = "BSD-3-Clause"
+
+    settings = "os", "arch", "compiler", "build_type"
+
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_pes_ts": [True, False],
+        "with_new_parser_api": [True, False],
+    }
+
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_pes_ts": True,
+        "with_new_parser_api": False,
+    }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version])
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["ENABLE_WEBMTS"] = self.options.with_pes_ts
+        tc.variables["ENABLE_WEBM_PARSER"] = self.options.with_new_parser_api
+        tc.variables["ENABLE_WEBMINFO"] = False
+        tc.generate()
+
+    def build(self):
+        apply_conandata_patches(self)
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+        copy(self, "LICENSE.TXT", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "WebM")
+        self.cpp_info.set_property("cmake_target_name", "WebM::webm")
+        self.cpp_info.set_property("pkg_config_name", "webm")
+        # TODO: back to global scope in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.components["libwebm"].libs = ["webm"]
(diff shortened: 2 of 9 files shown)


simoncent:
Sorry, I've opened this a bit prematurely it seems, fixing the build errors now 👍🏻

simoncent:
I have reviews open with Google for everything in the patch files for this PR, and while we're still tweaking around the edges, they agree in principle with each change, so future releases should not need the patch files. I will ask for a new tagged release once everything is merged and will update this PR or put in a new one if this is already merged (we could wait, but I would like to be able to use this sooner rather than later tbh).

recipes/libwebm/all/conandata.yml
```diff
@@ -0,0 +1,14 @@
+sources:
+  '1.0.0.29':
+    url: 'https://chromium.googlesource.com/webm/libwebm/+archive/2f9fc054ab9547ca06071ec68dab9d54960abb2e.tar.gz'
```
uilianries, line 3: could you please use the tag instead? it's available in the repo and is easier to identify.

Plus, is it possible to package 1.0.0.30 instead of 29?

recipes/libwebm/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
+import os
+
+required_conan_version = ">=1.53.0"
+
+
+class LibwebmConan(ConanFile):
+    name = "libwebm"
+    description = "Library for muxing and demuxing WebM media container files"
+    topics = ("conan", "libwebm", "webm", "container", "demuxing", "muxing", "media", "audio", "video")
```
uilianries, line 12:
```suggestion
    topics = ("webm", "container", "demuxing", "muxing", "media", "audio", "video")
```

topics are useful for conan.io/center for instance, so the project name and conan are ambiguous as it will find the project name as package name

recipes/libwebm/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version])
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["ENABLE_WEBMTS"] = self.options.with_pes_ts
+        tc.variables["ENABLE_WEBM_PARSER"] = self.options.with_new_parser_api
+        tc.variables["ENABLE_WEBMINFO"] = False
+        tc.variables["CMAKE_POSITION_INDEPENDENT_CODE"] = self.options.get_safe("fPIC", default=True)
```
uilianries, line 55:
```suggestion
```
The `CMAKE_POSITION_INDEPENDENT_CODE` is automatically filled by CMakeToolchain. https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html?highlight=cmake_position_independent_code#generated-files

Sometimes, we really need to enforce it, because the upstream customizes that flag.

recipes/libwebm/all/patches/1.0.0.29-cmake-dont-build-programs.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 8db3881..501d757 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -347,19 +347,6 @@ add_library(webm ${libwebm_common_sources}
+             $<TARGET_OBJECTS:mkvmuxer>
+             $<TARGET_OBJECTS:mkvparser>)
+ 
+-add_executable(mkvparser_sample ${mkvparser_sample_sources})
```
uilianries, line 9: It could be pushed to the upstream as a new option 😄

recipes/libwebm/all/patches/1.0.0.29-cmake-install-targets.patch
```diff
@@ -0,0 +1,139 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
uilianries, line 1: please, update the conandata.yml with the correct information. This patch is already in the master branch: https://chromium.googlesource.com/webm/libwebm/+/f7da30b247bde1bb662ec7d59687302b138f8268

It's not `conan`, but `official`. 

https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conandata_yml_format.md#patches-fields

recipes/libwebm/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+        self.cpp_info.set_property("cmake_target_name", "WebM::webm")
+        self.cpp_info.set_property("pkg_config_name", "webm")
+        # TODO: back to global scope in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.components["libwebm"].libs = ["webm"]
+
+        if self.settings.os in ["Linux", "FreeBSD", "Android"]:
+            self.cpp_info.components["libwebm"].system_libs.append("m")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.names["cmake_find_package"] = "WebM"
```
uilianries, line 80: Do you have reference about those target names? I don't see any reference in the upstream. By default, we don't customize cmake target if we don't any evidence from the upstream or official cmake modules.

uilianries (commented):
Thank you for your contribution. That's an interesting project, please, take a look in my review.

recipes/libwebm/all/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.1)
```
uilianries, line 1: You no longer need to include test_v1_package for new recipes: https://github.com/conan-io/conan-center-index/discussions/17406

recipes/libwebm/all/conandata.yml
simoncent: Yes, I was waiting for them to create the new tag and have had a long weekend.

uilianries: Thank you!

recipes/libwebm/all/patches/1.0.0.29-cmake-dont-build-programs.patch
simoncent: This was part of one of my patchsets, it has now been merged.

recipes/libwebm/all/patches/1.0.0.29-cmake-install-targets.patch
simoncent: All the patches should now be unnecessary because I've had them all merged into upstream. I will push an update to this PR soon.

recipes/libwebm/all/conanfile.py
simoncent: I'll look into fixing this 👍🏻

uilianries (approved):
LGTM

RubenRBS (approved):
Looks great to me, thanks a lot for your contribution :)

Issue #14120, opened 2022-11-09
[package] gfortran/10.2: cannot execute ‘f951’: execvp: No such file or directory


samuel-emrys:
### Description

After the changes in #12613, I'm still not able to use the binary `gfortran` package to compile any code. It fails with:

```
gfortran: fatal error: cannot execute ‘f951’: execvp: No such file or directory
```

Relates to #4696

### Package and Environment Details

* Package Name/Version: **gfortran/10.2**
* Operating System+version: **Arch Linux**
* Compiler+version: **Clang 14**
* Docker image: N/A
* Conan version: **conan 1.53.0**
* Python version: **Python 3.10.0**

### Conan profile

```
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=clang
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux
os_build=Linux
[options]
[build_requires]
[env]
CC=/usr/bin/clang
CXX=/usr/bin/clang++
[conf]
tools.build:skip_test=False
```

### Steps to reproduce

I modified the test package with the following to better use and test the executable produced by the `gfortran` package:

```diff
diff --git a/recipes/gfortran/all/conanfile.py b/recipes/gfortran/all/conanfile.py
index 209997c47..897d199be 100644
--- a/recipes/gfortran/all/conanfile.py
+++ b/recipes/gfortran/all/conanfile.py
@@ -77,3 +77,6 @@ class GFortranConan(ConanFile):
         self.cpp_info.includedirs = []
         self.env_info.PATH.append(bin_path)
         self.cpp_info.libs = ["gfortran"]
+        compiler = os.path.join(bin_path, "gfortran")
+        self.output.info(f"Exporting FC={compiler}")
+        self.buildenv_info.define("FC", compiler)

diff --git a/recipes/gfortran/all/test_package/conanfile.py b/recipes/gfortran/all/test_package/conanfile.py
index 9d3aaf213..1cfbb70a4 100644
--- a/recipes/gfortran/all/test_package/conanfile.py
+++ b/recipes/gfortran/all/test_package/conanfile.py
@@ -1,14 +1,70 @@
+import os
+import shutil
 from conan import ConanFile
-from conan.tools.build import can_run
+from conan.tools.build import cross_building
+from conan.tools.env import VirtualBuildEnv, VirtualRunEnv

 class TestPackageConan(ConanFile):
-    settings = "os", "arch", "compiler", "build_type"
[...]

PR #17767, opened 2023-05-31
gfortran: v2 compatibility

ericLemanissier:
Specify library name and version:  **gfortran/***

---


samuel-emrys:
This recipe is broken - if you need gfortran you should probably just use the gcc recipe. I actually think this recipe should be removed - i can't imagine anybody uses it in its current state. See https://github.com/conan-io/conan-center-index/issues/14120

samuel-emrys (approved):

uilianries (approved):
LGTM this package is broken for some scenarios, indeed gcc package is its better successor.

PR #17775, opened 2023-05-31
xoshiro-cpp: Conan 2.0 syntax fixes

jcar87:
Specify library name and version:  **xoshiro-cpp/all**

This PR fixes errors when using the `msvc` compiler (instead of Visual Studio) - as it was not handled in the case. Additionally for Conan 2.0, fixes legacy use of `self.output.warn` which is no longer valid syntax in Conan 2.0.


RubenRBS (approved):

jwillikers (approved):

SpaceIm:
There was https://github.com/conan-io/conan-center-index/pull/17297 opened for 1 month 😩

PR #17797, opened 2023-06-02
pixman 0.42.2

ericLemanissier:
Specify library name and version:  **pixman/0.42.2**

---

### DIFF
--- a/recipes/pixman/all/patches/0003-Adjustments-to-build-with-llvm.patch
+++ b/recipes/pixman/all/patches/0003-Adjustments-to-build-with-llvm.patch
@@ -1,0 +1,3211 @@
+diff --git a/pixman/pixman-arm-asm.h b/pixman/pixman-arm-asm.h
+index ee78541087377907073c02f15ac546e308c39a3f..bed7813489953f78999da3552f04a044aa00a882 100644
+--- a/pixman/pixman-arm-asm.h
++++ b/pixman/pixman-arm-asm.h
+@@ -25,13 +25,27 @@
+  *
+  */
+ 
++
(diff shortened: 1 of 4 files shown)


jwillikers (approved):

recipes/pixman/all/patches/0003-Adjustments-to-build-with-llvm.patch
```diff
@@ -0,0 +1,3211 @@
+diff --git a/pixman/pixman-arm-asm.h b/pixman/pixman-arm-asm.h
```
danimtb, line 1: wow, this path file is incredibly huge. I think this will be a burden if we ever have to tune it or even change some of its lines. The only positive thing is that it is well-referenced with a link to its origin in the conandata, but I still have my hesitations to add it here 😢

ericLemanissier: yeah, let's hope https://gitlab.freedesktop.org/pixman/pixman/-/merge_requests/71 will be merged as-is, so that we can drop the patch with the next version (and drop 0.42.2). The alternative is to disable macos for this version of pixman.

danimtb: yeah, I think it is too much right now. Let's wait for the upstream fix

ericLemanissier: ok. FTR, it could take a long time, seeing the release history https://gitlab.freedesktop.org/pixman/pixman/-/tags

PR #17822, opened 2023-06-04
cppdap: add recipe

gjasny:
Specify library name and version:  **cppdap/cci.20230526**

CMake 3.27 will have `cppdap` as a dependency (see [merge request](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8338))

---

### DIFF
--- a/recipes/cppdap/all/conanfile.py
+++ b/recipes/cppdap/all/conanfile.py
@@ -1,0 +1,99 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.microsoft import is_msvc
+from conan.tools.files import get, copy, rm, rmdir
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class PackageConan(ConanFile):
+    name = "cppdap"
+    description = "Debug Adapter Protocol SDK"
+    license = "Apache-2.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/google/cppdap"
+    topics = ("debug", "adapter", "protocol", "dap")
+    package_type = "library"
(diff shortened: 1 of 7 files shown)


memsharded (approved):

recipes/cppdap/all/conanfile.py
```diff
@@ -0,0 +1,99 @@
...
+        cmake.install()
+
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+
+    def package_info(self):
+        self.cpp_info.libs = ["cppdap"]
+
+        self.cpp_info.set_property("cmake_file_name", "cppdap")
+        self.cpp_info.set_property("cmake_target_name", "cppdap::cppdap")
```
memsharded, line 94: When this matches the package name, neither of these to properties are necessary, they are already the default.
No need to change it, but just you know it, for simpler recipes.

gjasny: Thanks for sharing!

recipes/cppdap/all/conanfile.py
```diff
@@ -0,0 +1,99 @@
...
+        elif self.options.with_json == "rapidjson":
+            self.requires("rapidjson/cci.20220822")        
+        elif self.options.with_json == "nlohmann_json":
+            self.requires("nlohmann_json/3.11.2")
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        if self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
```
SpaceIm, line 62: if it raises for shared, why providing shared option? And why message refers to Visual Studio and msvc specifically while it raises regardless of compiler?

SpaceIm: FYI, shared build is allowed since https://github.com/google/cppdap/pull/116

PR #17843, opened 2023-06-06
Package/microservice essentials

seboste:
Specify library name and version:  **microservice-essentials/0.1.0**

I am the author of this C++ open source library that takes care of typical recurring concerns that occur in microservice development.
I need this library to be retrievable from conan center because my open source CLI tool sapper (https://github.com/seboste/sapper) that enables you to rapidly create, extend, and update C++ microservices heavily relies on conan. 

---


recipes/microservice-essentials/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["microservice-essentials"]
+        if self.settings.os != "Windows":
+            self.cpp_info.system_libs = ["pthread"]
+        lib_path = os.path.join(self.package_folder, "lib")
+        self.output.info("Appending PATH environment variable with : {0}".format(lib_path))
+        self.runenv_info.prepend_path("PATH", lib_path)
```
czoido, line 106: Why is this needed?

seboste: Windows/Shared lib case: .dll is in the lib_path. The test_package (and any other user of the library) needs to be able to find the .dll file. Is there a better way to ensure this?

czoido: Ok, I see. This is not needed because Conan will add those paths to the `VirtualRunEnv` generator and create a `conanrun.bat/sh` file so you can activate the environment before running the program built in shared mode. Please check this section of the tutorial in the Conan docs: https://docs.conan.io/2/tutorial/consuming_packages/different_configurations.html#modifying-options-linking-the-application-dependencies-as-shared-libraries

recipes/microservice-essentials/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def package(self):
+        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["microservice-essentials"]
+        if self.settings.os != "Windows":
+            self.cpp_info.system_libs = ["pthread"]
```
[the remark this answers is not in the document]
seboste: thanks. Done.

recipes/microservice-essentials/all/conanfile.py
seboste: unfortunately, just removing isn't enough. The dll needs to be in the bin folder for this to work. I now explicitly copied the files into this folder.

recipes/microservice-essentials/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.16.3 <4]")
+
+    def requirements(self):
+        if self.options.with_examples:
+            self.requires("cpp-httplib/0.12.4")
+            self.requires("nlohmann_json/3.11.2")
+            self.requires("openssl/3.1.0")
```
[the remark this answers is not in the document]
seboste: done.

recipes/microservice-essentials/all/conanfile.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def package(self):
+        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        copy(self, "*.dll", src=self.source_folder, dst=os.path.join(self.package_folder, "bin"), keep_path=False)
```
czoido, line 99: I think you need this because probably the CMakeLists is installing the `dll's` to other place that's not `bin` (maybe also to `lib`?) which is the default, so in the end the dll's will end up in two places, the ones copied by `cmake.install()` and the ones by the `copy()`. You an set the `cpp_info.bindirs = ["lib"]` if that's the case but the expected thing would be that the dll's end up in the `bin` folder

seboste: CMake installs the dlls to the lib folder. That can be changed some time in the future but not for that initial 0.1.0 version. Other recipes use CMakeLists.txt patches for the same issue. If possible, I'd like to avoid that. Thus, I updated the .bindirs property. Maybe not perfect, but better than deploying two copies of the dlls.

czoido (dismissed):

recipes/microservice-essentials/all/conanfile.py
```diff
@@ -0,0 +1,104 @@
...
+        "with_examples": False
+    }
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "7",
+            "clang": "5",
+            "apple-clang": "10",
+            "Visual Studio": "15.7",
```
RubenRBS, line 38: Missing `msvc` compiler version check :)

```suggestion
            "Visual Studio": "15.7",
            "msvc": "191",
```

RubenRBS (changes requested):

czoido (approved):

RubenRBS (approved):

PR #17858, opened 2023-06-07
[cgal] compatibility with Conan 2.0, and cgal/5.5.2

### LINKED ISSUE #10011: [package] CGAL/5.3: recipe does not make consumers aware of -frounding-mode flag with CMakeToolchain/CMakeDeps generators & gcc/g++
# Context
[cgal_test.zip](https://github.com/conan-io/conan-center-index/files/8372793/cgal_test.zip)

CGAL makes use of interval arithmetic in many places; some of these uses are in header files,
which means that they may require the client/consumer to compile using certain flags on some compilers
to make the compiler aware that the code makes use of rounding modes different from the default (round-to-nearest, ties-to-even).

Failure to do so may or may not silently make the compiler output incorrect code, depending on the use case and whether the compiler finds opportunities to apply "optimizations" that break the code because they are based on the assumption of a specific rounding mode.

The relevant flags are /fp:strict for MSVC or -frounding-math for GCC.
See, for example, the following somewhat related CGAL issue: https://github.com/CGAL/cgal/issues/3180 for a discussion where the developers conclude that the flag may still be necessary.

CLang does not support such flags and GCCs support is somewhat flaky,
so CGAL developers use a macro (IA_OPACIFY) with compiler detection macros, typically involving volatile or inline asm to hide certain optimization opportunities from the compiler, preventing breakage in most cases (but not always);
it is therefore not trivial to produce a program that actually breaks without the flag.

Usually, e.g., when building and linking against CGAL without using conan, linking with the cgal targets in CMake makes sure that the flag is added to the compilation of the consumer (i.e., it is a PUBLIC compile_option or compile_definition of the cgal target) if the compiler is GCC or MSVC.
When using the CMakeDeps and CMakeToolchain generators, this flag is lost. I did not test any other generators (cmake or cmake_paths).

## Expected Behavior
Linking a consumer library/executable against the cgal::cgal target provided by the CMakeDeps/CMakeToolchain generators should add the flag `-frounding-math` to the compiler command line.

## Actual Behavior
Linking a consumer library/executable against the cgal::cgal target provided by the CMakeDeps/CMakeToolchain generators does not add the flag `-frounding-math` to the compiler command line.

### Package and Environment Details (include every applicable attribute)
  * Package Name/Version: cgal/5.3
  * Operating System+version: Mac OS, though probably every supported OS
  * Compiler+version: GCC 11.2
  * Docker image: Not a docker image, though conanio/gcc11 should also show this behaviour (not tested).
  * Conan version: 1.45.0
  * Python version: Python 3.9.12

### Conan profile (output of `conan profile show default` or `conan profile show <profile>` if custom profile is in use)
```
Configuration for profile gnu11:

[settings]
os=Macos
os_build=Macos
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=11.2
compiler.libcxx=libstdc++11
build_type=Release
[options]
[conf]
[build_requires]
[env]
CC=gcc-11
CXX=g++-11
```

### Steps to reproduce (Include if Applicable)
Unzip the attached (relatively small & simple) example, 
[cgal_test.zip](https://github.com/conan-io/conan-center-index/files/8372855/cgal_test.zip)
create & enter a build directory, run `conan install -s build_type=Release ..` in it.
Make sure that CMake finds and uses g++ and not clang.
Then run `cmake -DCMAKE_BUILD_TYPE=Release ..` and `VERBOSE=1 make` in it.
Check whether the compiler invocations are done using `-frounding-math`. It should be, but isn't (at least in my case).

### Logs (Include/Attach if Applicable)
[log.txt](https://github.com/conan-io/conan-center-index/files/8372849/log.txt)

lrineau:
Specify library name and version:  **cgal/5.5.2**

This pull-request adds cgal/5.5.2, but its major contribution is the compatibiity with Conan 2.0.

Fixes #10011.

---


uilianries:
@lrineau I started an internal CI job to build `mpfr/4.1.0`. Thank you for reporting.

lrineau:
@uilianries Houra! The last CI set of tests eventually succeeded! Thanks for your help.

recipes/cgal/all/conanfile.py
```diff
@@ -14,63 +17,162 @@ class CgalConan(ConanFile):
...
-        else:
-            files.replace_in_file(self, os.path.join(self._source_subfolder, "CMakeLists.txt"),
-                                "if(NOT PROJECT_NAME)", "if(TRUE)")
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.75.0")
-        self.requires("eigen/3.3.9")
+        self.requires("boost/1.78.0")
```
uilianries, line 46: Is it possible to use Boost 1.82.0 instead? 
```suggestion
        self.requires("boost/1.82.0")
```

recipes/cgal/all/conanfile.py
```diff
@@ -14,63 +17,162 @@ class CgalConan(ConanFile):
...
+        cmake = CMake(self)
         cmake.install()
-        files.rmdir(self, os.path.join(self.package_folder, "share"))
-        files.rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
-        files.rmdir(self, os.path.join(self.package_folder, "bin"))
+        copy(self, "LICENSE*", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rmdir(self, os.path.join(self.package_folder, "bin"))
+        rm(self, "*Config*.cmake", os.path.join(self.package_folder, "lib", "cmake", "CGAL"))
+        rm(self, "Find*.cmake", os.path.join(self.package_folder, "lib", "cmake", "CGAL"))
```
uilianries, line 88:
```suggestion
        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
```

You could just remove the cmake folder instead.

recipes/cgal/all/conanfile.py
```diff
@@ -14,63 +17,162 @@ class CgalConan(ConanFile):
...
+        self._create_cmake_module_variables(
+            os.path.join(self.package_folder, self._cmake_module_file_rel_path)
+        )
+
+    def _create_cmake_module_variables(self, module_file):
```
uilianries, line 93: Why do you need to create it only now? What error are you trying to fix?

recipes/cgal/all/conanfile.py
lrineau: No I cannot. The directory `lib/cmake/CGAL` in the CGAL installation still contains a lot of files, and some of them (most of them) are very important for the CMake API of CGAL. For example, the `CMakeLists.txt` of the test package uses the interface CMake target `CGAL::Eigen3_support` provided in the file `CGAL_Eigen3_support.cmake`. Those are documented, see for example [the section **4.2 Eigen** in the documentation](https://doc.cgal.org/5.5.2/Manual/thirdparty.html#title8).

I have removed the `FindXXX.cmake` module, and `CGALConfig.cmake`, to respect the Conan policy.

recipes/cgal/all/conanfile.py
lrineau: I am fixing a lot of issues, at least for CMake consumers. See #10011 for example.

CGAL requires C++14, and specific compilers flags to enable the possibility to set FPU rounding modes. That CMake function, from CGAL pull-request https://github.com/CGAL/cgal/pull/7512, takes care of all the known compilers CGAL has ever supported.

Everything is probably doable adding compilers flags using `cpp_info`, except for `target_compile_features(${target} INTERFACE cxx_decltype_auto)` for which Conan has currently no equivalent.

recipes/cgal/all/test_package/conanfile.py
```diff
@@ -1,20 +1,26 @@
...
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.build import can_run
 
 class TestPackageConan(ConanFile):
     settings = "os", "compiler", "build_type", "arch"
-    generators = "cmake"
+    generators = "CMakeDeps", "CMakeToolchain"
+
+    def requirements(self):
+        self.requires("eigen/3.4.0")
```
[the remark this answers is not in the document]
lrineau: Some parts of CGAL require Eigen3, but not CGAL core parts.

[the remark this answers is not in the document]
uilianries: Is it something new? We didn't have it before and cgal was working. If it's a new dependency, it would require a `if Version(self.version) < "5.5.2"` for instance.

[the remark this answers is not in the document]
uilianries: It should be solved as cgal dependency directly then, and could be added as dependency based on an option.

recipes/cgal/all/conanfile.py
uilianries: I would be great adding this detailed in description as a comment in the code. Usually, we receive this kind of change as a patch, over the current version, but I see applying it to each past version would require another patch too. Another option is you adding it as a cmake file side-by-side with the conanfile.py, and exporting it.

recipes/cgal/all/test_package/conanfile.py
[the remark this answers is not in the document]
lrineau: CGAL is a header-only library, with dependencies and optional dependencies. That is not new. What is new is that I have upgraded the `test_package` with a less trivial test.

During the build of the CGAL Conan package, the dependencies will not change anything in the produced package (that is header-only). I do not think Conan package options are a good way to deal with that. There are only three dependencies: boost, gmp, and mpfr. But currently **18** optional dependencies. That is not new.

If a user has a program with several compilation units using CGAL (like the CGAL 3D demo, with all its plugins) but only some of them using optional dependencies, then the use of Conan option would pull the dependencies for all the compilation units, and not only for the one really needing them.

uilianries (approved):
LGTM

lrineau:
@uilianries, @prince-chrismc: is there anything more I should do with this pull-request? Or just wait?

I am sorry for my questions: I am not completely familiar with the Conan community and its practices.

SSE4 (approved):

PR #17882, opened 2023-06-09
b2: Support CXX from a VirtualBuildEnv

PeteAudinate:
Specify library name and version:  **b2/4.9.6**

Parts of the b2 recipe read environment variables with `os.environ.get()`. This no longer works if the environment is provided from a VirtualBuildEnv.

Also, the `b2 install` step requires the correct compiler to be present, and will fail if the compiler isn't on the PATH. This is likely to be the case if the CXX environment variable is being used to specify the compiler binary. The best way to indicate the toolchain path to b2 seems to be to create a `project-config.jam` file with the path to the compiler (from the CXX flag).

---


PeteAudinate:
To reproduce:
- Remove g++ from your `PATH`
- Create a test_profile containing:
    ```
    [buildenv]
    CXX=/path/to/g++
    ```
- (for Conan v1) Run `conan create . b2/4.9.3@ --profile test_profile --options b2:toolset=gcc --options b2:use_cxx_env=True`

Without the first commit in this PR, `build.sh` fails to find g++.
Without the second commit in this PR, `b2 install` fails to find g++.

PeteAudinate:
@grafikrobot I'm not at all a b2 expert, so would appreciate your input :-). Is there a better way to do this?

PeteAudinate:
```
b2/4.3.0: Calling package()
ERROR: b2/4.3.0: Error in package() method, line 182
	copy(self, "*b2.exe", dst=self._pkg_bin_dir, src=self._b2_output_dir)
	OSError: [Errno 22] Invalid argument: 'C:\\J\\w\\prod-v2\\BuildSingleReference\\p\\t\\b21a65a25b24fc1\\b\\build\\output\\b2.exe'
```

@grafikrobot this failed once, then passed when I reran it above. I've also just seen this error with a local build (but again, it usually passes). Any idea what the cause might be? Should I raise a separate issue?

PeteAudinate:
> Conan v1 pipeline ❌
> Failure in build 3 (3674d446b433733fb313e372d6f4ef618d2e0c28):

No idea why this got kicked off (or why it failed). It passed previously, and there have been no subsequent updates.
I've merged from master again to attempt a retry.

jwillikers (approved):

recipes/b2/portable/conanfile.py
```diff
@@ -107,6 +108,12 @@ def _bootstrap_env(self):
             os.environ.clear()
             os.environ.update(saved_env)
 
+    def _write_project_config(self, cxx):
+        with open(os.path.join(self.source_folder, "project-config.jam"), "w") as f:
+            f.write(
+                f"using {self.options.toolset} : : {cxx} ;\n"
```
prince-chrismc, line 114: Is this for debug or confirm it works?

prince-chrismc (commented):
I am curious does this work with `tools.build:cxx` in the conf?

recipes/b2/portable/conanfile.py
PeteAudinate: It's writing to a project config file to tell b2 where the compiler is located. I'm a b2 novice, but this seemed like the best way to configure the project to use the compiler path from the CXX env var.

Without this, the b2 install stage uses the wrong compiler path.

PeteAudinate:
> I am curious does this work with `tools.build:cxx` in the conf?

Unfortunately not. The `use_cxx_env` option is specifically for instructing the recipe to respect the CXX/CXXFLAGS env vars. Having the ability to use the tools.build conf would be useful, but should probably be a separate feature (this PR is purely to fix a bug, and it's been waiting for a while to be merged so I'm a bit reluctant to add more functionality to it at this point :-))

PeteAudinate:
(that came out more passive-agressive than intended, I understand the demands on the team's time! 😄 )

PeteAudinate:
Raised in https://github.com/conan-io/conan-center-index/issues/18776

prince-chrismc:
That's perfectly fine! I was curious since you had a test vector it would be easy to check :)

prince-chrismc (approved):

PR #17888, opened 2023-06-10
fix(zlib): build_requirements missing

jusito:
Specify library name and version: zlib

I stumbled upon the fact that zlib doesn't define its build requirements, maybe not intentionally.
It seems that zlib supports cmake versions below 3.15.7 but it's currently the lowest version in conan center and the version I tested.

---

### DIFF
--- a/recipes/zlib/all/conanfile.py
+++ b/recipes/zlib/all/conanfile.py
@@ -78,6 +78,10 @@ class ZlibConan(ConanFile):
                                       '/* may be set to #if 1 by ./configure */',
                                       '#if defined(HAVE_STDARG_H) && (1-HAVE_STDARG_H-1 != 0)')
 
+        
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.15.7 <4]")
+
     def build(self):
         self._patch_sources()
         cmake = CMake(self)
--- a/recipes/zlib/all/test_package/conanfile.py
+++ b/recipes/zlib/all/test_package/conanfile.py
@@ -14,6 +14,9 @@ class TestPackageConan(ConanFile):
 
     def requirements(self):
         self.requires(self.tested_reference_str)
+    
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.15.7 <4]")
 
     def build(self):
         cmake = CMake(self)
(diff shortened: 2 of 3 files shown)


recipes/zlib/all/conanfile.py
```diff
@@ -78,6 +78,10 @@ def _patch_sources(self):
                                       '/* may be set to #if 1 by ./configure */',
                                       '#if defined(HAVE_STDARG_H) && (1-HAVE_STDARG_H-1 != 0)')
 
+        
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.15.7 <4]")
```
memsharded, line 83: This shouldn't be necessary, as Conan already assumes CMake >= 3.15. It will simply fail if not used. But adding this to the many hundreds of recipes in ConanCenter will probably be too much.

memsharded (commented):

However, I think this change will not be merged, as cmake>=3.15 is assumed as a working hypothesis for Conan and ConanCenter, so adding this to hundreds of recipes would be too much

recipes/zlib/all/conanfile.py
memsharded: You can see here: https://github.com/conan-io/conan-center-index/blob/master/docs/supported_platforms_and_configurations.md some details about the provisioning of the CI machines. Basically, it is the compiler + CMake + basic build tools in linux (automake, autotools, libtool, etc)

> Ok thanks a lot, I thought its a nice addition because it sounds like the idea would be to be as independent as possible from system tooling which could fail the build.

Yes, we have previously considered this, but at the end of the day, there is a minimum of setup that is necessary anyway, and adding this to many hundreds of recipes is also extra burden. Take into account that also there are many Conan users that expect that if they have CMake>=3.15 installed in their system (and in practice, like the vast majority of users will already have CMake installed), it should be ok and used for most of the recipes, instead of ConanCenter recipes bring yet another copy of CMake (there is a new tool in Conan 2.0 to avoid this, but it is not mainstream yet).

memsharded:
I am closing this PR as per the comments above. 
But thanks very much for your contribution, really appreciated anyway! 🙂

PR #17891, opened 2023-06-10
7bitconf: add new recipe

7bitcoder:
Specify library name and version: 7bitconf/1.0.0

This is my C++17 configuration provider library.
https://github.com/7bitcoder/7bitConf
I will use this library in my other projects

---

### DIFF
--- a/recipes/7bitconf/all/conanfile.py
+++ b/recipes/7bitconf/all/conanfile.py
@@ -1,0 +1,140 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeToolchain, CMakeDeps, cmake_layout
+from conan.tools.files import get, copy, rmdir
+from conan.tools.scm import Version
+import os
+
+required_conan_version = ">=1.53.0"
+
+
+class SevenBitConfConan(ConanFile):
+    name = "7bitconf"
+    homepage = "https://github.com/7bitCoder/7bitConf"
+    description = "7bitConf is a simple C++17 centralized configuration provider library."
+    topics = ("cpp17", "configuration", "provider", "configuration-files")
+    url = "https://github.com/conan-io/conan-center-index"
+    license = "MIT"
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "header_only": [True, False],
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "header_only": False,
+        "shared": False,
+        "fPIC": True,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 17
+    
+    @property
+    def _minimum_compilers_version(self):
+        return {
+            "Visual Studio": "14",
+            "msvc": "192",
+            "gcc": "8",
+            "clang": "7",
+            "apple-clang": "10",
+        }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            self.options.rm_safe("fPIC")
+
+    def configure(self):
+        if self.options.get_safe("shared") or self.options.header_only:
+            self.options.rm_safe("fPIC")
+        if self.options.header_only:
+            self.options.rm_safe("shared")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("taocpp-json/1.0.0-beta.13", transitive_headers=True)
+
+    def package_id(self):
+        if self.info.options.header_only:
+            self.info.clear()
+
+    def validate(self):
+        compiler = self.settings.compiler
+        compiler_name = str(compiler)
+
+        if compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+
+        minimum_version = self._minimum_compilers_version.get(compiler_name, False)
+        if minimum_version and Version(compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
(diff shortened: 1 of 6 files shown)


recipes/7bitconf/all/conanfile.py
```diff
@@ -0,0 +1,144 @@
...
+        if self.options.get_safe("shared") or self.options.header_only:
+            self.options.rm_safe("fPIC")
+        if self.options.header_only:
+            self.options.rm_safe("shared")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("taocpp-json/1.0.0-beta.13", transitive_headers=True, transitive_libs=True)
```
memsharded, line 60: Are you completely sure that the transitivity is required?
- The headers transitivity is more common, if your public package headers include headers from taocpp-json, then this is necessary. I guess this is the case
- The libs transitivity should be way more exceptional, but wouldn't be necessary to be specified for most regular libraries. So if it is really required, then it would be necessary to add some comment describing the reason.

recipes/7bitconf/all/conanfile.py
```diff
@@ -0,0 +1,144 @@
...
+            tc.variables["_7BIT_CONF_BUILD_EXAMPLES"] = False
+            tc.variables["_7BIT_CONF_BUILD_TESTS"] = False
+            tc.variables["_7BIT_CONF_BUILD_SINGLE_HEADER"] = False
+            tc.variables["_7BIT_CONF_INSTALL"] = True
+            tc.variables["_7BIT_CONF_LIBRARY_TYPE"] = self.getSevenBitConfLibraryType()
+            tc.generate()
+        cmake_deps = CMakeDeps(self)
+        cmake_deps.generate()
+
+    def getSevenBitConfLibraryType(self):
```
memsharded, line 94: This shouldn't be necessary. Conan is able to dynamically compute the ``package_type`` based on the options, so you can check directly either the options or the ``package_type``.

recipes/7bitconf/all/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.1)
```
memsharded, line 1: For new recipes test_v1_package can be dropped, no longer necessary.

recipes/7bitconf/all/conanfile.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            suffix = "d" if self.settings.build_type == "Debug" else ""
+            self.cpp_info.libs = ["7bitConf" + suffix]
+
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                self.cpp_info.system_libs.append("m")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.names["cmake_find_package"] = "7bitConf"
+        self.cpp_info.names["cmake_find_package_multi"] = "7bitConf"
```
memsharded, line 144: This is no longer necessary for new recipes, you can drop these lines

memsharded (commented):
Thanks very much for your contribution of your library!
It is looking very good, just a few minor comments.

recipes/7bitconf/all/test_v1_package/CMakeLists.txt
7bitcoder: test_v1_package is now removed.

recipes/7bitconf/all/conanfile.py
7bitcoder: Thank you, I have removed these lines.

recipes/7bitconf/all/conanfile.py
7bitcoder: Thank you, taocpp json library is header only and is included in public heareds so in this case I guess that transitive_libs is not necesary, I have removed it, now only transitive_headers option is set to True.

uilianries:
@7bitcoder Thank you for your contribution, please, do not forget to sign the CLA on the comment above.

recipes/7bitconf/all/conanfile.py
7bitcoder: Thank you, I didn't know about it, I have simplified this code with package_type check.

7bitcoder: There were some some problems with checking  package_type so I have reverted this code to previous version.

recipes/7bitconf/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,7 @@
+#include <SevenBit/Conf/LibraryConfig.hpp>
+#include <iostream>
+
+int main() {
+  std::cout << "7bitconf version: " << _7BIT_CONF_VERSION;
```
uilianries, line 5:
```suggestion
  std::cout << "7bitconf version: " << _7BIT_CONF_VERSION;
  IConfiguration::Ptr configuration = ConfigurationBuilder{} //
                                            .addAppSettings()
                                            .addEnvironmentVariables()
                                            .addCommandLine(argc, argv)
                                            .build();
```

Try to consume something that requires the library to validate the linkage too.

recipes/7bitconf/all/conanfile.py
uilianries: @memsharded the package type here is not the recipe attribute, but a custom project's cmake definition. Indeed it needs to declare the package type.

memsharded: Understood, thanks for clarifying!

memsharded (approved):

uilianries (approved):

PR #17892, opened 2023-06-10
Openblas - Conan 2.0

fcorso2016:
Upgraded openblas to work with conan 2.0

---


recipes/openblas/all/CMakeLists.txt
valgur, line 1: Can be removed.

recipes/openblas/all/conanfile.py
```diff
@@ -28,7 +32,6 @@ class OpenblasConan(ConanFile):
         "use_thread": True,
         "dynamic_arch": False,
     }
-    generators = "cmake"
     short_paths = True
 
     @property
```
valgur, line 37: `self._source_subfolder` and `self._build_subfolder` can be simply replaced with `self.source_folder` and `self.build_folder` or even omitted entirely in most cases with Conan 2.

valgur (changes requested):
LGTM as an intermediate step before merging the more comprehensive #17173.
Only the `_subfolder` properties and the `CMakeLists.txt` that is no longer needed should be dropped.

fcorso2016:
It seems like there's some kind of issue going on with the build server in regard to the older versions of this package when it tries to build for windows. I'm not sure what exactly the issue is and it's not happening for me locally. Does anyone know what might be causing the issue?

fcorso2016:
Okay so it seems to be stemming from some generated .c files adding a `#define RC` at the top of the file which then conflicts with one of the headers in the Windows Developer SDK which uses that as the name of a struct member. I'm trying to figure out what's different between the older and newer versions to see how I can best resolve this issue. Unless there's a way to manually specify to build the older versions with a compatible SDK

valgur:
Adding [fix_apple_shared_install_name(](https://docs.conan.io/2/reference/tools/apple/other.html) to package() should most likely fix the last warnings.

valgur (approved):

recipes/openblas/all/conanfile.py
```diff
@@ -28,66 +33,70 @@ class OpenblasConan(ConanFile):
...
         )
 
+        if Version(self.version) <= "0.3.15":
+            replace_in_file(self, os.path.join(self.source_folder, "cmake", "utils.cmake"),
+                            "set(obj_defines ${defines_in})", "set(obj_defines ${defines_in})\r\n\r\n" +
+                            "list(FIND obj_defines \"RC\" def_idx)\r\n" + "if (${def_idx} GREATER -1) \r\n\t" +
+                            "list (REMOVE_ITEM obj_defines \"RC\")\r\n\t" + "list(APPEND obj_defines  \"RC=RC\")\r\n" +
+                            "endif ()\r\n" + "list(FIND obj_defines \"CR\" def_idx)\r\n" +
+                            "if (${def_idx} GREATER -1) \r\n\t" + "list (REMOVE_ITEM obj_defines \"CR\")\r\n\t" +
+                            "list(APPEND obj_defines  \"CR=CR\")\r\n" + "endif ()")
```
franramirez688, line 64: If this is a patch, I think it's worth putting it as a Conan one. Please, have a look at our [Policy about patching](https://github.com/conan-io/conan-center-index/blob/56c1638e25453cc8e2ce4a676604900ebc7b62a2/docs/adding_packages/sources_and_patches.md#policy-about-patching)

franramirez688 (commented):
Hi @fcorso2016 - Thanks a lot for this contribution! 👏 

Only one comment about that patch for the `utils.cmake` file.

czoido (approved):

franramirez688 (approved):

recipes/openblas/all/conanfile.py
franramirez688: Let's move forward with that because there are more recipes depending on this migration. Let's add this patch in another iteration 😉

recipes/openblas/all/conanfile.py
```diff
@@ -28,66 +33,70 @@ class OpenblasConan(ConanFile):
...
         if self.options.build_lapack:
             self.output.warn("Building with lapack support requires a Fortran compiler.")
-        cmake.definitions["NOFORTRAN"] = not self.options.build_lapack
-        cmake.definitions["BUILD_WITHOUT_LAPACK"] = not self.options.build_lapack
-        cmake.definitions["DYNAMIC_ARCH"] = self.options.dynamic_arch
-        cmake.definitions["USE_THREAD"] = self.options.use_thread
+        tc.variables["NOFORTRAN"] = not self.options.build_lapack
+        tc.variables["BUILD_WITHOUT_LAPACK"] = not self.options.build_lapack
+        tc.variables["DYNAMIC_ARCH"] = self.options.dynamic_arch
+        tc.variables["USE_THREAD"] = self.options.use_thread
```
czoido, line 83:
```suggestion
        tc.cache_variables["NOFORTRAN"] = not self.options.build_lapack
        tc. cache_variables["BUILD_WITHOUT_LAPACK"] = not self.options.build_lapack
        tc. cache_variables["DYNAMIC_ARCH"] = self.options.dynamic_arch
        tc. cache_variables["USE_THREAD"] = self.options.use_thread
```

recipes/openblas/all/conanfile.py
```diff
@@ -28,66 +33,70 @@ class OpenblasConan(ConanFile):
...
 
         # Required for safe concurrent calls to OpenBLAS routines
-        cmake.definitions["USE_LOCKING"] = not self.options.use_thread
+        tc.variables["USE_LOCKING"] = not self.options.use_thread
```
czoido, line 86:
```suggestion
        tc. cache_variables["USE_LOCKING"] = not self.options.use_thread
```

recipes/openblas/all/conanfile.py
```diff
@@ -28,66 +33,70 @@ class OpenblasConan(ConanFile):
...
 
-        cmake.definitions[
+        tc.variables[
```
czoido, line 88:
```suggestion
        tc. cache_variables[
```

recipes/openblas/all/conanfile.py
```diff
@@ -28,66 +33,70 @@ class OpenblasConan(ConanFile):
...
             "MSVC_STATIC_CRT"
         ] = False  # don't, may lie to consumer, /MD or /MT is managed by conan
 
         # This is a workaround to add the libm dependency on linux,
         # which is required to successfully compile on older gcc versions.
-        cmake.definitions["ANDROID"] = self.settings.os in ["Linux", "Android"]
+        tc.variables["ANDROID"] = self.settings.os in ["Linux", "Android"]
```
czoido, line 94:
```suggestion
        tc.cache_variables["ANDROID"] = self.settings.os in ["Linux", "Android"]
```

franramirez688 (approved):

PR #17902, opened 2023-06-11
[grpc] do not depend on grpc-proto or googleapis

### LINKED ISSUE #17755: [proposal] reorganize grpc, grpc-proto, googleapis and google-cloud-cpp
## Objective 

Create recipes for `google-cloud-cpp`, `grpc`, and `Protobuf` that work correctly and can be upgraded with minimal effort.

## Background

[googleapis/googleapis] contains just service definitions as proto files, and Bazel rules to generate the gRPC and Protobuf code from these proto files. It is not considered a "package" by the teams that maintain it at google, just a location where other projects can find the proto files *and* they can use to generate the packages for each language.

[google-cloud-cpp](https://github.com/googleapis/google-cloud-cpp) vendors-in the protos from [googleapis/googleapis]. It contains code designed to work with specific versions of googleapis/googleapis.  It almost never works with older versions. It may work with newer versions, but there are no guarantees.  `google-cloud-cpp` depends on `grpc`.

[gRPC](https://github.com/grpc/grpc) *appears* to (but in fact does not) vendor some protos from [googleapis/googleapis]. It also appears to vendor some protos from [grpc-proto](https://github.com/grpc/grpc-proto).

* Most of the proto files that are copied into gRPC are used only for tests.
* A few proto files are used via [upb](https://github.com/protocolbuffers/upb). upb is intended for "direct consumption", meaning it is intended to be vendored in.  The team does not plan to release packages.
* The CMake commands to [download googleapis](https://github.com/grpc/grpc/blob/acbac3deee35708b2da1722948bce42caac7ac33/CMakeLists.txt#L354-L362) never execute in Conan builds, as the directory guarding their download always exists.

Note that there is no conflict between the `googleapis` files used by `google-cloud-cpp` and the files used in the `grpc`. Either the files are used in tests, or they are compiled with different "protobuf" libraries, designed to produce different symbols.

Meanwhile, it is impossible to make `grpc` or `grpc-proto` use a range of `googleapis` versions. Ranges for the googleapis versioning scheme (`cci.${date}`) do no work on CCI with Conan v1 (see #17598).

## Problem

Un-vendoring `googleapis` and `grpc-proto` from the `grpc` and `google-cloud-cpp` recipes has made these recipes hard to maintain, and practically impossible to upgrade.

## Proposal

Vendor-in `googleapis` into `google-cloud-cpp` and break the dependency between `grpc` and `grpc-proto`. This would appear to "vendor-in" some protos into `grpc`, but for the reasons mentioned above, this is not the case. The protos are unused in the recipe, or compiled with `upb`.

## See Also

* https://github.com/conan-io/conan-center-index/pull/17598#issuecomment-1568463325
* https://github.com/conan-io/conan-center-index/pull/16438#issuecomment-1520207786

[googleapis/googleapis]: https://github.com/googleapis/googleapis

coryan:
Specify library name and version:  **grpc/\***

Part of the work to fix #17755.  "unvendoring" `googleapis` and `grpc-proto` does not work and it is not really needed.  The protos copied into `grpc` are only used in tests (which are not compiled by Conan) or are pre-generated by the gRPC team using `upb`, which never conflicts with the code generated with `protoc`.

---


RubenRBS (approved):
As always, thanks a lot for contributing the changes to these recipes ensuring they are valuable.

Issue #17909, opened 2023-06-12
[package] linux-headers-generic/*: avoid non-existent lib and bin dirs for clients


nbukuli:
### Description

``linux-headers-generic`` is a header only library but it leavs libdirs and bindirs as defaults. That results in reporting non-existent directories for client libraries. Build system of some of the clients (such as qt/5.12.6) do not tolerate non-existent libdirs.

The dependency chain was: ``qt/5.12.6`` -> ``libdrm/2.4.114`` -> ``linux-headers-generic/5.14.9``

### Package and Environment Details

* Package Name/Version: **linux-headers-generic/5.14.9**
* Operating System+version: **Gentoo 2.8**
* Compiler+version: **GCC 7.5.0**
* Docker image: **N/A**
* Conan version: **conan 1.60.0**
* Python version: **Python 3.9.13**

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=7
compiler.libcxx=libstdc++11
build_type=Release
[options]
[build_requires]
[env]

### Steps to reproduce

conan create qt/5.12.6@

### Logs

<details><summary>Click to expand log</summary>

Excerpt from config.log of qt/5.12.6

```
looking for library drm
Trying source 0 (type pkgConfig) of library drm ...
+ /usr/bin/pkg-config --exists --silence-errors libdrm
+ /usr/bin/pkg-config --modversion libdrm
> 2.4.114
+ /usr/bin/pkg-config --libs-only-L libdrm
> -L/home/nbukuli/.conan/data/libdrm/2.4.114/_/_/package/8d7c74e595e1506231c11a8cedd00e74f333afee/lib -L/home/nbukuli/.conan/data/linux-headers-generic/5.14.9/_/_/package/c0519e2d9702ec12d057bb15adb7a02baaf18107/lib 
+ /usr/bin/pkg-config --libs-only-l libdrm
> -ldrm 
Library path /home/nbukuli/.conan/data/linux-headers-generic/5.14.9/_/_/package/c0519e2d9702ec12d057bb15adb7a02baaf18107/lib is invalid.
  => source produced no result.
```

</details>

PR #17911, opened 2023-06-12
(#17909) linux-headers-generic: bindirs and libdirs emptied

nbukuli:
linux-headers-generic/5.14.9

According to the documentation (https://docs.conan.io/2/tutorial/creating_packages/other_types_of_packages/header_only_packages.html), for header-only libraries they should be empty. Otherwise non-existent directories are told to the client libraries, and some of the clients' build systems do not tolerate that.

qt/5.12.6 depends on libdrm which in turn depends on linux-headers-generic. qt's own configuration system throws errors if non-existent include and link directories are given as compilation arguments.
---


ericLemanissier:
Already merged in https://github.com/conan-io/conan-center-index/pull/18129. this PR can be closed

PR #17931, opened 2023-06-14
gtsam: Add options new options

ffabiomar:
Add options to set use_system_eigen and with_march_native

Specify library name and version:  **gtsam**

---

- [ x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

### DIFF
--- a/recipes/gtsam/all/conanfile.py
+++ b/recipes/gtsam/all/conanfile.py
@@ -30,6 +30,7 @@ class GtsamConan(ConanFile):
         "with_TBB": [True, False],
         "with_eigen_MKL": [True, False],
         "with_eigen_MKL_OPENMP": [True, False],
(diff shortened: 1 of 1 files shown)


recipes/gtsam/all/conanfile.py
```diff
@@ -67,6 +70,7 @@ class GtsamConan(ConanFile):
         "install_matlab_toolbox": False,
         "install_cython_toolbox": False,
         "install_cppunitlite": True,
+        "use_system_eigen": True,
```
valgur, line 73: Should probably make the `self.requires("eigen/3.4.0")` conditional on this being set to `True` then as well.

ffabiomar:
Cancelled

PR #17933, opened 2023-06-14
gtsam: add v4.2.0a9, update package options

valgur:
* Adds [v4.2.0a9](https://github.com/borglab/gtsam/releases/tag/4.2a9).
* Adds `transitive_headers=True` and `transitive_libs=True` where necessary.
* Went thoroughly over all build options used in the build scripts and updated the package options based on these.
  - Should be mostly pretty straightforward. The only exception is the deprecation of `allow_deprecated_since_V4` in favor of `allow_deprecated`. This param gets replaced with an incremented new param in each version of the library while the meaning stays the same, so it made more sense to me to simplify it to one stable option name.
* Added a missing `program_options` Boost component requirement.
* Dropped v4.0.2 to speed up the full CI build a bit. It's superseded by v4.0.3.

---


recipes/gtsam/all/conanfile.py
```diff
@@ -75,29 +85,59 @@ def export_sources(self):
...
+        self.requires("eigen/3.4.0", transitive_headers=True)
         if self.options.with_TBB:
-            self.requires("onetbb/2020.3")
+            self.requires("onetbb/2020.3", transitive_headers=True, transitive_libs=True)
+        # TODO: port metis recipe to Conan v2
+        # if Version(self.version) >= "4.1" and self.options.support_nested_dissection:
+        #     # Used in a public header here:
+        #     # https://github.com/borglab/gtsam/blob/4.2a9/gtsam_unstable/partition/FindSeparator-inl.h#L23-L27
+        #     self.requires("metis/5.1.1", transitive_headers=True, transitive_libs=True)
+        # TODO: add gperftools to ConanCenter
```
valgur, line 117: https://github.com/conan-io/conan-center-index/pull/17950

jwillikers (approved):

recipes/gtsam/all/conandata.yml
```diff
@@ -1,14 +1,28 @@
...
+    - patch_file: "patches/4.2.0-0001-macos-rpath.patch"
+      patch_description: "Relocatable shared libs on Apple OS"
+      patch_type: "conan"
+    - patch_file: "patches/4.2.0-0002-eigen3-version.patch"
+      patch_description: "Get Eigen3 version info from Conan"
+      patch_type: "conan"
+    - patch_file: "patches/4.2.0a9-0001-fix-invalid-include.patch"
+      patch_description: "Fix an invalid include for pre-C++17 compatibility"
+      patch_type: "portability"
+      patch_source: "https://github.com/borglab/gtsam/pull/1545"
```
RubenRBS, line 25: Great catch! Thanks for porting it upstream :)

recipes/gtsam/all/conanfile.py
```diff
@@ -75,43 +120,91 @@ def export_sources(self):
...
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
         if self.options.with_TBB:
             self.options["onetbb"].tbbmalloc = True
+        if not self.options.support_nested_dissection:
+            self.options.rm_safe("use_vendored_metis")
+        elif not self.options.use_vendored_metis:
+            # GTSAM expects 32-bit integers for Metis
+            self.options["metis"].with_64bit_types = False
```
RubenRBS, line 140: Could you post logs of this configuration building correctly please :)? CCI's CI won't get to try these combinations, and it would be great to have some proof for future maintenance which shows that some of the new options were tested properly

recipes/gtsam/all/conanfile.py
```diff
@@ -75,43 +120,91 @@ def export_sources(self):
...
+            self.requires("gperftools/2.10.0")
+        if self.options.support_nested_dissection and not self.options.use_vendored_metis:
+            # Used in a public header here:
+            # https://github.com/borglab/gtsam/blob/4.2a9/gtsam_unstable/partition/FindSeparator-inl.h#L23-L27
+            self.requires("metis/5.1.1", transitive_headers=True)
 
     @property
     def _required_boost_components(self):
-        return ["serialization", "system", "filesystem", "thread", "date_time", "regex", "timer", "chrono"]
+        # Based on https://github.com/borglab/gtsam/blob/4.2a9/cmake/HandleBoost.cmake#L26
```
RubenRBS, line 159: Thanks _a lot_ for these links, they make life so much easier :)

RubenRBS (commented):
Thanks a lot! Not approving yet as I still need to check the package_info method but ran out of time, will circle back to it in a few days, but so far so good :)

recipes/gtsam/all/conanfile.py
```diff
@@ -75,43 +120,91 @@ def export_sources(self):
...
+        self.requires("eigen/3.4.0", transitive_headers=True)
         if self.options.with_TBB:
-            self.requires("onetbb/2020.3")
+            self.requires("onetbb/2021.9.0", transitive_headers=True)
+        if self.options.default_allocator == "tcmalloc":
+            self.requires("gperftools/2.10.0")
+        if self.options.support_nested_dissection and not self.options.use_vendored_metis:
+            # Used in a public header here:
+            # https://github.com/borglab/gtsam/blob/4.2a9/gtsam_unstable/partition/FindSeparator-inl.h#L23-L27
+            self.requires("metis/5.1.1", transitive_headers=True)
```
valgur, line 155:
```suggestion
            self.requires("metis/5.2.1", transitive_headers=True)
```

recipes/gtsam/all/conanfile.py
valgur: Hmm, it no longer builds with the new `metis` recipe, unfortunately. I'm reverting that feature for now and will possibly add it in a future PR instead.

franramirez688: Hi @valgur! Is this happening because of the latest Metis recipe version? Could we fix that before removing that option?

valgur: I think so, yes. I intend to look into the Metis issue, but I would leave that for a separate PR, since this one is quite comprehensive already. It's not affecting any existing functionality in the GTSAM recipe anyway.

recipes/gtsam/all/conandata.yml
```diff
@@ -1,14 +1,28 @@
 sources:
+  "4.2.0a9":
```
franramirez688, line 2: @valgur I realized that the official version is `4.2a9`, should not it be better to put the same one? Am I missing something?

valgur: I based it on the full version number in CMakeLists.txt: https://github.com/borglab/gtsam/blob/4.2a9/CMakeLists.txt#L10-L13
My reasoning was that the official release is going to be `4.2.0`, so this would be slightly more consistent and would avoid potential ambiguity when sorting the version numbers.
I don't have anything against renaming it to `4.2a9` either, though.

SSE4 (approved):

PR #17953, opened 2023-06-16
logr: support conan v2, update dependencies

toge:
Specify library name and version:  **logr/***

---


recipes/logr/all/conanfile.py
```diff
@@ -34,7 +36,7 @@ def requirements(self):
             fmt_ref = "fmt/9.1.0"
             spdlog_ref = "spdlog/1.11.0"
         else:
-            fmt_ref = "fmt/8.1.1"
+            fmt_ref = "fmt/8.0.1"
```
memsharded, line 39: Why going back in ``fmt`` version?

recipes/logr/0.1.0/conanfile.py
```diff
@@ -1,97 +1,104 @@
...
-
-        self._cmake.configure( build_folder=self._build_subfolder )
-        return self._cmake
-
-    def source(self):
-        tools.get(**self.conan_data["sources"][self.version])
-        extracted_dir = self.name + "-" + self.version
-        os.rename(extracted_dir, self._source_subfolder)
+    def requirements(self):
+        self.requires("fmt/9.1.0")
```
memsharded, line 49: Older version of logr using more modern fmt than newer versions of logr?

memsharded (approved):
Looking good, thanks for your contribution!

recipes/logr/all/conanfile.py
toge: Because there is a version conflict between fmt/8.1.1 and spdlog/1.9.0.

https://github.com/conan-io/conan-center-index/blob/0580eeb5763a18b4c7034b5092ad864dabc23ba2/recipes/spdlog/all/conanfile.py#L56-L63

I updated spdlog/1.10.0 to upgrade fmt to 8.1.1.

recipes/logr/0.1.0/conanfile.py
toge: Sorry I don't know details.
Because logr/0.1.0 can be build with spdlog/1.10.0, I upgrade fmt version for logr/0.1.0.

uilianries: spdlog is not able to use the latest fmt version, so we need to keep both fmt and spdlog aligned. I would consider something more dynamic if possible like:

```python
if self.options.backend == "spdlog":
    self.requires("spdlog/1.11.0")
    self.requires("fmt/9.1.0")
else:
    self.requires("fmt/10.0.0")
```

So we can use the latest or the older. Plus, `fmt` is a good case for version range adoption.

toge: @uilianries 
Thank you for your advice!
I fixed this recipe.

memsharded (approved):

PR #17957, opened 2023-06-17
gdcm: use charls from conan instead of internal one

Nekto89:
Specify library name and version:  **gdcm/***

Recently gdcm trunk started supporting charls >= 2.2.0.
https://github.com/malaterre/GDCM/pull/157

It was enough to just change case of library name and include path. I've backported this change to 3.0.9 and 3.0.20+ versions.

---


Nekto89:
I didn't change openssl version, how did it compile before?

recipes/gdcm/all/patches/0001-3.0.20-find-charls.patch
```diff
@@ -0,0 +1,28 @@
...
+https://github.com/malaterre/GDCM/pull/157
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -352,8 +352,8 @@ else()
+ endif()
+ 
+ if(GDCM_USE_SYSTEM_CHARLS)
+-  find_package(CharLS 2.0.0 REQUIRED)
+-  set(GDCM_CHARLS_LIBRARIES ${CHARLS_LIBRARIES})
++  find_package(charls 2.2.0 REQUIRED)
```
memsharded, line 11: But required version in ``conanfile.py`` is ``charls/2.4.2``, won't this be a problem?

recipes/gdcm/all/conanfile.py
```diff
@@ -37,7 +37,7 @@ class GDCMConan(ConanFile):
 
     @property
     def _min_cppstd(self):
-        return 11
+        return 14
```
memsharded, line 40: Why is the cppstd changing? the code is the same, there is no new version, in theory this should remain the same? It is fine if ``charls`` require 14, it should be the one raising the error, and not the current ``gdcm``?

memsharded (commented):
Looking good, thanks very much for your contribution.
Just a couple of questions.

recipes/gdcm/all/conanfile.py
Nekto89: We inject CMAKE_CXX_STANDARD (C++11 before the change) if conan profile doesn't have any value for C++ standard. Without it some compilers won't be able to compile gdcm. Newer charls (>= 2.2.0) can't be compiled with C++11 - it uses for example std::underlying_type_t. So we either need to update CMAKE_CXX_STANDARD to newer value or stop compiling gdcm for some compilers that are part of conan-center-index CI.

recipes/gdcm/all/patches/0001-3.0.20-find-charls.patch
Nekto89: I hope that CMakeToolchain/CMakeDeps always look for version that conan provides. So any value here should be ok.

recipes/gdcm/all/conanfile.py
uilianries: > Newer charls (>= 2.2.0) can't be compiled with C++11 - it uses for example std::underlying_type_t.

But if charls needs it, gdcm has nothing to do about. The restriction should restricted to the project itself only. When Conan solves the dependency graph, and charls requires C++14, Conan will validate it. So the standard should be 11 here, but updated in charls.

Nekto89: @uilianries  charls already has this restriction. https://github.com/conan-io/conan-center-index/blob/37ab4bd027654c453211528d3f4d411ea2389129/recipes/charls/all/conanfile.py#L48C9-L49C39
The problem is that conan CI uses several profiles without "compiler.cppstd". And some compilers without explicit C++ standard won't be able to compile C++14 code even if they support it. Is there some example where list of minimum supported compilers depends on whether profile contains "compiler.cppstd"?

uilianries: @Nekto89 The template https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L46 is an excellent example

Nekto89: @uilianries so do you recommend forbidding users from using GDCM on compilers that do not have C++14 (or later) as their default version, unless an explicit cppstd is specified?

Nekto89: > @uilianries so do you recommend forbidding users from using GDCM on compilers that do not have C++14 (or later) as their default version, unless an explicit cppstd is specified?

I feel that it will contradict approach from test_package where we allow not having explicit cppstd. From template:
```
# In case the target project need a specific C++ standard
```

uilianries: @Nekto89 No, the restriction should be based only in what the project needs, isolated from any dependency. Reading [GDCM's CMake](https://github.com/malaterre/GDCM/blob/master/CMakeLists.txt#L13) there is no need for C++14, but C++11 only, so we should respect it and use C++11.

What charls needs, should be fixed in charls' recipe, so when Conan client runs, it will solve by the dependencies graph, like gdcm will require charls, but the minimal compiler version does not fit for charls, then it will be an invalid configuration anyway, so you don't need to manage it in gdcm recipe.

The [Charls' recipe](https://github.com/conan-io/conan-center-index/blob/master/recipes/charls/all/conanfile.py#L52) has compiler restriction, but it seems like it does not fit your current case. What's your current that does not work? So we could fix in Charls' recipe.

> I feel that it will contradict approach from test_package where we allow not having explicit cppstd.

This is because Conan 1.x will not use cppstd by default, but for Conan 2.x we use it always. The cppstd feature was introduced only after 1.0 be released, so we could not fix it without breaking users 😞

Nekto89: @uilianries 
https://github.com/conan-io/conan-center-index/pull/17957#issuecomment-1595777414
```
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=7
os=Linux
[options]
gdcm:shared=False
```
```
In file included from /home/conan/w/prod/BuildSingleReference/.conan/data/charls/2.4.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/charls/jpegls_error.h:6:0,
                 from /home/conan/w/prod/BuildSingleReference/.conan/data/charls/2.4.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/charls/charls_jpegls_decoder.h:6,
                 from /home/conan/w/prod/BuildSingleReference/.conan/data/charls/2.4.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/charls/charls.h:7,
                 from /home/conan/w/prod/BuildSingleReference/.conan/data/gdcm/3.0.9/_/_/build/033fde7c0618a26e1fee1ef3915fb443e5226d60/src/Utilities/gdcm_charls.h:21,
                 from /home/conan/w/prod/BuildSingleReference/.conan/data/gdcm/3.0.9/_/_/build/033fde7c0618a26e1fee1ef3915fb443e5226d60/src/Source/MediaStorageAndFileFormat/gdcmJPEGLSCodec.cxx:24:
/home/conan/w/prod/BuildSingleReference/.conan/data/charls/2.4.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/charls/public_types.h: In function ‘constexpr charls::encoding_options_private::encoding_options charls::encoding_options_private::operator|(charls::encoding_options_private::encoding_options, charls::encoding_options_private::encoding_options)’:
/home/conan/w/prod/BuildSingleReference/.conan/data/charls/2.4.1/_/_/package/66c5327ebdcecae0a01a863939964495fa019a06/include/charls/public_types.h:511:20: error: ‘underlying_type_t’ in namespace ‘std’ does not name a template type
     using T = std::underlying_type_t<encoding_options>;
                    ^~~~~~~~~~~~~~~~~
```

uilianries: @Nekto89 The error is about charls, not gdcm. I'll open a RP charls adding a new restriction based on the compiler version.

uilianries: @Nekto89 After taking a look in gdcm and charls code, now I see your point. Indeed the C++14 is required since Charls uses it in its public headers. As Conan 1.x does not provide cppstd, we would need to drop more compilers for chars, but it won't be fair, as they are working. That's mean, your change makes sense, thank your for explaining us all the situation. I'll send a patch to GDCM, trying to configure the required cppstd version in the cmake file.

recipes/gdcm/all/patches/0001-3.0.20-find-charls.patch
uilianries: It looks for compatible version, so 2.6.2 fits. Only when `EXACT` is present in `find_package` CMake will requires only that specific version.

https://cmake.org/cmake/help/v3.15/command/find_package.html

uilianries (approved):
LGTM

SSE4 (approved):

PR #17968, opened 2023-06-19
cuda-kat: update recipe to Conan v2

valgur:
Specify library name and version:  **cuda-kat/0.2**

---

### DIFF
--- a/recipes/cuda-kat/all/conanfile.py
+++ b/recipes/cuda-kat/all/conanfile.py
@@ -73,4 +73,7 @@ class CudaKatConan(ConanFile):
         self.cpp_info.libdirs = []
 
         if is_msvc(self):
+            # Required for C++ version checks to work at
+            # https://github.com/eyalroz/cuda-kat/blob/44d381b/src/kat/detail/constexpr_by_cpp_version.hpp
+            # Otherwise MSVC will always report __cplusplus as 199711L
             self.cpp_info.cxxflags.append("/Zc:__cplusplus")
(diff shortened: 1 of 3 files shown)


recipes/cuda-kat/all/conanfile.py
```diff
@@ -29,26 +33,44 @@ def _compilers_minimum_version(self):
...
+        )
 
     def package_info(self):
-        self.cpp_info.names["cmake_find_package"] = "cuda-kat"
-        self.cpp_info.names["cmake_find_package_multi"] = "cuda-kat"
+        self.cpp_info.bindirs = []
+        self.cpp_info.libdirs = []
+
+        if is_msvc(self):
+            self.cpp_info.cxxflags.append("/Zc:__cplusplus")
```
memsharded, line 76: This is the typical line that we would appreciate a bit more of context, like one comment explaining why this flag is necessary, in what cases it fails if it is not there. For this one, it is not very clear why the consumers would really care about the pre-processor macro, if they rely on it, it should be their build systems that activate it?

memsharded (commented):
This is looking good,  many thanks for contributing it.
Just a minor comment, please have a look

recipes/cuda-kat/all/conanfile.py
valgur: Sure. I briefly mentioned the reasoning in the commit message, but this is a much better location for it, I agree. Added it here as well.

memsharded: That helps a lot to simplify future maintenance, being able to read things directly in recipe, saves a lot of hops to understand things, many thanks!

memsharded (approved):

SSE4 (approved):

PR #17982, opened 2023-06-20
highs: add version 1.5.3

hedtke:
Specify library name and version:  **highs/1.5.3**

---

### DIFF
--- a/recipes/highs/all/conanfile.py
+++ b/recipes/highs/all/conanfile.py
@@ -65,7 +65,7 @@ class HiGHSConan(ConanFile):
         apply_conandata_patches(self)
         cmake = CMake(self)
         cmake.configure()
-        cmake.build(target="libhighs")
+        cmake.build(target="highs")
 
     def package(self):
         copy(self, pattern="LICENSE", src=self.source_folder, dst=join(self.package_folder, "licenses"))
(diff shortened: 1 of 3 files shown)


recipes/highs/all/conanfile.py
```diff
@@ -65,7 +65,7 @@ def build(self):
         apply_conandata_patches(self)
         cmake = CMake(self)
         cmake.configure()
-        cmake.build(target="libhighs")
+        cmake.build(target="highs")
```
RubenRBS, line 68: I'm guessing the target changed between versions? If the old target is correct for the old one, this should be conditional on the current version, no?

RubenRBS (commented):
Hi! Thanks a lot for your contribution, we appreciate it :) I have one minor question regarding the target change

recipes/highs/all/conanfile.py
hedtke: Yes, target name changed. The new name also works with the old version :-)

RubenRBS: Oh, that's great then, thanks for clarifying!

SSE4 (approved):

jcar87 (approved):

PR #17995, opened 2023-06-21
fontconfig: migrate to util-linux-libuuid

ericLemanissier:
To avoid conflict with system libuuid

Specify library name and version:  **fontconfig/***

see https://github.com/conan-io/conan-center-index/pull/17664 and https://github.com/conan-io/conan-center-index/pull/17427#issuecomment-1545476420 for motivation and rationale

---

### DIFF
--- a/recipes/fontconfig/meson/conanfile.py
+++ b/recipes/fontconfig/meson/conanfile.py
@@ -52,11 +52,11 @@ class FontconfigConan(ConanFile):
         self.requires("freetype/2.13.0")
         self.requires("expat/2.5.0")
         if self.settings.os == "Linux":
-            self.requires("libuuid/1.0.3")
+            self.requires("util-linux-libuuid/2.39")
 
     def build_requirements(self):
         self.tool_requires("gperf/3.1")
-        self.tool_requires("meson/1.0.1")
+        self.tool_requires("meson/1.1.1")
         if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
             self.tool_requires("pkgconf/1.9.3")
 


ericLemanissier:
blocked by:
- https://github.com/conan-io/conan-center-index/pull/17664

ericLemanissier:
blocks:
- https://github.com/conan-io/conan-center-index/pull/17427
- https://github.com/conan-io/conan-center-index/pull/17485

jwillikers (approved):

samuel-emrys (approved):

jcar87:
Hi @ericLemanissier - thanks for your contribution. On the surface of this, this makes sense. However, would you able to provide more details (preferably in the PR description) about how these potential conflicts manifest when consuming this recipe? Thanks!

ericLemanissier:
@jcar87 see https://github.com/conan-io/conan-center-index/pull/17664

jcar87 (approved):

jcar87:
Thanks Eric! And @samuel-emrys  for the investigation and follow-up.

For future PRs please ensure there is enough information regarding what's motivating the changes - in particular in the PR description. This not only makes it easier (and speedier) for us to review, but users experiencing the very same errors may find it helpful as well.

czoido (approved):

prince-chrismc:
@ericLemanissier @jwillikers @samuel-emrys We are going to be rolling this back since it's causing lots of issues with the graph conflicts

ericLemanissier:
Why not moving forward, instead of backwards?

prince-chrismc:
It's easier to fix one then try to track down all the other recipes and get PRs through on a Friday :)
Now we know there's a problem we can plan how to get through it.

Hopefully that's a reasonable response :)

samuel-emrys:
If this is being rolled back, what's the way forward on system package conflicts? Ie xorg/system which brings in the util-linux-libuuid variant and conflicts with libuuid

prince-chrismc:
@samuel-emrys I do not see that searching the repo https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+util-linux-libuuid&type=code so I am not sure where that is being pulled in 🤔 If you can give me more imformation I'd appreciate that 🙏 

We have only 16 recipes using the libuuid 
https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index%20requires%20libuuid&type=code from a basic search and I think we can make a simple plan to introduce all those changes "at once" but what see

samuel-emrys:
@prince-chrismc see https://github.com/conan-io/conan-center-index/pull/17427#issuecomment-1545476420 for context. fontconfig is where libuuid is being pulled into the tree

samuel-emrys:
https://github.com/conan-io/conan-center-index/pull/17664 is where util-linux-libuuid was added. All packages currently using libuuid should migrate to util-linux-libuuid and libuuid should be deprecated

ericLemanissier:
I don't think we can migrate the 16 recipes "at once" because there are dependencies between these 16 packages, but it is true that we should all be ready to approve quickly the 16 PRs when the migration is started, to minimize the disruption (hoping that github or C3I don't let us down during the process)

The conflict Sam is talking about is one that cannot be detected by conan, because it is a conflict between a system package and a conan package.

prince-chrismc:
> All packages currently using libuuid should migrate to util-linux-libuuid and libuuid should be deprecated

Can we (i.e the Conan team) plan this next week so we cause users the least amount of disruption? Does that make sense?
I know historically it was one at a time by the community, but for cases like this, the reinvigorate team wants to find a better solution 🤞 

The issue is it's late at night and most have logged off so we decided to save the work for next week, a few more days will not delay other efforts -- it's bad timing for a human perspective

> The conflict Sam is talking about is one that cannot be detected by conan, because it is a conflict between a system package and a conan package.

Could the `xorg/system` package also list the requirement on `util-linux-libuuid` or a provides so that an error get's flagged? (thinking out loud) how do we as maintainers document that in the code? 

***

There's no objection to this change getting rolled out but we just want to do it correctly.  :)

samuel-emrys:
> Can we (i.e the Conan team) plan this next week so we cause users the least amount of disruption? Does that make sense?

Yep this sounds reasonable to me.

> Could the `xorg/system` package also list the requirement on `util-linux-libuuid` or a provides so that an error get's flagged? (thinking out loud) how do we as maintainers document that in the code? 

Yeah this isn't a bad idea. `xorg` wouldn't use it obviously, because it's a system package, but that would at least cause a conflict if `libuuid` exists elsewhere in the dep tree. This would have to be applied to any other system packages on CCI as well, if they have the dependency.

PR #18007, opened 2023-06-21
chaiscript: conan v2 support

holger-detering:
Specify library name and version:  **chaiscript/6.1.0**

---


RubenRBS (approved):
Looks great, thanks a lot!

I'm approving this one before it completes building, but checking upstream's repo, https://github.com/ChaiScript/ChaiScript#requirements says that it requires C++17, so this might fail in Windows, because unlike most platforms in CCI which C++17 is the oldest one they try,  msvc in Windows uses 14

Adding a `validate()` method that checks for it would fix that problem, [see the cmake template for an exmaple on how to do just this](https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L80)

RubenRBS (approved):
Thanks a lot! :)

recipes/chaiscript/all/conanfile.py
```diff
@@ -21,60 +29,78 @@ class ChaiScriptConan(ConanFile):
...
-
-    def source(self):
-        tools.get(**self.conan_data["sources"][self.version])
-        extracted_dir = "ChaiScript-" + self.version
-        os.rename(extracted_dir, self._source_subfolder)
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "7",
+            "clang": "7",
+            "apple-clang": "10",
```
RubenRBS, line 40: Sorry I didn't catch this earlier, seems like the msvc compiler is also missing here:

```suggestion
            "apple-clang": "10",
            "Visual Studio": "15.0",
            "msvc": "191",
```

RubenRBS (changes requested):
Jenkins auto-restarted due to some load issues, taking the chance to suggest some extra changes

RubenRBS (dismissed):

recipes/chaiscript/all/conanfile.py
jcar87: I believe the `msvc` and `Visual Studio` entries in this dictionary may not be evaluated since the check only happens conditionally on `not msvc(self)` on line 56. @RubenRBS, may be a good opportunity to look into updating the template, since other recipes are currently applying a newer pattern for this.

Either way not a blocker for this PR.

jcar87 (approved):

RubenRBS (approved):

PR #18014, opened 2023-06-22
[libinterpolate]: added version 2.6.4, which adds Windows support.

CD3:
Specify library name and version: libinterpolate/2.6.4

Finally figured out why Windows builds were failing.

---

### DIFF
--- a/recipes/libinterpolate/all/conanfile.py
+++ b/recipes/libinterpolate/all/conanfile.py
@@ -43,6 +43,10 @@ class PackageConan(ConanFile):
         self.requires("eigen/3.3.7", transitive_headers=True)
 
     def validate(self):
+        if Version(self.version) < "2.6.4" and self.settings.os != "Linux":
+            raise ConanInvalidConfiguration(f"{self.ref} is not supported by {self.settings.os}; Try the version >= 2.6.4")
+        if Version(self.version) >= "2.6.4" and self.settings.os not in ["Linux", "Windows"]:
+            raise ConanInvalidConfiguration(f"{self.ref} is not supported by {self.settings.os}.")
         if self.settings.compiler.get_safe("cppstd"):
             check_min_cppstd(self, self._min_cppstd)
         minimum_version = self._compilers_minimum_version.get(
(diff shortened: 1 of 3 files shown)


CD3:
I started work on adding another recipe and pushed it to my master branch not realizing it would be pulled into this merge request. I'll split it into a separate branch.

recipes/libinterpolate/all/conanfile.py
```diff
@@ -43,9 +43,6 @@ def requirements(self):
         self.requires("eigen/3.3.7", transitive_headers=True)
 
     def validate(self):
```
uilianries, line 45:
```suggestion
    def validate(self):
        if Version(self.version) < "2.6.4" and self.settings.os != "Linux":
            raise ConanInvalidConfiguration(f"{self.ref} is not supported by {self.settings.os}; Try the version >= 2.6.4")
        if Version(self.version) >= "2.6.4" and self.settings.os not in ["Linux", "Windows"]:
            raise ConanInvalidConfiguration(f"{self.ref} is not supported by {self.settings.os}.")
```

If I understand correctly, Windows has been supported since 2.6.4, but not before. Plus, only Linux and Windows are supported now, other OS like FreeBSD and Mac are not supported yet.

CD3: Yes, good catch. I wasn't thinking about the same recipe being used for the older versions. The library may build on the other systems but I can't test it, so they should be disabled until I am able to confirm.

uilianries: @CD3 if you can not confirm it, no problem, we can merge anyway, just add a TODO comment asking to check it in the future.

uilianries: No, as a comment in the code, like:

`# TODO: Verify if older versions than 2.6.4 can be built with Windows.`

CD3: Ah, got it.

uilianries (approved):
LGTM

jcar87 (approved):

PR #18020, opened 2023-06-22
entt: add version 3.12.2

RazielXYZ:
Specify library name and version:  **entt/3.12.2**

---

### DIFF
--- a/recipes/entt/3.x.x/conandata.yml
+++ b/recipes/entt/3.x.x/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "3.12.1":
+    url: "https://github.com/skypjack/entt/archive/refs/tags/v3.12.1.tar.gz"
+    sha256: "7DC4FC74CC32C1EC74B37419140D9334563BB22AB1B92AD9BE580703DA05B8AC"
   "3.11.1":
     url: "https://github.com/skypjack/entt/archive/refs/tags/v3.11.1.tar.gz"
     sha256: "0ac010f232d3089200c5e545bcbd6480cf68b705de6930d8ff7cdb0a29f5b47b"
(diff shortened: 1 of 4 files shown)


recipes/entt/3.x.x/conanfile.py
```diff
@@ -73,7 +73,3 @@ def package_info(self):
         self.cpp_info.set_property("cmake_target_name", "EnTT::EnTT")
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
-
-        # TODO: to remove in conan v2
-        self.cpp_info.names["cmake_find_package"] = "EnTT"
-        self.cpp_info.names["cmake_find_package_multi"] = "EnTT"
```
czoido, line 79: Thanks a lot for the contribution @RazielXYZ.
One thing, this should be left in the recipe, it's not required to add those for new recipes but we are not removing this for the moment to not break users that may use the legacy generators in Conan 1.X. Some time in the future we will remove this from recipes.

recipes/entt/3.x.x/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "3.12.1":
```
czoido, line 2: This would be a good opportunity to drop some versions. We could leave only the three latest minor versions: 3.12.1, 3.11.1, 3.10.3

recipes/entt/3.x.x/conanfile.py
RazielXYZ: Ah, makes sense, I'll put it back

recipes/entt/3.x.x/conandata.yml
RazielXYZ: That is fine by me!

recipes/entt/3.x.x/patches/0002-3.12.1-fix-gcc-7.0-7.3.patch
```diff
@@ -0,0 +1,20 @@
+--- a/src/entt/meta/meta.hpp
```
czoido, line 1: Just one comment for the record, we'll let the patch because it was previously applied but in general we prefer to not apply patches to source code.

RazielXYZ: I agree; not sure how important this patch is, it seems to be so that EnTT works with GCC 7, which I assume doesn't have full/correct constexpr support. That said, EnTT itself asks for a fully C++17 compiler, which GCC 7 is supposed to be as far as I know. Not sure how relevant it is anymore or how many people still use it, but the reason why I included an updated patch as well is indeed because it was already previously applied.

RazielXYZ:
I've updated this to 3.12.2 instead of 3.12.1, as that has just been released and we might as well be on latest!

jcar87 (approved):
LGTM, with agreement that patches to C++ code in this PR predate this PR.

PR #18034, opened 2023-06-23
opentelemetry: fixup requirements

Cogitri:
Some of these tool requirements are only needed with some options enabled

Specify library name and version:  **opentelemetry/***

---

### DIFF
--- a/recipes/opentelemetry-cpp/all/conanfile.py
+++ b/recipes/opentelemetry-cpp/all/conanfile.py
@@ -126,7 +126,7 @@ class OpenTelemetryCppConan(ConanFile):
             self.requires("thrift/0.17.0")
 
             if Version(self.version) >= "1.3.0":
-                self.requires("boost/1.81.0")
+                self.requires("boost/1.82.0")
 
     @property
     def _required_boost_components(self):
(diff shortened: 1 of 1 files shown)


RubenRBS:
Retriggering CI builds after the last ones failed due to some issue in CI on the weekend

RubenRBS:
Hi @Cogitri can you bump the boost dependency to 1.82 to resolve the conflict please? Thanks!

RubenRBS (approved):
Thanks a lot!

recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -440,6 +443,7 @@ def package_info(self):
             self.options.with_elasticsearch
         ):
             self.cpp_info.components[self._http_client_name].requires.append("libcurl::libcurl")
+            self.cpp_info.components[self._http_client_name].requires.append("openssl::openssl")
```
czoido, line 446: Hi @Cogitri! Thanks for the contribution.
Shouldn't this also be added to the `opentelemetry_exporter_etw` component?

SSE4 (approved):

recipes/opentelemetry-cpp/all/conanfile.py
SSE4: @czoido could you possibly explain it in more details why should it be added, from your point of view? what's the motivation, what's the benefit, etc?

czoido: @SSE4 It's because the recipe logic, when you have `with_etw=True` it adds _openssl_ as a requirement if I'm not mistaken.

PR #18038, opened 2023-06-23
chore: update version

tim-goto:
Specify library name and version:  **gurkenlaeufer/1.1.0**

---


memsharded (changes requested):
Hi @tim-goto 

Thanks for your contribution, a couple of comments:

- typically, older versions are maintained, unless there are many of them, too outdated, or irrelevant. So you can leave 1.0.1.
- You can add a new version, but it is also necessary to define the ``config.yml`` file that defines the mapping (almost always is mapping to "all" folder, but just in case some recipes diverge too much)

jwillikers (approved):

SSE4:
@memsharded can you unblock? I've read your comment, and don't understand what is needed to proceed.
you only said few things starting with `you can`, that looks like an optional recommendation (e.g. "you can do it, but it doesn't mean you shall and you will").

tim-goto:
I worked in the recommendations from memsharded so they are implemented

RubenRBS (approved):

jcar87 (approved):
@tim-goto  thanks for your contribution and for implementing the reviewer's requests :)

wrmbaron:
How long does it take, until this merge gets visible on https://conan.io/center/ ? It is not visible yet.

tim-goto:
The web interface is always a bit behind and lacks a little attention with the still ongoing efforts of conanv2, you can already use the package right after the merge, see:

Issue #17989, opened 2023-06-21
[request] sfml/2.6.0


ChrisThrasher:
### Package Name/Version

sfml/2.6.0

### Changelog

https://github.com/SFML/SFML/releases/tag/2.6.0

### Context about the new update

SFML 2.6.0 was released earlier today.

https://github.com/conan-io/conan-center-index/compare/master...ChrisThrasher:conan-center-index:sfml/2.6.0

I attempted this myself and made some progress. I was able to fix some broken patches and remove two patches that are now no longer necessary. The blocking issue was that we need to include glad/gl.h from the extlibs/ directory, however that whole directory gets deleted. I don't believe glad/gl.h is actually part of glad since as you can see in the above patch I depended on glad but the compiler still could not find that header. Using the in-tree version may be suitable.

There were also some issues not being able to find Audio module dependencies like Vorbis and FLAC that I couldn't make sense of either.

toge:
@ChrisThrasher 
Thank you for your information.
It help me a lot.

>  I don't believe glad/gl.h is actually part of glad since as you can see in the above patch I depended on glad but the compiler still could not find that header. Using the in-tree version may be suitable.

I agree with you.
`glad/gl.h` seems to be generated from https://gen.glad.sh/.
I decided to use `glad` in `extlibs`.

If possible, please review https://github.com/conan-io/conan-center-index/pull/18069.

PR #18069, opened 2023-06-27
sfml: add version 2.6.0

toge:
Specify library name and version:  **sfml/***

minimp3 and grad are needed since sfml/2.6.0.
However, cci's these recipes is too old to replace.
So this recipe uses extlibs included in sfml/2.6.0 tarball.

Try to close https://github.com/conan-io/conan-center-index/issues/17989.

---


ChrisThrasher (approved):
I think this looks good! You may just need to disable the `WARNINGS_AS_ERRORS` CMake option to ignore some warnings that are breaking the build. I believe this was another thing introduced between 2.5.1 and 2.6.0.

toge:
> I think this looks good! You may just need to disable the `WARNINGS_AS_ERRORS` CMake option to ignore some warnings that are breaking the build. I believe this was another thing introduced between 2.5.1 and 2.6.0. It defaults to `ON`.

@ChrisThrasher 
Thanks a lot!
I added it.

RubenRBS (approved):

recipes/sfml/all/conanfile.py
```diff
@@ -36,6 +38,24 @@ class SfmlConan(ConanFile):
...
+    def _compilers_minimum_version(self):
+        return {
+            "17": {
+                "gcc": "8",
+                "clang": "7",
+                "apple-clang": "12",
+                "Visual Studio": "16",
+                "msvc": "192",
+            },
+        }.get(self._min_cppstd, {})
```
ChrisThrasher, line 57: SFML 2 only requires C++03 so this should not be necessary. SFML 3 requires C++17 but that has not been released yet.

toge: @ChrisThrasher 
Thank you for your comment.
I misunderstand SFML policy.

But I found several files in 2.6.0 use C++11 features.
https://github.com/SFML/SFML/blob/2.6.0/src/SFML/Audio/SoundFileReaderMp3.cpp#L58

I will make `CMAKE_CXX_STANDARD` 11 in `conanfile.py`

ChrisThrasher: That's fine. We've tried to maintain C++03 compatibility but apparently some C++11 things have slipped in. Nearly every compiler used these days supports C++11 so this won't be a problem in practice.

recipes/sfml/all/patches/2.6.0-0006-disable-warning-flags.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/cmake/Macros.cmake b/cmake/Macros.cmake
+index e0118eb..85a7714 100644
+--- a/cmake/Macros.cmake
++++ b/cmake/Macros.cmake
+@@ -66,8 +66,6 @@ macro(sfml_add_library target)
+         add_library(${target} ${THIS_SOURCES})
+     endif()
+ 
+-    set_file_warnings(${THIS_SOURCES})
```
jcar87, line 9: A bit unfortunate that setting `WARNINGS_AS_ERRORS` to `OFF` was not sufficient due to newer warnings being referenced than gcc 5 supports.
I've opened an issue with upstream sfml: https://github.com/SFML/SFML/issues/2594

Chances are gcc 5 is no longer supported - if this is the case, we may chose to stop supporting it for this recipe as well.

jcar87 (approved):

recipes/sfml/all/patches/2.6.0-0006-disable-warning-flags.patch
toge: Thanks for quick action!

RubenRBS (approved):

recipes/sfml/all/patches/2.6.0-0006-disable-warning-flags.patch
ChrisThrasher: > Chances are gcc 5 is no longer supported

I'm simply a matter of nobody using GCC 5 with SFML 2 anymore. If you removed those warnings I bet it would work. It's not as if the maintainers made an intentional decision to stop supporting GCC 5.

ChrisThrasher:
Thank you so much!

PR #17364, opened 2023-05-02
Adding Botan 3.0.0 recipe

Maaown:
Specify library name and version:  **botan/3.0.0**

fixes https://github.com/conan-io/conan-center-index/issues/17247

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18079, opened 2023-06-27
Adding Botan 3.0.0 recipe and make it compatible with Conan 2.x

### LINKED ISSUE #15133: [package] botan/2.19.3 Unknown -fPIC flag
### Description

Build botan with MSVC. I am using Visual Studio 2017 x64 toolchain. The build log warns about an unknown `-fPIC`.

```
cl   /EHs /GR /D_WIN32_WINNT=0x0600 /MD /bigobj   /O2 /Oi -fPIC -DBOTAN_IS_BEING_BUILT /W4 /wd4250 /wd4251 /wd4275 /wd4127  /Ibuild\include /Ibuild\include\external /nologo /c botan_all.cpp /Fobuild\obj\lib\botan_all.obj
cl : Command line warning D9002 : ignoring unknown option '-fPIC'
```

* Package Name/Version: **botan/2.19.3**
* Compiler+version: **MSVC 19.16.27048**
* Conan version: **conan 1.54.0**
* Python version: **Python 3.10.8**

### Conan profile

[settings]
arch=x86_64
arch_build=x86_64
build_type=MinSizeRel
compiler=Visual Studio
compiler.runtime=MD
compiler.toolset=v141
compiler.version=15
os=Windows
os_build=Windows
[options]
[build_requires]
[env]

### Steps to reproduce

Just `conan install` on a conanfile.py containing botan

### Logs

```
botan/2.19.3: Copying sources to build folder
botan/2.19.3: Building your package in C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\build\e7993c7f0c660b49f544baf0c45805ef3bdd797d
botan/2.19.3: Generator txt created conanbuildinfo.txt
botan/2.19.3: Aggregating env generators
botan/2.19.3: Calling build()
   INFO: ./configure.py invoked with options "--build-targets=static --distribution-info=Conan --without-documentation --cc-abi-flags=  --extra-cxxflags=-fPIC --cc=msvc --cpu=x86_64 --prefix=C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\package\e7993c7f0c660b49f544baf0c45805ef3bdd797d --os=windows --amalgamation --msvc-runtime=MD --without-pkg-config"
   INFO: Configuring to build Botan 2.19.3 (revision unknown)
   INFO: Running under 3.10.8 (tags/v3.10.8:aaaf517, Oct 11 2022, 16:50:30) [MSC v.1933 64 bit (AMD64)]
   INFO: Autodetected platform information: OS="Windows" machine="AMD64" proc="Intel64 Family 6 Model 158 Stepping 13, GenuineIntel"
   INFO: Auto-detected compiler version 19.16
   INFO: Auto-detected compiler arch x86_64
   INFO: Target is msvc:19.16-windows-x86_64
   INFO: Assuming target x86_64 is little endian
   INFO: Skipping (dependency failure): asio certstor_sqlite3 sessions_sqlite3
   INFO: Skipping (incompatible CPU): aes_armv8 aes_power8 sha1_armv8 sha2_32_armv8 sm4_armv8
   INFO: Skipping (incompatible OS): certstor_system_macos commoncrypto dev_random fd_unix getentropy proc_walk
   INFO: Skipping (incompatible compiler): argon2_ssse3 serpent_avx2 sha2_32_bmi2 sha2_64_bmi2 sha3_bmi2
   INFO: Skipping (no enabled compression schemes): compression
   INFO: Skipping (requires external dependency): boost bzip2 lzma sqlite3 tpm zlib
   INFO: Loading modules: adler32 aead aes aes_ni aes_vperm aont argon2 aria asn1 auto_rng base base32 base58 base64 bcrypt bcrypt_pbkdf bigint blake2 block blowfish camellia cascade cast128 cast256 cbc cbc_mac ccm cecpq1 certstor_flatfile certstor_sql certstor_system certstor_system_windows cfb chacha chacha20poly1305 chacha_avx2 chacha_rng chacha_simd32 checksum cmac comb4p cpuid crc24 crc32 cryptobox ctr curve25519 des dh dl_algo dl_group dlies dsa dyn_load eax ec_group ecc_key ecdh ecdsa ecgdsa ecies eckcdsa ed25519 elgamal eme_oaep eme_pkcs1 eme_raw emsa1 emsa_pkcs1 emsa_pssr emsa_raw emsa_x931 entropy ffi filters fpe_fe1 gcm ghash ghash_cpu ghash_vperm gmac gost_28147 gost_3410 gost_3411 hash hash_id hex hkdf hmac hmac_drbg hotp http_util idea idea_sse2 iso9796 kasumi kdf kdf1 kdf1_iso18033 kdf2 keccak keypair lion locking_allocator mac mce mceies md4 md5 mdx_hash mem_pool mgf1 misty1 mode_pad modes mp newhope nist_keywrap noekeon noekeon_simd numbertheory ocb ofb par_hash passhash9 pbes2 pbkdf pbkdf1 pbkdf2 pem pgp_s2k pk_pad pkcs11 poly1305 poly_dbl prf_tls prf_x942 processor_rng psk_db pubkey rc4 rdrand_rng rdseed rfc3394 rfc6979 rmd160 rng roughtime rsa salsa20 scrypt seed serpent serpent_simd sessions_sql sha1 sha1_sse2 sha1_x86 sha2_32 sha2_32_x86 sha2_64 sha3 shacal2 shacal2_avx2 shacal2_simd shacal2_x86 shake shake_cipher simd simd_avx2 siphash siv skein sm2 sm3 sm4 socket sodium sp800_108 sp800_56a sp800_56c srp6 stateful_rng stream streebog system_rng thread_utils threefish_512 threefish_512_avx2 tiger tls tls_10 tls_cbc tss twofish utils uuid whirlpool win32_stats x509 x919_mac xmss xtea xts zfec zfec_sse2 zfec_vperm
   INFO: Using hardlink to link files into build dir (use --link-method to change)
   INFO: Writing amalgamation header to botan_all.h
   INFO: Writing amalgamation source to botan_all.cpp
   INFO: Botan 2.19.3 (revision unknown) (unreleased undated) build setup is complete
Conan:vcvars already set 

Microsoft (R) Program Maintenance Utility Version 14.16.27048.0
Copyright (C) Microsoft Corporation.  All rights reserved.

	cl   /EHs /GR /D_WIN32_WINNT=0x0600 /MD /bigobj   /O2 /Oi -fPIC -DBOTAN_IS_BEING_BUILT /W4 /wd4250 /wd4251 /wd4275 /wd4127  /Ibuild\include /Ibuild\include\external /nologo /c botan_all.cpp /Fobuild\obj\lib\botan_all.obj
cl : Command line warning D9002 : ignoring unknown option '-fPIC'
botan_all.cpp
botan_all.cpp(100497): warning C4101: 'ex': unreferenced local variable
	lib /nologo /OUT:./botan.lib build\obj\lib\botan_all.obj
botan/2.19.3: Package 'e7993c7f0c660b49f544baf0c45805ef3bdd797d' built
botan/2.19.3: Build folder C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\build\e7993c7f0c660b49f544baf0c45805ef3bdd797d
botan/2.19.3: Generated conaninfo.txt
botan/2.19.3: Generated conanbuildinfo.txt
botan/2.19.3: Generating the package
botan/2.19.3: Package folder C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\package\e7993c7f0c660b49f544baf0c45805ef3bdd797d
botan/2.19.3: Calling package()
Conan:vcvars already set 

Microsoft (R) Program Maintenance Utility Version 14.16.27048.0
Copyright (C) Microsoft Corporation.  All rights reserved.

	"C:\Python\Python310\python.exe" "src/scripts/install.py" --prefix="C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\package\e7993c7f0c660b49f544baf0c45805ef3bdd797d" --build-dir="build" --bindir="C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\package\e7993c7f0c660b49f544baf0c45805ef3bdd797d\bin" --libdir="C:\Source\conan-pkgs\.conan\data\botan\2.19.3\_\_\package\e7993c7f0c660b49f544baf0c45805ef3bdd797d\lib" --docdir="docs" --includedir="include"
   INFO: Botan 2.19.3 installation complete
botan/2.19.3 package(): Packaged 308 '.h' files
botan/2.19.3 package(): Packaged 1 '.lib' file: botan.lib
botan/2.19.3 package(): Packaged 1 '.txt' file: license.txt
botan/2.19.3: Package 'e7993c7f0c660b49f544baf0c45805ef3bdd797d' created
botan/2.19.3: Created package revision 5241133ec1aef36217537ff75224b4a6
```

</details>

### LINKED ISSUE #17247: [request] botan/2.19.3 - Upgrade conanfile to 2.x format
### Package Name/Version

botan/2.19.3

### Changelog

.

The recipe for botan v2.19.3 already exists, however we use it at work and since moving to 2.x is underway, we may soon have the need for an updated recipe.
I am available to perform the operation.

lieser:
Specify library name and version:  **botan/3.0.0**

- Make recipe compatible with Conan 2.x. Closes #17247.
- Add Botan 3.0.0, 3.1.0, 3.1.1, 3.2.0.
- Remove some older Botan versions.
- Fix unknown `-fPIC` flag warning in MSVC build. Fixes #15133.
- Support the `tools.build:sysroot` config option.

Replaces inactive PR #17364.

---


trns1997:
@lieser quick note, I just faced a problem where the `CXXFLAGS` defined in my profile file under the `[env]` tags is not used or parsed. Could you look into it? I will see on my end whats going on in further detail tomorrow :) .

lieser:
> I just faced a problem where the `CXXFLAGS` defined in my profile file under the `[env]` tags is not used or parsed.

@trns1997 Note that `[env]` is deprecated and as far as I know does not work in Conan 2. Try using `[buildenv]` instead. https://docs.conan.io/1/reference/profiles.html#buildenv

lieser:
@memsharded Currently the CI fails with `This commit cannot be built`. Probably because all compilers in the CI are to old to build Botan 3.0.0.

Does that mean it will not be possible to add Botan 3.0.0 until the CI gets updated, or Botan is somehow made compatible with at least one compiler?

There is a potential chance to make GCC 11.1 compatible, as GCC 11 has C++ 20 support. But unsure if it will be worth the effort to try working around the bugs in GCC 11.1 that affect Botan.

---

Another question is regarding the new `compiler.libcxx` setting now given in the Conan 2 profile.

How should a library like Botan with a custom build system that has a hard coded C++ language setting behave?

- Simply ignore it? (like it is done with the current recipe)
- In the `validate()` method check that `compiler.libcxx` is at least the hard coded version or higher?
- In the `configure()` method delete it with `self.settings.rm_safe("compiler.cppstd")` so it does not affect the `package_id`?

trns1997:
> @lieser quick note, I just faced a problem where the `CXXFLAGS` defined in my profile file under the `[env]` tags is not used or parsed. Could you look into it? I will see on my end whats going on in further detail tomorrow :) .

hey @lieser, I just figured out why it no works for us. Basically we used to pass the `CXXFLAGS` via the `[env]` tags in profiles when using Conan 1.55.0 but in the `conanfile.py` we explicitly fetch the cxxflags via the `[conf]` tag. So I just updated the profile and passed the cxxflags via `tools.build:cxxflags`

recipes/botan/all/test_package/CMakeLists.txt
```diff
@@ -1,11 +1,8 @@
...
 
-include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-conan_basic_setup(TARGETS)
-
 find_package(botan REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
 target_link_libraries(${PROJECT_NAME} botan::botan)
-set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 11)
+set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 20)
```
trns1997, line 8: @lieser `ERROR: botan/2.19.3 (test package): Error in build() method, line 26` the upgrade of compiler requirements is true for Botan 3 onwards. But building `test_package` for older version should not have this requirement otherwise it breaks the build.

lieser:
@trns1997 Did you also see my previous comment about the `[buildenv]`, which is the Conan 2 replacement for `[env]`?

The recipe in this PR should understand both the `CXXFLAGS` from the `[buildenv]` or the `tools.build:cxxflags` config in `[conf]`. What you use probably does not make a difference.

recipes/botan/config.yml
```diff
@@ -29,3 +9,5 @@ versions:
     folder: all
   "2.19.3":
     folder: all
+  "3.0.0":
+    folder: all
```
lieser, line 13: Note that Version 3.0.0 is not build in the CCI CI, as all compiler are to old.

recipes/botan/all/conanfile.py
```diff
@@ -140,59 +145,108 @@ def requirements(self):
...
+        # disabled compiler optimizations.
+        self._extra_cxxflags = self._cxxflags
+        self.buildenv.unset('CXXFLAGS')
+        VirtualBuildEnv(self).generate()
 
     def build(self):
-        for patch in self.conan_data.get('patches', {}).get(self.version, []):
-            tools.patch(**patch)
-        with tools.chdir(self._source_subfolder):
+        apply_conandata_patches(self)
```
lieser, line 235: Maybe a Conan expert can answer this with a reason why:
Is `build()` or `source()` the best place to put `apply_conandata_patches()`?

The Boost recipe calls it `source()`, and it make sense to me conceptually and from what is written in https://docs.conan.io/2/reference/conanfile/methods/source.html. But in most recipe it seems to be called in the `build()` method.

recipes/botan/all/conanfile.py
```diff
@@ -377,9 +421,15 @@ def _configure_cmd(self):
 
         build_flags.append('--without-pkg-config')
 
+        if self.settings.os == 'Linux':
+            # The current Ubuntu 18.04 runners in Conan Center Index don't have getrandom/getentropy,
+            # which is enabled by default starting in Botan 3.0.0
+            build_flags.append('--without-os-feature=getrandom')
+            build_flags.append('--without-os-feature=getentropy')
```
lieser, line 428: Maybe this should be removed again, as the CI is currently not not run for 3.0.0 (there this is enabled by default) because of old compilers anyway?

recipes/botan/all/conanfile.py
RubenRBS: Hi! Good question. The original idea is that patching in the source method should be done for cases where the patch is common to every version, and in build when you need to differenciate per configuration ([See here for an example in the docs](https://docs.conan.io/2/examples/tools/files/patches/patch_sources.html))

But when dealing with CCI patches, you might find some relaxing of thos ideas, and people either patch in `source()` because it's the "correct" thing to do, or fall back to patching in `build()` so they don't need to think about those cases, as patching in `build()` would always be correct

RubenRBS (commented):
Hi! Thanks a lot for your contribution (And patience while we processed the rest of the PR backlog!)
Most of the changes look good, but I have to double check with the team a few things, I'll ping you back once I know more!

Again, thanks a lot! :)

lieser:
@RubenRBS Any update on the review status? Currently nor urgent for us, but would still be nice if I could finish the work on this in the next weeks.

RubenRBS:
Hi @lieser thanks for the ping, will get this reviewed in the next few days, thanks a lot for your patience :)

recipes/botan/all/conanfile.py
uilianries: Not compilers, but glibc version. This error is listed in the upstream: https://github.com/randombit/botan/issues/2607#issuecomment-1382826977 but we can't do nothing about because Docker images used in CCI are not running on latest Ubuntu version to avoid breaking users with non-updated distros. I would suggest adding an option like `with_getrandom: False` where it uses `False` by default.

In CCI we try to preserve the upstream behavior always, but at same time, packages should be built, so this case deserves an exception.

Thank you for checking it !

recipes/botan/all/conanfile.py
RubenRBS, line 1: Locally, running `conan create . --version=3.0.0 -o "&:with_openssl=True"` fails because 

```
configure.py: error: no such option: --with-openssl
````

Looking at this the configure script seems to have changed between major versions, and in fact I can't see openssl mentioned in 3.0.0, which 2.19.3 does have them.

Note that trying to build 2.19.3 also fails, but for a different reason:
````
INFO: ./configure.py invoked with options "--build-targets=static --distribution-info=Conan --without-documentation --cc-abi-flags=  --extra-cxxflags=-fPIC --cc=gcc --cpu=armv8 --prefix=/root/.conan2/p/b/botanbfea1c417911d/p --os=linux --amalgamation --with-openssl --with-external-includedir=/root/.conan2/p/b/opense1670a4cbf775/p/include --with-external-libdir=/root/.conan2/p/b/opense1670a4cbf775/p/lib --without-pkg-config --without-os-feature=getrandom --without-os-feature=getentropy"
   INFO: Configuring to build Botan 2.19.3 (revision unknown)
   INFO: Running under 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]
   INFO: Autodetected platform information: OS="Linux" machine="aarch64" proc="aarch64"
   INFO: Using /etc/ssl/certs/ca-certificates.crt as system certificate store
   INFO: Canonicalized CPU target armv8 to arm64
   INFO: Auto-detected compiler version 11.4
   INFO: Auto-detected compiler arch arm64
   INFO: Target is gcc:11.4-linux-arm64
   INFO: Assuming target arm64 is little endian
  ERROR: Module not found: openssl
```

Can you take a look into that please?

Additionally, I've pushed some extra changes to improve the v2 compatibilty :)

recipes/botan/all/conanfile.py
RubenRBS: @uilianries is on the case with this one :)

RubenRBS (commented):
Thanks a lot for your contribution and patience! 

I've fixed some remaining issues with the v2 port, but I have a concern about the new 3.0.0 version not working properly when using `with_openssl` option (And maybe more!), please check :)

recipes/botan/all/conanfile.py
uilianries: I just created a PR to Botan: https://github.com/randombit/botan/pull/3685

The `getentropy` is indeed a new feature, not available as method, but also syscall.

I did some tests using the docker image `conanio/gcc5` with the follow code:

```c
#include <stdlib.h>
#include <stdio.h>
#include <unistd.h>

#include <sys/syscall.h>
#include <linux/random.h>


int main(int arg, char *argv[]) {
    char buffer[5];
    size_t length = 5;
    int result_getrandom = 0;
    int result_getentropy = 0;

#if defined(USE_SYSCALL)
    result_getrandom = syscall(SYS_getrandom, buffer, length, 0);
    result_getentropy = syscall(SYS_getentropy, buffer, length);
#else
    result_getrandom = getrandom(buffer, length, 0);
    result_getentropy = getentropy(buffer, length);
#endif

    if (result_getrandom == -1) {
        perror("getrandom");
        return EXIT_FAILURE;
    }
    if (result_getentropy == -1) {
        perror("getrandom");
        return EXIT_FAILURE;
    }

    return EXIT_SUCCESS;
}
```

When using syscall for getrandom, it works. The getentropy needs a newer glibc version.

recipes/botan/all/conanfile.py
securitykernel: The OpenSSL provider has been [removed](https://botan.randombit.net/news.html#version-2-19-2-2022-06-03) in Botan 2.19.2 due to being incompatible with OpenSSL 3.0.

recipes/botan/all/conanfile.py
```diff
@@ -117,82 +123,143 @@ def config_options(self):
...
-            if (compiler == 'apple-clang' and version < '10') or \
-               (compiler == 'gcc' and version < '8') or \
-               (compiler == 'clang' and version < '7'):
+        if Version(self.version) >= '2.14.0' and self.options.amalgamation:
+            if (self.settings.compiler == 'apple-clang' and compiler_version < '10') or \
+               (self.settings.compiler == 'gcc' and compiler_version < '8') or \
+               (self.settings.compiler == 'clang' and compiler_version < '7'):
                 raise ConanInvalidConfiguration(
-                    'botan amalgamation is not supported for {}/{}'.format(compiler, version))
+                    f"botan amalgamation is not supported for {compiler}/{compiler_version}")
```
lieser, line 214: In most cases you replaced the saved `compiler` with `self.settings.compiler`. Maybe use that here too, or use the `compiler_name`, and get rid of the saved `compiler`?

recipes/botan/all/conanfile.py
```diff
@@ -117,82 +123,147 @@ def config_options(self):
...
 
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
     def source(self):
-        tools.get(**self.conan_data['sources'][self.version], strip_root=True, destination=self._source_subfolder)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    @property
+    def _cxxflags(self):
```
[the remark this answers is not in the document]
lieser: `tools.build:sysroot` should now be forwarded to Botan's `--with-sysroot-dir` option. Admittedly did not test it yet, would be nice if you could report back if it works for you now.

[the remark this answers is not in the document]
fschoenm: Thanks, that was fast :) I hope I have time to try it out tomorrow.

recipes/botan/all/conanfile.py
uilianries: @RubenRBS @lieser Good news, my PR was merged into Botan, so you can apply it to the package and enable getrandom always: https://github.com/randombit/botan/pull/3685

recipes/botan/all/conanfile.py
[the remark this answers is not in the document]
fschoenm: This seemed to work, thanks again! 👍

securitykernel:
@RubenRBS @lieser Is this ready for merging?

reneme:
@securitykernel [This](https://github.com/conan-io/conan-center-index/pull/18079#discussion_r1314604929) still needs to be addressed. I'll do that now and I'll also add Botan 3.2.0 that was released yesterday.

recipes/botan/all/conanfile.py
```diff
@@ -133,72 +134,84 @@ def requirements(self):
         if self.options.with_bzip2:
             self.requires("bzip2/1.0.8")
         if self.options.with_openssl:
-            self.requires("openssl/1.1.1o")
+            self.requires("openssl/[>=1.1 <4]")
```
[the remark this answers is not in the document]
reneme: Done: https://github.com/conan-io/conan-center-index/pull/18079/commits/2d2b8a4f0901d09ad448f611ae4c94a80610adfb

Trafo (approved):

RubenRBS:
You'll need to pin a 1.1 release, as they dont use semver ordering and v1 did not handle that case for ranges

reneme:
> You'll need to pin a 1.1 release, as they dont use semver ordering and v1 did not handle that case for ranges

Thanks, done.

recipes/botan/all/conanfile.py
lieser: We probably want to remove the `build_flags.append('--without-os-feature=getrandom')` line (but not the one for `getentropy`), as we have patches to fix it for `getrandom` now.

securitykernel:
@RubenRBS Is there anything we should do here anymore to get this merged?

reneme:
@RubenRBS From my side, this is done. The force-push yesterday rebased to master and seems to have fixed [the remaining warnings](https://github.com/conan-io/conan-center-index/pull/18079#issuecomment-1758893311). There are two CI-related issues that I don't see an obvious reason for:

* `continuous-integration/jenkins/pr-merge` fails with "The build of this commit was aborted"
* the latest version Botan 3.2.0 does seem to be built in the CI (added in 646c21e97b49fcca358d8f64139cadcbcfa1d8dc)

Help and review/merge would be greatly appreciated. 😃

uilianries:
@lieser Thank you for your ping! Right there is a CI build running, but is taking longer than usual. It should show the build result in few hours.

reneme:
@uilianries @RubenRBS This should be finished and ready for prime time. I'm not sure why the CI didn't turn out green, so far.

RubenRBS:
Thanks for the ping @reneme - The cluster running our CI got some problems coming from Google's side so some PR jobs got lost, I've now restarted it, it should beiging to build the new commit again momentarily

reneme:
@uilianries @RubenRBS Thanks so much for being so incredibly responsive and helpful! This pull request is now (finally) done from our side and we'd be happy to have it reviewed and merged. 😄

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
+            return {
+                "gcc": "4.8",
+                "clang": "3.5",
+                "Visual Studio": "14",
+                "msvc": "190",
+            }
+        else:
+            # From https://github.com/randombit/botan/blob/master/doc/support.rst
+            return {
+                "gcc":  "11.2",
```
uilianries, line 172: I hope we can integrate new compilers in months, so we will be able to have these packages in Conan Center too.

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
 
     def requirements(self):
         if self.options.with_bzip2:
             self.requires("bzip2/1.0.8")
-        if self.options.with_openssl:
-            self.requires("openssl/1.1.1o")
+        if self.options.get_safe('with_openssl', False):
+            self.requires("openssl/1.1.1s")
```
uilianries, line 142:
```suggestion
            self.requires("openssl/[>=1.1 <4]")
```
OpenSSL and zlib can use range versions on CCI: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges

Hope it supports OpenSSL 3.x

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
         if self.options.with_zlib:
-            self.requires("zlib/1.2.12")
+            self.requires("zlib/1.2.13")
```
uilianries, line 144:
```suggestion
            self.requires("zlib/[>=1.2.11 <2]")
```
Same for zlib

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
         if self.options.with_sqlite3:
             self.requires("sqlite3/3.38.5")
         if self.options.with_boost:
-            self.requires("boost/1.79.0")
+            self.requires("boost/1.82.0")
```
uilianries, line 148:
```suggestion
            self.requires("boost/1.83.0")
```
The latest version available in Conan Center is 1.83.0. https://conan.io/center/recipes/boost

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
 
     @property
     def _required_boost_components(self):
         return ['coroutine', 'system']
 
+    @property
+    def _min_cppstd(self):
+        # From the same links as below
+        return 11 if Version(self.version).major < 3 else 20
```
uilianries, line 157:
```suggestion
        return 11 if Version(self.version).major < 3 else 20
```

Do not need to change, this is totally fine, only to know: `Version(self.version) < "3.0.0"` works too.

uilianries (commented):
Looking good, I only recommend updating some dependencies

recipes/botan/all/conanfile.py
lieser: No OpenSSL 3 is not supported by Botan

uilianries: @lieser thank you for your quick feedback. I just updated the pattern to not accept openssl 3. So we can move automatically in case having a newer version available.

reneme: Botan 2.x can (optionally) consume OpenSSL 1.x as a crypto provider. Support for the OpenSSL provider was removed in Botan 3.x. The `with_openssl` option is not available when requesting the Botan 3.x recipe.

Hence: This is intentional.

recipes/botan/all/conanfile.py
reneme:
```suggestion
        return 11 if Version(self.version) < "3.0.0" else 20
```

... self-suggesting for convenience.

recipes/botan/all/conanfile.py
reneme: Nevertheless, I applied your suggestion. Auto-updating won't hurt.

reneme:
@uilianries CI hanging again? :)

uilianries:
@reneme Sorry, it's seems be. I merge the master branch over this one to re-trigger the CI.

reneme:
@uilianries @RubenRBS This is done from our side and it [seems like it built successfully](https://github.com/conan-io/conan-center-index/pull/18079#issuecomment-1787976188). 🥳 

Let us know if there's anything more we can do.

reneme:
@uilianries @RubenRBS any blockers that we can help with? :)

uilianries (commented):
I can build on Mac OSX. Looking good, just a small question related to patches.

recipes/botan/all/conandata.yml
```diff
@@ -44,15 +14,32 @@ sources:
...
-      base_path: "sources"
   "2.18.2":
     - patch_file: "patches/fix-amalgamation-build.patch"
-      base_path: "sources"
+      patch_description: "Backport a fix for amalgamation build"
+      patch_type: "portability"
+  "3.0.0":
+    - patch_file: "patches/backport-getrandom-via-syscall-to-3.0.0.patch"
+      patch_description: "Backport a fix to support getrandom() with glibc < 2.25"
+      patch_type: "portability"
```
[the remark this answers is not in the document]
lieser: This backport is partly from one of the PRs you yourself created upstream 😁. It combines the merged PRs:
- https://github.com/randombit/botan/pull/3685
- https://github.com/randombit/botan/pull/3688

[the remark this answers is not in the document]
reneme: Please also note that the recipe applies these patches to Botan 3.0.0 and 3.1.x, only. Besides those two versions, this pull request also adds Botan 3.2.0, which already includes them in the upstream release.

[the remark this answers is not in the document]
reneme: @uilianries *poke, poke* 😅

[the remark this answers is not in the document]
reneme: That's a nice one! Thanks for pointing that out. Fixed now in: https://github.com/conan-io/conan-center-index/pull/18079/commits/cf6d35ef0a46e69c48f1f3674683f9adff146719 (along with changing the `patch_type` on the 2.18.2 release to "bugfix" as this is more suitable for the backported build fix).

uilianries:
The Conan 2.x is taking too long, I just restarted it right now. Sorry again about this delay.

uilianries:
@reneme I don't think so. It's more related number of builds waiting in the build quere vs number of nodes available. Building a package may take ours, need to consider that we build many configurations for each version. There is a relation listed here: https://github.com/conan-io/conan-center-index/blob/master/docs/supported_platforms_and_configurations.md#introduction

uilianries:
Sorry, I checked the build status and indeed something went wrong with Conan 2.x. It may take hours, but not days. Actually, there is a timeout (10h by default). Probably the build node disconnected for some reason, and Jenkins was waiting in an infinite loop 😞 

I'll restart it, but right now we are in the middle of a maintenance, so will be restarted as soon as we finish the maintenance. Sorry about this bad event.

lieser:
@uilianries Is the maintenance still ongoing or could you restart the CI?

uilianries:
@lieser I just restarted the build manually. The result should be available on https://ci-conan-prod.jfrog.team/job/prod-v2/job/cci/job/PR-18079/22/ as soon as it finishes.

lieser:
@uilianries Don't have access to the link you posted, and here on GitHub the *c3i/conan-v2/pr-merge* check is still red.
Would be really nice if this could be merged still this year. Maybe also a rebase would help?

securitykernel:
@uilianries This PR is now open almost half a year and quite some people want to use Botan 3.x from Conan Center. Can we speed up the merge process, i.e. merge now and fix whatever remains open on master as quick as possible? We would be there to help with anything after merging as quick as possible, just to get this PR merged now.

Otherwise, is there a person, group or institution I can escalate this to?

reneme:
Happy new year, everyone!

From my perspective, this is finished and ready for review. @RubenRBS, @jcar87, @uilianries we'd really appreciate to have this upstream. :)

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
+        copy(self, "license.txt", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        with chdir(self, self.source_folder):
+            # Note: this will fail to properly consider the package_folder if a "conan build" followed by a "conan export-pkg" is executed
             self.run(self._make_install_cmd)
+        if is_apple_os(self):
+            # fix rpath to make installed library relocatable (CCI KB-H077)
+            with chdir(self, os.path.join(self.package_folder, "lib")):
+                for dylib in glob.glob("*.dylib"):
+                    if not os.path.islink(dylib):
+                        self.run(f"install_name_tool -id @rpath/{os.path.basename(dylib)} {dylib}")
```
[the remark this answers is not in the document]
reneme: Nice! Thanks for pointing that out.

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +124,155 @@ def config_options(self):
...
+        with chdir(self, self.source_folder):
             self.run(self._configure_cmd)
             self.run(self._make_cmd)
 
     def package(self):
-        self.copy(pattern='license.txt', dst='licenses', src=self._source_subfolder)
-        with tools.chdir(self._source_subfolder):
+        copy(self, "license.txt", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        with chdir(self, self.source_folder):
+            # Note: this will fail to properly consider the package_folder if a "conan build" followed by a "conan export-pkg" is executed
```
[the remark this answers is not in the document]
reneme: That's just a remark, right? No need to change anything here?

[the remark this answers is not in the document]
uilianries: Correct, just a note 😄

reneme:
@uilianries I think, I addressed your recent comments and the pipeline is green. I'd appreciate another look. :)

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +125,148 @@ def config_options(self):
...
+                "clang": "14",
+                "apple-clang": "14",
+                "Visual Studio": "17",
+                "msvc": "193",
+            }
+
     def validate(self):
         if self.options.with_boost:
             miss_boost_required_comp = any(getattr(self.options['boost'], 'without_{}'.format(boost_comp), True) for boost_comp in self._required_boost_components)
             if self.options['boost'].header_only or self.options['boost'].shared or self.options['boost'].magic_autolink or miss_boost_required_comp:
```
uilianries, line 183: I tried to build locally and it does not work when boost is required:

```
$ conan create all --version=3.2.0 -s compiler.cppstd=20 --build=missing -pr gcc12.profile 

======== Exporting recipe to the cache ========
botan/3.2.0: Exporting package recipe: /home/conan/project/all/conanfile.py
botan/3.2.0: exports: File 'conandata.yml' found. Exporting it...
botan/3.2.0: Calling export_sources()
botan/3.2.0: Copied 1 '.yml' file: conandata.yml
botan/3.2.0: Copied 1 '.py' file: conanfile.py
botan/3.2.0: Exported to cache folder: /home/conan/.conan2/p/botan759963310d11c/e
botan/3.2.0: Exported: botan/3.2.0#9b62638872488502e0160f97996e959f (2024-01-15 15:43:57 UTC)

Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
[options]
botan/3.2.0:with_boost=True
botan/3.2.0:with_bzip2=True
botan/3.2.0:with_sqlite3=True
botan/3.2.0:with_zlib=True

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux


bzip2/1.0.8: Not found in local cache, looking in remotes...
bzip2/1.0.8: Checking remote: conancenter
bzip2/1.0.8: Downloaded recipe revision 457c272f7da34cb9c67456dd217d36c4
zlib/1.3: Not found in local cache, looking in remotes...
zlib/1.3: Checking remote: conancenter
zlib/1.3: Downloaded recipe revision 06023034579559bb64357db3a53f88a4
sqlite3/3.38.5: Not found in local cache, looking in remotes...
sqlite3/3.38.5: Checking remote: conancenter
sqlite3/3.38.5: Downloaded recipe revision 4b875d4249cdfb4c1235e6b3ea6c18e7
boost/1.83.0: Not found in local cache, looking in remotes...
boost/1.83.0: Checking remote: conancenter
boost/1.83.0: Downloaded recipe revision 658a6b1d7aa62a019cbba3f4ae363cb7
libbacktrace/cci.20210118: Not found in local cache, looking in remotes...
libbacktrace/cci.20210118: Checking remote: conancenter
libbacktrace/cci.20210118: Downloaded recipe revision ec1aa63bbc10145c6a299e68e711670c
b2/4.10.1: Not found in local cache, looking in remotes...
b2/4.10.1: Checking remote: conancenter
b2/4.10.1: Downloaded recipe revision 1b290403d8648c79f468f5a6496f829a
Graph root
    cli
Requirements
    boost/1.83.0#658a6b1d7aa62a019cbba3f4ae363cb7 - Downloaded (conancenter)
    botan/3.2.0#9b62638872488502e0160f97996e959f - Cache
    bzip2/1.0.8#457c272f7da34cb9c67456dd217d36c4 - Downloaded (conancenter)
    libbacktrace/cci.20210118#ec1aa63bbc10145c6a299e68e711670c - Downloaded (conancenter)
    sqlite3/3.38.5#4b875d4249cdfb4c1235e6b3ea6c18e7 - Downloaded (conancenter)
    zlib/1.3#06023034579559bb64357db3a53f88a4 - Downloaded (conancenter)
Build requirements
    b2/4.10.1#1b290403d8648c79f468f5a6496f829a - Downloaded (conancenter)
Resolved version ranges
    zlib/[>=1.2.11 <2]: zlib/1.3

boost/1.83.0: Checking 9 compatible configurations
boost/1.83.0: Compatible configurations not found in cache, checking servers
boost/1.83.0: '1f4d8d9bfb906fd36e9471d4ce2424cf0278c2ce': compiler.cppstd=11
boost/1.83.0: 'f331bab124723d370f6d6c8b733ba3ff9df40985': compiler.cppstd=gnu11
boost/1.83.0: '0444dbb02c7c03e0924add655328614c4069992c': compiler.cppstd=14
boost/1.83.0: '53fadd80e75df07d899e7a29532a2fc59cd7ab23': compiler.cppstd=gnu14
boost/1.83.0: '21c1ffdf955757b3eaf1428692e72e91c15d8446': compiler.cppstd=17
boost/1.83.0: 'ab3b66987b0ff007c45a5b90708ec0b68208653b': compiler.cppstd=gnu17
boost/1.83.0: 'b7691adcb83380a1f2a05c4982136fd2dcb71ab7': compiler.cppstd=gnu20
boost/1.83.0: 'b4575a8b8adc07f0f83596aa71fc38363b522735': compiler.cppstd=23
boost/1.83.0: '1d726159589d984732e608f539092dd02c6b8583': compiler.cppstd=gnu23
ERROR: botan/3.2.0: Error in validate() method, line 181
	miss_boost_required_comp = any(getattr(self.options['boost'], 'without_{}'.format(boost_comp), True) for boost_comp in self._required_boost_components)
while calling '<genexpr>', line 181
	miss_boost_required_comp = any(getattr(self.options['boost'], 'without_{}'.format(boost_comp), True) for boost_comp in self._required_boost_components)
	ConanException: option 'without_coroutine' doesn't exist
Possible options are []

```

So, doing a quick check, in Conan 2.x you should use `self.dependencies["boost"].options` instead. Could you please update and check it?

uilianries (commented):
Some notes:

- Tested locally with the docker image conanio/gcc11-ubuntu16.04 and it does not work because we have 11.1 installed
- Tested locally with the docker image conanio/gcc12-ubuntu16.04 and it does not work because it requires glibc >= 2.25 because it uses `getentropy`
- Tested locally with the docker image conanio/gcc12-ubuntu18.04 with Conan 2.0.17 and worked for static library build.

reneme:
Thanks, @uilianries for spotting this issue. I was able to reproduce and fix it with your suggestion. :+1:

Regarding the build on `conanio/gcc12-ubuntu16.04`: Botan expects certain features on Linux by default. Most notably: `getrandom` and `explicit_bzero`. Both could be disabled by passing `--without-os-features=getrandom,explicit_bzero` to Botan's `./configure.py`. However, the conan recipe currently does not expose this configuration flag. Nevertheless, I hacked this up locally and confirmed that it would compile.

There are many more switches in Botan's `./configure.py` that are not exposed in the recipe, for simplicity. We could, however, add a generic option `-o additional_configure_flags=...` (name is debatable). That would allow users to circumvent the recipe and directly manipulate their build configuration. Frankly, I'm not sure that is something we would want to do, though... What do you think?

uilianries:
@lieser Thank you for checking. It's a shame we still can run a newer docker image with a newer compiler to solve this problem. I would prefer not building a package, but distributing only the recipe in case not possible right in CCI. I see some problems like, Botan is going forward with new features, and we are trying to backport things, which makes even harder to maintain this recipe. 

About the option, you have a good observation about using the option, but I'm no inclined to accept a so generic feature, because it would result in a fragility not tested by us. Like, users starts to use configuration that does not work and report us, but in the end, it's a bug in the upstream. Maybe, using the [conan configuration](https://docs.conan.io/2/reference/config_files/global_conf.html ) `tools.build:cxxflags` could be an option, but only for local building, because right now we don't support custom config per recipe (yet, but we should in months).

My current suggestion is checking glibc version (it could be `platform.libc_ver()`) and raise ConanInvalidConfiguration if it does not fit. Sad, but not first case in CCI. Hope we can update those images in some months, using Ubuntu 20.04 or 22.04 (16.04 is close to be EOL)

reneme:
> My current suggestion is checking glibc version (it could be platform.libc_ver()) and raise ConanInvalidConfiguration if it does not fit.

Alternatively, we could disable certain features via Botan's `./configure.py --disable-os-features=` depending on what version is reported by `platform.libc_ver()`. I.e. instead of the generic `-o additional_configure_flags=`, the recipe would be smart enough to disable things as needed automatically.

Though, that would mean that consumers of the binaries compiled in your CI won't benefit from those features, either. Right, @uilianries? But, given that you are planning to update the build images soon, we could remove this workaround again soon.

reneme:
As discussed yesterday, I added [a workaround that allows compiling Botan 3.x with a glibc older than 2.25](https://github.com/conan-io/conan-center-index/pull/18079/commits/6c5240234984f3cfdee6158eee5917eedc714832). This disables certain OS features that the library just takes for granted starting in 3.x.

If the workaround is triggered, it prints a `self.output.warning()`. I hope CCI doesn't build with `core:warnings_as_errors`, @uilianries?

recipes/botan/all/conanfile.py
```diff
@@ -117,84 +128,149 @@ def config_options(self):
...
     def source(self):
-        tools.get(**self.conan_data['sources'][self.version], strip_root=True, destination=self._source_subfolder)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    @property
+    def _cxxflags(self):
+        global_cxxflags = " ".join(self.conf.get("tools.build:cxxflags", default=[], check_type=list))
+        env_cxxflags = VirtualBuildEnv(self).vars().get("CXXFLAGS", default="")
+        cxxflags = f"{env_cxxflags} {global_cxxflags}".strip()
+        return cxxflags if len(cxxflags) > 0 else None
```
uilianries, line 240:
```suggestion
        return cxxflags if len(cxxflags) > 1 else None
```

because cxxflags will contain at least an empty-space always.

recipes/botan/all/conanfile.py
```diff
@@ -375,11 +447,17 @@ def _configure_cmd(self):
         if is_msvc(self):
             build_flags.append(f"--msvc-runtime={msvc_runtime_flag(self)}")
 
+        if self._is_glibc_older_than_2_25_on_linux and Version(self.version) >= '3.0':
+            # Botan 3.0+ requires glibc >= 2.25, but CCI's CI Linux images are
+            # currently older than that. Remove as soon as CCI updates!
```
uilianries, line 452:
```suggestion
            # INFO: Botan 3.0+ requires glibc >= 2.25. Disable features to make it backward compatible
            # FIXME: CCI Docker images are running Ubuntu 16.04. Remove it after supporting later version.
```

Make it more explicit, so code editor will blink any FIXME in this repository.

recipes/botan/all/conanfile.py
```diff
@@ -445,3 +519,19 @@ def _is_linux_clang_libcxx(self):
             self.settings.compiler == 'clang' and
             self.settings.compiler.libcxx == 'libc++'
         )
+
+    @property
+    def _is_glibc_older_than_2_25_on_linux(self):
+        # glibc below 2.25 lacks support for certain syscalls that botan assumes
```
uilianries, line 525:
```suggestion
        # FIXME: glibc below 2.25 lacks support for certain syscalls that botan assumes
```

uilianries (commented):
Looking good, only few comments. I'll build locally with Boost to check how is working.

uilianries:
It's working on Mac: https://gist.github.com/uilianries/399359d2327602bcfe8ea5ed4183f0ab 🟢

uilianries (approved):
LGTM

recipes/botan/all/conanfile.py
reneme: Good to know!

reneme:
@RubenRBS @jcar87 Could we get your review on this, please? 😃

danimtb (approved):

reneme:
# 🥳 


Thank you all for the effort that went into this!

PR #18083, opened 2023-06-27
rapidcsv: add version 8.77, add package_type

toge:
Specify library name and version:  **rapidcsv/***

---


RubenRBS (approved):
If when you add new versions for new versions you can try to remove some old ones when the library has >5 versions, we (and our CI!) would appreciate it :)

Either way this is good, thanks a lot (again)

PR #18088, opened 2023-06-27
libtool: fix windows x86 build 

maksim-petukhov:
Specify library name and version:  **libtool/\***

The same problem as in #18086

Honorable mention: https://github.com/conan-io/conan/issues/12546 & https://github.com/conan-io/conan/issues/7460

---

### DIFF
--- a/recipes/libtool/all/conanfile.py
+++ b/recipes/libtool/all/conanfile.py
@@ -1,6 +1,7 @@
 from conan import ConanFile
 from conan.errors import ConanException
 from conan.tools.apple import is_apple_os, fix_apple_shared_install_name
+from conan.tools.build import cross_building
 from conan.tools.env import Environment
 from conan.tools.files import apply_conandata_patches, export_conandata_patches, copy, get, rename, replace_in_file, rmdir
 from conan.tools.gnu import Autotools, AutotoolsToolchain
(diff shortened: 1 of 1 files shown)


recipes/libtool/all/conanfile.py
```diff
@@ -92,6 +93,20 @@ def generate(self):
 
         tc = AutotoolsToolchain(self)
 
+        if cross_building(self) and is_msvc(self):
```
maksim-petukhov, line 96:
```suggestion
        if cross_building(self) and (is_msvc(self) or self.settings.compiler == "clang"):
```
Maybe it should be like that?

recipes/libtool/all/conanfile.py
```diff
@@ -92,6 +93,20 @@ def generate(self):
...
+            triplet_arch_windows = {"x86_64": "x86_64", "x86": "i686", "armv8": "aarch64"}
+
+            host_arch = triplet_arch_windows.get(str(self.settings.arch))
+            build_arch = triplet_arch_windows.get(str(self._settings_build.arch))
+
+            if host_arch and build_arch:
+                host = f"{host_arch}-w64-mingw32"
+                build = f"{build_arch}-w64-mingw32"
+                tc.configure_args.extend([
```
maksim-petukhov, line 105: Would it be better to use `tc.update_configure_args` or extend is fine?

uilianries:
@maksim-petukhov Thank you for your contribution. As you can read in [supported configurations](https://github.com/conan-io/conan-center-index/blob/master/docs/supported_platforms_and_configurations.md) document, the CCI does not build x86, so we can't validate your changes. Could you please share with us a full log of your build on Windows x86, so we could some prof that's have your desired effect?

PR #18089, opened 2023-06-27
b2: update 'standard' recipe to match the newer 'portable'

valgur:
TODO

- [ ] `Empty package`


recipes/b2/all/conanfile.py
```diff
@@ -13,62 +14,61 @@
...
-
-    Specifies the toolset to use for building. The default of 'auto' detects
-    a usable compiler for building and should be preferred. The 'cxx' toolset
-    uses the 'CXX' and 'CXXFLAGS' solely for building. Using the 'cxx'
-    toolset will also turn on the 'use_cxx_env' option. And the 'cross-cxx'
-    toolset uses the 'BUILD_CXX' and 'BUILD_CXXFLAGS' vars. This frees the
-    'CXX' and 'CXXFLAGS' variables for use in subprocesses.
-    '''
+    package_type = "application"
+    settings = "os", "arch", "compiler", "build_type"
```
grafikrobot, line 24: The current recipe intentionally does not include the compiler and build_type as those are irrelevant for running the produced b2 executable. Why are you adding them?

grafikrobot (commented):
AFAICT the changes you made are:

1. A bunch of style only changes. Why?
2. Adding compiler and build_type to settings. Why?
3. Using option_descriptions. Great!
4. Clearing of some cpp_info vars. Great!
5. Putting the old and new build code into one recipe by adding the _build_non_portable function and version check. Why?

I'm wondering why you did (1) as it just makes the history harder to follow.

I'm wondering why you did (2). As mentioned in the other comment.. This goes against an intentional choice.

Last, (5). I really don't understand the goal of this PR. Why complicate the new build and install mode with the old mode? Is there some use case or problem you are trying to fix?

valgur:
> 1. A bunch of style only changes. Why?

Sorry, these were incidental and not really relevant. Reverted them.

> 2. Adding compiler and build_type to settings. Why?

Added these to match the current recommendation to always list all values and then remove the irrelevant settings in `package_id()`: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conanfile_attributes.md#settings

> 5. Putting the old and new build code into one recipe by adding the _build_non_portable function and version check. Why?
> Last, (5). I really don't understand the goal of this PR. Why complicate the new build and install mode with the old mode? Is there some use case or problem you are trying to fix?

The intention behind this PR was (1) to eliminate unnecessary dead code in the repo by updating the `standard` recipe and really (2) to personally get some practice on how the environment variables should be handled in Conan v2 before starting work on migrating some of the more complex Autotools recipes.

I originally did not intend to modify the `portable` recipe at all, but the differences between the two recipes appeared to be quite small, so it made sense to merge them instead.

Anyway, I reverted the merge of the `standard` and `portable` recipes now and only kept minimal changes to the `portable` one. The `standard` recipe is now identical to `portable` with only the `build()` method differing. A minor benefit of keeping them separate is reducing the number of versions to build when the newer one gets updated, I suppose, since the list of versions is quite long.

valgur:

PR #18105, opened 2023-06-28
libpng: simplify patch logic

jcar87:
Specify library name and version:  **libpng/all**

- Replace the patches for versions 1.6.37 and 1.5.30 with a cmake project include instead - fewer lines and less duplication. Document why this is needed for those older versions.
- Remove `copy_if_different` patches via `replace_in_file` - after extensive testing, this patch does not seem to affect which files end in the package, or where inside the package they end up in, so removing this under the belief that the patch no longer has any observable effects whatsoever when os=Windows and compiler=gcc. If an issue is reported in the future and it is not considered to be a defect in the libpng buildscripts, we can address it.
- Very minor change: use `cache_variables` for libpng specific toggles, rather than defining this in the toolchain file, in line with CMake's guidelines and Conan 2.0 [documentation](https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#variables)


RubenRBS (approved):
👏

PR #18119, opened 2023-06-28
debug_assert: migrate to Conan v2

valgur:


czoido (approved):

franramirez688 (approved):

recipes/debug_assert/all/test_package/CMakeLists.txt
```diff
@@ -1,9 +1,8 @@
...
+project(test_package LANGUAGES CXX)
 
-include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-conan_basic_setup()
+find_package(debug_assert REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
-target_link_libraries(${PROJECT_NAME} ${CONAN_LIBS})
+target_link_libraries(${PROJECT_NAME} PRIVATE debug_assert::debug_assert)
 set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 11)
```
czoido, line 8: Same comment than here: https://github.com/conan-io/conan-center-index/pull/18124#discussion_r1246361883

czoido (dismissed):
Thanks for contributing!

czoido (approved):

recipes/debug_assert/all/test_package/CMakeLists.txt
valgur, line 8: Should I keep the `CXX_STANDARD` setting in test_package if `check_min_cppstd()` is used in the conanfile.py?
I suppose it should only be necessary if the standard needs to be lower than the compiler's default or the one set by `compiler.cppstd`?

CJCombrink (approved):

recipes/debug_assert/all/test_package/CMakeLists.txt
czoido: Yes, I think it's better keep it because we don't pass the cppstd to v1 builds, so it would fail

czoido (approved):

jwillikers (approved):

PR #18127, opened 2023-06-28
nextsilicon-cpp-subprocess: migrate to Conan v2

valgur:


czoido (approved):

RubenRBS (approved):

recipes/nextsilicon-cpp-subprocess/all/conanfile.py
```diff
@@ -1,27 +1,50 @@
...
-                   "The only goal was to develop something that is as close as"
-                   "python subprocess module in dealing with processes.")
-    no_copy_source = True
+    topics = ("subprocess", "os", "fork", "header-only")
+
+    package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
+    no_copy_source = True
+
+
```
franramirez688, line 26: I saw that the test_package is adding `set(CMAKE_CXX_STANDARD 11)`. I think that we should add a minimal check. Something like:

```python
    @property
    def _min_cppstd(self):
        return "11"

    def validate(self):
```

recipes/nextsilicon-cpp-subprocess/all/test_package/CMakeLists.txt
```diff
@@ -1,14 +1,12 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.15)
 project(test_package CXX)
 
-include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-conan_basic_setup()
-
 set(CMAKE_CXX_STANDARD 11)
```
franramirez688, line 4: Adding that check above, we could remove this one. Conan puts automatically this one through the CMakeToolchain generator.

recipes/nextsilicon-cpp-subprocess/all/conanfile.py
franramirez688: Let's not forget to add the proper import:

```python
```

PR #18128, opened 2023-06-28
effolkronium-random: migrate to Conan v2

valgur:


recipes/effolkronium-random/all/conanfile.py
```diff
@@ -1,31 +1,49 @@
...
+            src=os.path.join(self.source_folder, "include"),
+        )
 
     def package_info(self):
+        self.cpp_info.bindirs = []
+        self.cpp_info.libdirs = []
+
+        self.cpp_info.set_property("cmake_file_name", "effolkronium_random")
+        self.cpp_info.set_property("cmake_target_name", "effolkronium_random::effolkronium_random")
+
```
czoido, line 46: Checking upstream: https://github.com/effolkronium/random/blob/master/CMakeLists.txt I think the correct target name is `effolkronium_random` not `effolkronium_random::effolkronium_random`

valgur: Indeed: https://github.com/effolkronium/random/tree/master#cmake
Thanks! Fixed it.

recipes/effolkronium-random/all/test_package/CMakeLists.txt
```diff
@@ -1,9 +1,7 @@
...
+project(test_package LANGUAGES CXX)
 
-include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
-conan_basic_setup()
+find_package(effolkronium_random REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
-target_link_libraries(${PROJECT_NAME} ${CONAN_LIBS})
-set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 11)
+target_link_libraries(${PROJECT_NAME} PRIVATE effolkronium_random)
```
czoido, line 7: I think the fail in the ci is because this is missing:
```suggestion
target_link_libraries(${PROJECT_NAME} PRIVATE effolkronium_random)
target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_11)
```

recipes/effolkronium-random/all/test_package/CMakeLists.txt
```diff
@@ -1,9 +1,8 @@
...
+target_compile_features(${PROJECT_NAME} PUBLIC cxx_std_11)
```
czoido, line 8:
```suggestion
set_property(TARGET ${PROJECT_NAME} PROPERTY CXX_STANDARD 11)
```

czoido (approved):

czoido:
The problem here is that we have a target name for v1 that is `effolkronium_random::effolkronium_random` for v1 and just `effolkronium_random` for v2, if we want to align both target names we would have to do the ugly workaround with components (declare a `effolkronium_random` fictitious component and setting names for that component to `effolkronium_random`), but I think it would be better not to add that to the recipe and just use a different CMakeLists.txt for test_v1 (we will remove all those folders at some point in the future anyway)

czoido:
@valgur having a second look, I think we can drop the test_v1_package from this PR, we are anyway allowing removing those soon.

czoido (approved):

davidsanfal (approved):

PR #18138, opened 2023-06-29
#18137 Fix compilation of breakpad with gcc-11

### LINKED ISSUE #18137: [package] breakpad/cci.20210521: Compilation error with gcc 11
### Description

Fails to compile with gcc 11 on Ubuntu 22.04:
```cxx
exception_handler.cc:141:49: error: no matching function for call to ‘max(int, long int)’
polaris-jammy                |   141 |   static const unsigned kSigStackSize = std::max(16384, SIGSTKSZ);
polaris-jammy                |       |                                         ~~~~~~~~^~~~~~~~~~~~~~~~~

```

* Package Name/Version: **breakpad/cci.20210521**
* Operating System+version: **Linux Ubuntu 22.04**
* Compiler+version: **GCC 11**

### Conan profile

[settings]
os=Linux
arch=x64
compiler=gcc
compiler.version=111
compiler.libcxx=libstdc++11
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install -r conancenter breakpad/cci.20210521

### Logs

-

artem-ogre:
Specify library name and version:  **breakpad/cci.20210521**

Fixes #18137

### DIFF
--- a/recipes/breakpad/all/conanfile.py
+++ b/recipes/breakpad/all/conanfile.py
@@ -31,7 +31,7 @@ class BreakpadConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("linux-syscall-support/cci.20200813")
+        self.requires("linux-syscall-support/cci.20200813", transitive_headers=True)
 
     def validate(self):
         if self.settings.os != "Linux":
--- a/recipes/breakpad/all/conandata.yml
+++ b/recipes/breakpad/all/conandata.yml
@@ -6,3 +6,4 @@ patches:
   "cci.20210521":
     - patch_file: "patches/0001-Use_conans_lss.patch"
     - patch_file: "patches/0002-Remove-hardcoded-fpic.patch"
+    - patch_file: "patches/0003-Fix-gcc11-compilation.patch"
--- a/recipes/breakpad/all/patches/0003-Fix-gcc11-compilation.patch
+++ b/recipes/breakpad/all/patches/0003-Fix-gcc11-compilation.patch
@@ -1,0 +1,11 @@
+--- a/src/client/linux/handler/exception_handler.cc
++++ b/src/client/linux/handler/exception_handler.cc
+@@ -138,7 +138,7 @@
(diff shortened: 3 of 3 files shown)


RubenRBS:
Hi @artem-ogre, thanks a lot for your contribution. Has this patch been upstreamed in any way? If it's an official patch getting backported we usually find those easier to accept.

I also see that the available version in CCI is a bit old now, so maybe if it's already available the better approach would be to add a new release of the library, which contains the fix? Let me know what you think :)

artem-ogre:
@RubenRBS 
Hi, Ruben. Yes, that's a good point. I checked upstream and it is [patched there](https://chromium.googlesource.com/breakpad/breakpad/+/refs/heads/main/src/client/linux/handler/exception_handler.cc#144).

I updated the PR to back-port this patch verbatim.

artem-ogre:
Unfortunately I have no capacity for updating breakpad to the newer version.

RubenRBS:
> Unfortunately I have no capacity for updating breakpad to the newer version.

No worries about that! It's totally understandable!

artem-ogre:
This is really a trivial one-line patch (backported from upstream), is it possible to merge this sooner rather than later?

RubenRBS:
Hi, yes you are right, thanks for your patience while we processed the rest of the PR pipeline :)

RubenRBS (dismissed):

recipes/breakpad/all/conanfile.py
```diff
@@ -31,7 +31,7 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("linux-syscall-support/cci.20200813")
+        self.requires("linux-syscall-support/cci.20200813", transitive_headers=True)
```
czoido, line 34: Thanks for your patience @artem-ogre, I'm adding this change to take the chance to make it v2 compatible. I think it's worth it.

RubenRBS (approved):

czoido (approved):

PR #18147, opened 2023-06-29
Add embree4 package

jjcasmar:
Specify library name and version:  **embree/4.1.0**

Add embree4.  Since there is a folder for embree3, I decided to create embree4 folder. Maybe we should consider merging embree3 and embree4 recipes though.

The recipe should be compatible with linux and emscripten, but I haven't tested windows or mac platforms, as I dont have any sistem with any of those. 

---


jjcasmar:
Not sure why it fails for v1. I would probably need some help with that

PR #18153, opened 2023-06-30
st_tree: support conan v2

toge:
Specify library name and version:  **st_tree/***

---

### DIFF
--- a/recipes/st_tree/all/conanfile.py
+++ b/recipes/st_tree/all/conanfile.py
@@ -1,42 +1,58 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import get, copy, rmdir
+from conan.tools.layout import basic_layout
+from conan.tools.cmake import CMake, CMakeToolchain
 import os
 
-from conans.errors import ConanInvalidConfiguration
-from conans import ConanFile, CMake, tools
 
-required_conan_version = ">=1.33.0"
+required_conan_version = ">=1.52.0"
 
 class STTreeConan(ConanFile):
     name = "st_tree"
+    description = "A fast and flexible c++ template class for tree data structures"
     license = "Apache-2.0"
     url = "https://github.com/conan-io/conan-center-index"
-    description = "A fast and flexible c++ template class for tree data structures"
-    topics = ("stl", "container", "data-structures")
     homepage = "https://github.com/erikerlandson/st_tree"
+    topics = ("stl", "container", "data-structures", "tree", "header-only")
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
     no_copy_source = True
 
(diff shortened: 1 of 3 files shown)


recipes/st_tree/all/conanfile.py
```diff
@@ -1,42 +1,58 @@
...
+            check_min_cppstd(self, self._min_cppstd)
 
     def source(self):
-        tools.get(**self.conan_data["sources"][self.version], destination=self._source_subfolder, strip_root=True)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
-    def _configure_cmake(self):
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["ENABLE_TESTS"] = False
```
CJCombrink, line 41: Doesn't [cache_variables](https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#cache-variables) make more sense here?
```
tc.cache_variables["ENABLE_TESTS"] = False
```

CJCombrink (approved):

recipes/st_tree/all/conanfile.py
```diff
@@ -1,42 +1,58 @@
...
         cmake.install()
 
-        tools.rmdir(os.path.join(self.package_folder, "lib"))
-
-    def package_id(self):
-        self.info.header_only()
+        rmdir(self, os.path.join(self.package_folder, "lib"))
 
     def package_info(self):
-        self.cpp_info.filenames["cmake_find_package"] = "st_tree"
```
CJCombrink, line 41: Is this not needed anymore? Will they have proper defaults when removed?

I am just curious, the [Migrating legacy cpp_info attributes to set_property()](https://docs.conan.io/en/1.53/migrating_to_2.0/properties.html) does mention how to migrate these but does not mention removing is a valid migration option for Conan 2
(PS: I am looking at different PRs to understand how to migrate our packages once we start doing it)

CJCombrink (approved):

recipes/st_tree/all/conanfile.py
czoido: Hi @CJCombrink,
We are leaving those for some time to not break compatibility with v1 users that use the legacy generators, but it will be removed from all recipes soon. If you are using v1 with CMakeToolchain, CMakeDeps, and the other generators compatible with v2 you can remove those from your recipes.
Also note that adding test_v1_package for new recipes is not mandatory any more.

uilianries (approved):
@toge I simplified a bit your recipe. I checked the upstream's CMake and it does not deserve running all steps, it only copies headers files to `include` folder. The only plus is running tests and installing, but as we don't run unit tests, it would much better just copying headers to package folder directly.

SSE4 (approved):

PR #18166, opened 2023-06-30
aaf: add shared and fPIC, add package_type

valgur:


recipes/aaf/all/conanfile.py
```diff
@@ -10,26 +10,39 @@
...
     def export_sources(self):
         export_conandata_patches(self)
 
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
```
RubenRBS, line 44: `1.53` again :)

RubenRBS (commented):
Some build issues in Windows, let us know if you need help sorting them out!

jcar87 (changes requested):
Hi @valgur - thanks for your contribution.

I have a strong suspicion that this should be a `static-library`, please see the motivation in this PR: https://github.com/conan-io/conan-center-index/pull/18385

Evidence:
* The library's own SDK, which mention that consumers need to link against two static libraries (https://aaf.sourceforge.net/AAFProjectFAQ.html)
* The original build scripts, which:
   * Only package the `.a` files in the `.lib` folder: https://github.com/jhliberty/AAF/blob/master/dist/sdk-files.unix#L8-L14
   * Do not package the `.lib` files for the DLLs on windows: https://github.com/jhliberty/AAF/blob/master/dist/sdk-files.win#L18-L19
   
 This would strongly indicate that for the purposes of building, this is a static library (as it is referred explicitly in the documentation and the build scripts), and the the DLLs are merely runtime loaded modules.

MartinDelille (approved):

valgur:
Closing in favor of #18385.

PR #18167, opened 2023-06-30
access_private: add package_type

valgur:


RubenRBS:
@valgur thanks a lot for your contributions, but this seems to have an issue: We're not currently allowing the removal of `test_v1_package`s to recipes that already have them, only to new ones.
I see that you have open lots of PRs (It's going to take us a while to process all of them, give us some time!) and by quickly glancing at some of them, they also have those files removed, which is going to be a blocker for most of them, even the rest of the changes look good :/

We can revert the deletions manually, but maybe doing it on your end is better? Just keep in mind not to trigger the changes all at once so the CI systems are able to cope with the rest of the normal traffic :)

valgur:
Oh, my bad. Sorry about that. I will fix this in batch on my end asap. I'll take care to not overload the CI.

Also, regarding the PRs - I won't mind in the least if you drop them onto the bottom of the review queue. I will be away for the next two weeks and won't be able work on any change requests in the PRs for that duration anyway.

RubenRBS (approved):

czoido (approved):

PR #18175, opened 2023-06-30
android-ndk: add package_type, replace exports_sources

valgur:


RubenRBS (approved):
Thanks! Curious: Is the move to `export_soruces(self)` just a personal preference?

czoido (approved):

valgur:
> Is the move to `export_soruces(self)` just a personal preference? 

I suppose so. It wasn't strictly necessary. It's mostly based on the minor preference towards the method mentioned in the paciage template comments. I don't think it adds much if patches are not being exported, though.

PR #18176, opened 2023-06-30
angelscript: add package_type

valgur:


recipes/angelscript/all/conanfile.py
```diff
@@ -42,10 +43,7 @@ def config_options(self):
 
     def configure(self):
         if self.options.shared:
-            try:
-                del self.options.fPIC
-            except Exception:
-                pass
+            self.options.rm_safe("fPIC")
```
RubenRBS, line 46: This is a 1.53 feature, you should bump `required_conan_version` too :)

RubenRBS (changes requested):

RubenRBS (approved):

PR #18187, opened 2023-06-30
armadillo: add package_type

valgur:


recipes/armadillo/all/conanfile.py
```diff
@@ -25,7 +28,9 @@ class ArmadilloConan(ConanFile):
         "mkl",
         "hdf5",
     )
-    settings = "os", "compiler", "build_type", "arch"
+
+    package_type = "library"
```
samuel-emrys, line 32: Unnecessary as this was added in https://github.com/conan-io/conan-center-index/pull/17320

samuel-emrys (changes requested):

PR #18192, opened 2023-06-30
atk: add package_type, use is_msvc()

valgur:
TODO

- [ ] `../src/atk/meson.build:90:21: ERROR: Program 'glib-genmarshal' not found or not executable`

### DIFF
--- a/recipes/atk/all/conanfile.py
+++ b/recipes/atk/all/conanfile.py
@@ -50,7 +50,7 @@ class AtkConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("glib/2.76.3")
+        self.requires("glib/2.77.1")
 
     def validate(self):
         if self.options.shared and not self.dependencies["glib"].options.shared:
@@ -69,11 +69,11 @@ class AtkConan(ConanFile):
             )
 
     def build_requirements(self):
-        self.tool_requires("meson/1.1.1")
+        self.tool_requires("meson/1.2.0")
         if not self.conf.get("tools.gnu:pkg_config", check_type=str):
-            self.tool_requires("pkgconf/1.9.3")
+            self.tool_requires("pkgconf/1.9.5")
         if hasattr(self, "settings_build") and cross_building(self):
-            self.tool_requires("glib/2.76.3")
+            self.tool_requires("glib/2.77.1")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
--- a/recipes/atk/all/conanfile.py
+++ b/recipes/atk/all/conanfile.py
@@ -7,7 +7,7 @@ from conan.tools.files import apply_conandata_patches, copy, export_conandata_pa
 from conan.tools.gnu import PkgConfigDeps
 from conan.tools.layout import basic_layout
 from conan.tools.meson import Meson, MesonToolchain
-from conan.tools.microsoft import is_msvc_static_runtime
+from conan.tools.microsoft import is_msvc_static_runtime, is_msvc
 import os
 
 required_conan_version = ">=1.53.0"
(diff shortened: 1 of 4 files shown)


recipes/atk/all/test_package/CMakeLists.txt
```diff
@@ -1,7 +1,16 @@
...
 target_link_libraries(${PROJECT_NAME} PRIVATE atk::atk)
+
+if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
+    find_package(glib CONFIG REQUIRED)
+    target_link_libraries(${PROJECT_NAME} PRIVATE glib::glib-2.0)
+else()
+    find_package(PkgConfig REQUIRED)
+    pkg_check_modules(glib-2.0 REQUIRED IMPORTED_TARGET glib-2.0)
+    target_link_libraries(${PROJECT_NAME} PRIVATE PkgConfig::glib-2.0)
+endif()
```
valgur, line 16: @SpaceIm Sorry for bothering you again.
The `CMakeDeps` generator for `glib` does not populate the `includedir` values at all for some reason, so I had to apply the same logic that you have previously used in `glib`'s test_package: https://github.com/conan-io/conan-center-index/blob/master/recipes/glib/all/test_package/CMakeLists.txt#L6-L17
Was the PkgConfig approach used there as a workaround or simply as a test of the PkgConfig generator? 
The need for this workaround here looks like a bug to me, though, and downgrading to older v2 Conan does not help either.

valgur:

recipes/atk/all/conanfile.py
```diff
@@ -49,24 +50,30 @@ def layout(self):
...
 
     def build_requirements(self):
-        self.tool_requires("meson/1.1.1")
+        self.tool_requires("meson/1.2.0")
         if not self.conf.get("tools.gnu:pkg_config", check_type=str):
-            self.tool_requires("pkgconf/1.9.3")
+            self.tool_requires("pkgconf/1.9.5")
         if hasattr(self, "settings_build") and cross_building(self):
-            self.tool_requires("glib/2.76.3")
+            self.tool_requires("glib/2.77.1")
```
RubenRBS, line 76:
```suggestion
            self.tool_requires("glib/<host_version>")
```

RubenRBS (changes requested):
This can avoid errors in the future :)

SSE4:
> This can avoid errors in the future :)

what is "this" exactly? what kind of errors? could you possible elaborate a little bit more. I don't understand the reason why are you blocking this PR from your comment.

recipes/atk/all/conanfile.py
```diff
@@ -49,24 +50,30 @@ def layout(self):
...
     def validate(self):
         if self.options.shared and not self.dependencies["glib"].options.shared:
             raise ConanInvalidConfiguration(
                 "Linking a shared library against static glib can cause unexpected behaviour."
             )
 
-        if str(self.settings.compiler) == "Visual Studio" and not self.options.shared and \
-           is_msvc_static_runtime(self) and self.dependencies["glib"].options.shared:
-            raise ConanInvalidConfiguration("this specific configuration is prevented due to internal c3i limitations")
+        if (
```
SpaceIm, line 61: No please read the error message and compare to ci configurations currenty tested. There is no reason to extend this invalid configuration to compiler=msvc

valgur: That error message does not really provide any hint or context for this being strictly related to VS and not msvc. Could you elaborate a bit, please?

SpaceIm: It says its related to c3i issues, it's a v1 pipeline workaround, not a real compiler issue, that's all and you know that v2 pipeline doesn't test this configuration so there is no reason for this change.

ericLemanissier:
This recipe has been deprecated months ago because the upstream project is not maintained (it was merged in at-spi2-core). I see no value in maintaining this recipe any more.

valgur:
Oh, thanks for mentioning this, @ericLemanissier. Closing.

PR #18195, opened 2023-06-30
automake: tidy a bit

valgur:


RubenRBS (approved):

jcar87 (approved):
Hi @valgur ! Thanks for your contribution.

We are accepting this one as I believe adding `self.cpp_info.frameworkdirs = []` is correct given the package type and the settings.

However please bear in mind that we may not ordinarily accept purely stylistic changes if the functionality remains the same. This applies to PRs that are _mainly_ tidying and style changes, or when those changes are part of larger PRs. When it comes to style chances are different contributors are going to have different opinions. From the point of view of us as reviewers, we often see the same things in recipes change back and forth in PRs style wise (in particular in the test_package CMakeLists), and having these things show up in a PR diff detracts our attention from the actual functionality changes.

PR #18196, opened 2023-06-30
avahi: add package_type, set cmake_target_name, pkg_config_name

valgur:


recipes/avahi/all/conanfile.py
```diff
@@ -41,20 +48,14 @@ def requirements(self):
...
         if self.settings.os != "Linux":
             raise ConanInvalidConfiguration(f"{self.ref} only supports Linux.")
 
-    def configure(self):
-        if self.options.shared:
-            self.options.rm_safe("fPIC")
-        self.settings.rm_safe("compiler.cppstd")
-        self.settings.rm_safe("compiler.libcxx")
+    def build_requirements(self):
+        self.tool_requires("glib/2.75.2")
```
jwillikers, line 56: Can you bump the glib version while you're at it? Both here and in the `requirements` above?

```suggestion
        self.tool_requires("glib/2.76.3")
```

jwillikers (changes requested):
Thanks for the update. I understand adding `package_type`, but why remove the V1 test package and why are the `cmake_targe_name` and `pkg_config_name` properties necessary to set? Are no defaults generated? Can you update your description with that information?

recipes/avahi/all/conanfile.py
valgur: The removal of test_v1_package has been requested in most previous PRs, so I thought it's safe to simply include it in all new PRs.
The CMake properties were intended for the newer generators' support, but on second thought these are likely unnecessary if their default value is the component's name.

valgur: The pkg_config_name property replaces the v1 way of setting the same value. It should be compatible with v1.40+ or so versions.

jwillikers: We should probably drop them if the defaults work. I know that since Avahi doesn't provide native CMake Find modules, we're supposed to leave that as the default provided by the generator. Since the defaults for PkgConfig are fine, we can probably leave those off, too.

recipes/avahi/all/conanfile.py
```diff
@@ -98,9 +99,9 @@ def package(self):
     def package_info(self):
         for lib in ("client", "common", "core", "glib", "gobject", "libevent", "compat-libdns_sd"):
             avahi_lib = f"avahi-{lib}"
+            self.cpp_info.components[lib].set_property("pkg_config_name", avahi_lib)
             self.cpp_info.components[lib].names["cmake_find_package"] = lib
             self.cpp_info.components[lib].names["cmake_find_package_multi"] = lib
-            self.cpp_info.components[lib].names["pkg_config"] = avahi_lib
```
jwillikers, line 103: Will removing this break Conan V1 packages that use the pkg-config files?

jwillikers (commented):
Just a question on Conan V1 compatibility. Everything else looks perfect. Thanks!

jwillikers (approved):

recipes/avahi/all/conanfile.py
jwillikers: Oh wait, I think it will only effect the old Conan V1 `PkgConfig` generator and not the newer `PkgConfigDeps` generator. I'm fine with removing support for that. Thanks again for the contribution!

RubenRBS (dismissed):

jwillikers:
It might be something to do with bumping the `dbus` dependency, as this PR fixes #19271.

SSE4 (approved):

jcar87 (approved):

RubenRBS:
Im having problems trying to build `dbus` locally to test this yeah :/

ericLemanissier (approved):

jwillikers (approved):

danimtb (approved):

PR #18207, opened 2023-06-30
jsonformoderncpp: migrate to Conan v2

valgur:


recipes/jsonformoderncpp/all/conanfile.py
```diff
@@ -1,54 +1,60 @@
...
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, rmdir, rm
+
+required_conan_version = ">=1.52.0"
+
 
 class JsonformoderncppConan(ConanFile):
     name = "jsonformoderncpp"
-    homepage = "https://github.com/nlohmann/json"
+    deprecated = "nlohmann_json"
```
RubenRBS, line 12: This is a deprecated recipe, no need to upgrade those

RubenRBS (commented):
Thanks a lot for your contribution, I'm closing the PR as it upgrades a deprecated recipe which right now it's not in our policy to do :/

But thanks for all your effort :)

PR #18265, opened 2023-07-01
libunifex: migrate to Conan v2

valgur:
Requires
- #21061


Ahajha:
I think this might somehow be pulling a system liburing. In liburing 0.1, neither `IOURING_OP_ACCEPT` nor `IOURING_OP_TIMEOUT` was defined, in 0.2 they were both defined, but from [this commit](https://github.com/axboe/liburing/commit/b8712db0fa026e0892e66435e6ff7bb286f21587) it seems there was a brief period of time where only the latter was.

jwillikers (approved):

PR #18283, opened 2023-07-02
BugFix: Allow usage of clang+mingw with zlib

seppeon:

Without this change the following occurs:
```
libpng/1.6.39: RUN: cmake -G "Ninja" -DCMAKE_TOOLCHAIN_FILE="C:/Users/David/.conan2/p/b/libpn0015fee6450cb/b/build/Release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="C:/Users/David/.conan2/p/b/libpn0015fee6450cb/p" -DCMAKE_MACOSX_BUNDLE="OFF" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "C:\Users\David\.conan2\p\b\libpn0015fee6450cb\b\src"

-- Using Conan toolchain: C:/Users/David/.conan2/p/b/libpn0015fee6450cb/b/build/Release/generators/conan_toolchain.cmake
-- Conan toolchain: Setting BUILD_SHARED_LIBS = OFF
-- The C compiler identification is Clang 16.0.5
-- The ASM compiler identification is Clang with GNU-like command-line
-- Found assembler: C:/utils/msys64/mingw64/bin/clang.exe
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/utils/msys64/mingw64/bin/clang.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Conan: Target declared 'ZLIB::ZLIB'
CMake Error at C:/Users/David/.conan2/p/b/libpn0015fee6450cb/b/build/Release/generators/cmakedeps_macros.cmake:64 (message):
  Library 'zlib' not found in package.  If 'zlib' is a system library,
  declare it with 'cpp_info.system_libs' property
  C:/Users/David/.conan2/p/b/libpn0015fee6450cb/b/build/Release/generators/ZLIB-Target-release.cmake:24 (conan_package_library_targets)
  C:/Users/David/.conan2/p/b/libpn0015fee6450cb/b/build/Release/generators/ZLIBTargets.cmake:26 (include)
  C:/Users/David/.conan2/p/b/libpn0015fee6450cb/b/build/Release/generators/ZLIBConfig.cmake:16 (include)
  CMakeLists.txt:51 (find_package)
```
This occurs because in zlib, you have a detection of mingw which was not correct. It didn't identify mingw when it is used with clang. I've added a fix for this.
---


seppeon:
> other pull requests that are modifying zlib/all recipe

That was marked as stale before being seen to completion.

StellaSmith (approved):

recipes/zlib/all/conanfile.py
```diff
@@ -29,7 +29,11 @@ class ZlibConan(ConanFile):
 
     @property
     def _is_mingw(self):
-        return self.settings.os == "Windows" and self.settings.compiler == "gcc"
+        is_windows = self.settings.get_safe("os") == "Windows"
```
CJCombrink, line 32: This feels like a lot of "work" that will result in false in probably most cases (I don't have numbers but I assume mingw has quite a bit lower market share compared to gcc and msvc"
I guess it would still be better to get the short circuit behaviour compared to the old code where it returns false as soon as some condition is reached that makes it false.
I see later it is used for "if windows then _is_mingw()" but still thinks that it can do a lot lees most of the time.

CJCombrink (approved):

recipes/zlib/all/conanfile.py
StellaSmith: This a copy of conan-io/conan#12678 which still hasn't been merged. If you have any issues with it I recommend posting them there.

seppeon: Am I alright to resolve this?

CJCombrink: @seppeon I still think this is overly complex for a simple check. I would still recommend considering early return (see my comment [here](https://github.com/conan-io/conan/pull/12678#discussion_r1250971808)
I am also not a deciding voice so you can resolve if you don't intend to change 👍

PR #18289, opened 2023-07-02
libtiff: allow using zlib-ng instead of zlib

Nekto89:
Specify library name and version:  **libtiff/all**

zlib-ng provides better performance.

---

### DIFF
--- a/recipes/libtiff/all/conanfile.py
+++ b/recipes/libtiff/all/conanfile.py
@@ -24,7 +24,7 @@ class LibtiffConan(ConanFile):
         "fPIC": [True, False],
         "lzma": [True, False],
         "jpeg": [False, "libjpeg", "libjpeg-turbo", "mozjpeg"],
-        "zlib": [True, False],
+        "zlib": [False, "zlib", "zlib-ng"],
         "libdeflate": [True, False],
         "zstd": [True, False],
         "jbig": [True, False],
@@ -36,7 +36,7 @@ class LibtiffConan(ConanFile):
         "fPIC": True,
         "lzma": True,
         "jpeg": "libjpeg",
-        "zlib": True,
+        "zlib": "zlib",
         "libdeflate": True,
         "zstd": True,
         "jbig": True,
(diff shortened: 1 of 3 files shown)


recipes/libtiff/all/conanfile.py
```diff
@@ -24,7 +24,7 @@ class LibtiffConan(ConanFile):
         "fPIC": [True, False],
         "lzma": [True, False],
         "jpeg": [False, "libjpeg", "libjpeg-turbo", "mozjpeg"],
-        "zlib": [True, False],
+        "zlib": [False, "zlib", "zlib-ng"],
```
memsharded, line 27: This change will break users that had ``zlib=True`` option defined. I'd try to leave it True with meaning "zlib" and just add the new one.

RubenRBS:
Hi @Nekto89 thanks a lot for your patience - These PRs got put on hold while we could think about ways to have this be used more generally, supported within the Conan client itself.

There's now a PR https://github.com/conan-io/conan/pull/14871 (We might or might not end up merging it, we're still in the thinking and discussion phase) that would introduce the possibility for users to do these kinds of substitutions in a global way, without the need to modify the recipes.

If the feature does not go forward we'll then revisit the PRs, so either way, thanks a lot for your contribution :)

Nekto89:
> Hi @Nekto89 thanks a lot for your patience - These PRs got put on hold while we could think about ways to have this be used more generally, supported within the Conan client itself.
> 
> There's now a PR [conan-io/conan#14871](https://github.com/conan-io/conan/pull/14871) (We might or might not end up merging it, we're still in the thinking and discussion phase) that would introduce the possibility for users to do these kinds of substitutions in a global way, without the need to modify the recipes.
> 
> If the feature does not go forward we'll then revisit the PRs, so either way, thanks a lot for your contribution :)

Hi @RubenRBS

https://blog.conan.io/2024/02/20/Conan-2-graph-features.html

Should I close PRs with zlib-ng option? Is it something that should be completely on user side with new `[replace_requires]`? What about check of `zlib_ng.options.zlib_compat` option? How to implement it in case of `[replace_requires]` usage?

Issue #18155, opened 2023-06-30
[question] sqlite missing recommended compile-time options


wAuner:
### What is your question?

Hi, I have a question about the sqlite recipes. Maybe I'm missing something here, but to me it looks as if almost all of SQLite's [recommended compile-time options](https://www.sqlite.org/compile.html#recommended_compile_time_options) are missing from the conan recipes.
https://github.com/conan-io/conan-center-index/blob/d9de202b50396709fc85cab764a79a7e2f35b91d/recipes/sqlite3/all/CMakeLists.txt#L5

I just tried to add them to my conanfile, since we're using many of them. Is this on purpose? Otherwise, I'd be willing to make a PR to add those settings.

PR #18297, opened 2023-07-03
Add sqlite3 recommended options

wAuner:
Specify library name and version:  **sqlite3/***

This is an update to the `sqlite3` recipe which is missing the library's recommended compile options; see #18155. 
* adds the missing recommended compile options
* uses the library's default options for all recommended flags (default value is `None` not to overwrite the default settings)
* adds the official docstrings to the internal cmake option descriptions

Tested with conan 1 (conan center hooks activated) and 2.
I did not change the default values for the other (not in the recommended options listed) options. The existing options however, also listed in the recommended section, `threadsafe` (1), `use_alloca` (False) & `omit_deprecated` (False) have now a default value of `None` so that the default does not diverge from the maintainer's intend in case of a change.

Please let me know if there's anything that needs to be changed, I'm open to feedback.
---


wAuner:
Two CI checks are failing because the commit cannot be built. But I can't access the details. Is there anything I can do to speed things up? We need these changes for our conan adoption.

PR #18312, opened 2023-07-04
xerces-c: add cmake/3.26.4 to build requirements

Todiq:
Specify library name and version:  **xerces-c/3.2.4**

Add cmake/3.26.4 as a build requirement for xerces-c/3.2.4

---

### DIFF
--- a/recipes/xerces-c/all/conanfile.py
+++ b/recipes/xerces-c/all/conanfile.py
@@ -100,6 +100,7 @@ class XercesCConan(ConanFile):
     def build_requirements(self):
         if self.options.message_loader == "icu" and not can_run(self):
             self.tool_requires("icu/72.1")
+        self.tool_requires("cmake/3.26.4")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)


jcar87 (commented):
Hi @Todiq, thanks for your contribution.

What is the motivation behind this PR? We don't _generally_ add CMake as a tool requirement in recipes, unless the version required by the library is higher than the the version we have running on CI (which is currently 3.15 on Linux). I can't see this being the case here.

Please be advised that it is possible to list `tool_requires` in your profiles directly, rather than the recipes:
https://docs.conan.io/2.0/reference/config_files/profiles.html#tool-requires

With the current policies it is unlikely this PR will be accepted unless there is a clear motivation.

Todiq:
Hello @jcar87,

Thank you for the quick reply.
It is purely from a convenience point of view. Just like boost's recipe does mention b2 as its build requirement, I think that users should not be forced to manually install cmake on their own beforehand.
On clean docker images with only python, conan and g++ installed, it sometimes happen that I forget to install it and that leads to a fail during xerces-c's build. Such things do not happen with builds based of make, since it is installed alongside GCC in general.
I would find it much easier for it to be already included in the recipe, as conan handles that pretty well and that the option exists in the first place. I do not mind cmake's version, I input 3.26.4 because it is the latest version. Hope you understand. Thank you for taking care of this.

jcar87:
> Just like boost's recipe does mention b2 as its build requirement, I think that users should not be forced to manually install cmake on their own beforehand. 

Unfortunately the current assumption in Conan Center is that at least CMake 3.15 is installed at the system lvel. Unlike b2, CMake is a very common tool, to the point where it is probably the most common build mechanism at this point across all recipes. Most users in your scenario would typically install CMake in the docker images, alongside python, conan, g++. 

We are reviewing whether it would make sense for _all_ recipes that use CMake to list a requirement, but this is not something we currently support. I do suggest you do install CMake in your Docker images as most of our users in similar scenarios do, or that you consider adding this as an implicit tool requirement in your Conan profiles.

Todiq:
Sure, I understand. Hopefully that changes in the future. I'll manage to install it my own environments for now. Thank you :)

PR #18313, opened 2023-07-04
turtle: fix lib/bindirs

Cogitri:
Since turtle is header-only recipe, we should unset libdirs and bindirs as per https://github.com/conan-io/conan/issues/13702#issuecomment-1510507647

### DIFF
--- a/recipes/turtle/all/conanfile.py
+++ b/recipes/turtle/all/conanfile.py
@@ -24,6 +24,8 @@ class TurtleConan(ConanFile):
 
     def package_id(self):
         self.info.clear()
+
+    def package_info(self):
         self.cpp_info.libdirs = []
         self.cpp_info.bindirs = []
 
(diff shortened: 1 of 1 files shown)


recipes/turtle/all/conanfile.py
```diff
@@ -24,6 +24,8 @@ def requirements(self):
 
     def package_id(self):
         self.info.clear()
+        self.cpp_info.libdirs = []
```
jcar87, line 27: Hi @Cogitri - thanks for your contribution.

Indeed these two lines are advised for header only library - however they should be placed in the `package_info()` method, rather than the `package_id()` one.

Cogitri: Ah thanks :)

ericLemanissier (approved):

prince-chrismc (approved):

PR #18314, opened 2023-07-04
[quazip/x.x.x] Use version ranges for dependencies

MartinDelille:

---

### DIFF
--- a/recipes/quazip/all/conanfile.py
+++ b/recipes/quazip/all/conanfile.py
@@ -48,10 +48,10 @@ class QuaZIPConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("qt/5.15.9")
-        self.requires("zlib/1.2.13")
+        self.requires("qt/[>=5.15.2]")
+        self.requires("zlib/[>=1.2.11]")
         if Version(self.version) >= "1.4":
-            self.requires("bzip2/1.0.8")
+            self.requires("bzip2/[>=1.0.8]")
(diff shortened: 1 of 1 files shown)


recipes/quazip/all/conanfile.py
```diff
@@ -48,10 +48,10 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("qt/5.15.9")
-        self.requires("zlib/1.2.13")
+        self.requires("qt/[>=5.15.2]")
```
CJCombrink, line 51: Does this mean that this package is Qt 6 compatible?

CJCombrink (approved):

recipes/quazip/all/conanfile.py
MartinDelille: Thank you for the review! Yes it is compatible. I'm just wondering which qt version will be picked up in the test.

That fact that the target name contains the qt version can be problematic though.

RubenRBS:
Hi @MartinDelille thanks a lot for your contribution. I see that you closed the PR, but let me give you some insight:

Our currently policy for version ranges is to limit them to a handful of libraries for which we know and understand their compatibility matrixes. Once we're comfortable having more libraries use this method, we'll let everyone know so it can be applied elsewhere :)

MartinDelille:
I closed this PR because I had an issue with conan v2 and quazip/qt version conflict. I fixed it by using the requires override attribute!

Issue #18134, opened 2023-06-29
spdlog spdlog/1.11.0: implicitly changed required fmt without change of own version - broken builds


yshurik:
### Description

spdlog 1.11.0 quietly applied change to change the requirement require from fmt/9.1.0 to  fmt/10.0.0 without the upgrade of the version (like 1.11.1)

Now old build recipes become invalid - forced fmt update while a lot was depending on 9.1.0, a lot of object files got linking errors. Old artifacts become invalid.

Can the change be reverted and spd/1.11.1 be introduced with such changed requirement?

conan.io still have wrong dependency referent to 9.1.0 - https://conan.io/center/spdlog?version=1.11.0&revision=d0fdbaa523550b89156084bf42b41c90&tab=dependencies&os

### Package and Environment Details

* Package Name/Version: spdlog/1.11.0
* Operating System+version: Linux Ubuntu 
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

[settings]
os=Macos
os_build=Macos
arch=armv8
arch_build=armv8
compiler=apple-clang
compiler.version=14
compiler.libcxx=libc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

Try to rebuild existing package which was already successfully built with spdlog/1.11.0

ERROR: Conflict in spdlog/1.11.0:
    'spdlog/1.11.0' requires 'fmt/10.0.0' while 'Application/1.0.0@Common+Application/stable' requires 'fmt/9.1.0'.
    To fix this conflict you need to override the package 'fmt' in your root package.

### Logs

<details><summary>Click to expand log</summary>

```
ERROR: Conflict in spdlog/1.11.0:
    'spdlog/1.11.0' requires 'fmt/10.0.0' while 'Application/1.0.0@Common+Application/stable' requires 'fmt/9.1.0'.
    To fix this conflict you need to override the package 'fmt' in your root package.
```

</details>

Seadex-GmbH:
This is true and stops the development of other libraries/apps that are using libraries that have `spdlog` dependencies.
I want to build a library that uses `seadex-essentials` library. And `seadex-essentials` has a `spdlog` and `fmt` dependency.

Now I get the error and it is a show-stopper:

[...]

PR #17976, opened 2023-06-20
spdlog: bump fmt10.0.0 and apply patch to fix build error

pwqbot:
Specify library name and version:  **spdlog/1.11.0**

problems:
spdlog 1.11.0 do not support. fmt 10
https://github.com/gabime/spdlog/issues/2782
https://github.com/gabime/spdlog/issues/2767

relate pr:
https://github.com/gabime/spdlog/pull/2694
https://github.com/spack/spack/pull/38082

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18318, opened 2023-07-04
seadex-essentials: Bump versions, remove header option force

RubenRBS:
Specify library name and version:  **seadex-essentials/all**

Fixes the conflicts from https://github.com/conan-io/conan-center-index/issues/18134, while also improving simplifying logic that as @czoido pointed out, is no good either way.

### DIFF
--- a/recipes/seadex-essentials/all/conanfile.py
+++ b/recipes/seadex-essentials/all/conanfile.py
@@ -39,7 +39,7 @@ class SeadexEssentialsConan(ConanFile):
             "Visual Studio": "16",
             "msvc": "192",
             "apple-clang": "10"
-        }        
+        }
 
     def config_options(self):
         if self.settings.os == "Windows":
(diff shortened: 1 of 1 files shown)


czoido (approved):

prince-chrismc (approved):

prince-chrismc (changes requested):
I am not sure this is the correct place to fix the problem.

https://github.com/conan-io/conan-center-index/pull/17976 backport patches and breaks spdlog

prince-chrismc (approved):

recipes/seadex-essentials/all/conanfile.py
```diff
@@ -48,14 +48,12 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        self.options["fmt/*"].header_only = True
-        self.options["spdlog/*"].header_only = True
```
SpaceIm, line 52: This is a bad idea, because now it doesn't build out of the box, and c3i doesn't test this recipe anymore (all configurations raise Invalid Configuration)

PR #18344, opened 2023-07-05
resource_pool: migrate to Conan v2

valgur:

---


recipes/resource_pool/all/conanfile.py
```diff
@@ -1,32 +1,30 @@
...
-from conans.tools import check_min_cppstd
 import os
-import glob
+
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, get
+from conan.tools.layout import basic_layout
+from conans.errors import ConanInvalidConfiguration
+from conans.model.version import Version
```
valgur, line 8:
```suggestion
from conan.tools.scm import Version
```

recipes/resource_pool/all/conanfile.py
```diff
@@ -37,44 +35,55 @@ def _compilers_minimum_version(self):
             "apple-clang": "10",
         }
 
-    def _validate_compiler_settings(self):
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("boost/[<=1.79]")
```
jcar87, line 42: We are still not using version ranges for anything other than CMake and OpenSSL - I'm surprised this one does not work with boost 1.80 or greater . I'd suggest keeping this on 1.79 if that's the highest it supports for now.

```suggestion
        self.requires("boost/1.79.0")
```

RubenRBS (approved):

czoido (approved):

PR #18352, opened 2023-07-05
si: migrate to Conan v2

valgur:

---


recipes/si/all/conanfile.py
```diff
@@ -36,38 +38,47 @@ def _compilers_minimum_version(self):
...
     def source(self):
-        tools.get(**self.conan_data["sources"][self.version],
-                  destination=self._source_subfolder, strip_root=True)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["SI_BUILD_TESTING"] = False
+        tc.variables["SI_BUILD_DOC"] = False
+        tc.variables["SI_INSTALL_LIBRARY"] = True
```
RubenRBS, line 67: Is there a good reason to use variables instead of cache variables?

```suggestion
        tc.cache_variables["SI_BUILD_TESTING"] = False
        tc.cache_variables["SI_BUILD_DOC"] = False
        tc.cache_variables["SI_INSTALL_LIBRARY"] = True
```

RubenRBS (commented):
Question about cache variables vs variables

recipes/si/all/conanfile.py
memsharded: ``variables`` have an advantage over ``cache_variables``: they work easier for local-development, for all users using presets and using conan_toolchain, with the same result. Otherwise it is easier for local-development to build something different if they don't pass explicitly all those ``-DSI_BUILD_DOC=False`` on the command line.

But given that ConanCenter recipes are almost never developed locally, not a big deal.

RubenRBS (approved):

PR #18366, opened 2023-07-05
sdl_ttf: and 2.20.2, bump dep sdl, conan v2 fixes

prince-chrismc:

---

### DIFF
--- a/recipes/sdl_ttf/all/conandata.yml
+++ b/recipes/sdl_ttf/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "2.20.2":
+    url: "https://github.com/libsdl-org/SDL_ttf/releases/download/release-2.20.2/SDL2_ttf-2.20.2.tar.gz"
+    sha256: "9dc71ed93487521b107a2c4a9ca6bf43fb62f6bddd5c26b055e6b91418a22053"
   "2.20.1":
     url: "https://github.com/libsdl-org/SDL_ttf/releases/download/release-2.20.1/SDL2_ttf-2.20.1.tar.gz"
     sha256: "78cdad51f3cc3ada6932b1bb6e914b33798ab970a1e817763f22ddbfd97d0c57"
@@ -13,6 +16,10 @@ patches:
     - patch_file: "patches/2.20.1-0001-fix-cmake-min-version.patch"
       patch_description: "Disable useless .pc file install to avoid relying on CMake 3.21 features"
       patch_type: "conan"
+    - patch_file: "2.20.1-0002-remove-unsupported-property.patch"
+      patch_description: "remove DEPRECATION property is available on CMake 3.17+ (Linux C3i is 3.15 and the Conan min)"
+      patch_type: "portability"
+      patch_source: "https://github.com/libsdl-org/SDL_ttf/commit/e49b6030bd57b7aa3f0e4e42cb7ccc5fee876184"
   "2.0.18":
     - patch_file: "patches/2.0.18-0001-cmake-fix-link-target.patch"
       patch_description: "correct target name, disable PIC fixed"
--- a/recipes/sdl_ttf/all/conandata.yml
+++ b/recipes/sdl_ttf/all/conandata.yml
@@ -16,7 +16,7 @@ patches:
     - patch_file: "patches/2.20.1-0001-fix-cmake-min-version.patch"
       patch_description: "Disable useless .pc file install to avoid relying on CMake 3.21 features"
       patch_type: "conan"
-    - patch_file: "2.20.1-0002-remove-unsupported-property.patch"
+    - patch_file: "patches/2.20.1-0002-remove-unsupported-property.patch"
       patch_description: "remove DEPRECATION property is available on CMake 3.17+ (Linux C3i is 3.15 and the Conan min)"
       patch_type: "portability"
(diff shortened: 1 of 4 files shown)


prince-chrismc:
This was fix with https://github.com/libsdl-org/SDL_ttf/compare/release-2.20.1...release-2.20.2#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aL406

🤔

recipes/sdl_ttf/all/conandata.yml
```diff
@@ -13,6 +16,10 @@ patches:
     - patch_file: "patches/2.20.1-0001-fix-cmake-min-version.patch"
       patch_description: "Disable useless .pc file install to avoid relying on CMake 3.21 features"
       patch_type: "conan"
+    - patch_file: "2.20.1-0002-remove-unsupported-property.patch"
```
prince-chrismc, line 19:
```suggestion
    - patch_file: "patches/2.20.1-0002-remove-unsupported-property.patch"
```

recipes/sdl_ttf/all/conanfile.py
```diff
@@ -51,7 +51,7 @@ def layout(self):
 
     def requirements(self):
         self.requires("freetype/2.12.1")
-        self.requires("sdl/2.26.1")
+        self.requires("sdl/2.26.5")
```
jcar87, line 54:
```suggestion
        self.requires("sdl/2.26.5", transitive_headers=True)
```

jcar87: I've added this since it looks the sdl-ttf headers would bring the sdl ones transitively, but there may also be an argument for the test package to direectly require sdl as well if the code in the test package makes _direct_ calls to SDL functionality (rather than via sdl-ttf)

prince-chrismc:
This never built with v2 https://github.com/conan-io/conan-center-index/pull/15560#issuecomment-1411453915

And v2 is not required 🙄 https://github.com/conan-io/conan-center-index/blob/e950432268d73e42fa78a0de7e8ff5733f834992/.c3i/conan_v2_ready_references.yml#L675

recipes/sdl_ttf/all/conanfile.py
```diff
@@ -51,7 +51,7 @@ def layout(self):
 
     def requirements(self):
         self.requires("freetype/2.12.1")
-        self.requires("sdl/2.26.1")
+        self.requires("sdl/2.26.5", transitive_headers=True)
```
prince-chrismc, line 54:
```suggestion
        # https://github.com/conan-io/conan-center-index/pull/18366#issuecomment-1625464996
        self.requires("sdl/2.26.5", transitive_headers=True, transitive_libs=True)
```

czoido (approved):

recipes/sdl_ttf/all/conanfile.py
```diff
@@ -121,6 +123,10 @@ def package_info(self):
         self.cpp_info.components["_sdl2_ttf"].requires = ["freetype::freetype", "sdl::libsdl2"]
         if self.options.get_safe("with_harfbuzz"):
             self.cpp_info.components["_sdl2_ttf"].requires.append("harfbuzz::harfbuzz")
+        if Version(self.version) <= "2.0.18" and is_apple_os(self) and self.options.shared:
+            self.cpp_info.components["_sdl2_ttf"].frameworks = [
+                "AppKit", "CoreGraphics", "CoreFoundation", "CoreServices"
+        ]
```
SpaceIm, line 129: It makes no sense, it should be for static only, not shared.

prince-chrismc: Please open an issue with more information :) If the hooks are incorrect https://github.com/conan-io/conan-center-index/pull/18366#issuecomment-1625310577 please open an issue there so we can track it 🙏

SpaceIm: Hook is correct, but your interpretation of its result is incorrect.

SpaceIm: This hook just says that shared lib links to some system libs or frameworks, so it means that these system libs/frameworks must be added if static. This hook has always been an indirect way to find system libs or frameworks to add if static.

prince-chrismc: I'd argue the hooks are bad for providing misleading information

> Library './lib/libSDL2_ttf.dylib' links to system library 

implies the shared one needs it, if that incorrect -- then the messages are unclear -- it should have included the settings and options to be more concise :)

SpaceIm: Feel free to update message hook if you think it's misleading.

Instead of:

```
post_package(): WARN: [MISSING SYSTEM LIBS (KB-H043)] Library './lib/libSDL2_ttf.dylib' links to system library 'CoreServices' but it is not in cpp_info.frameworks.
```

it could be:

```
post_package(): WARN: [MISSING SYSTEM LIBS (KB-H043)] Library './lib/libSDL2_ttf.dylib' links to system library 'CoreServices', therefore it should be in cpp_info.frameworks if static, but it's missing.
```

prince-chrismc: Done, thanks for the suggestion!

PR #18367, opened 2023-07-06
openssl: add support for `install_ssldirs` and/or `install_fips`

gegles:
Specify library name and version:  **openssl/v3.x.x**

The various config files and certs-related files generated and installed via OpenSSL's `install_ssldirs` and `install_fips` targets can sometimes be needed by packages using `openssl`, either directly or repackaged when a final product embeds OpenSSL. Currently only the `install_sw` target is installed.

This PR does the following:
* It adds a package option called `with_ssldirs`
* The default value for `with_ssldirs` is `False` (i.e. same as current behavior where only `install_sw` is called)
* When `with_ssldirs` is `True`, `install_ssldirs` is called at install time
* Also, when `no_fips` is `False`, the `install_fips` target is called at install time

Because of how the `install_ssldirs` and `install_fips` targets heavily depend on the combo of `$(DESTDIR)` and `$(OPENSSLDIR)`, we have to override the previously set value of `OPENSSLDIR` in order for Make to install these files in a known/package-level subdir (i.e. `etc/ssl`). This does not affect the value of OPENSSLDIR that is set in the openssl binary at build time, especially when the user set it via the `openssldir` package option.

So, when the package was created using:
```
conan create --version 3.1.1 3.x.x -o with_ssldirs=True -o openssldir=/opt/mysoftware/etc/ssl
```
The resulting package's `openssl` will still embed the correct `OPENSSLDIR`:
```
~/.conan2/p/b/opensa94ddd3dbc68b/p/bin/openssl version -d                                                                                                                                                            
OPENSSLDIR: "/opt/mysoftware/etc/ssl"
```

Even though the ssl dir config files will be packaged at:
```
 tree ~/.conan2/p/b/opensa94ddd3dbc68b/p/etc/ssl 
├── certs
├── ct_log_list.cnf
├── ct_log_list.cnf.dist
├── fipsmodule.cnf
├── misc
│   ├── CA.pl
│   ├── tsget -> tsget.pl
│   └── tsget.pl
├── openssl.cnf
├── openssl.cnf.dist
└── private

4 directories, 8 files
```

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -83,6 +83,7 @@ class OpenSSLConan(ConanFile):
         "no_whirlpool": [True, False],
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
+        "with_ssldirs": [True, False],
     }
     default_options = {key: False for key in options.keys()}
     default_options["fPIC"] = True
@@ -397,7 +398,7 @@ class OpenSSLConan(ConanFile):
             ])
 
         for option_name in self.default_options.keys():
-            if self.options.get_safe(option_name, False) and option_name not in ("shared", "fPIC", "openssldir", "capieng_dialog", "enable_capieng", "zlib", "no_fips", "no_md2"):
+            if self.options.get_safe(option_name, False) and option_name not in ("shared", "fPIC", "openssldir", "with_ssldirs", "capieng_dialog", "enable_capieng", "zlib", "no_fips", "no_md2"):
                 self.output.info(f"Activated option: {option_name}")
                 args.append(option_name.replace("_", "-"))
         return args
@@ -482,6 +483,7 @@ class OpenSSLConan(ConanFile):
         command = [self._make_program]
         if install:
             command.append(f"DESTDIR={self.package_folder}")
+            command.append(f"OPENSSLDIR=/etc/ssl")
         if targets:
             command.extend(targets)
         if not self._use_nmake:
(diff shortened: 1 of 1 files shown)


jwillikers (approved):

gegles:
@Hopobcn @Croydon let us know what you think of this PR and/or if you think there is a better approach to the DESTDIR vs OPENSSLDIR...

For example, why is the `get_default_openssl_dir` different on Linux than [elsewhere](https://github.com/conan-io/conan-center-index/blob/master/recipes/openssl/3.x.x/conanfile.py#L339)?

Why could we not simply have the default openssldir be `/res/` on all platforms? This would always install the config files under `<package_folder>/<openssldir>` and would be `<package_folder>/res` by default (when no openssldir is specified.

Let us know. thx!

kulkarniamit:
According to https://github.com/openssl/openssl/blob/master/INSTALL.md:
```
Directory for OpenSSL configuration files, and also the default certificate and key store. Defaults are:
Unix:           /usr/local/ssl
Windows:        C:\Program Files\Common Files\SSL
OpenVMS:        SYS$COMMON:[OPENSSL-COMMON]
```
But this would violate the [DEFAULT-PACKAGE-LAYOUT rule](https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H013-DEFAULT-PACKAGE-LAYOUT).
A good solution would apply these official defaults as --openssldir (so binaries on installed machines have good default locations to look for config files) during compilation and use a standard location like `<package_folder>/res/`  while creating the package using `OPENSSLDIR`.
This will allow conan to create package binaries with official defaults and follow [DEFAULT-PACKAGE-LAYOUT](https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H013-DEFAULT-PACKAGE-LAYOUT) rules. Consumers can reliably extract config files from a known location and yet have official default path for config files.

PR #18374, opened 2023-07-06
libpng: add version 1.6.32

jalapenopuzzle:
Specify library name and version:  **libpng/1.6.32**

I am not the author of this library.
I am porting an application to conan. The application used this exact version of libpng, so I added it to conan.

---

- [*] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [*] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [*] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

### DIFF
--- a/recipes/libpng/all/conandata.yml
+++ b/recipes/libpng/all/conandata.yml
@@ -11,6 +11,9 @@ sources:
   "1.6.37":
     url: "https://sourceforge.net/projects/libpng/files/libpng16/1.6.37/libpng-1.6.37.tar.xz"
     sha256: "505e70834d35383537b6491e7ae8641f1a4bed1876dbfe361201fc80868d88ca"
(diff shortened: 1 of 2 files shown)


Nekto89:
just wondering: why do you need this older version? Is there some breaking change in newer releases?

jalapenopuzzle:
> just wondering: why do you need this older version? Is there some breaking change in newer releases?

I'm adapting existing software to use conan and I wanted to use exactly the same library versions in the first instance. I'll upgrade library versions one by one later and check for issues.

PR #18380, opened 2023-07-06
[mingw-builds] Recipe and test package compatibility for conan v2

danimtb:
closes https://github.com/conan-io/conan-center-index/pull/17768

Some of the changes in this PR are due to the fact that conan-center-index does not build packages for mingw, so this means it does not use mingw profiles to generate this package. In order to make this package comaptible, I had to turn some of the invalid configuration excetions into warnings. I know it is not the best solution but I think we should go this way for now because:
- It is more important to have a mingw tool recipe working for conan 2 asap than to make sure it raises when the compiler is not gcc and so.
- I believe, people using this recipe as tool require already know that they should use the "gcc" compiler in their profiles (in case they did not see the warnings"

I have also removed the compilation of the main.cpp in test_package to simplify it and for the same reason explained above. I believe, making sure that the mingw tools are in the path and can be run is enough for the purpose of the package. However, in case I am missing something, please let me know in the review.

### DIFF
--- a/recipes/mingw-builds/all/test_package/conanfile.py
+++ b/recipes/mingw-builds/all/test_package/conanfile.py
@@ -1,16 +1,14 @@
 import os
-from conans import ConanFile, tools
+from conan import ConanFile
 
 
 class MinGWTestConan(ConanFile):
-    generators = "gcc"
-    settings = "os", "arch", "compiler", "build_type"
+    generators = "VirtualBuildEnv"
+    settings = "os", "arch"
+    test_type = "explicit"
 
-    def build(self):
-        source_file = os.path.join(self.source_folder, "main.cpp")
-        self.run("gcc.exe {} @conanbuildinfo.gcc -lstdc++ -o main".format(source_file), run_environment=True)
+    def requirements(self):
+        self.tool_requires(self.tested_reference_str)
 
     def test(self):
-        if not tools.cross_building(self):
-            self.run("gcc.exe --version", run_environment=True)
-            self.run("main")
+        self.run("gcc.exe --version")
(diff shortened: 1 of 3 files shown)


recipes/mingw-builds/all/test_package/conanfile.py
```diff
@@ -1,16 +1,14 @@
...
-    settings = "os", "arch", "compiler", "build_type"
+    generators = "VirtualBuildEnv"
+    settings = "os", "arch"
+    test_type = "explicit"
 
-    def build(self):
-        source_file = os.path.join(self.source_folder, "main.cpp")
-        self.run("gcc.exe {} @conanbuildinfo.gcc -lstdc++ -o main".format(source_file), run_environment=True)
+    def requirements(self):
+        self.tool_requires(self.tested_reference_str)
```
jcar87, line 11: in v1 this still needs to be in the `build_requirements()` method :D

recipes/mingw-builds/all/test_package/conanfile.py
```diff
@@ -1,16 +1,14 @@
...
-    generators = "gcc"
-    settings = "os", "arch", "compiler", "build_type"
+    generators = "VirtualBuildEnv"
+    settings = "os", "arch"
+    test_type = "explicit"
 
-    def build(self):
-        source_file = os.path.join(self.source_folder, "main.cpp")
-        self.run("gcc.exe {} @conanbuildinfo.gcc -lstdc++ -o main".format(source_file), run_environment=True)
+    def requirements(self):
```
jcar87, line 10:
```suggestion
    def build_requirements(self):
```

jcar87 (approved):

PR #18388, opened 2023-07-06
Use version range for openssl dependency including version 3.x

dheater:
Specify library name and version:  **libssh2/1.11.0**

Using the version range for the dependency on openssl.

Support for openssl 3.x was marked at *not working* in a comment that I added previously, but that appears to have been a local configuration issue. It is working now.

---


RubenRBS (approved):
Thanks for double checking and improving the recipe! :)

ericLemanissier (approved):

PR #18389, opened 2023-07-06
Add pcl PointCloudLibrary package

EstebanDugueperoux2:
Inspired from https://github.com/conan-io/conan-center-index/pull/1891 but conan v2 compatible.

Specify library name and version:  **pcl/1.13.1**

---


recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,224 @@
...
+        self.requires("eigen/3.3.9")
+        self.requires("flann/1.9.2")
+        if self.options.with_png:
+            self.requires("libpng/1.6.39")
+        if self.options.with_qhull:
+            self.requires("qhull/8.0.1")
+        if self.options.with_apps:
+            self.requires("qt/6.5.0")
+        if self.options.with_libusb:
+            self.requires("libusb/1.0.26")
```
valgur, line 89: self.options.with_pcap likely requires libpcap.
Also opengl for with_opengl.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,224 @@
...
+        tc.variables["WITH_LIBUSB"] = self.options.with_libusb
+        tc.variables["WITH_PNG"] = self.options.with_png
+        tc.variables["WITH_QHULL"] = self.options.with_qhull
+        tc.variables["WITH_CUDA"] = self.options.with_cuda
+        tc.variables["WITH_PCAP"] = self.options.with_pcap
+        tc.variables["WITH_OPENGL"] = self.options.with_opengl
+        tc.variables["BUILD_TOOLS"] = self.options.with_tools
+        tc.variables["BUILD_APPS"] = self.options.with_apps
+        tc.cache_variables["WITH_QT"] = self.options.with_apps
+        tc.cache_variables["WITH_VTK"] = False
```
valgur, line 122: Maybe add with_vtk option, but display a self.output.warning() mentioning that it needs to be provided by the system?

recipes/pcl/all/patches/0001-flann-target-cmake.patch
```diff
@@ -0,0 +1,298 @@
...
+-# ^^^^^^^^^^^^^
+-#
+-# ::
+-#
+-#     find_package(FLANN REQUIRED)
+-#
+-#     add_executable(foo foo.cc)
+-#     target_link_libraries(foo FLANN::FLANN)
+-#
+-
```
valgur, line 55: More of a nitpick, but the patch size could be reduced by wrapping the unnecessary block of code in if(FALSE) ... endif() instead of deleting the file contents.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,224 @@
...
+required_conan_version = ">=1.53.0"
+
+
+class PclConan(ConanFile):
+    name = "pcl"
+    description = "Point Cloud Library"
+    license = "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/PointCloudLibrary/pcl"
+    topics = ("computer-vision", "point-cloud", "pcl", "pointcloud")
```
valgur, line 21: The pcl tag is redundant. Could instead add pcd tag for the suppoeted file format and possibly other formats as well.

valgur (changes requested):
Overall looks good. Should add the missing optional requirements, though.

recipes/pcl/all/conanfile.py
EstebanDugueperoux2: Done for pcd, for others file extensions, there are many according to https://pointclouds.org/documentation/group__io.html, then I prefer to limit to pcd.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,242 @@
...
+required_conan_version = ">=1.53.0"
+
+
+class PclConan(ConanFile):
+    name = "pcl"
+    description = "Point Cloud Library"
+    license = "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/PointCloudLibrary/pcl"
+    topics = ("computer-vision", "point-cloud", "pcd", "pointcloud")
```
valgur, line 21:
```suggestion
    topics = ("computer vision", "point cloud", "pointcloud", "3d", "pcd", "ply", "stl", "ifs", "vtk")
```
I the rest of the io format tags are worth including to make the package easier to find based on these. For example, there's an rply package for .ply in ConanCenter and PCL can be used as an alternative for it.
Also, I think the hyphen is not necessary.

EstebanDugueperoux2:
Hi @jcar87, could you add pcl recipe to xlarge node to avoid pipelines to be killed because of memory consumption?

Regards.

recipes/pcl/all/patches/0001-KB-H020_avoid_pkgconfig_files.patch
```diff
@@ -0,0 +1,200 @@
...
++            self.cpp_info.system_libs.append("dl")
++
++        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
++        self.cpp_info.filenames["cmake_find_package"] = "PACKAGE"
++        self.cpp_info.filenames["cmake_find_package_multi"] = "package"
++        self.cpp_info.names["cmake_find_package"] = "PACKAGE"
++        self.cpp_info.names["cmake_find_package_multi"] = "package"
+-- 
+2.41.0
+
```
valgur, line 200: You probably did not intend to include this?
```suggestion
```

recipes/pcl/all/patches/0001-KB-H020_avoid_pkgconfig_files.patch
```diff
@@ -0,0 +1,12 @@
...
+@@ -571,9 +571,6 @@ function(PCL_MAKE_PKGCONFIG _name)
+   else()
+     configure_file(${PROJECT_SOURCE_DIR}/cmake/pkgconfig.cmake.in ${_pc_file} @ONLY)
+   endif()
+-  install(FILES ${_pc_file}
+-          DESTINATION ${PKGCFG_INSTALL_DIR}
+-          COMPONENT pcl_${ARGS_COMPONENT})
+ endfunction()
+ 
+ ###############################################################################
```
valgur, line 12:
```suggestion
 ###############################################################################

```
To fix the linter error in CI.

EstebanDugueperoux2:
Hi @valgur 
It is not the first time that some pipeline jobs are killed because of too high memory consumption.
Do you know how to fix that?
I have asked to @jcar87  a xlarge node without answer.

Regards.

jcar87:
Hi @EstebanDugueperoux2, we're currently investigating. I can see that for the most recent build, there _are_ binaries produced for multiple versions of gcc, including one in the Debug configuration. For the one that most recently failed, I'll try to confirm whether the internal compiler error is repeatable (it's quite possible), or if this is random and thus would indicate an external factor (such as lack of memory). Currently investigating.

jcar87:
It appears that compiling the file `pcl_surface.dir/src/mls.cpp.o` can cause a failure if the compiler does not have enough RAM available. I make or ninja are building multiple files in parallel (in addition to `mls.cpp`) this is an additional problem. It appears that this is a problem that has already been identified upstream https://github.com/PointCloudLibrary/pcl/issues/3414#issuecomment-1334684151.

Currently working on determining what the minimum requirements are. This one is a bit of a significant outlier - memory exhaustion tend to happen due to parallelism when memory is not enough across all jobs, or at the linking stage (which is more memory intensive). A single invocation of the compiler using this much memory to the point where it crashes, it's quite the rare event.

EstebanDugueperoux2:
Hi @valgur,

On conan V1 pipeline, I see a killed meaning perhaps it was not runned on xlarge node, but I don't know how to check that.
On conan v2 pipeline, I see msvc compiler error on missing include on Eigen library, but I don't understand why.
Do you have an idea?

Regards.

valgur:
That's great! Hopefully adding this patch will solve the memory issues then: https://github.com/PointCloudLibrary/pcl/commit/0bc2f2cc369f7d013ee5c29075188c4e26ea6f87.patch
```patch
From 0bc2f2cc369f7d013ee5c29075188c4e26ea6f87 Mon Sep 17 00:00:00 2001
From: Markus Vieth <mvieth@techfak.uni-bielefeld.de>
Date: Sat, 15 Jul 2023 15:49:25 +0200
Subject: [PATCH] MovingLeastSquares: reduce the number of instantiations to
 reduce compile time PCL_XYZ_POINT_TYPES currently contains 18 types, so
 previously, MLS was instantiated for 18*18=324 different type combinations.
 However, among those were instantiations which are likely used by nobody
 (like `pcl::MovingLeastSquares<pcl::PointWithRange, pcl::PointDEM>`). With
 these changes, MLS is only instantiated 6*6+(18-6)=48 times. The most common
 type combinations should be covered, but if someone uses an uncommon
 combinations, they have to add `#define PCL_NO_PRECOMPILE` before including
 `pcl/surface/mls.h` to avoid linker errors.

---
 surface/src/mls.cpp | 11 ++++++++++-
 1 file changed, 10 insertions(+), 1 deletion(-)

diff --git a/surface/src/mls.cpp b/surface/src/mls.cpp
index ba2fd655300..c948f4d19e1 100644
--- a/surface/src/mls.cpp
+++ b/surface/src/mls.cpp
@@ -80,6 +80,15 @@ pcl::MLSResult::calculatePrincipalCurvatures (const double u, const double v) co
   PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointXYZ)(pcl::PointXYZI)(pcl::PointXYZRGB)(pcl::PointXYZRGBA)(pcl::PointXYZRGBNormal)(pcl::PointNormal))
                                               ((pcl::PointXYZ)(pcl::PointXYZI)(pcl::PointXYZRGB)(pcl::PointXYZRGBA)(pcl::PointXYZRGBNormal)(pcl::PointNormal)))
 #else
-  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, (PCL_XYZ_POINT_TYPES)(PCL_XYZ_POINT_TYPES))
+  // PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, (PCL_XYZ_POINT_TYPES)(PCL_XYZ_POINT_TYPES))
+  // All instantiations that are available with PCL_ONLY_CORE_POINT_TYPES, plus instantiations for all XYZ types where PointInT and PointOutT are the same
+  #define PCL_INSTANTIATE_MovingLeastSquaresSameInAndOut(T) template class PCL_EXPORTS pcl::MovingLeastSquares<T,T>;
+  PCL_INSTANTIATE(MovingLeastSquaresSameInAndOut, PCL_XYZ_POINT_TYPES)
+  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointXYZ))((pcl::PointXYZI)(pcl::PointXYZRGB)(pcl::PointXYZRGBA)(pcl::PointXYZRGBNormal)(pcl::PointNormal)))
+  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointXYZI))((pcl::PointXYZ)(pcl::PointXYZRGB)(pcl::PointXYZRGBA)(pcl::PointXYZRGBNormal)(pcl::PointNormal)))
+  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointXYZRGB))((pcl::PointXYZ)(pcl::PointXYZI)(pcl::PointXYZRGBA)(pcl::PointXYZRGBNormal)(pcl::PointNormal)))
+  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointXYZRGBA))((pcl::PointXYZ)(pcl::PointXYZI)(pcl::PointXYZRGB)(pcl::PointXYZRGBNormal)(pcl::PointNormal)))
+  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointXYZRGBNormal))((pcl::PointXYZ)(pcl::PointXYZI)(pcl::PointXYZRGB)(pcl::PointXYZRGBA)(pcl::PointNormal)))
+  PCL_INSTANTIATE_PRODUCT(MovingLeastSquares, ((pcl::PointNormal))((pcl::PointXYZ)(pcl::PointXYZI)(pcl::PointXYZRGB)(pcl::PointXYZRGBA)(pcl::PointXYZRGBNormal)))
 #endif
 #endif    // PCL_NO_PRECOMPILE
 ```

**Comment by EstebanDugueperoux2** [comment]
Hi @maksim-petukhov @valgur @jcar87  , I get always some jobs killed even using change from https://github.com/PointCloudLibrary/pcl/pull/5764.
Are you sure pipelines are executed in a xlarge node?

**Comment by maksim-petukhov** [comment]
@EstebanDugueperoux2 You should probably try to set `PCL_ONLY_CORE_POINT_TYPES=True` . I suppose this will also [reduce number of instantiations](https://github.com/PointCloudLibrary/pcl/blob/c577fbe7b42fa0f013deb6571e50dd0d0c2fdff7/features/src/integral_image_normal.cpp#L46) and will help to avoid killing the build situation. btw they [use ](https://github.com/PointCloudLibrary/pcl/blob/c577fbe7b42fa0f013deb6571e50dd0d0c2fdff7/.ci/azure-pipelines/build/ubuntu.yaml#L12)this option in their CI pipeline

**Comment by EstebanDugueperoux2** [comment]
@jcar87 , @maksim-petukhov, @valgur do you have an idea what happens with these pipelines?
Regards.

**Comment by jcar87** [comment]
> @jcar87 , @maksim-petukhov, @valgur do you have an idea what happens with these pipelines? Regards.

Hi @EstebanDugueperoux2 - we are currently experiencing issues with our underlying infrastructure and and working towards a solution. Thank you for your patience!


As for earlier issues, we are indeed running these in the more capable VMs - I have noticed the issues now seem to be specific to clang, it may be worth replicating that locally to discard any issues with that specific compiler?  Thank you for following up on the patch to reduce the number of instantiations!

**Comment by EstebanDugueperoux2** [comment]
> > @jcar87 , @maksim-petukhov, @valgur do you have an idea what happens with these pipelines? Regards.
> 
> Hi @EstebanDugueperoux2 - we are currently experiencing issues with our underlying infrastructure and and working towards a solution. Thank you for your patience!
> 
> As for earlier issues, we are indeed running these in the more capable VMs - I have noticed the issues now seem to be specific to clang, it may be worth replicating that locally to discard any issues with that specific compiler? Thank you for following up on the patch to reduce the number of instantiations!

Hi @jcar87 , thanks for your feedback, I have seen status.conan.io notification.
Ok I leave all my PRs pending to not disturb you.
About the failing builds, as they are about Windows and I have only Linux workstation it is difficult to diagnostic locally.

Regards.

File: recipes/pcl/all/conanfile.py [hunk 8]
```diff
@@ -0,0 +1,252 @@
...
+        tc.set_property("flann", "cmake_file_name", "FLANN")
+        tc.set_property("flann", "cmake_target_name", "FLANN::FLANN")
+        tc.set_property("glew", "cmake_file_name", "GLEW")
+        if self.options.with_qhull:
+            tc.set_property("qhull", "cmake_file_name", "QHULL")
+        tc.generate()
+
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+        for mod in ["Eigen", "FLANN", "GLEW", "PCap", "Qhull", "libusb"]:
```
valgur, line 151:
```suggestion
        for mod in ["Eigen", "FLANN", "GLEW", "Pcap", "Qhull", "libusb"]:
```
Oops, my bad. Sorry about that. A downside of testing recipes on Windows. :grin:

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,567 @@
...
+        tc.cache_variables["WITH_PCAP"] = self._is_enabled("pcap")
+        tc.cache_variables["WITH_PNG"] = self._is_enabled("png")
+        tc.cache_variables["WITH_QHULL"] = self._is_enabled("qhull")
+        tc.cache_variables["WITH_QT"] = self._is_enabled("qt")
+        tc.cache_variables["WITH_VTK"] = self._is_enabled("vtk")
+        tc.cache_variables["WITH_CUDA"] = self._is_enabled("cuda")
+        tc.cache_variables["BUILD_CUDA"] = self._is_enabled("cuda")
+        tc.cache_variables["BUILD_GPU"] = self._is_enabled("cuda")
+        tc.cache_variables["WITH_SYSTEM_ZLIB"] = True
+        tc.cache_variables["PCL_ONLY_CORE_POINT_TYPES"] = True
```
maksim-petukhov, line 439: I think this should be an option. C3I can't build without this, yes, but that doesn't mean that others can't.

recipes/pcl/all/conanfile.py
```diff
@@ -86,6 +86,8 @@ class PclConan(ConanFile):
         # "with_rssdk": [True, False],
         # "with_rssdk2": [True, False],
         # "with_qvtk": [True, False],
+        # If to enable https://github.com/PointCloudLibrary/pcl/pull/5764 
+        "with_compilation_optimisation": [True, False],
```
valgur, line 90:
```suggestion
        # Enables https://github.com/PointCloudLibrary/pcl/pull/5764
        "instantiate_only_core_point_types": [True, False],
```
The option name could be made more specific, IMO. It's a bit vague otherwise.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,570 @@
...
+        # "with_dssdk": [True, False],
+        # "with_ensenso": [True, False],
+        # "with_fzapi": [True, False],
+        # "with_metslib": [True, False],
+        # "with_openni": [True, False],
+        # "with_openni2": [True, False],
+        # "with_rssdk": [True, False],
+        # "with_rssdk2": [True, False],
+        # "with_qvtk": [True, False],
+        # If to enable https://github.com/PointCloudLibrary/pcl/pull/5764
```
maksim-petukhov, line 89:
```suggestion
```
The comment is a little misleading. The option itself doesn't have a lot in common with the mentioned PR. PR reduces compilation complexity when `PCL_ONLY_CORE_POINT_TYPES` is `False`. So the default value `True` doesn't enables this patch. It is good to have it though for those who will set this option to `False` so I wouldn't delete the `0001-ReportFixAboutMemoryConsumptionDuringBuild.patch`. But in my opinion there should be no comment there at all. The comment can also be changed to the documentation of the option [there](https://github.com/PointCloudLibrary/pcl/blob/3ed96c246e5c873713ec670b895469d09149a552/cmake/pcl_options.cmake#L49)

EstebanDugueperoux2: Is it ok with my last commit?

maksim-petukhov: I think the build will fail now. Setting `PCL_ONLY_CORE_POINT_TYPES` to `True` was the whole idea to get C3I building pcl. I was talking only about removing the comment. I apologize if I confused you with how I phrased things.

EstebanDugueperoux2:
Hi @jcar87 ,

Do you know what occurs in these pipelines?
I see "Unexpected Error" label and according to doc (https://github.com/conan-io/conan-center-index/blob/48d10319dedcc037ab7dc8452b691f4e76b57c5a/docs/labels.md#unexpected-error) these pipelines should be restarted but nothing occurs.

Regards.

EstebanDugueperoux2:
Thanks @jcar87 for the pipelines fixes.
All is green now, it miss a last reviewer to be able to merge this new recipe.

EstebanDugueperoux2:
Ping also @maksim-petukhov  and @memsharded  for a review.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,577 @@
...
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def system_requirements(self):
```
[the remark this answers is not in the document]
valgur: I would keep it until we have a working VTK package in CCI, such as one from this PR #10776. Might take a while, though, given the complexity of that package.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,577 @@
...
+                    )
+            for dep in self._internal_deps[component]:
+                if dep not in enabled_components:
+                    raise ConanInvalidConfiguration(
+                        f"'{dep}=True' is required when '{component}' is enabled."
+                    )
+
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
```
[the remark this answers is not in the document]
valgur: They are equivalent. I preferred not using `check_min_vs` since it keeps the code simpler. The only downside is that `Visual Studio` and `msvc` are listed separately in the min versions dict.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,577 @@
...
+        "with_opengl": True,
+        "with_openmp": False,
+        "with_pcap": True,
+        "with_png": True,
+        "with_qhull": True,
+        "with_qt": True,
+        "with_vtk": False,
+        # Enabled to avoid excessive memory usage during compilation in CCI
+        "precompile_only_core_point_types": True,
+        "add_build_type_postfix": False,
```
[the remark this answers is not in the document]
valgur: It avoids the headache of accounting for the different library suffixes depending on `build_type` and compiler. I did not see much value in keeping it enabled it with Conan, but it's easy enough to re-enable if necessary.

recipes/pcl/all/conanfile.py
```diff
@@ -0,0 +1,577 @@
...
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE.txt",
+             dst=os.path.join(self.package_folder, "licenses"),
+             src=self.source_folder)
+
+        cmake = CMake(self)
+        cmake.install()
```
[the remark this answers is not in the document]
EstebanDugueperoux2: Looking at https://c3i.jfrog.io/c3i/misc/logs/pr/18389/29-windows-visual_studio/pcl/1.13.1//f36e8de945c19577931147f7f0c36bdeef6741dc-build.txt we see that [PDB FILES NOT ALLOWED (KB-H017)] check is OK and we don't see pdb files installed by cmake install.
But I can add the missing line from the template to prevent PDBs installation.

Thanks for your review.

danimtb (approved):

jwillikers (approved):

PR #18391, opened 2023-07-06
Fix issue with building shared on windows

fcorso2016:
Upgrading onnxruntime 1.15.1 to fix bad paths on windows.

Currently, there's an issue with the fact that onnxruntime depends on Python to build the shared library. When it generates the path to the Python executable on Windows it uses backslashes as the path separators. This causes issues with CMake which sees those characters are the start of escape characters, causing an error with invalid escape characters. This fix performs a corrects replacing the backslashes with forward-slashes in order to allow CMake to build it correctly.

---


jcar87:
> Upgrading onnxruntime 1.15.1 to fix bad paths on windows and get the shared library binaries uploaded
> 
> * [x]  I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
> * [x]  I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
> * [x]  I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

Hi @fcorso2016 - thank you for your contribution. could you please complete the PR description with the errors that you get if this patch is not applied? thanks!

danimtb (dismissed):

fcorso2016:
> > Upgrading onnxruntime 1.15.1 to fix bad paths on windows and get the shared library binaries uploaded
> > 
> > * [x]   I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
> > * [x]   I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
> > * [x]   I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.
> 
> Hi @fcorso2016 - thank you for your contribution. could you please complete the PR description with the errors that you get if this patch is not applied? thanks!

Done. I managed to update the description.

jcar87:
> > > Upgrading onnxruntime 1.15.1 to fix bad paths on windows and get the shared library binaries uploaded
> > > 
> > > * [x]    I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
> > > * [x]    I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
> > > * [x]    I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.
> > 
> > 
> > Hi @fcorso2016 - thank you for your contribution. could you please complete the PR description with the errors that you get if this patch is not applied? thanks!
> 
> Done. I managed to update the description.

Hi @fcorso2016 - thank you, much appreciated.
I can see that there are changes in the recipe unrelated to the python issue in the description, that is the addition of the `onnx_no_static_registration`, but there is no information as to why this is needed so it's difficult to review.

Happy to approve the PR with only the change to fix the python path, and we can discuss the need of `onnx_no_static_registration` in a separate issue? Mind you that we would probably lean towards throwing an invalid configuration rather than set the option of a different package, but I can't comment without further information.

recipes/onnxruntime/all/conanfile.py
```diff
@@ -27,11 +27,13 @@ class OnnxRuntimeConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "with_xnnpack": [True, False],
+        "onnx_no_static_registration": [True, False],
```
Hopobcn, line 30:
```suggestion
```
I woudn't add a recipe option because (that I know) `onnxruntime` doesn't have a CMake knob to disable it's own registration.

So the best way to model this is to not add a recipe option, and force `onnx` to be compiled with `onnx:disable_static_registration=True`.

recipes/onnxruntime/all/conanfile.py
```diff
@@ -27,11 +27,13 @@ class OnnxRuntimeConan(ConanFile):
...
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "with_xnnpack": False,
+        "onnx_no_static_registration": False
```
Hopobcn, line 36:
```suggestion
```

recipes/onnxruntime/all/conanfile.py
```diff
@@ -60,6 +62,10 @@ def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
+        # We need this to avoid double registration by both onnx and onnxruntime
+        if self.options.onnx_no_static_registration:
+            self.options["onnx"].disable_static_registration = True
+
```
Hopobcn, line 68:
```suggestion
```

Hopobcn (changes requested):
Forcing options from dependencies in configure leads to very obscure behavior. Better handle this with `validate()`.


For @jcar87 , by default, both `onnxruntime` and `onnx` try to perform a static registration of the onnx operators schemas. Problem being that double registration generates an exception. See https://github.com/microsoft/onnxruntime/issues/8556#issuecomment-1005672965

danimtb (approved):

Hopobcn (approved):

jwillikers (approved):

jcar87 (approved):

jcar87:
Thanks @fcorso2016 - I tested this on Windows and indeed it was an issue if the path to the interpreter contained backslashes.

And thanks @Hopobcn  for providing additional context regarding the `disable_static_registration` - I agree that an approach in the `validate()` method would be more appropriate - this can be handled in a different PR. Please note that for pull requests that fix issues, we do request we are provided with enough information to validate this ourselves as part of the review process.

PR #17436, opened 2023-05-07
openmvg: workaround to fix "all shared" on Linux

SpaceIm:
Follow up of https://github.com/conan-io/conan-center-index/pull/16356#issuecomment-1529969598.

It's a workaround for https://github.com/conan-io/conan/issues/13560. I don't understand why this issue is ignored by conan team since it affects several recipes like opencv (which leads to these misleading workarounds: https://github.com/conan-io/conan-center-index/pull/16678#discussion_r1179649917).
@jcar87 pointed out in https://github.com/conan-io/conan/pull/13010#discussion_r1091019217 that current implementation was fragile.

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18400, opened 2023-07-07
openmvg: add with_jpeg option

maksim-petukhov:
Specify library name and version:  **openmvg/\***

The option allows to use libjpeg-turbo instead of libjpeg

---


valgur (approved):

recipes/openmvg/all/conanfile.py
```diff
@@ -30,13 +30,15 @@ class Openmvgconan(ConanFile):
         "with_openmp": [True, False],
         "with_avx": [False, "avx", "avx2"],
         "programs": [True, False],
+        "jpeg": ["libjpeg", "libjpeg-turbo"]
```
SpaceIm, line 33: Could you change this option name to be consistent with other CCI recipes please? There is already enough inconsistencies in CCI ;)

Also you can add mozjpeg as other CCI recipes already do.

valgur:
Could the default be switched to libjpeg-turbo as well, perhaps?

recipes/openmvg/all/conanfile.py
maksim-petukhov: Done

maksim-petukhov:
@valgur I don't think so. Other recipes have `libjpeg` as a default, so we shouldn't use `libjpeg-turbo` as a default there because it might result in a conflict.

recipes/openmvg/all/conanfile.py
maksim-petukhov: `"with_libjpeg":` - 10 recipes
`"with_jpeg":` - 21 recipes
`"jpeg":` - 5 recipes
Therefore I stopped on `with_jpeg`

recipes/openmvg/all/conanfile.py
```diff
@@ -66,7 +68,12 @@ def requirements(self):
         self.requires("coin-utils/2.11.6")
         self.requires("eigen/3.4.0", transitive_headers=True)
         self.requires("flann/1.9.2", transitive_headers=True, transitive_libs=True)
-        self.requires("libjpeg/9e")
+        if self.options.with_jpeg == "libjpeg":
+            self.requires("libjpeg/9e")
+        elif self.options.with_jpeg == "libjpeg-turbo":
+            self.requires("libjpeg-turbo/2.1.5")
```
valgur, line 74:
```suggestion
            self.requires("libjpeg-turbo/3.0.0")
```
Can probably bump the version soon after #18417 is merged.

ericLemanissier (approved):

maksim-petukhov:
v2 pipeline is waiting for #17436

jcar87 (approved):

maksim-petukhov:
@jcar87 There is a #17436 PR from @SpaceIm and I think your changes make it unnecessary. Should I mention #17436 in initial PR message?

SSE4 (approved):

PR #18411, opened 2023-07-07
android-ndk: add support to Apple Rosetta for r21d

kletoz:
Specify library name and version:  **android-ndk/r21d**

Run `ndk-build` with Rosetta in Apple Silicon, prefixing the command with `arch -x86_64`. Changed only `r21d`.

---


SSE4 (approved):

ericLemanissier (approved):

jcar87 (commented):
Hi @kletoz  - I have a question about this PR.
If the Android NDK is already compatible with arm64 on macOS, should it not already be properly compatible? What error is solved by forcing x86_64 mode? Would like to understand what problem this solves.

Also bear in mind that Rosetta is an optional macOS component that is only installed on demand, and that there have been indications that it may not be available in all countries. So perhaps I wouldn't go with the assumption that it _is_ available.

kletozan:
@jcar87 NDK is compatible with arm64 only in newer versions. This PR is changing only a old version, and could be used as example for someone else needing other older versions.

You can find more details [here](https://stackoverflow.com/a/69555276) and in the link there.

kletoz:
@jcar87 NDK is compatible with arm64 only in newer versions. This PR is changing only a old version, and could be used as example for someone else needing other older versions.

You can find more details [here](https://stackoverflow.com/a/69555276) and in the link there.

jcar87:
> @jcar87 NDK is compatible with arm64 only in newer versions. This PR is changing only a old version, and could be used as example for someone else needing other older versions.
> 
> You can find more details [here](https://stackoverflow.com/a/69555276) and in the link there.

Hi @kletoz for the clarification. I am confused, since:
* a new url is being added for `r21d` for arm64 on macOS - but you're saying that this version is not compatible with arm64? 
* the change you are proposing in the `build()` method is not being applied to an old version, but to _all_ versions of the android NDK, including the ones that are presumed to be compatible with arm64 without any workarounds.


Ir the NDK version `r21d` is _not_ compatible with arm64 on macOS, and users should use newer ones, I would be inclined to just mark that as an invalid configuration and not add any workarounds, if we don't know of the side effects. My understanding is that _newer_ versions of the Android NDK can be used to target older Android API levels, is this not the case? 

From https://github.com/android/ndk/issues/1299, r21d is never mention as one that would work on M1.  With this information we wouldn't be inclined to merge this PR - we are trying to avoid Conan Center supporting combinations or patches that _only_ exist in Conan Center, without proper validation. 
On the other hand, it should be possible for a consumer of the `android-ndk/r21d` to on an Apple Silicon based mac to just use the `x86_64` binaries, if the NDK itself (regardless of Conan), can be used in that way. Is that not the case?

kletoz:
@jcar87 We are currently using the change in this PR in a private Conan server. I thought it could be helpful to other users in the same situation to have access to that. For external reasons, we cannot upgrade NDK to a newer one that supports native arm64, without Rosetta.

I will close this request and keep our modified recipe version. When eventually we upgraded to a newer NDK hopefully I could get the recipe from Conan Center Index again.

jcar87:
Thanks @kletoz - we will continue monitoring this in case this is reported by more users, and we can revisit this in that case.
I'm inclined to believe that there won't be many users in this situation considering that `r21d` never properly supported M1 macs (so no reason to expect the Conan recipe to do so), and that it has been a few years and only new versions are currently supported by Android.

PR #18413, opened 2023-07-07
update msdf-atlas-gen to conan 2.0

Sahnvour:
**msdf-atlas-gen/1.2.2**


recipes/msdf-atlas-gen/all/conanfile.py
```diff
@@ -27,30 +34,31 @@ def requirements(self):
...
         cmakelists = os.path.join(
             self._source_subfolder, "CMakeLists.txt")
 
-        tools.replace_in_file(cmakelists,
+        replace_in_file(self, cmakelists,
                               "add_subdirectory(msdfgen)", "")
-        tools.save_append(cmakelists,
-                          "install(TARGETS msdf-atlas-gen-standalone DESTINATION bin)")
+        #tools.save_append(cmakelists,
+                          #"install(TARGETS msdf-atlas-gen-standalone DESTINATION bin)")
```
valgur, line 53: This should work
```suggestion
        save(cmakelists, "install(TARGETS msdf-atlas-gen-standalone DESTINATION bin)", append=True)
```

recipes/msdf-atlas-gen/all/conanfile.py
```diff
@@ -27,30 +34,31 @@ def requirements(self):
...
     def _configure_cmake(self):
         if self._cmake:
             return self._cmake
         self._cmake = CMake(self)
-        self._cmake.definitions["MSDF_ATLAS_GEN_BUILD_STANDALONE"] = True
-        self._cmake.configure()
+        self._cmake.configure(variables={
+            "MSDF_ATLAS_GEN_BUILD_STANDALONE": True
+        })
         return self._cmake
```
valgur, line 62: Replace this with `def generate(self):` containing a CMakeToolchain generator. See the cmake template under docs/ for an example.

recipes/msdf-atlas-gen/all/conanfile.py
```diff
@@ -12,8 +18,9 @@ class MsdfAtlasGenConan(ConanFile):
     description = "MSDF font atlas generator"
     topics = ("msdf-atlas-gen", "msdf", "font", "atlas")
     settings = "os", "arch", "compiler", "build_type"
+    package_type = "application"
 
-    generators = "cmake", "cmake_find_package_multi"
+    generators = "CMakeToolchain"
     exports_sources = ["CMakeLists.txt"]
```
valgur, line 24: If the project already uses CMake, then you should most likely delete the CMakeLists.txt in the recipe folder and drop this attribute. It's no longer needed with the new generators in Conan v2.

czoido:
Hi @Sahnvour,
Thanks a lot for contributing! To learn how to make v2 ready recipes and also compatible with v1, the package templates are always a good place to start: https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates
Also, the [creating package tutorial](https://docs.conan.io/2/tutorial/creating_packages.html) in the Conan Docs are useful to learn some basics regarding recipes.
I'll add some suggestions to your code :)

recipes/msdf-atlas-gen/all/conanfile.py
```diff
@@ -1,7 +1,13 @@
...
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import get, replace_in_file
+from conan.tools.cmake import CMake
+from conan.errors import ConanInvalidConfiguration
+from conan.errors import ConanException
+
 import os
 
-required_conan_version = ">=1.33.0"
+required_conan_version = ">=2.0"
```
czoido, line 10: For the moment we are also providing compatibility with v1, so depending on the features you use in the recipe this should be a conan 1.X version
```suggestion
required_conan_version = ">=1.53.0"
```

recipes/msdf-atlas-gen/all/conanfile.py
```diff
@@ -68,4 +76,4 @@ def package_info(self):
 
         bin_path = os.path.join(self.package_folder, "bin")
         self.output.info("Appending PATH environment variable: {}".format(bin_path))
-        self.env_info.PATH.append(bin_path)
+        self.runenv_info.PATH.append(bin_path)
```
czoido, line 79: This has a different syntax in v2, with this line the bin folder will be added to the [run environment](https://docs.conan.io/2/reference/tools/env/virtualrunenv.html), so to use the tool you'll have to activate the `conanrun.sh` and then it will be added to the PATH.
```suggestion
        self.runenv_info.prepend_path("PATH", bin_path)
```

recipes/msdf-atlas-gen/all/test_package/conanfile.py
```diff
@@ -1,4 +1,4 @@
-from conans import ConanFile, tools
+from conan import ConanFile, tools
```
czoido, line 1: This warning is because in general we try to only import specific tools we use in recipes instead of the whole package.

czoido:
Hi @Sahnvour,

I have created a PR to your repo with the changes to make this compatible with v2 and v1, please, check the changes and don't hesitate to ask any questions about them.: https://github.com/Sahnvour/conan-center-index/pull/1

Sahnvour:
> Hi @Sahnvour,
> 
> I have created a PR to your repo with the changes to make this compatible with v2 and v1, please, check the changes and don't hesitate to ask any questions about them.: [Sahnvour#1](https://github.com/Sahnvour/conan-center-index/pull/1)

Thanks for your help @czoido and @valgur, I had started going this way but you beat me to it :slightly_smiling_face:

recipes/msdf-atlas-gen/all/conanfile.py
```diff
@@ -12,60 +16,54 @@ class MsdfAtlasGenConan(ConanFile):
...
-        self._patch_sources()
-        cmake = self._configure_cmake()
+        cmake = CMake(self)
+        cmake.configure()
         cmake.build()
 
     def package(self):
-        self.copy("LICENSE.txt", dst="licenses", src=self._source_subfolder)
-        cmake = self._configure_cmake()
+        copy(self, pattern="LICENSE", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
```
czoido, line 57:
```suggestion
        copy(self, pattern="LICENSE.txt", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
```

Sahnvour:
I have an additional question : what's the intended way to find the built executable tool, `msdf-atlas-gen`, from outside the conan scripts ?
My build system relies on conan, it's calling `conan install` commands and consumes its pkg-config output, but it is not _driven_ by conan. I was able to find it on 1.x via `conan info` (admittedly it was not clean), which doesn't exist anymore. Should I use [deployers](https://docs.conan.io/2/reference/extensions/deployers.html) ?

czoido:
> I have an additional question : what's the intended way to find the built executable tool, `msdf-atlas-gen`, from outside the conan scripts ? My build system relies on conan, it's calling `conan install` commands and consumes its pkg-config output, but it is not _driven_ by conan. I was able to find it on 1.x via `conan info` (admittedly it was not clean), which doesn't exist anymore. Should I use [deployers](https://docs.conan.io/2/reference/extensions/deployers.html) ?

There are several options: 
- One of them could be using deployers to get the binaries out of the cache and have them in a fixed known location to invoke the executable
- Also, every time you do a `conan install`, call the `conanrun.sh` [environment scrip](https://docs.conan.io/2/reference/tools/env/virtualrunenv.html)t is generated, this script will add the folder where the binary is located to the `PATH` (this is because the `package_type` is `application`, so the binary folders are automatically added to that script). 
- The equivalent of what you did with Conan 1.x is calling `conan graph info --requires=msdf-atlas-gen/1.2.2 --format=json` and getting the information from there, but as you comment I think the first two are better ways of doing that.
Hope this helps!

PR #18414, opened 2023-07-07
cyclonedds: add support for type discovery option

wadehunk:
**cyclonedds/0.10.3**

Adds an option to enable type/topic discovery

---


recipes/cyclonedds/all/conanfile.py
```diff
@@ -131,20 +135,20 @@ def package(self):
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "CycloneDDS")
-        self.cpp_info.set_property("cmake_target_name", "CycloneDDS::ddsc")
+        self.cpp_info.set_property("cmake_target_name", "CycloneDDS::CycloneDDS")
```
RubenRBS, line 138: Can you give us more info into the rationality behind this change please? :)

Was the prior value wrong?

RubenRBS (commented):
Hi! Thanks for your contribution. I have a question into the second change to the recipe, I'd like to know a bit more into the reasoning for it to be able to properly review it, thanks!

recipes/cyclonedds/all/conanfile.py
wadehunk: This was changed to be more consistent in the target/component naming. The CycloneDDS::ddsc target is defined as a component.

prince-chrismc: That really required? This would be a breaking change for downstream recipes that were using `cyclonedds::ddsc` in their recipes

wadehunk: The cyclone::ddsc target is still defined as a component and is 100% backwards compatible. This just adds a top level target in preparation for providing additional components.

prince-chrismc: Oh I apologies, I mean the Conan component (but copied the wrong value).

This 

```py
def package_info(self):
     self.cpp_info.requires = ["cyclonedds::CycloneDDS"]
```

Would no longer work.

I do agree that the CMake targets are still correct 👍

wadehunk: Ah, I follow you now. I did not realize that it was possible to list requires like that on specific conan components. You are correct, this probably breaks that. I will revert that and just live it with being named in that fashion.

prince-chrismc: Thanks! trying to minimize breaking changes :)

https://github.com/conan-io/conan-center-index/pull/18414/files#diff-1f19ced64e4b1ee228a7ab56704e1bfeecef5cacf82dcb57e89bfcb5c8b581eeL159 has the old changed reference by the original comment so now the global is differ from the lib :)

With more cleanup we could remove the component since there's only 1 for the legacy 1.x generators

prince-chrismc (approved):

PR #18490, opened 2023-07-11
[protobuf-c] Only protobuf generator requires C++ support

uilianries:
Specify library name and version:  **protobuf-c/1.4.1**

* Enable shared build
* The `libprotobuf-c` is pure-C code
* The `protoc-gen-c` requires C++11 and Protobuf only
* The package only provides an application and a C-library. No cppstd and libcxx should be part of its package ID
* As we want to run protoc during the build, we need to use VirtualRunEnv:scope=build
* Reduce the patch to not customize test steps. We do not build tests.
* License is BSD-2-Clause: https://github.com/protobuf-c/protobuf-c/tree/master#contributing
* Add new patch `1.4.1-cmake-protobuf-defines.patch` that is related to https://github.com/protobuf-c/protobuf-c/pull/678. It fixes https://github.com/conan-io/conan-center-index/pull/18490#issuecomment-1631101286. When building protobuf-c as shared, it always expects libprotobuf as shared too, so when trying to link shared protobuf-c to static libprotobuf, it will not work because protobuf-c is injecting protobuf defines expected only when linking to shared <-> shared.

---


uilianries:
Everything working in Github Actions 🤔 : https://github.com/uilianries/conan-center-index/actions/runs/5541002016/jobs/10113823808

Issue #15101, opened 2023-01-04
[question] libsystemd/251.4: Fails when building for ARM platform


Tobulus:
### What is your question?

Building libsystemd for arch=ARM fails due to missing `libcrypt`. [This line](https://github.com/systemd/systemd/blob/34eebf08faf030ed73ad497fdeedbc8c7015fb5e/meson.build#L1046) in the meson build.

Is this expected (e.g. because libcrypt is a system lib)?

This can be fixed in my environment by installing the debian package `libcrypt1` but I'm interested whether this can/should be handled via conan.

Thanks!

uilianries:
@Tobulus It's an interesting case which needs more investigation. I know `libcrypt-dev` from Debian is actually `libxcrypt` project, which is already present in Conan Center: https://conan.io/center/libxcrypt

uilianries:
@Tobulus Could please share a **full** output log, with your failure?

Tobulus:
@uilianries Sure.

```
$ conan install . -pr:h $file_path -pr:b default --build missing -s build_type=Release

Configuration (profile_host):

[settings]

arch=armv8

build_type=Release

compiler=gcc

compiler.libcxx=libstdc++11

compiler.version=10.2

os=Linux

[options]

[build_requires]

[env]

CC=/usr/bin/aarch64-linux-gnu-gcc

CONAN_CMAKE_TOOLCHAIN_FILE=/workspaces/[…]/toolchain-aarch64.cmake

CXX=/usr/bin/aarch64-linux-gnu-g++

QEMU_LD_PREFIX=/usr/aarch64-linux-gnu

Configuration (profile_build):

[settings]

arch=x86_64

arch_build=x86_64

build_type=Release

compiler=gcc

compiler.libcxx=libstdc++11

compiler.version=10

os=Linux

os_build=Linux

[options]

[build_requires]

[env]

 

libsystemd/251.4: Not found in local cache, looking in remotes...

libsystemd/251.4: Trying with 'conan-dev-local'...

libsystemd/251.4: Trying with 'conancenter'...

Downloading conanmanifest.txt completed [0.87k]                                         

Downloading conanfile.py completed [8.35k]                                              

Downloading conan_export.tgz completed [0.49k]                                          

Decompressing conan_export.tgz completed [0.00k]                                         

libsystemd/251.4: Downloaded recipe revision 0

conanfile.py (...-logging/0.11.0-alpha+local.0): Installing package

Requirements

[...]

PR #18497, opened 2023-07-11
[libsystemd] add 253.6, add dependency to libcrypt

Tobulus:
Specify library name and version:  **libsystemd/253.6**

libsystemd's meson build script had a missing dependency declaration to `libcrypt`.

Problem was issued [here](https://github.com/systemd/systemd/issues/28161) and released in version `253.6`.

The missing dependency is specially a problem when [cross-compiling the library](https://github.com/conan-io/conan-center-index/issues/15101).

I don't know whether there will be more backports to other major versions. If not, this could be provided to other versions via patches.

---

### DIFF
--- a/recipes/libsystemd/all/conandata.yml
+++ b/recipes/libsystemd/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "253.6":
+    url: "https://github.com/systemd/systemd-stable/archive/v253.6.tar.gz"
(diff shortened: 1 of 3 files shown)


RubenRBS (approved):
Thanks a lot for linking back to the issue, makes reviewing work much more easy :)

czoido (approved):

PR #18513, opened 2023-07-12
adds range of openssl libs to have wider compatibility

jabbas:
jwt-cpp/0.6.0

This PR adds wider range of openssl libraries that can be used to build it

---

- [*] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [*] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [*] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

### DIFF
--- a/recipes/jwt-cpp/all/conanfile.py
+++ b/recipes/jwt-cpp/all/conanfile.py
@@ -24,7 +24,7 @@ class JwtCppConan(ConanFile):
         export_conandata_patches(self)
 
     def requirements(self):
-        self.requires("openssl/1.1.1s")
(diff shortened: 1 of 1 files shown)


patmantru:
This change breaks any recipe that attempts to use a range other than `[>1.1.1c,<1.1.1u]` for OpenSSL now that `1.1.1u` is the most recent release.

PR #18522, opened 2023-07-13
uvwasi: add recipe

toge:
Specify library name and version:  **uvwasi/***

uvwasi is required by wasm-micro-runtime with `libc=uvwasi`.

---


RubenRBS (dismissed):
Looks good! The patch seems to be a subset of the changes in the linked URL, but I think that's ok :)

Other than that, I like this simple recipe, congrats!

danimtb (dismissed):

recipes/uvwasi/all/patches/0.0.18-0001-fix-cmake.patch
```diff
@@ -0,0 +1,100 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index f71c36d..a92e629 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -14,13 +14,13 @@ cmake_dependent_option(UVWASI_BUILD_TESTS
+ 
+ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/out)
+ 
+-if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU")
++if(0)
```
jcar87, line 10: @toge, thanks for your contribution.

What is the motivation behind this change? The patch mentions https://github.com/nodejs/uvwasi/pull/210 but I cannot see this change there.

recipes/uvwasi/all/patches/0.0.18-0001-fix-cmake.patch
```diff
@@ -0,0 +1,100 @@
...
+   list(APPEND uvwasi_cflags -fvisibility=hidden --std=gnu89)
+   list(APPEND uvwasi_cflags -Wall -Wsign-compare -Wextra -Wstrict-prototypes)
+   list(APPEND uvwasi_cflags -Wno-unused-parameter)
+ endif()
+ 
+-if(APPLE)
++if(0)
```
jcar87, line 17: This is probably unnecessary - `CMAKE_MACOSX_RPATH` is `ON` by default if the cmake minimum required is recent enough, and `ON` is the value we want - I suspect that even without this change, the dylib would have an install name that starts with `@rpath`

toge: @jcar87 
Thank you for your review!
I disable this patch.

recipes/uvwasi/all/patches/0.0.18-0001-fix-cmake.patch
toge: Sorry for the wrong information.
uswasi#210 only adds the installer.

I split the patch file in two.

jcar87:
Thanks @toge ! It's still not clear why disabling visibility=hidden is needed?

toge:
@jcar87 
Sorry, I overlooked this.
When compiled with `visibility=hidden`, there is no uvwasi symbol in the shared library.
I think the only solution is to remove `visibility=hidden`.

jcar87:
> @jcar87 Sorry, I overlooked this. When compiled with `visibility=hidden`, there is no uvwasi symbol in the shared library. I think the only solution is to remove `visibility=hidden`.

I see @toge, thanks for providing a rationale behind this. In this case, and given that the upstream `uvwasi` repository already bundles a shared library, this seems like an issue in the upstream code that has nothing to do with Conan. My concern with the patch (that is, removing the hidden visibility compiler flag), is that if in the future this issue is addressed by the library authors by marking the exported symbols with the default visibility attribute, Conan Center would be generating a library that exports _all_ symbols. In this case before accepting this particular patch, I would report this issue to upstream first, since it appears their shared library is not usable due to a bug (and if that is the case, I'd rather this recipe not provide a shared library at all).

toge:
I agree with you.
Please wait a moment to do so.

RubenRBS:
Any updates @toge? :)

toge:
In gcc <= 10.0, uvwasi's test_package.cpp returns 134 error code.
It worked fine on uvwasi/0.0.18.

I'm investigating.
```
uvwasi/0.0.19 (test package): RUN: ./test_package
Segmentation fault (core dumped)

ERROR: uvwasi/0.0.19 (test package): Error in test() method, line 26
        ConanException: Error 139 while executing
```

PR #18525, opened 2023-07-13
Re-added support for gtest/1.8.1 as this is the latest version supported by qnx 7.0

klausholstjacobsen:
gtest/1.8.1

This is the latest version supported by qnx 7.0

---


jcar87:
Hi @klausholstjacobsen, thank you for your contribution. In which way are the current versions of gtest not compatible with qnx 7.0? Is qcc from the QNX toolchain unable compile versions newer than 1.8.1?

klausholstjacobsen:
When creating tests and linking to a gtest version newer than 1.8.1 it simply will not link.
QNX support has been looking into it confirmed that 1.8.1 is indeed the latest supported version. They have also stated that support for a newer gtest version will not be added to qnx 7.0! For that we must upgrade to QNX 7.1 which is a whole other can of worms!

jcar87:
Hi @klausholstjacobsen , thank you for your response.
I'm a bit confused here as to why QNX support would explicitly "not support" an open source library,  If I recall correctly the qcc compiler in QNX 7.0 is based on gcc 5.4, should still be able to build some of the versions we still maintain - is this not the case? Are there any specific linker issues/logs that could be attached to this PR?

I can also see that the QNX SDP appears to come with a version of GoogleTest: https://www.qnx.com/developers/docs/7.0.0/index.html#com.qnx.doc.ide.userguide/topic/writing_test_programs.html -  is it possible that the version from inside the QNX SDP is different than the one from Conan, and would you not be able to use the system-provided version instead, and that the incompatibility and lack of support has to do with the integration with between the QNX IDE and the version of GoogleTest? 

Also bear in mind that the Conan Center artifactory remote does _not_ remove older versions - removing them from the repository means that future changes are not validated against the more recent platform configurations (os/compiler versions), but the recipes for older versions can still be retrieved from the remote and built from source - is this not the case? I can see that there is already a recipe for 1.8.1.

klausholstjacobsen:
Hi @jcar87 

1. The gtest distributed with qnx sdp is 1.7.x. we want to go with the newest possible version.
2. The reason for not supporting >1.8.1 is that this is where google removed the ability to choose between googles tuple implementation or provide you own with whatever libc implemnetation you're going with.
3. I know 1.8.1 is indeed available in conan center, but with an old recipe not compliant with conan 2.0. So we cannot rebuilt gtest/1.8.1 with conan 2.0.

Regards
Klaus

klausholstjacobsen:
What is the procedure from here in order to get the PR merged?

jcar87 (approved):

jcar87:
> Hi @jcar87
> 
> 1. The gtest distributed with qnx sdp is 1.7.x. we want to go with the newest possible version.
> 2. The reason for not supporting >1.8.1 is that this is where google removed the ability to choose between googles tuple implementation or provide you own with whatever libc implemnetation you're going with.
> 3. I know 1.8.1 is indeed available in conan center, but with an old recipe not compliant with conan 2.0. So we cannot rebuilt gtest/1.8.1 with conan 2.0.
> 
> Here's the error I get when build the conancenter gtest/1.8.1 with conan 2.0:
> 
> image
> 
> Regards Klaus

Hi @klausholstjacobsen, thanks for providing more details. I think we can add gtest/1.8.1 again so that the recipe is compatible with Conan 2.0 - although we cant guarantee that it will remain a supported version in the long term, as it is old. However as I mentioned, once it's updated it will not be deleted from the Conan Center remote.

I can see that googletest moved to use std::tuple in this commit from 2018 - https://github.com/google/googletest/commit/7d3b73c85a42811309eac26e5cbe054c40b64785 

Out of curiosity, what are the build or linker errors in QNX 7.0? As far as I'm aware, the qcc from that version of QNX is based on gcc 5.4 - we still build gcc 5.x for Conan 1.x packages, and all versions of googletest seem to build correctly - so I'm surprised std::tuple is not usable in QNX?

ericLemanissier (approved):

Issue #18102, opened 2023-06-28
[package] asio-grpc: ability to not link any library


weatherhead99:
### Description

Due to an underlying issue in gRPC, infinite loops in programs can occur pretty easily if one accidentally links a program against both grpc++.so and grpc++_unsecure.so (see e.g. https://github.com/grpc/grpc/issues/24422, there are many roughly similar bugs that boil down to the same thing). 

Upstream asio-grpc provides the cmake variable "ASIO_GRPC_DISABLE_AUTOLINK" which, if set, causes the find module to not link against any grpc library. One of the (documented, e.g. in https://github.com/Tradias/asio-grpc/issues/61) uses of this is to choose yourself how to link against grpc from downstream build systems.

In the conan package for asio-grpc, this variable is (of course) not honored because it uses the autogenerated CMake module functionality built into the CMakeDeps generator. However, in order to get around this, I have ended up having to put extra code in my cmake files, like e.g. 

```
#NOTE: the below is to get around gRPC "infinite loop" bugs in their initialization
#if we are using asio-grpc directly from upstream this should work
set(ASIO_GRPC_DISABLE_AUTOLINK TRUE)
find_package(asio-grpc)

#however, if we're using the conan package, then the above is currently not supported.
#we can do the below ugly hack to mess around with the deps in that case
if(TARGET asio-grpc_DEPS_TARGET)
  message(WARNING "CONAN BUILD hack: patching asio-grpc deps...")
  set_target_properties(asio-grpc_DEPS_TARGET PROPERTIES INTERFACE_LINK_LIBRARIES
  "$<$<CONFIG:Release>:>;$<$<CONFIG:Release>:>;$<$<CONFIG:Release>:Boost::headers>")

endif()
```

note that this introduces a necessity to edit my own CMakeLists.txt file in order to work with conan, explicitly against the design goals of CMakeDeps/ CMakeToolchain.  I see a few of ways to resolve this:

1) instead of using autogenerated CMakeDeps, package the upstream asio-grpc Find module directly and install it into the conan package, for those using CMake downstream

2) somehow add a custom block to the CMakeDeps generating part which allows this variable to be used (I don't see an easy way to do this actually/ an obvious block where to put this)  **this is likely preferred but I don't immediately see how to do it**

[...]

PR #18529, opened 2023-07-13
Add option to choose grpc library to link to, bump version

weatherhead99:
Specify library name and version:  **asio-grpc/2.5.1**

addresses: #18102  by adding an option to choose which upstream grpc component we can link to. The default is grpc++_unsecure, which matches upstream package.

The version 2.5.1 is also now included.

The intention was to also allow linking to "no" component of grpc upstream (this would then be compatible with the cmake option ASIO_GRPC_DISABLE_AUTOLINK), however, if we do not include *something* in  self.cpp_info.requires for grpc we get an error. 

---

### DIFF
--- a/recipes/asio-grpc/all/conandata.yml
+++ b/recipes/asio-grpc/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "2.5.1":
+    url: "https://github.com/Tradias/asio-grpc/archive/refs/tags/v2.5.1.tar.gz"
+    sha256: "571779a25be6eed77a345088e3ded2cccf880c16800ee0075ece57116f14fc35"
   "2.4.0":
     url: "https://github.com/Tradias/asio-grpc/archive/refs/tags/v2.4.0.tar.gz"
     sha256: "d71de4f8de91dc0ad44d6a161fc628496b80622a6f9030dcd4c53b516629b8b7"
--- a/recipes/asio-grpc/all/conanfile.py
+++ b/recipes/asio-grpc/all/conanfile.py
@@ -57,7 +57,7 @@ class AsioGrpcConan(ConanFile):
             raise ConanInvalidConfiguration(f"{self.name} 'recycling_allocator' cannot be used in combination with the 'unifex' backend.")
 
     def requirements(self):
-        self.requires("grpc/1.50.1")
+        self.requires("grpc/1.50.1", transitive_headers=True)
         if self._local_allocator_option == "boost_container" or self.options.backend == "boost":
             self.requires("boost/1.81.0")
         if self.options.backend == "asio":
--- a/recipes/asio-grpc/all/conanfile.py
+++ b/recipes/asio-grpc/all/conanfile.py
@@ -22,12 +22,12 @@ class AsioGrpcConan(ConanFile):
     options = {
         "backend": ["boost", "asio", "unifex"],
         "local_allocator": ["auto", "memory_resource", "boost_container", "recycling_allocator"],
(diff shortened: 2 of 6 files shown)


jcar87:
I suspect `self.requires("grpc/1.50.1")` should have `transitive_headers=True` in Conan 2.0.

Tradias:
I don't quite understand why we need the `"none"` option. I think it is safe to assume that users who obtain asio-grpc through conan won't use the ASIO_GRPC_DISABLE_AUTOLINK variable, as it has no effect anyways (referring to https://github.com/conan-io/conan-center-index/issues/18102#issuecomment-1617858448).

On another note, wouldn't it be best to remove your newly added option and instead inspect `self.dependencies["grpc"].options.secure` and use that? If I create a conanfile like this;

```
[requires]
asio-grpc/2.4.0

[generators]
cmake_find_package_multi

[layout]
cmake_layout

[options]
grpc:secure=True
```
it will actually fail with 

```
conanfile.txt: ERROR: Generator cmake_find_package_multi(file:None) failed
Component 'grpc::grpc++_unsecure' not found in 'grpc' package requirement
ERROR: Component 'grpc::grpc++_unsecure' not found in 'grpc' package requirement
```
I think the asio-grpc recipe should just automatically pick grpc::grpc++ in that case.

weatherhead99:
Hiya,
yes indeed I was thinking the same however it turns out that in the *exact case* I myself need, the "none" option is required - my project has to build with *or without* conan, so I am indeed using GRPC_NO_AUTOLINK in my cmake code and then manually linking in grpc++. 

If I don't give the "none" option in conan, then I have to modify my CMake file to check whether GRPC was obtained by  conan, which if I already do I don't need this option in the first place.

weatherhead99:
as regards the "checking secure" option of upstream grpc, I did also consider that, however, it's not a 1:1 mapping. If you build grpc with "secure" as on, you just get grpc++ library. If you build it with "secure" as off, you get a choice of both (grpc++ and grpc++_unsecure).

Tradias:
I recommend taking the original asio-grpc recipe and change just this one line:

```python
self.cpp_info.requires = ["grpc::grpc++_unsecure"]
```

to 

```python
self.cpp_info.requires = ["grpc::grpc++"]
```

Why?

It seems that no other recipe handles the difference between gRPC secure and unsecure correctly. See for example google-cloud-cpp or opentelemetry-cpp. I think this is a problem with the gRPC recipe itself where `grpc++` should alias `grpc++_unsecure` when `grpc:secure=False` and it should not be addressed in dependent recipes.

It fixes the problem of `ERROR: Component 'grpc::grpc++_unsecure' not found in 'grpc' package requirement` that I mentioned earlier.

It fixes your problem.

It is a breaking change of this recipe but I do not know what the policy is regarding that in conan and I think it is ok. After all, if people use both `asio-grpc` and `opentelemtry-cpp` today then they would link with both `grpc++` and `grpc++_unsecure` and end up in the endless loop problem. In other words, asio-grpc is currently broken when other dependencies (from conan-center) that themselves depend on gRPC are used.

Tradias:
@weatherhead99 please take a look at https://github.com/conan-io/conan-center-index/pull/19070 and test whether it would fix your issue.

weatherhead99:
@Tradias I don't think it's fully ideal from the point of view of "not breaking existing usage and not having to change CMakeLists.txt files", but I also fully agree that the existing usage is quite a hack and I don't think that the case I described (where you actually _want_ both unsecure and secure) is more than a tiny tiny edge case, certainly I personally would never need this. Therefore (finally, sorry for the delay), closing this in lieu of your solution

PR #18376, opened 2023-07-06
Qt5: use layout

ericLemanissier:
Specify library name and version:  **qt/5.***

In an attempt to fix conan 2 windows failure

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18546, opened 2023-07-14
qarchive: add new version, bump deps, fix includedirs

maksim-petukhov:
Specify library name and version:  **qarchive/\***

This PR also makes QArchive header available as `#include <QArchive/QArchive>` (see [examples](https://github.com/antony-jr/QArchive/blob/e587f30507c0e6d92f79a2dc1a6aa7ebb1f8e679/examples/conan/main.cpp#L6))

---


ericLemanissier (approved):

recipes/qarchive/all/conanfile.py
```diff
@@ -104,7 +104,8 @@ def package_info(self):
         self.cpp_info.set_property("cmake_target_name", "QArchive")
         self.cpp_info.set_property("pkg_config_name", "QArchive")
         self.cpp_info.libs = ["QArchive"]
-        self.cpp_info.includedirs = [os.path.join("include", "QArchive")]
+        self.cpp_info.includedirs.append(os.path.join("include", "QArchive"))
+        self.cpp_info.requires = ["qt::qtCore", "libarchive::libarchive"]
```
ericriff, line 108: Why do we need to explicitly list the requires here?

maksim-petukhov: Because without that all qt components will be linked into final .so file and this will mess up `ldd` output (all Qt .so files will be mentioned there as dependencies for QArchive)

ericriff (approved):

ericriff:
That's a weird build error. I'm guessing we need the 2.0 builds to pass now. It's been a while since I pushed through a PR. 
Is there a Qt component missing on `self.cpp_info.requires = ["qt::qtCore", "libarchive::libarchive"]`?

maksim-petukhov:
> That's a weird build error. I'm guessing we need the 2.0 builds to pass now. It's been a while since I pushed through a PR. Is there a Qt component missing on `self.cpp_info.requires = ["qt::qtCore", "libarchive::libarchive"]`?

@ericriff I don't think that some Qt component is missing. Only `Core` [is used by QArchive ](https://github.com/antony-jr/QArchive/blob/e587f30507c0e6d92f79a2dc1a6aa7ebb1f8e679/CMakeLists.txt#L49)

ericriff:
Do you have any idea of what can be failing? Should we tag some of the conan folks?

maksim-petukhov:
> Do you have any idea of what can be failing? Should we tag some of the conan folks?

I think I've fixed it. We'll see

maksim-petukhov:
@uilianries @jcar87 @danimtb @franramirez688 Could someone please help me with v2 pipeline? The error is `ERROR: Missing prebuilt package for 'qt/5.15.10''`. How is this possible and what should I do?

ericriff:
I tried to see if there is a package with these settings / options / version available on conan.io and the site complains about it not compatible with 2.0
Maybe the qt recipe is not fully compatible with 2.0 and it needs an update
image

maksim-petukhov:
Thank you @ericriff. Qt recipe is not conan-v2-ready now. Waiting for #18376 then

maksim-petukhov:
@uilianries @jcar87 @danimtb @franramirez688 @RubenRBS this PR is not stale, could someone please review this?

RubenRBS:
Hi @maksim-petukhov thanks a lot for your patience, I've now added this to my review backlog next monday :)

maksim-petukhov:
@RubenRBS "next Monday" has already passed...

franramirez688 (approved):

PR #18551, opened 2023-07-15
[xapian-core] Use util-linux-libuuid instead of deprecated libuuid

samuel-emrys:
Transition from deprecated libuuid recipe to util-linux-libuuid, which is actively maintained. See https://github.com/conan-io/conan-center-index/pull/17664 for discussion

The util-linux-libuuid package relies on an actively maintained project, see https://github.com/conan-io/conan-center-index/pull/17664.
The libuuid package uses a stale project. Use the active project instead to fix conflicts.

This should be part of a group of packages merged rapidly. See https://github.com/conan-io/conan-center-index/pull/17995#issuecomment-1614987627 for discussion on challenges involved in a previous migration attempt.

---

### DIFF
--- a/recipes/xapian-core/all/conanfile.py
+++ b/recipes/xapian-core/all/conanfile.py
@@ -56,7 +56,7 @@ class XapianCoreConan(ConanFile):
     def requirements(self):
         self.requires("zlib/1.2.13")
         if self.settings.os != "Windows":
-            self.requires("libuuid/1.0.3")
+            self.requires("util-linux-libuuid/2.39")
 
     def validate(self):
         if self.options.shared and self.settings.os == "Windows":


jwillikers (approved):

ericLemanissier (approved):

jcar87 (approved):

mgehre-amd:
This has broken some of our builds, as we also depend on xapian-core/1.4.19 and libuuid. Is there a way to get the older version of xapian-core/1.4.18 which works with libuuid?

samuel-emrys:
> This has broken some of our builds, as we also depend on xapian-core/1.4.19 and libuuid. Is there a way to get the older version of xapian-core/1.4.18 which works with libuuid?

@mgehre-amd you should be able to override it in your tree with

```python
self.requires("libuuid/1.0.3", override=True)
```

Is there a reason you can't replace `libuuid` with `util-linux-libuuid` though?

Alternatively, yes, you can use the previous revision of the recipe, you'll just have to require it explicitly

mgehre-amd:
@samuel-emrys, thanks for responding so quickly. Where can I find the previous revision of the recipe? https://conan.io/center/recipes/xapian-core?version=1.4.19 seems to only list the latest one.

We will work on replacing libuuid by util-linux-libuuid eventually, but currently builds fail because `find_package(libuuid)` doesn't work with `util-linux-libuuid`.

samuel-emrys:
@mgehre-amd [it should](https://github.com/samuel-emrys/conan-center-index/blob/e05300c58b46b5dd6b929315b9d371243e776789/recipes/util-linux-libuuid/all/conanfile.py#L113). Are you using conan 1.x? which generator?

mgehre-amd:
> @mgehre-amd [it should](https://github.com/samuel-emrys/conan-center-index/blob/e05300c58b46b5dd6b929315b9d371243e776789/recipes/util-linux-libuuid/all/conanfile.py#L113). Are you using conan 1.x? which generator?
I'm using conan 1.60.2 with
```
[generators]
cmake
cmake_find_package
cmake_paths
```
I see `Findutil-linux-libuuid.cmake` in my build directory.

samuel-emrys:
Ah, yes - at the moment it only supports the modern generators for that find mode - `CMakeDeps`

samuel-emrys:
@mgehre-amd I'm not sure how to search for previous revisions with conan 1 - I assume the revisions haven't changed in conan 2 though. You can find them like this:

```
$ conan list "xapian-core/1.4.19#*" -r conancenter
conancenter
  xapian-core
    xapian-core/1.4.19
      revisions
        7a53af51867537057638eb8efc7e7c42 (2022-03-17 02:17:07 UTC)
        ff04ba9f37d90ddd5c29266a0f16e5c3 (2022-04-24 09:04:59 UTC)
        1b0e0ff1df9b8d0d426e29333532fdf3 (2022-08-31 08:46:13 UTC)
        60856b4e1c6680459fd9bc1c652f7cb7 (2022-11-17 12:33:55 UTC)
        e9b2e8ef1660f47fc5bcd7a05ec1a8e9 (2023-03-08 10:52:56 UTC)
```
So, I guess you'd have similar to the following (you might have to experiment a bit):

```python
self.requires("xapian-core/1.4.19#60856b4e1c6680459fd9bc1c652f7cb7")
```
But as I said, you could equally just force `libuuid` to be used instead of `util-linux-libuuid` with

```python
self.requires("libuuid/1.0.3", override=True)
```

jcar87:
Thanks @mgehre-amd and @samuel-emrys  - please continue the discussion here https://github.com/conan-io/conan-center-index/issues/19119 as it makes it easier for us to track this.

PR #18558, opened 2023-07-15
[cpython] Use util-linux-libuuid instead of deprecated libuuid

samuel-emrys:
Transition from deprecated libuuid recipe to util-linux-libuuid, which is actively maintained. See https://github.com/conan-io/conan-center-index/pull/17664 for discussion

The util-linux-libuuid package relies on an actively maintained project, see https://github.com/conan-io/conan-center-index/pull/17664.
The libuuid package uses a stale project. Use the active project instead to fix conflicts.

This should be part of a group of packages merged rapidly. See https://github.com/conan-io/conan-center-index/pull/17995#issuecomment-1614987627 for discussion on challenges involved in a previous migration attempt.

---

### DIFF
--- a/recipes/cpython/all/conanfile.py
+++ b/recipes/cpython/all/conanfile.py
@@ -187,7 +187,7 @@ class CPythonConan(ConanFile):
                 self.requires("mpdecimal/2.5.0")  # FIXME: no 2.5.1 to troubleshoot apple
         if self.settings.os != "Windows":
             if not tools.is_apple_os(self.settings.os):
-                self.requires("libuuid/1.0.3")
+                self.requires("util-linux-libuuid/2.39")
             self.requires("libxcrypt/4.4.25")
         if self.options.get_safe("with_bz2"):
             self.requires("bzip2/1.0.8")


samuel-emrys:
This is blocked by https://github.com/conan-io/conan-center-index/pull/18550, which is pulled in via the `tk` recipe causing a graph conflict with `libuuid`

ericLemanissier (approved):

jcar87:
Relaunched the build jobs for this PR now that fontconfig has been updated - that should resolve the conflict, but will monitor this.

samuel-emrys:
This is still necessary, but it looks like there's more broken with the cpython recipe that's blocking this issue. This might have to be incorporated into a conan 2 migration when the recipe is in a better state.

Ahajha:
> This is still necessary, but it looks like there's more broken with the cpython recipe that's blocking this issue. This might have to be incorporated into a conan 2 migration when the recipe is in a better state.

@valgur and I are working on the cpython recipe migration, but I've separated the dependency changes into a separate PR: https://github.com/conan-io/conan-center-index/pull/21426

samuel-emrys:
> @valgur and I are working on the cpython recipe migration, but I've separated the dependency changes into a separate PR: #21426

yes, this is why i tagged this PR in that issue - my observation above was that changes beyond just changing the dependency were required to enable the CI to pass, hence my suggestion that this should be incorporated into a conan 2 migration rather than segmented out on its own.

I might be wrong about that, but conan staff are slow enough to respond that it became intractable for me to troubleshoot in a reasonable amount of time which is why i closed it. Feel free to continue to attempt to resolve this issue with the same approach I took here, but my opinion is that your energy is probably better spent just getting the migration working.

PR #18556, opened 2023-07-15
[aaf] Use util-linux-libuuid instead of deprecated libuuid

samuel-emrys:
Transition from deprecated libuuid recipe to util-linux-libuuid, which is actively maintained. See https://github.com/conan-io/conan-center-index/pull/17664 for discussion

The util-linux-libuuid package relies on an actively maintained project, see https://github.com/conan-io/conan-center-index/pull/17664.
The libuuid package uses a stale project. Use the active project instead to fix conflicts.

This should be part of a group of packages merged rapidly. See https://github.com/conan-io/conan-center-index/pull/17995#issuecomment-1614987627 for discussion on challenges involved in a previous migration attempt.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18559, opened 2023-07-15
[util-linux-libuuid] Change CMake target from LibUUID::LibUUID to libuuid::libuuid

### LINKED ISSUE #18554: [package] util-linux-libuuid uses wrong cmake target
### Description

In the following lines, the `util-linux-libuuid` recipe sets the cmake target to be `LibUUID::LibUUID` with a filename of `LibUUID-config.cmake`:

https://github.com/conan-io/conan-center-index/blob/61c4f7819e6cd3594a57f6c3847f94ab86de623f/recipes/util-linux-libuuid/all/conanfile.py#L112-L113

This was based on the internal practice that Kitware has for their internal libuuid cmake module, however this is not public and a number of packages (czmq, cppcommon) seem to assume a `libuuid::libuuid` target. These change should be reverted such that these packages can utilise util-linux-libuuid without a requirement to be patched.

N/A

### Conan profile

N/A

### Steps to reproduce

N/A

### Logs

```
Put your log output here
```

</details>

samuel-emrys:
This is in line with what conan recipes seem to expect, which will minimise the number of patches required. LibUUID was chosen as the target name given this is what's used internally within cmake, but this isn't a public interface and so there is no common usage to take guidance from.

See discussion in original ticket adding `util-linux-libuuid` after merge: https://github.com/conan-io/conan-center-index/pull/17664#discussion_r1249000485

The following tickets are blocked until this change is merged:

* https://github.com/conan-io/conan-center-index/pull/18552
* https://github.com/conan-io/conan-center-index/pull/18553
* https://github.com/conan-io/conan-center-index/pull/18556
* https://github.com/conan-io/conan-center-index/pull/18557

Closes #18554

---

### DIFF
--- a/recipes/util-linux-libuuid/all/conanfile.py
+++ b/recipes/util-linux-libuuid/all/conanfile.py
@@ -109,7 +109,7 @@ class UtilLinuxLibuuidConan(ConanFile):
 
     def package_info(self):
         self.cpp_info.set_property("pkg_config_name", "uuid")
-        self.cpp_info.set_property("cmake_target_name", "LibUUID::LibUUID")
-        self.cpp_info.set_property("cmake_file_name", "LibUUID")
+        self.cpp_info.set_property("cmake_target_name", "libuuid::libuuid")
+        self.cpp_info.set_property("cmake_file_name", "libuuid")
         self.cpp_info.libs = ["uuid"]
         self.cpp_info.includedirs.append(os.path.join("include", "uuid"))
--- a/recipes/util-linux-libuuid/all/conanfile.py
+++ b/recipes/util-linux-libuuid/all/conanfile.py
@@ -111,5 +111,8 @@ class UtilLinuxLibuuidConan(ConanFile):
         self.cpp_info.set_property("pkg_config_name", "uuid")
         self.cpp_info.set_property("cmake_target_name", "libuuid::libuuid")
         self.cpp_info.set_property("cmake_file_name", "libuuid")
+        # Maintain alias to `LibUUID::LibUUID` for previous version of the recipe
+        self.cpp_info.set_property("cmake_target_aliases", ["LibUUID::LibUUID"])
+
         self.cpp_info.libs = ["uuid"]
         self.cpp_info.includedirs.append(os.path.join("include", "uuid"))
--- a/recipes/util-linux-libuuid/all/test_package/CMakeLists.txt
+++ b/recipes/util-linux-libuuid/all/test_package/CMakeLists.txt
@@ -1,8 +1,8 @@
 cmake_minimum_required(VERSION 3.8)
 project(test_package LANGUAGES C)
 
-find_package(LibUUID REQUIRED CONFIG)
+find_package(libuuid REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.c)
-target_link_libraries(${PROJECT_NAME} PRIVATE LibUUID::LibUUID)
+target_link_libraries(${PROJECT_NAME} PRIVATE libuuid::libuuid)
 target_compile_features(${PROJECT_NAME} PRIVATE c_std_99)


recipes/util-linux-libuuid/all/conanfile.py
```diff
@@ -109,7 +109,7 @@ def package(self):
 
     def package_info(self):
         self.cpp_info.set_property("pkg_config_name", "uuid")
-        self.cpp_info.set_property("cmake_target_name", "LibUUID::LibUUID")
-        self.cpp_info.set_property("cmake_file_name", "LibUUID")
+        self.cpp_info.set_property("cmake_target_name", "libuuid::libuuid")
+        self.cpp_info.set_property("cmake_file_name", "libuuid")
```
SpaceIm, line 113:
```suggestion
```

Where does it come from? Why not just keep `recipename::recipename`, it's conan-center convention when a library doesn't have any official config file or Find module file.

samuel-emrys: This recipe provides the `libuuid` interface and replaces/deprecates the `libuuid` recipe.  The failing libraries for this migration have already been patched for `libuuid::libuuid`, so this is the smallest change that would get everything on CCI functional.

SpaceIm: That's not because it's the smallest change that it the good one. You can also use set_property() method of CMakeDeps in these recipes to override these names in order to satisfy these patches.

SpaceIm: Example if in downstream library there is a custom `Findlibuuid.cmake` and its CMakeLists.txt uses `find_package(libuuid)` and `libuuid::libuuid` target:

```python
    def generate(self):
        deps = CMakeDeps(self)
        deps.set_property("util-linux-libuuid", "cmake_find_mode", "both")
        deps.set_property("util-linux-libuuid", "cmake_file_name", "libuuid")
        deps.set_property("util-linux-libuuid", "cmake_target_name", "libuuid::libuuid")
        deps.generate()
```

samuel-emrys: I don't disagree. I put this up to push the issue forward but I'm not particularly invested in an individual solution. I'm not sure your suggestion for `recipename::recipename` is a good one here - we're talking about the target for an interface for which there is no standard. Ideally these three libuuid recipes would be completely substitutable (independently of how advisable that would be). `recipename::recipename` doesn't satisfy that. If anything along those lines, I think it should be `provides::provides`. Whether we change what's provided is another question I guess.

samuel-emrys: For clarity, the three libraries that I see implementing this interface are:

* https://git.kernel.org/pub/scm/fs/ext2/e2fsprogs.git/tree/lib/uuid (not in conan)
* https://github.com/util-linux/util-linux (this library)
* https://sourceforge.net/projects/libuuid/ (libuuid in conan)

See https://github.com/conan-io/conan-center-index/pull/17427#issuecomment-1545476420 for discussion of each. The original implementation is `uuid`

samuel-emrys:
For traceability, there is also relevant conversation about this in https://cpplang.slack.com/archives/C41CWV9HA/p1689414965702109

recipes/util-linux-libuuid/all/conanfile.py
samuel-emrys: > Example if in downstream library there is a custom `Findlibuuid.cmake` and its CMakeLists.txt uses `find_package(libuuid)` and `libuuid::libuuid` target:
> 
> ```python
>     def generate(self):
>         deps = CMakeDeps(self)
>         deps.set_property("util-linux-libuuid", "cmake_find_mode", "both")
>         deps.set_property("util-linux-libuuid", "cmake_file_name", "libuuid")
>         deps.set_property("util-linux-libuuid", "cmake_target_name", "libuuid::libuuid")
>         deps.generate()
> ```

I like this way of patching/modifying the conan defaults to what individual recipes require - i think it would be good to preference this over patching the build system files where possible. Having said that, given that we're talking about an implementation interface, as above, I think that the `provides` interface should be used instead of `recipename`, i.e.

```diff
-        deps.set_property("util-linux-libuuid", "cmake_find_mode", "both")
+        deps.set_property("uuid", "cmake_find_mode", "both") # or libuuid/LibUUID, whatever value `provides` has
```

samuel-emrys: I'm coming around to using `uuid` as the `provides` value. It's the name of the lib that's built by both recipes:

https://github.com/conan-io/conan-center-index/blob/60fa77f44eafacf99b7f1d43650f13dd8db5d971/recipes/util-linux-libuuid/all/conanfile.py#L114

 https://github.com/conan-io/conan-center-index/blob/60fa77f44eafacf99b7f1d43650f13dd8db5d971/recipes/libuuid/all/conanfile.py#L88

so perhaps the target name should be `uuid::uuid`

gegles (approved):
Hopefully we can merge this in quick so we can adjust our names internally and stabilize on this. thx!

jwillikers (approved):

gegles:
@samuel-emrys @SpaceIm could we make progress on this so as to finalize the target names so that every depending recipe can unify and stabilize around the final name. thx!

samuel-emrys:
@gegles this is waiting on input from a team member. @jcar87 has been assigned and @danimtb is pending review

jcar87:
Hi @samuel-emrys - thank you for your contribution, and for the detailed analysis in https://github.com/conan-io/conan-center-index/pull/17427#issuecomment-1545476420.

You may have been this already, but indeed it does appear to be the case that the libuuid implementation on most Linux distros is the one from _this_ recipe (`util-linux-libuuid`) and *not* the one from Conan Center's `libuuid` recipe. I think aligning the implementation of `libuuid` across all of Conan Center with what the distros use is sensible, especially as it appears that https://sourceforge.net/projects/libuuid/ is no longer well maintained. 

The conflicts with a system installed version are unfortunate, as it means that there are dependencies in potentially two places due to reliance on xorg/system (not that we can do much about it) - which makes me wonder whether a `libuuid-system` may be a better solution.


As for the CMake target name, this is rather unfortunate and caused by a defect in the `libuuid` recipe. When there is no known upstream name from a CMake-config, we fall back to well-known names from CMake built-in modules, and if none is available, we use the name of the library. So the `libuuid` recipe should have used `LibUUID::LibUUID`, much as this one currently does.

I can see that _because_ of the Conan-Center-only `libuuid::libuuid` target, there are patches in a few places, for example:
https://github.com/conan-io/conan-center-index/blob/master/recipes/cppcommon/all/patches/0001-update-cmakelists-1-0-0-0.patch#L29-L30

I think I would be inclined to keep `util-linux-libuuid` having the case-sensitive `LibUUID` cmake package name, and `LibUUID::LibUUID` targets (as per CMake's [upstream find module](https://gitlab.kitware.com/cmake/cmake/blob/e1409101c99f7a3487990e9927e8bd0e275f564f/Source/Modules/FindLibUUID.cmake)), but add a `provides` as you have mention.

As we update recipes to depend on `util-linux-libuuid` instead of of `libuuid`, we should be able to remove the workarounds that existed to support the names that were specific to the `libuuid` recipe, does this make sense?

jwillikers:
@jcar87 Is it possible for us to have both a system and non-system recipe? I have some code that uses libuuid directly and I'd prefer to have it tied to a specific version of this library as opposed to the system one if possible, even if it would potentially conflict with the libuuid used by the xorg system package.

samuel-emrys:
> As we update recipes to depend on `util-linux-libuuid` instead of of `libuuid`, we should be able to remove the workarounds that existed to support the names that were specific to the `libuuid` recipe, does this make sense?

Yep, that makes sense. Before we move on, could you just confirm the following for me?

> The conflicts with a system installed version are unfortunate, as it means that there are dependencies in potentially two places due to reliance on xorg/system (not that we can do much about it) - which makes me wonder whether a `libuuid-system` may be a better solution.

This feels a bit like giving up as a package manager - I think the solution that @prince-chrismc proposed here is a better one - introduce a dependency in the `xorg` recipe on the conan-center-index `uuid` package (this one). That way `xorg` won't conflict and will raise an error if someone tries to use another package in that graph.

> As for the CMake target name, this is rather unfortunate and caused by a defect in the `libuuid` recipe. When there is no known upstream name from a CMake-config, we fall back to well-known names from CMake built-in modules, and if none is available, we use the name of the library. So the `libuuid` recipe should have used `LibUUID::LibUUID`, much as this one currently does.

Can you confirm that you're aware that `LibUUID::LibUUID` is an _internal_ target that isn't propagated downstream to users, and is unlikely to become a public target? It's not clear to me that the same inheritence rules should apply, as discussed in the thread above. If you think this is the way to go regardless, that's cool, just want to make sure you're aware that this isn't a public target.

> I think I would be inclined to keep `util-linux-libuuid` having the case-sensitive `LibUUID` cmake package name, and `LibUUID::LibUUID` targets (as per CMake's [upstream find module](https://gitlab.kitware.com/cmake/cmake/blob/e1409101c99f7a3487990e9927e8bd0e275f564f/Source/Modules/FindLibUUID.cmake)), but add a `provides` as you have mention.

So, just to make sure I understand, all of the `uuid` recipes should have

```python
class LibUUID(ConanFile):
    provides = "uuid"
```
and a cmake target of `LibUUID::LibUUID`? You don't think a general rule of defaulting to `<provides>::<provides>` in the case of packages that provide an implementation of the `<provides>` interface is the preferable way to go? So in this case it would be `uuid::uuid`. This is consistent with the linking rules (i.e. `-luuid` links to `libuuid.so`), and also the current naming convention of `<packagename>::<packagename>`, if you consider that any standalone package implicitly `provides = <packagename>`, so it seems like a natural choice imo.

jcar87:
> This feels a bit like giving up as a package manager - I think the solution that @prince-chrismc proposed here is a better one - introduce a dependency in the xorg recipe on the conan-center-index uuid package (this one). That way xorg won't conflict and will raise an error if someone tries to use another package in that graph.

To an extent, this could be valid as well. However this would be an artificial dependency - `libuuid` is not a direct dependency of `xorg/system` - and the dependency on the specific "new" libuuid may be distro dependent as well - so recording a dependency on `util-linux-libuuid` may be OK for some distros or some versions, but not others, and may not be valid in the future of the transitive dependency ceases to be in the future. We would also be recording a dependency on a Conan package that is not used by that package itself in any way, shape, or form - we still we a `libuuid.so` from the system.  So this is not without risks - I can see how adding that dependency in `xorg/system` would have resulted in an error that would have failed fast instead of letting this get to the linking stage - but I would only advocate recording the dependency if we could have stricter guarantees on a version-tagged version of libuuid, as well as specific version - but we cannot even guarantee (over time), that it is a dependency at all.


> Can you confirm that you're aware that LibUUID::LibUUID is an internal target that isn't propagated downstream to users, and is unlikely to become a public target? It's not clear to me that the same inheritence rules should apply, as discussed in the thread above. If you think this is the way to go regardless, that's cool, just want to make sure you're aware that this isn't a public target.

I'm unsure what you are referring to as "public" target.  If it is a target defined in a `package_info` as the way to consume the library, it is a public target name, in the sense that is the name of the target that _consumers_ need to express in their `CMakeLists.txt`. If we make this change, then consumers need to change the target name they use themselves (if they do so in CMakeLists). I think that captures the public-ness. 

When users consume any of the 10 recipes that depend on `libuuid` (I've posted the list here: https://github.com/conan-io/conan-center-index/issues/19084#issuecomment-1667650579) - then the name of the target is less relevant to them _unless_ they depend on `libuuid` themselves. In this case, however, having noticed that some libraries had patches to use all-lowercase `libuuid::libuuid` (that existed only in Conan Center), instead of `LibUUID::LibUUID` (which uses the public name of the cmake find module) - we may as well take the opportunity to correct that situation, and follow the policy that we have been following in Conan Center more recently, which is the following:
A) If the library authors provide a CMake config file - use the name of the package and target to match this, else:
B) If there exists a CMake find module that is widely used - use these names, else:
C) use `recipename::recipename`

I think there is a strong case here that this is the (B) situation and that the `libuuid` should have used that  convention as well.  This should hopefully result in _less_ things to patch and reduce the complexity of consumer libraries overall - assuming that they are more likely to rely on the built-in CMake module anyway.

samuel-emrys:
> The most robust solution here would be to make sure that there exists a libuuid/system or similar, and have the ability to ensure that the entire dependency graph can depend on that. This ensures only 1 is used, and for a given system, that it is always the same version.

I'm not sure this is the _most_ robust solution. I see other problems with it. What if both a source version of `uuid` exists in the dependency tree along with the `system` package? This would create material issues for anybody who actually cares to link to `uuid` directly.

The _most_ robust solution imo would be to package `xorg` in conan. A big task, for sure, and I'm not necessarily proposing it - I don't have a good understanding of the limitations in doing so. I agree that the faux dependency solution has issues. I think practically though, as long as all linux distro's are using `util-linux-libuuid`, most of the potential issues are likely to be unrealised.

> I think there is a strong case here that this is the (B) situation and that the `libuuid` should have used that convention as well. This should hopefully result in _less_ things to patch and reduce the complexity of consumer libraries overall - assuming that they are more likely to rely on the built-in CMake module anyway.

I see where there's been miscommunication. The discussion above was related to this statement:

> which uses the public name of the cmake find module

`LibUUID::LibUUID` is not the public name, propagated by CMake. This is an internal find module that CMake itself uses, but does not make available publicly.

I think the options you have are good:

> A) If the library authors provide a CMake config file - use the name of the package and target to match this, else:
B) If there exists a CMake find module that is widely used - use these names, else:
C) use recipename::recipename

But I think (C) can be abstracted to `provides::provides` instead of `recipename::recipename`. lets use `libjpeg` and `jpeg-turbo` to illustrate - they both provide `jpeg`, that is, they both contain `provides = "jpeg"`. I'm not sure what the specific circumstances surrounding this interface is - if theres a public CMake find module that already exists for it, then the target name for that should be used (i.e., (B) above). **but**, if it **doesn't** exist, then using your definition of (C) would lead to different targets depending on the library, i.e.:

1. `libjpeg::libjpeg` for libjpeg
2. `jpeg-turbo::jpeg-turbo` for jpeg-turbo

What i'm proposing is that both of these should use `jpeg::jpeg`, which would make each library completely interchangeable for the other, from a build system perspective. To switch between the two, the only thing an author would have to do is change the `self.requires("...")` statement to point at the library of choice.

The way this applies to `libuuid` is that if `libuuid`, `util-linux-libuuid` and a hypothetical `uuid` (if a package gets created) all provide the `uuid` interface, i.e. have `provides = "uuid"`, then the cmake target they should present is `uuid::uuid` (again, notwithstanding the existence of (A) or (B) above). This would mean that from a build perspective, you could make these libraries completely substitutible by just changing the implementation you refer to in `self.requires()`.

To be clear, all would share the same CMake code:

```
find_package(uuid REQUIRED)
target_link_libraries(mylib uuid::uuid)
```

Does that make sense? Again, no problems if you don't like this approach - just want to make sure I'm getting my point across before moving on.

jcar87:
i see what you mean re: `FindLibUUID.cmake` - if this is internal to CMake then yes, any name is just as good or as bad as any. In that case, indeed I'd much rather it had the same name as the other `libuuid` and be `libuuid::libuuid` as this PR proposes. 

Happy to proceed with this PR, but out of abundance of caution I would also keep `LibUUID::LibUUID` as an alias in the interim, to avoid breaking well-intentioned users that are already relying on that name (assuming they have used it directly) - otherwise they will experience disruption when they update the recipe revision.

I agree that it should be sensible for "provides" to share the same target name for transparent drop-in replacement. I'm not sure libjpeg and jpeg-turbo are the most perfect example of this, since jpeg-turbo is not a drop-in replacement in all cases (v7 and v8 emulation is opt-in, jpeg-turbo is not compatible with libjpeg v9 features), and jpeg-turbo also has its own API that is not compatible with libjpeg (should one choose to use that).



> I'm not sure this is the most robust solution. I see other problems with it. What if both a source version of uuid exists in the dependency tree along with the system package? This would create material issues for anybody who actually cares to link to uuid directly - it seems like this is @jwillikers use case.

A system libuuid would also need a provides, and would also need to be incompatible with the other versions. A consumer can link against the system uuid coming from a libuuid conan packages that acts as proxy for the system one. The incompatibility would be there on purpose to avoid having different ones.


> The most robust solution imo would be to package xorg in conan. A big task, for sure, and I'm not necessarily proposing it - I don't have a good understanding of the limitations in doing so. I agree that the faux dependency solution has issues. I think practically though, as long as all linux distro's are using util-linux-libuuid, most of the potential issues are likely to be unrealised.

I was not involved when the `xorg/system` approach was considered - but I can see why such an approach would be applicable. For build purposes (compiling and linking), it would be perfectly fine to have a dedicated xorg package that has all the needed components. It's an **runtime** on *desktop* Linux where an X server is running, where we could perhaps have issues if the instance of the X server cannot interface correctly with the runtime of the application (for example, an X11 runtime that is newer than the one running on the system). Those problems don't disappear with the current solution, but the Docker images used by Conan Center are kept "old enough" on purpose such that on newer systems things work due to backwards compatibility - but I assume that if we were relying on xorg/system on a too recent Linux distro, libraries that link against it may not work properly on distros with older versions. It's worth revisiting this if this is not the case - but I think building the x11 runtime or statically linking against it is not something that is _usually_ done.

jcar87 (approved):

ericLemanissier (approved):

franramirez688 (approved):

samuel-emrys:
> I agree that it should be sensible for "provides" to share the same target name for transparent drop-in replacement. I'm not sure libjpeg and jpeg-turbo are the most perfect example of this, since jpeg-turbo is not a drop-in replacement in all cases (v7 and v8 emulation is opt-in, jpeg-turbo is not compatible with libjpeg v9 features), and jpeg-turbo also has its own API that is not compatible with libjpeg (should one choose to use that).

fair comment - I'm not particularly familiar with either. Glad to see this is moving forward.

PR #18570, opened 2023-07-16
[xorg] Model uuid dependency

samuel-emrys:
Prevent erroneous linkage against an incompatible uuid interface when libuuid exists in the dependency tree with xorg by adding a dependency on the util-linux-libuuid conan package. This will ensure an error is raised if libuuid is in the same dependency graph as xorg.

This has caused issues in #17427, #17485 and this change is based on the suggestion by @prince-chrismc in https://github.com/conan-io/conan-center-index/pull/17995#issuecomment-1615034107. Read these linked issues for context.

---

### DIFF
--- a/recipes/xorg/all/conanfile.py
+++ b/recipes/xorg/all/conanfile.py
@@ -23,6 +23,12 @@ class XorgConan(ConanFile):
     def package_id(self):
         self.info.clear()
 
+    def requirements(self):
+        # This is necessary to prevent other packages in the dependency tree from using the "libuuid" recipe
+        # which conflicts with the system installation of `util-linux-libuuid` that `xorg` installs. See #17427, #17485
+        # This has no functional bearing on xorg, however, as it links against the system library installed in util-linux-libs
+        self.requires("util-linux-libuuid/2.39")
+
     def system_requirements(self):
         apt = package_manager.Apt(self)
         apt.install(["libx11-dev", "libx11-xcb-dev", "libfontenc-dev", "libice-dev", "libsm-dev", "libxau-dev", "libxaw7-dev",


jwillikers (approved):

SSE4 (approved):

samuel-emrys:
@RubenRBS is this being considered at all? If not, I'll close this out - trying to minimise the number of plates I've got spinning for CCI contributions.

valgur:
I think this is relevant anymore thanks to `libuuid` being replaced by `util-linux-libuuid` everywhere by now?

samuel-emrys:
@valgur actually not entirely. This is about modelling the dependency in xorg itself so that it's enforced as part of the dependency tree, so that we get meaningful errors if any package in the tree uses a different libuuid. We've solved this problem for cci packages, but not anything outside this curated ecosystem

valgur (approved):

recipes/xorg/all/conanfile.py
```diff
@@ -23,6 +23,12 @@ def validate(self):
     def package_id(self):
         self.info.clear()
 
+    def requirements(self):
+        # This is necessary to prevent other packages in the dependency tree from using the "libuuid" recipe
+        # which conflicts with the system installation of `util-linux-libuuid` that `xorg` installs. See #17427, #17485
+        # This has no functional bearing on xorg, however, as it links against the system library installed in util-linux-libs
+        self.requires("util-linux-libuuid/2.39")
```
valgur, line 30:
```suggestion
        self.requires("util-linux-libuuid/[*]")
```
Could it be set to accept any version, maybe, to avoid the need for version bumps? (the pattern works, I tested it locally)

samuel-emrys: I'm not opposed to this provided it gets maintainer buy in. @uilianries @RubenRBS thoughts?

lucaskdc: Please check: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges

> Conan maintainers may introduce this for other dependencies over time. Outside of the cases outlined above, version ranges are not allowed in ConanCenter recipes.

PR #18579, opened 2023-07-16
fix(protobuf): package can't be created when cross building

stefanpantic-pdftools:
Specify library name and version:  **protobuf/all**

Package cannot be created when cross-building.

---

### DIFF
--- a/recipes/protobuf/all/test_package/conanfile.py
+++ b/recipes/protobuf/all/test_package/conanfile.py
@@ -17,12 +17,12 @@ class TestPackageConan(ConanFile):
         self.requires(self.tested_reference_str)
 
     def build_requirements(self):
-        if cross_building(self) and hasattr(self, "settings_build"):
+        if not cross_building(self):
             self.tool_requires(self.tested_reference_str)
 
     def generate(self):
         VirtualRunEnv(self).generate()
-        if cross_building(self) and hasattr(self, "settings_build"):
+        if cross_building(self):
             VirtualBuildEnv(self).generate()
         else:
             VirtualRunEnv(self).generate(scope="build")
@@ -31,11 +31,13 @@ class TestPackageConan(ConanFile):
         tc.generate()
 
     def build(self):
-        cmake = CMake(self)
-        cmake.configure()
-        cmake.build()
+        if not cross_building(self):
+            cmake = CMake(self)
+            cmake.configure()
+            cmake.build()
 
     def test(self):
-        if can_run(self):
+        if not cross_building(self) and can_run(self):
             bin_path = os.path.join(self.cpp.build.bindirs[0], "test_package")
             self.run(bin_path, env="conanrun")
+


jcar87:
Hi @stefanpantic-pdftools - thank you for your contribution.

Could you provide more details as to which command you are running to cross-build, the profiles involved, and which version of Conan? Thanks!

stefanpantic-pdftools:
@jcar87 Hey, I'm using Conan 2.0.7/2.0.8. The scenarios are:

* Build a macOS x86_64 binary on a armv8 Mac
* Build a macOS armv8 binary on a x86_64 Mac
* Build a Windows x86 binary on a x86_64 machine

Command:
```
conan create . --version 3.21.12 -pr:b macos_armv8 -pr:h macos_x86_64
```

The error is always the same, a missing prebuilt binary:

```
Requirements
    protobuf/3.21.12#e85ddb467331c7d03dbe0b3228c71a1f:07e169a2626c721c8fae5fe9e5b62723c954c4dd#244ef6cd7879111cfff30842dc61cf1c - Cache
    zlib/1.2.13#ae7d9893156acc4e8d2b0c66026c7d7dfe2ba542:63a8f578a509f72f9f398a08df40405f787ccf34#3ab21376c3c4e0edbecb371a23f11d1f - Cache
Build requirements
    emsdk/3.1.31#ae7d9893156acc4e8d2b0c66026c7d7dfe2ba542:b2c94cb8a3129b120e67cc3b3edf8c07ec710a83#05693ec02e065f26f099158a63bcb08f - Cache
    protobuf/3.21.12#e85ddb467331c7d03dbe0b3228c71a1f:c0ed29e744dbdac2801b7270ff35d342d91f6c4f - Missing
Skipped binaries
    nodejs/18.12.0, zlib/1.2.13

ERROR: Missing binary: protobuf/3.21.12:c0ed29e744dbdac2801b7270ff35d342d91f6c4f

protobuf/3.21.12: WARN: Can't find a 'protobuf/3.21.12' package binary 'c0ed29e744dbdac2801b7270ff35d342d91f6c4f' for the configuration:
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=17
compiler.libcxx=libc++
compiler.version=14
os=Macos
os.version=10.10
[options]
debug_suffix=True
fPIC=True
lite=False
shared=False
with_rtti=True
with_zlib=True
[requires]
zlib/1.2.Z

ERROR: Missing prebuilt package for 'protobuf/3.21.12'
Check the available packages using 'conan list protobuf/3.21.12:* -r=remote'
or try to build locally from sources using the '--build=protobuf/3.21.12' argument

```

jcar87:
Hi @stefanpantic-pdftools, thank you for your reply. I would say this is expected and intentional. The `test_package` for protobuf exercises the functionality where the protobuf compiler is invoked, and thus it needs to be able to locate a package for the _build_ profile (for the compiler) in addition to the one for the _host_ profile (for the library). 

You have a couple of options - you can first create the package for the build profile, followed by the one you want to cross-build:
```
conan create . --version 3.21.12 -pr:b macos_armv8 -pr:h macos_armv8
conan create . --version 3.21.12 -pr:b macos_armv8 -pr:h macos_x86_64
```
This will ensure that the test package exercises the complete functionality. 

Or if you wish to create the package without running the `test_package`, you can pass `-tf=` (`--test-folder=` to indicate an empty `test_folder` and thus skip the test_package altogether) - this will achieve what you're trying to do in this PR, which amounts to a no-op test package when cross-building, but I believe that the behaviour you are changing from the test package in this PR is still valuable to be able to test the entire workflow.

stefanpantic-pdftools:
@jcar87 but that doesn't seem like an acceptable solution, the package is then not buildable in isolation, here IMHO the test package is malformed as it requires a pre-existing package in a different configuration to exist so it passes.

I've observed other packages do a no-op in these situations.

The setup here also makes it impossible to create a package in CI using a build matrix as all steps there are executed in parallel and waiting for a different platform to finish/upload would just not be possible or a good solution.

jcar87:
> @jcar87 but that doesn't seem like an acceptable solution, the package is then not buildable in isolation, here IMHO the test package is malformed as it requires a pre-existing package in a different configuration to exist so it passes.

I've opened this issue in the Conan client: https://github.com/conan-io/conan/issues/14314 - as it is relevant to this case. Indeed, I appreciate it is counter intuitive that a test package may not work if  it depends on a different configuration of the tested reference. However there may be situations in which it is valid for a test_package to have other dependencies - any dependency will need to exist beforehand for the test package to work. In this case it is slightly different because `conan create` will ONLY generate a single package ID and `--build=missing` will work for other references, but not for the test_one if multiple package IDs are needed. This is what we are going to be looking into in https://github.com/conan-io/conan/issues/14314 

> I've observed other packages do a no-op in these situations.

Unfortunately some test_packages in Conan Center have taken an approach to work around the Conan Center workflows - but if you could point out to specific examples I can look into those.

> The setup here also makes it impossible to create a package in CI using a build matrix as all steps there are executed in parallel and waiting for a different platform to finish/upload would just not be possible or a good solution.

When building packages of multiple configurations, I would say that in general anything that is listed as a `tool_requires` will need to be built _first_, in the same way that in order for `conan create .` to succeed in this instance, a package for `zlib` must already exist before. Here it is no exception - and when cross building in particular there are some cases where you _may_ need to build and upload a reference before it can be used somewhere else. That is no different here than if it was any other dependency. We can provide guidance on how to derive the correct build order when building references in bulk like this.

jcar87:
> @jcar87 but that doesn't seem like an acceptable solution, the package is then not buildable in isolation, here IMHO the test package is malformed as it requires a pre-existing package in a different configuration to exist so it passes.
> 
> I've observed other packages do a no-op in these situations.
> 
> The setup here also makes it impossible to create a package in CI using a build matrix as all steps there are executed in parallel and waiting for a different platform to finish/upload would just not be possible or a good solution.

I appreciate it has been a while since this came up.

Please note that since then - we have taken the approach of simplifying test packages in Conan Center in order to have the ability to build things in isolation - we have been reworking some test_packages so as to not depend on multiple package_ids of the tested reference, trying our best to make sure we still sufficiently test without risking additional regressions.  It should now be possible to build these in isolation, get the correct package, and get a passing test_package where the test package can run.

PR #18580, opened 2023-07-16
functions-framework-cpp: add version 1.2.0, support conan v2

toge:
Specify library name and version:  **functions-framework-cpp/***

---


recipes/functions-framework-cpp/all/conanfile.py
```diff
@@ -22,64 +25,59 @@ class FunctionsFrameworkCppConan(ConanFile):
...
-        return {
-            "gcc": "9",
-            "Visual Studio": "15.7",
-            "clang": "7",
-            "apple-clang": "11",
-        }
+    def requirements(self):
+        self.requires("abseil/20230125.3")
+        self.requires("boost/1.82.0")
+        self.requires("nlohmann_json/3.11.2")
```
RubenRBS, line 61: Seems like a missing `transitive_headers=True` trait?
```suggestion
        self.requires("nlohmann_json/3.11.2", transitive_headers=True)
```

uilianries (approved):
These new patches adding nlohmann to CMakeLists makes sense, cpp files are using it but there is no target in cmake. I just opened a PR to upstream adding it: https://github.com/GoogleCloudPlatform/functions-framework-cpp/pull/397

SSE4 (approved):

PR #18651, opened 2023-07-19
imguizmo: migrate to Conan v2

valgur:


recipes/imguizmo/all/conanfile.py
```diff
@@ -1,60 +1,67 @@
...
 
-    def source(self):
-        tools.get(**self.conan_data["sources"][self.version],
-                  destination=self._source_subfolder, strip_root=True)
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("imgui/1.87")
+        self.requires("imgui/[<=1.89.3]", transitive_headers=True)
```
czoido, line 44: Hi @valgur! Thanks a lot for the contribution. We still are trying version ranges in c3i only for CMake and OpenSSL, so this would be better set to a fixed version: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges

czoido (approved):

SSE4 (approved):

PR #18652, opened 2023-07-19
wineditline: migrate to Conan v2

valgur:


valgur:

recipes/wineditline/all/CMakeLists.txt
```diff
@@ -1,8 +1,17 @@
...
+project(WinEditLine C)
 
-project(WinEditLineWrapper C)
+add_library(edit
+    src/editline.c
+    src/fn_complete.c
+    src/history.c
+    src/libedit.def
+)
+target_include_directories(edit PRIVATE src)
```
uilianries, line 10:
```suggestion
target_include_directories(edit PRIVATE src)
set_target_properties(edit PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
```

Export symbols on Windows to produce .DLL, only when building with BUILD_SHARED_LIBS:

valgur: Is it really needed? The shared library built and linked fine in CI.

uilianries: It does not generate dll at all: https://c3i.jfrog.io/c3i/misc/logs/pr/18652/5-windows-visual_studio/wineditline/2.206//d057732059ea44a47760900cb5e4855d2bea8714-build.txt

recipes/wineditline/all/CMakeLists.txt
```diff
@@ -1,8 +1,17 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.15)
+project(WinEditLine C)
 
-project(WinEditLineWrapper C)
+add_library(edit
```
uilianries, line 4: The upstream uses different names. Why not keeping the upstream cmake instead?

https://sourceforge.net/p/mingweditline/code/HEAD/tree/mingweditline/trunk/src/CMakeLists.txt

danimtb: ping @valgur

valgur: Don't remember at all anymore why I ignored the project CMakeLists.txt. I'll look into it.

perseoGI: Hi @valgur, thanks for your contributions, any update on this?

CommanderCRM:
hello, any chance it'll be reviewed? found out it's still not migrated while working migration branch is out

recipes/wineditline/all/CMakeLists.txt
CommanderCRM: You can use the original CMake with some additions, I'll provide them below. With these additions, you don't need to have any custom CMakeLists inside the package at all

recipes/wineditline/all/conanfile.py
```diff
@@ -1,64 +1,62 @@
...
+
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", dst=self.export_sources_folder, src=self.recipe_folder)
+
+    def configure(self):
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
```
CommanderCRM, line 37: src_folder is not needed here, works fine without it. Root folder contains CMakeLists

RubenRBS (approved):

recipes/wineditline/all/conanfile.py
```diff
@@ -1,64 +1,62 @@
...
-            tools.patch(**patch)
-        self._configure_cmake().build()
+        cmake.build()
 
     def package(self):
-        self.copy("COPYING", "licenses", self._source_subfolder)
-        self._configure_cmake().install()
+        copy(self, "COPYING", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
+        cmake = CMake(self)
+        cmake.install()
```
CommanderCRM, line 59: If we want to use original CMake, the following additions are needed (install method doesn't populate package correctly for tests to work instead):

` copy(self, "*.h", src=self.source_folder, dst=os.path.join(self.package_folder, "include", "editline"), keep_path=False)  `
`copy(self, "*.exe", src=self.build_folder, dst=os.path.join(self.package_folder, "bin"), keep_path=False)  `
`copy(self, "*.dll", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)  `
`copy(self, "*.lib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)  `

recipes/wineditline/all/conanfile.py
```diff
@@ -1,64 +1,62 @@
...
 
     def package_info(self):
         self.cpp_info.libs = ["edit"]
```
CommanderCRM, line 62: original cmake variant (dll will be linked without it instead, which causes an exception when running tests):
`self.cpp_info.components["edit"].libs = ["edit_static"]`

recipes/wineditline/all/test_package/CMakeLists.txt
```diff
@@ -1,9 +1,6 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.15)
 project(test_package C)
 
-include("${CMAKE_BINARY_DIR}/conanbuildinfo.cmake")
-conan_basic_setup(TARGETS)
-
 find_package(wineditline REQUIRED CONFIG)
 
 add_executable(test_package test_package.c)
```
CommanderCRM, line 6: (refers to line 7) if original cmakelists is to be used:
`target_link_libraries(test_package PRIVATE wineditline::edit)`

in addition with `self.cpp_info.components["edit"].libs = ["edit_static"]` in recipe package_info()

CommanderCRM (changes requested):

recipes/wineditline/all/conanfile.py
RubenRBS: This is a CCI convention, src is expected to be the source folder for every package :)

RubenRBS (approved):

PR #18687, opened 2023-07-19
Add geogram recipe

EstebanDugueperoux2:
Specify library name and version:  **geogram/1.8.4**

---

### DIFF
--- a/recipes/geogram/all/conandata.yml
+++ b/recipes/geogram/all/conandata.yml
@@ -1,0 +1,11 @@
+sources:
+  "1.8.4":
+    # url: "https://github.com/BrunoLevy/geogram/archive/refs/tags/v1.8.4.tar.gz"
+    # sha256: "5453385c644b1a3411a0723281f40bef9cf9165984d0e0b08153d32508a1994b"
+    url: "https://github.com/BrunoLevy/geogram/releases/download/v1.8.4/geogram_1.8.4.tar.gz"
+    sha256: "98e9c2de1dfad795296ea62262c6b3d74000ef5d2cd903af33189592ea0a8a6f"
+patches:
+  "1.8.4":
+    - patch_file: "patches/0001-Map-VORPALINE_BUILD_DYNAMIC-to-BUILD_SHARED_LIBS.patch"
+      patch_description: "correct the order of cmake min and project"
+      patch_type: "conan"
--- a/recipes/geogram/all/conanfile.py
+++ b/recipes/geogram/all/conanfile.py
@@ -1,0 +1,147 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.microsoft import check_min_vs, is_msvc
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, copy, rm, rmdir
+from conan.tools.build import check_min_cppstd
+from conan.tools.scm import Version
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+import os
+
+required_conan_version = ">=1.53.0"
+
+class GeogramConan(ConanFile):
+    name = "geogram"
+    description = "a programming library with geometric algorithms"
+    license = "BSD 3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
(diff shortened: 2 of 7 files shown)


recipes/geogram/all/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+from conan.tools.scm import Version
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+import os
+
+required_conan_version = ">=1.53.0"
+
+class GeogramConan(ConanFile):
+    name = "geogram"
+    description = "a programming library with geometric algorithms"
+    license = "BSD 3-Clause"
```
uilianries, line 15:
```suggestion
    license = "BSD-3-Clause"
```
Please use the official SPDX short name only: https://spdx.org/licenses/BSD-3-Clause.html

recipes/geogram/all/conandata.yml
```diff
@@ -0,0 +1,11 @@
+sources:
+  "1.8.4":
+    # url: "https://github.com/BrunoLevy/geogram/archive/refs/tags/v1.8.4.tar.gz"
+    # sha256: "5453385c644b1a3411a0723281f40bef9cf9165984d0e0b08153d32508a1994b"
```
uilianries, line 4:
```suggestion
```
Yes, prefer release always. Prefer, remove that link to avoid confusion.

recipes/geogram/all/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        check_min_vs(self, 191)
+        if not is_msvc(self):
+            minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+            if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+                raise ConanInvalidConfiguration(
+                    f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+                )
+        # in case it does not work in another configuration, it should validated here too
```
uilianries, line 78: please, pay attention when using the template to avoid using everything from there. This condition is an example, not a rule, most of projects should work shared + windows .

recipes/geogram/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,278 @@
+// Copied from https://github.com/BrunoLevy/geogram/blob/main/src/tests/test_RVC/main.cpp
```
uilianries, line 1: Please, simplify it to a small test, link consuming a method that print the library version. The intention is testing the package, something small and functional. Not running examples neither unit tests that are more complex to maintain.

recipes/geogram/all/conandata.yml
```diff
@@ -0,0 +1,11 @@
...
+    url: "https://github.com/BrunoLevy/geogram/releases/download/v1.8.4/geogram_1.8.4.tar.gz"
+    sha256: "98e9c2de1dfad795296ea62262c6b3d74000ef5d2cd903af33189592ea0a8a6f"
+patches:
+  "1.8.4":
+    - patch_file: "patches/0001-Map-VORPALINE_BUILD_DYNAMIC-to-BUILD_SHARED_LIBS.patch"
```
uilianries, line 9: Could you please open a PR to the upstream with your patch? So we don't need to re-patch in case of a new release.

recipes/geogram/all/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+        "shared": False,
+        "fPIC": True,
+        "with_graphics": False,
+    }
+
+    short_paths = True
+
+    @property
+    def _min_cppstd(self):
+        return 11
```
uilianries, line 35: Are you sure this project requires C++11? Running grep I only found test, example and third party using C++11/14/17

EstebanDugueperoux2: then by default I should put "98"?

uilianries: Just skip and don't check anything. We usually add it to avoid breaking with new compilers.

EstebanDugueperoux2:
I close this PR because geogram use git submodules and many of its requirements don't follows semantic versionning.

PR #18709, opened 2023-07-20
cpp-ipc: migrate to Conan v2

valgur:


czoido (approved):

recipes/cpp-ipc/all/conanfile.py
```diff
@@ -21,64 +29,67 @@ class CppIPCConan(ConanFile):
...
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
     def validate(self):
-        if tools.is_apple_os(self.settings.os):
-            raise ConanInvalidConfiguration("{} does not support Apple platform".format(self.name))
+        if self.settings.os == "Macos":
```
jwillikers, line 57: Why not use [conan.tools.apple.is_apple_os](https://docs.conan.io/2/reference/tools/apple/other.html?highlight=is_apple_os#conan.tools.apple.is_apple_os)?

jwillikers (changes requested):

recipes/cpp-ipc/all/conanfile.py
valgur: No real reason, I mistakenly assumed the function has been deprecated in v2. Thanks for mentioning!

jwillikers: Yeah, I figured it wouldn't work on iOS etc, too. Thanks for updating this!

jwillikers (approved):

czoido (approved):

PR #18715, opened 2023-07-20
scc: migrate to Conan v2

valgur:


memsharded:
I am working on this, but still work to do. This PR is beyond 2.0 support, because it is fixing the previous recursive installation of packages hardcoded in scc CMakeLists.txt. 

I have hit the issue that ``systemc`` is still not Conan 2.0 ready, will check it.

valgur:
@RubenRBS @uilianries This recipe requires fmt < 10, but this conflicts with spdlog. How should this be handled, if `force=True` cannot be used?

valgur:
For anyone wondering where all of the `self.requires()` suddenly came from: the CMakeLists.txt in the project previously installed these deps from a separate conanfile, which is now disabled.

jwillikers (approved):

recipes/scc/all/conanfile.py
```diff
@@ -1,89 +1,176 @@
...
+        deps.set_property("systemc", "cmake_file_name", "SystemC")
+        deps.set_property("systemc-cci", "cmake_target_name", "systemc-cci::systemc-cci")
+        deps.set_property("yaml-cpp", "cmake_target_name", "yaml-cpp::yaml-cpp")
+        deps.generate()
+
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+        replace_in_file(self, os.path.join(self.source_folder, "third_party", "axi_chi", "CMakeLists.txt"),
+                        " STATIC", "")
+        if is_msvc(self):
```
memsharded, line 115: This is explicitly forbidden in ``validate()`` isn't it? Maybe can be removed?

valgur: Removed them.

recipes/scc/all/conanfile.py
```diff
@@ -1,89 +1,166 @@
...
+            #     see declaration of 'tlm::scc::tlm_gp_mm_t<16,false>'
+            #     src\sysc\tlm\scc\tlm_mm.h(116,114):
+            #     due to following members:
+            #     src\sysc\tlm\scc\tlm_mm.h(116,114):
+            #     'void tlm::tlm_extension<tlm::scc::tlm_gp_mm>::copy_from(const tlm::tlm_extension_base &)': is abstract
+            #     systemc-2.3.4\p\include\tlm_core\tlm_2\tlm_generic_payload\tlm_gp.h(78,18):
+            #     see declaration of 'tlm::tlm_extension<tlm::scc::tlm_gp_mm>::copy_from'
+            # and
+            #   src\sysc\tlm\scc\tlm_mm.h(31,20): error C2061: syntax error: identifier '__attribute__'
+            raise ConanInvalidConfiguration(f"{self.ref} is not supported on MSVC.")
```
memsharded, line 79: Is this documented and explicit by the upstream library? The message should be different if the recipe does not support it yet, or it is a library documented limitation. If it is a recipe issue, then: "This recipe does not support MSVC yet, contributions to ConanCenter to add it are welcome"

valgur: It's documented in the recipe.

valgur: Other than that - your guess is as good as mine. I'll leave it up to the user to find the deeper reason for the failure.

memsharded: > It's documented in the recipe.

The problem is that end users of the package doesn't read the recipe, they just receive an exception saying "scc/version is not supported in MSVC", which could be understood as the upstream library does not support MSVC which doesn't seem true. Then, it is important for end recipe users (as opposed to recipe creators and contributors) to get the right message.

valgur: Since the project appears to intend to have MSVC support, but I did not find any more details about it on the repo, I included a "contributions are welcome" message in the error.

recipes/scc/all/patches/2023.06-add-fmt-10-support.patch
```diff
@@ -0,0 +1,11 @@
+--- src/sysc/scc/trace/vcd_trace.hh
++++ src/sysc/scc/trace/vcd_trace.hh
+@@ -53,7 +53,7 @@
+ 
+ template<typename T>
+ inline void vcdEmitValueChangeReal(FPTR os, std::string const& handle, unsigned bits, T val){
+-    auto buf = fmt::format("r{:.16g} {}\n", val, handle);
++    auto buf = fmt::format("r{:.16g} {}\n", static_cast<double>(val), handle);
```
memsharded, line 8: This has been merged upstream in https://github.com/Minres/SystemC-Components/pull/51 and released in https://github.com/Minres/SystemC-Components/releases/tag/2024.03

memsharded (approved):

AbrilRBS (approved):
Thanks! Feel free to add the new version in a follow-up PR :)

PR #18732, opened 2023-07-20
cgltf: migrate to Conan v2

valgur:


valgur:

RubenRBS:
dont break

recipes/cgltf/all/conanfile.py
```diff
@@ -1,89 +1,88 @@
...
         )
 
+    def package(self):
+        copy(self, "LICENSE", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
+        cmake = CMake(self)
+        cmake.install()
+        for header_file in ["cgltf.h", "cgltf_write.h"]:
+            header_fullpath = os.path.join(self.package_folder, "include", header_file)
+            self._remove_implementation(header_fullpath)
+        for dll in (self.package_path / "lib").glob("*.dll"):
```
RubenRBS, line 84: You probably want to change this as you did for the other PRs

RubenRBS (changes requested):
Thanks a lot! One request, the rest looks as good as this recipe will do with all the wonky things it does to the source

recipes/cgltf/all/conanfile.py
valgur: For this one it was actually needed. Here's a CI run without it:
https://github.com/conan-io/conan-center-index/pull/18732#issuecomment-1746553400

The CMakeLists.txt should be tweaked to correctly handle this during the build, but simply copying the DLLs is sufficient. It at least avoids the need for additional patches.

RubenRBS (approved):

recipes/cgltf/all/conanfile.py
RubenRBS: Thanks for the extra insight then! :)

jwillikers (approved):

PR #18738, opened 2023-07-20
tidy-html5: migrate to Conan v2

valgur:
TODO

- [x] v1 test_package fails to find DLL

### DIFF
--- a/recipes/tidy-html5/all/conanfile.py
+++ b/recipes/tidy-html5/all/conanfile.py
@@ -1,17 +1,22 @@
-from conans import ConanFile, CMake, tools
 import os
 
-required_conan_version = ">=1.33.0"
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rm, rmdir
+from conan.tools.microsoft import is_msvc
+
+required_conan_version = ">=1.53.0"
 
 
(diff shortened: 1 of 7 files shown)


valgur:

recipes/tidy-html5/all/conanfile.py
```diff
@@ -1,17 +1,22 @@
...
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rm, rmdir
+from conan.tools.microsoft import is_msvc
+
+required_conan_version = ">=1.53.0"
 
 
 class TidyHtml5Conan(ConanFile):
     name = "tidy-html5"
+    description = "The granddaddy of HTML tools, with support for modern standards"
     license = "W3C"
```
RubenRBS, line 14: Just checked, and the library does not seem to actually be licensed under these terms:

https://spdx.org/licenses/W3C.html differs substantially from https://github.com/htacg/tidy-html5/blob/next/README/LICENSE.md

RubenRBS (changes requested):

recipes/tidy-html5/all/conanfile.py
valgur: Indeed. Turns out their license actually exists as a separate SPDX liecense ID: [`HTMLTIDY`](https://spdx.org/licenses/HTMLTIDY.html)

RubenRBS (approved):
TIL about the license! Thanks a lot for the quick response :)

czoido (approved):

PR #18750, opened 2023-07-20
co: migrate to Conan v2

valgur:


valgur:
Re-opened as #19253 due to rebased history.

PR #18756, opened 2023-07-20
quickfix: migrate to Conan v2

valgur:


czoido (dismissed):

SSE4 (approved):

valgur:

czoido (approved):

SSE4 (approved):

recipes/quickfix/all/conanfile.py
```diff
@@ -1,92 +1,110 @@
...
+
+        if self.options.with_mysql == "libmysqlclient":
+            self.requires("libmysqlclient/8.0.31")
 
     def validate(self):
         if self.settings.os == "Windows" and self.options.shared:
             raise ConanInvalidConfiguration("QuickFIX cannot be built as shared lib on Windows")
-        if self.settings.os == "Macos" and self.settings.arch == "armv8":
-            raise ConanInvalidConfiguration("QuickFIX doesn't support ARM compilation")  # See issue: https://github.com/quickfix/quickfix/issues/206
+        if is_apple_os(self) and self.settings.arch == "armv8":
```
SpaceIm, line 65: Fundamentally, it doesn't really make sense to replace Macos by is_apple_os in this specific case. It was just a way to pass CI when macOS M1 builds have beenn introduced. Actually it's broken for arm in general.

recipes/quickfix/all/patches/0005-inc.-required-CMake-version-add-conan-basic-setup-an.patch
```diff
@@ -25,8 +25,8 @@ index 48908ead..8789432c 100644
  endif()
```
SpaceIm, line 25: This patch still contains reference to `CONAN_LIBS`, so it has good chance to have broken logic when dependencies are enabled.

SpaceIm: Moreover patch name should be changed, because `add-conan-basic-setup` is a conan v1 thing.

PR #18779, opened 2023-07-21
tinyxml: enable with_stl by default

valgur:
Needed for #17347.

Including STL support in a C++ library by default should not be an issue, I assume.

### DIFF
--- a/recipes/tinyxml/all/conanfile.py
+++ b/recipes/tinyxml/all/conanfile.py
@@ -25,7 +25,7 @@ class TinyXmlConan(ConanFile):
     default_options = {
         "shared": False,
         "fPIC": True,
-        "with_stl": False,
+        "with_stl": True,
     }
 
     exports_sources = "CMakeLists.txt"


ericLemanissier (approved):

uilianries (changes requested):
As this PR changes the default behavior expected by TinyXML project, it needs more discussion first.

RubenRBS (approved):

uilianries (approved):
After taking a look into others package managers we found that STL option is enabled by default. So based on that, we are good to go.

Arch Linux: https://gitlab.archlinux.org/archlinux/packaging/packages/tinyxml/-/blob/main/PKGBUILD#L33 (BTW I use Arch)

Homebrew: https://github.com/Homebrew/homebrew-core/blob/055ae78415b61ecf1fa3de32b76b8a149855f903/Formula/t/tinyxml.rb#L25

Debian: https://git.launchpad.net/ubuntu/+source/tinyxml/tree/debian/rules#n8

valgur:
Oh, wow. Thank you so much for looking into this so thoroughly.

PR #18790, opened 2023-07-22
ozz-animation: migrate to Conan v2

valgur:


AlexRamallo:
Hey @valgur, sorry I didn't notice this PR was open when I started working on mine (#19191/#19126) last week.

Since it seems you didn't get a chance to finish working on this, I think it'd be best if you closed this to avoid duplicates, and review my PR since it seems like you might actually be depending on this package. I introduced new options to the existing recipe, so it might cause issues to existing users.

PR #18644, opened 2023-07-19
subunit: migrate to Conan v2

valgur:
TODO

- [ ] Remove `-Werror`.
- [ ] `ld: symbol(s) not found for architecture x86_64`

PR #18793, opened 2023-07-22
libcheck: migrate to Conan v2

valgur:
TODO
- Requires #18644 to be merged first.


valgur:

valgur:
The subunit dependency was temporarily disabled with this commit: https://github.com/conan-io/conan-center-index/pull/18793/commits/30cf12758ecab3372318c89bd556ab72d1b797ce
Should revert it once the that PR gets merged.

RubenRBS:
Oh, great to know, will focus on that one first then, thanks for the info!

recipes/libcheck/all/conanfile.py
```diff
@@ -24,96 +29,79 @@ class LibCheckConan(ConanFile):
...
             self.cpp_info.components["liblibcheck"].requires.append("subunit::libsubunit")
         if not self.options.shared:
             if self.settings.os in ["Linux", "FreeBSD"]:
                 self.cpp_info.components["liblibcheck"].system_libs = ["m", "pthread", "rt"]
 
+        # TODO: to remove in conan v2
         bin_path = os.path.join(self.package_folder, "bin")
-        self.output.info("Appending PATH environment variable: {}".format(bin_path))
+        self.output.info(f"Appending PATH environment variable: {bin_path}")
         self.env_info.PATH.append(bin_path)
```
jwillikers, line 97: You can drop this print statement and I believe you can even drop the following line as it should be added to `PATH` automatically.

```suggestion

```

jwillikers (changes requested):

recipes/libcheck/all/conanfile.py
valgur: Oh, really? Does that work with Conan v1 as well? I could have deleted so much dead code. :grin: 
Do you have an example of a recipe where `test_v1_package` works without one, maybe?

jwillikers (approved):

uilianries (approved):

PR #18794, opened 2023-07-22
qt-advanced-docking-system: migrate to Conan v2

valgur:


valgur:

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +35,108 @@ class QtADS(ConanFile):
...
-    _cmake = None
-    _qt_version = "5.15.6"
+    @property
+    def _qt_major(self):
+        return Version(self.dependencies["qt"].ref.version).major
 
     @property
-    def _source_subfolder(self):
-        return "source_subfolder"
+    def _min_cppstd(self):
```
fnadeau, line 44: Is there a reason to force cpp standard here? When building ads dynamically (most likely due to lgpl), it doesn't matter. If build statically, requirement will be forced by Qt anyway.

fnadeau:
Any progress on this? ADS is now at version 4.2.1. Is help needed for this?

valgur:
No progress and not actively working on this. You can look into the Windows build issue, if you feel like helping this PR along.

valgur:
Don't close.

valgur:
.

valgur:
.

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
-                  destination=self._source_subfolder)
+    def validate(self):
+        check_min_cppstd(self, self._min_cppstd)
 
-    def _configure_cmake(self):
-        if self._cmake:
-            return self._cmake
+    def build_requirements(self):
+        if not can_run(self):
+            self.tool_requires("qt/<host_version>", options={"gui": False, "widgets": False})
```
uilianries, line 61:
```suggestion
    def build_requirements(self):
        if not can_run(self):
            self.tool_requires("qt/<host_version>", options={"gui": False, "widgets": False})
```

Why does Qt is needed as build requirement as well? It's true the project supports qmake as well, but I could not find any Qt macro in src/CMakeLists.txt. What am I missing?

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
 
     def _patch_sources(self):
-        apply_conandata_patches(self)
-
-        replace_in_file(self,
-            f"{self.source_folder}/{self._source_subfolder}/src/ads_globals.cpp",
-            "#include <qpa/qplatformnativeinterface.h>",
-            f"#include <{self._qt_version}/QtGui/qpa/qplatformnativeinterface.h>"
+        qt_version = self.dependencies["qt"].ref.version
+        replace_in_file(self, os.path.join(self.source_folder, "src", "ads_globals.cpp"),
```
uilianries, line 77:
```suggestion
        replace_in_file(self, os.path.join(self.source_folder, "src", "ads_globals.cpp"),
```
Please, add a comment between lines 76 and 77 explaining why it is needed. My guess is adapting the include path to find from Conan package.

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
+        cmake = CMake(self)
+        cmake.configure()
         cmake.build()
 
     def package(self):
-        cmake = self._configure_cmake()
+        cmake = CMake(self)
         cmake.install()
-        self.copy("LICENSE", dst="licenses", src=self._source_subfolder)
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
```
uilianries, line 91:
```suggestion
        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
        copy(self, "gnu-lgpl-v2.1.md", self.source_folder, os.path.join(self.package_folder, "licenses"))
```

The upstream copies gnu-lgpl-2.1.md file as well when running install command. I would suggest copying it to the package too, just in case.

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
         rmdir(self, os.path.join(self.package_folder, "license"))
         rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
 
     def package_info(self):
+        if Version(self.version) >= 4:
```
uilianries, line 96: Please, update the expected cmake target as well:

The upstream generates the cmake target file lib/cmake/qt6advanceddocking/adsTargets.cmake, and it contains:

```
# Create imported target ads::qt6advanceddocking
add_library(ads::qt6advanceddocking STATIC IMPORTED)
```

Reference: https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/blob/4.2.0/src/CMakeLists.txt#L142

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
-            self.cpp_info.libs = ["qtadvanceddocking"]
+            self.cpp_info.libs = [lib_name]
         else:
             self.cpp_info.defines.append("ADS_STATIC")
-            self.cpp_info.libs = ["qtadvanceddocking_static"]
+            self.cpp_info.libs = [f"{lib_name}_static"]
+
+        if is_msvc(self) and self._qt_major >= 6:
+            # Qt 6 requires C++17 and a valid __cplusplus value
+            self.cpp_info.cxxflags.append("/Zc:__cplusplus")
```
uilianries, line 111:
```suggestion
        if is_msvc(self) and self._qt_major >= 6:
            # Qt 6 requires C++17 and a valid __cplusplus value
            self.cpp_info.cxxflags.append("/Zc:__cplusplus")
```

Not sure about this, it ensures the __cplusplus macro reflects the correct C++ standard version being used during compilation. But then you are exporting to the downstream. Is it the case that some MSVC versions do not fit a C++17 feature and are breaking this project?

recipes/qt-advanced-docking-system/all/conandata.yml
```diff
@@ -1,11 +1,7 @@
...
-    url: "https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/archive/refs/tags/3.8.3.tar.gz"
-    sha256: "bd5a9469b755bedf33baefd0b3dda6d167b7917a2888e2794eed5abee7d78f74"
-  "3.8.2":
-    url: "https://github.com/githubuser0xFFFF/Qt-Advanced-Docking-System/archive/refs/tags/3.8.2.tar.gz"
-    sha256: "e56811228fb4d5f5703c31cd83cb39ab2d5a849f581719d383db72f9322ec7f2"
-patches:
-  "3.8.2":
-    - patch_file: "patches/fix-cmake-license-install.patch"
-      base_path: "source_subfolder"
+  "4.2.0":
```
uilianries, line 2: The latest version available is 4.3.1. It may be worth updating it in case you are inclined and no one asked for 4.2.0 specifically.

uilianries (commented):
Thank you for your PR and pushing it toward Conan 2.x

Please, take a look at my review, some parts of these changes concern me about the recipe result. Best regards!

recipes/qt-advanced-docking-system/all/conanfile.py
valgur: I'll double check later, but in the process of adding cross-compilation support to Qt I learned that it always looks for native Qt executables in its .cmake modules and fails if they are not found, even if they are not eventually used anywhere.
Currently, in recipes consuming Qt, the executables are found simply based on their location relative to the .cmake modules, even if they have not been added to PATH. That is why building against shared Qt fails in some recipes that are using the executables - the necessary shared libraries have not been added to env vars.
To fix that in the native build context, the Qt dep will need a `run=True` and/or `VirtualRunEnv(self).generate(scope="build")`. Always adding Qt as a tool_requires would also work, but I'm sure you can imagine how fun it is to build Qt twice.

A somewhat related question - is `self.requires(..., run=True)` property propagated in the graph as well or does it apply only to the local recipe. I think it might have been transitive. In that case generating VirtualBuildEnv for build would be more correct and safer.

recipes/qt-advanced-docking-system/all/conandata.yml
valgur: Can do.

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
 
-    def configure(self):
-        if self.options.shared:
-            del self.options.fPIC
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires(f"qt/{self._qt_version}")
+        self.requires("qt/[>=6.0 <7]", transitive_headers=True, transitive_libs=True, run=can_run(self))
```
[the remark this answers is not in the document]
valgur: Probably can be reduced to just transitive_headers. I did not check whether transitive_libs is necessary or not when migrating this recipe. See the tool_requires comment for the run=True.

recipes/qt-advanced-docking-system/all/conanfile.py
```diff
@@ -26,70 +33,79 @@ class QtADS(ConanFile):
...
+        self.requires("libpng/[>=1.6 <2]")
```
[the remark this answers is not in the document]
valgur: I don't even remember why a saw a need for it anymore. I'll drop it.

recipes/qt-advanced-docking-system/all/conanfile.py
valgur: I think there was a check against the `__cplusplus` define value in the code somewhere, but I don't see one in the code anymore. I'll simply drop it for now.

recipes/qt-advanced-docking-system/all/conanfile.py
valgur: I'm puzzled by this change as well - it has been 1.5 years. I'll look into it.

valgur:
Thank you for the review, @uilianries!

recipes/qt-advanced-docking-system/all/conanfile.py
uilianries: @valgur Thank you for your explanation. For this specific recipe there is no register about problems related to cross-building, so, could you please remove it?

Plus, Qt6 does not support cross-building for now. Neither libmysqlclient/8.1.0 (another transitive dependency), so for this package is not worth adding extra complexity regarding cross-building.

valgur: Well... of course. Consider this a bug report for the cross-compilation issues. I can share a successful build log as well.

uilianries: @valgur That's not needed. Let's do that with the minimal PR that does that, that way we can move forward.  We can look at the cross-building situation in a follow-up PR, in particular, because if not can_run(self) is an unusual pattern to guard against for a tool requires.

valgur: The `not can_run(self)` will effectively never run until Qt is cross-compilable, so there's no harm in it. Can you not trust me on this? I've been through the guts of the Qt recipe and its .cmake modules. The only alternatives are to heavily patch the modules to not check for the executables every time or to unconditionally always add Qt as tool_requires.

I understand that you won't like me wasting your time on this, but the feeling is mutual.

valgur: fine

jcar87:
Merged (with preserved authorship) via https://github.com/conan-io/conan-center-index/pull/26496

PR #18795, opened 2023-07-22
srt: migrate to Conan v2

valgur:


valgur:

recipes/srt/all/conanfile.py
```diff
@@ -37,13 +37,7 @@ def _has_stdcxx_sync(self):
...
     @property
     def _has_posix_threads(self):
-        return not (
-            self.settings.os == "Windows"
-            and (
-                is_msvc(self)
-                or (self.settings.compiler == "gcc" and self.settings.compiler.get_safe("threads") == "win32")
-            )
-        )
+        return not (is_msvc(self) or self.settings.compiler.get_safe("threads") == "win32")
```
SpaceIm, line 40: No it's broken, keep previous condition please.

recipes/srt/all/conanfile.py
```diff
@@ -1,103 +1,118 @@
...
+            self,
+            os.path.join(self.source_folder, "CMakeLists.txt"),
+            'set (CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/scripts")',
+            'list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/scripts")',
+        )
+        if not self._has_posix_threads and not self._has_stdcxx_sync:
+            save(self, os.path.join(self.source_folder, "CMakeLists.txt"),
+                "find_package(pthreads4w REQUIRED)\n"
+                "link_libraried(pthreads4w::pthreads4w)\n",
+                append=True)
```
SpaceIm, line 95: there is a typo: `link_libraried`

andI don't think it can work anyway, because link_libraries() only affect targets created after, but you append to top CMakeLists, so it will be no-op.

recipes/srt/all/conanfile.py
valgur: Are you sure. Here's the reasoning for the simplifications I made:
1. The original expression with `is_msvc(self)` substituted:
  ```python
not (self.settings.os == "Windows" and (is_msvc(self) or (self.settings.compiler == "gcc" and self.settings.compiler.get_safe("threads") == "win32")))
```
2. The GCC check can be dropped since we are clearly not using Posix threads if `compiler.threads == "win32"`.
  ```python
not (self.settings.os == "Windows" and (is_msvc(self) or self.settings.compiler.get_safe("threads") == "win32"))
```
3. The `os == "Windows"` check is redundant, since both `is_msvc(self)` and `compiler.threads == "win32"` imply it.
  ```python
not (is_msvc(self) or self.settings.compiler.get_safe("threads") == "win32")
```

recipes/srt/all/conanfile.py
valgur: My bad. Thanks!

recipes/srt/all/conanfile.py
SpaceIm: Indeed, my bad 😃

uilianries (approved):
LGTM

PR #18812, opened 2023-07-22
libjxl: migrate to Conan v2, add v0.10.2

valgur:
.


valgur:

HappySeaFox:
Hi, any progress?

recipes/libjxl/all/conanfile.py
```diff
@@ -14,83 +17,114 @@ class LibjxlConan(ConanFile):
...
-        self.requires("lcms/2.11")
+        self.requires("brotli/1.1.0")
+        self.requires("highway/1.0.7")
+        self.requires("lcms/2.14")
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, 11)
+        if conan_version.major == 1 and self.settings.get_safe("compiler.libcxx") == "libc++":
+            raise ConanInvalidConfiguration("libc++ is not compatible with by libjxl due to limited std::atomic support")
```
uilianries, line 51:
```suggestion
```

It actually works with libc++, I'm to build locally. The problem is the `FindAtomic.cmake` in the project that is not able to check the library correctly. Unfortunately, CMake still does not provide a native way to check it, but llvm provides a robust macro: https://github.com/llvm-mirror/llvm/blob/master/cmake/modules/CheckAtomic.cmake

However, only enforcing atomic usage directly should be enough, because in case something is missing, the compilation would fail anyway.

So, you could add the follow patch:

```
diff --git a/recipes/libjxl/all/conanfile.py b/recipes/libjxl/all/conanfile.py
index 33b054036..e17db26f6 100644
--- a/recipes/libjxl/all/conanfile.py
+++ b/recipes/libjxl/all/conanfile.py
@@ -47,8 +47,6 @@ class LibjxlConan(ConanFile):
     def validate(self):
             check_min_cppstd(self, 11)
-        if conan_version.major == 1 and self.settings.get_safe("compiler.libcxx") == "libc++":
-            raise ConanInvalidConfiguration("libc++ is not compatible with by libjxl due to limited std::atomic support")
 
     def source(self):
@@ -110,6 +108,8 @@ class LibjxlConan(ConanFile):
         jxl_cmake = os.path.join(self.source_folder, "lib", "jxl.cmake")
         replace_in_file(self, jxl_cmake, '"$<BUILD_INTERFACE:$<TARGET_PROPERTY', "# ", strict=False)
         replace_in_file(self, jxl_cmake, "$<TARGET_PROPERTY", "# ", strict=False)
+        # Skip custom FindAtomic and force to use atomic library directly
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "find_package(Atomics REQUIRED)", "set(ATOMICS_FOUND TRUE)\nset(ATOMICS_LIBRARIES atomic)")
 
     def build(self):
         self._patch_sources()
```

Plus, need to update cpp_info to use atomic:

```
# In the docker container conanio/clang12-ubuntu16.04:1.62.0

$ ldd /home/conan/.conan/data/libjxl/0.8.2/_/_/package/340adc293c97c562de04770e7eb2b5a361e8ae6e/lib/libjxl.so
	linux-vdso.so.1 =>  (0x00007ffc946e6000)
	libatomic.so.1 => /usr/local/lib64/libatomic.so.1 (0x00007f6738963000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f673865a000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f673843d000)
	libc++.so.1 => /usr/local/lib/libc++.so.1 (0x00007f6739290000)
	libgcc_s.so.1 => /usr/local/lib64/libgcc_s.so.1 (0x00007f6738226000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f6737e5c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f673917c000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f6737c54000)
```

valgur: Wow, thank you so much for looking into this. Based on some googling, it looks like `-latomic` is only needed for `libstc++` and `libstdc++11`. I updated the recipe accordingly.

jwillikers (approved):

HappySeaFox:
Bump! Is it ready for merging?

valgur:
@uilianries @RubenRBS @franramirez688 pinging for a review, in case you have missed this one

jwillikers (approved):

valgur:
#24165

Julianiolo:
@valgur What does this mean? Is this closed because you are also quitting cci? Forgive me as a I am somewhat confused :)

uilianries (approved):
Built locally, looks okay.

AbrilRBS (approved):

recipes/libjxl/all/conanfile.py
```diff
@@ -14,114 +21,215 @@ class LibjxlConan(ConanFile):
...
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+        if self.settings.arch not in ["x86", "x86_64"] or Version(self.version) < "0.9":
+            del self.options.avx512
+            del self.options.avx512_spr
+            del self.options.avx512_zen4
+        # https://github.com/libjxl/libjxl/blob/v0.9.1/CMakeLists.txt#L52-L59
+        if self.settings.os in ["Linux", "FreeBSD"] and self.settings.arch == "x86_64":
+            self.options.with_tcmalloc = True
```
Cyriuz, line 55: Sorry for reviving this but I just got a surprise when updating this package with allocator mismatches due to this change in my static build and I just wanted to ask; Should we really do this here? From the linked cmake I gather that libjxl tries to use tcmalloc by default if it exists, but if we don't add it as a dependency, it shouldn't exists unless it is available on the system? Which I assume is a common problem in general with many libs. Could a better solution be to be able to disable it completely upstream instead either via a patch or a PR to them?

valgur: Thanks for reporting this. Please open it as a separate issue for more visibility.

PR #18819, opened 2023-07-22
tar: migrate to Conan v2

valgur:


valgur:

recipes/tar/all/conanfile.py
```diff
@@ -1,98 +1,109 @@
...
-            raise ConanInvalidConfiguration("bzip2:build_executable must be enabled")
-
-    def package_id(self):
-        del self.info.settings.compiler
+        if conan_version.major >= 2:
+            # Check does not work with Conan v2
+            if not self.dependencies.build["bzip2"].options.build_executable:
+                raise ConanInvalidConfiguration("bzip2:build_executable must be enabled")
+
+    def build_requirements(self):
```
uilianries, line 47: I don't see it as build requirement, but regular requires instead. In case you build with `-o */*:shared=True`, it will fail in runtime, because there is no other libraries available.

recipes/tar/all/conanfile.py
```diff
@@ -1,98 +1,112 @@
...
-        self.output.info("Appending PATH environment variable: {}".format(bin_path))
-        self.env_info.PATH.append(bin_path)
+        self.cpp_info.frameworkdirs = []
+        self.cpp_info.libdirs = []
+        self.cpp_info.resdirs = []
+        self.cpp_info.includedirs = []
 
         tar_bin = os.path.join(self.package_folder, "bin", "tar")
-        self.user_info.tar = tar_bin
+        self.conf_info.define("user.tar:tar", tar_bin)
```
uilianries, line 106:
```suggestion
        self.conf_info.define("user.tar:path", tar_bin)
```
A bit more specific about the which property of tar

uilianries (commented):
Looking good, only a small detail.

jwillikers (approved):

uilianries (approved):
LGTM

PR #18826, opened 2023-07-22
zstd/*: Added "build_programs" option

klausholstjacobsen:
Library: zstd/*

Added option ("build_programs") to let the consumer choose whether to build executables/programs for zstd along with the static library.
Also added tests to verify the built executables!

---

### DIFF
--- a/recipes/zstd/all/conandata.yml
+++ b/recipes/zstd/all/conandata.yml
@@ -46,7 +46,7 @@ patches:
     - patch_file: "patches/1.5.5-qnx_support.patch"
       patch_description: "Add qnx to platform"
       patch_type: "portability"
-      patch_souce: "https://github.com/facebook/zstd/pull/3745"
+      patch_source: "https://github.com/facebook/zstd/pull/3745"
   "1.5.4":
     - patch_file: "patches/1.5.2-cmake-remove-asm-except-x86_64.patch"
       patch_description: "use assembler codes only on x86_64"
--- a/recipes/zstd/all/conandata.yml
+++ b/recipes/zstd/all/conandata.yml
@@ -43,6 +43,9 @@ patches:
     - patch_file: "patches/1.5.2-cmake-remove-asm-except-x86_64.patch"
       patch_description: "use assembler codes only on x86_64"
       patch_type: "portability"
+    - patch_file: "patches/1.5.5-qnx_support.patch"
+      patch_description: "Add qnx to platform"
+      patch_type: "portability"
   "1.5.4":
     - patch_file: "patches/1.5.2-cmake-remove-asm-except-x86_64.patch"
       patch_description: "use assembler codes only on x86_64"
--- a/recipes/zstd/all/conandata.yml
+++ b/recipes/zstd/all/conandata.yml
@@ -46,6 +46,7 @@ patches:
     - patch_file: "patches/1.5.5-qnx_support.patch"
       patch_description: "Add qnx to platform"
       patch_type: "portability"
+      patch_souce: "https://github.com/facebook/zstd/pull/3745"
   "1.5.4":
     - patch_file: "patches/1.5.2-cmake-remove-asm-except-x86_64.patch"
       patch_description: "use assembler codes only on x86_64"
--- a/recipes/zstd/all/conanfile.py
+++ b/recipes/zstd/all/conanfile.py
@@ -1,11 +1,11 @@
 from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
-from conan.tools.files import apply_conandata_patches, collect_libs, copy, export_conandata_patches, get, replace_in_file, rmdir
+from conan.tools.files import apply_conandata_patches, collect_libs, copy, export_conandata_patches, get, replace_in_file, rmdir, rm
 from conan.tools.scm import Version
+from conan.tools.microsoft import is_msvc
 import os
 
 required_conan_version = ">=1.53.0"
-
 
 class ZstdConan(ConanFile):
     name = "zstd"
@@ -21,11 +21,13 @@ class ZstdConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "threading": [True, False],
+        "build_programs": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "threading": True,
+        "build_programs": True,
     }
 
     def export_sources(self):
@@ -49,10 +51,14 @@ class ZstdConan(ConanFile):
 
     def generate(self):
         tc = CMakeToolchain(self)
-        tc.variables["ZSTD_BUILD_PROGRAMS"] = False
-        tc.variables["ZSTD_BUILD_STATIC"] = not self.options.shared
+        tc.variables["ZSTD_BUILD_PROGRAMS"] = self.options.build_programs
+        tc.variables["ZSTD_BUILD_STATIC"] = not self.options.shared or self.options.build_programs
         tc.variables["ZSTD_BUILD_SHARED"] = self.options.shared
         tc.variables["ZSTD_MULTITHREAD_SUPPORT"] = self.options.threading
+
+        if not is_msvc(self):
+            tc.variables["CMAKE_C_FLAGS"] = "-Wno-maybe-uninitialized"
+
         if Version(self.version) < "1.4.3":
             # Generate a relocatable shared lib on Macos
             tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
@@ -77,6 +83,12 @@ class ZstdConan(ConanFile):
         cmake.install()
         rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+
+        if self.options.shared and self.options.build_programs:
+            #If we built programs we always build static libs,
+            #but if we only want shared libs in the package then remove the static libs
+            rm(self, "*.a", os.path.join(self.package_folder, "lib"))
 
     def package_info(self):
         zstd_cmake = "libzstd_shared" if self.options.shared else "libzstd_static"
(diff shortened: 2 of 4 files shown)


recipes/zstd/all/test_package/conanfile.py
```diff
@@ -25,3 +28,16 @@ def test(self):
...
             self.run(f"{bin_path} {png_file}", env="conanrun")
+
+            zstd_dep = self.dependencies["zstd"]
+            if(zstd_dep.options.build_programs):
+                zstd = os.path.join(zstd_dep.package_folder, zstd_dep.cpp_info.bindirs[0], "zstd")
+                zstd_version = zstd_dep.ref.version
+
+                output = StringIO()
+                self.run(f"{zstd} --version", output, env="conanrun")
+                s = re.search(f"{zstd_version}", output.getvalue())
```
valgur, line 39:
```suggestion
                zstd_version = zstd_dep.ref.version

                output = StringIO()
                self.run("zstd --version", output)
                s = re.search(zstd_version, output.getvalue())
```
The `zstd` executable should already be available on `PATH` thanks to the `VirtualRunEnv` generator.

klausholstjacobsen: That was my own initial thought. It will however not work if you build the package on a system that has zstd installed, because then that one will be executed.

`self.run(f"which zstd")` produces:

```
zstd/1.5.5 (test package): RUN: which zstd
/usr/bin/zstd
```

with or without `env="conanrun"`

klausholstjacobsen:
I see that the test recipe will not build with conan 1.x because self.dependencies is not available in `def test(self)`
Any suggestions to how to access information about the whereabouts of the package we're testing?

valgur:
> I see that the test recipe will not build with conan 1.x because self.dependencies is not available in `def test(self)`
> Any suggestions to how to access information about the whereabouts of the package we're testing?

You can take a look at how `ua-nodeset` handles this as a working example.

recipes/zstd/all/test_package/conanfile.py
valgur: Oh, yes, that's a very valid point.

valgur: Oh, yes, that's a very valid point.

klausholstjacobsen:
Pipeline says "system is under maintenance". System seem to be operational again. How can I make the pipelines rerun?

recipes/zstd/all/conanfile.py
```diff
@@ -21,13 +22,19 @@ class ZstdConan(ConanFile):
...
         "shared": False,
         "fPIC": True,
         "threading": True,
+        "build_programs": False,
     }
 
+    def validate(self):
+        if is_msvc(self):
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built with shared on Visual Studio and msvc, use mingw.")
+
```
uilianries, line 37:
```suggestion
    def validate(self):
        if is_msvc(self):
            raise ConanInvalidConfiguration(f"{self.ref} can not be built with shared on Visual Studio and msvc, use mingw.")

```

It's not clear about this new restriction. What's the error, why only fail now?

klausholstjacobsen: You are correct. Removed this check to allow msvc build!

recipes/zstd/all/test_package/conanfile.py
```diff
@@ -15,13 +19,28 @@ def layout(self):
...
+        save(self, "zstd_version", f"{zstd_version}")
+
+
     def build(self):
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
 
     def test(self):
-        if can_run(self):
```
uilianries, line 24: The original test must be preserved. It validates the packaged library. Please, revert it.

uilianries (dismissed):
I updated the test package to keep library validation too, so we can check both executable and libraries.

jwillikers (approved):

recipes/zstd/all/conanfile.py
```diff
@@ -90,3 +102,8 @@ def package_info(self):
         self.cpp_info.components["zstdlib"].libs = collect_libs(self)
         if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.components["zstdlib"].system_libs.append("pthread")
+
+        if self.options.build_programs:
+            # TODO: Remove after dropping Conan 1.x from ConanCenterIndex
+            bindir = os.path.join(self.package_folder, "bin")
+            self.runenv_info.prepend_path("PATH", bindir)
```
Hopobcn, line 109: Does conan detects that `self.package_folder/bin` exists, and add the path to self.runenv_info automatically? 

@uilianries is this actually necessary?

klausholstjacobsen: This is necessary for the test package to be able to run the produced executable. If We did not do this and the build system has zstd installed in the OS then the `self.run(..., env=conanrun)` in the test package will use OS installed one instead of the one produced here!

uilianries: @klausholstjacobsen It's not needed actually, the zstd can be added as VirtualBuildEnv, tool_requires, so it should be available. Real example: https://github.com/conan-io/conan-center-index/blob/230ab447b90a136dca3b2bf45875b39c62136103/recipes/yasm/all/test_package/conanfile.py#L6

klausholstjacobsen: Gotcha!

uilianries: @klausholstjacobsen I'll verify locally

uilianries, line 109:
```suggestion
            self.env_info.PATH.append(bindir)
```
@klausholstjacobsen the `runenv_info` is Conan 2.x feature: https://docs.conan.io/2/reference/conanfile/methods/package_info.html#buildenv-info-runenv-info

For Conan 1.x , you should use `self.env_info.PATH` instead. I tested locally and this change works.

klausholstjacobsen:
New platform.h patch added upstream: https://github.com/facebook/zstd/pull/3745

recipes/zstd/all/conanfile.py
klausholstjacobsen: I also have problem making VirtualBuildEnv and tool_requires work with conan 2.0.9:
It seem to calculate different required package id for the requires(...) and tool__requires(...)

See here:
```
zstd/1.5.5: Created package revision 1c38485aeb7744c655abe5f56c5f1c89
zstd/1.5.5: Package '43c10c705bb0f1d0030e597cb09a05ef5953fed9' created
zstd/1.5.5: Full package reference: zstd/1.5.5#4e5fbc9a14e898483676fb77fb4eef3b:43c10c705bb0f1d0030e597cb09a05ef5953fed9#1c38485aeb7744c655abe5f56c5f1c89
zstd/1.5.5: Package folder [hidden]/.conan2/p/b/zstd9d8fd2314131c/p
WARN: deprecated: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
WARN: deprecated:     'cpp_info.names' used in: zstd/1.5.5
WARN: deprecated:     'env_info' used in: zstd/1.5.5

======== Launching test_package ========

Graph root
    zstd/1.5.5 (test package): [hidden]/source/conan-center-index/recipes/zstd/all/test_package/conanfile.py
Requirements
    zstd/1.5.5#4e5fbc9a14e898483676fb77fb4eef3b - Cache
Build requirements
    zstd/1.5.5#4e5fbc9a14e898483676fb77fb4eef3b - Cache

Requirements
    zstd/1.5.5#4e5fbc9a14e898483676fb77fb4eef3b:43c10c705bb0f1d0030e597cb09a05ef5953fed9#1c38485aeb7744c655abe5f56c5f1c89 - Cache
Build requirements
    zstd/1.5.5#4e5fbc9a14e898483676fb77fb4eef3b:c1b14c3945d5adf1e5ae440c43c12f88117f2edc - Missing

ERROR: Missing binary: zstd/1.5.5:c1b14c3945d5adf1e5ae440c43c12f88117f2edc

zstd/1.5.5: WARN: Can't find a 'zstd/1.5.5' package binary 'c1b14c3945d5adf1e5ae440c43c12f88117f2edc' for the configuration:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.version=11
os=Linux
[options]
build_programs=True
fPIC=True
shared=False
threading=True

ERROR: Missing prebuilt package for 'zstd/1.5.5'
Check the available packages using 'conan list zstd/1.5.5:* -r=remote'
'conan test' tested packages must exist, and '--build' argument is used only for the 'test_package' dependencies, not for the tested dependencies

```

build command:
`conan create recipes/zstd/all/ -pr:h default -pr:b default --version 1.5.5 --build zstd/* --build missing -o:h *:shared=True`

As you can see the normal requires in the test packages resolves package id 43xxx and the tool_requires resolves c1xxx

klausholstjacobsen: Never got the above "multiple package id" issue solved.
But found a workaround from other inspirational packages.

In stead of doing this: 
```
    def requirements(self):

    def build_requirements(self):
```
I am doing this:
```
    def requirements(self):
        self.requires(self.tested_reference_str, run=True)
```
I guess this is due to the fact that `package_type = "library"` even though the package also produces binaries. But I cannot figure out how that translates to conan reolving mulitple dependent package versions of the package.

recipes/zstd/all/conandata.yml
```diff
@@ -43,6 +43,9 @@ patches:
     - patch_file: "patches/1.5.2-cmake-remove-asm-except-x86_64.patch"
       patch_description: "use assembler codes only on x86_64"
       patch_type: "portability"
+    - patch_file: "patches/1.5.5-qnx_support.patch"
+      patch_description: "Add qnx to platform"
+      patch_type: "portability"
```
uilianries, line 48:
```suggestion
      patch_type: "portability"
      patch_souce: "https://github.com/facebook/zstd/pull/3745"
```

klausholstjacobsen:
Can anybody help explain why c3i/conan-v2/pr-merge has been in "Pending — This commit is being built" for more than 24 hours now, while continuous-integration/jenkins/pr-merge says "The build of this commit was aborted"?

I see no pipeline log links in the comments as I normally do when a pipeline fails?

Help is appreciated!

Regards
Klaus

uilianries (approved):
LGTM

jwillikers (approved):

SpaceIm:
There is a regression in this PR. Now both shared and static are packaged on Windows when `shared=True`. The removal strategy is *nix centric but totally forgot Windows.

SpaceIm:
Another regression to fix: https://github.com/conan-io/conan-center-index/pull/20919

PR #18832, opened 2023-07-22
opentdf-client: migrate to Conan v2

valgur:
* Update recipe to support Conan 2.0

Other fixes:
* Add missing direct dependency on `ZLIB`
* Move patched logic to call `find_package` to a separate project include file
* Add missing component requirement in `package_info()` for `magic_enum`
* Support the `shared` option, and add `package_type` attribute
   - Add the `shared` option
   - Throw an exception in the `validate()` method when shared=True on Windows (not supported)
   - Patch the CMake scripts to _not_ install both static and shared variants into the package folder - only install one of them
   - Correctly set the library name in `package_info()` depending on the shared option value
* Other CMake fixes:
   - Comment out logic to build a "combined" static library with all static library dependencies in the same archive - this is not intended to be supported in Conan Center, and this file was not being used (as it was not advertised in `package_info()`


valgur:

valgur:
Thanks, @jcar87!

jcar87 (approved):

RubenRBS (approved):

PR #18847, opened 2023-07-23
gn: migrate to Conan v2

valgur:


valgur:

recipes/gn/all/conanfile.py
```diff
@@ -1,128 +1,165 @@
...
+            )
+
+            # Disable GenerateLastCommitPosition()
+            replace_in_file(self, os.path.join(self.source_folder, "build/gen.py"),
+                            "def GenerateLastCommitPosition(host, header):",
+                            "def GenerateLastCommitPosition(host, header):\n  return")
+
+            # Try sleeping to avoid time skew of the generated ninja.build file (and having to re-run build/gen.py)
+            err = None
+            for sleep in [0, 1, 2, 4, 8]:
```
RubenRBS, line 133: I need some context on this one. If the original sleep already raised some eyebrows, now with the loop it really needs some explanation 🤯

RubenRBS (commented):
Question about the sleep loop

maxnbk:
Does this PR need any assistance other than review? Hoping to pull in a conan-2-compatible `crashpad` which @valgur has also helpfully PRed and is blocked only on this.

IanBullard:
Please review this pull request.

recipes/gn/all/conanfile.py
valgur: @uilianries Unfortunately, it's not clear anymore what error to expect in this specific case, plus capturing the output without silencing it is kind of a pain, so I simply dropped this workaround for now.

jwillikers (approved):

IanBullard:
I'd appreciate it if we can get this merged.

recipes/gn/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,69 @@
+from conans import ConanFile, CMake, tools
```
danimtb, line 1: At this point, we can drop conan v1 test_packages

jwillikers (approved):

danimtb (approved):

PR #18849, opened 2023-07-23
libigl: migrate to Conan v2

### LINKED ISSUE #9694: [request] libigl/2.4.0
### Package Details
  * Package Name/Version: **libigl/2.4.0**
  * Website: **https://libigl.github.io/**
  * Source code: **https://github.com/libigl/libigl**

From their website:

> libigl is a simple C++ geometry processing library. We have a wide functionality including construction of sparse discrete differential geometry operators and finite-elements matrices such as the cotangent Laplacian and diagonalized mass matrix, simple facet and edge-based topology data structures, mesh-viewing utilities for OpenGL and GLSL, and many core functions for matrix manipulation which make [Eigen](http://eigen.tuxfamily.org/) feel a lot more like MATLAB.

valgur:
- Resolves #9469
- Resolves #9694


valgur:

RubenRBS:
Hi, thanks for this one. As I see, https://libigl.github.io/static-library/ comes in both header-only and static library, we'll need to model that. For an example on how to, fmt's recipe handles a similar case :)

RubenRBS:
I've pushed a change to manually override the package_type when header_only - The rest of the PR looks good to me :)

RubenRBS:
I lied, _now_ it looks good to me, feel free to take a look and let me know if you agree with my changes @valgur

AndreyMlashkin (approved):

jwillikers (approved):

recipes/libigl/all/conan_deps.cmake
```diff
@@ -0,0 +1 @@
+find_package(Eigen3 REQUIRED)
```
danimtb, line 1: Could you explain and document why this is needed? First time I see this approach in a recipe

valgur: It's pretty much a direct replacement for https://github.com/conan-io/conan-center-index/pull/18849/files#diff-81c0c5b750ed5544831cb54fb1e24013a3566ab140ffcbaa87c12e50f32e2931

It's not that relevant in this PR, but the `tc.variables["CMAKE_PROJECT_libigl_INCLUDE"] = "conan_deps.cmake"` pattern is something that I have used in several PRs to have tighter control over `find_package()` without needing to do a ton of patching in the CMake files. An example: https://github.com/conan-io/conan-center-index/pull/18900/files#diff-8175d97eafddd9bb81bd8992d385ce6790d86eef27d291becbbe687fd7bc660d

jwillikers (approved):

danimtb (approved):

PR #16558, opened 2023-03-14
Conan v2 support for LibTomMath

jowr:
Specify library name and version:  **libtommath/1.2.0-dev**

I regularly use this package and I would like to help porting to Conan v2.

I have not touched the old version 1.2.0, but it seems that it is outdated with respect to conan's Python tools. I am not sure what to do about it. Is it possible to completely remove the old release from the recipe?

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18852, opened 2023-07-23
libtommath: migrate to Conan v2

valgur:
Continues from #16558 with minimal changes (also resolves #16558).


valgur:

recipes/libtommath/cmake/conandata.yml
uilianries, line 1: both all and cmake folders are using CMake. Please, merge them if possible.

valgur: Good idea. Done.

jwillikers (approved):

recipes/libtommath/all/conanfile.py
```diff
@@ -21,119 +28,72 @@ class LibTomMathConan(ConanFile):
...
-    def build_requirements(self):
-        if self._settings_build.os == "Windows" and self.settings.compiler != "Visual Studio":
-            self.build_requires("make/4.3")
-        if self.settings.compiler != "Visual Studio" and self.settings.os != "Windows" and self.options.shared:
-            self.build_requires("libtool/2.4.6")
+        if is_msvc(self) and self.settings.build_type == "Debug":
+            # libtommath requires at least /O1 on MSVC for dead code elimination
+            # https://github.com/libtom/libtommath/blob/42b3fb07e7d504f61a04c7fca12e996d76a25251/s_mp_rand_platform.c#L120-L138
+            self.output.warning("Debug builds on MSVC are not supported, switching to RelWithDebInfo")
+            self.settings.build_type = "RelWithDebInfo"
```
RubenRBS, line 45: Hi! Thanks a lot for fixing the failing configuration. I'll talk with the team about your proposed solution, which while interesting, it's better if it has all our consensus on going ahead with it :)

recipes/libtommath/all/conanfile.py
```diff
@@ -21,119 +28,72 @@ class LibTomMathConan(ConanFile):
...
-                    ), run_environment=True)
+        if Version(self.version) >= "1.3":
+            get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        else:
+            get(self, **self.conan_data["sources"][self.version]["source"], strip_root=True)
+            download(self, **self.conan_data["sources"][self.version]["cmakelists"], filename="CMakeLists.txt")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
```
RubenRBS, line 59: As per the new policies we're trying to follow, if upstream does not support windows shared builds we should not rely on exporting all symbols by default

RubenRBS (commented):
Thanks! I have a few comments, will get back to you once I've talked with the team :)

recipes/libtommath/all/conanfile.py
RubenRBS, line 59: Accepting this as it's a C library, little risk for those

valgur: Thanks! Can we please extend that logic to https://github.com/conan-io/conan-center-index/pull/22197#discussion_r1579315523 :grin:

recipes/libtommath/all/conanfile.py
RubenRBS, line 45: Hi! We'll need to think about how to approach this in a better way, as per 
https://docs.conan.io/2/knowledge/guidelines.html#forbidden-practices recipes can NOT modify settings/confs, they are read-only.

As per @jcar87's suggestion, we can use `CMAKE_CXX_FLAGS_DEBUG_INIT` to set `/O1` if on debug, which will avoid the compilation errors:

Something like this in the `generate()` method might do the trick:

```py
if is_msvc(self) and self.settings.build_type == "Debug":
    tc.cache_variables["CMAKE_CXX_FLAGS_DEBUG_INIT"] = "/O1"
```

RubenRBS (changes requested):
Thanks! We'll need to change the approach to the msvc compilation errors, I've left a suggestion after talking with the team, let me know what you think :)

jwillikers (approved):

AbrilRBS (dismissed):
Thanks! :)

jwillikers (approved):

valgur:
.

valgur:
.

valgur:
.

uilianries (approved):
@valgur Thank you for continuing this PR.

I did a few changes on this PR, in order to simplify it:

- The version 1.3.0 includes CMakeLists.txt in its root folder. Supporting only this version simplifies the recipes a lot too.
- The upstream can not export symbols on Windows to generate a .dll. So I dropped shared library as well on Windows
- Simplified the test package to only validate the library.
- Dropped Windows debug support for now. First, CCI don't build package on Debug due the lower number the downloads and other topics, so we can not validate. Second, the upstream recommends inject /O1 instead of the default Cmake configuration /Og. In case an user really need Debug build type, then an issue can be opened to discuss this topic, but would need to pass CMAKE_C_FLAGS_DEBUG or similar.

PR #18861, opened 2023-07-24
pdcurses: migrate to Conan v2

valgur:


valgur:

recipes/pdcurses/all/conanfile.py
```diff
@@ -1,149 +1,197 @@
...
+        if self.options.get_safe("with_x11"):
             self.requires("xorg/system")
 
+    def validate(self):
+        if self.options.with_sdl and self.settings.os == "Windows":
+            raise ConanInvalidConfiguration("with_sdl option is not yet supported on Windows")
+        if self.settings.os != "Windows" and not self.options.get_safe("with_x11") and not self.options.with_sdl:
+            raise ConanInvalidConfiguration("At least one of with_x11 or with_sdl options must be enabled")
+        if cross_building(self):
+            raise ConanInvalidConfiguration("Cross-building is not supported")
```
uilianries, line 72: How cross_building is not working only now? Please, add more details.

valgur: I don't think the cross-building was ever tested. The most recent PR did not building for apple-clang at all: https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/5300/2/pdcurses/3.9//summary.json

It's a Makefile-only project and does not have the standard cross-compiling functionality of autotools, but I suppose it could be adapted for cross-compilation with some work. Better left for a separate PR, though.

uilianries (approved):

recipes/pdcurses/all/conanfile.py
```diff
@@ -1,149 +1,198 @@
...
+from conan.tools.microsoft import is_msvc, msvc_runtime_flag, unix_path, NMakeToolchain, NMakeDeps
+
+required_conan_version = ">=1.53.0"
+
 
 class PDCursesConan(ConanFile):
     name = "pdcurses"
     description = "PDCurses - a curses library for environments that don't fit the termcap/terminfo model"
-    topics = ("conan", "pdcurses", "curses", "ncurses")
+    license = "LicenseRef-LICENSE"
```
RubenRBS, line 20: It's my understanding (But I'm not a lawyer) that this should be something like 
```suggestion
    license = "DocumentRef-README.md:LicenseRef-pdcurses"
```

RubenRBS: (Because afaik you want to link the original document where the license text comes from, not where it's put into in the CCI package) - and as there's no "Public Domain" license (I really wish people would properly license it under CC0 or something!) then we're not sure that the best license id to put there is - that's why I use the package name itself

RubenRBS (commented):
Thanks! Some concerns about the license, otherwise seems good :)

recipes/pdcurses/all/conanfile.py
valgur: Going on a tangent: I understand the issue of "public domain" not being a valid and reliable license, but I'm not a huge fan of the SPDX word salad being used in its place either. I'm not against it either, but IMHO, a "Public Domain" would be a perfectly fine (non-)license identifier, which is about as useful as the DocumentRef/LicenseRef currently - it only provides a signal that you will need to look at the license in a case-by-case manner. Following the LicenseRef format is a bit cargo-culty and does not help much - it's not cross-referenced anywhere else and the specific license files can always found under the licenses folder. I would much rather the put effort into using the `x OR y` and `x AND y` notation to disambiguate the multiply-licensed projects, if I had to choose.

Anyway, sorry about the rant. What I really wanted to say is that the newer `DocumentRef-README.md:LicenseRef-pdcurses` should be included in the https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conanfile_attributes.md#license-attribute docs as well. :smile:

RubenRBS (approved):

PR #18862, opened 2023-07-24
godot-cpp: migrate to Conan v2

valgur:


valgur:

jwillikers (approved):

RubenRBS (approved):

RubenRBS (changes requested):
blocking, found something strange

recipes/godot-cpp/all/conanfile.py
```diff
@@ -12,20 +20,81 @@ class GodotCppConan(ConanFile):
...
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires(f"godot_headers/{self.version}", transitive_headers=True)
+
+    def package_id(self):
+        if self.info.settings.build_type != "Debug":
+            self.info.settings.build_type = "Release"
```
RubenRBS, line 45: Why is this needed? https://github.com/godotengine/godot-cpp/blob/master/CMakeLists.txt#L2 seems to imply this is an option. Does it actually always produce the same resulting binaries?

RubenRBS (commented):
Question about the package_id modification :)

recipes/godot-cpp/all/conanfile.py
valgur: `RelWithDebInfo` and other less common build types get built as `Release` by the project, so they are all equivalent. `Debug` is the only exception.

RubenRBS: Oh I missed that its looking for soemthing _other than_ debug  missed the negation. Thanks for clarifying

RubenRBS (approved):

Issue #5908, opened 2021-06-15
[request] liquid-dsp/latest (help needed)


ruilvo:
### Package Details
  * Package Name/Version: **liquid-dsp/latest**
  * Website: **https://github.com/jgaeddert/liquid-dsp**
  * Source code: **https://github.com/jgaeddert/liquid-dsp**

### Description Of The Library / Tool

So, this is an amazing DSP library, with a generous library.

I could do the recipe but I need some help: 
- It uses autotools;
- It uses C99 `<complex.h>` so it doesn't build with MSVC. 

However, since, this a C library, it can easily be linked with MSVC.  It's possible to build it with msys2/mingw-64 and then generate a "link library" for MSVC... 
Can this workflow be made with Conan? I mean, if the compiler is MSVC, have msys2/mingw-64 as a dependency and build it with it, and then generate a link library with MSVC, and etc?

uilianries:
@ruilvo I think you are looking for something like https://github.com/conan-io/conan-center-index/blob/c916710e54d587ba27f694a4678e7655a307ab76/recipes/re2c/all/conanfile.py#L43

PR #18866, opened 2023-07-24
liquid-dsp: migrate to Conan v2

valgur:


recipes/liquid-dsp/all/conanfile.py
```diff
@@ -49,138 +57,134 @@ def _target_name(self):
...
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def validate(self):
-        if hasattr(self, "settings_build") and tools.cross_building(self):
+        if hasattr(self, "settings_build") and cross_building(self):
             raise ConanInvalidConfiguration("Cross building is not yet supported. Contributions are welcome")
+        if is_msvc(self) and not self.options.shared:
+            # FIXME: export the correct libgcc library files for a static build
+            raise ConanInvalidConfiguration("Static builds are not supported on MSVC. Contributions are welcome")
```
SpaceIm, line 87: But it works in current recipe, it would be a regression.

valgur:
not stale

recipes/liquid-dsp/all/conanfile.py
```diff
@@ -49,138 +46,81 @@ def _target_name(self):
...
-            self.copy(patch["patch_file"])
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def validate(self):
-        if hasattr(self, "settings_build") and tools.cross_building(self):
+        if cross_building(self):
             raise ConanInvalidConfiguration("Cross building is not yet supported. Contributions are welcome")
+        if is_msvc(self):
+            raise ConanInvalidConfiguration("MSVC is not supported due to missing 'complex' support data type support")
```
perseoGI, line 70:
```suggestion
            raise ConanInvalidConfiguration("MSVC is not supported due to missing 'complex' data type support")
```

recipes/liquid-dsp/all/conanfile.py
```diff
@@ -49,138 +46,81 @@ def _target_name(self):
...
-        for patch in self.conan_data.get("patches", {}).get(self.version, []):
-            self.copy(patch["patch_file"])
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def validate(self):
-        if hasattr(self, "settings_build") and tools.cross_building(self):
+        if cross_building(self):
             raise ConanInvalidConfiguration("Cross building is not yet supported. Contributions are welcome")
+        if is_msvc(self):
```
perseoGI, line 69: In Conan 1 this recipe supports Windows, why are we removing support in this migration?

perseoGI (commented):
Looking good, but need answers for the decision of removing windows support

recipes/liquid-dsp/all/conanfile.py
valgur: It did so only through fetching MinGW via Conan and building with it instead of MSVC. I did try that approach and got it to work locally, but it kept failing on C3I due to a missing DLL despite at least 10 iterations of trying to fix this (and Windows is very unhelpful with its DLL error messages).
Relying on GCC from MinGW is a bit of a fragile edge-case for Conan, since it also becomes a build/runtime dependency unless you link GCC statically into a shared library. It's not well supported. See also https://github.com/conan-io/conan-center-index/pull/21193#issuecomment-2085602537

valgur:
don't close

recipes/liquid-dsp/all/conanfile.py
perseoGI: Okey, we will leave this for the future then. It is better to have this recipe working in `conan v2` for now. Even if it does not support MSVC

perseoGI (approved):
LGTM

uilianries (approved):
LGTM. Thank you for porting this recipe. 

- This library is not consumed by other recipes in CCI, so dropping oldest version is fine.
- Originally, it comes #5908, no MSVC supported requested so far, so dropping it is fine as well.
- An official CMake support will not happen, as there are 3 open PRs in the upstream and the author is not a fan of CMake. So we need to deal with Make and flags directly.

PR #18870, opened 2023-07-24
ada: add version 2.6.0

toge:
Specify library name and version:  **ada/***

---


danimtb:
Looks like there is a malformed package in artifactory. I have removed it and triggered the job again to regenerate it 😄

danimtb (approved):

recipes/ada/all/conanfile.py
```diff
@@ -69,15 +69,16 @@ def validate(self):
...
+        self.tool_requires("cmake/[>=3.16 <4]")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
         tc = CMakeToolchain(self)
         tc.variables["BUILD_TESTING"] = False
+        # Relocatable shared libs on macOS
+        tc.variables["CMAKE_MACOSX_RPATH"] = True
```
jcar87, line 81: Hi @toge  - I believe this is not necessary, as far as I'm aware the default behaviour of CMake >=3.0 is exactly this (for `CMAKE_MACOSX_RPATH` to be True by default). Given that this library (ada) has a minimum requirement of CMake 3.16, this would be redundant? Is there an error that we could reproduce that is fixed by applying this change?

jcar87: Hi @toge - I see, thanks!!
This is caused by this line here: https://github.com/ada-url/ada/blob/v2.6.0/cmake/ada-flags.cmake#L53

I'm a bit surprised if `tc.variables["CMAKE_MACOSX_RPATH"]` actually fixes the problem -  this defines the variable in the toolchain file, and thus, this is earlier than the call that overrides the value to `OFF` later on (`ada-flags.cmake` is processed _after_ the toolchain file) - according to the CMake documentation on evaluating variables, I interpret that the `OFF` value would prevail regardless.

I think most robust would be to actually patch `ada-flags.cmake` and remove that setting.

jwillikers (approved):

PR #18881, opened 2023-07-24
lua: specify `provides` attribute

uyha:

---


danimtb:
hi @uyha!

I have checked your PR and the `provides` attribute you are adding to this recipe is not needed. This recipe is called "lua" so it is already providing a package called that way. I think maybe it is not so clear in the docs, but you can check the "libjpeg" example to understand its usage https://docs.conan.io/2/reference/conanfile/attributes.html#provides

I have checked that you opened another PR related to this https://github.com/conan-io/conan-center-index/pull/18880. So in that case it should be enough with that one 😄 

Thank you!

PR #18883, opened 2023-07-24
cairo: add missing transitive_headers, add v1.17.8

valgur:


valgur:

RubenRBS (approved):

SSE4 (approved):

jwillikers (approved):

recipes/cairo/all/conanfile.py
```diff
@@ -84,20 +75,20 @@ def layout(self):
...
         if self.options.get_safe("with_glib", True):
-            self.requires("glib/2.76.3")
+            self.requires("glib/2.78.0", transitive_headers=True)
         self.requires("zlib/1.2.13")
         self.requires("pixman/0.40.0")
         self.requires("libpng/1.6.40")
 
     def package_id(self):
-        if self.options.get_safe("with_glib") and not self.dependencies["glib"].options.shared:
+        if self.info.options.get_safe("with_glib") and not self.dependencies["glib"].options.shared:
```
SpaceIm, line 91: it should be removed

recipes/cairo/meson/conanfile.py
```diff
@@ -75,12 +67,14 @@ def export_sources(self):
...
-        if self.settings.os != "Linux":
+        if self.settings.os not in ["Linux", "FreeBSD"]:
             del self.options.with_xlib
             del self.options.with_xlib_xrender
             del self.options.with_xcb
             del self.options.with_symbol_lookup
-        if self.settings.os in ["Macos", "Windows"]:
+        if self.settings.os == "Windows" or is_apple_os(self):
+            del self.options.with_opengl
+        if Version(self.version) >= "1.17.8":
```
SpaceIm, line 77: Here there might be a double delete of with_opengl option, this is the reason of v2 pipeline failure.

To replace with:

```python
if Version(self.version) >= "1.17.8" or self.settings.os == "Windows" or is_apple_os(self):
    del self.options.with_opengl
```

valgur:
@RubenRBS @jwillikers This PR should be ready to be merged and your review would be welcome. In particular, whether or not you agree with the dropping of the older versions using the autotools-based recipe. Squashing bugs in that recipe did not seem worth the effort as the versions are about three years old by now.

recipes/cairo/meson/conanfile.py
```diff
@@ -101,24 +93,24 @@ def requirements(self):
...
         if self.options.get_safe("with_opengl") == "desktop":
-            self.requires("opengl/system")
+            self.requires("opengl/system", transitive_headers=True)
             if self.settings.os == "Windows":
                 self.requires("glext/cci.20210420")
                 self.requires("wglext/cci.20200813")
                 self.requires("khrplatform/cci.20200529")
         if self.options.get_safe("with_opengl") and self.settings.os in ["Linux", "FreeBSD"]:
-            self.requires("egl/system")
+            self.requires("egl/system", transitive_headers=True)
```
jwillikers, line 113: Can we instead require the `egl` from `libglvnd` here, or does it require Mesa too?

jwillikers (approved):

recipes/cairo/meson/conanfile.py
valgur: I'm not that familiar with either of these libraries or Mesa. You will have to be more specific, sorry. :smile:

valgur: I can apply that change though, if it's necessary, sure. Is `transitive_libs=True` not required, though?

ericLemanissier (approved):

jwillikers (approved):

recipes/cairo/meson/conanfile.py
jwillikers: Looking at it again, it should probably stay as `egl/system`. That package just installs `libglvnd` and/or the Mesa EGL drivers, but I think it mostly depends on Mesa. I'm working on a Mesa package which will hopefully allow us to specify a dependency on a non-system Conan package. Hope I didn't confuse you too much.

jwillikers (approved):

SSE4 (approved):

uilianries (approved):
LGTM

PR #18884, opened 2023-07-24
cairomm: migrate to Conan v2

valgur:


valgur:

jwillikers (approved):

recipes/cairomm/all/conanfile.py
```diff
@@ -25,170 +35,118 @@ class CairommConan(ConanFile):
...
-        self.build_requires("pkgconf/1.7.4")
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("cairo/1.17.4")
-
-        if self._abi_version() == "1.16":
-            self.requires("libsigcpp/3.0.7")
+        self.requires("cairo/1.18.0", transitive_headers=True, transitive_libs=True)
```
uilianries, line 60: I see some headers indeed are consuming cairo headers, but is really needed all those transitive_headers and transitive_libs?

valgur: Yes. The `freetype` library, which is used by the project, is a transitive dependency from `cairo` and is not explicitly declared in `meson.build`, for example.

jwillikers (approved):

recipes/cairomm/all/conanfile.py
RubenRBS: This smells like upstream should then define that as a proper dependency no? Either way, seems like this is the best we can do as things stand, thanks! :)

recipes/cairomm/all/conanfile.py
```diff
@@ -25,170 +35,118 @@ class CairommConan(ConanFile):
...
-            raise ConanInvalidConfiguration(
-                "Linking against static cairo would cause shared cairomm to link "
-                "against static glib which can cause problems."
-            )
-
     @property
-    def _source_subfolder(self):
-        return "source_subfolder"
+    def _abi_version(self):
+        return "1.16" if Version(self.version) >= "1.16.0" else "1.0"
```
RubenRBS, line 40: This does not introduce too much variation complexity, so I'm ok with keeping the old version around for a bit longer, thanks for doing the extra work to keep its support when it would have been easier to just don't support 1.14 :)

RubenRBS (approved):
Looks good to me, thanks a lot! :)

Issue #13994, opened 2022-11-03
[service] Retroactive recipe changes make version numbering unreliable


patmantru:
**Real-life example:**
https://github.com/conan-io/conan-center-index/pull/13846#issuecomment-1297699165

opentdf-client requires libxml2 and boost.  A recent change to the libxml2 recipe broke builds that were consuming opentdf-client, while there were no changes at all to the opentdf-client recipe.  This recipe change retroactively altered the transitive version requirements of the package, thereby making it impossible to produce a repeatable build based on CCI.

```
ERROR: Conflict in libxml2/2.9.14:
    'libxml2/2.9.14' requires 'zlib/1.2.13' while 'boost/1.79.0' requires 'zlib/1.2.12'.
    To fix this conflict you need to override the package 'zlib' in your root package.
```

If the version number does not relate to a repeatable output, what's the point of having version numbers?  The root cause of all this seems to be steering recipe authors into using the same recipe for multiple versions of a package.  The recipe gets updated...and suddenly all the older versions of that same package are updated as well.  This is not a stable or reliable approach.  A backlevel version of a package may be fine one day and broken the next.  That's not a great foundation to build on.

My recourse appears to be:

**1) Add a requires statement in the opentdf-client recipe for zlib** 

Why should I have to add a requires statement for a package that is in fact not required by opentdf-client?  Why should I even have to know it exists?  I specified versions for boost and libxml2 that worked together a week ago, and now they don't.  

While I'd be very happy to use the option switches to disable unused functionality in boost and libxml2 in the opentdf-client recipe so this would not have been a problem in the first place, that sort of change is disallowed (as discussed here: https://github.com/conan-io/conan-center-index/pull/12697 ). I'm prevented from disabling a transitively-included package in my recipe, but somehow I'm being *required* to alter the version number of a transitively-included package in order to keep things building properly.  That also means anyone who was consuming client-opentdf and specifying boost+libxml options to disable zlib functionality will now get zlib anyway because of that requires statement.  To make matters worse, adding that zlib requirement statement would be a manual CCI PR review as it alters the recipe, and increases the maintenance footprint so it's more likely I'll have to have manual CCI PR reviews in the future.  

Am I supposed to have to version-pin the entire transitive tree of packages in my recipe "just in case they change"?  What's the purpose of having versions and recipes and packages in the first place?

**2) Use a private Conan repo and/or package cache**

Documented here as a CCI recommendation:
[...]

PR #18896, opened 2023-07-25
Adds opentdf-client versions 1.4.0 and 1.5.0

patmantru:
Specify library name and version:  **opentdf-client/1.5.0**

- Adds ICTDF support
- Upgrades to OpenSSL 3.1.1

---


recipes/opentdf-client/all/conanfile.py
```diff
@@ -66,7 +66,14 @@ def validate(self):
             raise ConanInvalidConfiguration(f'{self.name} can not be built with MT or MTd at this time')
 
     def requirements(self):
-        self.requires("openssl/1.1.1q")
+        # Uses openssl 3.x for 1.5.0 and newer
+        if Version(self.version) >= "1.5.0":
+            self.requires("openssl/3.1.1")
+        else:
+            self.requires("openssl/1.1.1q")
```
czoido, line 73: Hi @patmantru, thanks a lot for the contribution. One comment: we are now accepting version ranges for openssl to avoid overrides: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges
So maybe you can use one for both those requires.

czoido: Hi @patmantru, I may have not explained it properly, I mean that you can use separate version ranges for each `Version(self.version)` so that you avoid the need of doing overrides in other libraries that may be required in the same graph. Something similar to this:

```
        if Version(self.version) >= "1.5.0":
            self.requires("openssl/[>=3.1 <4]")
        else:
            self.requires("openssl/[>=1.1 <3]")
```

patmantru: OK, I understand now.  The issue I have (especially for the 1.1.x stuff) is that I don't want someone to accidentally pull in a random version that we don't know anything about.  Case in point, there are vulnerabilities in the 1.1.1[a-z] earlier versions that are fixed in later ones, so I'd rather go with a fixed version.  Unless there's a strict CCI requirement that I change this, I'd really prefer to leave it as-is.  Having a repeatable build (identical binaries every time you build it) is much more important to me for this package.

patmantru: @czoido any additional feedback?

czoido: Hi @patmantru,
There's no strict CCI requirement about that at the moment, It's just a recommendation that may avoid problems with versions conflicts with libraries extensively used such as openssl, but it there are legitimate reasons not to do that it's ok. Anyway, maybe you can depend on a fixed version for the case of openssl 1.X and use a version range for 3.X.

patmantru: I'm going to leave it as-is then, I'd rather have pinned versions.

patmantru:
@czoido any way to get some more reviews/approvals on this?  It's a very small change (adds 2 versions, adds 1 dependency, upgrades another dependency) so it should be a quick review.  Is there anything besides 2 more reviews that this needs to get merged?

Thanks...

patmantru:
Also, how come this isn't listed in 
https://github.com/prince-chrismc/conan-center-index-pending-review/issues/1

recipes/opentdf-client/all/conanfile.py
jcar87: Hi @partmantru. For the OpenSSL 3.x we will require the version range, unless it is known to not work in future versions (which should not be the case if OpenSSL follows their compatibility promises).

> I don't want someone to accidentally pull in a random version that we don't know anything about.
there is no randomness - `openssl/[>=3.1 <4]` will resolve to the most recent version that is visible by the users, _unless_ they themselves override it to a specific one.

> Case in point, there are vulnerabilities in the 1.1.1[a-z] earlier versions that are fixed in later ones, so I'd rather go with a fixed version.

because the version range will resolve to the most recent version, this most recent version is overwhelmingly more likely to include a critical security bugfix (usually those security bugfixes are what motivated a new release in the first place) - than keeping this anchored to an older version which may be vulnerable.

This also happens without having to make any future changes to the recipe, and the users have full control over which version is used if they should choose to.

patmantru: Is this a requirement or a recommendation?  Again...I'd prefer pinned versions so there are repeatable binaries.

jcar87: For OpenSSL 3 in Conan Center this is now the current convention, and not just a recommendation. Even if we were to accept this PR as is, we would happily accept a PR from a subsequent contributor wanting to relax the condition.

> I'd prefer pinned versions so there are repeatable binaries.

OpenSSL is a pure C library, and with Conan 2.0 in Conan center the generated binaries _do not_ and should not embed portions of OpenSSL - the generated code should always ensure the symbols are provided externally (from the OpenSSL package).

What this means is that in practice, the actual binary code of  packages of `opentdf-client` that depend on OpenSSL 3.x should be the same regardless of the version of OpenSSL used, as long as the signatures of the functions that `opentdf-client` is using remain the same - which they are guaranteed to be as per the [OpenSSL API/ABI guidelines](https://www.openssl.org/policies/releasestrat.html). 

That said, there will be different package IDs for `opentdt-client` for different versions of OpenSSL (up to the .minor version, Conan considers that changes in the patch version are compatible and should not need a new package ID). 

So in all cases, the binaries should be repeatable:
* because we are dealing with a pure C library with some of the strongest API/ABI guarantees - the generated code of `opentdf-client` should actually be the same w.r.t OpenSSL 3.x
* because Conan uses different package IDs - so each of them is reproducible

> Even allowing the idea that 'newer is always better', 

We are not saying that newer is always better. Although when it comes to OpenSSL, newer tends to be safest.

> and then not updated to any newer OpenSSL releases that came out afterwards.

We are planning to change this in the future, such that when new versions of OpenSSL are released (that is, new _minor_ releases), new package IDs are re-generated for all the libraries that depend on OpenSSL. We don't expect any issues here - given the strong compatibility promises of OpenSSL, everything should be source compatible.

> It becomes very difficult to determine exactly what version of OpenSSL you are running with in the case where conan is downloading the CCI binaries,

The `conan install` command will _always_ print exactly which version of OpenSSL it resolved to, exactly which version it is going to use, and it will make sure that the binaries were generated with the same `major.minor`, and if not, it will _require_ that new binaries are generated with the new version. 

Also please keep in mind that using overrides or lockfiles, users can control exactly what version of OpenSSL should be resolved, if they themselves have stronger requirements.

> Given that CCI disallows running unit test suites, it seems dangerous to allow a recipe that pulls in different subcomponents than I've run my full unit test suite against locally before we publish a new version.

For most libraries, Conan center will currently generate about 50 individual binary packages with different settings (os, compiler, compiler version) and options (shared True/False). I say it's highly unlikely that library authors run their test suites against all of these configurations. I would hope that both library authors and consumers of C and C++ libraries are aware that this is the prevalent case regardless of which package manager they decide to use. From what I can see, taking `curl` (or `libcurl`) as an example, [Homebrew](https://github.com/Homebrew/homebrew-core/blob/113f300e68390edbfd8e975034b70c28824658da/Formula/curl.rb#L43), [conda](https://github.com/conda-forge/curl-feedstock/blob/main/recipe/meta.yaml#L27),  and [vcpkg](https://github.com/microsoft/vcpkg/blob/master/ports/curl/vcpkg.json#L87) will all use the most recent version of OpenSSL found at the time the packages are built. Even [Ubuntu](https://packages.ubuntu.com/jammy/libcurl4) lists an open ended version range. Conan Center having strict version requirements (for some libraries that are well known to be backwards compatible, such as OpenSSL) has traditionally been an inconvenience to users - hence we are allowing and recommending version ranges for OpenSSL, which should enable the best of both worlds: get the latest if you don't care, get a specific one if you do.

patmantru: @jcar87 Thank you for taking the time for such a detailed reply, I appreciate that.

The core issue that I have is if a consumer contacts me and says "I used opentdf-client 1.5.0 and I had this problem", I have no way to know what version of OpenSSL was in use.  That makes the investigation that much more difficult.  Since not all consumers keep good notes on 'the output of the `conan install` command', that information is lost to me.  Pointing the finger at the consumer to figure out what version(s) of subordinate libraries were used at that point does nothing to cool their frustration.

When a vulnerability is announced, the first question we get asked is "Are we using that version of that package?"  Now I can't answer that, or at least I have to do my own record-keeping or come up with other ways to figure it out.  That makes things more difficult.

And to your point...

> because we are dealing with a pure C library with some of the strongest API/ABI guarantees - the generated code of opentdf-client should actually be the same w.r.t OpenSSL 3.x

...then why does it matter if the version number doesn't float?  You just argued that the binaries will be the same anyway, so...why does it matter if it's version-pinned?  Q: Why bother going back to re-compile packages that use OpenSSL if/when there is a new release if it's generating the same code anyway?  A: Because it might not.

> because Conan uses different package IDs - so each of them is reproducible

I disagree with this, I can't tell you the number of times I've had builds fail because a 3rd party CCI recipe changed underneath me with zero changes to my recipe, code or environment.  Yes, there is a way to pin the recipe versions of subordinate libraries to use, but that's (at best) cumbersome, and poorly supported on the CCI website.  And pinning recipe versions is essentially the same as pinning library versions because there can be no newer releases in a pinned recipe, so how is that different? 

I'm finding CCI difficult to use for production-quality code, where 100% repeatability is vitally important.  I also am finding it difficult to recommend to customers as a basis for producing production-quality applications.  For 'proof-of-concept' work, keep playing until it mostly works, it's great.  Given that CCI disallows recipe options...with a default of 'off'...for fear of generating different binaries, and disallowing the use of tags other than what's in `conandata.yml`, claiming it's OK to have a floating version number for 'some packages' is difficult to understand.  Making the build side of things less strict at the cost of making the support and problem-determination side very much more difficult is not a good tradeoff IMO.

patmantru:
@RubenRBS is this in limbo again?

RubenRBS:
No, but something weird is going on, let me investigate

patmantru:
@RubenRBS Thank you!

recipes/opentdf-client/all/conanfile.py
jcar87: > The core issue that I have is if a consumer contacts me and says "I used opentdf-client 1.5.0 and I had this problem", I have no way to know what version of OpenSSL was in use. That makes the investigation that much more difficult. Since not all consumers keep good notes on 'the output of the conan install command', that information is lost to me. Pointing the finger at the consumer to figure out what version(s) of subordinate libraries were used at that point does nothing to cool their frustration.

We are in a similar position as maintainers of Conan and Conan Center. When users make a report, we do expect them to give us complete information so that we can identify and troubleshoot their issue. We generally don't have problems identifying the exact versions from the logs users post, and users are generally happy to run conan install and paste the output for us. If that is not your experience and if there is anything that the `conan` tool could do to facilitate this gathering of information, we would be very happy to hear about actual experiences so that we can improve this (link to [new issue](https://github.com/conan-io/conan/issues/new/choose)).

> When a vulnerability is announced, the first question we get asked is "Are we using that version of that package?" 

If a vulnerability in OpenSSL is announced - I would argue that it is the responsibility of the user to identify that they are a) using OpenSSL, either directly or indirectly and b) take the necessary measures to update to a version that does not display the vulnerability. 

> Now I can't answer that, or at least I have to do my own record-keeping or come up with other ways to figure it out. That makes things more difficult.

Even if this recipe requests a _specific_ version of OpenSSL without a version range, it is still possible via overrides for the resolver to use a different version altogether. Conan 2.0 is more strict (it will raise a conflict rather than implicitly override it), but it still possible. 

Using a version range greatly facilitates that if a new version has a security fix, that it can be readily used _without_ making changes to the recipe. Our responsibility when a new version exist that is considered more secure, is to make it available and useable. The rest is up to the users. Using a version range means that the recipe doesn't have to be modified in order for this to happen - which can actually lengthen the amount of time users are potentially using a vulnerable version, as there would need to be a new PR in Conan Center, and go through the CI and review process, and then be published. This is one of the primary reasons we are now using version ranges for OpenSSL - so that we are aligned with the practices of other package managers in this regard.

> ...then why does it matter if the version number doesn't float? You just argued that the binaries will be the same anyway, so...why does it matter if it's version-pinned? 

If a new vulnerability is identified in OpenSSL 3.1.1, and the users themselves also use OpenSSL, the following can happen:
* If the users are using OpenSSL themselves, in a specific version, COnan 2.0 will raise a conflict - their only recourse is to modify the `opentdf-client` and align it - and wait for that to happen here in Conan Center index.
* If all other requirements use OpenSSL as a version range, `opentdf-client` will pin this down to 3.1.1 for the entire dependency graph - thus exposing the users to a now vulnerable version of OpenSSL (nevermind the fact that opentdf-client binaries are the same regardless). 

That's why we want a version range. 

> Q: Why bother going back to re-compile packages that use OpenSSL if/when there is a new release if it's generating the same code anyway? A: Because it might not.

Conan 2.0 default behaviour is to re-compile when the minor component of the version changes. In some cases, all other things being equal, this may very well result in exactly the same binaries being re-generated anyway, under a different package ID.  Some libraries do not have strict guarantees API/ABI guarantees to the minor component, so we rather be safe.

>  A: Because it might not.
When it comes to a stable, strictly backwards compatible C API/ABI, the answer should be "it will not". There's no "might", although I would happily stand corrected if OpenSSL causes these sort of issues. All binaries published by Conan Center use specific versions of the C and C++ runtimes, and users are able to consume these binaries on systems that use entirely different and newer version of these runtimes. They work on all systems where such strict backwards compatibility policies exist. When it comes to OpenSSL we are merely aligning the policy with longstanding convention that exists and predates Conan Center, as I mentioned in my earlier reply.

> I disagree with this, I can't tell you the number of times I've had builds fail because a 3rd party CCI recipe changed underneath me with zero changes to my recipe, code or environment. Yes, there is a way to pin the recipe versions of subordinate libraries to use, but that's (at best) cumbersome, and poorly supported on the CCI website. And pinning recipe versions is essentially the same as pinning library versions because there can be no newer releases in a pinned recipe, so how is that different?

I think there are two issues here which are related, but not the same. I stand by what I said: behaviour in Conan is reproducible when revisions are enabled. That's does not contradict your statement about how you can have builds fail because of other changes coming from CCI, or that it may be cumbersome to operate and mitigate this issues. Conan 2.0 provides significant improvements in this regards, and we will be moving forward with this and updating our recommendations once Conan Center no longer has to support Conan 1.x compatibility. 

> I'm finding CCI difficult to use for production-quality code, where 100% repeatability is vitally important. I also am finding it difficult to recommend to customers as a basis for producing production-quality applications. For 'proof-of-concept' work, keep playing until it mostly works, it's great. For production builds, not so much.

This is something that we are aware of and we are working towards improving. For starters, we are currently in the process of updating our guidelines to discourage using CCI _directly_ in production environments ([PR open](https://github.com/conan-io/docs/pull/3322)), and for teams that have stricter requirements, we will be updating our guidelines on how to mitigate this (for example, the use of lockfiles). In this regard Conan Center is not all that different to repositories/package managers in other languages such as Python/pip/PyPi, Rust/Cargo, javascript/npm - where version ranges are commonplace and users are strongly encouraged to use lockfiles. 

> claiming it's OK to have a floating version number for 'some packages' is difficult to understand.

Version ranges need to play well together with the compatibility promises of library API and ABI with regards to the version numbers. OpenSSL is one of the libraries with the strongest promises, and as I mentioned and have provided links for, it is an open-ended requirement in all other C/C++ package managers as far as I can see.
Hence, we are taking a cautious approach of only enabling version ranges for libraries that have very well-known compatibility promises. This will allow our users to use the latest version of published libraries that are known to be compatible (so should not introduce any changes), also allows user to pin them down to specific versions if they so wish, and allow us in Conan Center to automatically re-build binaries that need to be rebuilt due to a version change, without the need to open a PR and wait for the process - it would just work. This flexibility is what we are trying to achieve.

For years, Conan Center recipes have been manually updated to move to "newer" versions of OpenSSL 1.1.1.x - I've not seen any issues happen in this regard, and I don't expect any to happen as a result of OpenSSL 3.x being a version range.

RubenRBS:
It's progressing properly now :)

patmantru:
Wonderful.  Thanks again @RubenRBS !!

recipes/opentdf-client/all/conanfile.py
patmantru: As someone developing this code for a data protection company, security is job 1.  That includes build repeatability.  That means knowing exactly what is being used.  That means version pinning.  Everything.  Even the big stuff.  I guess the answer here is to (as you suggest) not depend on the CCI recipe for production builds.  TBH, since it generally takes days-to-weeks to get a recipe change through CCI, I've started switching many of our builds to use a recipe pulled from our GH repo, effectively bypassing CCI.  

That makes me sad, as I had hoped CCI would be like (as you mentioned) Pypi/Cargo/npm, a central 'this is the gold standard for package recipes' for C++ modules.  The CCI focus now seems to be on doing whatever is necessary to reduce the volume of recipe updates (short of allowing tags not mentioned in conandata.yml, which would actually be extremely useful), including floating versions 'on some packages' to force more of the builds to work without intervention.  'Flexibility' is not always better.

If lockfiles are the answer, then can I upload them to CCI with the recipe and ensure they are used when the recipe is referenced by a consumer?  Guessing not, because that would defeat CCI's 'flexibility' goal.

As I mentioned in another issue (https://github.com/conan-io/conan-center-index/issues/13994), it seems like the CCI guidance continues to be 'Please submit your recipes to CCI, but you should really make a local copy of everything and use that instead of CCI because we're unreliable', which falls far short of what CCI could be.

patmantru: @jcar87 Since it's clear that I will not be successful in my attempt to retain pinned versions, I have updated the recipe as you require.  I look forward to a smooth path to approvals and a successful merge of this change.

patmantru: @jcar87 After half a dozen attempts, I've discovered that range specs simply create a whole new variety of problems.  Notably, the recipe for jwt-cpp specs "openssl/[>1.1.1c, <1.1.1u]", which means my specification of [>=1.1 <=1.2] fails because 1.1.1u (most recent openSSL version) is not acceptable to jwt-cpp.  Rather than picking 1.1.1t, which would be acceptable to both my package and jwt-cpp, it just fails.  This is like playing whack-a-mole, it moves the versioning problem instead of fixing it.

patmantru: I have been building this using a local recipe, and it works fine using pinned versions.  Now that I am specifying a version range...the opentdf-client version 1.5.0 that includes the OpenSSL 3.1.1 upgrade changes fails because jwt-cpp disallows it.  Going back to pinned versions to see if that works.

patmantru:
@RubenRBS looks like the 'unexpected error' is back again.  :(

recipes/opentdf-client/all/conanfile.py
patmantru: @jcar87 After 12 (!!!) attempts and the better part of 2 full work days, I have this PR building.

Notes:
`jwt-cpp` has its requirement for OpenSSL set as the range `[>1.1.1c, <1.1.1u]` (note the less-than on the upper limit) which means any range spec that pulls in something newer than `1.1.1t` will fail.  The `opentdf-client` package uses OpenSSL 1.1.1 for releases up through 1.4.0.  Since the current 1.x version is `1.1.1u`, setting a range spec of `[>1.1.1q <1.2]` will pick 1.1.1u, which will fail `jwt-cpp`'s range check.  As a concession to your request, even though it is actually propagating the problem, I've added `[>1.1.1q <1.1.1u]` as the range to `[>=1.1.1q <1.1.1u]` just so I can say I did my best here.

Similarly, specifying an OpenSSL 3.1.x version range for my release 1.5.0 recipe will fail, because it's ALSO not within `jwt-cpp`'s `[>1.1.1c, <1.1.1u]` range check.

The ONLY WAY TO GET MY RELEASE 1.5.0 RECIPE FOR TO BUILD CLEANLY given the `jwt-cpp` recipe, is to pin the OpenSSL version as `3.1.1`, which overrides the `jwt-cpp` range check.  This is evident from looking at the 11 previous build attempts that all failed for one range-check reason or another.

I think I've done more than my share of attempting to comply with your request to use version ranges for OpenSSL.  I've proven that it is not possible in this situation given all the constraints.  All I wanted to do was to publish our release 1.4.0 and 1.5.0 recipe updates.  That was supposed to be a couple of new SHA256 lines, and bumping some requirement versions.  Now this has dragged on for weeks, and it's still not in.

Please approve this so it can be merged and I can move on to more important things.

Thanks-
Pat

czoido:
> @czoido still waiting for additional reviews, and the CI doesn't look like it's run yet. Can you provide any guidance on how long I should expect this to take to get merged?

Hi @patmantru, sorry for not responding I was out of the office for some days. I'll have a look again now to this PR. Thanks for your patience.

recipes/opentdf-client/all/conanfile.py
jcar87: > As someone developing this code for a data protection company, security is job 1. That includes build repeatability. That means knowing exactly what is being used. That means version pinning. Everything.

Absolutely agreed, this is important, and possible with Conan Center and where we are headed. I think we have a disagreement as to where this version pinning should be happening. We want to make sure that it is consumers that are able to do version pinning. This is the purpose of lockfiles. [This](https://owasp.org/Top10/A06_2021-Vulnerable_and_Outdated_Components/) is a good summary of recommendations, that places the responsibility on consumers to be aware and known their entire dependency tree  - we give them the options to both pin the version and update to a newer one if necessary, provided the newer one is known to be compatible (as is the case with the OpenSSL 3.x series).

> That makes me sad, as I had hoped CCI would be like (as you mentioned) Pypi/Cargo/npm, a central 'this is the gold standard for package recipes' for C++ modules. 

on the one hand, you have hopes that CCI would be like Pypi/Cargo/npm, but on the other hand these tools and repositories use version ranges by default, which you are arguing against - Some example

* [These](https://github.com/boto/boto3/blob/develop/setup.py#L15-L19) are the requirements of `boto3`, the top most downloaded package in PyPI - notice how each of them uses a version range, as we are advocating to use in Conan Center for OpenSSL. 
* Cargo for Rust uses [version ranges](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) by default, even when a full version is specified - this is because it assumes SemVer compatibility, same as is the case for OpenSSL.
* The [semver](https://www.npmjs.com/package/semver?activeTab=dependencies) npm package (which has 276 million weekly downloads) lists has a dependency that uses version ranges: https://github.com/npm/node-semver/blob/main/package.json#L51 - the `^` notation here is also exactly equivalent as what we are doing for OpenSSL.

In all these cases, it means that users doing `pip install`, `npm install` or cargo builds would get a combination of versions today, and will very likely get a completely different one if they run the same commands 2 weeks from now.  In all of these cases there are well described mechanisms to lock the version resolver to specific versions:
* pip: https://pip.pypa.io/en/stable/user_guide/#constraints-files
* npm: https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-json
* cargo: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html

The practices that we are proposing for Conan Center are in line with  widely adopted practices elsewhere. Conan is no differente: https://docs.conan.io/2.0/tutorial/versioning/lockfiles.html


>  The CCI focus now seems to be on doing whatever is necessary to reduce the volume of recipe updates 

This is categorically untrue. 

>  including floating versions 'on some packages' to force more of the builds to work without intervention. 'Flexibility' is not always better.

The "some packages" is for packages that have known compatibility promises. For libraries that don't, we won't be using them. Widely adopted practice elsewhere. 

>  then can I upload them to CCI with the recipe and ensure they are used when the recipe is referenced by a consumer? Guessing not, because that would defeat CCI's 'flexibility' goal.

While there are some differences, Cargo explains this better than I can: https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries 

We want the end consumers to be in control - they depend on OpenSSL, they should control which version, provided all the libraries they are using are compatible. 

> As I mentioned in another issue (https://github.com/conan-io/conan-center-index/issues/13994), it seems like the CCI guidance continues to be 'Please submit your recipes to CCI, but you should really make a local copy of everything and use that instead of CCI because we're unreliable', which falls far short of what CCI could be.

The PR is still open and subject to change. The advise is, and always has been for Conan Center, to avoid consuming recipes and binaries from Conan Center in a fully unconstrained manner. The advice (especially with Conan 2.0) would be to use lockfiles, and only update the lockfiles when you need to update your dependencies - then you can inspect and solve conflicts if needed or fix compatibility issues. It shouldn't happen "from under you" - this is in line with the recommendations for python packages, npm, cargo, and so on.

jcar87: >  Notably, the recipe for jwt-cpp specs "openssl/[>1.1.1c, <1.1.1u]", which means my specification of [>=1.1 <=1.2] fails because 1.1.1u (most recent openSSL version) is not acceptable to jwt-cpp. Rather than picking 1.1.1t, which would be acceptable to both my package and jwt-cpp, it just fails. This is like playing whack-a-mole, it moves the versioning problem instead of fixing it.

Thanks for reporting this. `openssl/[>1.1.1c, <1.1.1u]` is not accepted range in Conan Center. We have discovered there is a bug with the bump dependencies automerge - and the team is working to fix this. I have opened a PR to address this: https://github.com/conan-io/conan-center-index/pull/19220

I apologise you experienced this issues, I can understand this has been frustating. 


> The ONLY WAY TO GET MY RELEASE 1.5.0 RECIPE FOR TO BUILD CLEANLY given the jwt-cpp recipe, is to pin the OpenSSL version as 3.1.1, which overrides the jwt-cpp range check.

Unfortunately this is a side effect of the `openssl/[>1.1.1c, <1.1.1u]` dependency in the `jwt-cpp` recipe - this is malformed and should have not been merged.  The team is looking into this to avoid this in the future. in Conan 2.0 this is interpreted as `openssl/[>1.1.1c]`, hence why it works when you pin to 3.1.1: in Conan 1 it does an override, and in Conan 2.0 it just matches the range.

Please note that this silent overrides (as experienced in Conan 1.x) are no longer a thing in Conan 2.0: if `jwt-cpp` were only compatible with OpenSSL 1.x and `opentdf-client` is only compatible with OpenSSL 3.x onwards (or in the case where both depend on a specific version), I'm fairly certain this would result in a conflict immediately, and there is no override.  Version ranges avoid the conflicts when things are known to be compatible.

jcar87: We are fixing this issue here: https://github.com/conan-io/conan-center-index/pull/19220 and will make sure `opentdf-client` is relaunched and merged today!

patmantru:
@jcar87 Yes, I tried that.  No, it didn't work for me either.

jcar87 (approved):

czoido (approved):

PR #18901, opened 2023-07-25
mpdecimal/2.4.2: migrate to Conan v2

valgur:


valgur:

jwillikers (approved):

samuel-emrys (approved):

recipes/mpdecimal/2.4.2/conanfile.py
RubenRBS, line 1: Is 2.4.2 version really necessary? I'd be all in for removing this version and simplifying the recipes a lot

RubenRBS (commented):
Thanks! Looks good, but I have a suggestion for removing 2.4.2, let me know what you think @valgur

recipes/mpdecimal/2.4.2/conanfile.py
samuel-emrys: The comment in the [original commit adding cpython](https://github.com/conan-io/conan-center-index/commit/e5e3def4ace88be9118b86245179b4d98180416f) that's relevant to this is:

> * cpython: newer mpdecimal is not compatible with older cpython releases

Which goes along with this code block:

https://github.com/conan-io/conan-center-index/blame/edf1c5a297df013a55cb2d6ac7bf284681a86f33/recipes/cpython/all/conanfile.py#L182-L185

So, to support these older python versions I'd suggest that yes it is necessary

valgur: @RubenRBS I would have favored dropping it as well, but it's still required for older Python versions. I don't see it causing any further maintenance overhead, so keeping it around should not be much of an iasue.

RubenRBS (approved):

PR #18936, opened 2023-07-28
glib/*: QNX 7.1 support

klausholstjacobsen:
glib/2.76.1
glib/2.76.2
glib/2.76.3
glib/2.77.0

Added support for qnx 7.1 and onwards.

---


valgur:
Since glib releases new versions fairly frequently, the amount of additional patches to maintain this PR adds is not ideal.
1. Could these be and have you perhaps already shared these patches upstream as well?
2. Could some of the less stable patches be replaced with a `replace_in_file()` or a `save(..., append=True)` call?

jcar87:
Hi @klausholstjacobsen - thank you for your contribution. While we're happy to support platforms beyonds the one covered by our CI service, this PR introduces patches to C/C++ code that goes beyond we can validate. We don't know to what extent this was tested or validated, and we would not offer this to our users without minimal due diligence. 

I would strongly suggest that the changes needed to support QNX 7.0 are proposed upstream to the glib maintainers, rather than the Conan Center recipe. If these patches already exist in an upstream version, please mention this in the PR description for traceability, along with a justification for backporting them to all other older versions.

Otherwise, I'm afraid this is not something we would be inclined to consider for merge.

klausholstjacobsen:
@jcar87 You are absolutely correct. I was just sitting this morning contemplating this.
I have thus chosen to remove all patches related to qnx 7.0
The good news is that the patched code is built into qnx 7.1, so glib will build for qnx 7.1 with only a few remaining modifications to the recipe!

recipes/glib/all/conanfile.py
```diff
@@ -97,10 +100,17 @@ def generate(self):
...
         tc.project_options["libmount"] = "enabled" if self.options.get_safe("with_mount") else "disabled"
-        if self.settings.os == "FreeBSD":
+        if self.settings.os == "FreeBSD" or self.settings.os == "Neutrino":
             tc.project_options["xattr"] = "false"
         tc.project_options["tests"] = "false"
         tc.project_options["libelf"] = "enabled" if self.options.get_safe("with_elf") else "disabled"
+
+        if self.settings.os == "Neutrino":
+            tc.cross_build["host"]["system"] = "qnx"
+            tc.c_args = ["-Wno-nonnull", "-Wno-format-nonliteral", "-Wno-unused-variable", "-Wno-unused-function", "-Wno-sign-compare", "-Wno-error=implicit-function-declaration", "-Wno-int-conversion"]
```
jcar87, line 110: Are this strictly necessary? They're only disabling warnings - if glib has "warnings as errors" i would disable that one setting, rather than disable some specific ones here. The qnx compiler is actually gcc - I would expect the same warnings to appear when using gcc for any other platform

danimtb: Let's try without this flags and try to move this PR forward 😄 
```suggestion
```

jcar87 (approved):

danimtb (approved):

PR #18944, opened 2023-07-28
nmslib: migrate to Conan v2

valgur:


RubenRBS (dismissed):

recipes/nmslib/all/conanfile.py
```diff
@@ -1,66 +1,81 @@
...
+        cmake = CMake(self)
         cmake.install()
+        fix_apple_shared_install_name(self)
 
     def package_info(self):
         self.cpp_info.libs = ["NonMetricSpaceLib"]
-        if self.settings.os == "Linux":
-            self.cpp_info.system_libs = ["pthread"]
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs = ["pthread", "m", "mvec"]
```
jcar87, line 81:
```suggestion
            self.cpp_info.system_libs = ["pthread", "m"]
            if self.settings.arch in ["x86", "x86_64"]:
                self.cpp_info.system_libs.append("mvec")
```

The `mvec` library from gcc is only available on Intel systems as far as I know - I suspect this has come up before where users on aarch64 Linux had issues with `mvec` being unconditionally listed as a system lib

RubenRBS (approved):

recipes/nmslib/all/conanfile.py
valgur: I agree, it looked suspect to me as well. On my Ubuntu machine the ld script in `libm.a` adds `mvec` already adds transitively, so it is probably not necessary to separately list it in the recipe:
```
/* GNU ld script
*/
OUTPUT_FORMAT(elf64-x86-64)
GROUP ( /usr/lib/x86_64-linux-gnu/libm-2.31.a /usr/lib/x86_64-linux-gnu/libmvec.a )
`

SSE4 (approved):

PR #18947, opened 2023-07-28
libest: migrate to Conan v2

valgur:


recipes/libest/all/conanfile.py
```diff
@@ -1,80 +1,100 @@
...
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("openssl/1.1.1q")
+        self.requires("openssl/1.1.1w", transitive_headers=True, transitive_libs=True)
```
uilianries, line 53:
```suggestion
        self.requires("openssl/[>=1.1 <4]", transitive_headers=True, transitive_libs=True)
```

In case it does not support OpenSSL 3.x, you still can use range, but `[>=1.1 <3]`

valgur: Thanks! Using `<2` did not work for some reason.

valgur: @uilianries Yeah, this fails on Conan v1 for some reason. Quite annoying.

uilianries: What error?

valgur: ERROR: Version range '>=1.1 <3' from requirement 'openssl/[>=1.1 <3]' required by 'libest/3.2.0' could not be resolved in remote 'conan-center'

Only in Conan v1.

uilianries: Thank you, I'll take a look.

uilianries (approved):
Conan 1.x can not solve ranged versions using the format 1.1.1(suffix).

jwillikers (approved):

PR #18956, opened 2023-07-28
resiprocate: migrate to Conan v2

valgur:
.


recipes/resiprocate/all/conanfile.py
```diff
@@ -1,94 +1,104 @@
...
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
 
     def requirements(self):
         if self.options.with_ssl:
-            self.requires("openssl/1.1.1q")
+            self.requires("openssl/[>=1.1 <4]")
```
SpaceIm, line 59: Didn't you say in a previous commit that it was not compatible with openssl 3.x?

SpaceIm: But because pre-built packages of libmysqlclient are built against openssl 3.x, you'll have to disable with_ssl or with_mysql option in this recipe, otherwise it will never pass c3i due to missing packages.

valgur:
@uilianries It looks like the OpenSSL version ranges are really not working as well as might be expected. `libmysqlclient` has been built with `openssl/3.x` and the build fails due to a missing `libysqlclient` binary when using `openssl/1.1.1w` as the requirement, but finds the binary when OpenSSL 3.x is used.
I'm not sure whether this happens only when the exact version differs or just the major version, but anyway, this makes renders the OpenSSL version ranges a bit counterproductive in CCI. It replaces a clear version conflict with a very vague "missing binary" one.

jwillikers (approved):

uilianries (approved):
LGTM. Conan 1.x has limitations with openssl 1.1.1 version range and reciprocate does not support OpenSSL 3.x https://github.com/resiprocate/resiprocate/discussions/242

PR #18957, opened 2023-07-28
liboping: migrate to Conan v2

valgur:


RubenRBS (approved):
Just double checked the license in https://noping.cc/#license and seems we're ok, as we are packaging the utilities too as far as I can see

recipes/liboping/all/conanfile.py
```diff
@@ -1,19 +1,26 @@
...
-required_conan_version = ">=1.33.0"
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import is_apple_os, fix_apple_shared_install_name
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rm, rmdir
+from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.layout import basic_layout
+from conan.tools.microsoft import is_msvc
+
+required_conan_version = ">=1.47.0"
```
SpaceIm, line 11: it's 1.53.0 due to rm_safe

PR #18965, opened 2023-07-29
vsg ImGui integration

psyinf:
**vsgImGui 0.1.0**

Add-on for vsg(VulkanSceneGraph) to integrate imgui
---

### DIFF
--- a/recipes/vsgimgui/all/conanfile.py
+++ b/recipes/vsgimgui/all/conanfile.py
@@ -74,7 +74,7 @@ class VsgImGuiConan(ConanFile):
                 )
 
     def source(self):
-       get(self, **self.conan_data["sources"][self.version], destination=self.source_folder, strip_root=True)
+        get(self, **self.conan_data["sources"][self.version], destination=self.source_folder, strip_root=True)
 
     def generate(self):
         tc = CMakeToolchain(self)
--- a/recipes/vsgimgui/all/conanfile.py
+++ b/recipes/vsgimgui/all/conanfile.py
@@ -26,8 +26,6 @@ class VsgImGuiConan(ConanFile):
         "shared": False,
         "fPIC": True, 
     }
-    #to keep backends for imgui
-    keep_imports = True
 
     @property
     def _min_cppstd(self):
@@ -90,9 +88,10 @@ class VsgImGuiConan(ConanFile):
 
         deps.generate()
 
-    def imports(self):
-        self.copy("imgui_impl_vulkan.*", dst=os.path.join(self.source_folder,"src", "imgui", "backends"), src="./res/bindings")
-                  
+        copy(self, "imgui_impl_vulkan.*", os.path.join(self.dependencies["imgui"].package_folder, "res", "bindings"), 
+                   os.path.join(self.source_folder, "src", "imgui", "backends"))
+ 
+
     def build(self):
         apply_conandata_patches(self)
         cmake = CMake(self)
@@ -123,8 +122,3 @@ class VsgImGuiConan(ConanFile):
         if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.system_libs.append("pthread")
 
-        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
-        self.cpp_info.filenames["cmake_find_package"] = "vsgImGui"
-        self.cpp_info.filenames["cmake_find_package_multi"] = "vsgImGui"
-        self.cpp_info.names["cmake_find_package"] = "VSGIMGUI"
-        self.cpp_info.names["cmake_find_package_multi"] = "vsgImGui"
--- a/recipes/vsgimgui/all/conanfile.py
+++ b/recipes/vsgimgui/all/conanfile.py
@@ -53,7 +53,7 @@ class VsgImGuiConan(ConanFile):
        
     def requirements(self):
         self.requires("vulkan-loader/1.3.239.0", transitive_headers=True)
-        self.requires("vsg/[>1.0.5]")
+        self.requires("vsg/1.0.5")
         self.requires("imgui/1.89.5")
         self.requires("implot/0.14")
         #self.requires("vsgxchange/1.0.2")
--- a/recipes/vsgimgui/all/conanfile.py
+++ b/recipes/vsgimgui/all/conanfile.py
@@ -1,0 +1,130 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.microsoft import check_min_vs, is_msvc_static_runtime, is_msvc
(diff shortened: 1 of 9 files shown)


recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+       
+    def requirements(self):
+        self.requires("vulkan-loader/1.3.239.0", transitive_headers=True)
+        self.requires("vsg/[>1.0.5]")
```
czoido, line 56: Hi @psyinf, thanks a lot for the contribution. 
Please, note that we are only accepting ranges for openssl and cmake for the moment: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges
Also, does this need >1.0.5? There's only until 1.0.5 i Conan Center
```suggestion
        self.requires("vsg/1.0.5")
```

recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+    
+        tc.generate()
+        
+        deps = CMakeDeps(self)
+
+        deps.generate()
+
+    def imports(self):
+        self.copy("imgui_impl_vulkan.*", dst=os.path.join(self.source_folder,"src", "imgui", "backends"), src="./res/bindings")
+
```
czoido, line 95: Now, in Conan 2.X (compatible with 1.X) the way to do this is in the generate() method. Please, check this code as an example: https://github.com/conan-io/examples2/blob/50f1192b220107e9135bae972356500a1dd05a66/examples/libraries/imgui/introduction/conanfile.py#L17-L22

czoido:
```suggestion
```

recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True, 
+    }
+    #to keep backends for imgui
+    keep_imports = True
```
czoido, line 30: Please, check my comment in the imports() method about this
```suggestion
```

recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        get(self, **self.conan_data["sources"][self.version], destination=self.source_folder, strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+
+        if is_msvc(self):
+            tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+        
+        tc.variables["VSG_MACRO_PATH"] = os.path.join(self.deps_cpp_info["vsg"].rootpath, "lib/cmake/vsg/").replace("\\","/")
+
```
czoido, line 86: I think this probably should be handled in the vsg recipe adding the macro, something like:

```
self.cpp_info.set_property("cmake_build_modules", [os.path.join("lib", "cmake", "vsg", "vsgMacros.cmake")])
```

recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        self.cpp_info.set_property("cmake_target_name", "vsgImGui::vsgImGui")
+
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs.append("pthread")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.filenames["cmake_find_package"] = "vsgImGui"
+        self.cpp_info.filenames["cmake_find_package_multi"] = "vsgImGui"
+        self.cpp_info.names["cmake_find_package"] = "VSGIMGUI"
+        self.cpp_info.names["cmake_find_package_multi"] = "vsgImGui"
```
czoido, line 130:
```suggestion
```
Being this a new recipe, at the moment we are I think we can omit these and also remove the `test_v1_package`

recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+       
+    def requirements(self):
+        self.requires("vulkan-loader/1.3.239.0", transitive_headers=True)
+        self.requires("vsg/[>1.0.5]")
+        self.requires("imgui/1.89.5")
+        self.requires("implot/0.14")
```
czoido, line 58:
```suggestion
        self.requires("vsg/1.0.5", transitive_headers=True)
        self.requires("imgui/1.89.8", transitive_headers=True)
        self.requires("implot/0.16")
```

recipes/vsgimgui/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        if is_msvc(self):
+            tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+        
+        tc.variables["VSG_MACRO_PATH"] = os.path.join(self.deps_cpp_info["vsg"].rootpath, "lib/cmake/vsg/").replace("\\","/")
+    
+        tc.generate()
+        
+        deps = CMakeDeps(self)
+
+        deps.generate()
```
czoido, line 91:
```suggestion
        deps.generate()

        copy(self, "imgui_impl_vulkan.*", os.path.join(self.dependencies["imgui"].package_folder, "res", "bindings"), 
                   os.path.join(self.source_folder, "src", "imgui", "backends"))
 
```

recipes/vsgimgui/all/patches/0.1.0-0001-includeMacro.patch
```diff
@@ -0,0 +1,98 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index eb0a39c..1ae1ff5 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -23,34 +23,11 @@ if (VULKAN_SDK)
+ endif()
+ 
+ find_package(vsg 1.0.5)
++include(${VSG_MACRO_PATH}/vsgMacros.cmake)
```
czoido, line 9: This should not be necessary if we declare `cmake_build_modules` property in vsg for that macro

psyinf: I'll update the vsg package and then re-integrate this

PR #18978, opened 2023-07-31
rdma-core: add new recipe

valgur:
Adds https://github.com/linux-rdma/rdma-core

Required by `openmpi` (which uses `libibverbs`), `libfabric` (uses `libefa` and `librdmacm`) and possibly `libpcap` packages.


wadimklincov:
Building with `build_librdmacm` enabled fails, at least on my system (Debian 12, gcc 11):

> build_assert.h:23:33: error: size of unnamed array is negative
   23 |         do { (void) sizeof(char [1 - 2*!(cond)]); } while(0)
      |                                 ^
src/librdmacm/acm.c:105:17: note: in expansion of macro ‘BUILD_ASSERT’
  105 |                 BUILD_ASSERT(sizeof(IBACM_SERVER_PATH) <=
      |                 ^~~~~~~~~~~~
[90/138] Building C object libibverbs/CMakeFiles/ibverbs.dir/marshall.c.o

Looks like the generated string for `IBACM_SERVER_PATH` exceeds the (108 byte?) limit of `addr.unx.sun_path`: `/home/developer/.conan/data/rdma-core/49.0/_/_/package/c965b2ab63caf8e11f0c51cc66e6a9f195743f00/var/run/ibacm.port`. Patching `config.h.in` seems to help. 

Since this doesn't depend on systemd, perhaps could be patched by default?

recipes/rdma-core/all/conanfile.py
```diff
@@ -0,0 +1,133 @@
...
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("libnl/3.8.0")
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.requires("libudev/system")
```
franramirez688, line 53: The validation is already in the `validate()` function.

```suggestion
        self.requires("libudev/system")
```

franramirez688 (approved):
LGTM 😁 Only a minor suggestion

franramirez688:
> Looks like the generated string for `IBACM_SERVER_PATH` exceeds the (108 byte?) limit of `addr.unx.sun_path`: `/home/developer/.conan/data/rdma-core/49.0/_/_/package/c965b2ab63caf8e11f0c51cc66e6a9f195743f00/var/run/ibacm.port`. Patching `config.h.in` seems to help.
> 
> Since this doesn't depend on systemd, perhaps could be patched by default?

Oops, I did not see that. @valgur do you have any idea on how to solve this?

recipes/rdma-core/all/conanfile.py
```diff
@@ -0,0 +1,133 @@
...
+from conan.tools.files import get, copy, rmdir, load, save, replace_in_file
+from conan.tools.gnu import PkgConfigDeps
+
+required_conan_version = ">=1.53.0"
+
+class PackageConan(ConanFile):
+    name = "rdma-core"
+    description = ("RDMA core userspace libraries and daemons. "
+                   "Provides userspace components for the Linux Kernel's drivers/infiniband subsystem.")
+    license = ("GPL-2.0", "Linux-OpenIB", "BSD-2-Clause")
```
jwillikers, line 17: I think this is GPL-2.0-only.

```suggestion
    license = ("GPL-2.0-only", "Linux-OpenIB", "BSD-2-Clause")
```

PR #18982, opened 2023-07-31
depot_tools: migrate to Conan v2

valgur:


recipes/depot_tools/all/conanfile.py
```diff
@@ -1,24 +1,36 @@
...
+        export_conandata_patches(self)
 
-    @property
-    def _source_subfolder(self):
-        return os.path.join(self.source_folder, "source_subfolder")
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def package_id(self):
+        del self.info.settings.arch
```
RubenRBS, line 31: How is arch deleted from an application? That seems like it's going to break super easy no?

RubenRBS (commented):
Concern about removing arch from an application

recipes/depot_tools/all/conanfile.py
valgur: This project consists of Python scripts only, so it should not be dependent on architecture. Looking at it again, `self.info.clear()` might be even more appropriate here. I don't think it's os-dependent either.

RubenRBS: Ohhh, that makes a lot of sense!

recipes/depot_tools/all/conanfile.py
```diff
@@ -30,63 +39,68 @@ def _dereference_symlinks(self):
...
             for symlink in symlinks:
                 dest = os.readlink(symlink)
                 os.remove(symlink)
                 shutil.copy(os.path.join(root, dest), symlink, follow_symlinks=False)
                 self.output.info("Replaced symlink '%s' with its destination file '%s'" % (symlink, dest))
 
-    def source(self):
-        tools.get(**self.conan_data["sources"][self.version], destination=self._source_subfolder)
+    def build(self):
+        get(self, **self.conan_data["sources"][self.version], destination=self.source_folder)
```
RubenRBS, line 51: Why is this part of the build method instead of the source one? It seems like the sources are configuration independent

RubenRBS (commented):
Question about downloading in build()

recipes/depot_tools/all/conanfile.py
valgur: Mostly just following the example in the docs:
https://github.com/conan-io/conan-center-index/blob/ff8b4aaf4ea7e4af4306ab956ce31f1ff026308c/docs/package_templates/prebuilt_tool_package/all/conanfile.py#L40-L46
As far as I can tell it serves the same purpose as `no_copy_source=True`.

valgur:
@RubenRBS Ping.

RubenRBS (approved):

juansblanco (approved):

PR #19021, opened 2023-08-03
add ethash

pwqbot:
Specify library name and version:  **ethash/1.0.1**

---


recipes/ethash/all/conanfile.py
```diff
@@ -0,0 +1,102 @@
...
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "ethash")
+        self.cpp_info.set_property("cmake_target_name", "ethash::ethash")
+
+        self.cpp_info.names["cmake_find_package"] = "ethash"
+        self.cpp_info.names["cmake_find_package_multi"] = "ethash"
+
+        if self.settings.os == "Linux":
+            self.cpp_info.system_libs.append("m")
+        self.cpp_info.components["keccak"].set_property("cmake_file_name", "keccak")
```
valgur, line 95:
```suggestion
```
cmake_file_name has no effect for components.

recipes/ethash/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,19 @@
+from conans import ConanFile, CMake
```
valgur, line 1: test_v1_package should no longer be included for new packages.

recipes/ethash/all/conanfile.py
```diff
@@ -0,0 +1,102 @@
...
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "ethash")
+        self.cpp_info.set_property("cmake_target_name", "ethash::ethash")
+
+        self.cpp_info.names["cmake_find_package"] = "ethash"
+        self.cpp_info.names["cmake_find_package_multi"] = "ethash"
+
```
valgur, line 92:
```suggestion
```
These values are already the defaults.

recipes/ethash/all/test_v1_package/conanfile.py
pwqbot: Thanks! Should I just delete test_v1_package/ ?

valgur: Yes, it can be dropped entirely.

recipes/ethash/all/conanfile.py
```diff
@@ -0,0 +1,95 @@
...
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        if self.settings.os == "Linux":
+            self.cpp_info.system_libs.append("m")
+        self.cpp_info.components["keccak"].set_property("cmake_target_name", "keccak::keccak")
+        self.cpp_info.components["keccak"].libs = ["keccak"]
+
+        self.cpp_info.components["ethash"].set_property("cmake_file_name", "ethash")
```
valgur, line 92:
```suggestion
```
Remove this one as well.

danimtb (approved):

recipes/ethash/all/conanfile.py
```diff
@@ -0,0 +1,94 @@
...
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.generate()
+
+    def validate(self):
+        if self.settings.os == "Windows":
+            raise ConanInvalidConfiguration("ethash does not support windows")
```
uilianries, line 53: Do you have any reference that it does not work on Windows from the upstream? I could not find some valid argument in the upstream: https://github.com/search?q=repo%3Achfast%2Fethash%20windows&type=code

Here it's noted that it can be built on Windows: https://github.com/chfast/ethash/issues/88

uilianries (changes requested):
Blocking to have more information about Windows build.

uilianries (dismissed):
I did few changes in this PR:

- After checking the upstream, it requires C++14, not C++20;
- Both ethash and global-context libraries are optional (enabled by default);
- CMake variable should be cached because options are declared before project();
- The global-context has its own cmake target;
- The global-context is only enabled in case ethash is built.

recipes/ethash/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+               "build_global_context": [True, False]}
+    default_options = {"shared": False,
+                       "fPIC": True,
+                       "build_ethash": True,
+                       "build_global_context": True}
+
+    @property
+    def _min_cppstd(self):
+        return 14
+
```
uilianries, line 34:
```suggestion

    def config_options(self):
        if self.settings.os == "Windows":
            del self.options.fPIC

    def configure(self):
        if self.options.shared:
            self.options.rm_safe("fPIC")

```

uilianries (approved):
LGTM

RubenRBS (approved):
Thanks a lot! :)

PR #19030, opened 2023-08-03
openpam/20190224: Added openpam recipe

klausholstjacobsen:
openpam/20190224

Added recipe

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #19031, opened 2023-08-03
openssh/9.1p1: Added openssh recipe

klausholstjacobsen:
openssh/9.1p1
openssh/8.1p1

Added recipe!

---


recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("zlib/[>=1.2.12]")
+        if self.options.with_openssl:
+            self.requires("openssl/1.1.1q")
```
RubenRBS, line 68: Is this library only compatible with this version of openssl or is it possible to add a [>=1.1 <4] range?

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def configure(self):
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("zlib/[>=1.2.12]")
```
RubenRBS, line 66:
```suggestion
        self.requires("zlib/[>=1.2.13 <2]")
```

We're planning on adding zlib to the list of allowed recipes with evrsion ranges as we being to feel more confident that version ranges are the right aapproach, but we're not quite yet there.

BUT, as this might still take a few days to get applied, I'm suggesting the provisional range for zlib :)

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        if self.options.with_openssl:
+            self.requires("openssl/1.1.1q")
+        if self.options.with_pam:
+            self.requires("openpam/20190224")
```
RubenRBS, line 70: For traceability, needs https://github.com/conan-io/conan-center-index/pull/19030 to get merged first

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+        if self.options.with_sandbox != 'auto':
+            tc.configure_args.append("--with-sandbox={}".format(self.options.with_sandbox))
+
+        tc.generate()
+
+    def build(self):
+        self._patch_sources()
+
+        autotools = Autotools(self)
+        env = VirtualRunEnv(self)
```
RubenRBS, line 105: Any explanation as to why this is needed?

RubenRBS (commented):
Thanks a lot! I have a few comments, but the overall structure of the recipe looks good, thank you as always :)

recipes/openssh/all/conandata.yml
```diff
@@ -0,0 +1,12 @@
+sources:
+  "9.1p1":
+    url: "https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-9.1p1.tar.gz"
+    sha256: "19f85009c7e3e23787f0236fbb1578392ab4d4bf9f8ec5fe6bc1cd7e8bfdd288"
+  "8.1p1":
+    url: "https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-8.1p1.tar.gz"
+    sha256: "02f5dbef3835d0753556f973cd57b4c19b6b1f6cd24c03445e23ac77ca1b93ff"
+patches:
+  "9.1p1-neutrino-7.0":
```
RubenRBS, line 9: This needs to be changed. You should add a `patch_system: "neutrino-7.0"` and filter by that key instead. Else, changes in the patches for an unrelated version will change the revisions of every other version!

recipes/openssh/all/conanfile.py
klausholstjacobsen: It will not work with openssl 3:
```
checking OpenSSL library version... configure: error: Unknown/unsupported OpenSSL version ("30100020 (OpenSSL 3.1.2 1 Aug 2023)")
```
But I can change it to [>=1.1 <3]

recipes/openssh/all/conanfile.py
klausholstjacobsen: Done!

recipes/openssh/all/conanfile.py
klausholstjacobsen: When build with `*:shared=True`
This error is produced during configure:
```
configure:15472: cc -o conftest  -m64 -O3 -pipe -Wno-error=format-truncation -Wall -Wpointer-arith -Wuninitialized -Wsign-compare -Wformat-security -Wsizeof-pointer-memaccess -Wno-pointer-sign -Wno-unused-result -Wimplicit-fallthrough -Wmisleading-indentation -fno-strict-aliasing -D_FORTIFY_SOURCE=2 -ftrapv -fzero-call-used-regs=all -fno-builtin-memset -fstack-protector-strong -fPIE -I/home/khj/.conan2/p/b/opensb564e5c3c60b5/p/include  -I/home/khj/.conan2/p/b/opensb564e5c3c60b5/p/include -I/home/khj/.conan2/p/zlib5fd5adbde7f5c/p/include -DNDEBUG -D_XOPEN_SOURCE=600 -D_BSD_SOURCE -D_DEFAULT_SOURCE -L/home/khj/.conan2/p/b/opensb564e5c3c60b5/p/lib  -L/home/khj/.conan2/p/b/opensb564e5c3c60b5/p/lib -L/home/khj/.conan2/p/zlib5fd5adbde7f5c/p/lib -m64 -Wl,-z,relro -Wl,-z,now -Wl,-z,noexecstack -fstack-protector-strong -pie conftest.c -lcrypto  -lssl -lcrypto -lz -ldl -lpthread -lrt >&5
configure:15472: $? = 0
configure:15472: ./conftest
./conftest: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory
configure:15472: $? = 127
```

The only way I can let it know where openssl .so files are is with a virtual run env.

recipes/openssh/all/conandata.yml
klausholstjacobsen: So what you mean is that I should apply this filtering when copying in `export_sources(...)` in order to only export patches for the package I am building?

klausholstjacobsen: No, my above suggestion cannot be true as sources must be configuration independent.
Can you elaborate on what you mean? 
How can this happen: `changes in the patches for an unrelated version will change the revisions of every other version!`?

klausholstjacobsen: @RubenRBS I have committed a patch solution and added os and os_version keys to conan_data. Is that what you were looking for?

klausholstjacobsen: That, unfortunately, failed due to a hook preventing patch data having a `patch_os` and `patch_os_version` key

RubenRBS: Agghh, let me check with the team to see if taking care of the hook is the way to go, thanks for your patience in this

klausholstjacobsen: @RubenRBS Any news?

klausholstjacobsen: @RubenRBS Any news, or should I go another route implementing this?

Regards
Klaus

RubenRBS: Hi, sorry for the delay, thank you a lot for your patience - This is similar to an issue raised in https://github.com/conan-io/conan-center-index/pull/19529, which we still need to come to a conclusion on.

Note that @uilianries has started work on the linters migration, so they can probably chime in if they think it would be faster to move away from this implementation, or else wait for the new linters to allow for this

RubenRBS: Hi @klausholstjacobsen disregard my last message, we've come to the conclusion that this is something worth supporting, will make sure the linter does not complain for now, thanks again for your patience :)

RubenRBS:
Sorry about the delay, this got internally labelled as waiting for you by mistake so we didn't come back around the PR until now - Will try to push this along, you have waited enough, thanks a lot for your patience :)

danimtb (approved):

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,132 @@
...
+    name = "openssh"
+    description = "The OpenSSH (portable) suite of secure connectivity tools"
+    license = "BSD"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/openssh/openssh-portable"
+    topics = ("security", "cryptography", "login", "keychain", "file-sharing")
+    package_type = "application"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "with_openssl": [True, False],
```
jwillikers, line 25: Does the project support using `libressl` as an alternative?

```suggestion
        "with_libcrypto": [None, "libressl", "openssl],
```

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,132 @@
...
+        "with_pam": [True, False],
```
jwillikers, line 26: While we don't have a Linux PAM package now, it might be good build this option out in a way where that's possible in the future, i.e.

```suggestion
        "with_pam": [None, "OpenPAM"],
```

```suggestion
        "with_pam": [True, False],
```

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,132 @@
...
+from conan.tools.layout import basic_layout
+from os.path import join
+
+required_conan_version = ">=1.54.0"
+
+
+class PackageConan(ConanFile):
+    name = "openssh"
+    description = "The OpenSSH (portable) suite of secure connectivity tools"
+    license = "BSD"
```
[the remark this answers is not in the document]
valgur: Here's the SPDX string used by Fedora (https://packages.fedoraproject.org/pkgs/openssh/openssh-clients/):
`BSD-3-Clause AND BSD-2-Clause AND ISC AND SSH-OpenSSH AND ssh-keyscan AND sprintf AND LicenseRef-Fedora-Public-Domain AND X11-distribute-modifications-variant`

recipes/openssh/all/patches/pselect_timeout.patch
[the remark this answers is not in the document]
klausholstjacobsen: The patch is not necessary. I will remove it!

recipes/openssh/all/conanfile.py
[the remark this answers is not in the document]
klausholstjacobsen: So, should I put "SSH-OpenSSH" inthere?

recipes/openssh/all/conanfile.py
klausholstjacobsen: Updating recipe to [>=1.1 <4]

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,132 @@
...
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "with_openssl": [True, False],
+        "with_pam": [True, False],
+        "with_sandbox": ["auto", "no", "capsicum", "darwin", "rlimit", "seccomp_filter", "systrace", "pledge"]
+    }
+    default_options = {
+        "with_openssl": True,
+        "with_pam": False,
+        "with_sandbox": "auto"
```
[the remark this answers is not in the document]
klausholstjacobsen: adding options!

recipes/openssh/all/conanfile.py
klausholstjacobsen: No it does not seem to have config options for that!

recipes/openssh/all/conanfile.py
klausholstjacobsen: Updating option

recipes/openssh/all/conanfile.py
jwillikers: I think it should still work with `libressl` without a config option. The `libressl` package should work since it should be a drop-in replacement.

recipes/openssh/all/conanfile.py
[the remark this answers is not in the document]
jwillikers: Thanks!

jwillikers (changes requested):

recipes/openssh/all/conanfile.py
klausholstjacobsen: Added support for libressl

jwillikers (approved):
Thanks!

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("zlib/[>=1.2.11 <2]")
+        if self.options.with_libcrypto  == "openssl":
+            self.requires("openssl/[>=1.1 <=3.1]")
+        elif self.options.with_libcrypto  == "libressl":
+            self.requires("libressl/[>=2.9 <=3.5]")
```
RubenRBS, line 56:
```suggestion
            self.requires("libressl/3.5.3")
```

Assuming the original upper limit meant <3.6 (Else 3.5 would be in, but 3.5.1 wouldn't)

I'm not cnfident in the backwards compatibility of libress to allow version ranges for its version, so pinning for the latest available one currently

RubenRBS (commented):
Hi! Sorry about the delay - the PR fell thru the cracks and got lost on the sea of PRs

The good news is that the PR looks great, one minor thing remaining before merging - feel free to ping me directly once that's done so that I can approve asap :)

klausholstjacobsen:
@RubenRBS your suggestion is committed! Have a nice weekend!

jwillikers (approved):

recipes/openssh/all/conandata.yml
```diff
@@ -0,0 +1,7 @@
+sources:
+  "9.6p1":
+    url: "https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-9.6p1.tar.gz"
+    sha256: "910211c07255a8c5ad654391b40ee59800710dd8119dd5362de09385aa7a777c"
+  "9.1p1":
+    url: "https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-9.1p1.tar.gz"
+    sha256: "19f85009c7e3e23787f0236fbb1578392ab4d4bf9f8ec5fe6bc1cd7e8bfdd288"
```
valgur, line 7: Is there a specific need for the non-latest version? If not, I would drop it.

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+    homepage = "https://www.openssh.com/portable.html"
+    topics = ("security", "cryptography", "login", "keychain", "file-sharing", "ssh")
+    package_type = "application"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "with_libcrypto": [None, "libressl", "openssl"],
+        "with_pam": [None, "openpam"],
+        "with_selinux": [True, False],
+        "with_libedit": [True, False],
+        "with_sandbox": ["auto", "no", "capsicum", "darwin", "rlimit", "seccomp_filter", "systrace", "pledge"]
```
valgur, line 27:
```suggestion
        "with_sandbox": [False, "auto", "capsicum", "darwin", "rlimit", "seccomp_filter", "systrace", "pledge"]
```
Would be slightly more consistent with other recipes, imo.

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+    description = "The OpenSSH (portable) suite of secure connectivity tools"
+    license = "SSH-OpenSSH"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://www.openssh.com/portable.html"
+    topics = ("security", "cryptography", "login", "keychain", "file-sharing", "ssh")
+    package_type = "application"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "with_libcrypto": [None, "libressl", "openssl"],
+        "with_pam": [None, "openpam"],
```
valgur, line 24:
```suggestion
        "with_libcrypto": [False, "libressl", "openssl"],
        "with_pam": [False, "openpam"],
```
Not a fan of `None` as an option value. It makes the output of `self.options.get_safe("xyz")` ambiguous. The convention in other recipes is to use `False` in such cases as well.

valgur (changes requested):
LGTM, but I would adjust the negative option values a bit to match other recipes better.

recipes/openssh/all/conandata.yml
klausholstjacobsen: there is a specific need for 9.1p1!

recipes/openssh/all/conanfile.py
```diff
@@ -0,0 +1,130 @@
...
+            openssl = self.dependencies["openssl"]
+            tc.configure_args.append("--with-ssl-dir={}".format(openssl.package_folder))
+        elif self.options.with_libcrypto == "libressl":
+            libressl = self.dependencies["libressl"]
+            tc.configure_args.append("--with-ssl-dir={}".format(libressl.package_folder))
+        else:
+            tc.configure_args.append("--without-openssl")
+
+        if self.options.with_sandbox != 'auto':
+            tc.configure_args.append("--with-sandbox={}".format(self.options.with_sandbox))
```
valgur, line 99:
```suggestion
            with_sandbox = "no" if not self.options.with_sandbox else self.options.with_sandbox
            tc.configure_args.append("--with-sandbox={}".format(with_sandbox))
```

recipes/openssh/all/conanfile.py
franramirez688: IMO, I think that we should have kept the "no" in this case. Variables that are "yes"/"no", ok, let's have True or False as many other recipes, but I don't see the point of mixing boolean and strings in this kind of option. Indeed, it'd avoid the replacement like the comment below https://github.com/conan-io/conan-center-index/pull/19031#discussion_r1538839241

franramirez688 (commented):
Hi @klausholstjacobsen 

Thanks a lot for adding this recipe, and sorry for the long delay in reviewing this PR.

I just added one commit to simplify a couple of things, but it looks great! 👏

Before approving, I have a question. I just tried this command and got some errors in macOS Sonoma 14.4.1  (failed with both *:shared=True and False):

```bash
$ conan create . --version 9.6p1 --build missing -o with_libcrypto="libressl"
. . . 
cc -m64 -O3 -pipe -Wunknown-warning-option -Qunused-arguments -Wall -Wextra -Wpointer-arith -Wuninitialized -Wsign-compare -Wformat-security -Wsizeof-pointer-memaccess -Wno-pointer-sign -Wno-unused-parameter -Wno-unused-result -Wmisleading-indentation -Wbitwise-instead-of-logical -fno-strict-aliasing -mretpoline -D_FORTIFY_SOURCE=2 -ftrapv -fzero-call-used-regs=used -ftrivial-auto-var-init=zero -fno-builtin-memset -fstack-protector-strong   -fPIC -I. -I.. -I/Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat -I/Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat/.. -I/Users/conanuser/.conan2/p/b/libreef807320845de/p/include  -I/Users/conanuser/.conan2/p/b/libreef807320845de/p/include -I/Users/conanuser/.conan2/p/b/zlibf6519252f237d/p/include -DNDEBUG -DHAVE_CONFIG_H -c /Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat/fmt_scaled.c
/Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat/xcrypt.c:88:4: error: call to undeclared function 'explicit_bzero'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
                        explicit_bzero(passwd, strlen(passwd));
                        ^
1 error generated.
make[1]: *** [xcrypt.o] Error 1
/Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat/bcrypt_pbkdf.c:108:2: error: call to undeclared function 'explicit_bzero'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        explicit_bzero(ciphertext, sizeof(ciphertext));
        ^
/Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat/bcrypt_pbkdf.c:177:2: error: call to undeclared function 'freezero'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        freezero(countsalt, saltlen + 4);
        ^
/Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/src/openbsd-compat/bcrypt_pbkdf.c:178:2: error: call to undeclared function 'explicit_bzero'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        explicit_bzero(out, sizeof(out));
        ^
3 errors generated.
make[1]: *** [bcrypt_pbkdf.o] Error 1
make: *** [openbsd-compat/libopenbsd-compat.a] Error 2

openssh/9.6p1: ERROR:
Package '9831a80989c890ca8957350d5a90e23b88ab109c' build failed
openssh/9.6p1: WARN: Build folder /Users/conanuser/.conan2/p/b/opensc8a0037d6a765/b/build-release
ERROR: openssh/9.6p1: Error in build() method, line 110
	autotools.make()
	ConanException: Error 2 while executing
```
Could it be a bug in openssh using `libressl` in macOS? Maybe related to this https://github.com/openssh/openssh-portable/pull/203 ??

For the record, I just tried it in Linux and it works fine.

franramirez688:
> @franramirez688 I recently worked with the `util-linux` project to fix a usage of `explicit_bzero` on macOS that produced the same error. That symbol isn't available on macOS. It appears that their Autotools script checks for this function and defines its own if it is not available. Something must be off in the build system regarding this.

Thanks @jwillikers for the explanation. Given that, would it make sense to add a `validate` step for raising an invalid configuration exception when using `with_libcrypto == "libressl"` and `macOS`?

jwillikers:
@franramirez688 The configure output should note whether `explicit_bzero` was detected or not. In CI, I see `checking for explicit_bzero... no`. Do you see that as well? I don't think this issue has anything to do with `libressl`, since OpenSSH Portable should be declaring and defining this function as needed on macOS. It's required regardless of the implementation of libcrypto, unless OpenSSL is declaring the function.

jwillikers:
@franramirez688 Perhaps `explicit_bzero` is being picked up from the `libressl` project in the configure checks? I wonder if this was fixed in libressl/portable#961.

franramirez688:
> @franramirez688 Perhaps `explicit_bzero` is being picked up from the `libressl` project in the configure checks? I wonder if this was fixed in [libressl/portable#961](https://github.com/libressl/portable/pull/961).

@jwillikers thanks for that one. Indeed, it was the key point. Using `libressl/3.9.1` (https://github.com/conan-io/conan-center-index/pull/23325), that problem would be solved 👏 

Waiting for your PR to be merged, so we can add it directly here.

franramirez688 (approved):

jwillikers (approved):

PR #19070, opened 2023-08-06
asio-grpc: Fix grpc++ linkage

Tradias:
Specify library name and version:  **asio-grpc/***

Fixes https://github.com/conan-io/conan-center-index/issues/18102 in the following ways:

* It seems that no other recipe handles the difference between gRPC secure and unsecure correctly. See for example google-cloud-cpp or opentelemetry-cpp. I think this is a problem with the gRPC recipe itself where `grpc++` should alias `grpc++_unsecure` when `grpc:secure=False` and it should not be addressed in dependent recipes. Linking with both `grpc++` and `grpc++_unsecure` leads to the issues described in this issue: https://github.com/grpc/grpc/issues/21213#issuecomment-740190485.

* It fixes another problem given the following conanfile.txt:
```
[requires]
asio-grpc/2.4.0

[generators]
CMakeDeps
CMakeToolchain

[layout]
cmake_layout

[options]
grpc:secure=True
```
fails with
```
> conan install --build=missing ..

======== Finalizing install (deploy, generators) ========
conanfile.txt: Writing generators to C:\3YOURMIND\test-conan\build\generators
conanfile.txt: Generator 'CMakeDeps' calling 'generate()'
conanfile.txt: ERROR: Traceback (most recent call last):
  File "conan\tools\cmake\cmakedeps\templates\__init__.py", line 35, in render
  File "conan\tools\cmake\cmakedeps\templates\target_configuration.py", line 24, in context
  File "conan\tools\cmake\cmakedeps\templates\target_configuration.py", line 251, in get_deps_targets_names
  File "conan\tools\cmake\cmakedeps\templates\__init__.py", line 90, in get_component_alias
conans.errors.ConanException: Component 'grpc::grpc++_unsecure' not found in 'grpc' package requirement

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "conans\client\generators\__init__.py", line 92, in write_generators
  File "conan\tools\cmake\cmakedeps\cmakedeps.py", line 43, in generate
  File "conan\tools\cmake\cmakedeps\cmakedeps.py", line 85, in content
  File "conan\tools\cmake\cmakedeps\cmakedeps.py", line 103, in _generate_files
  File "conan\tools\cmake\cmakedeps\templates\__init__.py", line 37, in render
conans.errors.ConanException: error generating context for 'asio-grpc/2.5.1': Component 'grpc::grpc++_unsecure' not found in 'grpc' package requirement

ERROR: Error in generator 'CMakeDeps': error generating context for 'asio-grpc/2.5.1': Component 'grpc::grpc++_unsecure' not found in 'grpc' package requirement
```

* It is a breaking change of this recipe but I do not know what the policy is regarding that in conan and I think it is ok. After all, it is also a bug fix. If people use, for example, asio-grpc and opentelemetry-cpp today then they would link with grpc++ and grpc++_unsecure. In other words, asio-grpc is currently broken when other dependencies (from conan-center) that themselves depend on gRPC are used.

---

- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated. ~~No because these instructions don't seem to work anymore with conan 2.0.9: https://github.com/conan-io/hooks#conan-config-as-installer~~ Tested with conan v1 now.
```
PS D:\Programming\conan-center-index\recipes\asio-grpc> conan config set hooks.conan-center
usage: conan config [-h] [-v [V]] {home,install,list,show} ...
conan config: error: argument subcommand: invalid choice: 'set' (choose from 'home', 'install', 'list', 'show')
ERROR: Exiting with code: 2
````

### DIFF
--- a/recipes/asio-grpc/all/conanfile.py
+++ b/recipes/asio-grpc/all/conanfile.py
@@ -111,7 +111,7 @@ class AsioGrpcConan(ConanFile):
 
         build_modules = [os.path.join("lib", "cmake", "asio-grpc", "AsioGrpcProtobufGenerator.cmake")]
 
-        self.cpp_info.requires = ["grpc::grpc++_unsecure"]
+        self.cpp_info.requires = ["grpc::grpc++"]
         if self.options.backend == "boost":
             self.cpp_info.defines = ["AGRPC_BOOST_ASIO"]
             self.cpp_info.requires.append("boost::headers")


Tradias:
This pull request is independent from https://github.com/conan-io/conan-center-index/pull/19069

and supersedes https://github.com/conan-io/conan-center-index/pull/18529 (tbd).

weatherhead99:
I don't think this addresses the exact problem I mentioned in #18102 - my CMakeLists.txt file uses ASIO_GRPC_DISABLE_AUTOLINK because it has to build without using conan in some instances. Choosing to link either grpc++ or grpc++_unsecure in asio-grpc therefore inevitably has _some_ library linked twice unless I manually remove it (as shown in that bug report)

weatherhead99:
to be clear, this is because in my own CMakeLists.txt (having used DISABLE_AUTOLINK) I then _explicitly_ link in grpc++ to the target.  If conan does that as well I think we end up linking it twice... this actually isn't a problem in my particular build as I'm using it as a shared library, but if you used it as static you end up with duplicated code...

THen again, using it as static I'm not sure triggers the original issue anyway...

weatherhead99:
a second issue (that I also brought up in another thread somewhere) is that grpc build with "secure=False" does _not_ mean you always want to link against grpc++_unsecure does it? Both libraries are built in that case as far as I can see...

Tradias:
>  I think we end up linking it twice...

I don't see how that can happen when you use CMake, it will automatically deduplicate. 

> grpc build with "secure=False" does not mean you always want to link against grpc++_unsecure does it?

As is, the option isn't really applicable, if you link with `grpc++_unsecure` in your application while dependencies are linking with `grpc++`. This option has to be reworked in the grpc recipe, it should force you and every dependency to link with `_unsecure`.

Tradias:
@ conan-team, this pull request is now ready for review. The conflicting pull request has been closed, see https://github.com/conan-io/conan-center-index/pull/18529#issuecomment-1723476670

Tradias:
Contained in https://github.com/conan-io/conan-center-index/pull/22910

PR #19074, opened 2023-08-06
ignition-tools: migrate to Conan v2

valgur:
Also unvendors `backward-cpp` and removes `includedirs`, which the package does not actually export since the built shared library is only meant to be used from a Ruby script.


recipes/ignition-tools/all/conanfile.py
```diff
@@ -1,23 +1,25 @@
...
-    settings = "os", "compiler", "build_type", "arch"
-    options = {"shared": [True, False], "fPIC": [True, False]}
-    default_options = {"shared": False, "fPIC": True}
-    generators = "cmake", "cmake_find_package_multi"
-    exports_sources = "CMakeLists.txt", "patches/**"
-    _cmake = None
+    homepage = "https://ignitionrobotics.org/libs/tools"
+    topics = ("ignition", "robotics", "tools", "gazebo")
+
+    package_type = "application"
```
uilianries, line 21:
```suggestion
    package_type = "library"
```
This package contains a library. https://docs.conan.io/2/reference/conanfile/attributes.html#package-type

recipes/ignition-tools/all/conanfile.py
```diff
@@ -27,79 +29,104 @@ def _minimum_cpp_standard(self):
...
-        if self.settings.os == "Windows":
-            del self.options.fPIC
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
-    def configure(self):
-        if self.options.shared:
-            del self.options.fPIC
+    def requirements(self):
+        self.requires("backward-cpp/1.6")
```
uilianries, line 42: Why does `backward-cpp` is required only now? Its backward library in ignition-tools is not the same from backward-cpp project. https://github.com/gazebosim/gz-tools/tree/ignition-tools_1.4.0/src/Backward#installation

recipes/ignition-tools/all/conanfile.py
```diff
@@ -27,79 +29,104 @@ def _minimum_cpp_standard(self):
...
-        self.cpp_info.components["libignition-tools"].names["pkg_config"] = "ignition-tools{}".format(version_major)
+        self.cpp_info.includedirs = []
+
+        version_major = Version(self.version).major
+        pkg_name = f"ignition-tools{version_major}"
+        self.cpp_info.set_property("cmake_file_name", pkg_name)
+        self.cpp_info.set_property("cmake_target_name", f"{pkg_name}::{pkg_name}")
+
+        component = self.cpp_info.components["libignition-tools"]
+        component.includedirs = []
```
uilianries, line 122: PLease, revert it, the package backward library provided by ignition-tool is not the `backward-cpp` package.

recipes/ignition-tools/all/conanfile.py
valgur: It does build a library, but only as a shared library to be used dynamically in the exported `ign.rb` Ruby script. No headers are installed by the project's CMakeLists.txt.

recipes/ignition-tools/all/conanfile.py
valgur: Are you sure? It even includes a `revision` file in the directory, which links to https://github.com/bombela/backward-cpp/tree/3bb9240cb15459768adb3e7d963a20e1523a6294, which is the v1.6 tag. The backward.hpp file is identical. Since the backward-cpp library is used as header-only here, I would only add `libs=False` to the `requires()` call.

IMO it is preferable to unvendor unmodified external dependencies if possible, since it provides the consumer the option to override the version and it avoids potential silent conflicts if the dependency gets added somewhere else in the graph as well.

recipes/ignition-tools/all/conanfile.py
uilianries: Yes, I read your PR description, but still, you provide a library that affects the package ID.

recipes/ignition-tools/all/conanfile.py
uilianries: If it's identical, so is fine using the external dependency 👍

recipes/ignition-tools/all/conanfile.py
valgur: Ok, thanks, I was not aware of that difference. Would the same logic not apply to any other typical application as well, though, if they package .so/.dll files to be only used by the packaged executables and not as stand-alone libraries?

recipes/ignition-tools/all/test_package/test_package.cpp
```diff
@@ -1,7 +1,5 @@
 #include <iostream>
 
-
-int main(int argc, char **argv)
-{
+int main() {
```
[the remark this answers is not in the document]
valgur: Ugh... I looked into what this package actually should provide and this recipe is quite incorrect.

The built library is meant to be a pure runtime library for the `ign` Ruby script packaged by this recipe. CMake and pkg-config targets should not be exported by the recipe, since they are effectively empty. Here's the .pc file installed by the project:
```
# This packages provides no headers or libraries, so those fields are omitted
Name: Ignition tools
Description: A set of tools classes for robot applications
Version: 1.4.0
```

I updated the recipe accordingly.

valgur:
Don't close.

valgur:
.

PR #19081, opened 2023-08-07
libraw: update libjpeg-turbo

toge:
Specify library name and version:  **libraw/***

---


recipes/libraw/all/conanfile.py
```diff
@@ -66,6 +67,8 @@ def source(self):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        if Version(self.version) <= "0.19.5":
+            tc.variables["CMAKE_CXX_STANDARD"] = 11
```
jcar87, line 71: in Conan 2.0 I think this effectively nullifies all the effects of the `compiler.cppstd` setting. 

If this recipe is not compatible with C++17 or higher, I believe we should be using: https://docs.conan.io/2/reference/tools/build.html#conan-tools-build-check-max-cppstd

Setting it to C++11 here when the package ID has c++17 (or higher) can be problematic if the `register` keyword is in a public header.

jcar87 (changes requested):

recipes/libraw/all/conanfile.py
toge: @jcar87 
Thank your for your review!
I updated this PR to use check_max_cppstd.

Please review it again.

toge: fixed

SSE4:
@jcar87 can you unblock? your comment was addressed long time ago already

toge:
@jcar87 
Please review again or unblock it.

recipes/libraw/all/test_package/CMakeLists.txt
```diff
@@ -3,6 +3,11 @@ project(test_package LANGUAGES CXX)
 
 find_package(libraw REQUIRED CONFIG)
 
+# libraw/0.19.5 uses std::auto_ptr 
+if (libraw_VERSION VERSION_EQUAL "0.19.5")
+    set(CMAKE_CXX_STANDARD 11)
+endif()
```
SpaceIm, line 9: auto_ptr has been removed in C++17, so C++14 is valid (please fix your check_max_cppstd)

SpaceIm: But honestly, you should remove this version.

recipes/libraw/all/conanfile.py
```diff
@@ -60,14 +62,20 @@ def requirements(self):
...
     def source(self):
-       get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
         tc = CMakeToolchain(self)
         tc.variables["RAW_LIB_VERSION_STRING"] = self.version
         tc.variables["LIBRAW_SRC_DIR"] = self.source_folder.replace("\\", "/")
+        if conan_version.major == 1:
+            tc.variables["CMAKE_CXX_STANDARD"] = 11
```
uilianries, line 78:
```suggestion
        if conan_version.major == 1:
            tc.variables["CMAKE_CXX_STANDARD"] = self.settings.get_safe("compiler.cppstd", 11)
```

In case a version does not support C++17/20, for instances, let's not try workaround, but just avoid building with check_max_cppstd, like: https://github.com/conan-io/conan-center-index/blob/master/recipes/ccfits/all/conanfile.py#L52

uilianries (commented):
Thank for your PR!

SpaceIm:
This PR can be closed, libjpeg-turbo version has been updated in https://github.com/conan-io/conan-center-index/pull/20935

PR #17272, opened 2023-04-27
elfutils: Revise the recipe to support Conan V2 and add the latest version 0.189

0xFireWolf:
Specify library name and version:  **elfutils/0.180, 0.186, 0.189**

**This PR contains the following changes:**
- Revise the recipe to make it compatible with both Conan v1 and v2.
- Upgrade all build requirements and dependencies to the latest version.
- Implement universal patches that are necessary to compile the library in Python.
- Remove version-specific patch files.
- Prevent users from building the library with GCC or Clang on macOS.
    - elfutils does not support macOS.

Note that this PR is needed to fix another build error (https://github.com/conan-io/conan-center-index/issues/16799) of the package `backward-cpp`.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #19095, opened 2023-08-08
backward-cpp: add header_only option

valgur:
The library is nearly header-only anyway and is being used as such in #19074, so it made sense to perhaps add the option to the recipe.

Also removed the redundant `transitive_libs=True` params and made Linux OS detection more consistent.

### DIFF
--- a/recipes/backward-cpp/all/conanfile.py
+++ b/recipes/backward-cpp/all/conanfile.py
@@ -3,6 +3,7 @@ from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
+from conan.tools.layout import basic_layout
 from conan.tools.scm import Version
 import os
 
@@ -20,12 +21,14 @@ class BackwardCppConan(ConanFile):
     package_type = "library"
     settings = "os", "arch", "compiler", "build_type"
     options = {
+        "header_only": [True, False],
         "shared": [True, False],
         "fPIC": [True, False],
         "stack_walking": ["unwind", "libunwind", "backtrace"],
         "stack_details": ["dw", "bfd", "dwarf", "backtrace_symbol"],
     }
     default_options = {
+        "header_only": False,
         "shared": False,
         "fPIC": True,
         "stack_walking": "unwind",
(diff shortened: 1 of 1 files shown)


recipes/backward-cpp/all/conanfile.py
```diff
@@ -57,23 +60,33 @@ def config_options(self):
...
+            self.info.clear()
 
     def requirements(self):
-        if self.settings.os in ["Linux", "Android"]:
+        if self.settings.os in ["Linux", "FreeBSD", "Android"]:
             if self._has_stack_walking("libunwind"):
                 self.requires("libunwind/1.6.2", transitive_headers=True)
             if self._has_stack_details("dwarf"):
-                self.requires("libdwarf/20191104", transitive_headers=True, transitive_libs=True)
+                self.requires("libdwarf/20191104", transitive_headers=True)
```
RubenRBS, line 84: After checking locally, this seems to be needed for shared=True, so I wonder if the `dw` and `bfd` options are also neded. I can't test `dw` because `elfutils` is not Conan 2 ready yet (Waiting on https://github.com/conan-io/conan-center-index/pull/17272), and for `bfd`, I get a missing header compilation error:
```
In file included from /root/.conan2/p/b/backw4b891165fe452/b/src/backward.cpp:36:
/root/.conan2/p/b/backw4b891165fe452/b/src/backward.hpp:229:10: fatal error: bfd.h: No such file or directory
  229 | #include <bfd.h>
      |          ^~~~~~~
```

RubenRBS (commented):
Thanks a lot. I only have 1 concern about removing the `transitive_libs` trait for `shared=True` in `dwarf` mode

recipes/backward-cpp/all/conanfile.py
valgur: Interesting. As far as I understand it, `transitive_libs=True` is set by default anyway as long as `libs=True` is set (also by default). Is there something I'm missing? I would love to know more, if that's the case.

RubenRBS: @valgur Sorry I missed this conversation: In Conan 1, it all linked in the end due to how Conan worked back then afaik, but yeah Conan 2 does not transitively make the libs available unless told to otherwise (Except for the case of `package_type=header-library` which has both transitive traits set to true)

valgur:
@RubenRBS Another ping, just in case.

RubenRBS (approved):

RubenRBS:
Thanks!

jwillikers (approved):

PR #19114, opened 2023-08-09
fontconfig: minimal change to ensure most recent rrev

jcar87:
The recent changes in `fontconfig` w.r.t to the libuuid dependency were introduced and then subsequently reversed - causing an interesting case where the recipe revision for the most recent commit still has the "earlier" timestamp for the same changes.

This PR touches the recipe minimally to ensure that a new recipe revision with a fresh timestamp is published.


RubenRBS (approved):

samuel-emrys:
Oh good, thanks for this. I was wondering why `tk` still had `libuuid` in the graph

PR #19115, opened 2023-08-09
Add new recipe: libmpdclient/2.20

SpaceIm:

---

### DIFF
--- a/recipes/libmpdclient/all/conanfile.py
+++ b/recipes/libmpdclient/all/conanfile.py
@@ -78,6 +78,7 @@ class LibmpdclientConan(ConanFile):
         meson = Meson(self)
         meson.install()
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
         fix_apple_shared_install_name(self)
         fix_msvc_libname(self)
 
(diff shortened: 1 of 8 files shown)


recipes/libmpdclient/all/conanfile.py
```diff
@@ -0,0 +1,111 @@
...
+        fix_apple_shared_install_name(self)
+        fix_msvc_libname(self)
+
+    def package_info(self):
+        self.cpp_info.set_property("pkg_config_name", "libmpdclient")
+        self.cpp_info.libs = ["mpdclient"]
+        if self.settings.os == "Windows":
+            self.cpp_info.system_libs.append("ws2_32")
+
+def fix_msvc_libname(conanfile, remove_lib_prefix=True):
```
RubenRBS, line 97: You might have had this asked before but I couldnt find anything in the few PRs I checked:

Is there any strong reason besides conventions for you to prefer doing it this way instead of conditionally setting `self.cpp_info.libs`? I'm not against this way either, but it's true that I feel it would leave the code simpler to understand and be more "true", as it leaves the filename unchanged to what might otherwise be found in the wild

RubenRBS (approved):
Thanks a lot @SpaceIm, really appreciated! This looks good so approving, but I have one question regarding alternatives to `fix_msvc_libname` (But Im not opposed to using it either) :)

SSE4 (approved):

recipes/libmpdclient/all/conanfile.py
SpaceIm: It would be another solution yes. Currently in all meson recipes, lib file is renamed for msvc, so it's consistent. Actually I was hoping that conan would provide this helper function (https://github.com/conan-io/conan/issues/11866) in the near future.

Conditional lib names in package_info() tend to be fragile (many issues in CCI recipes with clang variant on Windows), I think the current workaround is more robust.

PR #19135, opened 2023-08-10
libe57format: add new versions 3.0.[12]

marxin:
The added patch disables LTO as it's not feasible for static libraries where one needs to have exactly the same GCC version (LTO is a bytecode).

Disable E57_BUILD_TEST tests as they want to clone a separate git repository.


marxin:
The package tests fail on `MacOS` for unknown reason:

```
libe57format/3.0.2 (test package): Running test()
libe57format/3.0.2 (test package): RUN: ./test_package
libc++abi: terminating with uncaught exception of type e57::E57Exception: E57 exception
/bin/sh: line 1: 68134 Abort trap: 6           ./test_package
```

Is there a way I can get it merged even w/o the MacOS support?

Issue #16639, opened 2023-03-21
[package] qt/5.15.8: Qt webengine link error: unknown option: -rpath-link=...


nmgwddj:
### Description

Build Qt with WebEngine under M1.

### Package and Environment Details

* Package Name/Version: **qt/5.15.8**
* Operating System+version: **Macos 13.2.1**
* Compiler+version: **Apple clang 14.0**
* Conan version: **conan 1.59.0**
* Python version: **Python 3.7.x**

### Conan profile

Configuration for profile default:

[settings]
os=Macos
os.version=10.14
os_build=Macos
arch=armv8
arch_build=armv8
compiler=apple-clang
compiler.version=14.0
compiler.libcxx=libc++
compiler.cppstd=14
build_type=Release
[options]
[conf]
[build_requires]
[env]
OBJCFLAGS=-Wno-error=deprecated-declarations -fvisibility=hidden -fvisibility-inlines-hidden

### Steps to reproduce

add new patch files to qt 5.15.8 patchs folder:

Patch1: https://github.com/qt/qtwebengine-chromium/commit/019a6b9282e7e162ca21832a4724a7e27dfea0b1
Patch2: https://github.com/qt/qtwebengine-chromium/commit/2082566249c5a096f3e1ef3706d1a87cb46ac970
Patch3: https://github.com/qt/qtwebengine-chromium/commit/376934168123aaa977c0e93dbc6f0afb32125eee
Patch4: https://github.com/qt/qtwebengine-chromium/commit/08b4e141cc17a9e6105380da74041069b44dccf5

These fixes can ensure smooth compilation of qt 5.15.8 webengine. Than add new patch to conandata.yml

```
  - base_path: qt5/qtwebengine/src/3rdparty
    patch_file: patches/fix-arm64-skia-build-for-macos.patch
  - base_path: qt5/qtwebengine/src/3rdparty
    patch_file: patches/fix-building-with-xcode-14.1.patch
  - base_path: qt5/qtwebengine/src/3rdparty
    patch_file: patches/fix-clang-set-but-unused-variable-warning.patch
  - base_path: qt5/qtwebengine/src/3rdparty
    patch_file: patches/on-arm64-hosts-set-host_cpu-to-arm64-not-arm.patch
```

In the receipt/qt/5.x.x, execute following command:

```
conan create . qt/5.15.8 \
    -s os=Macos \
    -s arch=armv8 \
    -s compiler.cppstd=14 \
    -s build_type=Release \
[...]

PR #19137, opened 2023-08-10
clang does not support -rpath-link

boussaffawalid:
Specify library name and version:  **qt/5.15**

Compaling qt with qtwebengine enabled on m1 mac with clang 13 I got an error
```
linking ../../lib/libQt5WebEngineCore.5.15.10.dylib
ld: unknown option: -rpath-link=
```

it seems like Clang does not support `-rpath-link`

This is the same issue as https://github.com/conan-io/conan-center-index/issues/16639

---

### DIFF
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -124,7 +124,6 @@ class QtConan(ConanFile):
     no_copy_source = True
     short_paths = True
 
-    @property
     def _is_clang(self):
         return str(self.settings.compiler) in ["clang", "apple-clang"]
 
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -125,6 +125,10 @@ class QtConan(ConanFile):
     short_paths = True
 
     @property
+    def _is_clang(self):
+        return str(self.settings.compiler) in ["clang", "apple-clang"]
+
+    @property
     def _settings_build(self):
         return getattr(self, "settings_build", self.settings)
 
@@ -737,7 +741,7 @@ class QtConan(ConanFile):
 
         libdirs = [l for dependency in self.dependencies.host.values() for l in dependency.cpp_info.aggregated_components().libdirs]
         args.append("QMAKE_LIBDIR+=\"%s\"" % " ".join(libdirs))
-        if not is_msvc(self):
+        if not is_msvc(self) and not self._is_clang():
             args.append("QMAKE_RPATHLINKDIR+=\"%s\"" % ":".join(libdirs))
 
         if "libmysqlclient" in [d.ref.name for d in self.dependencies.direct_host.values()]:


recipes/qt/5.x.x/conanfile.py
```diff
@@ -737,7 +740,7 @@ def build(self):
 
         libdirs = [l for dependency in self.dependencies.host.values() for l in dependency.cpp_info.aggregated_components().libdirs]
         args.append("QMAKE_LIBDIR+=\"%s\"" % " ".join(libdirs))
-        if not is_msvc(self):
+        if not is_msvc(self) and not self._is_clang():
```
jwillikers, line 743: Is this a problem on newer versions of Clang or only on version 13? If it's only an issue for 13, I think we should only do this for that version of Clang. I think we want to keep RPATH enabled otherwise.

jwillikers (changes requested):

recipes/qt/5.x.x/conanfile.py
boussaffawalid: well, I cannot really tell, I can confirm that the issue exist for clang 13 and 14 on Macos m1

jcar87: I beleive this is an issue with the linker and not the compiler. The linker in Apple platforms doesn't have `-rpath-link`, but on Linux (even with Clang), this should work.


Perhaps this conditional can be changed to `if not is_msvc(self) and not is_apple_os(self):` - or something along those lines.

boussaffawalid: probably `if not is_msvc(self) and not (is_apple_os(self) and self._is_clang()) :` is more accurate, assuming `-rpath-link` is available for gcc on macos

jcar87:
I'm investigating this issue and I'm a bit puzzled as to why it happens in the first place.

`QMAKE_RPATHLINKDIR` should _only_ take effect on platforms where it is supported, and looking at the source code, it should do nothing on platforms where it isn't.

https://github.com/qt/qtbase/blob/v5.15.14-lts-lgpl/qmake/generators/unix/unixmake.cpp#L149-L155

From the above, the `-rpath-link` flags would only be passed if `QMAKE_LFLAGS_RPATHLINK` is defined.

`QMAKE_LFLAGS_RPATHLINK` is only defined for some mkspecs, `gcc-base-unix`, `qcc-base-qnx` and `linux-icc` 

The mkspec for macos is `macx-clang`, which the recipe is passing here: 
https://github.com/conan-io/conan-center-index/blob/master/recipes/qt/5.x.x/conanfile.py#L527

The looking at the  `macx-clang` mkspec from https://github.com/qt/qtbase/blob/v5.15.14-lts-lgpl/mkspecs/macx-clang/qmake.conf#L27-L30 and the files it includes... I cannot see `QMAKE_LFLAGS_RPATHLINK` defined anywhere for this configuration - it _really_ should prevent `-rpath-link` from being passed altogether. 

I'm building this locally (takes a while), but will double check. Perhaps since this is qtwebengine we're talking about, there's additional logic somewhere.

jcar87:
Fixed in https://github.com/conan-io/conan-center-index/pull/24826

Issue #19125, opened 2023-08-09
[request] influxdb-cxx/0.7.1


steinerthomas:
### Package Name/Version

influxdb-cxx/0.7.1

### Webpage

https://github.com/offa/influxdb-cxx

### Source code

https://github.com/offa/influxdb-cxx

### Description of the library/tool

InfluxDB C++ client library.

Opened for discussion, see also:
https://github.com/offa/influxdb-cxx/issues/53
@offa You want to bring it up here at conan-center-index?

offa:
> @offa You want to bring it up here at conan-center-index?

I'm fine with the submission :+1:, thanks.

steinerthomas:
@toge Awesome, thank you for the contribution!

toge:
@steinerthomas  
Thank you for your request!
Please wait a little to merge PR.

PR #19148, opened 2023-08-11
influxdb-cxx: add recipe

toge:
Specify library name and version:  **influxdb-cxx/0.7.1**

Try to close https://github.com/conan-io/conan-center-index/issues/19125.

---

### DIFF
--- a/recipes/influxdb-cxx/all/conanfile.py
+++ b/recipes/influxdb-cxx/all/conanfile.py
@@ -77,6 +77,7 @@ class InfluxdbCxxConan(ConanFile):
         tc.cache_variables["INFLUXCXX_TESTING"] = False
         tc.cache_variables["INFLUXCXX_WITH_BOOST"] = self.options.boost
         if self.options.shared:
+            # See https://github.com/offa/influxdb-cxx/issues/194
             tc.preprocessor_definitions["InfluxDB_EXPORTS"] = 1
         tc.generate()
         deps = CMakeDeps(self)
--- a/recipes/influxdb-cxx/all/conanfile.py
+++ b/recipes/influxdb-cxx/all/conanfile.py
@@ -76,6 +76,8 @@ class InfluxdbCxxConan(ConanFile):
         tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
         tc.cache_variables["INFLUXCXX_TESTING"] = False
         tc.cache_variables["INFLUXCXX_WITH_BOOST"] = self.options.boost
+        if self.options.shared:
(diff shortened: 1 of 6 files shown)


steinerthomas (approved):

recipes/influxdb-cxx/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        # BUILD_SHARED_LIBS is defined explicitly in CMakeLists.txt
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["INFLUXCXX_TESTING"] = False
+        tc.cache_variables["INFLUXCXX_WITH_BOOST"] = self.options.boost
+        if self.options.shared:
+            tc.preprocessor_definitions["InfluxDB_EXPORTS"] = 1
```
jcar87, line 80: This one looks a little unusual - is it not possible to build the shared variant without the consumers defining preprocessor definitions externally?

jcar87 (commented):
Thanks @toge - recipe looks great!! Just one small question

recipes/influxdb-cxx/all/conanfile.py
offa: This shouldn't be necessary as the define is handled by CMakes GenerateExportHeader. However, there are reports of issues on MSVC. I wasn't able to reproduce the issue, but I'm no MSVC dev.

danimtb (approved):

recipes/influxdb-cxx/all/conanfile.py
jcar87: Thanks @toge  - there is indeed a pre-existing issue, I've updated the recipe to reference it

I've had a chance to investigate the root cause and made a comment here:
https://github.com/offa/influxdb-cxx/issues/194#issuecomment-1935661442

recipes/influxdb-cxx/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,16 @@
+#include <InfluxDBFactory.h>
+#include <InfluxDBException.h>
+#include <iostream>
+
+int main()
+{
+    try {
+        auto influxdb = influxdb::InfluxDBFactory::Get("xyz://foobar");
```
jcar87, line 8: I've changed this to _not_ reference localhost - in the odd chance something is actually running.  Test packages in Conan center _must not_ make network requests to any resource whatsover.

jcar87 (approved):

franramirez688 (approved):

PR #19124, opened 2023-08-09
[cgal] add cgal/5.6

lrineau:
Specify library name and version:  **cgal/5.6**

Update to latest CGAL upstream release: https://github.com/CGAL/cgal/releases/tag/v5.6

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #18720, opened 2023-07-20
flint: migrate to Conan v2

valgur:
TODO

- [x] `Could NOT find PThreads (missing: PThreads_LIBRARIES)`
- [x] `ERROR: Missing prebuilt package for 'mpfr/4.1.0'`

PR #19150, opened 2023-08-11
[mpfr] Fix when building shared library on Linux

uilianries:
Specify library name and version:  **mpfr/4.2.0**

When building with Conan 2.0, and `*/*:shared=True`, the `mpfr` executes a validation with autotools to check if is possible to use `gmp`. However, as `VirtualRunEnv` was not configured, the `LD_LIBRARY_PATH` was not present, resulting in failure when trying to run its test:

```
configure:4522: gcc -o conftest  -m64 -O3  -I/home/conan/.conan2/p/gmp0bfdd57764b0f/p/include -DNDEBUG -I/home/conan/.conan2/p/gmp0bfdd57764b0f/p/include  -L/home/conan/.conan2/p/gmp0bfdd57764b0f/p/lib -m64 -L/home/conan/.conan2/p/gmp0bfdd57764b0f/p/lib conftest.c  -lgmpx
x -lgmp -lm >&5
configure:4526: $? = 0
configure:4533: ./conftest
./conftest: error while loading shared libraries: libgmpxx.so.4: cannot open shared object file: No such file or directory
configure:4537: $? = 127
configure:4544: error: in `/home/conan/.conan2/p/b/mpfrbaa826e03f170/b/build-release':
configure:4546: error: cannot run C compiled programs.
```

This PR fixes the situation by adding `VirtualRunEnv`.

Required by #18720 and #19124

---

### DIFF
--- a/recipes/mpfr/all/conanfile.py
+++ b/recipes/mpfr/all/conanfile.py
@@ -5,6 +5,8 @@ from conan.tools.cmake import CMakeToolchain, CMakeDeps, CMake, cmake_layout
 from conan.tools.gnu import Autotools, AutotoolsToolchain, AutotoolsDeps
 from conan.tools.microsoft import is_msvc, check_min_vs, unix_path
 from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.env import VirtualRunEnv
+from conan.tools.build import cross_building
 from conan.errors import ConanException
 import os
 import re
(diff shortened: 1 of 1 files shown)


danimtb (approved):

jcar87 (approved):
We need to find a way to solve this that doesn't require exposing the run environment at build time!

nextsilicon-itay-bookstein:
Writing it out here for posterity: apparently this trick causes libgmp **from the conan cache** to load into the compiler process itself! I only found this out as I was doing an address-sanitized build (with shared-libasan) and this caused weird errors to pop up when the configure script tried interrogating the compiler itself (missing symbols that normally would be retrieved from libasan DSO).

PR #19181, opened 2023-08-14
Added kdbindings/1.0.3

bog-dan-ro:
**kdbindings/1.0.3**

KDBindings enables reactive programming & data binding in C++ From plain C++ you get:
- Signals + Slots.
- Properties templated on the contained type.
- Property bindings allowing reactive code to be written without having to do all the low-level, error prone plumbing by hand. Lazy evaluation of property bindings.
- No more broken bindings.
- Totally stand-alone "header-only" library.

I'm working for KDAB (the company behind this library). I have their support for adding and maintaining this library to conan center index.

---

### DIFF
--- a/recipes/kdbindings/all/conanfile.py
+++ b/recipes/kdbindings/all/conanfile.py
@@ -1,0 +1,54 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import get, copy
+from conan.tools.scm import Version
+import os
+
+class KDBindingsConan(ConanFile):
+    name = "kdbindings"
+    license = "MIT"
+    topics = ("c++17", "reactive", "kdab", "header-only")
+    description = "Reactive programming & data binding in C++"
+    homepage = "https://github.com/KDAB/KDBindings"
+    url = "https://github.com/conan-io/conan-center-index"
+    package_type = "header-library"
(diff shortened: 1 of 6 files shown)


RubenRBS (approved):

RubenRBS:
Thanks a lot! We appreciate it, the recipe looks good :)

czoido (approved):

RubenRBS:
@bog-dan-ro nice to hear :)

There's actually no requirement for header-only libraries to define _any_ setting. You're in fact encouraged not to define the attribute at all if the `generate()` method does not use its values/invokes any compiler, etc.
But if you do need them, adding them all usually avoids issues with generators misbehaving

bog-dan-ro:
@RubenRBS I added them after cci hook warned that I didn't have any `settings` attribute...

recipes/kdbindings/all/conanfile.py
```diff
@@ -0,0 +1,54 @@
...
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True, destination=self.source_folder)
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "9",
+            "Visual Studio": "15.7",
+            "clang": "7",
+            "apple-clang": "11",
```
SpaceIm, line 31: `"msvc": "191"` should be added to this dictionary

PR #19187, opened 2023-08-14
[highs] Enable shared build on Windows

hedtke:
Specify library name and version:  **highs**

Enable shared build on Windows and drop v1 test package

---


recipes/highs/all/conanfile.py
```diff
@@ -1,14 +1,11 @@
 from conan import ConanFile
 from conan.tools.apple import fix_apple_shared_install_name
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
-from conan.errors import ConanInvalidConfiguration
 from conan.tools.files import apply_conandata_patches, collect_libs, copy, export_conandata_patches, get
 from conan.tools.microsoft import is_msvc
 from conan.tools.scm import Version
 from os.path import join
 
-required_conan_version = ">=1.53.0"
```
RubenRBS, line 10: We want to keep this, it provides a nicer error for old versions :)

recipes/highs/all/conanfile.py
```diff
@@ -27,10 +24,6 @@ class HiGHSConan(ConanFile):
         "fPIC": True,
     }
 
-    def validate(self):
```
RubenRBS, line 30: So I take that there was no issue with it and ir always worked? Oh

RubenRBS (approved):
Weird that it works, I wonder why it was disabled in the first case then 🤔

recipes/highs/all/conanfile.py
hedtke: Added again

recipes/highs/all/conanfile.py
hedtke: I forgot to copy the .lib file on Windows in the initial version of the recipe. I didn't know how to fix the problem of missing symbols. My only solution was to deactivate shared libs on Windows. I know better now.

RubenRBS (approved):
Thanks! Sorry for the delay, I missed the notification about the pending review after the new commit :)

danimtb (approved):

PR #19200, opened 2023-08-15
Follow the convention and also provide etl::etl alias.

jputcu:
Add etl::etl library alias to follow the convention.
---

### DIFF
--- a/recipes/etl/all/conanfile.py
+++ b/recipes/etl/all/conanfile.py
@@ -60,7 +60,6 @@ class EmbeddedTemplateLibraryConan(ConanFile):
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "etl")
-        self.cpp_info.set_property("cmake_target_name", "etl")
         self.cpp_info.set_property("cmake_target_name", "etl::etl")
 
         self.cpp_info.bindirs = []
--- a/recipes/etl/all/conanfile.py
+++ b/recipes/etl/all/conanfile.py
@@ -61,6 +61,7 @@ class EmbeddedTemplateLibraryConan(ConanFile):
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "etl")
         self.cpp_info.set_property("cmake_target_name", "etl")
+        self.cpp_info.set_property("cmake_target_name", "etl::etl")
 
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
--- a/recipes/etl/all/conanfile.py
+++ b/recipes/etl/all/conanfile.py
@@ -61,6 +61,7 @@ class EmbeddedTemplateLibraryConan(ConanFile):
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "etl")
         self.cpp_info.set_property("cmake_target_name", "etl::etl")
+        self.cpp_info.set_property("cmake_target_aliases", ["etl"]) # legacy target before 20.27.2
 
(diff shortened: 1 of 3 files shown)


recipes/etl/all/conanfile.py
```diff
@@ -60,7 +60,7 @@ def _module_file_rel_path(self):
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "etl")
-        self.cpp_info.set_property("cmake_target_name", "etl")
+        self.cpp_info.set_property("cmake_target_name", "etl::etl")
```
SpaceIm, line 63:
```suggestion
        self.cpp_info.set_property("cmake_target_name", "etl::etl")
        self.cpp_info.set_property("cmake_target_aliases", ["etl"]) # legacy target before 20.27.2
```

It might be a good idea to still provide legacy imported target (which was `etl` before 20.27.2, see https://github.com/ETLCPP/etl/pull/523) to not break users.

jputcu: Thanks, I was thinking the same with my first attempt: calling set_property twice, which wasn't correct.

recipes/etl/all/conanfile.py
```diff
@@ -60,7 +60,8 @@ def _module_file_rel_path(self):
...
```
czoido, line 63: Hi @jputcu, thanks a lot for the contribution. The set of both properties can be removed as the default for those properties is the same as set by the `set_property` method. So better remove to simplify the recipe.

SpaceIm: I don't recommend to remove, because it's the only way for maintainers to track the difference between a library without any official config file, and library with official config file which turns out to have the same name that recipe name.

jcar87: I see, thanks for the explanation.  We understand the motivation but this is a non-obvious  convention. A recipe having a `cmake_target_name` does not really signal that the CMake name is correct, it just signals that it has an explicit name. To some of us, when the name is lowercase and matches the name of the recipe, it makes us wonder why it is needed in the first place. 

> Moreover, it enforces the name and avoid to break when a recipe is renamed.

I think this is a very good point and one that would indeed avoid a breage. But that's also the purpose of the test_package, which explicitly encodes the name of the expected target in the CMakeLists. Changing the name of a recipe in a way that causes the name of the generated cmake targets to change, would also cause the `test_package` to fail: since the name of the expected target is explicitly encoded in the test_package. It would far more useful to me as a maintainer if this was actually documented in the test packages CMakeLists themselves, and failing that, the recipe. That should be strong enough a signal if a recipe is renamed (which I dont think happens often enough, but still good to protect us from that).

We're tentatively okay with being explicit in recipes but this is not a convention we will enforce in any way until this is reviewed by the team - so while we will not remove this from any recipe, we ask that we don't burden contributors during code reviews with adding this to recipes that don't need them.

jputcu: Maybe a small related note, on the conan website, I think the those targets are used in the documentation: "use-it->targets". As a library user I find that interesting, but without the `cmake_target_name` I don't think we get that info.

czoido (approved):

recipes/etl/all/conanfile.py
jcar87: Thanks @jputcu, that's a very good observation! in the website backend we have implemented logic to derive the target even when `cmake_target_name` is not defined explicit, so this info should be visible for libraries regardless.

Issue #19204, opened 2023-08-15
[package] libmysqlclient/8.1.0: Error when building library on RHEL8


carloshimura:
### Description

This new package version can't build on RHEL8 (Alma linux distro)

Cmake error:
```
CMake Error at CMakeLists.txt:2024 (ADD_SUBDIRECTORY):
  ADD_SUBDIRECTORY given source "extra/tirpc" which is not an existing
  directory.
```
This seems to be due to the fact that extra directory is fully removed in the conanfile.py

### Package and Environment Details

* Package Name/Version: **zlib/1.2.8**
* Operating System+version: **Alma Linux 8.8**
* Compiler+version: **GCC 12**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.60.0**
* Python version: **Python 3.11.2**

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=12
compiler.libcxx=libstdc++11
build_type=Release
compiler.cppstd=17
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install -r conancenter libmysqlclient/8.1.0 --build=missing

### Logs

<details><summary>Click to expand log</summary>

```
...
-- Conan: Component target declared 'OpenSSL::Crypto'
-- Conan: Component target declared 'OpenSSL::SSL'
-- Conan: Target declared 'openssl::openssl'
-- Conan: Including build module from '/ccache/conandata/openssl/3.1.1/_/_/package/772675ace4978e609416253f5f352cadffbfbb53/lib/cmake/conan-official-openssl-variables.cmake'
-- Found OpenSSL: 3.1.1 (found version "3.1.1") 
-- SSL_LIBRARIES = OpenSSL::SSL;OpenSSL::Crypto;dl
-- KERBEROS path is none, disabling kerberos support.
-- HAVE_KRB5_KRB5_H  
CMake Warning at cmake/sasl.cmake:272 (MESSAGE):
  Could not find SASL
Call Stack (most recent call first):
  CMakeLists.txt:1814 (MYSQL_CHECK_SASL)


-- Looking for lber.h
-- Looking for lber.h - not found
-- Looking for ldap.h
-- Looking for ldap.h - not found
CMake Warning at cmake/ldap.cmake:158 (MESSAGE):
  Could not find LDAP
Call Stack (most recent call first):
[...]

PR #19209, opened 2023-08-15
libmysqlclient: fix compilation error on RHEL

toge:
Specify library name and version:  **libmysqlclient/***

Try to fix https://github.com/conan-io/conan-center-index/issues/19204.

---


RubenRBS (approved):
I don't think we're fully handling upstream's options by looking at the cmakelist, maybe this recipe needs a few enhancements in the future, but looks like this works without breaking anything that wasn't broken before, so nice! Thanks for your contribution :)

I wonder if we could just keep the needed library and remove the rest, it's wasted space otherwise

Issue #19219, opened 2023-08-16
[opencv] 4.5.5: with_jpeg2000=openjpeg fails


marxin:
### Description

The package cannot be build with `openjpeg` Conan dependency.

### Package and Environment Details

OpenCV 4.5.5
openSUSE Tumbleweed
GCC 13.2
Conan 1.6.1

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=13
compiler.libcxx=libstdc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

`$ conan create . opencv/4.5.5@ --build=missing -o opencv:with_jpeg2000=openjpeg -o opencv:with_ffmpeg=False`

### Logs

```
...
-- Found system OpenJPEG:  (found version "")
...
In file included from /home/marxin/.conan/data/opencv/4.5.5/_/_/build/d17281ef8f37f975e5c92a82c7d4acfcec9ee902/src/modules/imgcodecs/src/grfmt_jpeg2000_openjpeg.cpp:10:
/home/marxin/.conan/data/opencv/4.5.5/_/_/build/d17281ef8f37f975e5c92a82c7d4acfcec9ee902/src/modules/imgcodecs/src/grfmt_jpeg2000_openjpeg.hpp:13:10: fatal error: openjpeg.h: No such file or directory
   13 | #include <openjpeg.h>
      |          ^~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [modules/imgcodecs/CMakeFiles/opencv_imgcodecs.dir/build.make:216: modules/imgcodecs/CMakeFiles/opencv_imgcodecs.dir/src/grfmt_jpeg2000_openjpeg.cpp.o] Error 1
gmake[2]: *** Waiting for unfinished jobs....
In file included from /home/marxin/.conan/data/opencv/4.5.5/_/_/build/d17281ef8f37f975e5c92a82c7d4acfcec9ee902/src/modules/imgcodecs/src/grfmts.hpp:54,
                 from /home/marxin/.conan/data/opencv/4.5.5/_/_/build/d17281ef8f37f975e5c92a82c7d4acfcec9ee902/src/modules/imgcodecs/src/loadsave.cpp:47:
[...]

PR #19222, opened 2023-08-16
opencv 4.5.x: add openjpeg="openjpeg_embedded" option

marxin:
This is a workaround for #19219.

### DIFF
--- a/recipes/opencv/4.x/conanfile.py
+++ b/recipes/opencv/4.x/conanfile.py
@@ -127,7 +127,7 @@ class OpenCVConan(ConanFile):
         "with_jpeg": [False, "libjpeg", "libjpeg-turbo", "mozjpeg"],
         "with_png": [True, False],
         "with_tiff": [True, False],
-        "with_jpeg2000": [False, "jasper", "openjpeg"],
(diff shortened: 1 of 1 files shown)


SpaceIm:
I don't think it's the correct way to handle issue described in https://github.com/conan-io/conan-center-index/issues/19219. If external openjpeg header is not found, it means that something has to be patched in upstream CMakeLists or maybe defined differently in CMakeToolchain.

SpaceIm:
It's likely a casing issue: https://github.com/opencv/opencv/blob/4.5.5/modules/imgcodecs/CMakeLists.txt#L44-L45

For openjpeg recipe, `CMakeDeps` define `OpenJPEG_INCLUDE_DIRS` & `OpenJPEG_LIBRARIES` while OpenCV CMakeLists expects `OPENJPEG_INCLUDE_DIRS` & `OPENJPEG_LIBRARIES` (OpenCV build files are horrible because they almost never rely on CMake targets). So it's either something missing in openjpeg recipe to also define these variables, or a small patch to apply in OpenCV build file `modules/imgcodecs/CMakeLists.txt`.

Issue #19211, opened 2023-08-15
[package] rtmidi/6.0.0: sha256 signature failed for 'rtmidi-6.0.0.tar.gz' file


SpaceIm:
### Description

sha256 of 6.0.0 is incorrect (I guess it has changed)

### Package and Environment Details

* Package Name/Version: **rtmidi/6.0.0**
* Operating System+version: **macOS Ventura**
* Compiler+version: **apple-clang 14**
* Conan version: **conan 2.0.9**
* Python version: **Python 3.11.4**

### Conan profile

[settings]
os=Macos
arch=x86_64
compiler=apple-clang
compiler.version=14
compiler.libcxx=libc++
compiler.cppstd=20
build_type=Release

### Steps to reproduce

```
conan install --requires rtmidi/6.0.0 -b rtmidi
```

### Logs

<details><summary>Click to expand log</summary>

```
ERROR: rtmidi/6.0.0: Error in source() method, line 55
	get(self, **self.conan_data["sources"][self.version], strip_root=True)
	ConanException: sha256 signature failed for 'rtmidi-6.0.0.tar.gz' file.
 Provided signature: 3336248e8c1f054ea5e51a4449558490dca51edd324fcde0eea27df33b80a9ed
 Computed signature: 5960ccf64b42c23400720ccc880e2f205677ce9457f747ef758b598acd64db5b
```

</details>

PR #19233, opened 2023-08-16
rtmidi: fix sha256 of 6.0.0

SpaceIm:
closes https://github.com/conan-io/conan-center-index/issues/19211


jcar87 (changes requested):
Is there any indication as to why the sha changed, whether the contents are the same as the previously released package?

RubenRBS:
Also note that as of the 10th of Aug (public announcement is still pending), we have enabled source backups in CCI for all sources downloaded by recipes.

Among one of its upsides is the fact that this will allow us to compare sha mismatches should this happen again in the future, as the original files will still be retrievable :)

jcar87 (approved):
Confirmed my upstream maintainers that it is the same source code and that only metadata changed: https://github.com/thestk/rtmidi/issues/318

PR #19253, opened 2023-08-17
co: migrate to Conan v2

valgur:


recipes/co/all/conandata.yml
```diff
@@ -1,8 +1,7 @@
 sources:
   "2.0.2":
     url: "https://github.com/idealvin/co/archive/refs/tags/v2.0.2.tar.gz"
-    sha256: "5dbb90e18627ebe563a6a437e8e28329a6bfc07f266718318bd3adf26027d156"
```
RubenRBS, line 4: Blocking this PR until we understand the reason for this SHA change.

recipes/co/all/conanfile.py
```diff
@@ -30,75 +32,76 @@ class CoConan(ConanFile):
...
-            del self.options.fPIC
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
         if self.options.with_libcurl:
-            self.requires("libcurl/7.79.1")
+            self.requires("libcurl/[>=7]")
```
RubenRBS, line 53: Not allowed, please pin it for now :)

BUT, I'm discussing with the team on allowing it for libcurl as it has a solid compatibility behaviour between versions, will let you know once we come to a conclussion!

RubenRBS (changes requested):
As the recipe is deprecated, I'm reviewing it just as far as "it seems to work" goes, which is looks like it does! However, I'm blocking the merge until we understand why the sha256 has changed

PR #19262, opened 2023-08-17
nss: conan v2 support

valgur:
Continues from #16174.

- Resolves #19579


RubenRBS:
Blocked by https://github.com/conan-io/conan-center-index/pull/18892

recipes/nss/all/conanfile.py
```diff
@@ -17,59 +19,74 @@ class NSSConan(ConanFile):
...
-            del self.options.fPIC
-        del self.settings.compiler.libcxx
-        del self.settings.compiler.cppstd
-
     def requirements(self):
-        self.requires("nspr/4.35")
-        self.requires("sqlite3/3.41.2")
+        self.requires("nspr/4.35", transitive_headers=True)
+        self.requires("sqlite3/3.42.0", run=not cross_building(self))
         self.requires("zlib/1.2.13")
```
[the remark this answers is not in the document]
valgur: Ok, sure. Is this something I can apply to all of the open PRs, maybe? I would take care not push it to all of them at once, of course.

jwillikers (approved):

recipes/nss/all/conanfile.py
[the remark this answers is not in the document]
RubenRBS: Sorry, missed the message - Yeah if you're creating PRs for those, feel free to add the zlib ranges to them :)

recipes/nss/all/conanfile.py
```diff
@@ -17,166 +19,180 @@ class NSSConan(ConanFile):
...
-            del self.options.fPIC
-        del self.settings.compiler.libcxx
-        del self.settings.compiler.cppstd
-
     def requirements(self):
-        self.requires("nspr/4.35")
-        self.requires("sqlite3/3.41.2")
-        self.requires("zlib/1.2.13")
+        self.requires("nspr/4.35", transitive_headers=True, transitive_libs=True)
+        self.requires("sqlite3/3.43.2", run=not cross_building(self))
```
RubenRBS, line 51: Is the not crossbuilding thingy necessary still?

recipes/nss/all/conanfile.py
```diff
@@ -17,166 +19,180 @@ class NSSConan(ConanFile):
...
-        if hasattr(self, "settings_build"):
-            self.build_requires("sqlite3/3.41.2")
-
     def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
         self.options["nspr"].shared = True
         self.options["sqlite3"].shared = True
```
uilianries, line 47:
```suggestion
```
This can not be configured. Qt uses nss, and it will result in missing package when building with Conan 2.x

RubenRBS (commented):
One question about the run trait for sqlite

recipes/nss/all/conanfile.py
mayeut, line 51:
```suggestion
        self.requires("sqlite3/3.44.2", run=not cross_building(self))
```

recipes/nss/all/conanfile.py
valgur: How should this be handled then? Move them to validate()?

jwillikers (approved):

recipes/nss/all/conanfile.py
valgur: It built fine without it.

valgur:
@RubenRBS Apparently the cross-building check was not necessary.

tsondergaard:
@valgur, nss/3.93 appears to be necessary when building qt with qtwebengine. What is required to get this PR merged?

tsondergaard:
What is the motivation for commit " nss: set package_type to static-library, do not set shared for deps"?

Rodarin (approved):
This solves build problems with other libraries that are needed. Approving in an attempt to get more attention on this.

RubenRBS (dismissed):

jwillikers (approved):

recipes/nss/all/conanfile.py
```diff
@@ -17,182 +18,182 @@ class NSSConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     description = "Network Security Services"
     topics = ("network", "security", "crypto", "ssl")
-    settings = "os", "compiler", "build_type", "arch"
-    options = {"shared": [True, False], "fPIC": [True, False]}
```
Rodarin, line 21: Removal of the shared option breaks the Qt recipe with shared=True.

Rodarin (changes requested):

jwillikers (approved):

recipes/nss/all/conanfile.py
uilianries: @Rodarin break how? Could you please give a build log or some consistent detail to understand? For Conan 2.x build in CCI, the CI service uses `-o */*:shared=True/False`, so in case using static library, it will link shared x static. Not the best scenario, but should work as well.

Rodarin: The break is actually caused by a slightly different issue, so I will place a comment at that spot. Sorry for the confusion.

recipes/nss/all/conanfile.py
```diff
@@ -17,182 +18,182 @@ class NSSConan(ConanFile):
...
-    def build_requirements(self):
-        if self.settings.compiler == "Visual Studio" and not tools.get_env("CONAN_BASH_PATH"):
-            self.build_requires("msys2/cci.latest")
-        if self.settings.os == "Windows":
-            self.build_requires("mozilla-build/3.3")
-        if hasattr(self, "settings_build"):
-            self.build_requires("sqlite3/3.41.2")
-
     def configure(self):
-        self.options["nspr"].shared = True
```
Rodarin, line 42: This change forces explicitly setting the nspr:shared=True option, even though that is required in order to build. Yes, it will fail validation below if not set, but it makes no sense to force setting the option. This problem also manifests when building Qt, as you are forced to explicitly set an option for a transitive dependency, which seems even more counterintuitive.

valgur: Good point. Thanks! Reverted.

valgur:
Don't close.

valgur:
.

valgur:
Please trigger the build (or do something about the ill-conceived absolute open PR limit - there's no way I'm going to close 90% of my open PRs just to make one line in some CI script happy).

This package is the only remaining blocker for QtWebEngine support on Linux for Qt6 on CCI.

valgur:
@AbrilRBS @uilianries

valgur:
@jcar87 I patched out the `vswhere` dependency in the build scripts.

valgur:
not stale

recipes/nss/all/conandata.yml
```diff
@@ -1,40 +1,16 @@
 sources:
+  "3.107":
+    url: "https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_107_RTM/src/nss-3.107.tar.gz"
+    sha256: "7f7e96473e38150771a615f5d40e8c41ba3a19385301ae0c525091f2fc9d6729"
   "3.93":
     url: "https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_93_RTM/src/nss-3.93.tar.gz"
     sha256: "15f54bb72048eb105f8c0e936a04b899e74c3db9a19bbc1e00acee2af9476a8a"
-  "3.92":
```
darkrift, line 5: Why are you removing these ?

We are migrating from debian based builds to conan based which currently brings 3.87.1 (so that version would even be missing for us) and in order to minimize risk, we try as much as we can to use the same lib version.

Was there any problem with this or it was solely to minimize the number of patches to make for all those versions ?

valgur: > was solely to minimize the number of patches to make for all those versions ?

Yes. This was an extremely difficult recipe to get right (i.e. to add support on all platforms and cross-compilation support) to due to its very uncommon build system, so I preferred to keeps simple. I can restore any specific versions you might need.

darkrift: Since you already apply the patch to 2 versions (3.107 and 3.93), can't it be applied to all of them ? I mean the worse case was to find out what to change, are those build files changing across version ?

We'd need at least version `3.87.1` (which wasn't in the original list) to help us with our migration to conan based build (before we attempt to upgrade to a later version of nss)

JulianGro:
Thank you for this PR, by the way. We have been using it downstream for months now to build Qt Webengine.

PR #19263, opened 2023-08-17
libgettext/*: Set license as LGPL instead of GPL

jwillikers:
The portions of libgettext that we package are licensed under the LGPL, not the GPL. We only package the libgnuintl portion of the libgettext package, which is under the LGPL. This can easily be confirmed by checking the packaged libintl.h header file. This is also documented explicitly by the project here: https://www.gnu.org/software/gettext/manual/gettext.html#Licenses.

---


recipes/libgettext/all/conanfile.py
```diff
@@ -26,7 +26,7 @@ class GetTextConan(ConanFile):
     topics = ("gettext", "intl", "libintl", "i18n")
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://www.gnu.org/software/gettext"
-    license = "GPL-3.0-or-later"
```
czoido, line 29: Maybe we can add a comment here, to be aware of the different licenses in case we decide to change it in the future?

jwillikers: Done.

czoido (approved):

SSE4 (approved):

PR #19277, opened 2023-08-18
glslang version 13.0.0

psyinf:
Specify library name and version:  **glslang/13.0.0**

This introduces version 13.0.0 which did some restructuring and exposes additional headers needed also for [this PR](https://github.com/conan-io/conan-center-index/pull/16528). 
Unfortunately the patch management concerning the PIC got messy so I decided to move this all to a single patch file

Required by #18996

---

### DIFF
--- a/recipes/glslang/all/conandata.yml
+++ b/recipes/glslang/all/conandata.yml
@@ -38,36 +38,36 @@ sources:
 patches:
   "1.3.239.0":
     - patch_file: "patches/1.3.236.0-0001-no-force-glslang-pic.patch"#
-      patch_type: "conan" 
+      patch_type: "conan"
   "1.3.236.0":
     - patch_file: "patches/1.3.236.0-0001-no-force-glslang-pic.patch"
-      patch_type: "conan" 
+      patch_type: "conan"
   "1.3.231.1":
     - patch_file: "patches/0001-no-force-glslang-pic.patch"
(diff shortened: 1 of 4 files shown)


uilianries:
@psyinf Thank you for your contribution, it seems like some libraries/components are no longer present in this new version. The Conan hook spotted that mismatch. Could you please check the cpp_info vs what is installed in fact? During the package step, all installed artifacts are listed, so it can give a help.

psyinf:
> @psyinf Thank you for your contribution, it seems like some libraries/components are no longer present in this new version. The Conan hook spotted that mismatch. Could you please check the cpp_info vs what is installed in fact? During the package step, all installed artifacts are listed, so it can give a help.
@uilianries I removed the components for the new version/shared build, but I honestly don't feel to confident here

psyinf:
@conan-io/barbarians Anything I can do about the failing builds due to the cmake version on the build-hosts is too low? Since the version is from the upstream package, there is not much I can do here

uilianries:
@psyinf That new version requires C++ Filesystem, that's separated distributed:

https://en.cppreference.com/w/cpp/filesystem#Notes

Please, use the follow workaround to fix it:

https://github.com/conan-io/conan-center-index/blob/master/recipes/metall/all/conanfile.py#L85C48-L85C48

PR #19281, opened 2023-08-19
libhal: add version 2.0.0, add package_type

toge:
Specify library name and version:  **libhal/***

---


RubenRBS (dismissed):

recipes/libhal/all/conanfile.py
```diff
@@ -50,28 +58,37 @@ def lazy_lt_semver(v1, v2):
...
     def source(self):
-        get(self, **self.conan_data["sources"][self.version],
-            destination=self.source_folder, strip_root=True)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def build(self):
+        if Version(self.version) == "0.0.0":
+            replace_in_file(
+                self, 
+                os.path.join(self.source_folder, "include", "libhal", "percentage.hpp"),
```
jcar87, line 72: I can't find any details in the PR description about the need for this patch - is there a compiler error on some specific compiler/version? 

On the other hand, I can see that version `0.0.0` is described as pre-alpha, perhaps we can consider removing the version altogether from the recipe instead.

jcar87 (changes requested):
pending questions re: need for patch of C++ sources, and feedback on removing 0.0.0 altogether

recipes/libhal/all/conanfile.py
jwillikers: It looks like the spaceship operator is supposed to return an `int` and not a `bool`: https://en.cppreference.com/w/cpp/language/operator_comparison#Three-way_comparison

A patch file for this would be nice.

recipes/libhal/all/conanfile.py
```diff
@@ -34,9 +32,19 @@ def _compilers_minimum_version(self):
...
-            "apple-clang": "13.1.6"
+            "apple-clang": "14.0.0"
         }
 
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        if Version(self.version) >= "2.0.0":
+            self.requires("boost-leaf/1.82.0", transitive_headers=True)
```
jwillikers, line 43: I believe the `boost-leaf` package is deprecated as of #16536.

```suggestion
            self.requires("boost/1.83.0", transitive_headers=True)
```

jwillikers (changes requested):

jwillikers (approved):

franramirez688 (approved):

recipes/libhal/all/conanfile.py
toge: @jcar87 
As jwillikers adviced, I create a patch file with patch_desscription.
Please review again.

RubenRBS (approved):
Thanks a lot! The patch makes sense in my eyes, and the rest of the changes look good

RubenRBS (changes requested):
Let's remove the prealpha verison :)

SSE4:
> Let's remove the prealpha verison :)

why remove? what's the motivation and what's the advantage? could you possibly explain a little bit more.

also, this PR's goal is to add version 2.0.0. why is it needed to block it for the reason unrelated to the changes made? if removal of prealpha is really required, it could be safely done in an another PR. why block people from doing changes?

RubenRBS:
It simplifies the recipe, and removes an iffy patch :)

As for why blocking PRs for "unrelated" reasons, we understand that ti might cause _some_ frustration (So we try not to do it for new contributors/if the PR really has nothing to do with the request), but as this PR modifies the recipe to take into account an old version, it's totally fair to ask for changes based on that

SSE4 (approved):

RubenRBS (approved):

franramirez688 (approved):

czoido (approved):

uilianries:
pinging @jcar87 for final review/unblock 😄

uilianries (approved):

SSE4:
@jcar87 ping. I think question was answered a long time ago already.

PR #19318, opened 2023-08-20
sbepp: update dependencies

toge:
Specify library name and version:  **sbepp/***

---

### DIFF
--- a/recipes/sbepp/all/conanfile.py
+++ b/recipes/sbepp/all/conanfile.py
@@ -64,8 +64,8 @@ class PackageConan(ConanFile):
 
     def requirements(self):
         if self.options.with_sbeppc:
-            self.requires("fmt/9.1.0")
-            self.requires("pugixml/1.12.1")
+            self.requires("fmt/10.1.0")
+            self.requires("pugixml/1.13")
 
     def validate(self):
         if self.settings.compiler.cppstd:


RubenRBS (approved):

jwillikers (approved):

OleksandrKvl:
@toge and others, may I ask what issue this PR solved? I'm the author and the project was initially developed and tested using `fmt/9.1.0` and `pugixml/1.12.1`, users that want to use higher versions can always override them. What was the need to hardcode higher versions of those? Imagine I have a project with fmt/9 and sbepp/1.1.0, this change silently breaks it.

RubenRBS:
Hi @OleksandrKvl - Thanks a lot for your question. There are a few reasons that CCI has been doing this historically. One of the most important ones is graph divergence. We usually try to keep libraries with dependencies aligned, so that using any two libraries does not cause a version conflict. For this, version bumps of dependencies are needed.

It's true that this might break libraries, so we appreciate when we get reports of bumps not being valid and we can then leave a comment in the recipes themselves to warn future bumps.
What other recipes usually do is to only have bumps to dependencies in their latest version

But yeah, in the end this is the age old question of staying updated vs having a consistent graph. I hope this helps a little, but please feel free to discuss your thoughts if needed.

PR #19338, opened 2023-08-22
update openssl dependency to use lts version by default

pgeler:
Specify library name and version:  grpc/1.54.3 and 1.48.4

new openssl release model is tracking multiple major releases, so that prefer by default build package with lts version instead of sts

---

### DIFF
--- a/recipes/grpc/all/conanfile.py
+++ b/recipes/grpc/all/conanfile.py
@@ -93,7 +93,8 @@ class GrpcConan(ConanFile):
         else:
             self.requires("abseil/20230125.3", transitive_headers=True, transitive_libs=True)
         self.requires("c-ares/1.19.1")
-        self.requires("openssl/[>=1.1 <4]")
+        # openssl latest avaliable lts is 3.0.x
+        self.requires("openssl/[>=1.1 <3.1]")
         self.requires("re2/20230301")
         self.requires("zlib/1.2.13")
         self.requires("protobuf/3.21.12", transitive_headers=True, transitive_libs=True, run=can_run(self))


jcar87:
Hi @pgeler - thanks for this PR.

Our policy for this is for the version range to be open for all the 3.x series - assuming that this [policy](https://www.openssl.org/policies/releasestrat.html) is followed and that  no new release of 3.x is ABI/API breaking. Thus, we happily want the version range to extend to new releases beyond the current LTS one.

The version ranges do allow users to override the dependency - for uses that wish to use a specific LTS they can simply have an override in their root conanfile.py, or use lockfiles to lock to a specific version. Is this mechanism not suitable for your workflows?

pgeler:
that is exactly the point of this PR, the only lts version is 3.0.x while 3.1.x is not an lts, missing something?

pgeler:
> Hi @pgeler - thanks for this PR.
> 
> Our policy for this is for the version range to be open for all the 3.x series - assuming that this [policy](https://www.openssl.org/policies/releasestrat.html) is followed and that  no new release of 3.x is ABI/API breaking. Thus, we happily want the version range to extend to new releases beyond the current LTS one.
> 
> The version ranges do allow users to override the dependency - for uses that wish to use a specific LTS they can simply have an override in their root conanfile.py, or use lockfiles to lock to a specific version. Is this mechanism not suitable for your workflows?
> 

interesting (thanks for the link). desiring something not necessary how its all landed... anyway to give some background its all started from 1.1.1s to 1.1.1v upgrade which has api/abi  guards under defines. In general I don't have any issues with using conanfile, with exception of tracking old packages/receipts and breaking receipts, so that outdated recept does not support new dependency package etc... let me dig my problem first...

recipes/grpc/all/conanfile.py
```diff
@@ -93,7 +93,8 @@ def requirements(self):
         else:
             self.requires("abseil/20230125.3", transitive_headers=True, transitive_libs=True)
         self.requires("c-ares/1.19.1")
-        self.requires("openssl/[>=1.1 <4]")
+        # openssl latest avaliable lts is 3.0.x
+        self.requires("openssl/[>=1.1 <3.1]")
```
Nekto89, line 97: how does this recipe already work with OpenSSL 3? PR with OpenSSL 3 was committed to master (so no release yet) just yesterday
https://github.com/grpc/grpc/pull/31256

jcar87: if the changes were merged it is because all the packages built for all ConanCenter supported configurations - so at the very least it is compatible at the source level and linking level.

From what I can see, other package managers have also convered to OpenSSL 3, even for grpc and certainly before this commit that landed yesterday. 

Chances are the APIs used are still in OpenSSL3 and that PR just moves to use newer ones in some instances? Also the PR does address some tests as well (which as we know, are not covered by Conan Center CI).

pgeler:
@jcar87 ok, problem with my upgrade was in combination of broken cprypto ABI 1.1.1s - 1.1.1v and incomplete conanfile dependencies for one of the sub-projects, which cause two separate set of openssl headers for main project and another one which was coming from outdated grpc receipt... 

So I'm good with the workaround... Well not that much :), now I have to go and update all projects to use specific version of the openssl library, which another pain point due to conan search statement search slowness(was it fixed at all?).

 IMHO knowing that openssl just failed to follow their own rules, I'm not in very optimistic about long term approach for such wide range of dependencies, as well as it's establishing incorrect practice where versioning in one library could be different than another and conan maintainers will start injecting search style dependencies into receipts without deep considerations. Which might be well suited for small projects, but users who maintain more stable environments, would have to spend much more time maintaining conan dependencies and that would limit conan benefits to a tool for upload/download antifactory packages.

jcar87:
>  broken cprypto ABI 1.1.1s - 1.1.1v 

do you have more details?

> incomplete conanfile dependencies for one of the sub-projects,
incomplete? Do you mean a recipe depended on something but it missed expressing a dependency on that something? that could be a problem, but sounds less like an OpenSSL problem and more of a recipe problem

> now I have to go and update all projects to use specific version of the openssl library, which another pain point due to conan search statement search slowness(was it fixed at all?).

I'm surprised you have to go and update all your projects. Given a dependency graph, if you are using lockfiles to lock to a specific version of OpenSSL, or use an override from the top-level conanfile, the solver will resolve to specifically that version, repeatedly - you do not have to change any recipe (other than your top-level one).

> IMHO knowing that openssl just failed to follow their own rules, 

Do you have more details? From what you are saying, it sounds like a library that depended on OpenSSL did not record this dependency, thus when the solver elsewhere used a different version, rather than causing Conan to rebuild that library against the different (incompatible) version, it just reused binaries built with an older version. Which is obviously not going to work - that's why libraries need to have requirements on the things they actually and directly depend on.

> I'm not in very optimistic about long term approach for such wide range of dependencies, as well as it's establishing incorrect practice where versioning in one library could be different than another and conan maintainers will start injecting search style dependencies into receipts without deep considerations. 

The Conan solver will only resolve to a version that is compatible with the requirements across the graph. Additionally, Conan will _require_ new binaries to be rebuilt the circumstances that require it - which should avoid all cases of API or linker errors, assuming the library does maintain their documented promises. Thus, this should not cause issues. 

> Which might be well suited for small projects, but users who maintain more stable environments, would have to spend much more time maintaining conan dependencies and that would limit conan benefits to a tool for upload/download antifactory packages.

Using version ranges is not exclusive to Conan, and the approach is similar to the default behaviours in pip/Pypi, npm, Rust cargo, Conda, etc - In all of those scenarios, maintain stable environments is not only possible, by highly recommended, by using lockfiles on the consumer end - this ensures repeatability. Without removing the flexibility of using new versions of dependencies if new versions are released that are known to be compatible.

pgeler:
> > broken cprypto ABI 1.1.1s - 1.1.1v
> 
> do you have more details?
original change caused incompatibility was in following typedef: https://github.com/openssl/openssl/blob/6399d7856c75abde9ed23782d10960013de03810/include/openssl/crypto.h.in#L220
, however, I don't have a good answer why it was raised for the upgrade and it's not priority right now as receipt issue already addressed.

> 
> > incomplete conanfile dependencies for one of the sub-projects,
> 
> incomplete? Do you mean a recipe depended on something but it missed expressing a dependency on that something? that could be a problem, but sounds less like an OpenSSL problem and more of a recipe problem
no doubts about that, the problem more in the order of those dependencies
> 
> > which cause two separate set of openssl headers for main project and another one which was coming from outdated grpc receipt...
> 
> Conan should only resolve to a single version on OpenSSL.
> 

I believe this is project setup issue

> > now I have to go and update all projects to use specific version of the openssl library, which another pain point due to conan search statement search slowness(was it fixed at all?).
> 
> I'm surprised you have to go and update all your projects. Given a dependency graph, if you are using lockfiles to lock to a specific version of OpenSSL, or use an override from the top-level conanfile, the solver will resolve to specifically that version, repeatedly - you do not have to change any recipe (other than your top-level one).

we are not using lockfiles and might want to start doing so. All sub-projects holding their own dependencies, so this was perfect way to `shot ourself to the foot`, and right now I don't see any way to build single dependency graph using conanfiles and msvc except link everything with all dependencies... well actually might be usage of $(ConanIncludeDirectories)/$(ConanLinkerFlags) was not perfect way of doing so

> 
> > IMHO knowing that openssl just failed to follow their own rules,
> 
> Do you have more details? From what you are saying, it sounds like a library that depended on OpenSSL did not record this dependency, thus when the solver elsewhere used a different version, rather than causing Conan to rebuild that library against the different (incompatible) version, it just reused binaries built with an older version. Which is obviously not going to work - that's why libraries need to have requirements on the things they actually and directly depend on.
> 
> > I'm not in very optimistic about long term approach for such wide range of dependencies, as well as it's establishing incorrect practice where versioning in one library could be different than another and conan maintainers will start injecting search style dependencies into receipts without deep considerations.
> 
> The Conan solver will only resolve to a version that is compatible with the requirements across the graph. Additionally, Conan will _require_ new binaries to be rebuilt the circumstances that require it - which should avoid all cases of API or linker errors, assuming the library does maintain their documented promises. Thus, this should not cause issues.
> 

Agree.

> > Which might be well suited for small projects, but users who maintain more stable environments, would have to spend much more time maintaining conan dependencies and that would limit conan benefits to a tool for upload/download antifactory packages.
> 
> Using version ranges is not exclusive to Conan, and the approach is similar to the default behaviours in pip/Pypi, npm, Rust cargo, Conda, etc - In all of those scenarios, maintain stable environments is not only possible, by highly recommended, by using lockfiles on the consumer end - this ensures repeatability. Without removing the flexibility of using new versions of dependencies if new versions are released that are known to be compatible.

we are using conan from early 1.1 versions, without too much attention to the new features... and there is something to address here, however as of right now it seems that lockfiles are in beta version and `could introduce breaking changes`... so dealing with few issues with conan using such feature to us sounds very dangerous 🤔

pgeler:
I'm open to continue this conversation, but it's clearly out of subject, so closing this PR

PR #19339, opened 2023-08-22
qt/6.5.x: Only output a warning when using GCC 9

jwillikers:
Fail explicitly for GCC 9.1 or 9.2 which won't work. Users using any version of GCC 9 may be using newer versions of GCC 9. A warning should suffice in this case, without requiring users to change their settings model to incorporate minor version numbers of the compiler. For details, see https://bugreports.qt.io/browse/QTBUG-112920

Specify library name and version:  **qt/6.5.x**

---

### DIFF
--- a/recipes/qt/6.x.x/conandata.yml
+++ b/recipes/qt/6.x.x/conandata.yml
@@ -117,6 +117,10 @@ patches:
     - "base_path": "qtwebengine"
       "patch_description": "Workaround for too long .rps file name"
       "patch_file": "patches/c72097e.diff"
+    - base_path: "qtbase"
+      patch_description: "Fix compilation with point releases of GCC 9 prior to 9.3"
+      patch_file: "patches/148895_0001-Fix-FTBFS-qt-6.5.0-on-gcc-9.patch"
+      patch_source: "https://bugreports.qt.io/browse/QTBUG-112920"
   "6.5.1":
     - base_path: "qtwebengine"
       patch_description: "Workaround for too long .rps file name"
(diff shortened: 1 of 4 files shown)


MartinDelille (approved):

RubenRBS (approved):
We're checking the possibility of updating to gcc >=9.3 in the CI machines, but some issues arose - let's get this in and check back once an updated version (if any) is added :)

ericLemanissier (changes requested):
we should be able to build with any version of gcc 9 by applying this patch on qt 6.5.0 to 6.6.0 : https://bugreports.qt.io/secure/attachment/148895/148895_0001-Fix-FTBFS-qt-6.5.0-on-gcc-9.patch
The patch has been reviewed upstream : https://bugreports.qt.io/browse/QTBUG-112920?focusedId=757912&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-757912

RubenRBS:
Thanks for the insight @ericLemanissier! Let's do that :)

ericLemanissier (changes requested):
Thanks. You should also remove the exceptions and warning triggered when the profile has gcc 9

jwillikers:
> Thanks. You should also remove the exceptions and warning triggered when the profile has gcc 9

Duh! Thanks!

ericLemanissier (approved):

ericLemanissier (changes requested):
The patch must also be applied to 6.6.0

jwillikers:
Closing as it looks as though this check was removed entirely.

PR #19351, opened 2023-08-22
boost: make sure libdirs is empty with header-only=True and used via `boost::boost` target

gegles:
Specify library name and version:  **boost/1.83**

This PR helps remove the following warning on macOS when boost is used as header-only, but the `boost::boost` target is used in CMake as opposed to the `Boost::headers`:
```
ld: warning: directory not found for option '-L/Users/gegles/.conan2/p/boost3ea68282e5511/p/lib'
```

I faced this issue when using `restinio` that transitively brings in `boost` but internally uses the `boost::boost` CMake target.

---


jcar87 (approved):
Thanks @gegles, for the contribution and for the PR description.

PR #18106, opened 2023-06-28
arrow: add version 12.0.1, dirty workaround for conan v2

toge:
Specify library name and version:  **arrow/***

I took the easy workaround to the problem of `package_id()` prohibiting access to `self.options` in conan v2.0.
I know that this is not the official way from previous discussions:
https://github.com/conan-io/conan/issues/13227
If someone give me a hint of a better solution, I will try to implement it.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #19380, opened 2023-08-24
[fix] Arrow: Make it conan v2 comaptible, refactor auto options and add 12.0.1 version

danimtb:
Specify library name and version:  **arrow/***

Includes changes from and closes https://github.com/conan-io/conan-center-index/pull/18106

I wanted to fix the workaround proposed at https://github.com/conan-io/conan-center-index/pull/18106, but it turned out to be more complex than expected due to the logic implemented in the recipe.

I tried to simplify the management of the "auto" options following these rules:
- The final value of the options should never be "auto".
- The final value of the options should be computed as soon as possible in the recipe.
- The final value of the options should be the ones used as conditions in the rest of the recipe methods.

cc/ @toge


valgur (approved):

recipes/arrow/all/conanfile.py
```diff
@@ -49,8 +49,9 @@ class ArrowConan(ConanFile):
         "with_glog": ["auto", True, False],
         "with_grpc": ["auto", True, False],
         "with_jemalloc": ["auto", True, False],
-        "with_mimalloc": ["auto", True, False],
```
danimtb, line 52: I removed the "auto" value here as there was no real automatic management in place for this options and I didn't investigate the library's build system to know about it. I think it is fine as is.

recipes/arrow/all/conanfile.py
```diff
@@ -49,8 +49,9 @@ class ArrowConan(ConanFile):
...
+        "with_mimalloc": [True, False],
         "with_json": [True, False],
+        "with_thrift": ["auto", True, False],
```
danimtb, line 54: this option did not exist and was conditional to other one. I think it is clear to have it as a real option.

RubenRBS (approved):

jwillikers (approved):

recipes/arrow/all/conanfile.py
```diff
@@ -155,39 +157,55 @@ def config_options(self):
...
+        self.options.with_gflags = self._with_gflags()
+        self.options.with_glog = self._with_glog()
+        self.options.with_grpc = self._with_grpc()
+        self.options.with_jemalloc = self._with_jemalloc()
+        self.options.with_thrift = self._with_thrift()
+        self.options.with_llvm = self._with_llvm()
+        self.options.with_openssl = self._with_openssl()
+        self.options.with_protobuf = self._with_protobuf()
+        self.options.with_re2 = self._with_re2()
+        self.options.with_utf8proc = self._with_utf8proc()
```
SpaceIm, line 174: I just want to mention that all this logic is broken, since all these functions try to access options values which don't even exist at `config_options()` time.

dvirtz: what is the proper way to define dependencies between options in conan 2?

PR #18016, opened 2023-06-22
[intel-ipsec-mb] Add new recipe: Intel IPSec MB 1.4

uilianries:
Specify library name and version:  **intel-ipsec-mb/1.4**

In 2021 there was an effort to add Intel IPSec MB in CCI, but we had an infra limitation.

After all these months, the project now has CMake supported. Let's see if we can bring it to CCI now.

closes #6758

/cc @gokhanettin 

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #19394, opened 2023-08-24
[nasm] Configure assembler compiler by cpp_info

uilianries:
Specify library name and version:  **nasm/2.15.2**

Related to https://github.com/conan-io/conan-center-index/pull/18016

The `nasm` Conan package only stores binaries but forget to configure them properly to packages that will use as a tool requirement.

CMake will try use the system nasm and ignore that one provided by the package, in case not configured. Plus, `AS` variable is used for assembler.

This PR fixes a situation found when building `intel-ipsec-mb` where I needed to pass:

`tc.variables["CMAKE_ASM_NASM_COMPILER"] = os.path.join(self.dependencies.direct_build["nasm"].package_folder, "bin", "nasm")`

So I'm able to use nasm from Conan. Otherwise, CMake (using CMakeToolchain + CMakeDeps) will consume from system.

---

### DIFF
--- a/recipes/nasm/all/conanfile.py
+++ b/recipes/nasm/all/conanfile.py
@@ -17,13 +17,27 @@ class NASMConan(ConanFile):
     homepage = "http://www.nasm.us"
     description = "The Netwide Assembler, NASM, is an 80x86 and x86-64 assembler"
     license = "BSD-2-Clause"
-    topics = ("nasm", "installer", "assembler")
+    topics = ("asm", "installer", "assembler",)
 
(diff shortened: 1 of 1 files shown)


uilianries:
I wanted to bump nasm version to 2.16.01 too, but when building locally, I've affected by the known bug https://bugzilla.nasm.us/show_bug.cgi?id=3392905

RubenRBS (approved):

prince-chrismc (approved):

PR #19399, opened 2023-08-24
freertos-kernel/11.1.0: Add FreeRTOS package

jwillikers:
Specify library name and version:  **freertos-kernel/11.1.0**

Still working out how to manage the `FreeRTOSConfig.h` header in a way that makes sense.

---


recipes/freertos-kernel/all/conanfile.py
```diff
@@ -0,0 +1,403 @@
...
+from conan.tools.cmake import cmake_layout, CMake, CMakeToolchain
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, save
+
+required_conan_version = ">=1.60.0"
+
+
+
+default_configs = {
+    "GCC_POSIX": textwrap.dedent("""\
+        #ifndef FREERTOS_CONFIG_H
```
Ju1He1, line 15: I came across an interesting article regarding target_sources
target_sources()

https://crascit.com/2016/01/31/enhanced-source-file-handling-with-target_sources/

If i saw it right the whole FreeRTOS is already and interface library. Is it possible to add the FreeRTOSconfig.h via target_sources within the application itself?
Would it be possible to add something like this to the actual application?

```
find_library(FREERTOS_LIB FreeRTOS::FreeRTOS)
target_sources(${FREERTOS_LIB} "FreeRTOSConfig.h")
```

jwillikers: I don't think that work when the recipe itself builds FreeRTOS because `FreeRTOSConfig.h` needs to be made available to the recipe when it is being built.

The difficulties of handling `FreeRTOSConfig.h` directly are determining how to pass `FreeRTOSConfig.h` to Conan such that it is exported as part of the recipe, its contents are included as part of the package_id to distinguish the binary, and used to compile the recipe itself.

Ju1He1: on second thought you are probably right :(

jcar87:
After talking with the team, we have determined this is not currently a priority for Conan Center. Our experience with recipes with this many options is that in practice, most users don't deviate from the defaults, and when they do, things don't work because they're untested (by CI or by us), increasing the maintenance burden on us to find a solution. 

I would encourage opening an issue to request this, detailing in which contexts this recipe would be useful - any links to documentation with usage examples are appreciated. It's also helpful so that other users can also articulate their needs to help us guide a decision.

PR #19411, opened 2023-08-25
Kdbindings fix

bog-dan-ro:
**kdbindings/1.0.3**

 - added "msvc": "191" to compilers list
 - clean settings attribute

---

### DIFF
--- a/recipes/kdbindings/all/conanfile.py
+++ b/recipes/kdbindings/all/conanfile.py
@@ -13,7 +13,7 @@ class KDBindingsConan(ConanFile):
     homepage = "https://github.com/KDAB/KDBindings"
     url = "https://github.com/conan-io/conan-center-index"
     package_type = "header-library"
-    settings = "os", "arch", "compiler", "build_type"
+    settings = "compiler"
     no_copy_source = True
 
     def package_id(self):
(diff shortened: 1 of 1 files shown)


recipes/kdbindings/all/conanfile.py
```diff
@@ -13,7 +13,7 @@ class KDBindingsConan(ConanFile):
     homepage = "https://github.com/KDAB/KDBindings"
     url = "https://github.com/conan-io/conan-center-index"
     package_type = "header-library"
-    settings = "os", "arch", "compiler", "build_type"
+    settings = "compiler"
```
valgur, line 16:
```suggestion
    settings = "os", "arch", "compiler", "build_type"
```
The convention in Conan 2.0 is to always set all settings and clear them in `package_id()` instead.

valgur (changes requested):

perseoGI (approved):
LGTM

recipes/kdbindings/all/conanfile.py
uilianries: @bog-dan-ro Please, avoid changing settings, so we can use any settings in the recipe. The package ID will not change due `self.info.clear()`.

uilianries (approved):
LGTM

Issue #19379, opened 2023-08-24
[package] clickhouse-cpp/2.4.0: with_openssl and enable_benchmark options not functional


tgurriet:
### Description

Hello,

The `with_openssl` and `enable_benchmark` options of the `clickhouse-cpp` package are not functional.

I believe this is because these 2 options are passed to the cmake toolchain as `variables` instead of `cache_variables`:
https://github.com/conan-io/conan-center-index/blob/4adbf0e5da9409b1a34dd3e07f65f42323bcad97/recipes/clickhouse-cpp/all/conanfile.py#L90
https://github.com/conan-io/conan-center-index/blob/4adbf0e5da9409b1a34dd3e07f65f42323bcad97/recipes/clickhouse-cpp/all/conanfile.py#L92
and the clickhouse-cpp [CMakeLists.txt](https://github.com/ClickHouse/clickhouse-cpp/blob/a133b50db0160f4e2b1e0288a99f123ae76793b7/CMakeLists.txt#L9-L11) defines these options without the `CACHE` attribute:
```
OPTION (BUILD_BENCHMARK "Build benchmark" OFF)
OPTION (BUILD_TESTS "Build tests" OFF)
OPTION (BUILD_SHARED_LIBS "Build shared libs" OFF)
```
so the generated options in the conan_toolchain.cmake:
```
set(BUILD_BENCHMARK "False" CACHE STRING "Variable BUILD_BENCHMARK conan-toolchain defined")
set(WITH_OPENSSL "True" CACHE STRING "Variable WITH_OPENSSL conan-toolchain defined")
```
get overwritten.

Passing these 2 options as `cache_variables` does fix the issue, but I'm wondering if it could have something to do with the fact that `variables` generated from recipe options generate `True/False` strings CMake variables instead of `ON/OFF` boolean CMake variables!?

`cache_variables` get properly converted in the `CMakePresets.json`:
```
...
            "cacheVariables": {
                "BUILD_BENCHMARK": "OFF",
                "BUILD_SHARED_LIBS": "OFF",
                "WITH_OPENSSL": "ON",
                "WITH_SYSTEM_ABSEIL": "ON",
                "WITH_SYSTEM_LZ4": "ON",
                "WITH_SYSTEM_CITYHASH": "ON",
                "CMAKE_POLICY_DEFAULT_CMP0091": "NEW",
                "CMAKE_BUILD_TYPE": "Debug"
            },
...
```

### Package and Environment Details

[...]

PR #19449, opened 2023-08-27
clickhouse-cpp: enable `enable_benchmark` and `with_openssl` options

toge:
Specify library name and version:  **clickhouse-cpp/***

- set `BUILD_BENCHMARK` and `WITH_OPENSSL` by `cache_variables`
- When `enable_benchmark` is True, clickhouse-cpp requires benchmark and gtest.

Try to close https://github.com/conan-io/conan-center-index/issues/19379.

---

### DIFF
--- a/recipes/clickhouse-cpp/all/conanfile.py
+++ b/recipes/clickhouse-cpp/all/conanfile.py
@@ -1,6 +1,6 @@
 from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeToolchain,CMakeDeps, cmake_layout
-from conan.tools.files import copy, get
+from conan.tools.files import copy, get, export_conandata_patches, apply_conandata_patches
 from conan.tools.build import check_min_cppstd
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.scm import Version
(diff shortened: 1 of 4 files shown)


SpaceIm:
Why `enable_benchmark` is not removed from package_id() in this recipe, and the first question should be why such option has been accepted to begin with?

SpaceIm:
Why is there an `enable_benchmark` option? Why has it been accepted?

recipes/clickhouse-cpp/all/conanfile.py
```diff
@@ -81,15 +63,36 @@ def configure(self):
...
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support.")
+        if self.settings.os == "Windows" and self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} does not support shared library on Windows.")
+            # look at https://github.com/ClickHouse/clickhouse-cpp/pull/226
+
+    def build_requirements(self):
+        if self.options.enable_benchmark:
+            self.requires("benchmark/1.8.2")
+            self.requires("gtest/1.14.0")
```
SpaceIm, line 86: Seems to be broken logic. I guess you want test_requires. But enable_benchmark option should be removed I think.

toge: @SpaceIm 
Oh. You are right.
It is not a necessary option.

I will update this PR to remove `enable_benchmark` option and write a comment to the source issue.

toge: @Jihadist 
Thanks!

recipes/clickhouse-cpp/all/conanfile.py
```diff
@@ -99,6 +94,7 @@ def generate(self):
         cd.generate()
 
     def build(self):
+        apply_conandata_patches(self)
```
Jihadist, line 97: Looks like we do not need it too

toge: @Jihadist 
Thanks!

SpaceIm (approved):

czoido (approved):

PR #19460, opened 2023-08-28
Add ohnet

DoomHammer:
Specify library name and version:  **ohnet/1.36.5113**

This library is a dependency of another library: ohPipeline. Both are required to build audio products based on the OpenHome protocol.

---

- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated. <- I'm not sure how to make them work with Conan 2.0 as the 1.0 instructions don't work.


DoomHammer:
@uilianries @memsharded is it good to go?

recipes/ohnet/all/conanfile.py
```diff
@@ -0,0 +1,141 @@
...
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def validate(self):
+        if is_msvc(self) and (self.options.shared or msvc_runtime_flag(self).startswith('MD')):
```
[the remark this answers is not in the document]
DoomHammer: The Makefiles in the upstream source basically only use MT in the most recent version. Previously they had some magic conditions on when to use one versus another. Rather that working around upstream problems I'd block the unsupported configurations for now.

If you have any suggestions how to better approach this, I'd be happy to try them.

recipes/ohnet/all/conanfile.py
```diff
@@ -0,0 +1,141 @@
...
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def generate(self):
+        if is_msvc(self):
+            tc = NMakeToolchain(self)
+            tc.generate()
+        else:
+            tc = AutotoolsToolchain(self)
+            tc.make_args.append("-j1")
```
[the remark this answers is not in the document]
DoomHammer: The upstream Makefiles can't run in parallel. So it was either fixing the upstream Makefiles which might have been difficult or just running serial ;_;

recipes/ohnet/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,775 @@
+#include <memory>
```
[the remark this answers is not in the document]
DoomHammer: OK!

recipes/ohnet/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+cmake_minimum_required(VERSION 3.15)
+
+project(test_package CXX)
+
+find_package(ohNet REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} test_package.cpp)
+target_compile_definitions(${PROJECT_NAME} PUBLIC "DEFINE_LITTLE_ENDIAN")
```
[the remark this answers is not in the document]
DoomHammer: Good question. Because I had no idea that it's possible to do so :)

recipes/ohnet/all/conanfile.py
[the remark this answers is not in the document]
uilianries: @DoomHammer thank you for clarifying, so in case the upstream does not support it very well, I agree in not supporting it here too.

recipes/ohnet/all/conanfile.py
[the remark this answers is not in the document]
uilianries: Oh gosh! Thank you for the information!

DoomHammer:
@uilianries is it better now?

recipes/ohnet/all/test_v1_package/CMakeLists.txt
uilianries, line 1: @DoomHammer You actually can remove all the test_v1_package folder. It's no longer needed for newer recipes in CCI: https://github.com/conan-io/conan-center-index/discussions/19104

It results in less maintenance to this recipe in the future 😄

DoomHammer: Man, how much time I've spent trying to make it work 😅

uilianries: We used the test_v1_package to validate legacy generators like `cmake ` and `cmake_find_package_multi`, but with Conan 2.x and backward compatibility of `CMakeDeps` and `CMakeToolchain`, the recommendation is using only them. And, as we are planning at some point dropping Conan 1.x from ConanCenterIndex, we will first remove it by parts, like `test_v1_package` and some configurations with the CI.

recipes/ohnet/config.yml
```diff
@@ -0,0 +1,5 @@
+versions:
+  "1.36.5113":
+    folder: all
+  "1.36.5182":
```
uilianries, line 4: This newer version is missing in conandata.yml. But, please, let's only the latest version in case you don't need both now.

recipes/ohnet/all/conandata.yml
```diff
@@ -1,7 +1,7 @@
 sources:
-  "1.36.5113":
-    url: "https://github.com/openhome/ohNet/archive/refs/tags/ohNet_1.36.5113.tar.gz"
-    sha256: "fb71d92fac37cb6fd0688c0d53689c5ab96cf2de7060b72555d05a1831f4e2a4"
+  "1.36.5182":
+    url: "https://github.com/openhome/ohNet/archive/refs/tags/ohNet_1.36.5182.tar.gz"
+    sha256: "1489407b9bae13affa8f933a81fb5a43f16b29f5bf8048b68ea645afc0e53ced"
 patches:
   "1.36.5113":
```
uilianries, line 6:
```suggestion
  "1.36.5182":
```

Need to update the patch too. Hope it works without extra change.

recipes/ohnet/all/conanfile.py
```diff
@@ -0,0 +1,145 @@
...
+        return args
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
```
uilianries, line 45:
```suggestion

    def layout(self):
        basic_layout(self, src_folder="src")

```

The advantage using layout is not mixing source files with build cache in the build folder, so when developing a package, you can list/run much easier in case an error.

https://docs.conan.io/2/tutorial/developing_packages/package_layout.html

recipes/ohnet/all/conanfile.py
```diff
@@ -0,0 +1,145 @@
+from conan import ConanFile
+from conan.tools.files import get, chdir, copy, mkdir, export_conandata_patches, apply_conandata_patches
+from conan.tools.apple import fix_apple_shared_install_name, is_apple_os
+from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.microsoft import is_msvc, msvc_runtime_flag, NMakeToolchain
+from conan.errors import ConanInvalidConfiguration
```
uilianries, line 6:
```suggestion
from conan.tools.layout import basic_layout
from conan.errors import ConanInvalidConfiguration
```

uilianries (commented):
Just a last comment, but looking good. I tested locally and it's working.

uilianries (approved):
LGTM Thank you @DoomHammer !!!

recipes/ohnet/all/conanfile.py
```diff
@@ -0,0 +1,149 @@
...
+                autotools = Autotools(self)
+                autotools.make(args=args, target="install-libs install-includes")
+                if is_apple_os(self):
+                    fix_apple_shared_install_name(self)
+
+    def package_info(self):
+        self.cpp_info.components["ohNet"].libs = ["ohNet"]
+        self.cpp_info.components["ohNet"].builddirs = [os.path.join("lib", "ServiceGen")]
+        self.cpp_info.components["ohNet"].set_property("cmake_target_name", "ohNet")
+        self.cpp_info.components["ohNet"].defines.extend(["DEFINE_LITTLE_ENDIAN"])
```
jwillikers, line 115: This seems architecture specific here and like something we shouldn't be defining without respect to the target architecture. Of course, we don't really do anything to define the endianness of architectures which can be big or little endian. Can the upstream not solve this in a more portable way?

jwillikers (changes requested):
I have a concern that always defining `DEFINE_LITTLE_ENDIAN` won't be portable.

recipes/ohnet/all/conanfile.py
DoomHammer: There are many things I'd wish upstream solved in a more portable way, starting with using CMake instead of Makefiles that do everything including cross-compilation guesses.

There used to be a PR with CMake even, but  don't think any maintainer even looked at it.

That said, from a pragmatic point of view, I prefer to work around such issues on a higher level (Conan package in this instance) and thus have a working solution as opposed to waiting indefinitely for upstream support.

Is there a way in Conan to check and set the endianness in a way that should be compatible accross platforms?

jwillikers: I don't think so since the `arch` settings don't really include endianness. Arm can be either big or little endian. It's likely it would require a patch to the source code to calculate it at compile time. Another option is to omit it here and require that consumers define it themselves. You could just define `DEFINE_LITTLE_ENDIAN` in the test package then to have it build on CCI.

DoomHammer: That's what I did in the first place, but @uilianries directed me at the defines and that they might be a better option. And I kinda agree.

Funny things, whereas upstream deifines `DEFINE_LITTLE_ENDIAN` several times in the Makefiles, `DEFINE_BIG_ENDIAN` is never there, so it seems like they don't even support any Big Endian platforms.

jwillikers (approved):

PR #19470, opened 2023-08-28
mimalloc: relax cppstd requirements because we don’t compile tests

tankeco:
Specify library name and version:  **mimalloc/\***

the source code of mimalloc is written almost in plain c, only the tests needs c++17 (see [link](https://github.com/microsoft/mimalloc/blob/master/test/CMakeLists.txt#L5)) and the recipe set `MI_BUILD_TESTS=OFF`

---


RubenRBS:
I've been looking for some documentation upstream that mentions the C++ 11 requirement but I haven't been able to find anything yet. Could you please point me to it? (Or did you set on 11 any other way?)
Thanks! :)

tankeco:
> 

I think there is no c++11 requirement. I changed 17 to 11 only because it is the smallest change :)

RubenRBS:
Hi, sorry for the delayed follow up. I think removing the cppstd check as you have already done is the way to go (Or if it's a plain C lib, compiling it as such would also work!)

Thanks for the fix :)

valgur:
If it's effectively a C-only project, then clearing the libcxx and cppstd settings would be a welcome addition as well: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L67

tankeco:
> If it's effectively a C-only project, then clearing the libcxx and cppstd settings would be a welcome addition as well: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L67

I'm not very sure. The implementation is pure c, but the header files support c++ std containers:
https://github.com/microsoft/mimalloc/blob/master/include/mimalloc.h#L421

RubenRBS (approved):
Thanks a lot!

PR #19480, opened 2023-08-29
poco: Add option for enabling debug and trace logs

steinerthomas:
Specify library name and version:  **poco/***

With the preprocessor definition `POCO_LOG_DEBUG` the logging levels `debug` and `trace` can also be logged with Release builds. With the option `log_debug` this switch can be activated.

---


recipes/poco/all/conanfile.py
```diff
@@ -29,12 +29,14 @@ class PocoConan(ConanFile):
         "fPIC": [True, False],
         "enable_fork": [True, False],
         "enable_active_record": [True, False, "deprecated"],
+        "log_debug": [True, False],
```
steinerthomas, line 32: The `log_debug` option could be deleted in the `config_options` method when building Debug and read with `get_save` below. But I think creating the package will be more challenging then...

recipes/poco/all/conanfile.py
```diff
@@ -200,6 +202,9 @@ def generate(self):
         tc.preprocessor_definitions["POCO_NO_AUTOMATIC_LIBS"] = "1"
         # Picked up from conan v1 CMake wrapper, don't know the rationale
         tc.preprocessor_definitions["XML_DTD"] = "1"
+        # Enable debug and trace logs for Release builds
+        if self.options.log_debug:
+            tc.preprocessor_definitions["POCO_LOG_DEBUG"] = "1"
```
uilianries, line 207:
```suggestion
```
It's not needed, because is only consumed by public headers. Only when consuming Poco it makes sense.

uilianries (changes requested):
@steinerthomas Thank you for your contribution. The `log_debug` does not affect the generated library, because it's only consumed by public headers. In that case, you should remove it from the package ID:


```
def package_id(self):
    del self.info.options.log_debug
```

recipes/poco/all/conanfile.py
steinerthomas: Oh you are right!
I guess I can close this pull request and set this definition in my consumer project with `add_definitions(-DPOCO_LOG_DEBUG)`...

I was probably too fast creating that PR, because we fixed a similar issue with `POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX` and this definitely affects the generated library. So I've the following question: 

Should I abandon this pull request and create a new one for switching off `POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX`?
Or do you think it won't get merged anyways because it's not needed by most of the users to disable the debug-suffix?

Thanks for your help!

uilianries: You can keep this PR, but it should not affect the package ID at all. It can set a definition in cpp_info, like `self.cpp_info.defines.append('POCO_LOG_DEBUG=1')`

steinerthomas: Sorry for the late response, I've applied your suggestions!
The cpp_info is already there: `self.cpp_info.components["poco_foundation"].defines.append("POCO_LOG_DEBUG")`

uilianries (dismissed):
LGTM

ericLemanissier (approved):

uilianries (approved):

jwillikers (approved):

recipes/poco/all/conanfile.py
steinerthomas: > I was probably too fast creating that PR, because we fixed a similar issue with `POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX` and this definitely affects the generated library. So I've the following question:
> 
> Should I abandon this pull request and create a new one for switching off `POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX`? Or do you think it won't get merged anyways because it's not needed by most of the users to disable the debug-suffix?

#22807
Now the `POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX` issue is also resolved!

PR #19488, opened 2023-08-29
libvault: fix import of Version + decrease min Visual Studio Version + many small fixes

SpaceIm:
- fix import of Version (`from conan import Version` was illegal and breaks in conan 2.0.10) 
- use `export_conandata_patches()`
- remove useless `VirtualBuildEnv` generator (there is no build requirements in this recipe
- bump libcurl
- remove useless catch2 in requirements (it's used for tests, but tests are disabled...)
- add package_type
- fix min Visual Studio (it was ... 19, which doesn't make sense) and add min version for msvc
- use self.settings instead of self.info.settings in validate()
- remove useless CMakeLists wrapper (it had old conan v1 stuff with conan_basic_setup() and was not used at all...)
- add a layout
- add libm to system libs on Linux & FreeBSD only

---


SpaceIm:
Last PR of libvault is https://github.com/conan-io/conan-center-index/pull/14498, far before v2 pipeline. There might be more fixes needed to pass this pipeline.

SpaceIm:
libvault shared, msvc:

```
ERROR: libvault/0.52.0 (test package): Error in test() method, line 19
	self.run(bin_path, run_environment=True)
	ConanException: Error 3221225785 while executing bin\test_package
```

Ok so the problem on Windows is that there is a `Vault.dll` system library in System32 (found first by runtime loader), and libvault creates `vault.dll` file, it's an unfortunate name.

SpaceIm:
> libvault shared, msvc:
> 
> ```
> ERROR: libvault/0.52.0 (test package): Error in test() method, line 19
> 	self.run(bin_path, run_environment=True)
> 	ConanException: Error 3221225785 while executing bin\test_package
> ```
> 
> Ok so the problem on Windows is that there is a `Vault.dll` system library in System32 (found first by runtime loader), and libvault creates `vault.dll` file, it's an unfortunate name.

https://github.com/abedra/libvault/issues/116

recipes/libvault/all/patches/0.48.0-0001-fix-cmake.patch
```diff
@@ -0,0 +1,42 @@
...
+ include_directories("${PROJECT_SOURCE_DIR}/lib")
+@@ -120,9 +111,11 @@ set_target_properties(vault PROPERTIES
+ 
+ target_include_directories(vault PRIVATE src)
+ 
+-if(LINK_CURL)
+-  target_link_libraries(vault curl)
+-endif(LINK_CURL)
++target_link_libraries(vault PUBLIC CURL::libcurl)
++target_compile_features(vault PUBLIC cxx_std_17)
```
RubenRBS, line 36: As this is for `vault`, does this not just override whatever I set in my `compiler.cppstd` setting?

RubenRBS (commented):
Thanks! One question about one of the lines of the CMake patch, otherwise looks nice

Thanks a lot as well for the detailed PR message, helps a lot :)

recipes/libvault/all/patches/0.48.0-0001-fix-cmake.patch
SpaceIm: `CMAKE_CXX_STANDARD` injected by `CMakeToolchain` when `compiler.cppstd` is set has precedence over cxx_std compile feature.

`compiler.cppstd` is not honored when upstream CMakeLists hardcode `CMAKE_CXX_STANDARD`. The modern approach for CMake is "compile features" in libraries to express requirements, and CMAKE_CXX_STANDARD externally to force a specific standard if needed.

SpaceIm:
Is there still anything preventing this PR to be accepted and merged?

SpaceIm:
@RubenRBS @uilianries could you review this one please?

recipes/libvault/all/patches/0.48.0-0001-fix-cmake.patch
uilianries: Thank you for your explanation, it could be pushed to the upstream too.

SpaceIm:
@RubenRBS @uilianries could you accept this PR please? It's been 3 months :(

RubenRBS (approved):

PR #17894, opened 2023-06-10
librdkafka: fix Visual Studio Debug ssl >= 1.8.0

sourcedelica:
Specify library name and version:  **librdkafka/1.8.0**

Fixes #17879 
---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #19496, opened 2023-08-30
Librdkafka/openssl patch/all

KevDi:
Specify library name and version:  **librdkafka/all**

Starting from librdkafka Version 1.8.0 the following `pragma` calls for `openssl` where added [librdkafka/src/rdkafka_ssl.c](https://github.com/confluentinc/librdkafka/blob/49f180a36c247100dc246379ace84833b91f4038/src/rdkafka_ssl.c#L43C1-L44C35)

```
#pragma comment(lib, "libcrypto.lib")
#pragma comment(lib, "libssl.lib")
```
On a Debug Build this causes problems, because Openssl builds the Debug Libraries with an additional `d` (`libcryptod.lib/libssld.lib`). Linking the Program then fails because it won't find the correct libraries.

Adding these with the correct ending fixes this.
I also opened a PR on the Librdkafka Project with those changes but i'm not sure if they will be added. 

---


KevDi:
As @ericLemanissierBot mentioned there is another PR fixing this Problem. The question is what would be the best way? Adding the additional Pragmas or just replacing them like in the PR #17894

KevDi:
@RubenRBS 👍 and is this fix as it is ok or is the one mentioned above where the replace is done in python a better solution? I also made a pr at the librdkafka project with those changes so maybe if it gets accepted this patches are not needed then anymore. But we will see 😃☺️

RubenRBS:
> @RubenRBS 👍 and is this fix as it is ok or is the one mentioned above where the replace is done in python a better solution? I also made a pr at the librdkafka project with those changes so maybe if it gets accepted this patches are not needed then anymore. But we will see 😃☺️

Oh my @KevDi sorry about the radio silence, I totally dropped the ball on this one and did not see the mention notification until now, so I didn't circle back to the PR until now

Do you have a link to the PR you mention you submitted upstream? Having backport links usually helps a lot for future maintenance

KevDi:
@RubenRBS this is the other PR which deals with this problem #17894

KevDi:
Okay sorry you mean the PR on the librdkafka Project. Here it is: [Added Pragmas for the Debug Libs from Openssl](https://github.com/confluentinc/librdkafka/pull/4408)

KevDi:
not needed anymore

PR #19497, opened 2023-08-30
pkgconf: less verbose package_info() + bump meson

SpaceIm:
Printing env vars in `package_info()` is useless (why not printing other informations like lib name etc for example? It's not the job of the recipe to print these informations, they are available through other way provided by conan) and can be very verbose during conan install because this method can be called multiple consecutive times (I have a conanfile where it's called 13 consecutive times for pkgconf/1.9.3 and 7 times for pkgconf/1.9.5...).

---

### DIFF
--- a/recipes/pkgconf/all/conanfile.py
+++ b/recipes/pkgconf/all/conanfile.py
@@ -52,7 +52,7 @@ class PkgConfConan(ConanFile):
         self.settings.rm_safe("compiler.cppstd")
 
     def build_requirements(self):
-        self.tool_requires("meson/1.0.0")
+        self.tool_requires("meson/1.2.1")
 
     def source(self):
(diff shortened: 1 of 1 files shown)


ericLemanissier (approved):

toge (approved):

jwillikers (approved):

RubenRBS (commented):
Quick question: Why is the enable_lib check in package_info removed? I'm way less knowledgeable about old v1 so I might be missing something super obvious, sorry!

SpaceIm:
> Quick question: Why is the enable_lib check in package_info removed? I'm way less knowledgeable about old v1 so I might be missing something super obvious, sorry!

legacy `self.cpp_info.names["pkg_config"]` inherits from pkg_config_name property, so it's useless.

RubenRBS (approved):

ericLemanissier (approved):

PR #19509, opened 2023-08-30
tre: add recipe

valgur:
Specify library name and version:  **tre/cci.20230717**: https://github.com/laurikari/tre

Did not include the last official release v0.8.0, since it's 14 years old and there have been minor improvements since without a version bump.

[Packaging status](https://repology.org/project/tre/versions)

### DIFF
--- a/recipes/tre/all/conanfile.py
+++ b/recipes/tre/all/conanfile.py
@@ -1,14 +1,12 @@
 import os
 
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import fix_apple_shared_install_name
 from conan.tools.build import cross_building
-from conan.tools.cmake import cmake_layout
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout
 from conan.tools.env import VirtualBuildEnv, VirtualRunEnv
-from conan.tools.files import copy, get, replace_in_file, rm, rmdir
+from conan.tools.files import copy, get, rm, rmdir
 from conan.tools.gnu import Autotools, AutotoolsToolchain
-from conan.tools.microsoft import is_msvc
 
 required_conan_version = ">=1.53.0"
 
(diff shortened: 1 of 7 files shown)


recipes/tre/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+        env.generate()
+        if not cross_building(self):
+            env = VirtualRunEnv(self)
+            env.generate(scope="build")
+        tc = AutotoolsToolchain(self)
+        tc.generate()
+
+    def _patch_sources(self):
+        if self.settings.os == "Windows" and not is_msvc(self):
+            replace_in_file(self, os.path.join(self.source_folder, "win32", "tre.def"), "tre.dll", "libtre.dll")
```
RubenRBS, line 63: So this can't actually happen right now right? The validate() method above removes any Windows configurations. Do you expect then that once it works this would be needed?

recipes/tre/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.os == "Windows":
+            raise ConanInvalidConfiguration("Windows builds are not supported yet")
```
RubenRBS, line 44: Conan or recipe limitation?

recipes/tre/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        env = VirtualBuildEnv(self)
+        env.generate()
+        if not cross_building(self):
+            env = VirtualRunEnv(self)
+            env.generate(scope="build")
```
RubenRBS, line 57: Just a heads-up: Afaik this bit is only needed when the recipe has direct _host_ requirements so could be dropped, but leaving it as-is is ok too :)

RubenRBS (commented):
Thanks! I have a question about the validate() method, but other than that, looks good :)

recipes/tre/all/conanfile.py
valgur: I would like to add Windows support for both this and the related libmagic recipe. tre needs some minor fixes in Windows, but I would like to merge the simpler version wihout it first, so the dependent packages can at least be developed further.
This bit of logic is a carry-over from the Vcpkg port I based this recipe on.

franramirez688 (approved):

RubenRBS (approved):
Thanks!

PR #19522, opened 2023-08-31
[docs] Change conan v2 feedback message

uilianries:
Following Conan 2.x migration schedule. From now, all PRs should pass by Conan 2.x CI build.

More information in https://github.com/conan-io/conan-center-index/discussions/19104

Most of recipes are ported now to Conan 2.x, and the recommendation for those that are still missing is pinging us in the PR, so we can help to migrate it too.

This PR only updates the feedback message, not the CI behavior. The behavior should be done in a separated PR.

/cc @jcar87 @danimtb @RubenRBS 

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #19523, opened 2023-08-31
[config] Conan v2 ready references is no longer required

uilianries:
This PR is the second part of #19522 

It changes the need of a temporary list of packages when doing a regression test with those packages already working with Conan v2.

It's not possible to merge it right now, first we will need to do some changes in C3I_Jenkins, as it may be affected by a current running feature. 

The current reference list will be maintained, so people will have a reference about those package ported to Conan v2 already.

---


uilianries:
Closing this PR as we will keep the list, so anyone will be able to know all packages that are ported to Conan 2.x already.

danimtb (approved):

RubenRBS (approved):

PR #19533, opened 2023-08-31
Add a recipe for nghttp3

cjbradfield:
Specify library name and version:  **libnghttp3/0.15.0**

Libnghttp3 is an http3 implementation from the authors of nghttp2.

---


cjbradfield:
The v2 pipeline is failing due to some bug in the windows build image. I'll need an admin to rerun this once the builder is fixed.

recipes/libnghttp3/all/conanfile.py
```diff
@@ -0,0 +1,88 @@
...
+from conan.tools.microsoft import is_msvc
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class Nghttp3Conan(ConanFile):
+    name = "libnghttp3"
```
uilianries, line 16:
```suggestion
    name = "nghttp3"
```
Please, keep the project's name to be consistent. That means you will need to change the folder name too.

recipes/libnghttp3/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "0.14.0":
```
uilianries, line 2: Is it possible to update to 0.15.0 ? 

https://github.com/ngtcp2/nghttp3/tags

recipes/libnghttp3/all/conanfile.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def package(self):
+        copy(self, pattern="COPYING", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+
+    def package_info(self):
+        self.cpp_info.components["nghttp3"].set_property("pkg_config_name", "libnghttp3")
```
uilianries, line 82: Looking this cpp_info, you should not need a separated component. What situation are you trying deal with?

recipes/libnghttp3/all/conanfile.py
```diff
@@ -0,0 +1,88 @@
...
+        if is_apple_os(self):
+            # workaround for: install TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
+            tc.cache_variables["CMAKE_MACOSX_BUNDLE"] = False
+        tc.generate()
+
+        tc = CMakeDeps(self)
+        tc.generate()
+        tc = PkgConfigDeps(self)
+        tc.generate()
+
```
uilianries, line 68:
```suggestion
```
I see no requirement listed. The tool requirement for cmake should be solved by VirtualBuildEnv

recipes/libnghttp3/all/conanfile.py
```diff
@@ -0,0 +1,88 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "6":
```
uilianries, line 48: please, in case this project needs C++14, use the template for validation:

https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L78

recipes/libnghttp3/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,21 @@
...
+    const nghttp3_info* info = nghttp3_version(NGHTTP3_VERSION_NUM);
+    if (info) {
+        printf("nghttp3 ver=%d version=%s\n", info->version_num, info->version_str);
+    } else {
+        printf("nghttp3: cannot get version\n");
+    }
+    return 0;
+}
+
+// vim: et ts=4 sw=4
```
uilianries, line 21:
```suggestion
```

Please, avoid passing editor configuration

recipes/libnghttp3/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,17 @@
+from conans import ConanFile, CMake, tools
```
uilianries, line 1: Please, remove the entire test_v1_package folder, it's no longer needed: https://github.com/conan-io/conan-center-index/discussions/17406

uilianries (commented):
Thank you for your contribution

recipes/libnghttp3/all/test_v1_package/conanfile.py
cjbradfield: Thanks for the review! I used libnghttp2 as a template for this recipe since both projects have the same authors and fall under the same administrative umbrella. Many of the issues you have pointed out appear in that recipe as well.

I'll post and update with these changes shortly!

recipes/libnghttp3/all/conanfile.py
cjbradfield: This came from libnghttp2s recipe. I'll remove it.

recipes/libnghttp3/all/conanfile.py
cjbradfield: This is a C project. It shouldn't have any build dependency on C++14.

recipes/libnghttp3/all/conandata.yml
cjbradfield: I'll bump the version. It was released while this was awaiting review.

recipes/libnghttp3/all/conanfile.py
cjbradfield: I'm going to make it consistent with libnghttp2 by changing everything to libnghttp3 and removing nghttp3 references. Libnghttp2 has these inconsistencies, but at the end of the day, that recipe lives in a folder of that name and produces a package of that name.

recipes/libnghttp3/all/conanfile.py
cjbradfield: Could you please elaborate on this comment? I don't understand how that applies to the code suggestion.

For CMake recipes, as I understand it, it is common to run the CMakeToolchain generator as well as the CMakeDeps generator. This project also has pkg-config support and that generator is run as well.

recipes/libnghttp3/all/conanfile.py
cjbradfield: I guess I should say the qpack examples require C++-17; however, this recipe explicitly disables the examples.

recipes/libnghttp3/all/test_package/test_package.cpp
cjbradfield: removed

cjbradfield:
Thanks for the review! I used libnghttp2 as a template for this recipe since both projects have the same authors and fall under the same administrative umbrella. Many of the issues you have pointed out appear in that recipe as well.

I've posted the requested changes and posted follow-up questions as necessary.

recipes/libnghttp3/all/conanfile.py
uilianries: The libnghttp2 was not well reviewed and was long time ago, it should be `nghttp2` instead. In the very past we had a wrong idea of creating separated packages for binaries and libraries. 

As you can see, the repology website finds many references to https://repology.org/project/nghttp2/information and same for nghttp3.

Please, let's do it right now, rename it `nghttp3`.

recipes/libnghttp3/all/conanfile.py
cjbradfield: I believe I've addressed this correctly. Let me know if I have not.

recipes/libnghttp3/all/conanfile.py
uilianries: So why does it need GCC 6? Does it use C-11 ?

recipes/nghttp3/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+        tc = CMakeToolchain(self)
+        tc.variables["ENABLE_SHARED_LIB"] = self.options.shared
+        tc.variables["ENABLE_STATIC_LIB"] = not self.options.shared
+        tc.variables["ENABLE_LIB_ONLY"] = True
+        if is_apple_os(self):
+            # workaround for: install TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
+            tc.cache_variables["CMAKE_MACOSX_BUNDLE"] = False
+        tc.generate()
+        tc = CMakeDeps(self)
+        tc.generate()
```
uilianries, line 64:
```suggestion
```

The CMakeDeps is useful to generate `Findxxx.cmake` for those packages listed in `requirements()` method. Here, in `nghttp3`, it does not require dependencies, so `CMakeDeps` can be disposed, because is not used.

But what about `cmake` as dependency? CMake will be solved via PATH environment variable, configured in `conanbuildenv.sh`, but it's generated by `VirtualBuildEnv`.

recipes/nghttp3/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import is_apple_os
+from conan.tools.cmake import CMake, CMakeToolchain, CMakeDeps, cmake_layout
```
uilianries, line 4:
```suggestion
from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
```

CMakeDeps is not needed because does not require an external dependency.

uilianries (commented):
Only a small comment about CMakeDeps not being needed. But it's looking really good!

recipes/libnghttp3/all/conanfile.py
cjbradfield: Since their build files were ported from ngtcp2 (which were ported from nghttp2), I assumed this needed to stay the same as nhgttp2. I can remove it and see if it breaks.

recipes/nghttp3/all/conanfile.py
cjbradfield: Got it. I'll remove CMakeDeps.

uilianries (approved):
LGTM

jwillikers (approved):

PR #19549, opened 2023-09-01
opencolorio: add version 2.3.0, update dependencies

toge:
Specify library name and version:  **opencolorio/***

---

### DIFF
--- a/recipes/opencolorio/all/conandata.yml
+++ b/recipes/opencolorio/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "2.3.0":
+    url: "https://github.com/AcademySoftwareFoundation/OpenColorIO/archive/v2.3.0.tar.gz"
+    sha256: "32b7be676c110d849a77886d8a409159f0367309b2b2f5dae5aa0c38f42b445a"
   "2.2.1":
     url: "https://github.com/AcademySoftwareFoundation/OpenColorIO/archive/v2.2.1.tar.gz"
     sha256: "36f27c5887fc4e5c241805c29b8b8e68725aa05520bcaa7c7ec84c0422b8580e"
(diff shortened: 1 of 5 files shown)


recipes/opencolorio/all/patches/2.2.1-0001-fix-cmake-source-dir-and-targets.patch
```diff
@@ -12,15 +12,24 @@ index 17e188d..91af0ec 100755
...
-index 5455a08..eb91a37 100644
+index 5455a08..420fbd4 100644
 --- a/share/cmake/modules/FindExtPackages.cmake
 +++ b/share/cmake/modules/FindExtPackages.cmake
+@@ -41,7 +41,7 @@ find_package(expat 2.4.1 REQUIRED)
+ 
+ # yaml-cpp
+ # https://github.com/jbeder/yaml-cpp
+-find_package(yaml-cpp 0.7.0 REQUIRED)
++find_package(yaml-cpp 0.8.0 REQUIRED)
```
irieger, line 23:
```suggestion
+find_package(yaml-cpp REQUIRED)
```
I have never used the version parameter in cmake myself, but does it make sense to fix the version here? Maybe someone has a reason to overwrite the version of yaml-cpp etc. and as long as there is no hard reason I'd personally not restrict it to a fixed version maybe.

recipes/opencolorio/all/patches/2.2.1-0001-fix-cmake-source-dir-and-targets.patch
```diff
@@ -12,15 +12,24 @@ index 17e188d..91af0ec 100755
...
+ 
+ # pystring
+ # https://github.com/imageworks/pystring
 @@ -138,7 +138,7 @@ endif()
  
  # minizip-ng
  # https://github.com/zlib-ng/minizip-ng
 -find_package(minizip-ng 3.0.7 REQUIRED)
-+find_package(minizip 3.0.7 REQUIRED)
++find_package(minizip 3.0.9 REQUIRED)
```
irieger, line 32:
```suggestion
+find_package(minizip REQUIRED)
```
As above.

irieger (commented):
As I'm new to this but thought I'd use a pull request that is changing something I'm using, here is my two cents. Feel free to ignore this as I'm not fully understand, was just wondering why version are fixed in the CMake files anyway.

Also couldn't fully understand how approvals are handled here from reading the contribution guides and just got access today as I started working on an upgrade for OpenImageIO on the weekend. To me the changes look sensible and build locally also. But From looking at it and as a user of OCIO this looks good to me and I'd approve it.

recipes/opencolorio/all/patches/2.2.1-0001-fix-cmake-source-dir-and-targets.patch
toge: @irieger 
I'm sorry for missing your comments.

PR #19554, opened 2023-09-01
Fix constexpr on clang for eastl

Kalixio:
Specify library name and version: eastl/3.21.12

When building with clang 16, it was not possible to use constexpr with eastl
This is due to GNUC being defined on clang and EA_COMPILER_VERSION being less than 9000

With clang, EA_COMPILER_VERSION is defined to:

`#define EA_COMPILER_VERSION (__clang_major__ * 100 + __clang_minor__)`

see https://github.com/electronicarts/EABase/blob/521cb053d9320636f53226ffc616216cf532f0ef/include/Common/EABase/config/eacompiler.h#L341C53-L341C53

I propose to exclude clang from

```
	#elif defined(__GNUC__) && (EA_COMPILER_VERSION < 9000)   // Before GCC 9.0
		#define EA_CPP14_CONSTEXPR  // not supported
		#define EA_NO_CPP14_CONSTEXPR 
```

Since it only applies to gcc before 9.0, according to the comment

Also, added a constexpr test case in the test package

Tested locally on MSVC 19.37.32822.0
Tested locally on Clang 16

---


RubenRBS:
Hi @Kalixio thanks a lot for your contribution. Given that this is a patch with uncertain implications, I think it'd be better if a PR was submitted upstream for this code. Once (if) it gets merged, as EASTL's releases are sporadic at best, we can backport the fix. 

Can you open the PR in EA's repo and link it here? I'm sure they'll appreciate it :)

Kalixio:
We applied this patch in  #3643 because they updated the upstream to not use this code anymore and only rely on EA_COMPILER_CPP14_ENABLED (https://github.com/electronicarts/EASTL/blob/05f4b4aef33f2f3ded08f19fa97f5a27ff35ff9f/include/EASTL/internal/config.h#L146), which broke constexpr on msvc 14

We reverted that change with this patch and notified the upstream of the issue (https://github.com/electronicarts/EASTL/issues/402) but no changes or update since then.

Genuinely curious: since we're the ones that created and applied this patch to revert the changes in the upstream, shouldn't we maintain it as well ? (although I understand this is not ideal...)

PR #19556, opened 2023-09-02
luau: add version 0.593, remove older versions, make static-library

toge:
Specify library name and version:  **luau/***

---


recipes/luau/all/conanfile.py
```diff
@@ -17,18 +16,14 @@ class LuauConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://luau-lang.org/"
     topics = ("lua", "scripting", "typed", "embed")
-    package_type = "library"
+    package_type = "static-library"
```
RubenRBS, line 19: When making these changes, it's good if the PR message contains the rationale behind it to save some reviewing time :)

Since when is the library static-only? Was this an initial recipe bug or was support just dropped?

RubenRBS (commented):
Thanks! A question about the rationale to drop shared (Not that I don't trust you! It's just better to have it written down where possible :))

recipes/luau/all/conanfile.py
toge: @RubenRBS 
Sorry for the lack of information.
When I first wrote the recipe, I had created a patch to allow shared builds as well.
My goal was to make it versatile.

However, now that I think about it, this versatility is to be implemented for upstream, and at this time when upstream is restricting itself to static builds, I thought the conan recipe should also provide only static builds.

danimtb (approved):
I am ok with the changes! In case someone ask for this library to support shared, better ask for support upstream. Thank you!

PR #19571, opened 2023-09-03
crowcpp-crow: update dependencies, fix requires names

toge:
Specify library name and version:  **crowcpp-crow/***

---

### DIFF
--- a/recipes/crowcpp-crow/all/conanfile.py
+++ b/recipes/crowcpp-crow/all/conanfile.py
@@ -47,7 +47,7 @@ class CrowConan(ConanFile):
             self.requires("openssl/[>=1.1 <4]")
         if Version(self.version) >= "1.0":
             if self.options.with_ssl:
-                self.requires("openssl/1.1.1v")
+                self.requires("openssl/[>=1.1 <3]")
             if self.options.with_compression:
                 self.requires("zlib/1.2.13")
 
--- a/recipes/crowcpp-crow/all/conanfile.py
+++ b/recipes/crowcpp-crow/all/conanfile.py
@@ -42,12 +42,12 @@ class CrowConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.81.0")
+        self.requires("boost/1.83.0")
         if self.version == "0.2":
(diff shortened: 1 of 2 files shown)


recipes/crowcpp-crow/all/conanfile.py
```diff
@@ -42,12 +42,12 @@ def layout(self):
...
     def requirements(self):
-        self.requires("boost/1.81.0")
+        self.requires("boost/1.83.0")
         if self.version == "0.2":
-            self.requires("openssl/1.1.1s")
+            self.requires("openssl/[>=1.1 <4]")
         if Version(self.version) >= "1.0":
             if self.options.with_ssl:
-                self.requires("openssl/1.1.1s")
+                self.requires("openssl/1.1.1v")
```
RubenRBS, line 50: Why does 0.2 allow for version range but no >=1.0?

RubenRBS (commented):
Thanks! Question about the openssl range :)

recipes/crowcpp-crow/all/conanfile.py
toge: @RubenRBS 
I met following execution error with OpenSSL/3.x.x:
```
terminate called after throwing an instance of 'boost::wrapexcept<boost::system::system_error>'
  what():  context: unknown option (digital envelope routines) [asio.ssl:50331817]
```

It seems to be caused by OpenSSL/3.x.x providers.
https://wiki.openssl.org/index.php/OpenSSL_3.0#Providers

uilianries: @toge you still can use ranges, but in that case: `openssl/[>=1.1 <3]`

uilianries, line 50:
```suggestion
                self.requires("openssl/[>=1.1 <3]")
```

toge: @uilianries 
Thanks a lot!

RubenRBS: Weird that an old version supports openssl 3, but newer ones dont!

RubenRBS (approved):

PR #19578, opened 2023-09-03
make : Add alternative mirror.

p-groarke:
Specify library name and version:  **make/\***

https://ftpmirror.gnu.org is down, preventing the build of libraries dependent on `make`.

As per loosely this thread https://github.com/conan-io/conan/issues/6349 and this pr https://github.com/conan-io/conan-center-index/pull/632, add a mirror.

---


recipes/make/all/conandata.yml
```diff
@@ -1,10 +1,14 @@
 sources:
   "4.3":
-    url: "http://ftpmirror.gnu.org/gnu/make/make-4.3.tar.gz"
-    sha256: "e05fdde47c5f7ca45cb697e973894ff4f5d79e13b750ed57d7b66d8defc78e19"
+    - url: "https://ftpmirror.gnu.org/gnu/make/make-4.3.tar.gz"
```
p-groarke, line 3: https seems like the better choice

recipes/make/all/conandata.yml
```diff
@@ -1,10 +1,14 @@
...
+      sha256: "e05fdde47c5f7ca45cb697e973894ff4f5d79e13b750ed57d7b66d8defc78e19"
+    - url: "https://ftp.gnu.org/gnu/make/make-4.3.tar.gz"
+      sha256: "e05fdde47c5f7ca45cb697e973894ff4f5d79e13b750ed57d7b66d8defc78e19"
```
p-groarke, line 6: same file same sha

recipes/make/all/conanfile.py
```diff
@@ -38,8 +39,19 @@ def package_id(self):
...
     def source(self):
-        get(self, **self.conan_data["sources"][self.version],
-            destination=self.source_folder, strip_root=True)
+        for data in self.conan_data["sources"][self.version]:
+            url, sha = data.values()
+            filename = url[url.rfind("/")+1:]
+            try:
+                download(self, url, filename, sha256=sha, retry=0)
+            except ConanException as e:
+                if "NewConnectionError" in str(e):
```
p-groarke, line 48: I don't know how / if you can filter exceptions in python so I used this. Better than nothing, lmk if there's a better way.

recipes/make/all/conandata.yml
```diff
@@ -1,12 +1,20 @@
...
-    url: "http://ftpmirror.gnu.org/gnu/make/make-4.2.1.tar.bz2"
-    sha256: "d6e262bf3601b42d2b1e4ef8310029e1dcf20083c5446b4b7aa67081fdffc589"
+    - url: "https://ftpmirror.gnu.org/gnu/make/make-4.2.1.tar.bz2"
+      sha256: "d6e262bf3601b42d2b1e4ef8310029e1dcf20083c5446b4b7aa67081fdffc589"
+    - url: "https://ftp.gnu.org/gnu/make/make-4.2.1.tar.bz2"
+      sha256: "d6e262bf3601b42d2b1e4ef8310029e1dcf20083c5446b4b7aa67081fdffc589"
 patches:
   "4.3":
     - patch_file: "patches/4.3-0001-clang.patch"
+      patch_description: "Fix __alloca define check."
```
p-groarke, line 15: Suppress schema warning to the best of my abilities.

recipes/make/all/conanfile.py
RubenRBS: There's no need for this. The `get`/`download` functions should already be able to handle mirrors

RubenRBS (changes requested):
Thanks a lot!
The mirror URLs are ok, but the changes to the recipes seem unnecessary :)

recipes/make/all/conandata.yml
RubenRBS: Thanks! We don't enforce this for PRs that do not directly add the patch, but it's unlucky that the linter still complains for every other case

recipes/make/all/conanfile.py
RubenRBS: Or does this come from the current get implementation not handling this failing mode properly?

p-groarke: @RubenRBS Could you point to a recent example? All my research seemed to indicate mirrors coming from conandata.yml needed some custom logic in the recipe, but a lot of the info was old (google is bad nowadayts). If it doesn't that's great :)

p-groarke: Roger :] I'll ping you re-review when I have a some time to teak

danimtb (approved):

SSE4 (approved):

jwillikers (approved):

PR #19588, opened 2023-09-04
[docs] Add `zlib` and `libcurl` to the list of allowed version ranges

RubenRBS:
This PR adds `zlib` and `libcurl` as part of the allowed ones to be declared with version ranges in recipe's dependencies.

The CCI maintainers will create the necessary PRs to change the requirements so as to not miss any :)

Zlib: 
For libcurl: https://curl.se/libcurl/abi.html

We're also discussing if allowing `sqlite` too

### DIFF
--- a/docs/adding_packages/dependencies.md
+++ b/docs/adding_packages/dependencies.md
@@ -174,15 +174,18 @@ for consumer, we do impose some limits on Conan features to provide a smoother f
 * [`python_requires`](https://docs.conan.io/1/reference/conanfile/other.html#python-requires) are not allowed.
 
 ### Version Ranges
- 
+
 Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/2/tutorial/versioning/version_ranges.html).
 With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful of dependencies.
 Currently, these are:
 
 * OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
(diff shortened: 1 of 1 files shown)


docs/adding_packages/dependencies.md
```diff
@@ -174,15 +174,18 @@ for consumer, we do impose some limits on Conan features to provide a smoother f
...
- 
+
 Version ranges are a useful Conan feature, [documentation here](https://docs.conan.io/2/tutorial/versioning/version_ranges.html).
 With the introduction of Conan 2.0, we are currently working to allow the use of version ranges and are allowing this for a handful of dependencies.
 Currently, these are:
 
 * OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
-* CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts.
+* CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts
+* Zlib: `[>=2.X.Y <3]`, where `2.X.Y` is the minimum version of Zlib required, starting from `2.1.10`
```
uilianries, line 184: But zlib is still 1.x! 🤔

danimtb (approved):

prince-chrismc (approved):

prince-chrismc:
It's a touch confusing because all 3 for different numbering formats.

Some with 1 `Y` Other with 2 and the places are not the same ... 

Well need to bump the docs if libcurl 10 comes out


Perhaps just saying is allow with a min and max known major is sufficient?

PR #19593, opened 2023-09-04
libdwarf: add version 0.7.0, fix missing header in v0.5.0

### LINKED ISSUE #17403: [package] libdwarf/0.5.0: dwarf.h missing from package
### Description

dwarf.h is not present in the v0.5.0 package. The headers defines all DWARF types and is needed to use the library. 

The update to v0.5.0[ change a test](https://github.com/conan-io/conan-center-index/blob/7581062d7195d756a82c9e1adb136709c777bff6/recipes/libdwarf/all/test_package/test_package.c#L6-L11) to only check for the libdwarf,h. There's is no mention in libdwarf about dropping dwarf.h from consumer applications, it even states the [opposite](https://github.com/davea42/libdwarf-code/blob/81e5968d506aa50562aa3df64783f030d9e316cc/src/lib/libdwarf/libdwarf.h#L35-L39).

The issue comes from using cmake install. libdwarf is not installing the header file (probably a bug in the CMakeLists, because it tries to copy [all header files](https://github.com/davea42/libdwarf-code/blob/81e5968d506aa50562aa3df64783f030d9e316cc/src/lib/libdwarf/CMakeLists.txt#L126-L127)). It only installs the libdwarf.h.

We can solve the issue by copying the file from conan or applying a patch to libdwarf´s CMakeLists to fix the directory install line and actually export all the headers.

* Package Name/Version: **libdwarf/0.5.0**
* Compiler+version: **Visual Studio 2022**
* Conan version: **conan 1.59.0**
* Python version: **Python 3.10.2**

### Conan profile

[settings]
os=Windows
os_build=Windows
arch=x86
arch_build=x86_64
compiler=Visual Studio
compiler.version=17
build_type=RelWithDebInfo
compiler.cppstd=20
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install libdwarf/0.5.0@ 

Try using the dwarf.h

### Logs

```

```

</details>

cesenaLA:
Specify library name and version:  **libdwarf/0.7.0**

Updated CMake patch to also install dwarf.h. The patch was already inplace for version v20191104 but was missed when the recipe got updated to v0.5.0.

Also remove a dependency on libelf from any new version if `with_dwarfgen=False`. libdwarf does not depend in libelf since version 0.1.0.

Added version 0.7.0.

---

Fixes #17403


cesenaLA:
The failures in conan v1 seems to be related to make the `find_package(libelf)` optional in the CMakeList https://github.com/conan-io/conan-center-index/pull/19593/commits/775c32c7810cab252a72d4686c689918908527dc. Should that be removed entirely by the patch file?

recipes/libdwarf/all/conanfile.py
```diff
@@ -47,7 +47,8 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("libelf/0.8.13")
+        if self.options.with_dwarfgen or self.version == "20191104":
```
uilianries, line 50: That's bad. The options `with_dwarfgen` should be `True` by default:

CMakeLists.txt:203

    option(DWARF_WITH_LIBELF "Use libelf (default is YES)" TRUE)


I would suggest changing the default value for `with_dwarfgen` to `True`. I may break some users, but then we will follow the correct behavior.

cesenaLA: > The options `with_dwarfgen` should be True by default

I guess I disagree, the option to enable `dwarfgen` is False by default in the CMakeList:

CMakeLists.txt:353
```
option(BUILD_DWARFGEN "Build dwarfgen (default is NO)" FALSE)
```

dwarfgen is more a utility for testing dwarf-generation by libdwarf and as an example of how one uses libdwarf to generate DWARF. I don't think that's something you want to get by default if you only want to use the library.

I made the change based on the release notes from v0.1.0, you can check them here in the Releases Section https://www.prevanders.net/dwarf.html. `libelf` is only needed when building `dwarfgen`.

Also, recent changes upstream are removing the remaining references to `libelf`:
https://github.com/davea42/libdwarf-code/commit/77d9cad797963942497d004103b602d55f5a7882
https://github.com/davea42/libdwarf-code/commit/a5d44c690636aa9c49ba4c6656f042c05f42c7ed

uilianries: In that case, if you pass DWARF_WITH_LIBELF=False, would work for the current error, or do you need another patch?

cesenaLA: if you mean the build errors in this PR, I can change that second commit to not call `find_package(libelf REQUIRED)` instead of just dropping the `REQUIRED`

uilianries: >  I can change that second commit to not call find_package(libelf REQUIRED) instead of just dropping the REQUIRED

@cesenaLA It sounds perfect for the current situation. Could you please update it?

cesenaLA: I made the change, but it turns out that was not the issue. libdwarf started exporting the libdwarf.pc file in https://github.com/davea42/libdwarf-code/commit/7368a1062271e08e7df0f188d8b02c9991a0d03a. I removed installing that file to the patch.

cesenaLA:
Updated PR to fix conflicts.

@uilianries let me know if I need to make more changes to get this PR ready

uilianries (approved):
LGTM! Thank you @cesenaLA

jwillikers (approved):

uilianries:
The CI is lazy for some reason. I just restarted to Conan v2 pipeline. All reviews are okay, just waiting for CI result now.

PR #19596, opened 2023-09-05
add test for c_subscripts

AndreyMlashkin:
Specify library name and version:  **newmat/all**

---


RubenRBS:
Hi! Thanks for your contribution. Can you explain the reasonale behind changing this default? It would be helpful in order to be able to validate the change

AndreyMlashkin:
It is more or less a test if flag with_c_subscripts is relyably working. Because it does not on my end.

RubenRBS:
I've pushed a change to fix the URL and get the option back to its default, as unless there's a good reason, we normally don't differ from upstream defaults :)

RubenRBS (approved):
Great job catching the defines bug! :)

PR #19600, opened 2023-09-05
objectbox: add version 0.19.0, add package_type

toge:
Specify library name and version:  **objectbox/***

---

### DIFF
--- a/recipes/objectbox/all/conandata.yml
+++ b/recipes/objectbox/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.19.0":
+    url: "https://github.com/objectbox/objectbox-c/archive/v0.19.0.tar.gz"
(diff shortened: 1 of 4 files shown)


RubenRBS:
I found (And later saw that @SpaceIm mentioned it in the previous PR for this recipe) that this recipe might be distributing libs which the team and I are not sure if we have license/permissions to do so. While it _might_ be ok, I think a safer solution for now would be setting `upload_policy="skip"` in the recipe so binaries are not uploaded to the CCI central repo.

This will break any PR for packages that depend on this because the CI only builds the 1 recipe and does not do `--build=missing`, but at least users would still be able to use it locally. 

Wdyt? I've also asked the team for further input

jwillikers (approved):

uilianries (approved):
LGTM

PR #19606, opened 2023-09-05
[vulkan-validationlayers] Fix discovery of libVkLayer_khronos_validation

### LINKED ISSUE #19605: [bug] vulkan-validationlayers/1.3.239.0: Cannot load libVkLayer_khronos_validation.so at runtime
### Description

`libVkLayer_khronos_validation.so` is a runtime requirement of any application that consumes `vulkan-validationlayers`, which technically produces a static library (it produces both the static and dynamic libraries, but only the static library should be linked). `libVkLayer_khronos_validation.so` should not be linked with the consumer, but is instead dynamically loaded via `vulkan-loader`, which uses `$VK_LAYER_PATH` to read `libVkLayer_khronos_validation.json` to identify the library, which contains the path to `libVkLayer_khronos_validation.so`. This looks like the following:
```
{
    "file_format_version": "1.2.0",
    "layer": {
        "name": "VK_LAYER_KHRONOS_validation",
        "type": "GLOBAL",
        "library_path": "libVkLayer_khronos_validation.so",
        "api_version": "1.3.239",
...
```
Unfortunately, the recipe doesn't populate `LD_LIBRARY_PATH` on linux, and so when searching for `libVkLayer_khronos_validation.so`, the application can't find it.

There are a few potential fixes to this:

1. Manually add `self.libdirs` to `LD_LIBRARY_PATH`, `PATH`, `DYLD_LIBRARY_PATH`
2. Set `package_type="shared-library"`
3. Require consumers to depend on this with `self.requires("vulkan-validationlayers/1.3.239.0", run=True)`
4. Try to patch `libVkLayer_khronos_validation.json` after creation so that `library_path` contains a relative path to the `lib` directory in its own package

* Package Name/Version: **vulkan-validationlayers/1.3.239.0**
* Operating System+version: **Arch Linux**
* Compiler+version: **GCC 13.2**
* Docker image: N/A
* Conan version: **conan 2.0.10**
* Python version: **Python 3.11.1**


### Conan profile

Profile host:                                                                                                                               
[settings]                                                                                                                                  
arch=x86_64                                                                                                                                 
build_type=Release                                                                                                                          
compiler=gcc                                                                                                                                
compiler.cppstd=20                                                                                                                          
compiler.libcxx=libstdc++11                                                                                                                 
compiler.version=12                                                                                                                         
os=Linux                                                                                                                                    
&:build_type=Debug                                                                                                                          
[conf]                                                                                                                                      
tools.build:compiler_executables={'c': '/usr/bin/gcc', 'cpp': '/usr/bin/g++'}                                                               
                                                                                                                                            
Profile build:                                                                                                                              
[settings]                                                                                                                                  
arch=x86_64                                                                                                                                 
build_type=Release                                                                                                                          
compiler=clang                                                                                                                              
compiler.cppstd=20                                                                                                                          
compiler.libcxx=libstdc++11                                                                                                                 
compiler.version=15                                                                                                                         
os=Linux                                                                                                                                    
[conf]                                                                                                                                      
tools.build:compiler_executables={'c': '/usr/bin/clang', 'cpp': '/usr/bin/clang++'}

### Steps to reproduce

1. `git clone -b bug/vulkan-validationlayers https://github.com/samuel-emrys/vulkan-tutorial.git`
2. `conan build . -s "&:build_type=Debug"` # The build type is important here
3. `source build/Debug/generators/conanrun.sh`
4. `./build/Debug/src/vulkantutorial`
5. Observe the following:

```bash
$ ./build/Debug/src/vulkantutorial
validation layer: Searching for ICD drivers named libGLX_nvidia.so.0                                                                        
validation layer: Searching for ICD drivers named /usr/lib32/amdvlk32.so                                                                    
validation layer: Build ICD instance extension list                                                                                         
validation layer: libVkLayer_khronos_validation.so: cannot open shared object file: No such file or directory                               
validation layer: Requested layer VK_LAYER_KHRONOS_validation failed to load!                                                               
failed to create instance!  
```

### Logs

```
Put your log output here
```

</details>

samuel-emrys:
**Please read #19605 for a full description of the bug this PR solves**

Add libdirs to the runtime lib discovery path to enable `libVlLayer_khronos_validation.{so,dll,dylib}` to be discovered appropriately at runtime. This package exports both a static library and a shared library, of which, only the static library should be linked against. The shared library is loaded dynamically by vulkan-loader, and still needs to be discoverable. Because this is not a shared library, these paths are not exported automatically and therefore are required to be explicitly set.

I felt that option 1 was the most desirable solution here because:
* It doesn't introduce the unwanted semantics that setting `package_type="shared-library"` would have, and potential complications when people actually want to use this only as the static library that it is. Particularly as this relates to the difference in package_id implications
* It doesn't impose requirements on the user to depend on this library differently by manually specifying `run=True` in the `requires()` call
* It doesn't require patching of artifacts

Additional discussion available in https://cpplang.slack.com/archives/C41CWV9HA/p1693906666122339

Closes #19605

---


samuel-emrys:
// cc @ericLemanissier @SpaceIm

recipes/vulkan-validationlayers/all/conanfile.py
```diff
@@ -225,5 +225,18 @@ def package_info(self):
...
+        # Update runtime discovery paths to allow libVkLayer_khronos_validation.{so,dll,dylib} to be discovered
+        # and loaded by vulkan-loader when the consumer executes
+        # This is necessary because this package exports a static lib to link against and a dynamic lib to load at runtime
+        runtime_lib_discovery_path = "LD_LIBRARY_PATH"
+        if self.settings.os == "Windows":
+            runtime_lib_discovery_path = "PATH"
+        if self.settings.os == "Macos":
+            runtime_lib_discovery_path = "DYLD_LIBRARY_PATH"
+        for libdir in [os.path.join(self.package_folder, libdir) for libdir in self.cpp_info.libdirs]:
+            self.output.info(f"Prepending {libdir} to {runtime_lib_discovery_path}")
```
SpaceIm, line 238:
```suggestion
```

I would remove these messages, it can be called many times. It should be the job of VirtualRunEnv to display something on demand.

SpaceIm (approved):

SSE4 (approved):

jwillikers (approved):

StellaSmith (approved):

samuel-emrys:
@RubenRBS how's this one going? Last update you provided you were going to check with the team that this was the best way forward

franramirez688 (approved):

RubenRBS (approved):
Thanks a lot for your patience. After looking into it with the team, we've seen that while Conan lacks some built-in support for cases like these (See https://github.com/conan-io/conan/issues/15216 for tracking), this PR is correct :)

valgur:
IMO, since the `vulkan-validationlayers` packages more-or-less packages only a runtime library (i.e. a plugin), the `package_type` would be more accurate and useful as `application`. It sets the `self.requires()` defaults for the package more appropriately as `headers=False, libs=False, run=True`.

See also #21495. The static utility library has been removed in the latest versions and only the runtime one remains.

samuel-emrys:
@RubenRBS @franramirez688 could either of you trigger a build of the necessary `spirv-tools` binary?

RubenRBS:
Sorry that it took so long to get this merged after approval, my bad on that one, thanks a lot for your patience :)

PR #19615, opened 2023-09-06
[SoPlex] bump version to 6.0.4

hedtke:
Specify library name and version:  **soplex/6.0.4**

I also enable shared builds on Windows. They have always been possible, it was just a missing `copy` in the recipe.

I also bump the versions of the dependencies.

---


recipes/soplex/all/conanfile.py
```diff
@@ -62,8 +63,6 @@ def validate(self):
                 raise ConanInvalidConfiguration(
                     f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
                 )
-        if is_msvc(self) and self.options.shared:
-            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
```
RubenRBS, line 66: Thanks a lot for checking this, it's really appreciated! :)

RubenRBS (approved):
Thanks a lot for your contribution and for adding the comments and traits, they are super helpful for future maintenance

Quick question: Why didn't the missing `transitive_libs` trait in zlib make the test package fail? is it _too_ simple?

hedtke:
> Thanks a lot for your contribution and for adding the comments and traits, they are super helpful for future maintenance
> 
> Quick question: Why didn't the missing `transitive_libs` trait in zlib make the test package fail? is it _too_ simple?

the test package failed in shared mode, but only with `-o with_boost=False`, now this also works. This setting is not tested in the CI as `with_boost=True` is the default and shared on Windows was disabled

czoido (approved):

PR #19618, opened 2023-09-06
cn-cbor: remove invalid check_min_cppstd

valgur:
Required by the latest version of `cose-c`.

### DIFF
--- a/recipes/cn-cbor/all/conandata.yml
+++ b/recipes/cn-cbor/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "cci.20200822":
+    url: "https://github.com/jimsch/cn-cbor/archive/f713bf67bcf3e076d47e474ce060252ef8be48c7.zip"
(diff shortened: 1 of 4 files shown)


RubenRBS:
Maybe we can get upstream to generate a release we can pin?
In their READMe they have Conan instructions to use ranges and it would be a pitty if someone followed that and got an old commit

valgur:
> Maybe we can get upstream to generate a release we can pin? In their READMe they have Conan instructions to use ranges and it would be a pitty if someone followed that and got an old commit

@RubenRBS That would be ideal, indeed, but I don't really feel like taking that on currently, tbh. Can we merge this one as is, maybe?

franramirez688 (commented):
The project has had no activity for 4 years, so I think it's not worth adding the last commit as a new version. If they release a new version, let's add it for sure.

franramirez688 (approved):

RubenRBS (approved):

PR #19633, opened 2023-09-06
[sentry-crashpad] add #include to fix compile with gcc13

paulharris:
Specify library name and version:  **sentry-crashpad/0.6.5**

GCC13 can't compile the code without a couple of #includes added.
... Just adding `#include <cstdint>`

I also deleted a  patch for an old version that is no longer included in this recipe.


RubenRBS (commented):
Thanks! Has this been reported upstream?

paulharris:
Reported here: https://github.com/chromium/mini_chromium/pull/3
Reported early by someone else too.

RubenRBS:
Thanks! Could you update the patch to use their suggested approach to change the header, and add a link to the PR that added ir upstream as the patch source? :)

PR #19694, opened 2023-09-09
qgeoview: add qgeoview/1.0.4

Jihadist:
Specify library name and version:  **qgeoview/1.0.4**

QGeoView recipe forces shared library https://github.com/AmonRaNet/QGeoView/blob/master/lib/CMakeLists.txt#L19 so I removed shared option

---


recipes/qgeoview/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+            del self.options.fPIC
+
+    def configure(self):
+        self.options.rm_safe("fPIC")
+        # FIXME: we shouldn't have to force shared in qt, but config file
+        #        generated by conan in qt static is likely broken, or maybe env vars.
+        self.options["qt"].shared = True
+
+    def requirements(self):
+        self.requires("qt/[>=5.15.2 <7]")
```
RubenRBS, line 38: Hi! Thanks for the idea, but QT is not yet one of the allowed libraries to be required with version ranges - We're working towards that future, but understanding the implications for each library is a slow process :) Please pin this for now with the version that makes most sense, usually the latest available one in CCI

recipes/qgeoview/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        self.options.rm_safe("fPIC")
+        # FIXME: we shouldn't have to force shared in qt, but config file
+        #        generated by conan in qt static is likely broken, or maybe env vars.
```
RubenRBS, line 34: If this is a bug, could you report an issue to the client/this repo, depending on where you feel it fits better? Thanks!

recipes/qgeoview/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class QGeoViewConan(ConanFile):
+    name = "qgeoview"
+    description = "QGeoView is a Qt / C ++ widget for visualizing geographic data. "
+    license = "LGPL-3.0"
+    topics = ("geo", "geospatial", "qt")
+    homepage = "https://github.com/AmonRaNet/QGeoView"
+    url = "https://github.com/conan-io/conan-center-index"
+    settings = "os", "arch", "compiler", "build_type"
```
RubenRBS, line 19: missing `package_type` attribute. [Docs here](https://docs.conan.io/2/reference/conanfile/attributes.html#package-type)

RubenRBS (changes requested):
Thanks a lot! I have a request due to QT's version range usage and a question about a comment in the code, but otherwise the recipe looks good

recipes/qgeoview/all/conanfile.py
Jihadist: Hi! Thank you for your attention. I tried to fix qt5 or qt6. Conan v1 pipeline works with qt5 but v2 pipeline works only with qt6.

recipes/qgeoview/all/conanfile.py
Jihadist: I cannot build any qt gui library with qt static and conan v1. It's an old problem that was fixed by v2.

recipes/qgeoview/all/conanfile.py
Jihadist: done

Jihadist:
Blocked by #22788

valgur (approved):

Jihadist:
thank you for no reviews ^)

PR #19757, opened 2023-09-12
libmysqlclient: Remove Doxygen requirement, remove old 8.0.25 version

### LINKED ISSUE #19654: [question] libmysqlclient : missing tool requirement ?
### What is your question?

When trying to consume _libmysqlclient_ (as a requirement) on a far-from-homogenous Linux system (project build with some user-built compiler, different from the system one, with different _glibc_), the build of _libmysqlclient_ fails because of _glibc_ incompatibilities.  
Something like

```
-- Found Doxygen: /usr/bin/doxygen  found components: doxygen dot 
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /usr/bin/doxygen)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /usr/bin/doxygen)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `CXXABI_1.3.13' not found (required by /usr/bin/doxygen)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /lib/x86_64-linux-gnu/libclang-14.so.13)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /lib/x86_64-linux-gnu/libclang-cpp.so.14)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /lib/x86_64-linux-gnu/libclang-cpp.so.14)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `CXXABI_1.3.13' not found (required by /lib/x86_64-linux-gnu/libclang-cpp.so.14)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `CXXABI_1.3.13' not found (required by /lib/x86_64-linux-gnu/libLLVM-14.so.1)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /lib/x86_64-linux-gnu/libLLVM-14.so.1)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /lib/x86_64-linux-gnu/libLLVM-14.so.1)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /lib/x86_64-linux-gnu/libz3.so.4)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /lib/x86_64-linux-gnu/libz3.so.4)
/usr/bin/doxygen: /opt/gcc-10.3.0/lib64/libstdc++.so.6: version `GLIBCXX_3.4.30' not found (required by /lib/x86_64-linux-gnu/libicuuc.so.72)
```

Well, there are many ways to solve this issue.  The easiest and generalised one being to add _Doxygen_ as a tool requirement in the profile used for the build, like
```
[tool_requires]
doxygen/1.9.4
```

But it still seems to me that _Doxygen_ *should* have been added as a tool requirement in the recipe, as it is mandatory for the build to success in this kind of setup

Am I wrong ?

If requirement miss is confirmed, I can submit the PR.

Thanks

RubenRBS:
Specify library name and version:  **libmysqlclient/all**

Doxygen is only used for the docs. As there's no configure option to disable the docs, we comment its related FindPackage out

Closes #19654


recipes/libmysqlclient/all/conandata.yml
```diff
@@ -31,10 +31,3 @@ patches:
     - patch_file: "patches/0006-fix-cpp20-build-8.0.29.patch"
       patch_description: "Fix C++20 compilation"
       patch_type: "portability"
-  "8.0.25":
```
SpaceIm, line 34: I guess you also want to remove 8.0.25 under `sources` field.

recipes/libmysqlclient/all/conanfile.py
```diff
@@ -196,6 +188,10 @@ def _patch_sources(self):
                         "  INSTALL_DEBUG_SYMBOLS(",
                         "  # INSTALL_DEBUG_SYMBOLS(")
 
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
+                        "FIND_PACKAGE(Doxygen)",
+                        "# FIND_PACKAGE(Doxygen)")
```
SpaceIm, line 193: Maybe add a comment explaining why we are doing this?

recipes/libmysqlclient/all/conandata.yml
RubenRBS: Good catch

ericLemanissier (approved):

oleurodecision:
(thanks for the PR, I was on it but had network issues last days)

Removing 8.0.25 has the drawback that it is the only one conanified version supporting pre c++17. Won't argue on the pros & cons of removing or keeping this (old) version, I just wanted to underline this fast and possibly get an explanation (not keeping too many versions ?)

SpaceIm (approved):

RubenRBS:
@oleurodecision that's a fair question. Yes, it was due to simplification of the recipe and not keeping _too_ many old versions around, but you're right, I think we should keep this one for a bit longer for the reasons you mention

Thanks a lot for bringing it up :)

PR #19773, opened 2023-09-13
[SCIP] bump version to 8.0.4

hedtke:
Specify library name and version:  **scip/8.0.4**

SCIP and SoPlex are coupled together: there is a 1-to-1 correspondence between the SCIP version and the SoPlex version, thus the new map in this recipe.

---

### DIFF
--- a/recipes/scip/all/conanfile.py
+++ b/recipes/scip/all/conanfile.py
@@ -80,7 +80,6 @@ class SCIPConan(ConanFile):
         if self.options.with_sym == "bliss":
             self.requires("bliss/0.77")
         self.requires("soplex/6.0.3")
-        self.requires("zlib/1.2.13")
 
     def configure(self):
         self.options["soplex"].with_gmp = self.options.with_gmp
--- a/recipes/scip/all/conanfile.py
+++ b/recipes/scip/all/conanfile.py
@@ -84,6 +84,7 @@ class SCIPConan(ConanFile):
         if self.options.with_sym == "bliss":
             self.requires("bliss/0.77")
         self.requires(f"soplex/{self.soplex_version_belonging_to_me[self.version]}")
+        self.requires("zlib/[>=1.2.11 <2]")
 
     def configure(self):
         self.options["soplex"].with_gmp = self.options.with_gmp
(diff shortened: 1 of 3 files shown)


recipes/scip/all/conanfile.py
```diff
@@ -76,11 +80,10 @@ def config_options(self):
 
     def requirements(self):
         if self.options.with_gmp:
-            self.requires("gmp/6.2.1")
+            self.requires("gmp/6.3.0")
         if self.options.with_sym == "bliss":
             self.requires("bliss/0.77")
-        self.requires("soplex/6.0.3")
-        self.requires("zlib/1.2.13")
+        self.requires(f"soplex/{self.soplex_version_belonging_to_me[self.version]}")
```
RubenRBS, line 86: If this is 1 to 1, couldn't we use `self.version` as the version? Or are you opening the door to possibly using newer patches?

hedtke: It is 1-to-1, but Soplex 6.0.4 belongs to SCIP 8.0.4, so versions are not exactly the same.

RubenRBS: That makes sense, thanks for the extra insight :)

RubenRBS (approved):

recipes/scip/all/conanfile.py
```diff
@@ -76,11 +80,10 @@ def config_options(self):
 
     def requirements(self):
         if self.options.with_gmp:
-            self.requires("gmp/6.2.1")
+            self.requires("gmp/6.3.0")
         if self.options.with_sym == "bliss":
             self.requires("bliss/0.77")
-        self.requires("soplex/6.0.3")
-        self.requires("zlib/1.2.13")
```
uilianries, line 83: @hedtke Why did you remove zlib? I see it listed in scip files: https://github.com/scipopt/scip/blob/v804/src/xml/xmldef.h#L52

hedtke: it is automatically included by soplex

uilianries: @hedtke please, revert that change. Each package should manage their own dependencies, getting a hide over other dependencies to include a package will create a problem related to the package ID.

hedtke: Re-added and also using ranges instead of a specific version

uilianries (approved):
LGTM

RubenRBS (approved):

RubenRBS:
Thanks for your patience @hedtke :)

PR #19776, opened 2023-09-13
cgal: add CMake variable CGAL_USE_GMP for compatibility

lrineau:
Specify library name and version:  **cgal/5.6**

In a private discussion, a user of the `cgal` package pointed out the lack of the CMake variable `CGAL_USE_GMP`, that they want to use in the downstream project https://bitbucket.org/taucgl/cgal-python-bindings

The upstream CGAL CMake scripts define that variable, as well as `CGAL_USE_MPFR`, when CGAL is configured with GMP, but the Conan package `cgal` does not. This pull-request fixes that.

---

### DIFF
--- a/recipes/cgal/all/conanfile.py
+++ b/recipes/cgal/all/conanfile.py
@@ -163,6 +163,10 @@ function(CGAL_setup_CGAL_flags target)
 endfunction()
 
 CGAL_setup_CGAL_flags(CGAL::CGAL)
+
+# CGAL use may rely on the presence of those two variables
+set(CGAL_USE_GMP  TRUE CACHE INTERNAL "CGAL library is configured to use GMP")
+set(CGAL_USE_MPFR TRUE CACHE INTERNAL "CGAL library is configured to use MPFR")
 ''')
         save(self, module_file, content)
 


RubenRBS (approved):
Hi @lrineau thanks a lot for your contribution!
As you're an author of the library, I'm trusting you with the reasoning behind your changes :)

I'm wondering if we can avoid a similar situation like this for the future - this one was eventually caught, but what if a new versin makes us diverge from upstream more? Do you have any ideas on how to tackle it?

czoido (approved):

lrineau:
@RubenRBS 
> I'm wondering if we can avoid a similar situation like this for the future - this one was eventually caught, but what if a new versin makes us diverge from upstream more? Do you have any ideas on how to tackle it?

I do not see a clean way to check that. Currently I would say that CGAL CMake scripts are way too complex, and what I did in the CGAL conan package is probably too simple to cover all use cases. But I do not see a simple way to avoid that.

Maybe I should try to compile and launch the CGAL examples and tests, using the CGAL conan package, but it takes hours, and we cannot afford to launch that in every recompilation of the conan package, and for all platforms. A launch of the CGAL test suite is about 5 hours per platform, if you use about 4 CPU cores (on modern architectures).

In my opinion, the best we can hope is to fix bugs incrementally when user report any.

PR #19808, opened 2023-09-14
openblas: add version 0.3.24

joakimono:
Specify library name and version:  **openblas/0.3.24**

This PR adds version 0.3.24 with some minor modifications to the recipe.
OpenBLAS can since 0.3.21 be built with option `build_lapack=True` without a Fortran compiler. This PR adds the functionality to exploit this, by checking `self.conf.get("tools.build:compiler_executables")` and setting appropriate CMake variables to trigger Fortran-less build when appropriate.

This work is a simplified variant of PR #17173, where handling of Fortran runtime and other issues are skipped.
---

### DIFF
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -34,6 +34,14 @@ class OpenblasConan(ConanFile):
         "dynamic_arch": False,
     }
     short_paths = True
+    package_type = "library"
+
+    @property
+    def _fortran_compiler(self):
+        comp_exe = self.conf.get("tools.build:compiler_executables")
+        if comp_exe and 'fortran' in comp_exe:
+            return comp_exe["fortran"]
+        return None
 
     def config_options(self):
         if self.settings.os == "Windows":
@@ -75,9 +83,19 @@ class OpenblasConan(ConanFile):
     def generate(self):
         tc = CMakeToolchain(self)
 
+        tc.cache_ variables["NOFORTRAN"] = not self.options.build_lapack
+        # This checks explicit user-specified fortran compiler
         if self.options.build_lapack:
-            self.output.warning("Building with lapack support requires a Fortran compiler.")
-        tc.cache_variables["NOFORTRAN"] = not self.options.build_lapack
+            if not self._fortran_compiler:
+                if Version(self.version) < "0.3.21":
+                    self.output.warning(
+                        "Building with LAPACK support requires a Fortran compiler.")
+                else:
+                    tc.cache_variables["C_LAPACK"] = True
+                    tc.cache_variables["NOFORTRAN"] = True
+                    self.output.info(
+                        "Building LAPACK without Fortran compiler")
+
         tc.cache_variables["BUILD_WITHOUT_LAPACK"] = not self.options.build_lapack
         tc.cache_variables["DYNAMIC_ARCH"] = self.options.dynamic_arch
         tc.cache_variables["USE_THREAD"] = self.options.use_thread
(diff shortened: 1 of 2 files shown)


samuel-emrys:
Looks like 0.3.24 has added an OpenBLASConfig.cmake file that you'll need to remove to make the 1.x hooks happy:

```diff
    def package(self):
        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
        cmake = self._configure_cmake()
        cmake.install()
+       rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
```

recipes/openblas/all/conanfile.py
```diff
@@ -75,19 +83,28 @@ def layout(self):
     def generate(self):
         tc = CMakeToolchain(self)
 
-        if self.options.build_lapack:
-            self.output.warning("Building with lapack support requires a Fortran compiler.")
         tc.cache_variables["NOFORTRAN"] = not self.options.build_lapack
+        # This checks explicit user-specified fortran compiler
+        if self.options.build_lapack:
+            if not self._fortran_compiler:
+                if Version(self.version) < "0.3.21":
```
samuel-emrys, line 90: This should probably be less than this version, as this is the version in which the fortran requirement is being removed
```suggestion
                if Version(self.version) < "0.3.24":
```

recipes/openblas/all/conanfile.py
```diff
@@ -137,7 +157,9 @@ def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "OpenBLAS")
         self.cpp_info.set_property("cmake_target_name", "OpenBLAS::OpenBLAS")
         self.cpp_info.set_property("pkg_config_name", "openblas")
-        cmake_component_name = "pthread" if self.options.use_thread else "serial" # TODO: ow to model this in CMakeDeps?
+        cmake_component_name = "pthread" if self.options.use_thread else "serial"  # TODO: how to model this in CMakeDeps?
+        self.cpp_info.components["openblas_component"].set_property(
+            "cmake_target_name", "OpenBLAS::" + cmake_component_name)  # 'pthread' causes issues without namespace
```
samuel-emrys, line 162: I'm a bit confused about what this is doing - do you know why this is necessary? It might be a bit strange to add `OpenBLAS::pthread` and `OpenBLAS::serial` as targets - I can't imagine this is what anybody would expect.

Having said that, this seems to be a reasonable deviation from what upstream CMake suggests, which would be more akin to [[1](https://cmake.org/cmake/help/latest/module/FindBLAS.html)][[2](https://cmake.org/cmake/help/latest/module/FindLAPACK.html)]:

```cmake
set(BLA_VENDOR OpenBLAS)
find_package(BLAS)
find_package(LAPACK)

target_link_libraries(mylib BLAS::BLAS LAPACK::LAPACK)
```

So, what this package is doing is unexpected anyway, and I don't have a clear idea how we would even model the above in conan - maybe this is the best way to handle it

joakimono: The reason for adding the targets as I have suggested is because they are defined in the upstream [OpenBLAS config targets](https://github.com/OpenMathLib/OpenBLAS/blob/9f815cf1bf16b4e64d4aee681b33558fc090b62a/cmake/OpenBLASConfig.cmake.in#L19). Or, that is, they do not have the `OpenBLAS::` namespace. `pthread` as target name causes issues without this prefix on Linux. I do not know why it works for the native OpenBLAS target script, but not for the conan generated one.

It could be useful to add [cmake target aliases](https://docs.conan.io/2/reference/tools/cmake/cmakedeps.html#properties) besides the standard OpenBLAS target names, e.g. `BLAS::BLAS` and `LAPACK::LAPACK` as you suggest. We should at the very least retain the `OpenBLAS::OpenBLAS` target, which an OpenBLAS would expect.

samuel-emrys: Fair enough. We could probably add aliases for `BLAS::BLAS` and `LAPACK::LAPACK` in this recipe as well (I think this would be suitable in the case where a consumer links against either of these), but I think matching the interface above by setting `BLA_VENDOR` is probably beyond the capabilities of `CMakeDeps` generator at the moment, so don't worry about this too much. I think we can defer these changes for a future PR. I know there were issues with complexity in the previous PR, so lets keep this one simple and targeted.

I hadn't realised that OpenBLAS exported its own config file, so thanks for directing me to that.

samuel-emrys (approved):

samuel-emrys (approved):

uilianries (approved):
Okay, it seems be following https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.21

PR #19830, opened 2023-09-15
tsl-hat-trie: add package_type

SpaceIm:

---

### DIFF
--- a/recipes/tsl-hat-trie/all/conanfile.py
+++ b/recipes/tsl-hat-trie/all/conanfile.py
@@ -21,7 +21,7 @@ class TslHatTrieConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("tsl-array-hash/0.7.1", transitive_headers=True)
+        self.requires("tsl-array-hash/0.7.1")
 
     def package_id(self):
         self.info.clear()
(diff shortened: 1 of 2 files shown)


recipes/tsl-hat-trie/all/conanfile.py
```diff
@@ -14,14 +14,14 @@ class TslHatTrieConan(ConanFile):
...
-
+    package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
 
     def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("tsl-array-hash/0.7.1")
+        self.requires("tsl-array-hash/0.7.1", transitive_headers=True)
```
RubenRBS, line 24: No need for transitive_headers in a `package_type = "header-library"` recipe, it's automatically inferred ([Docs here](https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-type-trait-inferring))

Approving either way as it does not change the behaviour of the program and we avoid recompiling it :)

RubenRBS (dismissed):

recipes/tsl-hat-trie/all/conanfile.py
SpaceIm: ok thanks, I was not sure.

SpaceIm, line 24:
```suggestion
        self.requires("tsl-array-hash/0.7.1")
```

RubenRBS (approved):

ericLemanissier (approved):

PR #19833, opened 2023-09-15
fmi1: Add new recipe for fmi

joakimono:
Specify library name and version:  **fmi1/1.0.1**

The functional mock-up interface is a leading standard to share dynamic simulation models. 
https://fmi-standard.org/
A so-called functional mock-up unit (FMU) make use of the FMI headers to implement the interface.
This PR adds a recipe to provide the header files for version 1 of this standard. 
Currently, there are three major versions of this standard and they can be (and actually are) used at the same time within a consuming library. For this reason, fmi2 and fmi3 are two forthcoming PRs with these major version variants.

---


recipes/fmi1/all/patches/001-add-license-file.patch
RubenRBS, line 1: Where does this file come from? (Any reason why it is not included in the release files?)

RubenRBS (commented):
Thanks! The recipe is looking good - I only have one concern about the library license

Just to double check, the reasoning behind creating 2 other recipes instead of adding them as new versions here is because they can coexist all 3 at the same time, and often do?

recipes/fmi1/all/patches/001-add-license-file.patch
joakimono: I extracted the license text from one of the headers (fmiFunctions.h). I could not find a separate license file in version 1 of this standard.

RubenRBS: Thanks for the explanation :)

joakimono: I have added a commit that uses the same mechanism as you describe to extract license text instead of using a patch file.

RubenRBS (approved):

joakimono:
This is a blocker for another package **fmilibrary** to be added.

czoido (approved):

RubenRBS:
Thanks for the ping @joakimono - I'll manually merge this later to unblock you as no PRs are automatically merged while under maintenance mode :)

PR #19856, opened 2023-09-15
zint: add package_type + fix test package for conan v2 + use version range for zlib

SpaceIm:
- add package_type
- modernize more (use rm_safe)
- fix test package for conan v2 client (legacy self.options["zint"].with_qt and new self.dependencies can't be used in test(), therefore we use the know trick with ctest)
- bump libpng & qt
- use version range for zlib

---


RubenRBS:
Thanks! Will un-draft when it's this library's turn to get the zlib range :)

czoido (approved):

PR #19857, opened 2023-09-15
llvm-openmp: add versions v13, v14, v15, v16 and v17-rc4

valgur:


uilianries (commented):
All those version? 🤔  What about adding only latest for now, then adding other in case someone asks in an issue? Remember that anything from LLVM consumes tons of resources to be built. E.g. I need 1h30 to build llvm-clang a single configuration in my machine and is not a bad a hardware at all.

valgur:
> All those version? 🤔  What about adding only latest for now, then adding other in case someone asks in an issue? Remember that anything from LLVM consumes tons of resources to be built. E.g. I need 1h30 to build llvm-clang a single configuration in my machine and is not a bad a hardware at all. 

It fortunately does not depend on the core llvm components and builds in seconds due to being pure C.

uilianries (approved):
LGTM

uilianries:
I checked locally and they are really small packages.

marxin:
@valgur : About `# All versions except for v11 crash with a segfault for the simple test_package.cpp test`: Have you filled an upstream issue about it? Is it something known by upstream?

valgur:
@marxin No, I have not. Feel free to investigate the issue and report it upstream.

marxin:
I bet it's caused by the following hairy issue: https://github.com/llvm/llvm-project/issues/49923. Apparently, the Apple interest for OpenMP on aarch64 is pretty low :/

PR #19877, opened 2023-09-16
Updated ImGui. Added the imgui_stdlib source and header file

GavinNL:
Specify library name and version:  **ImGui/1.89.9.0**

ImGui provides some additional functionality for C++ std::strings, but the source is not compiled into the library in the current recipe. 

This patch adds the source and header to CMakeLists.txt

Tested it against ImGui `1.89.9`, `1.89.8-docking` and `1.85` on Linux Mint 21 

My access may have expired. I have re-requested ( https://github.com/conan-io/conan-center-index/issues/4#issuecomment-1722120840 )

--


czoido (dismissed):

samuel-emrys (approved):

RubenRBS:
Question: If we want this one, why not the rest of the listed in https://github.com/ocornut/imgui/wiki/Useful-Extensions#cness?

RubenRBS (commented):
I have a question: If we want this one, why not the rest of the listed in https://github.com/ocornut/imgui/wiki/Useful-Extensions#cness?

GavinNL:
> Hi! Thanks a lot for your contribution :) I have a question: If we want this one, why not the rest of the listed in https://github.com/ocornut/imgui/wiki/Useful-Extensions#cness?

Hi Ruben,

I don't think those extensions are part of the imgui repo.

I think the only exception is the free_type extension which is in the `misc/freetype` folder. But including that will require us to link against freetype. I don't think people will want to be forced to do that.

One option is to include source/header files from `misc/cpp` and `misc/freetype` in the installed header directory, rather than compiled into the library. Then the user can manually insert these source files by adding the following lines to the bottom of one of their source files if they want to use the extension. Similar to how they would choose which backend they want to use


```c++
#include <imgui.h>

// source not included in static library
#include <imgui_stdlib.h>
#include <imgui_freetype.h>
int main()
{
.....
}

// include extensions's source
#include <imgui_stdlib.cpp>
#include <imgui_freetype.cpp>
```

RubenRBS:
Oh, I see! Thanks for the extra context. Let me check with @czoido, but this looks good, ty

GavinNL:
> Oh, I see! Thanks for the extra context. Let me check with @czoido, but this looks good, ty

Actually. Now that I think about it, it might be better to include the source/header files for the extensions rather than compile it into the library. This won't change the library that is built, but it will still allow people to include those extensions.

I can make the change

GavinNL:
Okay, updated.

The cpp files are no longer compiled into the library, but they are included in the install so they can be manually compiled into the user's application using the following code:

```cpp
#include <imgui.h>

// source not included in static library
#include <imgui_stdlib.h>
#include <imgui_freetype.h>
int main()
{
.....
}

// include extensions's source
#include <imgui_stdlib.cpp>
#include <imgui_freetype.cpp>
```

czoido:
Checking this again, I was thinking that maybe it would be better to explicitly package those files in the `package()` method, like we do with the bindings and then import the files in the consumer, like:

```python

    def package(self):
    ...
        copy(self, pattern="*.*",
            dst=os.path.join(self.package_folder, "res", "misc", "cpp"),
            src=os.path.join(self.source_folder, "misc", "cpp"))
        copy(self, pattern="*.*",
            dst=os.path.join(self.package_folder, "res", "misc", "freetype"),
            src=os.path.join(self.source_folder, "misc", "freetype"))
...
```

And then in the consumer:

```python
    def generate(self):
        copy(self, "*.*", os.path.join(self.dependencies["imgui"].package_folder,
            "res", "misc", "cpp"), os.path.join(self.source_folder))
        copy(self, "*.*", os.path.join(self.dependencies["imgui"].package_folder,
            "res", "misc", "freetype"), os.path.join(self.source_folder))
       ...

```
cc @RubenRBS

GavinNL:
> ```python
>         copy(self, pattern="*.*",
>             dst=os.path.join(self.package_folder, "res", "misc", "cpp"),
>             src=os.path.join(self.source_folder, "misc", "cpp"))
>         copy(self, pattern="*.*",
>             dst=os.path.join(self.package_folder, "res", "misc", "freetype"),
>             src=os.path.join(self.source_folder, "misc", "freetype"))
> ```

GavinNL:
Doh! I accidentally closed the PR...

I pushed the changes that @czoido suggested. The src/header files are now included in the res folder

GavinNL:
> I tried applying the patch from @GavinNL in an _imgui_ app that defines **IMGUI_ENABLE_FREETYPE** before including `imgui.h`. The consuming conanfile.py also copies `imgui_freetype.{c,hpp}` from the package and includes them in the build. It works (fonts are smoother due to switch from STB to Freetype), but with two caveats.
> 
> First, you have to instantiate the Freetype builder object yourself (if you don't then Freetype is not used). The reason is that **IMGUI_ENABLE_FREETYPE** is only defined in the consuming app and not when Conan builds the _imgui_ library. This means that automatic instantiation of the builder is not activated in `imgui_draw.cpp`:
> 
> ```c++
> bool    ImFontAtlas::Build()
> {
> [snip]
>     const ImFontBuilderIO* builder_io = FontBuilderIO;
>     if (builder_io == NULL)
>     {
> #ifdef IMGUI_ENABLE_FREETYPE
>         builder_io = ImGuiFreeType::GetBuilderForFreeType();
> #elif defined(IMGUI_ENABLE_STB_TRUETYPE)
>         builder_io = ImFontAtlasGetBuilderForStbTruetype();
> #else
>         IM_ASSERT(0); // Invalid Build function
> #endif
>     }
> [snip]
> }
> ```
> 
> This is not what the user will expects but can be worked around in the consuming apps code:
> 
> ```c++
> // Create the font builder before building the atlas (implicitly or through Build())
> ImGui::GetIO().Fonts->FontBuilderIO = ImGuiFreeType::GetBuilderForFreeType()
> ```
> 
> The second caveat is that code for the inferior STB font builder is no longer removed even if **IMGUI_ENABLE_FREETYPE** is defined. The code in `imgui_internal.h` is included from `imgui_draw.cpp` and does not see **IMGUI_ENABLE_FREETYPE** and thus does not skip defining **IMGUI_ENABLE_STB_TRUETYPE**:
> 
> ```c++
> // Enable stb_truetype by default unless FreeType is enabled.
> // You can compile with both by defining both IMGUI_ENABLE_FREETYPE and IMGUI_ENABLE_STB_TRUETYPE together.
> #ifndef IMGUI_ENABLE_FREETYPE
> #define IMGUI_ENABLE_STB_TRUETYPE
> #endif
> ```
> 
> The end result is a slightly larger binary.
> 
> Would it not make sense to side-step these quirks by introducing options to the _imgui_ package?
> 
> ```python
>     options = {
>         "enable_cpp": [True, False],
>         "enable_freetype": [True, False],
>         "enable_freetype_lunasvg": [True, False],
>     }
>     default_options = {
>         "enable_cpp": False,
>         "enable_freetype": False,
>         "enable_freetype_lunasvg": False,
>     }
> ```
> 
> Would require
> 
> * Some checking in validate() to ensure lunasvg only is enabled if freetype also is
> * Adding conditional requirements (freetype, lunasvg).
> * Including the freetype/cpp files in the library build
> 
> The consumer of _imgui_ would do
> 
> ```python
>     def configure(self)
>         self.options["imgui"].enable_cpp = True
>         self.options["imgui"].enable_freetype = True
>         self.options["imgui"].enable_freetype_lunasvg = True
> ```
> 
> or set the corresponding options using Conan profiles.
> 
> No need to mess with the extra cpp files for the consumer since they are already included in the library.
> 
> Thoughts?
> 
> EDIT: The _imgui_ freetype_lunasvg support got a compilation bugfix recently in [ocornut/imgui#6842](https://github.com/ocornut/imgui/issues/6842) which might require a package patch until a new release of _imgui_ is available.

Hi @technoyes , thanks for testing that out. 

Whenever I've used ImGui, I have never compiled it into an external library. I always manually included the cpp files in one of my sources, that way I have explicit control over which definitions are set.

One thing good about ImGui is that it does not have any dependencies. If we have freetype as a explicit conan option, then that will force freetype as a dependency, requiring us to set a specific version and then maintain both configurations going forward. I fear that will make this whole conan package a lot more complicated to deal with in the future

I think if we want to keep the simplicity of the original package, while still having the flexibility for the user to use the extensions, we should include all the cpp files for imgui as a resource. That way, there is only one set of options that conan has to maintain, but users will still have access to the imgui cpp files to compile the way they want.

GavinNL:
I updated the recipe to include the source files in the res folder. I tested it against a project that I'm working on and it allows me to do exactly what I need. Hopefully this is acceptable

technoyes:
> Whenever I've used ImGui, I have never compiled it into an external library. I always manually included the cpp files in one of my sources, that way I have explicit control over which definitions are set.

Yes, and I believe that is actually how Dear ImGui is intended to be used. Quoting the README:

```
No specific build process is required. You can add the .cpp files into your existing project.
```

And the Conan "adding packages" README states

``` 
In ConanCenter, our belief is recipes should always match upstream, in other words, what the original author(s) intended.
```

I think your patch to include the ImGui sources as resources is a good thing and helps achieve the above goals.

I did try and add some options to the package just to learn a bit more about Conan. Seems to mostly work except for `find_package(lunasvg)` failing for some reason.  It does admittedly add a little complexity, but not too much. I don't know how much of a maintenance burden it would impose.

Leaving a link here in case someone is interested: https://github.com/technoyes/conan-center-index/commit/1394901fb76cd47a3c1a686f9b64a56ab3b78b82 .

technoyes:
Is the PR as it stands good enough to merge? I see some of the CI builds were aborted. @czoido  @GavinNL

Looking forward to import the Conan-installed ImGui source and enable freetype support :-)

GavinNL:
Hi @technoyes , I'm not planning on making any changes, so if everyone is happy, its good to merge/review again.

technoyes:
Ok. Will see what @RubenRBS  and @czoido think when they get  time to have a look.

RubenRBS (approved):

czoido (approved):

technoyes:
c3i/conan-v2/pr-merge 

> Failure in build 8 (`2cc4c77d7caf46a96f9aa55888f84a7ea2a66b48`):
> 
> An unexpected error happened and has been reported. Help is on its way! 🏇

Is this some kind of internal error in the v2 CI pipeline, or is there a problem with the PR as such @RubenRBS ?

czoido:
> c3i/conan-v2/pr-merge
> 
> > Failure in build 8 (`2cc4c77d7caf46a96f9aa55888f84a7ea2a66b48`):
> > An unexpected error happened and has been reported. Help is on its way! 🏇
> 
> Is this some kind of internal error in the v2 CI pipeline, or is there a problem with the PR as such @RubenRBS ?

I think there was some problem in the ci job, I have relaunched it. Let's wait for ci 🤞

PR #19907, opened 2023-09-17
[armadillo] Add version 12.6.4

samuel-emrys:
Specify library name and version:  **armadillo/12.6.4**

Add version 12.6.4 of armadillo. The patch file here makes the same changes as previous patches, but added to account for differences in the upstream CMakeLists.txt that prevents existing patch files from being applied.

---

### DIFF
--- a/recipes/armadillo/all/conanfile.py
+++ b/recipes/armadillo/all/conanfile.py
@@ -186,7 +186,7 @@ class ArmadilloConan(ConanFile):
         # See https://gitlab.com/conradsnicta/armadillo-code/-/issues/227 for more information.
         if self.options.use_hdf5 and Version(self.version) < "12":
             # Use the conan dependency if the system lib isn't being used
-            self.requires("hdf5/1.14.0")
+            self.requires("hdf5/1.14.0", transitive_headers=True)
 
         if self.options.use_blas == "openblas":
             self.requires("openblas/0.3.20")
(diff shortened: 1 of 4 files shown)


recipes/armadillo/all/conanfile.py
```diff
@@ -186,10 +186,10 @@ def requirements(self):
         # See https://gitlab.com/conradsnicta/armadillo-code/-/issues/227 for more information.
         if self.options.use_hdf5 and Version(self.version) < "12":
             # Use the conan dependency if the system lib isn't being used
-            self.requires("hdf5/1.14.0")
+            self.requires("hdf5/1.14.0", transitive_headers=True, transitive_libs=True)
```
RubenRBS, line 189: Per conversation in Slack:

> Without the traits, consumers fail during linking with home/conan/w/prod-v2/bsr/9620/afdcf/p/openb04dc4c8cf208a/p/lib/libopenblas.so.0: error adding symbols: DSO missing from command line
https://c3i.jfrog.io/c3i/misc-v2/logs/pr/19907/2-linux-gcc/armadillo/10.7.0//6057bdc3db7276f815dd06db0a25b3e779201d99-test.txt When shared = True

> An explanation https://linuxpip.org/how-to-fix-dso-missing-from-command-line. This is a tightening of requirements in later versions of binutils requiring consumers to explicitly link shared libs even when the symbol can be identified in a transitive dep. Setting transitive_libs=True adds -lopenblas to the consumer link args which is the suggested fix and resolves this

RubenRBS (approved):
Thanks!

PR #19920, opened 2023-09-17
godot-cpp: add version 4.1.1

fmorgner:
Specify library name and version:  **godot-cpp/4.1.1**

---


recipes/godot-cpp/config.yml
```diff
@@ -5,3 +5,5 @@ versions:
     folder: 3.x.x
   "cci.3.2-20200130":
     folder: 3.x.x
+  "4.1.1":
+    folder: all
```
Croydon, line 9: I would name the folder `4.x.x`

`all` only makes sense if there are no further directories

recipes/godot-cpp/config.yml
```diff
@@ -1,7 +1,9 @@
 versions:
   "3.5.1":
-    folder: all
+    folder: 3.x.x
   "3.3.4":
-    folder: all
+    folder: 3.x.x
   "cci.3.2-20200130":
+    folder: 3.x.x
+  "4.1.1":
```
valgur, line 8: Please place the newest version on top.

recipes/godot-cpp/all/conanfile.py
```diff
@@ -12,131 +10,43 @@ class GodotCppConan(ConanFile):
...
+        platform = {
+            "Android": "android",
+            "Linux": "linux",
+            "Macos": "darwin",
+            "Windows": "windows",
+        }[self.settings.get_safe("os")]
+        build_type = self.settings.get_safe("build_type").lower()
+
+        if self.settings.get_safe("os") == "Android":
+            self.cpp_info.libs= [f"godot-cpp.{platform}.{build_type}"]
```
valgur, line 43:
```suggestion
        }[self.settings.os]
        build_type = self.settings.build_type.lower()

        if self.settings.os == "Android":
            self.cpp_info.libs= [f"godot-cpp.{platform}.{build_type}"]
```

valgur: `build_type`, `os` and `arch` are standard settings that are guaranteed to be available.

recipes/godot-cpp/all/conanfile.py
```diff
@@ -12,131 +10,43 @@ class GodotCppConan(ConanFile):
...
         else:
-            self.cpp_info.libs = [self._libname]
+            bits = 64 if self.settings.get_safe("arch") in ["x86_64", "armv8"] else 32
```
valgur, line 45:
```suggestion
            bits = 64 if self.settings.arch in ["x86_64", "armv8"] else 32
```

recipes/godot-cpp/config.yml
valgur: I agree. Would also make the PR diff cleaner.

recipes/godot-cpp/all/conanfile.py
fmorgner: adjusted by using `str(...)` where a string-like or hashable object is required.

fmorgner:
I will look into the V2 build/lint of the old versions on the weekend.

fmorgner:
I fixed the linter issues in 3.x.x (and one in 4.x.x). I tried to keep the changes small (although i did fix the format string warnings as well), since this PR is primarily focused on adding 4.1.1. I think that modernizing the 3.x.x conanfile should be done separately in a future PR (or by rebasing and adjusting #18862).

fmorgner:
Is the v2 build expected to be re-triggered automatically (like the v1 one) or is there something I (or someone else) has to do?

valgur (approved):

Croydon:
Can we pls get a CI re-run here? @uilianries @RubenRBS 

ConanV2 was in maintenance when it did run

uilianries:
@Croydon Danke! I just re-triggered now.

Croydon:
No new CI feedback?

uilianries:
I just fixed some git conflicts, it triggers a new CI build that should post the result now.

PR #19923, opened 2023-09-17
cprocessing: fix min msvc version + do not suggest unofficial CMake names + bump glfw & stb

SpaceIm:
- fix msvc min version
- do not suggest unofficial CMake names
- bump glfw & stb
- remove no_copy_source since there is a patch

---

### DIFF
--- a/recipes/cprocessing/all/conanfile.py
+++ b/recipes/cprocessing/all/conanfile.py
@@ -18,10 +18,8 @@ class CProcessingConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/maksmakuta/CProcessing"
     topics = ("processing", "opengl", "sketch", "header-only")
-
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
-    no_copy_source = True
 
     @property
     def _min_cppstd(self):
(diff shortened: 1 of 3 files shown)


recipes/cprocessing/all/conanfile.py
```diff
@@ -18,10 +18,8 @@ class CProcessingConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/maksmakuta/CProcessing"
     topics = ("processing", "opengl", "sketch", "header-only")
-
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
-    no_copy_source = True
```
RubenRBS, line 24: Isn't it the other way around? As we don't have patches, we can set `no_copy_source` to True to avoid unnecessary copies, no?

RubenRBS (commented):
Question about the removal of `no_copy_source`

recipes/cprocessing/all/conanfile.py
SpaceIm: there is a patch, a replace_in_file()

RubenRBS: Oh I missread your PR comment and thought that there werent, thanks for the clarification :)

RubenRBS (approved):

ericLemanissier (approved):

czoido (approved):

PR #19929, opened 2023-09-17
btyacc: conan v2 support

SpaceIm:

---

### DIFF
--- a/recipes/btyacc/all/test_package/conanfile.py
+++ b/recipes/btyacc/all/test_package/conanfile.py
@@ -1,28 +1,33 @@
 from conan import ConanFile
 from conan.tools.build import can_run
-from conan.tools.cmake import CMake, CMakeDeps, cmake_layout
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.env import VirtualRunEnv
 
 
 class TestPackageConan(ConanFile):
     settings = "os", "arch", "compiler", "build_type"
-    generators = "CMakeToolchain", "VirtualBuildEnv"
+    generators = "CMakeDeps", "CMakeToolchain"
     test_type = "explicit"
 
     def layout(self):
         cmake_layout(self)
 
-    def build_requirements(self):
-        self.tool_requires(self.tested_reference_str)
+    # FIXME: we should test btyacc as a build requirement and
+    #        add btyacc to build_context_activated of CMakDeps but there is a
+    #        conan client bug: https://github.com/conan-io/conan/issues/14758
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
 
     def generate(self):
-        deps = CMakeDeps(self)
-        deps.build_context_activated = ["btyacc"]
-        deps.generate()
+        if can_run(self):
+            VirtualRunEnv(self).generate(scope="build")
 
     def build(self):
-        cmake = CMake(self)
-        cmake.configure()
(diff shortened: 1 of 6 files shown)


SpaceIm:
I don't test btyacc in build requirements of test package due to https://github.com/conan-io/conan/issues/14758

recipes/btyacc/all/test_package/conanfile.py
```diff
@@ -1,23 +1,48 @@
...
+        if hasattr(self, "settings_build") or can_run(self):
+            cmake = CMake(self)
+            cmake.configure()
+            cmake.build()
 
     def test(self):
-        if not tools.cross_building(self):
-            self._configure_cmake().test()
+        if can_run(self):
+            self.run(f"ctest --output-on-failure -C {self.settings.build_type}", env="conanrun")
```
RubenRBS, line 48: I was going to complain about the ctest usage again, but I see that this comes from the original recipe, so won't block on this

franramirez688, line 48: @SpaceIm I don't get anything about this test_package. I know that this code was somehow there, but let's discuss this one.

This recipe is an `application` one, so, IMO, this conanfile should look like:

```python
from conan import ConanFile


    generators = "VirtualBuildEnv"
    test_type = "explicit"

    def build_requirements(self):

    def test(self):
        self.run("btyacc -b y [THE REST OF THE PARAMETERS HERE]")
```
Asking because I'm sure I'm missing something somewhere...

SpaceIm: This test package also test that we can consume btyacc with find_package() and that Cmake imported target of executable works as expected, so we can't rely on such naive test.

franramirez688: Yeah, but it's only using the `BTYACC_EXECUTABLE` variable from there, so I was wondering if it seemed too much, and could be simpler than that.

Perhaps:

CMakeLists.txt
```cmake
cmake_minimum_required(VERSION 3.1)
project(test_package LANGUAGES C)

find_package(btyacc REQUIRED CONFIG)
if(NOT DEFINED BTYACC_EXECUTABLE)
    message(FATAL_ERROR "BTYACC_EXECUTABLE is not defined")
endif()
```
conanfile
```python
import os

from conan import ConanFile
from conan.tools.cmake import cmake_layout, CMake


    generators = "CMakeDeps", "CMakeToolchain", "VirtualBuildEnv"
    test_type = "explicit"

    def layout(self):
        cmake_layout(self)

    def build_requirements(self):

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build()

    def test(self):
        self.run("btyacc -b y [THE REST OF THE PARAMETERS HERE]")
```
It's only an idea. Testing both CMake and env executable at the same time could be a better fit.

SpaceIm: It wouldn't be a representative usage. I don't understand what's the problem with current test package. It works and it's a representative test.

franramirez688: No no, it's fine for me. I was only thinking of other alternatives to test the same thing, but this one was already there so I have no objections.

franramirez688 (approved):

jwillikers (approved):

PR #19942, opened 2023-09-18
imagl: Patching sources' CMakeLists.txt to use find_package instead of legacy conan 1.X cmake generator

Woazim:
imaGL sources use conan to deal with its own dependencies. Some patches already remove conan calls in sources' CMakeLists.txt, but CMakeLists.txt expects that conan use the old conan 1.X cmake generator.

I made new patches to remove the need of cmake generator and use find_package instead which seems to be a better practice.

---


valgur (approved):

jcar87 (approved):
thanks @Woazim !

RubenRBS (approved):
Thanks! SOrry for the delay on getting this reviewed :)

PR #19944, opened 2023-09-18
Clhep bump version

weatherhead99:
Specify library name and version:  **clhep/2.4.6.4**

This is a version bump for the CLHEP library. I intend to start working on packaging quite a few other downstream cern/HEP libraries in the coming months and bumping this helps with being able to package recent versions of other things.

---


SpaceIm:
Title is misleading, this recipe is already v2 compatible.

recipes/clhep/all/conanfile.py
```diff
@@ -6,8 +6,7 @@
 from conan.tools.microsoft import is_msvc
 import os
 
-required_conan_version = ">=1.53.0"
-
+required_conan_version = ">=2.0.9"
```
SpaceIm, line 9: Is it allowed by cci policy? I don't see any value to break conan v1 users.

recipes/clhep/config.yml
```diff
@@ -1,3 +1,5 @@
 versions:
   "2.4.1.3":
     folder: all
+  "2.4.6.4":
```
SpaceIm, line 4: newest versions at the top please

recipes/clhep/all/conanfile.py
weatherhead99: it doesn't break conan v1 users, they have available the previous recipe revisions still (don't they?). They just won't get the new version. But, for example, given that "implements" attribute is only available in conan2 that's why I bumped the required conan version. I wasn't aware that use of conan 2 only features was disallowed but I do see now in FAQ it states things should be 1 and 2 compatible so I will revert this

recipes/clhep/all/conanfile.py
```diff
@@ -52,6 +44,7 @@ def validate(self):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
```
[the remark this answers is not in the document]
weatherhead99: will do, but, can I ask - given that this applies to all feasible build configurations, why?

weatherhead99:
> Title is misleading, this recipe is already v2 compatible.

sorry, name of my branch before I realised that. PR is renamed

recipes/clhep/all/conanfile.py
SpaceIm: It will break conan v1 users if they want to update recipe RREV. `implements` avoids some boilerplate, it's not worth breaking v1 users.

For the moment, recipes must support conan v1, see https://github.com/conan-io/conan-center-index/discussions/19104#discussion-5500696

recipes/clhep/all/conanfile.py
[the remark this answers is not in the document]
SpaceIm: Immutability of source code. Some patches may be conditional (depending on os, compiler etc), so it cannot be done in source(). conancenter prefers to keep all the modification of source code in the same place.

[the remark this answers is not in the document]
weatherhead99: I understand that, but none of _these_ patches are conditional (notwithstanding 2 of them are for windows compatibility, they get applied to all configurations anyway).  For myself in this case then I'd prefer it in source() because precisely because it conveys that meaning.  You have the original source in your cache as well anyway, at least in the core.sources:download_cache ;-) .

But I'm not arguing of course if that's the rule that's the rule.

recipes/clhep/all/conanfile.py
```diff
@@ -8,14 +8,14 @@
...
 
-
 class ClhepConan(ConanFile):
     name = "clhep"
     description = "Class Library for High Energy Physics."
     license = "LGPL-3.0-only"
     topics = ("cern", "hep", "high energy", "physics", "geometry", "algebra")
     homepage = "http://proj-clhep.web.cern.ch/proj-clhep"
     url = "https://github.com/conan-io/conan-center-index"
+    implements = ["auto_shared_fpic"]
```
RubenRBS, line 18: I love that you want to add such new features to CCI, but sadly it will have to wait until we stop supporting v1 alongside v2, you still need to manually implement this by hand :)

RubenRBS (changes requested):
Hi! Thanks a lot for your contribution, I have a request before being able to move forward with this PR :)

RubenRBS (dismissed):

recipes/clhep/all/conanfile.py
```diff
@@ -41,6 +37,9 @@ def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
+    def export_sources(self):
+        export_conandata_patches(self)
+
```
SpaceIm, line 42: keep this where it was before please

recipes/clhep/all/conanfile.py
```diff
@@ -8,7 +8,6 @@
 
 required_conan_version = ">=1.53.0"
 
-
```
SpaceIm, line 11: same here, so that you avoid a unecessary build

SpaceIm (approved):

RubenRBS (approved):
Hi @weatherhead99 thanks a lot for your patience while I got to circle back to this PR, the waiting time exceeds what I'm comfortable with, my bad

Other than that, the changes now look great, so let's get this finally merged! Thanks again :)

PR #18693, opened 2023-07-19
s2let: migrate to Conan v2

valgur:
TODO

- [x] Merge #18665 first.
- [x] Merge #19957

PR #19957, opened 2023-09-19
astro-informatics-so3: add transitive_libs=True

valgur:
Required for
- #18693

I'm still puzzled why `transitive_libs=True` is required at all and so often, here and in other PRs.


recipes/astro-informatics-so3/all/conanfile.py
```diff
@@ -57,12 +59,18 @@ def generate(self):
         tc.variables["conan_deps"] = False
         tc.variables["CONAN_EXPORTED"] = True
         tc.variables["BUILD_TESTING"] = False
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
         tc.generate()
         deps = CMakeDeps(self)
         deps.set_property("fftw", "cmake_target_name", "FFTW3::FFTW3")
         deps.generate()
 
+    def _patch_sources(self):
```
uilianries, line 68: I still prefer asking to the upstream supporting shared, instead of forcing it here. Changing it will generate a content that's not expect by upstream's users.

valgur: I don't prefer the asking part, unless I'm personally invested in the project, tbh. :grin:
I can revert the forced shared libs support.

valgur: Reverted to static-only.

danimtb (approved):

valgur:
@jwillikers @ericLemanissier A community review would be welcome to get this merged.

ericLemanissier (approved):

PR #19968, opened 2023-09-19
cgif: bump meson

SpaceIm:
Without this bump, it fails during configuration with apple-clang 15:

```
../src/meson.build:1:0: ERROR: Unable to detect linker for compiler `clang -Wl,--version -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk -arch x86_64`
stdout:
stderr: ld: unknown options: --version
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I guess that all meson based recipes are affected and will need a bump (version range for meson would help).

---

### DIFF
--- a/recipes/cgif/all/conanfile.py
+++ b/recipes/cgif/all/conanfile.py
@@ -47,7 +47,7 @@ class CgifConan(ConanFile):
             raise ConanInvalidConfiguration(f"{self.ref} doesn't support shared build with Visual Studio")
 
     def build_requirements(self):
-        self.tool_requires("meson/1.0.0")
+        self.tool_requires("meson/1.2.1")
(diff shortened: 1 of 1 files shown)


RubenRBS:
Ohh, thanks for this! I've found other recipes with which I had the same issue, now I know the reason

SpaceIm:
> Ohh, thanks for this! I've found other recipes with which I had the same issue, now I know the reason

Yep, I don't know what has been fixed in meson and when, but I know that even with meson 1.1.1 it fails. I guess it has been fixed in meson 1.2.0 or 1.2.1

PR #18807, opened 2023-07-22
shaderc: migrate to Conan v2, simplify patching, add v2023.6

valgur:

PR #19978, opened 2023-09-20
spirv-headers: add new versions, remove old ones

valgur:
Kept only versions from the past year.

### DIFF
--- a/recipes/spirv-headers/all/conandata.yml
+++ b/recipes/spirv-headers/all/conandata.yml
@@ -1,4 +1,13 @@
 sources:
+  "1.3.261.1":
+    url: "https://github.com/KhronosGroup/SPIRV-Headers/archive/refs/tags/sdk-1.3.261.1.tar.gz"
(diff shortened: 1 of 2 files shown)


RubenRBS:
Thanks! did you check if any of the removed ones is in use by any other recipe in CCI?

valgur:
> Thanks! did you check if any of the removed ones is in use by any other recipe in CCI?

Yes. It's only used transitively via `spirv-tools` in `diligent-core` and `shaderc`. `spirv-tools` uses 1.3.224.0, which can be safely bumped to 1.3.224.1. And as for `shaderc`, the motivation for it was to bump its dependencies in the first place for https://github.com/conan-io/conan-center-index/pull/18807/files#diff-8081cb7851a6cf6ba55b29260e409058efb299d48cc9b8c21d0fd5e46480f805R46-R58

jwillikers (approved):

valgur:
Closing in favor of #20838.

SpaceIm:
Sorry, didn't see there was an old PR already.

PR #19979, opened 2023-09-20
spirv-tools: add new versions, remove old ones

valgur:
Depends on
- #19978

### DIFF
--- a/recipes/spirv-tools/all/conanfile.py
+++ b/recipes/spirv-tools/all/conanfile.py
@@ -43,8 +43,8 @@ class SpirvtoolsConan(ConanFile):
         return {
             "17": {
                 "apple-clang": "10",
-                "clang": "7",
-                "gcc": "8",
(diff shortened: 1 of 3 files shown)


SpaceIm:
Could you also add 1.3.268.0 please? https://github.com/conan-io/conan-center-index/pull/20838 has been merged.

recipes/spirv-tools/all/conanfile.py
```diff
@@ -43,8 +43,8 @@ def _compilers_minimum_version(self):
         return {
             "17": {
                 "apple-clang": "10",
-                "clang": "5",
-                "gcc": "7",
+                "clang": "7",
+                "gcc": "8",
```
SpaceIm, line 47: Is it no specific to this new version? Why would it raise for clang 5 and gcc 7 for previous versions?

franramirez688 (approved):

Issue #19951, opened 2023-09-19
[package] `cryptopp-pem`, `microprofile`, `nas` and `tweetnacl` recipes broken because https://unlicense.org is down since July


SpaceIm:
### Description

`cryptopp-pem`, `microprofile`, `nas` and `tweetnacl` recipes download an extra file from https://unlicense.org but this website is down since July: https://github.com/unlicense/unlicense.org/issues/89

Therefore they can't be built from source anymore since `source()` method fails to download this file.

### Logs

```
nas/1.9.4: ERROR: Error downloading file https://unlicense.org/UNLICENSE: 'HTTPSConnectionPool(host='unlicense.org', port=443): Max retries exceeded with url: /UNLICENSE (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1007)')))'
nas/1.9.4: Waiting 5 seconds to retry...
nas/1.9.4: ERROR: Error downloading file https://unlicense.org/UNLICENSE: 'HTTPSConnectionPool(host='unlicense.org', port=443): Max retries exceeded with url: /UNLICENSE (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1007)')))'
nas/1.9.4: Waiting 5 seconds to retry...
ERROR: nas/1.9.4: Error in source() method, line 65
        download(self, filename="LICENSE", **self.conan_data["sources"][self.version][1])
        ConanException: Error downloading file https://unlicense.org/UNLICENSE: 'HTTPSConnectionPool(host='unlicense.org', port=443): Max retries exceeded with url: /UNLICENSE (Caused by SSLError(SSLError(1, '[SSL: SSLV3_ALERT_HANDSHAKE_FAILURE] sslv3 alert handshake failure (_ssl.c:1007)')))'
```

RubenRBS:
Progress (Still need to push the branches):
* nas -> fixed
* microprofiles -> actually uses unlicense from the repo, keeping as-is (But needs porting to v2!)
* tweetnacl -> same as microprofiles
* cryptopp-pem -> same as last 2

RubenRBS:
I'm just now seeing that CCI might not be correctly using the `LicenseRef` SPDX attribute, I'll double check with the team, thanks for bringing this up to our attention though

[...]

PR #19981, opened 2023-09-20
nas: Fix license

RubenRBS:
Specify library name and version:  **nas/all**

thanks to @SpaceIm for noticing and reporting it in #19951

### DIFF
--- a/recipes/nas/all/conanfile.py
+++ b/recipes/nas/all/conanfile.py
@@ -3,7 +3,7 @@ from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.layout import basic_layout
 from conan.tools.env import VirtualBuildEnv
-from conan.tools.files import chdir, get, download, export_conandata_patches, apply_conandata_patches, rm, copy, load, save
+from conan.tools.files import chdir, get, export_conandata_patches, apply_conandata_patches, rm, copy, load, save
 from conan.tools.gnu import AutotoolsToolchain, Autotools, AutotoolsDeps
 import os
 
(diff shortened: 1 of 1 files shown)


czoido (approved):

recipes/nas/all/conanfile.py
```diff
@@ -17,7 +17,7 @@ class NasRecipe(ConanFile):
     topics = ("audio", "sound")
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://www.radscan.com/nas.html"
-    license = "Unlicense"
+    license = "DocumentRef-wave.h:LicenseRef-MIT-advertising"
```
SpaceIm, line 20: What does it mean?

RubenRBS: We're reevaluating our understanding of the spdx specification. The docs right now point contributors to do something like "LicenseRef-FILENAME", but upon further inspection, I realized that the filename should be i the DocumentRef section, and that LicenseRef should be an id for the license. Probably this one one is actually incorrect too, and should be something like `DocumentRef-wave.h:LicenseRef-MIT-advertising-nas-variant` or something like that

Link here https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/

RubenRBS: Good question, I honestly don't know yet - licensing is always confusing!

uilianries: I asked in opensource stackexchange about it. We could use this pattern as well, but the answer was using something simpler like `LicenseRef-nas`, then add some fields in the license file. 

https://opensource.stackexchange.com/a/14309/31147

Anyway, I'm fine with this current approach too, as it's documented.

uilianries (approved):
Asked about it on https://opensource.stackexchange.com/a/14309/31147

It looks fine according to SPDX docs.

PR #19982, opened 2023-09-20
spirv-cross: add new versions, remove old ones

valgur:
Kept only versions from the past year.


RubenRBS (dismissed):

ericLemanissier (changes requested):
Could you also please remove the unused patches in this one ? thanks!

recipes/spirv-cross/config.yml
```diff
@@ -7,29 +13,5 @@ versions:
...
-  "cci.20211113":
-    folder: all
-  "cci.20210930":
-    folder: all
-  "cci.20210823":
-    folder: all
-  "cci.20210621":
-    folder: all
-  "20210115":
-    folder: all
```
danimtb, line 27: Please keep this option as it is an active requirement of "shaderc" recipe at https://github.com/conan-io/conan-center-index/blob/91e67bdec23d59d4e2a104a3343762142b300058/recipes/shaderc/all/conanfile.py#L71

recipes/spirv-cross/config.yml
```diff
@@ -7,29 +13,5 @@ versions:
     folder: all
   "1.3.231.1":
     folder: all
-  "1.3.224.0":
```
danimtb, line 10: Please keep this version as it is a requirement for "diligent-core" recipe https://github.com/conan-io/conan-center-index/blob/91e67bdec23d59d4e2a104a3343762142b300058/recipes/diligent-core/all/conanfile.py#L116

RubenRBS (changes requested):
As @danimtb pointed out, there are some used versions still - we should either keep those or bump their dependencies

RubenRBS (approved):
Thanks! I know it can be frustrating to have to keep old versions around

ericLemanissier (approved):

PR #19989, opened 2023-09-20
wt: add version 4.10.1

toge:
Specify library name and version:  **wt/4.10.1**

---


RubenRBS:
Any idea what the issue might be?

uilianries:
> Any idea what the issue might be?

My guess is some missing support from that compiler version. I'll check locally.

uilianries:
Just like I suspected ... Some features of C++ has been removed in C++17 causing compilation errors.

I was able to reproduce the same error as the CI, then I tried with `compiler.cppstd=14` and worked.

In that case, we will need [check_max_cppstd](https://docs.conan.io/2/reference/tools/build.html#conan-tools-build-check-max-cppstd), but there is a problem: It's not ported to Conan 1.x and will not be ... So we will need a workaround:


https://github.com/conan-io/conan-center-index/blob/7483e5be48215583c829f9bf34173759aa4093bd/recipes/opencv/2.x/conanfile.py#L413

uilianries:
@toge I was added the workaround with `check_max_cppstd`. It should fix the Conan 2.x situation.

uilianries (approved):
LGTM

jwillikers (approved):

SpaceIm:
> Just like I suspected ... Some features of C++ has been removed in C++17 causing compilation errors.
> 
> I was able to reproduce the same error as the CI, then I tried with `compiler.cppstd=14` and worked.
> 
> In that case, we will need [check_max_cppstd](https://docs.conan.io/2/reference/tools/build.html#conan-tools-build-check-max-cppstd), but there is a problem: It's not ported to Conan 1.x and will not be ... So we will need a workaround:
> 
> https://github.com/conan-io/conan-center-index/blob/7483e5be48215583c829f9bf34173759aa4093bd/recipes/opencv/2.x/conanfile.py#L413

Did you report this issue upstream? It's quite unfortunate to have this kind of regression in latest version.

uilianries:
@SpaceIm no, but looking the upstream, there is a workaround using a define:

https://github.com/emweb/wt/commit/7956a40c736aae9c5be87ed68d19e2b00bb81bc2

SpaceIm:
So why is there max cppstd? Is it not supposed to work with C++17 and above? This commit is part of 4.10.1

uilianries:
Yes, but will need a new PR reverting and passing a new define to the compiler. I'll open it.

PR #20025, opened 2023-09-20
hlslpp: add version 3.3.1

valgur:


recipes/hlslpp/all/conanfile.py
```diff
@@ -17,6 +19,16 @@ class HlslppConan(ConanFile):
...
 
+    @property
+    def _min_cppstd(self):
+        return 11
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "6",
+        }
```
uilianries, line 30:
```suggestion
        return {"gcc": "6"} if Version(self.version) >= "3.3" else {}
```

Do not restrict older versions, in case they work with Conan 1.x

franramirez688 (approved):

jwillikers (approved):

PR #20068, opened 2023-09-21
Recipe/20067 armadillo fix cmake targets

### LINKED ISSUE #20067: [package] armadillo/12.2.0 Doesn't use upstream cmake target
### Description

CMake has a published [`FindArmadillo.cmake`](https://cmake.org/cmake/help/latest/module/FindArmadillo.html) module that defines the filename to be `Armadillo` and the target to be `Armadillo::Armadillo`. This is in contrast to the default recipe `armadillo::armadillo` currently used.

This should be updated to reflect the upstream convention.

N/A

### Conan profile

N/A

### Steps to reproduce

N/A

### Logs

```
Put your log output here
```

</details>

samuel-emrys:
Specify library name and version:  **armadillo/12.6.4**

CMake has a published [FindArmadillo.cmake](https://cmake.org/cmake/help/latest/module/FindArmadillo.html) module that defines the filename to be Armadillo and the target to be Armadillo::Armadillo. This is in contrast to the default recipe armadillo::armadillo currently used.

This should be updated to reflect the upstream convention. This recipe builds on top of #19907 so that it builds in conan 2. #19907 should be merged prior to this because of that.

Closes #20067

---


samuel-emrys:
This obviously brings in too many commits. I'll fix these once the other armadillo PRs are merged.

SpaceIm:
```
CMake has a published [FindArmadillo.cmake](https://cmake.org/cmake/help/latest/module/FindArmadillo.html) module that defines the filename to be Armadillo and the target to be Armadillo::Armadillo. This is in contrast to the default recipe armadillo::armadillo currently used.
```

That's not true for CMake target. See https://github.com/conan-io/conan-center-index/issues/20067#issuecomment-1730155244

samuel-emrys:
@SpaceIm @RubenRBS a review would be appreciated, noting my comment in https://github.com/conan-io/conan-center-index/issues/20067#issuecomment-1839749295

valgur (approved):

recipes/armadillo/all/conanfile.py
```diff
@@ -271,10 +306,24 @@ def package(self):
...
         rmdir(self, os.path.join(self.package_folder, "share"))
+        self._create_cmake_module_variables(os.path.join(self.package_folder, self._module_vars_rel_path))
+
 
     def package_info(self):
         self.cpp_info.libs = ["armadillo"]
         self.cpp_info.set_property("pkg_config_name", "armadillo")
+        self.cpp_info.set_property("cmake_find_mode", "both")
+        self.cpp_info.set_property("cmake_file_name", "Armadillo")
+        self.cpp_info.set_property("cmake_target_name", "Armadillo::Armadillo")
```
MartinDelille, line 317: I'll be happy to approve your PR but I wanted to understand first what is the logic for setting the `cmake_file_name` and `cmake_target_name` ?

I generally see it lowercase in other recipe ? Is it a convention from the conan-center-index or something related to the upstream repository ?

samuel-emrys: Good question. The file name is being overridden here to match the upstream cmake module, which is named FindArmadillo.cmake. The target, since it's inherited from the module name, has also been overridden so that it's not interpreted as `armadillo::armadillo`. As per my comment in https://github.com/conan-io/conan-center-index/issues/20067#issuecomment-1839749295, I'm a little uncertain as to what to make the target name as it's not defined upstream. However, a couple of data points:

* the [`ensmallen` library uses `Armadillo::Armadillo`](https://github.com/mlpack/ensmallen/blob/5bb9fac6cf543ded1c69993a49b6fad285eef356/CMakeLists.txt#L61) (maintained by the same author as armadillo)
* The [`carma` library uses `armadillo::armadillo`](https://github.com/RUrlus/carma/blob/1a984a44e849fa632315d196dd71f8f29e670534/CMakeLists.txt#L73) (also maintained by the same author as armadillo), but they define their own [`GetArmadillo.cmake`](https://github.com/RUrlus/carma/blob/stable/cmake/GetArmadillo.cmake)
* [vcpkg uses `armadillo` as the target](https://github.com/Sentinel-One/vcpkg/blob/2b1757dd1d5ed7d9d5bb65fd8316631b157e7afc/ports/armadillo/usage#L4)
* UncommonCMakeModules (unofficial) [uses `Armadillo::Armadillo`](https://github.com/markjolah/UncommonCMakeModules/blob/f55b3ec18868c8ccfdd68ec4333c4e3b963e618d/FindArmadillo.cmake#L43)

[GitHub search results](https://github.com/search?q=%2Ftarget_link_libraries%5C%28.*%28PUBLIC%7CPRIVATE%7CINTERFACE%29+Armadillo%2F&type=code)

So, there's a bit of a smattering and no clear example of which target to choose. My thoughts are that `Armadillo::Armadillo` is the most consistent target, especially given usage in `ensmallen` and the module name. Having said that, I'm keen to hear community/team suggestions on consensus.

valgur: So how about an additional `self.cpp_info.set_property("cmake_target_aliases", ["armadillo::armadillo"])` as a compromise?

samuel-emrys: Great idea, thanks @valgur

samuel-emrys: I've updated the recipe with four different aliases to provide reasonable coverage of possible existing usage.

recipes/armadillo/all/conanfile.py
```diff
@@ -271,10 +306,25 @@ def package(self):
...
+        self.cpp_info.set_property("cmake_target_aliases", ["Armadillo", "armadillo", "armadillo::armadillo", "Armadillo::armadillo"])
```
MartinDelille, line 318: If you add alias, I would limit them to existing case then: 

```suggestion
        self.cpp_info.set_property("cmake_target_aliases", ["armadillo", "armadillo::armadillo"])
```

recipes/armadillo/all/test_package/CMakeLists.txt
```diff
@@ -1,12 +1,32 @@
 cmake_minimum_required(VERSION 3.15)
 project(PackageTest CXX)
 
-find_package(armadillo CONFIG REQUIRED)
+find_package(Armadillo CONFIG REQUIRED)
```
MartinDelille, line 4: Same question here: isn't the CCI convention to use the recipe name (lowercase) ?

recipes/armadillo/all/test_package/CMakeLists.txt
```diff
@@ -1,12 +1,32 @@
...
 if (LINK_HDF5)
     find_package(HDF5)
     set(HDF5_TARGETS HDF5::HDF5)
 endif()
 
 add_executable(example src/example.cpp)
-target_link_libraries(example armadillo::armadillo ${HDF5_TARGETS})
+target_link_libraries(example Armadillo::Armadillo ${HDF5_TARGETS})
```
MartinDelille, line 11: Same here:

```suggestion
target_link_libraries(example armadillo::armadillo ${HDF5_TARGETS})
```

recipes/armadillo/all/test_v1_package/CMakeLists.txt
```diff
@@ -4,8 +4,28 @@ project(PackageTest CXX)
 include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake)
 conan_basic_setup(TARGETS)
 
-find_package(armadillo REQUIRED)
+find_package(Armadillo REQUIRED)
```
MartinDelille, line 7:
```suggestion
find_package(armadillo REQUIRED)
```

recipes/armadillo/all/test_v1_package/CMakeLists.txt
```diff
@@ -4,8 +4,28 @@ project(PackageTest CXX)
...
 
 add_executable(example ../test_package/src/example.cpp)
-target_link_libraries(example armadillo::armadillo)
+target_link_libraries(example Armadillo::Armadillo)
```
MartinDelille, line 10:
```suggestion
target_link_libraries(example armadillo::armadillo)
```

recipes/armadillo/all/test_package/CMakeLists.txt
samuel-emrys: In this case, no. CCI convention is to match the upstream interface. As mentioned in the description, CMake publish a FindArmadillo.cmake module, which enforces the use of `find_package(Armadillo)` and CCI convention is that the conan package matches the upstream interface wherever provided.

recipes/armadillo/all/test_package/CMakeLists.txt
samuel-emrys: I don't really like it given that there's been a weak steer from the package maintainer that `Armadillo:Armadillo` is preferred, but since this isn't enforced I guess this does [match the CCI policy on CMake targets](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/test_packages.md#cmake-targets). I'll swap the primary target to `armadillo::armadillo` and make the aliases `Armadillo::Armadillo` and `armadillo`

recipes/armadillo/all/test_package/CMakeLists.txt
MartinDelille: Thank you for the clarification!

MartinDelille (approved):

samuel-emrys:
I've reverted back to using `Armadillo::Armadillo` as the primary target, as it seems like this is all conan 1.x can support when the module name is `FindArmadillo.cmake`. I've retained `armadillo::armadillo` as an alias in conan 2.x, so the net effect is nil in conan 2.x, but using `Armadillo::Armadillo` as the cmake target allows us to maintain a consistent interface between 1.x, 2.x and upstream.

danimtb (approved):

valgur (approved):

PR #20090, opened 2023-09-23
onnx: always use protobuf from build context in case of 2 profiles

SpaceIm:
Follow recommendations from conan team and potentially avoids weird side effects.
The workaround is kept for one profile (default native build in conan v1), because one profile has a bug preventing to have different versions in host & build contexts (can easily happen when consumer overrides versions in host context).

---


Hopobcn:
Conan v1 using two profiles would arrive to set `self.tool_requires("protobuf/<host_version>")`?

Hopobcn (approved):

SpaceIm:
> Conan v1 using two profiles would arrive to set `self.tool_requires("protobuf/<host_version>")`?

yes

jwillikers (approved):

PR #20097, opened 2023-09-24
Mp-units-2.0

mpusz:
Specify library name and version:  **mp-units/2.0**

---

### DIFF
--- a/recipes/mp-units/2.0.0/conanfile.py
+++ b/recipes/mp-units/2.0.0/conanfile.py
@@ -5,6 +5,7 @@ from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import copy, get, rmdir
+from conan.tools.microsoft import is_msvc
 from conan.tools.scm import Version
 
 required_conan_version = ">=1.59.0"
@@ -41,6 +42,7 @@ class MPUnitsConan(ConanFile):
 
     @property
     def _minimum_compilers_version(self):
+        # Note that apple-clang and msvc are disabled for now, their C++ 20 implementations are not up to speed
         return {"gcc": "11", "clang": "16"}
 
     @property
@@ -79,6 +81,11 @@ class MPUnitsConan(ConanFile):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires at least {compiler} {min_version} ({compiler.version} in use)"
             )
+        
+        # Note that apple-clang and msvc are disabled for now, their C++ 20 implementations are not up to speed
+        # Re-enable once newer versions with better support come out
+        if is_msvc(self) or compiler == "apple-clang":
+            raise ConanInvalidConfiguration(f"{self.ref} disabled for {compiler} as their C++20 implementation is not up to speed yet")
 
     def layout(self):
         cmake_layout(self, src_folder="src")
--- a/recipes/mp-units/2.0.0/conanfile.py
+++ b/recipes/mp-units/2.0.0/conanfile.py
@@ -1,0 +1,158 @@
+import os
+
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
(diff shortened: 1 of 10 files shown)


recipes/mp-units/2.0.0/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+    package_type = "header-library"
+    no_copy_source = True
+
+    @property
+    def _min_cppstd(self):
+        return "20"
+
+    @property
+    def _minimum_compilers_version(self):
+        return {"gcc": "11", "clang": "16"}
```
RubenRBS, line 44: Do you know the min required versions for `msvc` and `apple-clang`?

recipes/mp-units/2.0.0/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+        "quantities",
+        "dimensional-analysis",
+        "physical-quantities",
+        "physical-units",
+        "system-of-units",
+        "system-of-quantities",
+        "isq",
+        "si",
+        "library",
+        "quantity-manipulation",
```
RubenRBS, line 29:
```suggestion
        "quantity-manipulation",
        "header-only",
```

RubenRBS (commented):
Hi! Thanks a lot :)

I have a small question about some missing compiler checks, but otherwise the recipe looks nice, thanks a lot!

recipes/mp-units/2.0.0/conanfile.py
mpusz: Even though we require "only" C++20 support MSVC is too buggy to support that.

mpusz: Regarding apple-clang, for now we do not have enough experience to know if it works on the latest compiler version already. I have other priorities now and did not have time to investigate if we can make it work. But taking into account that mp-units 2.0 barely compiles (with many hacks) on clang-16, we expect it will take a while for apple-clang to catch up.

recipes/mp-units/2.0.0/conanfile.py
mpusz: It might be true for 2.0.0 but 2.1.0 will support C++ modules already which is hard to name "header-only".

mpusz: BTW, do you have any plans on how to handle and distribute C++ modules in Conan?

RubenRBS: Pinging @jcar87 which has some interesting stuff coming up on this front :)

jcar87: Hi @mpusz  - we are currently exploring how to best support modules and currently experimenting with a few libraries that have some support - if you have a preview of 2.1.0 with module support I can have a look and include that in our (internal, preliminary) testing.

First there's the question of what to package in the first place, then how it is consumed, and for the latter support is still needed in the wider tooling ecosystem - gcc only merged dependency scanning support last week (assuming this will make it to the gcc 14.1 release) and I expect CMake to have preliminary support for modules in imported targets in 3.28 - we'll try to very closely follow these developments so that for libraries that support modules, there are options  - even if they are experimental - so that tooling and library authors can make progress

mpusz: > if you have a preview of 2.1.0 with module support I can have a look and include that in our (internal, preliminary) testing.

I am about to merge the PR with modules (https://github.com/mpusz/mp-units/pull/350) in the next days. I will let you know when it is done,

recipes/mp-units/2.0.0/conanfile.py
RubenRBS: Wondering just now if we should disallow apple-clang and msvc builds for now then?

recipes/mp-units/2.0.0/conanfile.py
mpusz: https://github.com/mpusz/mp-units/pull/350 now compiles!

RubenRBS:
Will re-trigger once the CI queue clears a little

mpusz:
Hi, do we have any update on this receipe? Is everything OK?

RubenRBS (approved):

jwillikers (approved):
Nice!

@RubenRBS Is `VirtualBuildEnv` applied automatically now? I noticed it wasn't in the `generate` method but that doesn't appear to be causing problems in CI.

PR #20106, opened 2023-09-25
[docs] Add changelog September, 25 2023

uilianries:
- When a PR fails due missing dependencies, the CI should mark the PR with the label `Missing dependencies`. It should be used not only filter those PRs, but also to be improved in the future with other CCCI features.
- Some internal CI jobs were executed daily, but cost +10h be executed each, impacting directly Jenkins working queue. We re-scheduled them to be executed only during weekends.
- When Conan 2.x was introduced in CCI, there was a regression test the be sure once a recipe was working with Conan 2.x, it should not fail in a next PR. This feature has been removed, because any new PR should now pass by both Conan 1.x and 2.x as mandatory
- The epochs feature, a configuration to avoid mixing old and new configurations when builds, is now being deprecated. This is the first step.

---

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### 25-Sep-2023 - 14:33 CEST
+
+- [feature] Label PRs that have missing dependencies.
+- [feature] Remove check regression for Conan v2 pipeline.
+- [feature] Start deprecating epochs support in profile configurations.
+- [fix] Run TapaholesRepo job only once per week.
+- [fix] Run ListPackages job only once per week.
+
 ### 01-Sep-2023 - 19:58 CEST
 
 - [fix] Use Unix separators for Windows folder path when creating CI workspace.


danimtb (approved):

RubenRBS (approved):

SpaceIm:
> When Conan 2.x was introduced in CCI, there was a regression test the be sure once a recipe was working with Conan 2.x, it should not fail in a next PR. This feature has been removed, because any new PR should now pass by both Conan 1.x and 2.x as mandatory

What happen when a new recipe is submitted but depends on a recipe which is not v2 ready?

uilianries:
> What happen when a PR is submitted on a recipe which depends on another recipe not v2 ready?

@SpaceIm it breaks! But considering the low number of recipes not ported to Conan 2.x yet, it should not be a problem, most of those recipes are not a base dependency (we checked dependencies graphs before taking this decision). Plus, in case someone is blocked and will take time to fix, we can merge the PR manually, enabling to fix/update the PR in question.

PR #20110, opened 2023-09-25
Bump openssl's zlib requirement to [>=1.2.11 <2] range

RubenRBS:


uilianries (approved):

ericLemanissier (approved):

recipes/openssl/1.x.x/conanfile.py
```diff
@@ -228,7 +228,7 @@ def layout(self):
 
     def requirements(self):
         if self._full_version < "1.1.0" and not self.options.get_safe("no_zlib"):
-            self.requires("zlib/1.2.13")
+            self.requires("zlib/[>=1.2.11 <2]")
```
Croydon, line 231: I'm really not keen on these potential downgrades. This could have been a more forward-looking change

RubenRBS: The range will always prefer to use the latest released version, but allowing a lower version range avoids downstream consumers from conflicting while we transition to a consistent version range

PR #20133, opened 2023-09-25
Bump sentry-crashpad's zlib requirement to [>=1.2.11 <2] range

RubenRBS:


uilianries (approved):

recipes/sentry-crashpad/all/conanfile.py
```diff
@@ -60,7 +60,7 @@ def build_requirements(self):
             self.tool_requires("jwasm/2.13")
 
     def requirements(self):
-        self.requires("zlib/1.2.13")
+        self.requires("zlib/[>=1.2.11 <2]")
         if self.settings.os in ("Linux", "FreeBSD"):
             self.requires("libcurl/8.2.0")
```
SpaceIm, line 65: Maybe version range for libcurl in order to save one PR?

ericLemanissier (approved):

RubenRBS:
Weird that one of the statuses shows up as erroring 
cc/ @uilianries

recipes/sentry-crashpad/all/conanfile.py
uilianries: @SpaceIm It could save but also could create another problem with missing packages if we start to use version range for libcurl everywhere. Right now, we are focused to fix only zlib, otherwise, it will take even longer.

PR #19904, opened 2023-09-17
[doxygen] Model iconv dependency as a conan package

samuel-emrys:
Specify library name and version:  **doxygen/1.9.2**

This resolves an issue where `iconv` wasn't being appropriately modelled as a conan dependency in the doxygen recipe. This lead to unresolved symbol errors on Macos builds with conan 2. This was previously part of https://github.com/conan-io/conan-center-index/pull/18415, but is being split to a separate PR to separate concerns and make review easier.

Closes #19903 

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20134, opened 2023-09-25
Bump doxygen's zlib requirement to [>=1.2.11 <2] range

RubenRBS:
- Add iconv as dependency: Required when building doxygenapp
- Fix iconv linkage by using cmake targets: It uses cmake variables not compatible with current cmakedeps generated using iconv conan package
- Add option to build apps: The doxygenapp is optional
- Use version ranges for zlib
- Remove test_v1_package to avoid glibc symbol errors: https://github.com/conan-io/conan-center-index/pull/20134#issuecomment-1739386435
- Skip part of test_package result to avoid glibc symbol errors: https://github.com/conan-io/conan-center-index/pull/20134#issuecomment-1739386435

closes #19904


uilianries (dismissed):

ericLemanissier (approved):

uilianries:
@RubenRBS We built Doxygen with GCC9, using the docker image conanio/gcc9, based on Ubuntu 19.10 (GLIBC 2.30). However, when building test package, it runs all docker images, including those based on 16.04 which are running older glibc, resulting in missing symbols.

My suggesting is doing a workaround in test package to ignore those errors. Of course, it should be managed by c3i_jenkins, but it's another step.

uilianries (dismissed):

uilianries (approved):

recipes/doxygen/all/test_v1_package/conanfile.py
```diff
@@ -1,22 +1,17 @@
...
-            if not os.path.isdir(os.path.join(self.build_folder, "test_package", "html")):
-                raise ConanException("doxygen did not create html documentation directory")
-
-            self.output.info("Version:")
-            self.run("doxygen --version", run_environment=True)
+        if not tools.cross_building(self):
+            self.output.info("Doxygen Version:")
+            stringio = StringIO()
+            try:
+                self.run(f"{self.recipe_folder}/doxygen --version", run_environment=True, output=stringio)
```
uilianries, line 14:
```suggestion
                self.run(f"doxygen --version", run_environment=True, output=stringio)
```

czoido (dismissed):

uilianries (approved):

jwillikers (approved):

Croydon:
Hi @RubenRBS 

> Fix iconv linkage by using cmake targets: It uses cmake variables not compatible with current cmakedeps generated using iconv conan package

What exactly is the problem there? 🤔

PR #20139, opened 2023-09-25
Bump cassandra-cpp-driver's zlib requirement to [>=1.2.11 <2] range

RubenRBS:


ericLemanissier (approved):

uilianries (approved):

jcar87 (approved):

recipes/cassandra-cpp-driver/all/conanfile.py
```diff
@@ -146,6 +147,7 @@ def package(self):
         copy(self, pattern="LICENSE.txt", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
         cmake = CMake(self)
         cmake.install()
+        fix_apple_shared_install_name(self)
```
SpaceIm, line 150: @uilianries it's overkill, injecting `tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"` is more efficient (indeed it comes from old default CMake policies induced by https://github.com/datastax/cpp-driver/blob/2.17.0/CMakeLists.txt#L1).

SpaceIm: I've submitted https://github.com/conan-io/conan-center-index/pull/20174

uilianries: Good catch, thank you @SpaceIm

PR #20155, opened 2023-09-26
Add LASlib package

CJaccarino:
**LASLib/2.0.2**

The goal of this PR is to add laslib wich is part of LASTools. This library is used to read/write or process las/laz file which are file containing Lidar point data. This library can be used with the existing conan package laszip. I am not the author of the library.

---


RubenRBS:
Thanks for your contribution! The bot run twice per week to give you access to the CI, but it's skipped the last few due to the current maintenance, you'll be added in the next one, thanks a lot for your patience :)

recipes/laslib/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.generate()
+
+    def build(self):
+        apply_conandata_patches(self)
```
valgur, line 56: The patches can be avoided (almost?) entirely by replacing the root `CMakeLists.txt` as follows:
```suggestion
        save(self, os.path.join(self.source_folder, "CMakeLists.txt"), "add_subdirectory(LASlib/src)")
        save(self, os.path.join(self.source_folder, "LASlib", "example", "CMakeLists.txt"), "")
```

CJaccarino: I will update the root and the patch accordingly. Thanks :)

recipes/laslib/all/patches/0001-build-only-laslib.patch
```diff
@@ -0,0 +1,34 @@
+diff --git a/LASlib/CMakeLists.txt b/LASlib/CMakeLists.txt
+index 5efb193..6037262 100644
+--- a/LASlib/CMakeLists.txt
++++ b/LASlib/CMakeLists.txt
+@@ -9,4 +9,4 @@ else()
+ endif()
+ 
+ add_subdirectory(LASlib/src)
+-add_subdirectory(LASlib/example)
++
```
valgur, line 10:
```suggestion
```
This part of the patch should no longer be necessary since the `LASlib/example/CMakeLists.txt` is now truncated in the build step.

recipes/laslib/all/patches/0001-build-only-laslib.patch
valgur, line 1: Is this patch really necessary? The installed .cmake files can probably be easily removed in `package()`, which is preferable as it avoid the need to update the patch files when adding new versions to the package.

CJaccarino: As i understand i need this patch, without it i am not able to properly build the lib.

danimtb (approved):

jwillikers (approved):

Issue #20158, opened 2023-09-26
[package] <harfbuzz>/<8.1.1>: Incorrect Include Directory Structure in harfbuzz Conan Recipe


tuduongquyet:
### Description

Issue Title: Incorrect Include Directory Structure in harfbuzz Conan Recipe

Description:

Problem:

The harfbuzz Conan recipe currently generates an include directory structure that deviates from the standard behavior of harfbuzz when used in CMake projects. Normally, when we include harfbuzz in CMake projects, we expect to use headers like `#include <harfbuzz/hb.h>`. However, the current Conan recipe packages the harfbuzz headers in a way that necessitates a different include path, such as #include <hb.h>. This inconsistency in the include directory structure can lead to confusion and issues for projects using harfbuzz via Conan.

Expected Behavior:

The harfbuzz Conan package should align with the standard usage of harfbuzz, ensuring that the include directory structure remains consistent with the typical `harfbuzz/hb.h` format. This will ensure a smoother integration of harfbuzz into projects that rely on the standard harfbuzz header paths.

Proposed Solution:

To address this issue, it's recommended to modify the Conan recipe for harfbuzz. Specifically, in the conan.py file, we suggest adjusting the `self.cpp_info.includedirs` configuration to match the expected directory structure. By doing so, projects consuming harfbuzz through Conan will be able to use the standard `harfbuzz/hb.h` include path without any modifications.

Steps to Reproduce:

To reproduce this issue, simply create a CMake-based project that depends on the harfbuzz package from Conan, and attempt to include harfbuzz headers using `#include <harfbuzz/hb.h>`. You will observe that the headers are not found using the standard path, leading to compilation errors.

### Package and Environment Details

* Package Name/Version: **harfbuzzz/8.1.1**
* Operating System+version: **Linux Ubuntu 18.04**
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

os=Macos
os_build=Macos
arch=x86_64
arch_build=x86_64
compiler=apple-clang
compiler.version=12.0
compiler.libcxx=libc++
build_type=Release
compiler.cppstd=11
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

[...]

PR #20159, opened 2023-09-26
Fix Incorrect Include Directory Structure in harfbuzz Conan Recipe #20158

tuduongquyet:
Specify library name and version:  **harfbuzz/8.1.1**

Resolving #20158

---

### DIFF
--- a/recipes/harfbuzz/all/conanfile.py
+++ b/recipes/harfbuzz/all/conanfile.py
@@ -180,7 +180,6 @@ class HarfbuzzConan(ConanFile):
         if self.options.with_subset:
             self.cpp_info.libs.append("harfbuzz-subset")
         self.cpp_info.libs.append("harfbuzz")
-        self.cpp_info.includedirs.append(os.path.join("include", "harfbuzz"))
         if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.system_libs.append("m")
         if self.settings.os == "Windows" and not self.options.shared:


ericLemanissier (changes requested):
on my machine, (ubuntu), I have:
```
$ cat /usr/lib/x86_64-linux-gnu/pkgconfig/harfbuzz.pc
prefix=/usr
exec_prefix=/usr
libdir=/usr/lib/x86_64-linux-gnu
includedir=/usr/include

Name: harfbuzz
Description: HarfBuzz text shaping library
Version: 2.7.4

Libs: -L${libdir} -lharfbuzz
Libs.private: -lm
Requires.private: glib-2.0 >= 2.19.1 freetype2 >= 12.0.6 graphite2 >= 1.2.0
Cflags: -I${includedir}/harfbuzz
```
so the include path which is passed is `/usr/include/harfbuzz`, which contains all the hb*.h files

Also, `/usr/lib/x86_64-linux-gnu/cmake/harfbuzz/harfbuzz-config.cmake` has:
```
set_target_properties(harfbuzz::harfbuzz PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_harfbuzz_prefix}/${_harfbuzz_includedir}/harfbuzz"
  ...)
```

Also also,  `https://github.com/harfbuzz/harfbuzz/blob/main/test/fuzzing/hb-draw-fuzzer.cc#L4` does not specify `harfbuzz/` when including.

also also also, https://harfbuzz.github.io/a-simple-shaping-example.html does not specify `harfbuzz.` when including either

Where does your assumption that it should be included as `#include "harfbuzz/hb*.h` comes from ?

SpaceIm:
That's incorrect: I don't think so: https://github.com/harfbuzz/harfbuzz/blob/c8c97864e8c810068123ef62947be13675df54c2/CMakeLists.txt#L452-L454

These lines clearly say that CMake config files should expose headers under harfbuzz folder directly. You are breaking recipe actually here, while your incorrect include convention was also supported actually, I don't know how you consume this recipe.

tuduongquyet:
I'm currently working on a project that involves using harfbuzz, and in the codebase, we have references to harfbuzz/hb.h. I also conducted a search on GitHub and found numerous instances where harfbuzz/hb.h is used in other projects as well. This seems to be a common convention when working with harfbuzz.

Furthermore, it's a widely followed practice to structure include paths with the format <folder>/<header> when dealing with packages. This helps maintain clarity and organization in larger codebases.

However, I should have also considered the specific structure of the harfbuzz Debian package, which utilizes Cflags: -I${includedir}/harfbuzz. In this case, the package itself defines the include path as /usr/include/harfbuzz, and it's acceptable to directly include headers without the harfbuzz/ prefix due to this packaging configuration.

So, while #include "harfbuzz/hb.h" is a common convention, the Debian package configuration you mentioned allows for a simplified #include "hb.h" when working within the context of an Ubuntu system with the harfbuzz package installed.

So I think we should keep the same.
I will close the issue and the PR.

Thank you for reviewing. @ericLemanissier

tuduongquyet:
Thank you @SpaceIm

PR #20185, opened 2023-09-28
Extend assimp recipe to provide v5.3.1

Adnn:
assimp/5.3.1

Latest assimp version was lagging ~1.5 years. The majority of the work was to adapt the "unvendor-deps" patch.


RubenRBS:
Thanks a lot! I've restarted the build now to get proper build errors out :)

Adnn:
Thank you @RubenRBS for your message (otherwise the wall of automated notifications is a bit daunting for a first PR).

As far as I see, most CI builds were canceled because parts of `assimp/5.3.1@ Linux, GCC` failed. From [this link](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/20185/3-linux-gcc/assimp/5.3.1//summary.json), I think it only fails on GCC 5.

I do not know if assimp 5.3.1 is supposed to build on GCC 5, and I am not sure what I should be doing next to get this PR closer to being merged in?

Any advice will be greatly appreciated.

Adnn:
Hi,

  As I do not know what to do next to get thing going, could someone please advise what are my next steps?
@RubenRBS maybe?

>As far as I see, most CI builds were canceled because parts of assimp/5.3.1@ Linux, GCC failed. From [this link](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/20185/3-linux-gcc/assimp/5.3.1//summary.json), I think it only fails on GCC 5.
>
>I do not know if assimp 5.3.1 is supposed to build on GCC 5, and I am not sure what I should be doing next to get this PR closer to being merged in?
>
>Any advice would be greatly appreciated.

FranzPoize:
@RubenRBS Is there a way to not build this recipe with gcc 5 in CI ? It seems the crash is due to a warning that gcc that should not be a warning
https://stackoverflow.com/questions/25594644/warning-specialization-of-template-in-different-namespace

RubenRBS:
Yes! For issues that are not trivial to fix in recipes, we're more than happy to accept the usage of the `validate()` method raising `ConanInvaidConfigurationException`.

Check how this is done on the header-only package template https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/header_only/all/conanfile.py#L74 (But the proper condition here would be something like `if self.settings.compiler == "gcc" and Version(self.settings.compiler.version).major == "5"`:

recipes/assimp/5.x/patches/0007-unvendor-deps-5.3.x.patch
```diff
@@ -0,0 +1,315 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -132,12 +132,12 @@ OPTION ( ASSIMP_IGNORE_GIT_HASH
+ IF (WIN32)
+   OPTION( ASSIMP_BUILD_ZLIB
+     "Build your own zlib"
+-    ON
```
Croydon, line 7: Why are we overwriting default option values, when we can just set their value in the recipe?

recipes/assimp/5.x/patches/0007-unvendor-deps-5.3.x.patch
```diff
@@ -0,0 +1,315 @@
...
++    OFF
+   )
+ ELSE()
+   OPTION( ASSIMP_BUILD_ZLIB
+     "Build your own zlib"
+-    ON
```
Croydon, line 13: same here

jcar87:
closing as superseded by https://github.com/conan-io/conan-center-index/commit/df3ba9ab669394474b250ccfab072ca18992ae8e which includes the new version

Thanks for your contribution!

PR #20187, opened 2023-09-28
s2geometry: add recipe

alex-700:
Specify library name and version:  **s2geometry/0.10.0**

---

### DIFF
--- a/recipes/s2geometry/all/conanfile.py
+++ b/recipes/s2geometry/all/conanfile.py
@@ -1,0 +1,103 @@
+import os
+
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import get, copy, export_conandata_patches, apply_conandata_patches
+from conan.tools.microsoft import check_min_vs, is_msvc
+from conan.tools.scm import Version
+
+required_conan_version = ">=1.53.0"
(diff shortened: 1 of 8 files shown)


RubenRBS (approved):
Thanks a lot! The recipe looks great, and I see that the patches are needed, so approving this :)

czoido (approved):

recipes/s2geometry/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["s2"]
+        self.cpp_info.set_property("cmake_target_name", "s2")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.names["cmake_find_package"] = "s2"
+        self.cpp_info.names["cmake_find_package_multi"] = "s2"
```
SpaceIm, line 103: Please remove this. There is no config file upstream.

alex-700: IMHO, there is a reason to force `s2` target. While s2geometry does not provide `cmake` config file, the common targets-aware way to have this dependency is "building in a tree" (like submodule + `add_subdirectory` e.g.). In this case the target will be [`s2`](https://github.com/google/s2geometry/blob/master/CMakeLists.txt#L95). Saving the name allows users to follow this [one-line guideline](https://github.com/google/s2geometry/blob/master/CMakeLists.txt#L244).

PR #20214, opened 2023-09-29
Bump google-cloud-cpp's zlib requirement to [>=1.2.11 <2] range

RubenRBS:


RubenRBS:
Weird, seems like the graph conflict detection I used to generate this PR is faulty, seems like grpc should be moved to ranges first! I wonder why the script marked this as ready then

RubenRBS:
Oh my 🤦 
The last change I made was missing a negation - This level seems to have been recipes that did not have zlib as a direct dependency in the graph! No wonder this showed up before glib then :)

RubenRBS:
After re-running with proper logic, seems like grpc can now be opened - will re-trigger this one after https://github.com/conan-io/conan-center-index/pull/20223 is merged

czoido (approved):

coryan:
You may want to consider using #20721 instead.  There are other requirements (beyond `zlib`) that need bumps, and new patches are needed to compile the code [^1].  I expect that PR will build cleanly in this round, should be just a few more hours.


[^1]: I am guessing because the toolchain(s) and/or the base OS changed.

coryan:
> I expect that PR will build cleanly in this round, should be just a few more hours.

"a few" turned out to be about 13 hours. But it is now passing.

RubenRBS:
Closing in favour of https://github.com/conan-io/conan-center-index/pull/20721, thanks @coryan!

PR #20256, opened 2023-10-02
fmilib: add recipe for version 2.4.1

joakimono:
Specify library name and version:  **fmilib/2.4.1**

fmilibrary is a software package that enables integration of Functional Mock-up Units (FMUs). This library is an implementation of the [FMI open standard](https://fmi-standard.org/). 

This package is a dependency of the software provided by [Open Simulation Platform](https://opensimulationplatform.com/).

---


recipes/fmilibrary/all/conanfile.py
```diff
@@ -0,0 +1,131 @@
...
+        copy(self, "*.h", self.dependencies["fmi2"].cpp_info.includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI2"))
+
+        tc = CMakeToolchain(self)
+        tc.variables["FMILIB_BUILD_STATIC_LIB"] = not self.options.shared
+        tc.variables["FMILIB_BUILD_SHARED_LIB"] = self.options.shared
+        tc.variables["FMILIB_BUILD_TESTS"] = self.options.with_fmus
+        tc.variables["FMILIB_FMI_STANDARD_HEADERS"] = posixpath.join(self.build_folder, "fmis").replace("\\", "/")
+        tc.variables["FMILIB_GENERATE_DOXYGEN_DOC"] = False
+
```
kyllingstad, line 81: There is a variable [`FMILIB_BUILD_FOR_SHARED_LIBS`](https://github.com/modelon-community/fmi-library/blob/2.4.1/CMakeLists.txt#L48) which you might consider tying to the `fPIC` option. Like `fPIC`, it is only available when building a static library.

joakimono: Added this as suggested.

recipes/fmilibrary/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0026"] = "OLD"
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0045"] = "OLD"
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0046"] = "OLD"
+
+        tc.generate()
+        tc = CMakeDeps(self)
+        tc.generate()
+
+        vre = VirtualRunEnv(self)
+        vre.generate(scope="build")
```
uilianries, line 107:
```suggestion
```

VirtualRunEnv should not be needed as there is no build requirements in this recipe.

recipes/fmilibrary/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+        apply_conandata_patches(self)
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+        if not self.conf.get("tools.build:skip_test", default=True):
+            env = Environment()
+            env.define("CTEST_OUTPUT_ON_FAILURE", "ON")
+            with env.vars(self).apply():
+                cmake.test()
```
uilianries, line 119:
```suggestion
```
Please, try to avoid unit test in recipes. The main idea is packaging a project, not running its unit tests. But in case you really need/want it, use https://docs.conan.io/2/reference/tools/env/envvars.html#envvars with the CMakeToolchain, so you can use a separated script and does not propagate to your environment.

uilianries (commented):
Thank you for your pull request. I see many patches that could be sent to the upstream as well. Using find_package is not only a recommendation by Cmake part but also much safer than mimicking where artifacts are installed.

recipes/fmilibrary/all/conanfile.py
joakimono: For some reason this is needed for a successful build with MSVC for shared library.

recipes/fmilibrary/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+             path.join(self.build_folder, "fmis", "FMI1"))
+        copy(self, "fmiFunctions.h", self.dependencies["fmi1"].cpp_info.components["cosim"].includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI1"))
+        copy(self, "*.h", self.dependencies["fmi2"].cpp_info.includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI2"))
+
+        tc = CMakeToolchain(self)
+        tc.variables["FMILIB_BUILD_STATIC_LIB"] = not self.options.shared
+        tc.variables["FMILIB_BUILD_SHARED_LIB"] = self.options.shared
+        tc.variables["FMILIB_BUILD_TESTS"] = self.options.with_fmus
```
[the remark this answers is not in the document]
joakimono: Yes, this is correct. In order to build so-called FMUs (zips) to be part of package in the res folder, building tests must be enabled. They are used both for the unit tests in this project, but also for a (soon-to-appear on CCI) downstream application, FMU compliance checker.

recipes/fmilibrary/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+            raise ConanInvalidConfiguration(
+                f"{self.ref} does not support architecture "
+                f"'{self.settings.arch}' on {self.settings.os}")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+
+        copy(self, "fmiModel*.h", self.dependencies["fmi1"].cpp_info.components["modex"].includedirs[0],
```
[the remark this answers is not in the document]
joakimono: This approach essentially replicates what was originally done in fmilibrary. The build scripts assumes the header files of header-only libraries fmi1 and fmi2 to be located at specific locations in the source tree. It could be that using target_link_libraries works well, but if I recall correctly, this warrants for quite a few changes in the build scripts. I will take another look at this before deciding on what to do, either move to `build()` or add patches.

recipes/fmilibrary/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+        self.requires("fmi2/2.0.4")
+        if self.version >= Version("3.0a2"):
+            self.requires("fmi3/3.0.1")
+        self.requires("expat/2.4.9")
+        self.requires("minizip/1.2.13")
+        self.requires("zlib/[>=1.2.11 <2]")
+        # c99_snprintf -> should be externalised
+
+    def validate(self):
+        if self.settings.os == "Macos" and self.settings.arch == "armv8":
```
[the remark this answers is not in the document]
joakimono: I doubt it will be fixed for v2, but there is an v3 in alpha release where this is more relevant. I have created an issue upstream (https://github.com/modelon-community/fmi-library/issues/93), and will add it as a comment in the recipe for reference.

recipes/fmilibrary/all/conanfile.py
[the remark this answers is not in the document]
joakimono: Investigations reveal that target link libraries will not work unless several build script changes are done. Also, the include paths are different; all fmi includes must be changed as well. The header files in question are also bundled with this package, so if not copying them to the source tree, copying them in `package()` is still required. My suggestion is to not add patches, but move to `build()`.

I commit is coming.

recipes/fmilibrary/all/conandata.yml
```diff
@@ -0,0 +1,38 @@
...
+      patch_type: "conan"
+    - patch_file: "patches/2.4.1-005-ensure-conan-friendly-cmakelists.patch"
+      patch_description: "Do minor changes to CMakeLists for conan friendliness"
+      patch_type: "conan"
+    - patch_file: "patches/2.4.1-006-link-external-targets.patch"
+      patch_description: "Link external targets no longer merged in fmilib"
+      patch_type: "conan"
+    - patch_file: "patches/2.4.1-007-enable-testing.patch"
+      patch_description: "Ensure that tests can be run with cmake"
+      patch_type: "conan"
```
uilianries, line 29: This patch should be sent to the upstream too. It's not a port to work with Conan, but an improvement to the original project.

recipes/fmilibrary/all/conandata.yml
```diff
@@ -0,0 +1,38 @@
...
+    - patch_file: "patches/2.4.1-008-add-missing-stdlib.h-include.patch"
+      patch_description: "exit() is defined in stdlib.h"
+      patch_type: "conan"
```
uilianries, line 32:
```suggestion
    - patch_file: "patches/2.4.1-008-add-missing-stdlib.h-include.patch"
      patch_description: "exit() is defined in stdlib.h"
      patch_type: "conan"
```

Same, this patch should be sent to the upstream too. It's not a port to work with Conan, but an improvement to the original project.

recipes/fmilibrary/all/patches/2.4.1-009-link-zlib-in-fmilib.patch
```diff
@@ -0,0 +1,28 @@
...
+
+diff --git a/Config.cmake/fmizip.cmake b/Config.cmake/fmizip.cmake
+index 7618d51..213b5f8 100644
+--- a/Config.cmake/fmizip.cmake
++++ b/Config.cmake/fmizip.cmake
+@@ -42,9 +42,10 @@ set(FMIZIPHEADERS
+ #include_directories("${FMILIB_THIRDPARTYLIBS}/zlib/lib/VS2005/win32")
+ 
+ find_package(minizip CONFIG REQUIRED)
++find_package(ZLIB CONFIG REQUIRED)
```
uilianries, line 18: Is it a limitation from Conan zlib lib package, or the upstream is just missing it and should fix its cmake file instead?

recipes/fmilibrary/all/patches/2.4.1-010-install-runtime-to-bin.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index cc2a25c..0bb24f6 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -304,7 +304,7 @@ file(COPY "${FMILIBRARYHOME}/Config.cmake/fmilib.h" DESTINATION "${FMILibrary_BI
+ install(TARGETS ${FMILIB_TARGETS}
+ 	ARCHIVE DESTINATION lib
+ 	LIBRARY DESTINATION lib
+-	RUNTIME DESTINATION lib
++	RUNTIME DESTINATION bin
```
uilianries, line 19: I would prefer moving DLL from lib to bin, using `shutil.move` than creating a patch. Patches are not portable from a version to another in case that file change, and we will become maintainers of this file once merged.

uilianries (commented):
I'm really inclined to rename this package to `fmilib` instead of keeping `fmilibrary`. My concern is due the popularity over the name `fmilib`, used not only by other packages managers like AUR and Conda, but also I can find many references in headers like `FMILIB_EXPORT`.

recipes/fmilibrary/all/conanfile.py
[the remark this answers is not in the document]
uilianries: Thank you for checking it! I see there is no easy. It would be great if fmilib supports some option for external dependencies without all these tricks.

recipes/fmilibrary/all/conandata.yml
joakimono: This is no longer needed, since ctest is not run in the recipe. Will be removed.

recipes/fmilibrary/all/patches/2.4.1-009-link-zlib-in-fmilib.patch
joakimono: fmilib builds it internally and links it. Patch 005 removes this linking and patch 002 includes source code from minizip, which in turn calls zlib-functions..

recipes/fmilibrary/all/patches/2.4.1-010-install-runtime-to-bin.patch
joakimono: I will ensure this is done in the recipe instead.

recipes/fmilibrary/all/conandata.yml
joakimono: I will include a reference to patch_source.

joakimono:
Are there any actions I can take to move the review process further ahead? @uilianries
I believe I've addressed the reviewer comments, but there may be emerging comments I need to address?

recipes/fmilib/all/patches/2.4.1-009-make-project-pure-c.patch
```diff
@@ -0,0 +1,39 @@
+From dcf9863528e64edcad83ff0f9242b94b14df1705 Mon Sep 17 00:00:00 2001
```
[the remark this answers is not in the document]
joakimono: This patch makes sure that the library can be built with a C compiler only, which the upstream cannot, since it builds third party libraries that require CMake to look for C++ compiler. By applying this patch, we can remove libcxx and cppstd settings from the recipe. Looking at the upstream code, fmilib 3 will require c++ (at least building tests), so it can be dropped when fmilib 3 is released, and therefore no need to maintain it beyond fmilib 2.4.1.

recipes/fmilib/all/patches/2.4.1-007-add-missing-stdlib.h-include.patch
```diff
@@ -0,0 +1,23 @@
+From 79a37f7a6173f8bdd3324a0442ab377bcb6f238a Mon Sep 17 00:00:00 2001
```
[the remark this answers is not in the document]
joakimono: I have reported this upstream: https://github.com/modelon-community/fmi-library/issues/95
The reason for not avoiding to build test is that this is the only way to produce specific resources, which are needed by a downstream (to-be-submitted) package.

recipes/fmilib/all/patches/2.4.1-004-use-minizip-and-zlib-from-conan.patch
```diff
@@ -0,0 +1,55 @@
...
++  ${FMIZIPDIR}/src/minizip.c
++  ${FMIZIPDIR}/src/miniunz.c
+ )
+ 
+ set(FMIZIPHEADERS
+ #  src/fmi_zip_unzip_impl.h
+   ${FMIZIPDIR}/include/FMI/fmi_zip_unzip.h
+   ${FMIZIPDIR}/include/FMI/fmi_zip_zip.h
++  ${FMIZIPDIR}/include/minizip.h
++  ${FMIZIPDIR}/include/miniunz.h
```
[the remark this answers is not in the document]
joakimono: The third party minizip source that `fmilib` uses are modified so that `minizip` and `miniunz` are functions rather than executables. This makes it necessary to add both the source and header files, which have modified function signatures (`main -> mini{zip,unz}`)

recipes/fmilib/all/patches/2.4.1-006-link-external-targets.patch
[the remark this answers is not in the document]
joakimono: If `fmilib` is to be built without tests, the `with_fmus` option will not work, since the so-called fmu-resources are built as part of the tests. The point of adding `with_fmus` is that this will be needed by FMU Compliance Checker executable, which will soon arrive as a recipe as well.

recipes/fmilib/all/patches/2.4.1-007-add-missing-stdlib.h-include.patch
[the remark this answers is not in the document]
joakimono: I think the library will work without the resources being generated. However, these resources are needed for FMU-Compliance-Checker executable, which is planned to appear as a recipe as well.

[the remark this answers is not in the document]
joakimono: It is an executable that has `fmilib` as dependency: https://github.com/modelica-tools/FMUComplianceChecker .
However, the repository is archived, so I do not think I will include a recipe for that application on CCI.

With that in mind, I believe the option `with_fmus` is less relevant and can be set `False` by default. I still think it is useful to include in the res/ folder for consumers of `fmilib` that might want access to these resources.

recipes/fmilib/all/patches/2.4.1-004-use-minizip-and-zlib-from-conan.patch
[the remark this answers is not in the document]
joakimono: Yes, because `minizip.c` includes `zip.h`, and `miniunz.c` includes `unzip.h`; those files are provided by `minizip`.

recipes/fmilib/all/conanfile.py
```diff
@@ -0,0 +1,152 @@
...
+        # The variable is an option only if the following condition is true
+        if not self.options.shared and not self.settings.os in ["Windows", "Macos"]:
+            tc.variables["FMILIB_BUILD_FOR_SHARED_LIBS"] = self.options.get_safe("fPIC", False)
+
+        if is_msvc(self):
+            tc.variables["FMILIB_BUILD_WITH_STATIC_RTLIB"] = is_msvc_static_runtime(self)
+
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0026"] = "OLD"
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0045"] = "OLD"
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0046"] = "OLD"
```
valgur, line 93:
```suggestion
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0046"] = "OLD"
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
Needed because the project's CMake policy version is less than 3.12.

recipes/fmilib/all/conanfile.py
```diff
@@ -0,0 +1,152 @@
...
+
+        tc.generate()
+
+        cd = CMakeDeps(self)
+        cd.generate()
+
+        vre = VirtualRunEnv(self)
+        vre.generate(scope="build")
```
valgur, line 101: This is normally not required for CMake projects. What is the purpose of it?

recipes/fmilib/all/conandata.yml
```diff
@@ -0,0 +1,30 @@
...
+    sha256: "8199d3e9423494b714b9c4e42f055248457a7c9162df3d4652000aa9a10b8316"
+patches:
+  "2.4.1":
+    - patch_file: "patches/2.4.1-001-static-c99snprintf.patch"
+      patch_description: "Always build c99snprintf as static library"
+      patch_type: "bugfix"
+      patch_source: "https://github.com/open-simulation-platform/conan-fmilibrary/blob/master/build-static-c99snprintf.patch"
+    - patch_file: "patches/2.4.1-002-add-minizip-tools-as-functions.patch"
+      patch_description: "Add minizip and miniunz as functions instead of executables"
+      patch_type: "conan"
```
valgur, line 14: This is a bit much for CCI. It would be better suited as a patch for the upstream.

Why not simply rely on the executables?

recipes/fmilib/all/patches/2.4.1-003-use-expat-from-conan.patch
valgur, line 1: A `if(FALSE) ... endif()` would be preferable to keep the patch smaller and more maintainable.

recipes/fmilib/all/patches/2.4.1-003-use-expat-from-conan.patch
```diff
@@ -0,0 +1,110 @@
...
++include_directories("${FMILIB_THIRDPARTYLIBS}/FMI/" "${FMIXMLGENDIR}/FMI1" "${FMIXMLGENDIR}/FMI2")
+ 
+ PREFIXLIST(FMIXMLSOURCE  ${FMIXMLDIR}/)
+ PREFIXLIST(FMIXMLHEADERS ${FMIXMLDIR}/)
+@@ -232,6 +157,6 @@ if(MSVC)
+     target_compile_definitions(fmixml PUBLIC XML_STATIC)
+ endif()
+ 
+-target_link_libraries(fmixml ${JMUTIL_LIBRARIES} expat)
++target_link_libraries(fmixml ${JMUTIL_LIBRARIES} expat::expat)
```
valgur, line 105: This can probably be replaced with a `cd.set_property("expat", "cmake_target_name", "expat")`.

recipes/fmilib/all/patches/2.4.1-002-add-minizip-tools-as-functions.patch
valgur, line 1: Could these be downloaded directly from a repo instead? This is a bit excessive for CCI.

recipes/fmilib/all/patches/2.4.1-003-use-expat-from-conan.patch
joakimono: Applied your suggesting.

recipes/fmilib/all/patches/2.4.1-002-add-minizip-tools-as-functions.patch
joakimono: The source code is now downloaded from the same archive as the minizip dependency. I admit this is not a good solution, since the download url and checksum must be copied whenever there is a new version of minizip on CCI. Also the recipe now executes `replace_in_file` to perform necessary changes to the copied source code.

recipes/fmilib/all/patches/2.4.1-003-use-expat-from-conan.patch
joakimono: Done as you suggested.

recipes/fmilib/all/conanfile.py
joakimono: This is need when option `with_fmu=True`. Then, during the build process some executables need runtime access to what can be found in the virtual run environment.

recipes/fmilib/all/conanfile.py
joakimono: I added this to the newest commit.

recipes/fmilib/all/conanfile.py
```diff
@@ -0,0 +1,197 @@
...
+            "1.2.13":
+            {
+                "url": "https://zlib.net/fossils/zlib-1.2.13.tar.gz",
+                "sha256": "b3a24de97a8fdbc835b9833169501030b8977031bcb54b3b3ac13740f846ab30"
+            }
+        }
+        get(self,
+            **minizip_code[str(self.dependencies["minizip"].ref.version)],
+            pattern="*/minizip/*",
+            strip_root=True, destination=path.join(self.build_folder))
```
RubenRBS, line 122: Afaik, the library needs the actual source code for minizip instead of relying in the compiled library (Would appreciate insight into this too!)

You actually get access to the dependency's conandata thru the `self.dependencies` accessor, so this could be simplified to something like
 ```suggestion
        minizip_version = self.dependencies["minizip"].ref.version
        get(self,
            **self.dependencies["minizip"].conandata["sources"][minizip_version],
            pattern="*/minizip/*",
            strip_root=True, destination=path.join(self.build_folder))
```

which allivieates some of your concerns - I would still like some way to avoid this, but alas seems like this will be that way to go forward

**Review by RubenRBS — COMMENTED** [review]
Hi! Sorry for the delay. I have some concerns/suggestions and questions, would love to get some insight on some of the changes :)

File: recipes/fmilib/all/conanfile.py [hunk 19]
```diff
@@ -0,0 +1,197 @@
...
+        self._patch_sources()
+
+        copy(self, "fmiModel*.h", self.dependencies["fmi1"].cpp_info.components["modex"].includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI1"))
+        copy(self, "fmiPlatformTypes.h", self.dependencies["fmi1"].cpp_info.components["cosim"].includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI1"))
+        copy(self, "fmiFunctions.h", self.dependencies["fmi1"].cpp_info.components["cosim"].includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI1"))
+        copy(self, "*.h", self.dependencies["fmi2"].cpp_info.includedirs[0],
+             path.join(self.build_folder, "fmis", "FMI2"))
```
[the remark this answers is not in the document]
joakimono: There is no reason. I moved them to the source folder instead.

recipes/fmilib/all/conanfile.py
joakimono: `fmilib` introduce `minizip.c` and `minunz.c` as functions instead of applications by changing their `main` functions. To do this, `fmilib` needs the `minizip` sources. It also uses other functionality in `minizip`, so the precompiled library is also needed. In patch file `*002*.patch`, you can see the new header files for the patched functions.

joakimono:
> Hi! Sorry for the delay. I have some concerns/suggestions and questions, would love to get some insight on some of the changes :)
@AbrilRBS: Hi, have I addressed your suggestions and questions, or do you have further inquiries?

AbrilRBS:
Hi @joakimono sorry for the radio silence - I've been talking with @uilianries and we're checking some changes to the recipe to make it a bit more maintenable, sorry that I didn't communitcate it better!

joakimono:
> Hi @joakimono sorry for the radio silence - I've been talking with @uilianries and we're checking some changes to the recipe to make it a bit more maintenable, sorry that I didn't communitcate it better!

Hi @AbrilRBS, have you been able to check the changes for maintainability yet? I'm was hoping to get this recipe merged soon.

uilianries:
@joakimono Sorry our delay, I spent some hours analyzing fmi and it's really complicated to get unvendorized as I see, that's why you needed to apply so many patches. Only AUR (Arch Linux) and vcpkg are packaging fmilib as well, and both are not trying to unvendor everything due complexity/fragility. 

I'll bring this PR to be discussed with Conan team, but personally I'm more inclined to avoid all those patches related to unvendor. This is not the first recipe in CCI that we have this situation (see BehaviorTree). I would suggest discussing first with the upstream, to enable external dependencies instead via CMake options, it could be re-used by any user in the future. 

Using Conan package instead of those vendorized is good as we can control the dependencies, but not always the author wants it, we had a project using vendorized dependencies and the author did not want to use external instead, to keep all aligned and tested with that specific versions, so users would have the same behavior always.

joakimono:
Thank you for your reply. I agree that the option to unvendor dependencies is best handled by the upstream package. I posted an issue on this, https://github.com/modelon-community/fmi-library/issues/100, last November, but I do not think it is prioritized. I will see if I can provide a pull request to fmi-library to achieve it, perhaps it would expedite matters.

uilianries:
@joakimono Thank you for your feedback and pointing that issue. I'll try to take a look during this weekend as side-project too. 😄

uilianries:
I sent the first PR to separate Expat this weekend: https://github.com/modelon-community/fmi-library/pull/134

I think the best way is sending separated PRs, otherwise it would be hard not only to be tested, but also reviewed, as there are several dependencies. Once accepted/merged, I can send a next PR.

PR #20264, opened 2023-10-02
cyrus-sasl: bump libpq to 14.9

czoido:
This PR: https://github.com/conan-io/conan-center-index/pull/19653/files removed the 14.7 version and left cyrus-sasl depending on an unmaintained version in CC, bumping to 14.9


RubenRBS (approved):

recipes/cyrus-sasl/all/conanfile.py
```diff
@@ -80,7 +80,7 @@ def requirements(self):
         if self.options.with_openssl:
             self.requires("openssl/[>=1.1 <4]")
         if self.options.with_postgresql:
-            self.requires("libpq/14.7")
+            self.requires("libpq/14.9")
```
SpaceIm, line 83:
```suggestion
            self.requires("libpq/15.4")
```

PR #20266, opened 2023-10-02
mpg123: fix missing glibc symbols on newer systems

valgur:


SpaceIm:
Honestly, I don't think it's responsibility of conancenter recipes to check this. There should be a settings in conan allowing to tell what is glibc version.

valgur:
The fix worked and the PR is ready for merging.

ericLemanissier (approved):

ericLemanissier (approved):

RubenRBS:
> There should be a settings in conan allowing to tell what is glibc version.

@SpaceIm this is something we're carefully considering :)

czoido (approved):

PR #20267, opened 2023-10-02
aeron: fix missing glibc symbols on newer systems

valgur:


valgur:
The fix worked and the PR is ready to be merged. Thanks @czoido!

ericLemanissier (approved):

czoido (approved):

PR #20275, opened 2023-10-02
pulseaudio: fix missing glibc symbols on newer systems

valgur:

Depends on:
- #20266
- #20270


valgur:
@uilianries @RubenRBS This build has been restarting over and over again for 8 times already today. You should probably look into it.

RubenRBS:
Sorry for the spam @valgur, will check what's triggering the builds so often

RubenRBS:
Ok I've discovered the root cause and we're working on fixing the issue, it's some combination of unlucky timing and maintenance mode checks for v2 builds. I've restarted the build, shouldn't happen again

valgur:
The fix works. I created a separate PR to have something to actually merge: #20732

PR #20276, opened 2023-10-02
soundtouch: fix missing glibc symbols on newer systems

valgur:

### DIFF
--- a/recipes/soundtouch/all/conanfile.py
+++ b/recipes/soundtouch/all/conanfile.py
@@ -54,8 +54,10 @@ class SoundTouchConan(ConanFile):
         tc.cache_variables["OPENMP"] = self.options.with_openmp
         # The finite-math-only optimization has no effect and can cause linking errors
         # when linked against glibc >= 2.31
-        tc.variables["CMAKE_C_FLAGS"] = "-fno-finite-math-only"
-        tc.variables["CMAKE_CXX_FLAGS"] = "-fno-finite-math-only"
+        tc.blocks["generic_system"].template = tc.blocks["cmake_flags_init"].template + """
+               string(APPEND CMAKE_CXX_FLAGS_INIT " -fno-finite-math-only")
+               string(APPEND CMAKE_C_FLAGS_INIT " -fno-finite-math-only")
+           """
         tc.generate()
 
     def build(self):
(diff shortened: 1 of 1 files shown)


ericLemanissier (approved):

RubenRBS (changes requested):
Just realized with @czoido that this would overwrite CXX flags, we'll submit a better way to do this

recipes/soundtouch/all/conanfile.py
```diff
@@ -48,10 +48,14 @@ def source(self):
...
-        tc.variables["SOUNDSTRETCH"] = self.options.with_util
-        tc.variables["OPENMP"] = self.options.with_openmp
+        tc.cache_variables["INTEGER_SAMPLES"] = self.options.integer_samples
+        tc.cache_variables["SOUNDTOUCH_DLL"] = self.options.with_dll
+        tc.cache_variables["SOUNDSTRETCH"] = self.options.with_util
+        tc.cache_variables["OPENMP"] = self.options.with_openmp
+        # The finite-math-only optimization has no effect and can cause linking errors
+        # when linked against glibc >= 2.31
+        tc.variables["CMAKE_C_FLAGS"] = "-fno-finite-math-only"
+        tc.variables["CMAKE_CXX_FLAGS"] = "-fno-finite-math-only"
```
czoido, line 58: This would be a more appropriate way of adding extra flags. Because using `tc.variables`, you are overwriting all the all the flags set by the Conan toolchain. Maybe we should provide an interface in CMakeToolchain to make this more straightforward. 
```suggestion
        tc.blocks["generic_system"].template = tc.blocks["cmake_flags_init"].template + """
               string(APPEND CMAKE_CXX_FLAGS_INIT " -fno-finite-math-only")
               string(APPEND CMAKE_C_FLAGS_INIT " -fno-finite-math-only")
           """
```

RubenRBS (approved):

ericLemanissier (approved):

recipes/soundtouch/all/conanfile.py
valgur: Thanks a lot for a working solution! I agree, though. Dedicated toolchain variables similar to `AutotoolsToolchain`'s `extra_cflags`, `extra_ldflags` would be very welcome. This solution has almost zero discoverability in the API.

RubenRBS (approved):

PR #20278, opened 2023-10-02
hwdata: Add recipe

jwillikers:
Add the [hwdata](https://github.com/vcrhonek/hwdata) package.
This package is required to create a wlroots Conan package.

Specify library name and version:  **hwdata/0.374**

---


RubenRBS (dismissed):

RubenRBS:
These are the v2 errors, similar to v1 https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/20278/2-windows-msvc/hwdata/0.374//summary.json

jwillikers:
> These are the v2 errors, similar to v1 https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/20278/2-windows-msvc/hwdata/0.374//summary.json

Thanks! Looks like a simple-enough fix.

recipes/hwdata/all/conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class HwDataConan(ConanFile):
+    name = "hwdata"
+    description = "hwdata contains various hardware identification and configuration data, such as the pci.ids and usb.ids databases"
+    license = "GPL-2.0-or-later OR XFree86-1.1"
```
uilianries, line 15:
```suggestion
    license = ("GPL-2.0-or-later", "XFree86-1.1")
```
Each license should a separated item, so we can filter them by conan inspect command

RubenRBS: Great catch

jwillikers: No problem, but I do get the feeling will want to distinguish between `AND` and `OR` in the future. No need to have that discussion here though. I'll update that right now.

jwillikers: Fixed.

RubenRBS: I agree, thanks for the heads up

recipes/hwdata/all/test_package/conanfile.py
```diff
@@ -0,0 +1,36 @@
+from conan import ConanFile
```
uilianries, line 1: @jwillikers I would suggest simplifying this test package to only check if those files are installed. The hwdata package only contains "text" files, nothing to be linked actually, so you could check only if are installed and it should be enough 😸

uilianries (dismissed):

recipes/hwdata/all/test_package/conanfile.py
```diff
@@ -0,0 +1,39 @@
...
+        pkg_config_deps = PkgConfigDeps(self)
+        if self._settings_build:
+            pkg_config_deps.build_context_activated = ["hwdata"]
+        pkg_config_deps.generate()
+        virtual_build_env = VirtualBuildEnv(self)
+        virtual_build_env.generate()
+
+    def build(self):
+        pkg_config = PkgConfig(self, "hwdata", self.generators_folder)
+        self.output.info(pkg_config.variables["pkgdatadir"])
```
uilianries, line 35:
```suggestion
```
Please, avoid adding logs to the output. There is an weird behavior with the CI that affects the json parser and breaks the build 😞

uilianries (commented):
@jwillikers it seems like you reach a bug in C3I, where there is no explicit build profile for conan test command. We will fix it, but please, reduce your test package to only check if those files are installed meanwhile, otherwise, this PR will be blocked.

recipes/hwdata/all/test_package/conanfile.py
jwillikers: Sorry, that was just there for local debugging. I accidentally committed it.

jwillikers:
> @jwillikers it seems like you reach a bug in C3I, where there is no explicit build profile for conan test command. We will fix it, but please, reduce your test package to only check if those files are installed meanwhile, otherwise, this PR will be blocked.

@uilianries I think I've simplified the test package like you were asking. It checks the PkgConfig file as a reqular requires instead of a tool_requires now. It just checks that the pkg-config is found correctly and that custom variable that is expected is defined.

uilianries:
> > @jwillikers it seems like you reach a bug in C3I, where there is no explicit build profile for conan test command. We will fix it, but please, reduce your test package to only check if those files are installed meanwhile, otherwise, this PR will be blocked.
> 
> @uilianries I think I've simplified the test package like you were asking. It checks the PkgConfig file as a reqular requires instead of a tool_requires now. It just checks that the pkg-config is found correctly and that custom variable that is expected is defined.

I meant, only checking installed files, nothing related to load them with pkgconfig

jwillikers:
> > > @jwillikers it seems like you reach a bug in C3I, where there is no explicit build profile for conan test command. We will fix it, but please, reduce your test package to only check if those files are installed meanwhile, otherwise, this PR will be blocked.
> > 
> > 
> > @uilianries I think I've simplified the test package like you were asking. It checks the PkgConfig file as a reqular requires instead of a tool_requires now. It just checks that the pkg-config is found correctly and that custom variable that is expected is defined.
> 
> I meant, only checking installed files, nothing related to load them with pkgconfig

It's important that the custom variable exist, so I figured that should be checked in the test package. Is there another package that checks that pkg-config files just exist that I could pull from?

uilianries:
> It's important that the custom variable exist, so I figured that should be checked in the test package. Is there another package that checks that pkg-config files just exist that I could pull from?

Nope, but just add a comment in the conanfile.py like `# INFO: pkgdatadir should not be removed. It's consumed by pkgconfig generator.`. I believe it will be enough.

jwillikers:
> > It's important that the custom variable exist, so I figured that should be checked in the test package. Is there another package that checks that pkg-config files just exist that I could pull from?
> 
> Nope, but just add a comment in the conanfile.py like `# INFO: pkgdatadir should not be removed. It's consumed by pkgconfig generator.`. I believe it will be enough.

Should I remove the test package then?

recipes/hwdata/all/test_package/conanfile.py
```diff
@@ -0,0 +1,32 @@
...
+        pkg_config_deps.generate()
+        virtual_build_env = VirtualBuildEnv(self)
+        virtual_build_env.generate()
+
+    def build(self):
+        pkg_config = PkgConfig(self, "hwdata", self.generators_folder)
+        assert len(pkg_config.variables["pkgdatadir"]) > 0
+
+    def test(self):
+        pass
```
uilianries, line 32:
```suggestion
from conan.tools.files import save, load
import os


class TestPackageConan(ConanFile):
    settings = "os", "arch", "compiler", "build_type"
    test_type = "explicit"

    def requirements(self):
        self.requires(self.tested_reference_str)
        save(self, "hwdata_pkg_dir", self.dependencies[self.tested_reference_str].package_folder)

    def test(self):
        pkg_dir = load(self, "hwdata_pkg_dir")
        assert os.path.isfile(os.path.join(pkg_dir, "res", "hwdata", "usb.ids"))
```

jwillikers: Thanks @uilianries!

uilianries (approved):
LGTM

jwillikers:
@ericLemanissier Would you mind adding your review real quick?

czoido (approved):

PR #20283, opened 2023-10-02
numpy: add new recipe

### LINKED ISSUE #19961: [request] NumPy/1.26.0
### Package Name/Version

NumPy/1.26.0

### Webpage

https://numpy.org/

### Source code

https://github.com/numpy/numpy

NumPy is the fundamental package for scientific computing with Python.

This looks buildable with meson given that it has a [meson.build](https://github.com/numpy/numpy/blob/main/meson.build) file in the root. This is a dependency of carma (#19959)

valgur:
Specify library name and version:  **numpy/1.26.4**


Ahajha:
Conan 2 error is a version conflict, I can put out the necessary PR(s) to fix. Conan 1 error seems to only happen on GCC 9, but not 11 (I'll check 10), not quite sure why yet but at least I can reproduce it locally.

Ahajha:
Little bit of progress: Version range was fixable by just updating to the maintained patch version of cpython (related, maybe we should use version ranges for that? anyways...). The issue that I'm running into on Windows is that it's looking for `python3.lib`, which exposes potentially a few issues:
1. python3.lib isn't even packaged (only python3(minor).lib) - easy fix
2. The source of `python3.lib` is from a `#pragma comment(lib, ...)`, which in my personal experience generally just cause headaches when trying to do things properly. It uses `python3.lib` specifically when using the Python Limited API.
3. Unsure if this is Meson specific, but probably not since a few quick hacks to the cpython package can cause a similar failure.

All in all I think it's (mostly) a cpython package issue, though removing the `Py_LIMITED_API` definition still causes it to fail with a missing `python3.10.lib`, so maybe there's some other stuff going on.

Ahajha:
Added a workaround for a GCC bug (see here: https://github.com/numpy/numpy/issues/16970), I've tested it locally but it now all legacy GCC versions get skipped: numpy itself requires 8.4+, and cpython doesn't work with GCC 9.

Ahajha:
@RubenRBS, any idea why the v2 pipeline is failing? As far as I can tell no `ConanInvalidConfiguration` errors are being thrown. Also need a binary for the v1 pipeline.

Ahajha:
@valgur BTW I believe this should work on all platforms, in case you wanted to add it to your new recipe repo. I've at least tested locally with Windows/VS2022/x64/Release + Linux/GCC(several)/x64/Release + Mac/apple-clang(15)/armv8/Release.

valgur:
.

PR #20287, opened 2023-10-02
Bump nmos-cpp's zlib requirement to [>=1.2.11 <2] range

RubenRBS:


recipes/nmos-cpp/all/conanfile.py
```diff
@@ -65,7 +67,7 @@ def requirements(self):
 
     def build_requirements(self):
         # nmos-cpp needs CMake 3.17 or higher but CCI doesn't allow version ranges
-        self.build_requires("cmake/3.24.2")
+        self.build_requires("cmake/[>=3.17 <4]")
```
SpaceIm, line 70:
```suggestion
        self.tool_requires("cmake/[>=3.17 <4]")
```

recipes/nmos-cpp/all/conanfile.py
```diff
@@ -65,7 +67,7 @@ def requirements(self):
 
     def build_requirements(self):
         # nmos-cpp needs CMake 3.17 or higher but CCI doesn't allow version ranges
```
SpaceIm, line 69:
```suggestion
```

This comment is not relevant anymore if you use version range.

SpaceIm:
There is also this code to remove:

```python
    def package_id(self):
        self.info.requires["boost"].minor_mode()
        self.info.requires["nlohmann_json"].patch_mode()
```

recipes/nmos-cpp/all/conanfile.py
uilianries: Done.

recipes/nmos-cpp/all/conanfile.py
uilianries: Done.

uilianries:
@SpaceIm Thank you for your review, all suggestions are applied now.

@RubenRBS I started a Tapaholes to generate `avahi/0.8`

jwillikers (approved):

uilianries:
I just rebuilt `avahi/0.8` internally and now restarted this PR.

RubenRBS:
Relaunched again, this time due to libselinux

czoido (approved):

jwillikers (approved):

uilianries:
Avahi missing again?! I'll check it further.

uilianries:
I created another PR (https://github.com/conan-io/conan-center-index/pull/20706) that should fix avahi. That's why we still have it missing.

uilianries (approved):

PR #20290, opened 2023-10-02
libsystemd: Include libudev

jwillikers:
Specify library name and version:  **libsystemd/***

Include the udev library as a component of the libsystemd package.
Update Meson version.
Add `provides = "libudev"` to allow other providers of the `libudev` library, i.e. `eudev`.
It seems like the `provides` should be related to the `libudev` component instead of the entire package.
Is this possible?

I wasn't sure whether to add a separate package for libudev, i.e. something under `libudev/all`, or to include it as part of the libsystemd package.
In light of the fact that a separate package requires more resources to maintain, I opted for including it in the libsystemd package.
This will require only updating one package in CCI instead of two, at the cost of flexibility for managing libudev and libsystemd in distinct ways, i.e. trying to build one shared and the other static.

---


RubenRBS:
> It seems like the provides should be related to the libudev component instead of the entire package.
Is this possible?

Not as far as I know, no. I've added this recipe to our next team review meeting, will let you

jwillikers:
Looks like it is most likely a compiler bug causing the compilation error and not an actual issue with the project. I wonder if it's the version of GCC 9 in CI causing problems.

jwillikers:
Yep, works with GCC 9.4. I'll add a workaround similar to Qt 6.

recipes/libsystemd/all/conanfile.py
```diff
@@ -68,9 +69,17 @@ def requirements(self):
     def validate(self):
         if self.settings.os != "Linux":
             raise ConanInvalidConfiguration("Only Linux supported")
+        if os.getenv('NOT_ON_C3I', '0') == '0' and Version(self.version) >= "253.10" and self.settings.compiler == "gcc" and Version(self.settings.compiler.version) == "9":
```
uilianries, line 72: I would suggest using config instead, much safer than env vars, and be customized by package. It could be `user.c3i.skip:libsystemd=True`, so in case needing this workaround again (IIRC we already used it in the past) we just add a new package name, or a wild card like `*`.


jwillikers: That sounds great to me!

jwillikers: Done.

valgur (approved):

jwillikers:
This PR is necessary for #20358, which requires a newer version of `libudev` than is available in CI.

PR #20297, opened 2023-10-03
nanobind: add a new recipe

### LINKED ISSUE #20095: [request] nanobind
### Package Name/Version

nanobind

### Webpage

https://github.com/wjakob/nanobind

### Source code

https://github.com/wjakob/nanobind

nanobind is a small binding library that exposes C++ types in Python and vice versa. It is reminiscent of [Boost.Python](https://www.boost.org/doc/libs/1_64_0/libs/python/doc/html) and [pybind11](https://github.com/pybind/pybind11) and uses near-identical syntax. In contrast to these existing tools, nanobind is more efficient: bindings compile in a shorter amount of time, produce smaller binaries, and have better runtime performance.

More concretely, [benchmarks](https://nanobind.readthedocs.io/en/latest/benchmark.html) show up to ~4× faster compile time, ~5× smaller binaries, and ~10× lower runtime overheads compared to pybind11. nanobind also outperforms Cython in important metrics (3-12× binary size reduction, 1.6-4× compilation time reduction, similar runtime performance).

### LINKED ISSUE #26592: [request] nanobind/2.5.0
### Package Name/Version

nanobind/2.5.0

### Webpage

https://github.com/wjakob/nanobind

### Source code

https://github.com/wjakob/nanobind

https://github.com/wjakob/nanobind/blob/master/README.md

valgur:
Specify library name and version:  **nanobind/***

* Closes #20095
* Closes #26592


valgur:
@Ahajha The issue is caused by the Conan package not exporting the `Development.Module` component and others, I think? Should this already be supported by the cpython recipe?

Ahajha:
@valgur The recipe currently only exports `cpython::python`, `cpython::embed`, and `cpython::_hidden`, and to top it off the CMake config file is named `cpython`, so the built-in `FindPython` knows nothing about it. It needs to be fed some hints in order to find the Conan Python, otherwise it will just look for a system one.

Ahajha:
RE: https://github.com/conan-io/conan-center-index/pull/25500#issuecomment-2423550874

I don't think it's going to be easy (or even possible in its current state) to fix the cpython Mac cross-builds(unless we cheat and call a universal build a cross build. Wouldn't be the worst thing I suppose). That could probably be added as an invalid config for this recipe in the short term and we can revisit if/when we find a fix. Perhaps when C3I allows additional configs to be added, that could help the situation.

The Windows issues (in your PR) are certainly fixable though.

Ahajha:
As a dumb idea - can you do something like
```python
def requirements(self):
    try:
        self.requires("cpython/whatever"...)
        self.should_test = True
    catch ConanInvalidConfiguration:
        self.should_test = False

def test(self):
    if self.should_test:
        self.run(whatever...)
```

Doesn't have to be anything like this, but essentially just limit the test package to requiring cpython when it is valid, if it isn't then don't build or test anything.

hasB4K:
Hello @valgur, do you plan to continue your work on this PR? nanobind is very useful since its the successor of pybind11.

Ahajha:
If not, I can probably work on getting it to the finish line. I think he's in a similar boat where long review times get frustrating, so there's not much motivation to work on things.

valgur:
I actually pushed an updated version to my dev branch recently, but have not gotten around to updating this PR.
It would not matter much anyway since it's one of the ~200 PRs I guess I would have to close to appease the otherwise great new CI pipeline deployed in November, to get a passing build and any subsequent reviews on the PR.

valgur:
.

valgur:
Logs for GCC armv8 cross-compilation:
- static: [nanobind.log](https://github.com/user-attachments/files/18812065/nanobind.log)
- shared: [nanobind.log](https://github.com/user-attachments/files/18812068/nanobind.log)

PR #20319, opened 2023-10-03
hayai: add a new recipe

valgur:
Adds https://github.com/nickbruun/hayai, a C++ benchmarking framework

### DIFF
--- a/recipes/hayai/all/conanfile.py
+++ b/recipes/hayai/all/conanfile.py
@@ -45,6 +45,9 @@ class HayaiConan(ConanFile):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_HAYAI_TESTS"] = False
+        tc.cache_variables["BUILD_HAYAI_SAMPLES"] = False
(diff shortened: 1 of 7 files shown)


recipes/hayai/all/conanfile.py
```diff
@@ -0,0 +1,61 @@
...
+required_conan_version = ">=1.53.0"
+
+
+class HayaiConan(ConanFile):
+    name = "hayai"
+    description = "C++ benchmarking framework"
+    license = "Apache"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/nickbruun/hayai"
+    topics = ("benchmarking", "header-only")
```
RubenRBS, line 16: `header-only`?

RubenRBS (commented):
Thanks! One question about the topic of header-only thought

recipes/hayai/all/conanfile.py
valgur: Oops. Thanks! Fixed it.

danimtb (approved):

SSE4 (approved):

PR #20321, opened 2023-10-03
libsgp4: add v1.0

valgur:


RubenRBS:
Hi @valgur! Thanks a lot for your enthusiasm with keeping ConanCenterIndex with updated versions, but so many at once are overloading our CI systems, as we're also trying to build the older PRs to finally get rid of the backlog, and this is creating lots of contention for our resources.

Could you please refrain from opening bump dependencies/new version PRs until we've processed the rest? You currently have 250 PR opened that are advancing slowly, and we would much rather focus on those before creating new ones! We feel that adding new versions of libraries in big bursts is mostly counterproductive, as both our review bandwidth and CI capacity get quickly saturated and just makes the backlog grow even longer

(Normal PRs that have recipe maintenance and/or bugfixing are still ok though!)

Thanks!

RubenRBS (approved):

jwillikers (approved):

PR #20324, opened 2023-10-03
libdisplay-info: Add recipe

jwillikers:
Specify library name and version:  **libdisplay-info/0.1.1**

`libdisplay-info` is required to create a `wlroots` Conan package.

Requires #20278.

Note that I might need to disable Windows / macOS builds if they don't work. I didn't see anything that wouldn't allow them to build there, so I left them enabled at first.

---

### DIFF
--- a/recipes/libdisplay-info/all/conanfile.py
+++ b/recipes/libdisplay-info/all/conanfile.py
@@ -18,7 +18,7 @@ class LibdisplayInfoConan(ConanFile):
     license = "MIT"
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://gitlab.freedesktop.org/emersion/libdisplay-info"
-    topics = ("display", "DisplayID", "EDID", "topic2", "topic3")
+    topics = ("display", "DisplayID", "EDID")
     package_type = "library"
     settings = "os", "arch", "compiler", "build_type"
     options = {
--- a/recipes/libdisplay-info/all/conanfile.py
+++ b/recipes/libdisplay-info/all/conanfile.py
@@ -1,6 +1,5 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
-from conan.tools.apple import is_apple_os
 from conan.tools.build import cross_building
 from conan.tools.env import VirtualBuildEnv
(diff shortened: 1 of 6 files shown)


valgur (approved):

recipes/libdisplay-info/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+required_conan_version = ">=1.53.0"
+
+
+class LibdisplayInfoConan(ConanFile):
+    name = "libdisplay-info"
+    description = "EDID and DisplayID library."
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://gitlab.freedesktop.org/emersion/libdisplay-info"
+    topics = ("display", "DisplayID", "EDID", "topic2", "topic3")
```
valgur, line 21:
```suggestion
    topics = ("display", "DisplayID", "EDID")
```

valgur (approved):
LGTM (except for the minor topics typo).

recipes/libdisplay-info/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+        if not self._has_build_profile:
+            self.requires("hwdata/0.374")
+
+    def validate(self):
+        if not self.settings.os in ["FreeBSD", "Linux"]:
+            raise ConanInvalidConfiguration(f"{self.ref} is not supported on {self.settings.os}")
+
+    def build_requirements(self):
+        if self._has_build_profile:
+            self.tool_requires("hwdata/0.374")
```
uilianries, line 60:
```suggestion
            self.tool_requires("hwdata/<host_version>")
```

https://docs.conan.io/2/reference/conanfile/methods/build_requirements.html#host-version

uilianries (commented):
looking good, just. a minor

recipes/libdisplay-info/all/conanfile.py
jwillikers: I don't think that will work in this case since `hwdata` is only a required in the `host` context for single-profile builds, otherwise it is only a `tool_requires`.

uilianries (approved):
LGTM

RubenRBS (approved):

PR #20336, opened 2023-10-03
boost: remove `b2-default` value of `layout` option

SpaceIm:
logic of this value in configure() can't work in conan v2, it raises a ConanException (see https://cpplang.slack.com/archives/C41CWV9HA/p1696374024080289)

If folks want default layout for Windows, they can set layout to versioned instead of this removed b2-layout, so this value was more or less useless.

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -81,7 +81,7 @@ class BoostConan(ConanFile):
         "filesystem_no_deprecated": [True, False],
         "filesystem_use_std_fs": [True, False],
         "filesystem_version": [None, "3", "4"],
-        "layout": ["system", "versioned", "tagged", "b2-default"],
+        "layout": ["system", "versioned", "tagged"],
         "magic_autolink": [True, False],  # enables BOOST_ALL_NO_LIB
         "diagnostic_definitions": [True, False],  # enables BOOST_LIB_DIAGNOSTIC
         "python_executable": [None, "ANY"],  # system default python installation is used, if None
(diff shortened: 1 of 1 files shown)


SpaceIm:
See also https://github.com/conan-io/conan/issues/14868 for explanations

RubenRBS (changes requested):
This is going to change moving forward - We'll need to validate the options instead of forcing them, as per your conversation in the Conan repo issue, this is no longer valid in v2

SpaceIm:
I don't understand. As explained, this value was a mistake actually. Users can still change layout, so what's the point of this value?

jwillikers (approved):
Yeah, that looks like it's effectively a dead value for that option.

SSE4:
@RubenRBS can you unblock?

from my point of view, `b2-default` may produce missed expectations. e.g. if you build for Android on Windows host vs on Linux host, you will get different library names for the same package ID (which is kinda weird).

valgur (approved):

uilianries (commented):
It looks good, considering:

The `b2-default` is a virtual auto value for Boost project, the same does not exist in Boost project, only `system`, `versioned` and `tagged` are available: https://github.com/boostorg/boost/blob/e4b2e2d6a11c9596a022ddfaa49c7a3c6261e297/boostcpp.jam#L64

It's true in Windows, Boost uses `versioned` by default, but in this recipe we have been enforced the `system` by default to any platform.

Plus, I'm not confident in auto values, as the upstream may change in future and we miss it when reviewing. 

I just talked to @AbrilRBS first, so I'll wait her thoughts first.

AbrilRBS (approved):

uilianries (approved):

PR #20335, opened 2023-10-03
eudev: Add recipe

jwillikers:
Specify library name and version:  **eudev/3.2.12**

[eudev](https://github.com/eudev-project/eudev) is an alternative implementation of systemd's libudev library.
This package indicates this with `provides = "libudev"`.
A `libblkid` Conan package is missing, but should be added and default enabled in this package. Here's an attempt at that:  #22426

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20320, opened 2023-10-03
mtdev: Add recipe

jwillikers:
Specify library name and version:  **mtdev/1.1.6**

The mtdev package is required to create a libinput package.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20318, opened 2023-10-03
libevdev: Add recipe

jwillikers:
Specify library name and version:  **libevdev/1.13.1**

The libevdev library is required to make a libinput package.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20346, opened 2023-10-04
libinput: Add recipe

jwillikers:
Specify library name and version:  **libinput/1.25.0**

[libinput](https://www.freedesktop.org/wiki/Software/libinput/) is a library that handles input devices for display servers and other applications that need to directly deal with input devices.

libinput is required to enable the libinput backend in wlroots.

The following PRs are required first:

- [x] #20335
- [x] #20320
- [x] #20318

Optionally, a libwacom package could be added to enable libwacom support.

---

### DIFF
--- a/recipes/libinput/all/conanfile.py
+++ b/recipes/libinput/all/conanfile.py
@@ -1,0 +1,133 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import copy, get, rmdir
+from conan.tools.gnu import PkgConfigDeps
+from conan.tools.layout import basic_layout
+from conan.tools.meson import Meson, MesonToolchain
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class LibinputConan(ConanFile):
+    name = "libinput"
+    description = "libinput is a library that handles input devices for display servers and other applications that need to directly deal with input devices."
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://www.freedesktop.org/wiki/Software/libinput/"
+    topics = ("device", "display", "event", "input")
+    package_type = "shared-library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "epoll_dir": [None, "ANY"],
+        "debug_gui": [True, False],
+        "with_libudev": ["eudev", "systemd"],
+        "with_libwacom": [True, False],
+        "with_wayland": [True, False],
+        "with_x11": [True, False],
+    }
+    default_options = {
+        "epoll_dir": None,
+        "debug_gui": True,
+        "with_libudev": "systemd",
+        # todo Package libwacom and enable this option by default.
+        "with_libwacom": False,
+        "with_wayland": True,
+        "with_x11": True,
+    }
+
+    def configure(self):
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+        if not self.options.debug_gui:
+            self.options.rm_safe("with_wayland")
+            self.options.rm_safe("with_x11")
+        if self.options.get_safe("with_wayland"):
+            self.options["gtk"].with_wayland = True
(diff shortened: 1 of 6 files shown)


recipes/libinput/all/conanfile.py
```diff
@@ -0,0 +1,127 @@
...
+            self.requires("cairo/1.18.0")
+            self.requires("glib/2.78.3")
+            self.requires("gtk/system")
+            if self.options.with_wayland:
+                self.requires("wayland/1.22.0")
+                self.requires("wayland-protocols/1.33")
+            if self.options.with_x11:
+                self.requires("xorg/system")
+
+        if self.options.with_libudev == "systemd":
```
RubenRBS, line 65: Should we remove/validate out this until it's supported?

recipes/libinput/all/conanfile.py
```diff
@@ -0,0 +1,127 @@
...
+        if self.settings.os not in ["FreeBSD", "Linux"]:
+            raise ConanInvalidConfiguration(f"{self.ref} is not supported on {self.settings.os}.")
+        if self.options.with_libwacom:
+            raise ConanInvalidConfiguration(f"The with_libwacom option for {self.ref} is not yet supported. Contributions welcome.")
+
+    def build_requirements(self):
+        self.tool_requires("meson/1.3.2")
+        if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
+            self.tool_requires("pkgconf/2.1.0")
+        if self.options.get_safe("with_wayland"):
```
RubenRBS, line 82: Is wayland needed when `debug_gui` is false?

RubenRBS (commented):
Hi! Step by step we have finally merged the blocking PRs so good to finally be able to come back to this one, I have some initial questions about this recipe, but overall looks great :)

recipes/libinput/all/conanfile.py
jwillikers: I think I just need to remove the comment to avoid confusion. I need to figure out how to get libudev building from source and come back to this. The `systemd` value still makes sense since `libudev/system` is just referring to the system package, which is just libudev from the systemd project.

recipes/libinput/all/conanfile.py
jwillikers: Nope, nice catch!

recipes/libinput/all/conanfile.py
jwillikers: Comment removed.

recipes/libinput/all/conanfile.py
jwillikers: Fixed.

RubenRBS (approved):

ericLemanissier (approved):

jwillikers:
> FYI, I just tried this:
> 
> ```
> $ conan create . --version 1.25.0 -o debug_gui=True --build missing
> ...
> 
> ERROR: Version conflict: Conflict between expat/2.6.0 and expat/2.5.0 in the graph.
> Conflict originates from fontconfig/2.14.2
> 
> Run conan graph info with your recipe and add --format=html to inspect the graph errors in an easier to visualize way.
> ```

Oh, that's from `cairo` most likely. PR #22696 should fix that or at least one of the conflicts.

recipes/libinput/all/conanfile.py
```diff
@@ -0,0 +1,125 @@
...
+        self.requires("mtdev/1.1.6")
+        self.requires("libevdev/1.13.1")
+
+        if self.options.debug_gui:
+            self.requires("cairo/1.18.0")
+            self.requires("glib/2.78.3")
+            self.requires("gtk/system")
+            if self.options.with_wayland:
+                self.requires("wayland/1.22.0")
+                self.requires("wayland-protocols/1.33")
```
SpaceIm, line 61: wayland-protocols should be in build requirements

jwillikers: @SpaceIm Right you are! This slipped through the recipes I needed to update to do this. I'll take care of it in a follow up PR. Thanks!

jwillikers: PR #23301

PR #20373, opened 2023-10-05
libblkid: Add recipe

jwillikers:
Specify library name and version:  **libblkid/2.39**

libblkid is part of the util-linux project.
It is currently packaged as part of the libmount package, but I need to depend on the libblkid library specifically in the eudev recipe, so I'm creating a separate package for it.
This will be the third Conan package broken out from the util-linux project.
I'm not sure if these could somehow be brought together in the future to reduce the maintenance overhead.
Anyways, the libmount recipe should be updated after this is merged to depend on this libblkid package instead of bundling the library itself.

---

### DIFF
--- a/recipes/libblkid/all/conandata.yml
+++ b/recipes/libblkid/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "2.39":
+    url: "https://mirrors.edge.kernel.org/pub/linux/utils/util-linux/v2.39/util-linux-2.39.tar.xz"
+    sha256: "32b30a336cda903182ed61feb3e9b908b762a5e66fe14e43efb88d37162075cb"
--- a/recipes/libblkid/all/conanfile.py
+++ b/recipes/libblkid/all/conanfile.py
@@ -1,0 +1,73 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.files import copy, get, rm, rmdir
+from conan.tools.gnu import Autotools, AutotoolsToolchain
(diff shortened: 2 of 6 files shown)


RubenRBS:
> I'm not sure if these could somehow be brought together in the future to reduce the maintenance overhead.

I'm adding this to our next team discussion, I'll let you know if we come up with anything for this :)

jwillikers:
> > I'm not sure if these could somehow be brought together in the future to reduce the maintenance overhead.
> 
> I'm adding this to our next team discussion, I'll let you know if we come up with anything for this :)

Thanks!

RubenRBS:
Once concern: If I now have both this and libmount in the same graph, that will cause problems down the line, right?

jwillikers:
> Once concern: If I now have both this and libmount in the same graph, that will cause problems down the line, right?

@RubenRBS Dang, I think you're right. The only way to mitigate that in a separate recipe is if util-linux allows building `libmount` with an external `libblkid`, which I doubt it supports.

Unfortunately, I think it would probably require a larger refactor of components to properly support these kinds of packages. The util-linux project in particular has licenses that apply to different components, which would require specifying the license field per component. It would also be great to somehow tie the components to a options to enable or disable them, which would make it easier to control which components get enabled by particular projects.

recipes/libblkid/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "2.39":
```
valgur, line 2: Can be bumped to 2.39.3.

valgur (approved):

PR #20381, opened 2023-10-05
sdl: bump deps and modernize more for conan v2

SpaceIm:
- bump dependencies
- use `<host_version>` for wayland in build requirements
- sort methods by order of execution
- use version range for cmake
- fix several access to informations of dependencies (`self.dependencies[foo].options` instead of `self.options[foo]`, or robust collection of cpp_info in dependencies with components)
- remove `PKG_CONFIG_PATH`  workaround since `required_conan_version` is greater than 1.55.0 now
- simplify test package and use C instead of C++

---


RubenRBS (approved):

RubenRBS:
Missing package for pkgconf/2.0.3 in v2 - that's totally unexpected, will check

mayeut:
> Missing package for pkgconf/2.0.3 in v2 - that's totally unexpected, will check

@RubenRBS, the commit  f00f9308d65323edecec3360253fa91424506168 is still being built on master according to the status check in GitHub (#19497 pkgconf: less verbose package_info() + bump meson). Could this be the reason ?

#20393 seems stuck.

jwillikers (approved):

recipes/sdl/all/test_package/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
 cmake_minimum_required(VERSION 3.15)
-project(test_package)
+project(test_package LANGUAGES C)
 
 find_package(SDL2 REQUIRED CONFIG)
 
-add_executable(${PROJECT_NAME} test_package.cpp)
-target_link_libraries(${PROJECT_NAME} SDL2::SDL2main)
+add_executable(${PROJECT_NAME} test_package.c)
+target_link_libraries(${PROJECT_NAME} PRIVATE SDL2::SDL2main)
```
sykhro, line 7: This causes a hard error on our end because we consume SDL with`-o:h sdl/*:sdl2main=False`.

It can also cause errors with `sdl2main=True` because the final test application will end up with two copies of `main()`. Since the test package doesn't use it, I would just remove this.

sykhro: We also needed
```python
self.cpp_info.components["libsdl2"].defines.append("SDL_MAIN_HANDLED")
```
It seems like the install target produces `SDL_main.h` headers regardless of the CMake settings. Might be an upstream issue though

SpaceIm:
hum, now I have to rebase for something removed anyway in this PR :(

recipes/sdl/all/test_package/CMakeLists.txt
RubenRBS: @SpaceIm Are you implementing this? Looks like a good solution to me

SpaceIm: It's out of scope of this PR. It can be done in another one.

jwillikers (approved):

recipes/sdl/all/conanfile.py
```diff
@@ -321,17 +273,58 @@ def define_toolchain(self):
...
+        env.vars(self).save_script("sdl_library_path")
+
+    def _patch_sources(self):
+        cmakelists = os.path.join(self.source_folder, "CMakeLists.txt")
+        if self.settings.os == "Macos":
+            if self.options.iconv:
+                # If using conan-provided iconv, search for the symbol "libiconv_open"
+                replace_check = "check_library_exists(iconv libiconv_open"
+            else:
+                # When no tusing conan-provided icon, don't check for iconv at all
```
valgur, line 309:
```suggestion
                # When not using conan-provided iconv, don't check for iconv at all
```

recipes/sdl/all/conanfile.py
```diff
@@ -321,17 +273,58 @@ def define_toolchain(self):
...
+        PkgConfigDeps(self).generate()
+
+        env = Environment()
+        lib_paths = [lib for dep in all_deps for lib in dep.cpp_info.aggregated_components().libdirs]
+        env.define_path("LIBRARY_PATH", os.pathsep.join(lib_paths))
+        env.vars(self).save_script("sdl_library_path")
+
+    def _patch_sources(self):
+        cmakelists = os.path.join(self.source_folder, "CMakeLists.txt")
+        if self.settings.os == "Macos":
```
valgur, line 304:
```suggestion
        if is_apple_os(self):
```
Can be generalized to other Apple OS-s, perhaps.

recipes/sdl/all/conanfile.py
```diff
@@ -178,56 +171,25 @@ def validate(self):
             if self.options.directfb:
                 raise ConanInvalidConfiguration("Package for 'directfb' is not available (yet)")
 
-    def package_id(self):
-        if Version(self.version) < "2.0.22":
-            del self.info.options.sdl2main
-
     def build_requirements(self):
         if self.settings.os == "Macos" and cross_building(self):
```
valgur, line 175:
```suggestion
        if is_apple_os(self) and cross_building(self):
```

valgur (approved):

PR #20296, opened 2023-10-03
pcapplusplus: bump version to v23.09

seladb:
Specify library name and version:  **pcapplusplus/23.09**

This PR updates the PcapPlusPlus recipe to the latest version v23.09.
This version includes a major change in the build system - moving from a custom build system to CMake.
I asked in the Slack channel what's the best way to approach this change, and I was told it's better to create a new folder and a new `conanfile.py`, which is what this PR contains.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20399, opened 2023-10-07
npcap 1.70: migrate to Conan 2.0

seladb:
Specify library name and version:  **npcap/1.70**

This recipe and version already exist in ConanCenter, but probably not in the Conan 2.0 repo. That's probably why the PcapPlusPlus PR fails: https://github.com/conan-io/conan-center-index/pull/20296

This PR mostly fixes the test to be compatible with Conan 2.0.

---

### DIFF
--- a/recipes/npcap/all/test_package/conanfile.py
+++ b/recipes/npcap/all/test_package/conanfile.py
@@ -1,4 +1,4 @@
-from conan import ConanFile, conan_version
+from conan import ConanFile
 from conan.tools import files
 from conan.tools.build import can_run
 from conan.tools.cmake import cmake_layout, CMake
@@ -23,6 +23,9 @@ class TestPackageConan(ConanFile):
     def layout(self):
         cmake_layout(self)
 
+    def generate(self):
+        files.save(self, os.path.join(self.build_folder, "libpcap_bin_path"), self.dependencies["libpcap"].cpp_info.bindirs[0])
+
     def build(self):
         cmake = CMake(self)
         cmake.configure()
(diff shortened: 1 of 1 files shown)


recipes/npcap/all/test_package/conanfile.py
```diff
@@ -31,17 +30,13 @@ def build(self):
...
     def test(self):
         if can_run(self):
-            bindir = self.cpp.build.bindirs[0]
+            bindir = self.cpp.build.bindir
             # Use libpcap DLL as a replacement for npcap DLL
             # It will not provide all the functions
             # but it will cover enough to check that what we compiled is correct
             files.rm(self, "wpcap.dll", bindir)
-            files.copy(self, "pcap.dll", src=self.deps_cpp_info['libpcap'].bin_paths[0], dst=bindir)
+            libpcap_bin_path = self.deps_cpp_info["libpcap"].bin_paths[0] if conan_version < "2" else self.dependencies["libpcap"].cpp_info.bindirs[0]
```
RubenRBS, line 38: This check should not be necessary. V1 also supports `self.dependencies` attribute

RubenRBS (changes requested):
Thanks! One question about why the check for the conan version, as it does not seem necessary for me

recipes/npcap/all/test_package/conanfile.py
seladb: That's what I thought too, but Conan v1 pipeline failed saying that `self.dependencies` was not available.
Unfortunately I don't have the logs because the current CI report overrides the previous one. Maybe you have access to it?

seladb: Here is the commit where I changed it: https://github.com/conan-io/conan-center-index/pull/20399/commits/320dfaab399c6f9819a86acbcfcaa0a89cff7710

uilianries: @seladb Please, avoid checking conan version and use only `self.dependencies`. As workaround for Conan 1.x you can use: https://github.com/conan-io/conan-center-index/blob/master/recipes/opentelemetry-proto/all/test_package/conanfile.py

Store the value in a file, then get it back in `test()`

seladb: @uilianries @RubenRBS thanks for your review! Addressed here: https://github.com/conan-io/conan-center-index/pull/20399/commits/ea3ff475dd78e1dc52e98231f3353131a552cad5

uilianries (approved):
LGTM

recipes/npcap/all/test_package/conanfile.py
RubenRBS: Oops, sorry about the confusion, I should have made myself more clear along the lines of what @uilianries did, thanks a lot for addressing the review so quickly :)

PR #20442, opened 2023-10-08
chunkio: add recipe

toge:
Specify library name and version:  **chunkio/***

---


recipes/chunkio/all/patches/1.1.6-0003-disable-pic.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index ebf3c4c..915fb28 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -7,7 +7,7 @@ set(CIO_VERSION_PATCH  6)
+ set(CIO_VERSION_STR "${CIO_VERSION_MAJOR}.${CIO_VERSION_MINOR}.${CIO_VERSION_PATCH}")
+ 
+ # CFLAGS
+-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
++#set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
SpaceIm, line 10: It must be kept if shared because there is a vendored libs always build a static lib, so when chunkio is shared, it fails if PIC is not enabled.

SpaceIm:
[revert disable PIC](https://github.com/conan-io/conan-center-index/pull/20442/commits/2678950eccf393be94c75727c2f75d8739620a86)

it's not what I said in https://github.com/conan-io/conan-center-index/pull/20442#discussion_r1350807241

toge:
@SpaceIm 
Thank you for your comment.
I misunderstand your review.
I will fix it.

SSE4:
@RubenRBS can you unblock this PR please?

> * The upstream CMakeLists defines `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` in https://github.com/fluent/chunkio/blob/master/CMakeLists.txt#L10, you'll ant to remove that

sorry, I don't understand what does it mean "you'll ant to remove that", can you elaborate a little bit more? ant as a verb (https://en.wiktionary.org/wiki/ant#Verb) doesn't make much sense in that context, so any details are appreciated.
as a side note, if upstream wishes to support only PIC-code, that's their choice to be honored by conan recipe (they might have had some good reasoning behind that).

> * @SpaceIm has had som recent PRs fixing the apple hook warning you are getting, you might want to check their recent PRs for examples :)

that's just a warning, not an error, so shouldn't be a strict blocker for the PR. "you might want to check", but also "you might not want to check", so that's just an optional thing, up to the contributor.

toge:
@RubenRBS 
Can you check again?
I am afraid it will be automatically closed.

RubenRBS:
Will do so tomorrow, thanks a lot for your patience :)

recipes/chunkio/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_filesystem": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_filesystem": True,
```
franramirez688, line 25: This option is not being used in the recipe. Is it aimed at managing the `CIO_BACKEND_FILESYSTEM` cmake variable?

recipes/chunkio/all/patches/1.1.6-0003-disable-pic.patch
```diff
@@ -0,0 +1,15 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -7,7 +7,9 @@ set(CIO_VERSION_PATCH  6)
+ set(CIO_VERSION_STR "${CIO_VERSION_MAJOR}.${CIO_VERSION_MINOR}.${CIO_VERSION_PATCH}")
+ 
+ # CFLAGS
+-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
++if(BUILD_SHARED_LIBS)
++  set(CMAKE_POSITION_INDEPENDENT_CODE ON)
++endif()
```
franramirez688, line 12: Should we remove the `fpic` option from the Conan recipe?

franramirez688 (commented):
@toge Thanks a lot for the recipe! 😄 

A few doubts about the PR, but it's almost done 👍

recipes/chunkio/all/patches/1.1.6-0003-disable-pic.patch
franramirez688: IMO, I think that Conan should manage it through the option. Wdyt? @RubenRBS @SpaceIm @uilianries

uilianries: I'm fine with this change. fPIC management is more or less regular in CCI, to avoid cases where we build a package without fPIC and try to link to another package with fPIC.

recipes/chunkio/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
+        tc.variables["CIO_LIB_STATIC"] = not self.options.shared
+        tc.variables["CIO_LIB_SHARED"] = self.options.shared
+        # Relocatable shared libs on macOS
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
```
valgur, line 53:
```suggestion
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
For `tc.variables` due to CMake version being set to 3.0.

recipes/chunkio/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
+        tc.variables["CIO_LIB_STATIC"] = not self.options.shared
+        tc.variables["CIO_LIB_SHARED"] = self.options.shared
```
franramirez688, line 51: Following the comment above https://github.com/conan-io/conan-center-index/pull/20442/files#r1431569021
```suggestion
        tc.variables["CIO_LIB_SHARED"] = self.options.shared
        tc.variables["CIO_BACKEND_FILESYSTEM"] = self.options.with_filesystem
```

franramirez688 (commented):
Following-up! 😁

franramirez688 (approved):
LGTM! 👏

RubenRBS (approved):

PR #20443, opened 2023-10-08
arrow: fix issue when building with aws-cpp-sdk

jcar87:
When building `arrow` with aws s3 support, e.g.: 

```
conan install --requires=arrow/11.0.0 -o arrow*:dataset_modules=True -o arrow*:with_re2=True -o arrow*:with_s3=True -o aws-sdk-cpp*:config=True --build=missing
```

This PR fixes issue where this fails compiling with:
```
//b/src/cpp/src/arrow/filesystem/s3fs.cc:43:10: fatal error: aws/core/Aws.h: No such file or directory
   43 | #include <aws/core/Aws.h>
      |          ^~~~~~~~~~~~~~~~
compilation terminated.
```

The issue is caused by the Arrow CMakeLists scripts expecting the variable `AWSSDK_LINK_LIBRARIES`, which is not generated by CMakeDeps. The fix is making sure this variable exists with the correct value (in this case, it's simply the target `aws-cpp-sdk::aws-cpp-sdk` which transitively sets up the right things.

Note: this fixes an issue that does not occur with the default options for this recipe, so CI will not verify these changes.
I've verified this locally on Linux.


SpaceIm:
> The issue is caused by the Arrow CMakeLists scripts expecting the variable AWSSDK_LINK_LIBRARIES, which is not generated by CMakeDeps. The fix is making sure this variable exists with the correct value (in this case, it's simply the target aws-cpp-sdk::aws-cpp-sdk which transitively sets up the right things.

Is `AWSSDK_LINK_LIBRARIES` a variable defined by upstream aws-cpp-sdk config file? If so, aws-sdk-cpp recipe should be fixed, no arrow recipe.

jcar87:
> > The issue is caused by the Arrow CMakeLists scripts expecting the variable AWSSDK_LINK_LIBRARIES, which is not generated by CMakeDeps. The fix is making sure this variable exists with the correct value (in this case, it's simply the target aws-cpp-sdk::aws-cpp-sdk which transitively sets up the right things.
> 
> Is `AWSSDK_LINK_LIBRARIES` a variable defined by upstream aws-cpp-sdk config file? If so, aws-sdk-cpp recipe should be fixed, not arrow recipe.

Please feel free to do so if that is the case - the changes in this PR are in such a way that if the variable _is_ defined in the future, this becomes redundant.

SpaceIm:
Hum... I've opened https://github.com/conan-io/conan-center-index/pull/20446 it can avoid to make a recipe confusing while issue is elsewhere.

memsharded (approved):

RubenRBS (approved):

Issue #20471, opened 2023-10-09
[package] assimp/5.2.2: error: 'M_PI' was not declared in this scope


algarcia-ujaen:
### Description

IDE: CLion 2023.2.2 with Conan plugin 2.0.2 (using the conan installed in the system)
CMake 3.27.4 installed in the system as standalone (added to system path)
Conan 2.0.10 installed in the system as standalone (added to system path)

When assimp is added to an empty project, the download is correct, but when the source code is compiled, an "error: 'M_PI' was not declared in this scope" message appears, and the compilation is aborted.

### Package and Environment Details

* Package Name/Version: **assimp/5.2.2**
* Operating System+version: **Windows 10**
* Compiler+version: **GCC 13.1.0**
* Conan version: **conan 2.0.10**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=13
os=Windows

### Steps to reproduce

- Create an empty project
- Configure the conan plugin (use Conan installed in the system)
- Add assimp to the project through the conan plugin
- Add the corresponding find_package and target_link_libraries instructions to CMakeLists.txt
- Reload the CMake project

### Logs

....

poly2tri/cci.20130502: Running CMake.build()
poly2tri/cci.20130502: RUN: cmake --build "C:\Users\admin\.conan2\p\b\poly2ca8dc7da62214\b\build\Debug" -- -j8
[1/6] Building CXX object CMakeFiles/poly2tri.dir/src/poly2tri/sweep/advancing_front.cc.obj
[2/6] Building CXX object CMakeFiles/poly2tri.dir/src/poly2tri/sweep/cdt.cc.obj
[3/6] Building CXX object CMakeFiles/poly2tri.dir/src/poly2tri/sweep/sweep.cc.obj
FAILED: CMakeFiles/poly2tri.dir/src/poly2tri/sweep/sweep.cc.obj 
[...]

Issue #17054, opened 2023-04-17
[package] poly2tri/cci.20130502: M_PI was not declared


StellaSmith:
### Description

poly2tri fails to compile under MSYS2 (mingw64).

### Package and Environment Details

* Package Name/Version: **poly2tri/cci.20130502**
* Operating System+version: **Microsoft Windows Server 2022 10.0.20348**
* Compiler+version: **GCC 12.2.0**
* Docker image: **[window-latest](https://github.com/actions/runner-images/releases/tag/win22%2F20230407.1)**
* Conan version: **conan 2.0.4**
* Python version: **CPython 3.10.10**

### Conan profile

 [settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=12
os=Windows

### Steps to reproduce

`msys2 "conan install --requires poly2tri/cci.20130502"`

### Logs

<details><summary>Click to expand log</summary>

```
poly2tri/cci.20130502: Sources downloaded from 'conancenter'
poly2tri/cci.20130502: Calling source() in C:/Users/runneradmin/.conan2/p/poly2cda31f58dc326/s/src
Downloading 88de49021b6d9bef6faa1bc94ceb3fbd85c3c204.zip
poly2tri/cci.20130502: Unzipping 471.7KB, this can take a while


-------- Installing package poly2tri/cci.20130502 (17 of 34) --------
poly2tri/cci.20130502: Building from source
poly2tri/cci.20130502: Package poly2tri/cci.20130502:e0cb4c8abd08827791498b17e032db28adba565b
poly2tri/cci.20130502: Copying sources to build folder
poly2tri/cci.20130502: Building your package in C:/Users/runneradmin/.conan2/p/t/poly23223c25a40ad5/b
poly2tri/cci.20130502: Calling generate()
poly2tri/cci.20130502: Generators folder: C:/Users/runneradmin/.conan2/p/t/poly23223c25a40ad5/b/build/Debug/generators
poly2tri/cci.20130502: CMakeToolchain generated: conan_toolchain.cmake
poly2tri/cci.20130502: CMakeToolchain generated: CMakePresets.json
poly2tri/cci.20130502: Generating aggregated env files
poly2tri/cci.20130502: Generated aggregated env files: ['conanbuild.bat', 'conanrun.bat']
poly2tri/cci.20130502: Calling build()
poly2tri/cci.20130502: Running CMake.configure()
[...]

PR #20496, opened 2023-10-10
poly2tri: fix MinGW by injecting `_USE_MATH_DEFINES` to avoid `M_PI` not defined

SpaceIm:
closes https://github.com/conan-io/conan-center-index/issues/17054
closes https://github.com/conan-io/conan-center-index/issues/20471

@StellaSmith @algarcia-ujaen please test this PR with your compiler and report whether it fixes your issue.

---

### DIFF
--- a/recipes/poly2tri/all/CMakeLists.txt
+++ b/recipes/poly2tri/all/CMakeLists.txt
@@ -11,6 +11,11 @@ add_library(poly2tri
     ${POLY2TRI_SRC_DIR}/sweep/sweep_context.cc
 )
 
+# _USE_MATH_DEFINES is defined in utils.h where M_PI is used, but it can break because math.h or cmath
+# may have already been included indirectly by other headers included by sweep.cc before inclusion of utils.h
+# Therefore we ensure that _USE_MATH_DEFINES is defined before any inclusion of headers.
+target_compile_definitions(poly2tri PRIVATE _USE_MATH_DEFINES)
+
 if(MSVC AND BUILD_SHARED_LIBS)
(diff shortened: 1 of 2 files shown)


algarcia-ujaen (approved):
Works like a charm :-)

Thanks

recipes/poly2tri/all/CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ add_library(poly2tri
     ${POLY2TRI_SRC_DIR}/sweep/sweep_context.cc
 )
 
+# _USE_MATH_DEFINES is defined in utils.h where M_PI is used, but it can break because math.h or cmath
+# may have already been included indirectly by other headers included by sweep.cc before inclusion of utils.h
+# Therefore we ensure that _USE_MATH_DEFINES is defined before any inclusion of headers.
+target_compile_definitions(poly2tri PRIVATE _USE_MATH_DEFINES)
```
RubenRBS, line 17: Question, any reason to prefer this against adding `tc.defines["..` in the `generate()` method?

RubenRBS (approved):
Seems like a proper fix in this case, thanks a lot - I do have one question for further insight, but approving now

jwillikers (approved):

recipes/poly2tri/all/CMakeLists.txt
jwillikers: For CMake projects, using the `target_*` scoped functions is generally considered best practice. I think this also fits better as part of the build system where possible.

SpaceIm: Yep, we have a custom CMakeLists, there is no reason to inject definitions through tc.preprocessor_definition rather than CMakeLists itself.

PR #20507, opened 2023-10-10
libmount: Add 2.40.2 and use Meson

jwillikers:
Specify library name and version:  **libmount/2.40.2**

The util-linux project is switching to Meson.
This PR switches version 2.40.2 to use Meson instead of Autotools.

Version 2.39 has the following error when cross-compiling under Meson:

```
libmount/lib_mount.a.p/src_hook_mount.c.o -MF libmount/lib_mount.a.p/src_hook_mount.c.o.d -o libmount/lib_mount.a.p/src_hook_mount.c.o -c ../src/libmount/src/hook_mount.c
../src/libmount/src/hook_mount.c: In function ‘hook_create_mount’:
../src/libmount/src/hook_mount.c:302:40: error: ‘struct statx’ has no member named ‘stx_mnt_id’; did you mean ‘stx_uid’?
  302 |                 cxt->fs->id = (int) st.stx_mnt_id;
      |                                        ^~~~~~~~~~
      |                                        stx_uid
```

Version 2.40 doesn't hit this snag, so I'm using Meson support for version 2.40.

---


jwillikers:
I've modified this PR like #20508 to switch to Meson only for the new version 2.40 to avoid cross-compilation problems in 2.39 with Meson. It will be a draft until version 2.40 is released, but the rc2 is working great so far.

jwillikers:
Hitting this issue: https://github.com/util-linux/util-linux/issues/2830

recipes/libmount/meson/conandata.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+  "2.40-rc2":
+    url: "https://github.com/util-linux/util-linux/archive/refs/tags/v2.40-rc2.tar.gz"
+    sha256: "0fdf4a47eb415639cfcdf2ca4cb76919ca971dc173ce5bb034857a77188a7c27"
+patches:
+  "2.40-rc2":
+    - patch_file: "patches/4ebdd569049ac48a3867a59537cea769fa319e8f.patch"
+    - patch_file: "patches/0001-meson-Only-build-blkzone-and-blkpr-if-the-required-l.patch"
```
valgur, line 8:
```suggestion
    - patch_file: "patches/0001-meson-Only-build-blkzone-and-blkpr-if-the-required-l.patch"
      patch_type: "bugfix"
      patch_source: "https://github.com/util-linux/util-linux/issues/2850"
```

valgur (approved):

recipes/libmount/meson/conandata.yml
jwillikers: Ah, sorry about that. I left these patches in to test in CCI temporarily. Both have landed in the 2.40 branch, so they won't be necessary when 2.40 is properly released. I've got this marked as a draft until then.

recipes/libmount/meson/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+        # Enable libutil for older versions of glibc which still provide an actual libutil library.
+        tc.project_options["libutil"] = "enabled"
+        tc.project_options["program-tests"] = False
+        tc.generate()
+        virtual_build_env = VirtualBuildEnv(self)
+        virtual_build_env.generate()
+
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+        # Disable translations.
```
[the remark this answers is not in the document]
jwillikers: Ha! Nice catch, turns out it's not needed. I've just removed it.

recipes/libmount/meson/patches/2.40-0001-meson-Only-require-the-crypt-library-when-necessary.patch
uilianries, line 1: This patch file is not listed in conandata.yml to be applied.

jwillikers: That was upstreamed for 2.40.1, and I forgot to remove it. Thanks, I've removed it now.

danimtb (approved):

recipes/libmount/meson/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    def validate(self):
+        if self.settings.os != "Linux":
+            raise ConanInvalidConfiguration(f"{self.ref} only supports Linux")
+
+    def build_requirements(self):
+        self.tool_requires("bison/3.8.2")
+        self.tool_requires("flex/2.6.4")
+        self.tool_requires("meson/1.3.2")
+        if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
+            self.tool_requires("pkgconf/2.1.0")
```
valgur, line 54:
```suggestion
        self.tool_requires("meson/[>=1.2.3 <2]")
        if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
            self.tool_requires("pkgconf/[>=2.2 <3]")
```

valgur (approved):

uilianries:
Hello again, and thank you for bringing this new option of building! I just started to review (again) this PR.

> The util-linux project is switching to Meson.

@jwillikers Do you have a reference about it from the upstream? What do you mean by "switching", will autotools be deprecated? I see you contributed directly to libmount meson script as well.

I'm curious about the advantage of migrating it to Meson now in CCI, if is still working with Autotools so far and the current recipe is really straightforward, no patches and even smaller than this new one? IMO I don't see a need for this migration, and when migrating we had few problems related to library name in the past, so we need to double-check everything.

jwillikers:
> Hello again, and thank you for bringing this new option of building! I just started to review (again) this PR.
> 
> > The util-linux project is switching to Meson.
> 
> @jwillikers Do you have a reference about it from the upstream? What do you mean by "switching", will autotools be deprecated? I see you contributed directly to libmount meson script as well.
> 
> I'm curious about the advantage of migrating it to Meson now in CCI, if is still working with Autotools so far and the current recipe is really straightforward, no patches and even smaller than this new one? IMO I don't see a need for this migration, and when migrating we had few problems related to library name in the past, so we need to double-check everything and plus keep 2 recipes for the same project.

Meson is much easier for maintainers to manage. It's much simpler than Autotools, better documented, more standard, and more well-behaved leading to much more consistency and ease of use, especially when it comes to cross-compilation and multi-platform support. Similar tools from the `util-linux` project packages in CCI will benefit from moving to Meson as well as `libmount`. To avoid potential issues with the migration that happened for `freetype`, I'm migrating for a new version of the package and not migrating an existing version. That would have avoided the issues migrating the `freetype` recipe.

I could ask the upstream what the migration plans, as I'm not sure their exact plans, but in general, I think the projects that have been adding Meson build systems in addition to Autotools are planning on slowly migrating away from Autotools and to Meson. I can't imagine that anyone wants to support multiple build systems indefinitely.

danimtb (approved):

jcar87 (changes requested):
Thanks @jwillikers  for your contribution.

We've had a meeting with the team to discuss this PR and have decided to not move forward with the meson version of the recipe at the present time.

We fully agree that meson is easier and more modern than Autotools, but that is a concern for the upstream maintaners of libmount, not for us. Our main concern is the recipe, and the autotools version of the recipe is as trivial as it gets for an Autotools based, as far as we can see. We see no benefit presently - this would cause us to maintain two versions of the recipe (autotools and meson), and the new recipe has patches and additional dependency whereas the autotools version doesn't have these. So from a maintainability from of view, we're more comfortable where we are.

Previous meson migrations have caused issues - as you have mentioned, maintaining two build systems is not sustainable, and what we have found is that they are often not equivalent, even if both are checked by CI upstream. Meson also also follows different conventions when it comes to file names and soversions, in ways that can cause breakages. Oftentimes, the upstream CI do not validate these as long as the tests pass. 

Given that the meson support in libmount is too recent, we will reconsider this in the future once there is more evidence of upstream authors and linux distro package managers using this new build method.

Feel free to amend the PR to only add the new version if this is still required.

jcar87:
Superseded by https://github.com/conan-io/conan-center-index/pull/25841

PR #20522, opened 2023-10-11
bump glib dependencies

AndreyMlashkin:
Specify library name and version:  **glib/all**

---

### DIFF
--- a/recipes/glib/all/conanfile.py
+++ b/recipes/glib/all/conanfile.py
@@ -81,9 +81,9 @@ class GLibConan(ConanFile):
             self.requires("libiconv/1.17")
 
     def build_requirements(self):
-        self.tool_requires("meson/[>=1 <2]")
+        self.tool_requires("meson/1.2.2")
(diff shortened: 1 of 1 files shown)


recipes/glib/all/conanfile.py
```diff
@@ -81,9 +81,9 @@ def requirements(self):
             self.requires("libiconv/1.17")
 
     def build_requirements(self):
-        self.tool_requires("meson/1.2.1")
+        self.tool_requires("meson/[>=1 <2]")
         if not self.conf.get("tools.gnu:pkg_config", check_type=str):
-            self.tool_requires("pkgconf/2.0.3")
+            self.tool_requires("pkgconf/[>=2 <3]")
```
SpaceIm, line 86: it's not allowed currently

AndreyMlashkin: Is there some policy? Where to find it?
It's a pity, because tool_requires is exactly where the version ranges would make the most sense
It's very common situation, that a dev want to define the minimum version of cmake, meson, pkgconf and don't care if it's higher.

PR #20502, opened 2023-10-10
xorg/system: Add xcb-dri2, xcb-dri3, xcb-glx, and xcb-present

jwillikers:
Specify library name and version:  **xorg/system**

Motivation: https://github.com/conan-io/conan-center-index/pull/20502#issuecomment-1757742671

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20528, opened 2023-10-11
mesa: Add recipe

### LINKED ISSUE #6506: [request] mesa/21.1.5
### Package Details
  * Package Name/Version: **mesa/21.1.5**
  * Website: **https://github.com/mesa3d/mesa**
  * Source code: **https://github.com/mesa3d/mesa**

3D Graphics Library

jwillikers:
Specify library name and version:  **mesa/24.0.3**

[Mesa](https://www.mesa3d.org/) is an open-source implementation of various graphics APIs.

This PR adds a recipe for Mesa, which overlaps with some of the libraries provided by the `opengl/system`, `egl/system`, and `libglvnd` packages. This occurs when Mesa is built without `libglvnd` support on Linux and FreeBSD, where it will provide the OpenGL libraries that are enabled.

The Mesa package provides the `libgbm` library, which is required for several Wayland compositors, such as Weston and wlroots. I'm working on packages for these compositors.

Fixes #6506.

Required:

- [x] #23204
- [x] #20502
- [x] Windows support (MSVC) - I was able to take `opengl32.dll` produced by Mesa, drop it in to a Qt application, and get software rendering inside a VM.

Optional:

- [x] #21387 - This will allow building a newer version of Python as required for CCI. I've already got this PR tested and working with the Mesa recipe, so all that is required is for it to be merged.
- [ ] llvm - Mesa integrates tightly with LLVM to provide a substantial amount of functionality. The LLVMPipe software rasterizer is one of these features. Ideally, #17509 will be merged so that a proper dependency can exist.
- [ ] macOS - I don't have a Mac unfortunately, so I haven't been testing this extensively. Needless to say, only GLX is supported for macOS and which limits it to working in XQuartz. A lot of work involving HomeBrew to pull in XQuartz is currently needed to get this working.
- [ ] OpenCL - Requires llvm / clang libraries and modules like clang-cpp, Polly, PollyISL, etc. There is also [LLVMSPIRVLib](https://github.com/KhronosGroup/SPIRV-LLVM-Translator).
- [x] DirectX support - need to make a directx-headers package. This will allow building the Microsoft experimental Vulkan drivers.  #20623
- [x] libva - missing a libva package
- [ ] omx - missing dependencies such as [libomxil-bellagio](https://github.com/felipec/libomxil-bellagio) or libtizonia, libtizplatform, and tizilheaders
- [x] vdpau - missing a vdpau package
- [ ] Windows Cygwin
- [ ] Windows mingw64
- [ ] Tests could be improved. Tests are missing for Vulkan, GLX, and GLES. Runtime loading of EGL is not tested when running without glvnd support. Nothing is currently tested on Windows, either. Contributions on this end would be incredibly helpful.

---


uilianries:
@jwillikers Please, consider merging https://github.com/jwillikers/conan-center-index/pull/15

Ahajha:
Re: Mac support. Not sure how helpful I would be, but I have an M2 Mac mini that I can use for testing. Feel free to ping me if you want me to test some things.

arnoldad:
@jwillikers im also happy to test on m1/m2 macs and worst case can even send one out to you.

jwillikers:
@Ahajha @arnoldad Thanks! I've got a couple of devs around with M2 macs, so I can do some minimal testing so long as they keep putting up with me. I probably should update the description with what I ended up finding out for macOS, and that is that Mesa is extremely limited in its macOS support due to the fact that it only supports GLX. There's no EGL support on macOS, and the GLX support can only be used with XQuartz, which would require that and any related dependencies to be packaged in CCI first in order to support this in CCI.

I was hoping to use zink to generate OpenGL bindings from [moltenvk](https://github.com/KhronosGroup/MoltenVK) in order to get support for newer versions of OpenGL, but since we're targeting Qt, XQuartz isn't an option. That renders the Mesa route futile. I think that using Vulkan via MoltenVK directly is the best way to go about this instead of targeting OpenGL. An alternative option is Asahi Linux, but that requires dual-booting Linux on your mac.

In case there's anyone that would actually find the ability to use Mesa on macOS useful still, I've left all the bits in that should make it possible, though having it built in CCI isn't something I'll be working. Contributions for that are certainly welcome. I did get it building at one point, but it required a few workarounds involving HomeBrew.

recipes/mesa/all/conanfile.py
uilianries, line 1: @jwillikers do you really all these options in this recipe? You see, is really hard to review this recipe, are 1k5 lines and the complexity of enabling/disabling options is really hard to follow. Still, this is first step, because giving maintenance will be even harder. As example, see Boost recipe, we avoid touching there because it has 1k8 lines and is really easy to break something.

jwillikers: I'll think about how it can be further simplified. I mostly just translated `meson_options.txt` for the available options. I imagine that some of the ones having to do with naming things could be removed at the very least. To set expectations, though, I'd expect this recipe to be one of the more complicated recipes in CCI with Boost, Qt, and LLVM.

On the optimistic side of things, there's a lot of complexity in the Boost recipe just for dealing with `b2` and the whole modules thing. Since Meson is used here, most of the logic is just for dealing with options and various inter-dependencies. I might be able to think of a way to simplify some of this down to make the relationships more explicit and defined in a single place.

I primarily need the `gbm` library as a dependency for my work to add recipes for wlroots and Weston. However, I would suspect that those interested in the recipe in #6506 are interested in using the graphics drivers and GL implementations, which requires the majority of the options.

jwillikers: @uilianries I've removed several extraneous options, though none of these added much complexity. Hopefully that simplifies it at least a bit. There might yet be a way to simplify inter-option dependencies.

PR #20530, opened 2023-10-12
nmslib: add support for WITH_EXTRAS

valgur:
https://github.com/nmslib/nmslib/blob/master/similarity_search/src/CMakeLists.txt#L16-L19
https://github.com/nmslib/nmslib/blob/master/similarity_search/CMakeLists.txt#L100C17-L108

I set it to enabled by default since it only adds one .cpp file and an internal Eigen dependency to the build.


recipes/nmslib/all/conanfile.py
```diff
@@ -45,6 +50,11 @@ def configure(self):
     def layout(self):
         cmake_layout(self, src_folder="src")
 
+    def requirements(self):
+        if self.options.build_extras:
+            # Eigen is only used internally, no need for transitive_*
+            self.requires("eigen/3.4.0")
```
franramirez688, line 56: @valgur is it not requiring boost and gsl too?

recipes/nmslib/all/conanfile.py
```diff
@@ -25,10 +25,15 @@ class Nmslib(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "build_extras": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "build_extras": True,
```
franramirez688, line 33: I'd use `with_extras` instead, what do you think?
```suggestion
        "with_extras": True,
```

recipes/nmslib/all/conanfile.py
franramirez688: Sorry, I just saw that it was for python-bindings.

recipes/nmslib/all/conanfile.py
```diff
@@ -57,13 +67,23 @@ def source(self):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.variables["WITH_EXTRAS"] = self.options.build_extras
+        if self.options.build_extras:
+            tc.preprocessor_definitions["WITH_EXTRAS"] = "1"
```
franramirez688, line 72: I think that we can remove this one https://github.com/nmslib/nmslib/blob/master/similarity_search/CMakeLists.txt#L80 It seems that it's already adding the preprocessor definition

recipes/nmslib/all/conanfile.py
franramirez688: Oh, that makes sense. Agree 😄

jwillikers (approved):

recipes/nmslib/all/conanfile.py
```diff
@@ -41,10 +46,19 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+        if is_msvc(self):
+            # Not available on MSVC
+            # https://github.com/nmslib/nmslib/blob/v2.1.1/similarity_search/include/space/space_sqfd.h#L19
+            self.options.build_extras = False
```
franramirez688, line 52: Instead of changing the value, it'd be better to add this step to the `validate()` function and raise an exception saying something like `"WITH_EXTRAS is not available for MSVC, please, set this option to False"`.

franramirez688 (changes requested):

recipes/nmslib/all/conanfile.py
valgur: That would not work too well with the default value being `True` and would cause it to be unnecessarily skipped in the CI. I simply disabled the option on MSVC instead.

franramirez688 (approved):

jwillikers (approved):

PR #20533, opened 2023-10-12
libzen: fix build of shared lib with MinGW

SpaceIm:
Patch `0002-export-data-in-windows-dlls.patch` was broken. msvc didn't complain, but MinGW did.

---

### DIFF
--- a/recipes/libzen/all/patches/0002-export-data-in-windows-dlls.patch
+++ b/recipes/libzen/all/patches/0002-export-data-in-windows-dlls.patch
@@ -1,18 +1,3 @@
---- Source/ZenLib/Conf.cpp
-+++ Source/ZenLib/Conf.cpp
-@@ -24,8 +24,8 @@
- 
- //End of line
- #ifdef WINDOWS
--    const Char* EOL=__T("\r\n");
--    const Char  PathSeparator=__T('\\');
-+    ZEN_IMPEXP const Char* EOL=__T("\r\n");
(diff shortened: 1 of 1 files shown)


RubenRBS:
Thanks :)

For scenarions not covered by CCI, it would be great to get your local compilation logs so that we can double check that they match with our local testing in these cases, specially for things that are tricky like this

SpaceIm:
I'm not on my Windows computer currently, but yep I'll share error fixed by this PR.

Basically the problem comes from the fact that `__declspec(dllexport)` was put in front of declaration AND definition of `EOL` and `PathSeparator`, while it should be only put in front of declaration. It was leading to some compilation error (or link error I don't remember) related to `PathSeparator`.

SpaceIm:
So error with MinGW shared was:

```
C:/Users/spaceim/.conan2/p/b/libze48e7b0cafc132/b/src/Source/ZenLib/Conf.cpp:28:28: error: external linkage required for symbol 'ZenLib::PathSeparator' because of 'dllexport' attribute
   28 |     ZEN_IMPEXP const Char  PathSeparator=__T('\\');
      |                            ^~~~~~~~~~~~~
```

jwillikers (approved):

SpaceIm:
@RubenRBS could you review please?

RubenRBS (approved):
Hi! Etxremelly sorry about the delay, this one fell thru the cracks :)

PR #20537, opened 2023-10-12
teemo: make deprecate

toge:
Specify library name and version:  **teemo/***

teemo has been renamed to zoe.

---

### DIFF
--- a/recipes/teemo/all/conandata.yml
+++ b/recipes/teemo/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
   "2.7":
     url: "https://github.com/winsoft666/teemo/archive/refs/tags/v2.7.tar.gz"
-    sha256: "97be8ae62cd3e2e92ed0ff405dfaf5e6cb6e323b21c53e798adfa5872a2f7084"
+    sha256: "0cf4f90d22b1c19845db08ff1addc5872327b33f020e8d5409f3819391afaf79"
 
 patches:
   "2.7":
--- a/recipes/teemo/all/conanfile.py
+++ b/recipes/teemo/all/conanfile.py
@@ -27,6 +27,7 @@ class TeemoConan(ConanFile):
         "shared": False,
         "fPIC": True,
     }
+    deprecated = "zoe"
 
     @property
     def _min_cppstd(self):
--- a/recipes/teemo/all/conanfile.py
+++ b/recipes/teemo/all/conanfile.py
@@ -48,13 +48,13 @@ class TeemoConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("libcurl/7.86.0")
+        self.requires("libcurl/7.86.0", transitive_headers=True)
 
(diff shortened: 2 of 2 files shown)


recipes/teemo/all/conandata.yml
```diff
@@ -1,7 +1,7 @@
 sources:
   "2.7":
     url: "https://github.com/winsoft666/teemo/archive/refs/tags/v2.7.tar.gz"
-    sha256: "97be8ae62cd3e2e92ed0ff405dfaf5e6cb6e323b21c53e798adfa5872a2f7084"
```
RubenRBS, line 4: Why has the checksum changed?

recipes/teemo/all/conanfile.py
```diff
@@ -48,12 +49,13 @@ def layout(self):
 
     def requirements(self):
         self.requires("libcurl/7.86.0")
+        self.requires("openssl/[>=1.1 <4]", transitive_headers=True)
```
RubenRBS, line 52: Has this always been missing then?

RubenRBS (commented):
Thanks! A few questions, but the deprecation seems to make sense :)

recipes/teemo/all/conandata.yml
toge: Sorry, I don't know why.

recipes/teemo/all/conanfile.py
toge: @RubenRBS 
Thank you for your review.
I want to modify requiring openssl when libcurl's with_ssl is true.

```
        self.requires("libcurl/7.86.0")
        if self.dependencies["libcurl"].options.with_ssl:
            self.requires("openssl/[>=1.1 <4]", transitive_headers=True)
```
But this code is failed with following error message.
```
ERROR: teemo/2.7: Error in requirements() method, line 53
        if self.dependencies["libcurl"].options.with_ssl:
        KeyError: "'libcurl' not found in the dependency set"
```

Is there any way to solve it?

jwillikers: Are the `dependencies` even available in the `requirements` method? Or perhaps a full reference is needed?

RubenRBS: Yeah `self.dependencies` has no chance of being valid inside the `requirements()` method, the dependencies have not been expanded at that point (It happens after the requirements() and build_requirements() methods have been called)

RubenRBS: I can't actually come up with a clear way to do what you want in this case @toge, let me think for a bit :)

recipes/teemo/all/conandata.yml
RubenRBS: Just found out they are redirecting the files to the new zoe location, and the name changed. Otherwise seems identical, so approving now, thanks a lot for your patience on this one :)

RubenRBS (approved):
Thanks!

danimtb (approved):

PR #20538, opened 2023-10-12
mtdev: Call autoreconf to allow cross-compiling for armv8

jwillikers:
Specify library name and version:  **mtdev/***

Unless autoreconf is called when cross-compiling for armv8, the follow error occurs:

```
checking build system type... x86_64-pc-linux-gnu
checking host system type... Invalid configuration `aarch64-linux-gnu': machine `aarch64' not recognized
```

---


RubenRBS (approved):
Thanks! Looks good to me :)
(When cross-compiling is involved, it would be really nice if you could provide your logs of that scenario working, just to double check that our local testing matches, as CCI does not crosscompile in any of its configurations)

danimtb (approved):

PR #20540, opened 2023-10-12
libglvnd: Fix the egl component to only require xorg::x11 when enabled

jwillikers:
Specify library name and version:  **libglvnd/***

This fixes an error in the components specification for the `egl` component.
When disabling X11 support, the package does not require `xorg/system`.
This is problematic because the `egl` component still specifies a dependency on the `xorg/system` component.
This results in an error in consumers calling `PkgConfigDeps(self).generate()` when there is nothing pulling in `xorg/system`.
It's a very obscure error and should probably result in an error at package creation time instead of pkg-config generation in consumers anywhere up the line.

---


RubenRBS (approved):
Great catch!

ericLemanissier (approved):

jwillikers:
@RubenRBS Looks like CI is stuck.

RubenRBS:
Not even stuck, seems like the webhook didn't make it to Jenkins!

Issue #20542, opened 2023-10-12
[package] xkbcommon/1.5.0: build fails on Linux (wayland-client >= 1.2.0, wayland-protocols >= 1.7 which were not found)


SpaceIm:
### Description

xkbcommon/1.5.0 build fails on Linux:

```
Run-time dependency wayland-client found: NO (tried pkgconfig and cmake)
Found pkg-config: /home/spaceim/.conan2/p/b/pkgco720a404f4464b/p/bin/pkgconf (2.0.3)
Build-time dependency wayland-protocols_build found: YES 1.31
Build-time dependency wayland-scanner_build found: YES 1.22.0

../src/meson.build:481:12: ERROR: Problem encountered: The Wayland xkbcli programs require wayland-client >= 1.2.0, wayland-protocols >= 1.7 which were not found.
You can disable the Wayland xkbcli programs with -Denable-wayland=false.

A full log can be found at /home/spaceim/.conan2/p/b/xkbco9f60d54155a8e/b/build-release/meson-logs/meson-log.txt
```

### Package and Environment Details

* Package Name/Version: **zlib/1.2.8**
* Operating System+version: **Linux Ubuntu 18.04**
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

```ini
[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=12
compiler.libcxx=libstdc++11
compiler.cppstd=gnu20
build_type=Release
```

### Steps to reproduce

```
conan install --requires xkbcommon/1.5.0 -o "*:shared=True" -b missing
```

### Logs

<details><summary>Click to expand log</summary>

```
xkbcommon/1.5.0: Calling build()
xkbcommon/1.5.0: Meson configure cmd: meson setup --native-file "/home/spaceim/.conan2/p/b/xkbco9f60d54155a8e/b/build-release/conan/conan_meson_native.ini" "/home/spaceim/.conan2/p/b/xkbco9f60d54155a8e/b/build-release" "/home/spaceim/.conan2/p/b/xkbco9f60d54155a8e/b/src" -Dprefix="/home/spaceim/.conan2/p/b/xkbco9f60d54155a8e/p"
[...]

PR #20545, opened 2023-10-12
xkbcommon: Bump wayland-protocols and required_conan_version

jwillikers:
Specify library name and version:  **xkbcommon/***

Fix building in the build context.

---

### DIFF
--- a/recipes/xkbcommon/all/conanfile.py
+++ b/recipes/xkbcommon/all/conanfile.py
@@ -40,7 +40,7 @@ class XkbcommonConan(ConanFile):
 
     @property
     def _has_build_profile(self):
-        return hasattr(self, "settings_build") and cross_building(self)
+        return hasattr(self, "settings_build")
 
     @property
     def _has_xkbregistry_option(self):
--- a/recipes/xkbcommon/all/conanfile.py
+++ b/recipes/xkbcommon/all/conanfile.py
@@ -2,7 +2,6 @@ import os
 
 from conan import ConanFile
 from conan.tools.apple import fix_apple_shared_install_name
-from conan.tools.build import cross_building
 from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import copy, get, replace_in_file, rmdir
 from conan.tools.gnu import PkgConfigDeps
--- a/recipes/xkbcommon/all/conanfile.py
+++ b/recipes/xkbcommon/all/conanfile.py
@@ -2,6 +2,7 @@ import os
 
 from conan import ConanFile
 from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.build import cross_building
 from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import copy, get, replace_in_file, rmdir
 from conan.tools.gnu import PkgConfigDeps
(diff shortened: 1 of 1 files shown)


recipes/xkbcommon/all/conanfile.py
```diff
@@ -39,7 +40,7 @@ class XkbcommonConan(ConanFile):
 
     @property
     def _has_build_profile(self):
-        return hasattr(self, "settings_build")
+        return hasattr(self, "settings_build") and cross_building(self)
```
SpaceIm, line 43: No change is required here I would say.

jwillikers: You're right, that would change the method meaning. I was trying to fix the issue that you also ran into in #20542. I'm not sure id didn't cause other problems. I've reverted.

danimtb (dismissed):

recipes/xkbcommon/all/conanfile.py
```diff
@@ -2,6 +2,7 @@
 
 from conan import ConanFile
 from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.build import cross_building
```
ericLemanissier, line 5:
```suggestion
```

ericLemanissier (changes requested):

ericLemanissier (approved):

recipes/xkbcommon/all/conanfile.py
```diff
@@ -82,7 +82,7 @@ def build_requirements(self):
             self.tool_requires("pkgconf/2.0.3")
         if self._has_build_profile and self.options.get_safe("with_wayland"):
             self.tool_requires("wayland/<host_version>")
-            self.tool_requires("wayland-protocols/1.31")
+            self.tool_requires("wayland-protocols/1.32")
```
uilianries, line 85:
```suggestion
            self.tool_requires("wayland-protocols/<host_version>")
```

https://docs.conan.io/2/reference/conanfile/methods/build_requirements.html#host-version

recipes/xkbcommon/all/conanfile.py
```diff
@@ -82,7 +82,7 @@ def build_requirements(self):
             self.tool_requires("pkgconf/2.0.3")
         if self._has_build_profile and self.options.get_safe("with_wayland"):
             self.tool_requires("wayland/<host_version>")
-            self.tool_requires("wayland-protocols/1.31")
+            self.tool_requires("wayland-protocols/<host_version>")
```
ericLemanissier, line 85:
```suggestion
            self.tool_requires("wayland-protocols/1.32")
```
host_version cannot work when the dependency is either a build requirement or a requirement, but not both.

ericLemanissier (changes requested):

ericLemanissier (approved):

SpaceIm:
wayland-protocols should be removed from requirements, and another workaround found for 1 profile.

jwillikers:
> wayland-protocols should be removed from requirements, and another workaround found for 1 profile. It's bad to have requirements depending on 1 or 2 profiles, because it corrupts dependency graph. Requirements (not build requirements) must never depend on settings_build.

Would it be possible to just require two profiles using the `validate` method? That would make things simpler.

SpaceIm:
I've submitted https://github.com/conan-io/conan-center-index/pull/21239. I think it should work but I can't test it right now.

jwillikers:
> I've submitted #21239. I think it should work but I can't test it right now.

Thanks! I'll double check that yours cross-compiles correctly tomorrow.

jwillikers:
Closing in favor of #21239

PR #20550, opened 2023-10-12
Websocketpp

bemehiser:
Specify library name and version:  websocketpp/*

Allow a version range for boost in the websocketpp conanfile.

---

### DIFF
--- a/recipes/websocketpp/all/conanfile.py
+++ b/recipes/websocketpp/all/conanfile.py
@@ -37,7 +37,7 @@ class WebsocketPPConan(ConanFile):
             self.requires("openssl/[>=1.1 <4]", transitive_headers=True, transitive_libs=True)
 
         if self.options.with_zlib:
-            self.requires("zlib/1.2.13", transitive_headers=True, transitive_libs=True)
+            self.requires("zlib/[>=1.2.11 <2]", transitive_headers=True, transitive_libs=True)
 
         if self.options.asio == "standalone":
             self.requires("asio/1.28.1", transitive_headers=True)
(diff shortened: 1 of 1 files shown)


recipes/websocketpp/all/conanfile.py
```diff
@@ -42,7 +42,7 @@ def requirements(self):
         if self.options.asio == "standalone":
             self.requires("asio/1.28.1", transitive_headers=True)
         elif self.options.asio == "boost":
-            self.requires("boost/1.83.0", transitive_headers=True)
+            self.requires("boost/[>=1.76.0 <2]", transitive_headers=True)
```
franramirez688, line 45: Hi @bemehiser,

Thanks a lot for contributing!

Unfortunately, this is not allowed at this moment. See more information about [Version Ranges](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges).
Bringing back that piece of documentation:
```
OpenSSL: [>=1.1 <4] for libraries known to be compatible with OpenSSL 1.x and 3.x
CMake: [>3.XX <4], where 3.XX is the minimum version of CMake required by the relevant build scripts
Libcurl: [>=X.YY <9], where X.YY is the minimum version of Libcurl required, starting from 7.78
```
Those are the libraries allowed to use the version ranges for now (and zlib is coming soon too).

bemehiser: Oh, that's too bad. This would be a very useful feature. We've been using boost 1.75/1.76 for a while now with this recipe. Is there a reason they are only allowing those few recipes to use version ranges? Is it so they can manually ensure api compatibility between the versions?

Should I close this pull request, or leave it open in the hope they will decide boost can use version ranges too?

PR #20551, opened 2023-10-12
Allow transitive headers for openssl on cpprestsdk

bemehiser:
Specify library name and version:  cpprestsdk/*

Allow transitive headers for openssl. This is required when linking cpprestsdk to azure-storage-cpp.

---

### DIFF
--- a/recipes/cpprestsdk/all/conanfile.py
+++ b/recipes/cpprestsdk/all/conanfile.py
@@ -1,10 +1,11 @@
-from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import (
     apply_conandata_patches, collect_libs, copy, export_conandata_patches, get,
     replace_in_file, rmdir
 )
 import os
+
+from conan import ConanFile
 
 required_conan_version = ">=1.53.0"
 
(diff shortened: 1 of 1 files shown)


recipes/cpprestsdk/all/conanfile.py
```diff
@@ -1,11 +1,12 @@
-from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import (
     apply_conandata_patches, collect_libs, copy, export_conandata_patches, get,
     replace_in_file, rmdir
 )
 import os
 
+from conan import ConanFile
```
SpaceIm, line 8: Why do you move this import?

recipes/cpprestsdk/all/conanfile.py
```diff
@@ -57,8 +58,8 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.83.0")
-        self.requires("openssl/[>=1.1 <4]")
+        self.requires("boost/[>=1.75.0 <2]")
```
SpaceIm, line 61: it's not allowed yet

recipes/cpprestsdk/all/conanfile.py
```diff
@@ -57,8 +58,8 @@ def layout(self):
...
+        self.requires("openssl/[>=1.1 <4]", transitive_headers=True)
```
SpaceIm, line 62: It would be nice to add a comment explaining why (in which public header it is included for example)

PR #20557, opened 2023-10-13
libsystemd/253.*: Fix cross-compilation

jwillikers:
Specify library name and version:  **libsystemd/253.***

Bump versions of pkgconf and Meson.
Add  a patch to ensure that necessary compiler flags are passed through when the build system is invoking the compiler to generate files.
Without the proper compiler flags when cross-compiling, the system include directories and such are used which causes problems.
The patch has been applied upstream in systemd/systemd#29538.

---

### DIFF
--- a/recipes/libsystemd/all/conandata.yml
+++ b/recipes/libsystemd/all/conandata.yml
@@ -34,6 +34,10 @@ patches:
     - patch_file: "patches/251.15/0001-Remove-dependency-from-coreutils.patch"
       patch_description: "allow to build in environments without 'realpath --relative-to' by replacing it with conan-specific build variable"
       patch_type: "conan"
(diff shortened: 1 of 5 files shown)


RubenRBS (approved):
It's great to see that you yourself fixed it upstream 🫶

jwillikers:
> It's great to see that you yourself fixed it upstream 🫶

Thanks! 😊 It would feel way cooler if I wasn't simply unreverting a single-line change from 5 years ago that ended up fixing the problem after updates to Meson 4 years ago, but that seems to be the way of Open Source at times. I've yet to grasp why no one had fixed this in the past 4 years, though. 🤔

danimtb (approved):

Issue #15276, opened 2023-01-15
[fltk] <fltk>/<1.3.8>: fails to link to libxft on Linux


erikogenvik:
### Description

When building on Linux it fails to link to required libxft. Symbols are missing, like this:
```
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: /home/erik/.conan/data/fltk/1.3.8/_/_/package/7fb45cb79434bbfd311c6fcceb81f069245765ae/lib/libfltk.a(fl_font.cxx.o): in function `Fl_Font_Descriptor::Fl_Font_Descriptor(char const*, int, int)':
fl_font.cxx:(.text+0x469): undefined reference to `XftFontOpenXlfd'
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: fl_font.cxx:(.text+0x6b7): undefined reference to `XftFontMatch'
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: fl_font.cxx:(.text+0x6c8): undefined reference to `XftFontOpenPattern'
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: fl_font.cxx:(.text+0x715): undefined reference to `XftFontOpen'
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: /home/erik/.conan/data/fltk/1.3.8/_/_/package/7fb45cb79434bbfd311c6fcceb81f069245765ae/lib/libfltk.a(fl_font.cxx.o): in function `Fl_Xlib_Graphics_Driver::width(char const*, int)':
fl_font.cxx:(.text+0xb9c): undefined reference to `XftTextExtents32'
/usr/lib64/gcc/x86_64-suse-linux/12/../../../../x86_64-suse-linux/bin/ld: /home/erik/.conan/data/fltk/1.3.8/_/_/package/7fb45cb79434bbfd311c6fcceb81f069245765ae/lib/libfltk.a(fl_font.cxx.o): in function `Fl_Xlib_Graphics_Driver::width(unsigned int)':
fl_font.cxx:(.text+0xc82): undefined reference to `XftTextExtents32'
[...]

PR #20558, opened 2023-10-13
fltk: Disable xft to fix compilation on Linux

jputcu:
fltk/1.3.8

On Linux I got undefined references when using the `fltk` library.

~~~~
undefined reference to `XftFontMatch'
undefined reference to `XftFontOpenPattern'
undefined reference to `XftFontOpen'
undefined reference to `XftFontOpenXlfd'
undefined reference to `XftTextExtents32'
undefined reference to `XftDrawChange'
undefined reference to `XftDrawSetClip'
undefined reference to `XftDrawString32'
~~~~

Found a stale pull request #15277 and bug report  #15276.

--

### DIFF
--- a/recipes/fltk/all/conanfile.py
+++ b/recipes/fltk/all/conanfile.py
@@ -25,6 +25,7 @@ class FltkConan(ConanFile):
         "with_threads": [True, False],
         "with_gdiplus": [True, False],
         "abi_version": ["ANY"],
+        "with_xft": [True, False],
     }
     default_options = {
         "shared": False,
@@ -32,6 +33,7 @@ class FltkConan(ConanFile):
         "with_gl": True,
         "with_threads": True,
         "with_gdiplus": True,
+        "with_xft": False,
     }
 
     def export_sources(self):
@@ -72,6 +74,8 @@ class FltkConan(ConanFile):
             self.requires("glu/system")
             self.requires("fontconfig/2.14.2")
             self.requires("xorg/system")
+            if self.options.with_xft:
+                self.requires("libxft/2.3.6")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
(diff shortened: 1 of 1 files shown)


RubenRBS:
Hi! Thanks a lot for taking the time to fix this. I think that the proper fix is along the lines of your first PR
The error you got in v2 is due to a missing binary which I've launched a build for, but the v1 one seems to be a library/recipe issue that will take some time to debug.

If you don't have any insights into it, I'm ok with merging as-is, and trying to "properly" fix it later, so linux at least builds meanwhile! Let me know what you prefer :)

jputcu:
Hello @RubenRBS.
We can add `xft` as an option, enabled on v2 and disabled on v1. 
Although a known GUI library, it appears not popular via conan, so if new users come they are hopefully on v2.
V1 users can still enable it if their machine.

recipes/fltk/all/conanfile.py
```diff
@@ -68,10 +70,13 @@ def requirements(self):
         self.requires("libjpeg/9e")
         self.requires("libpng/1.6.40")
         if self.settings.os in ["Linux", "FreeBSD"]:
-            self.requires("opengl/system")
-            self.requires("glu/system")
+            if self.options.with_gl:
+                self.requires("opengl/system")
```
jwillikers, line 74: Since this is only for Linux and FreeBSD, is it possible to require `libglvnd` instead of `opengl/system`?

jwillikers (changes requested):

recipes/fltk/all/conanfile.py
jputcu: @jwillikers I'm currently not using opengl, so I don't know if I can confirm the correct operation. Maybe keep that for a next iteration?

jwillikers: Yeah, that's alright. Thanks!

jwillikers (approved):

PR #20562, opened 2023-10-14
yajl: fix MinGW + package either static or shared lib

SpaceIm:
- Fix build with MinGW (cleanup bloated CMakeLists which was injecting unnecessary msvc warning flags if windows regardless of compiler, leading to `gcc.exe: error: /W4: linker input file not found: No such file or directory` with MinGW).
- Package either static or shared, not both. Indeed yajl builds and installs both flavors, and cci hook didn't catch this issue because static and shared lib names are different. I've preferred removing unwanted flavor afterwards instead of patching CMakeLists, because several executables are always link to static flavor.
- Disable build of doc, example, test, perf
- Better way to install relocatable shared lib for macOS (by removing hardcoded `INSTALL_NAME_DIR` in CMakeLists, and bumping cmake_minimum_required which was very very old and deprecated).
- Better way to install dll into bin folder (fix `RUNTIME DESTINATION` in CMakeLists instead of moving files afterwards).
- Add `YAJL_SHARED` interface definition if shared in `package_info()`.

---

### DIFF
--- a/recipes/yajl/all/conandata.yml
+++ b/recipes/yajl/all/conandata.yml
@@ -2,3 +2,8 @@ sources:
   "2.1.0":
     url: "https://github.com/lloyd/yajl/archive/refs/tags/2.1.0.tar.gz"
     sha256: "3fb73364a5a30efe615046d07e6db9d09fd2b41c763c5f7d3bfb121cd5c5ac5a"
+patches:
+  "2.1.0":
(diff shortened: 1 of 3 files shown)


jwillikers (approved):

RubenRBS (approved):
Thanks!

Issue #20383, opened 2023-10-05
[package] qt6/6.5.2: Cannot build with "compiler.runtime=static"


matheusgomes28:
### Description

When attempting to build my Qt6 application on my Windows machine, it fails on the Qt6 checks for the `static` runtime build. See the logs for the exact details.

Note that, building with the `MSVC_RUNTIME_LIBRARY` set to `MultiThreaded{Debug}DLL` (i.e. `compiler.runtime=dynamic`) then it builds just fine.

As far as I can tell, Qt6 does provide support for building with a static runtime as outlined in their docs, so this is something that should also be supported in the conan recipe. 

### Package and Environment Details

* Package Name/Version: **qt/6.5.2**
* Operating System+version: **Windows 11**
* Conan version: **conan 2.0.13**
* Python version: **Python 3.12.0**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.version=193
os=Windows

### Steps to reproduce

conan install . -sbuild_type=Debug -of=conan --build=missing

### Logs

<details>
<summary>Click to expand logs</summary>
<pre>
> conan install . -sbuild_type=Debug -of=conan --build=missing

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Debug
compiler=msvc
compiler.cppstd=17
compiler.runtime=static
compiler.runtime_type=Debug
compiler.version=193
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=static
compiler.runtime_type=Release
compiler.version=193
os=Windows

======== Computing dependency graph ========
Graph root
    conanfile.py: C:\Users\mathe\development\discotec\conanfile.py
Requirements
    brotli/1.1.0#d56d7bb9ca722942aba17369cb5c0519 - Cache
    bzip2/1.0.8#411fc05e80d47a89045edc1ee6f23c1d - Cache
    double-conversion/3.3.0#33321c201741cc32b51169c6d2d05e60 - Cache
    fmt/10.1.1#cd63809a79574a2f9eb73ca35f16a243 - Cache
    freetype/2.13.0#28bb2425d43c8bf7dbcfb024d847f5fb - Cache
    glib/2.77.0#9616df60560ec0c4548263d838a2c834 - Cache
[...]

PR #20565, opened 2023-10-14
Fixing Qt6 Multimedia Dependency Issue With ShaderTools 

matheusgomes28:
Specify library name and version:  **qt/6.5.2**

I've noticed that my statically built Qt6 on Windows failed when I set the `qtmultimedia` option.

From a quick glance at the code, I noticed that the `default_options` are frozen inside the `configure(...)` method, where the function that is supposed to be setting the module dependency options (like `qtmultimedia` should set `qtshadertools`) runs. This option setting fails with the error `  ConanException: Incorrect attempt to modify option 'qtshadertools' from 'False' to 'True'`. This is documented [in this issue](https://github.com/conan-io/conan-center-index/issues/20383#issuecomment-1762088147).

As a quick fix, we can remove the default value of `qtshadertools` so it does not get frozen, allowing the `_enablemodule` calls in the recipe to succeed setting the dependencies from the `qtmodules6.5.2.conf` file as expected.

I also suspect that other module dependencies may be broken, as the freezing behaviour seems to be a breaking change from Conan 1.x as documented [here](https://github.com/conan-io/conan/issues/10104). I haven't had time to build other optional Qt6 modules yet.

---

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -118,6 +118,7 @@ class QtConan(ConanFile):
         "disabled_features": "",
     }
     default_options.update({module: False for module in _submodules})
+    del default_options["qtshadertools"]
 
     short_paths = True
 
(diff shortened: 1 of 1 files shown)


matheusgomes28:
@ericLemanissier It may be useful to take a look at this PR.

PR #20584, opened 2023-10-15
[scippp] bump version to 1.1.0

hedtke:
Specify library name and version:  **scippp/1.1.0**

I am the author of the library. I just released v1.1.0

One dependency was also bumped in upstream: SCIP to 8.0.4: https://github.com/scipopt/SCIPpp/commit/236dec8b700e281cef5d89744ce587cb5fb3cbfd

Minimum compiler version updated, see upstream https://github.com/scipopt/SCIPpp/pull/15

---

### DIFF
--- a/recipes/scippp/all/conanfile.py
+++ b/recipes/scippp/all/conanfile.py
@@ -33,12 +33,21 @@ class ScipPlusPlus(ConanFile):
 
     @property
     def _compilers_minimum_version(self):
-        return {
-            "gcc": "8",
-            "clang": "7",
-            "apple-clang": "11",
-            "msvc": "192"
-        }
+        # see https://github.com/scipopt/SCIPpp/commit/faa80e753f96094004467c1daa98a7ab4d86f279
+        if Version(self.version) >= "1.1.0":
+            return {
+                "gcc": "8",
+                "clang": "7",
+                "apple-clang": "11",
+                "msvc": "192"
+            }
+        else:
+            return {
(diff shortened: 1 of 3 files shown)


TiloW (approved):

hedtke:
I need help. "This commit cannot be built" but there is no error message

hedtke:
@RubenRBS can I ask for your help here? Where is the problem?

RubenRBS:
Pinging @uilianries, as I'm currently on vacation :)

recipes/scippp/all/conanfile.py
```diff
@@ -34,9 +34,9 @@ def _min_cppstd(self):
     @property
     def _compilers_minimum_version(self):
         return {
-            "gcc": "7",
+            "gcc": "8",
```
uilianries, line 37: I'm curious why did you need to change the minimum compiler version? In case this new version is not compatible (e.g. required C++ standard), it should be validated by package version instead. Otherwise, users will lose access to older versions of this library.

hedtke: Indeed the old versions works fine with the older compiler. The new compiler is now only enforced for the new version.

uilianries (approved):
LGTM

recipes/scippp/all/conanfile.py
uilianries: Thank you for checking it!

jwillikers (approved):

PR #20619, opened 2023-10-17
onnxruntime: simplify patching

valgur:
Replaces the https://github.com/microsoft/onnxruntime/blob/v1.16.1/cmake/external/onnxruntime_external_deps.cmake
file with a custom version, which applies `find_package()` against Conan targets instead of using `FetchContent`.

Includes changes from #20392.


recipes/onnxruntime/all/conanfile.py
```diff
@@ -71,34 +74,32 @@ def configure(self):
...
         self.requires("date/3.0.1")
-        self.requires("re2/20230901")
-        self.requires(f"onnx/{self._onnx_version}")
-        self.requires("flatbuffers/1.12.0")
-        self.requires("boost/1.83.0", headers=True, libs=False, run=False)  # for mp11, header only, no need for libraries to link/run
-        self.requires("safeint/3.0.28")
+        self.requires("re2/20231101")
+        self.requires("flatbuffers/1.12.0")  # v1.* is required, newer versions are not compatible
+        self.requires("boost/1.83.0", headers=True, libs=False)  # for mp11, header only, no need for libraries
+        self.requires("safeint/3.24")
```
mayeut, line 92: 3.24 is in fact older than 3.0.28 & should not be upgraded.

```suggestion
        self.requires("safeint/3.0.28")
```

PR #16453, opened 2023-03-08
protobuf 3.21.9: backport bugfix for GID_MAX macro crash on macOS

jcar87:
Specify library name and version:  **protobuf/3.21.9**

When trying to build `googleapis` with `protobuf/3.21.9`, we may see an error like this on macOS:
```
error: expected member name or ';' after declaration specifiers
  static constexpr GID GID_MAX =
  ~~~~~~~~~~~~~~~~~~~~ ^
/Applications/Xcode_14.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/sys/syslimits.h:86:28: note: In file included from /Users/luisc/dev/conan-io/conan-center-index/recipes/protobuf/all/test_package/test_package.cpp:4:
expanded from macro 'GID_MAX'/Users/luisc/dev/conan-io/conan-center-index/recipes/protobuf/all/test_package/build/apple-clang-13-armv8-gnu17-release/addressbook.pb.h:
834:24: error: expected member name or ';' after declaration specifiers
#define GID_MAX            2147483647U  /* max value for a gid_t (2^31-2) */
                           ^
  static constexpr GID GID_MAX =
  ~~~~~~~~~~~~~~~~~~~~ ^
/Applications/Xcode_14.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/sys/syslimits.h:86:28: note: expanded from macro 'GID_MAX'
#define GID_MAX            2147483647U  /* max value for a gid_t (2^31-2) */
                           ^
1 error generated.
```

This is caused by one of the message types in `googleapis` containing a class called `GID`, which causes the protobuf compiler to declare a symbol called `GID_MAX`, which is defined as a macro by system headers on macOS.

Newer versions of protobuf handle this case by un-defining the macro on macOS.

This PR:
* Backports the fix for this specific case
* Adds a type called `GID` in the test package to ensure that this fixes the issue. The test would fail without the patch being applied. 

After this is merged, it is safe to bump the version of protobuf in the `googleapis` recipe without introducing regressions on macOS (this has been verified locally).

PR #20627, opened 2023-10-17
Adding protobuf/3.5.1

luizfeldmann:
Specify library name and version:  **protobuf/3.5.1**

This is an older version of the library but it's important to have it here because it's the last one to support C++98.

---


lucaskdc:
Regarding the macOS build failure. It happens because there is not patch to the GID fields in the test_package. Other versions of protobuf received those patches (see [](https://github.com/conan-io/conan-center-index/pull/16453/))

https://github.com/conan-io/conan-center-index/pull/16453/

lucaskdc:
Regarding the macOS build failure. It happens because there is no patch to the GID fields for 3.5.1 in the test_package for this version. Other versions of protobuf received those patches (see the PR https://github.com/conan-io/conan-center-index/pull/16453). That bug is officially fixed since 3.22.0, and probably can be easily be backported to 3.7.1 (using the port_def.inc and port_undef.inc mechanism). 

The reason the mentioned PR give to backport this bugfix to 3.21.12 is to allow the googleapis to use it. See [googleapis conanfile](https://github.com/conan-io/conan-center-index/blob/master/recipes/googleapis/all/conanfile.py).

I suggest to change the test_package to let this bug pass in 3.5.1.

uilianries:
Saudações @luizfeldmann and @lucaskdc I don't recommend skipping the test package, it's important to keep reproducibility and validate the generated package. Instead, I would ask to backport patch if possible. You see, without the test package we can't confirm if that package is really working when distributed.

luizfeldmann:
Running the command `conan create -s build_type=Debug -o protobuf:shared=True -tf test_v1_package . protobuf/3.5.1@ --build=missing` I was able to reproduce the error:

```
dyld[13117]: Library not loaded: 'libprotocd.dylib'
  Referenced from: '/Users/jenkins/w/prod-v1/bsr@3/36213/dccff/.conan/data/protobuf/3.5.1/_/_/package/6ba6920b22b03a7cda05599853afac26e7fdfce2/bin/protoc'
  Reason: tried: 'libprotocd.dylib' (no such file), '/usr/local/lib/libprotocd.dylib' (no such file), '/usr/lib/libprotocd.dylib' (no such file), '/Users/jenkins/w/prod-v1/bsr@3/cci-a6bfc911/recipes/protobuf/all/test_v1_package/build/9be9c4d74b252db4497c80d51dbbc27e25fa4789/test_package/libprotocd.dylib'
```

As I'm not familiar with macOS, I'm assuming it's related to `@rpath` warning produced by the [hook](https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H077).

@uilianries Do you maybe have any comments ?

luizfeldmann:
This [PR](https://github.com/protocolbuffers/protobuf/pull/4620) seems to address this issue.

luizfeldmann:
New issue related to source file generated at build-time, see:
https://github.com/protocolbuffers/protobuf/pull/4694
https://github.com/protocolbuffers/protobuf/issues/4704

The build process generates a custom executable `js_embed` which in turn generates the file `src/google/protobuf/compiler/js/well_known_types_embed.cc` before the build can continue.

This method seems to break cross-compilations and because the host cannot run this executable (which is built for the cross compilation target and not for the host architecture).

This [PR](https://github.com/protocolbuffers/protobuf/pull/4709) later resolved the issues by completely removing this step and instead having this file being part of the source instead of auto-generated.

luizfeldmann:
Let's try to backport these patches into the upstream:
https://github.com/protocolbuffers/protobuf/pull/14644
https://github.com/protocolbuffers/protobuf/pull/14641

jcar87:
Hi @luizfeldmann - thank you for your contribution.

unforunately as you have seen - this is too old a version (2017) and keeping it working for newer systems and compilers is too much effort - this wouldn't be `3.5.1`, this would be 3.5.1 with a series of backports. It's also telling that protobuf themselves have not paid https://github.com/protocolbuffers/protobuf/pull/14644 or https://github.com/protocolbuffers/protobuf/pull/14641 much attention.

Accepting this would impose too high a maintenance burden for us. 

I appreciate this is a common use case that in certain users need to stay on too-old versions of certification purposes - that is why we have a variety of strategies for users to maintain their own versions of recipes:
* https://docs.conan.io/2/devops/using_conancenter.html#control-and-customization
* https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires
* https://docs.conan.io/2/tutorial/conan_repositories/setup_local_recipes_index.html

PR #20628, opened 2023-10-17
shapelib: Fix RPATH on macOS

jwillikers:
Specify library name and version:  **shapelib/1.5.0**

Disabling the `USE_RPATH` CMake option in the conanfile.py currently has no effect on the shapelib build system.
This is because the `USE_RPATH` variable is set as a normal variable and not a cache variable and due to the fact that the minimum required version of CMake is 3.7.
As of CMake 3.13, policy CMP0077 fixes this behavior so that the `option()` function in CMake accepts normal variables for setting options.

For now, this PR sets `USE_RPATH` as a cache variable to workaround this issue.
Additionally, the CMake version required in the test packages is updated.

---

### DIFF
--- a/recipes/shapelib/all/conanfile.py
+++ b/recipes/shapelib/all/conanfile.py
@@ -47,7 +47,7 @@ class ShapelibConan(ConanFile):
     def generate(self):
         tc = CMakeToolchain(self)
         tc.variables["BUILD_TESTING"] = False
-        tc.variables["USE_RPATH"] = False
+        tc.cache_variables["USE_RPATH"] = False
         tc.generate()
 
     def build(self):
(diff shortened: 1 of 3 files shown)


RubenRBS (approved):
Thanks a lot for addressing the issue and your patience, both are really appreciated :)

uilianries (approved):
LGTM

SpaceIm:
```
Disabling the USE_RPATH CMake option in the conanfile.py currently has no effect on the shapelib build system.
This is because the USE_RPATH variable is set as a normal variable and not a cache variable and due to the fact that the minimum required version of CMake is 3.7.
As of CMake 3.13, policy CMP0077 fixes this behavior so that the option() function in CMake accepts normal variables for setting options.

For now, this PR sets USE_RPATH as a cache variable to workaround this issue.
Additionally, the CMake version required in the test packages is updated.
```

This change doesn't hurt, but for me it doesn't change anything, it was working fine before (when I run otool -l on dylib file, there is no LC_RPATH), and it still works fine. `USE_RPATH` is a regular CMake option, so variables or cache_variables are both OK.

PR #20638, opened 2023-10-18
proj: remove unused versions

ericLemanissier:
Specify library name and version:  **proj/***

---


recipes/proj/config.yml
```diff
@@ -5,15 +5,3 @@ versions:
...
-  "9.0.1":
-    folder: "all"
-  "9.0.0":
-    folder: "all"
-  "8.2.1":
-    folder: "all"
-  "7.2.1":
-    folder: "all"
-  "6.3.1":
-    folder: "all"
```
SpaceIm, line 19: Shouldn't we keep at least one flavor of each major version?

ericLemanissier: you are right. I misread the official policy. I feel like it could be acceptable if any of these versions are used on conan-center-index. WDYT ?
The call to reduce both CI time and recipe volume is very strong in me

PR #20650, opened 2023-10-18
zstd: do not package both static & shared on windows

SpaceIm:
fix regression from https://github.com/conan-io/conan-center-index/pull/18826 where both shared & static were packaged on Windows if shared=True (see https://github.com/conan-io/conan-center-index/pull/18826#issuecomment-1769360351)

---

### DIFF
--- a/recipes/zstd/all/conanfile.py
+++ b/recipes/zstd/all/conanfile.py
@@ -86,24 +86,25 @@ class ZstdConan(ConanFile):
         rmdir(self, os.path.join(self.package_folder, "share"))
 
         if self.options.shared and self.options.build_programs:
-            #If we built programs we always build static libs,
-            #but if we only want shared libs in the package then remove the static libs
+            # If we built programs we always build static libs,
+            # but if we only want shared libs in the package then remove the static libs
             rm(self, "*.a", os.path.join(self.package_folder, "lib"))
+            rm(self, "*_static.*", os.path.join(self.package_folder, "lib"))
 
     def package_info(self):
         zstd_cmake = "libzstd_shared" if self.options.shared else "libzstd_static"
         self.cpp_info.set_property("cmake_file_name", "zstd")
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):
Great catch @SpaceIm! :)

SpaceIm:
Actually lib name can be hard to predict, so let's keep `collect_libs`, and if static lib has to be removed, we remove `*.a` and `*_static.*` files: https://github.com/facebook/zstd/blob/v1.5.5/build/cmake/lib/CMakeLists.txt#L119-L124

uilianries (approved):
Thank you for spotting it and providing a fix!

recipes/zstd/all/conanfile.py
```diff
@@ -86,24 +86,25 @@ def package(self):
         rmdir(self, os.path.join(self.package_folder, "share"))
 
         if self.options.shared and self.options.build_programs:
```
uilianries, line 88:
```suggestion
        if self.options.shared:
```

Programs are statically linked already, we don't need to package them anyway.

SpaceIm: This branch removes static libs when you have both shared & build_programs options enabled, it's correct as it is. See generate() method.

recipes/zstd/all/conanfile.py
```diff
@@ -86,24 +86,25 @@ def package(self):
...
-            #If we built programs we always build static libs,
-            #but if we only want shared libs in the package then remove the static libs
+            # If we built programs we always build static libs,
+            # but if we only want shared libs in the package then remove the static libs
```
SpaceIm, line 90:
```suggestion
            # If we build programs we have to build static libs (see logic in generate()),
            # but if shared is True, we only want shared lib in package folder.
```

uilianries (approved):

jwillikers (approved):

PR #20672, opened 2023-10-18
Bump crashpad's zlib requirement to [>=1.2.11 <2] range (v1 only)

RubenRBS:
Part of the zlib range migration, this PR only supports Conan v1


uilianries:
It's so outdated, that not even our current Conan 1.x pipeline is able to build it 💥

RubenRBS:
Closing then, if noone has yet submitted a PR to port it for v2 we shall then wait

PR #20686, opened 2023-10-19
crossguid: add recipe

toge:
Specify library name and version:  **crossguid/***

---

### DIFF
--- a/recipes/crossguid/all/conanfile.py
+++ b/recipes/crossguid/all/conanfile.py
@@ -56,6 +56,7 @@ class CrossGUIDConan(ConanFile):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
         tc.variables["XG_TESTS"] = False
         tc.generate()
         dpes = CMakeDeps(self)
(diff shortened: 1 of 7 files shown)


czoido (approved):

uilianries (approved):
LGTM

recipes/crossguid/all/conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.settings.os == "Linux":
+            self.requires("libuuid/1.0.3", transitive_headers=True, transitive_libs=True)
```
SpaceIm, line 48: Shouldn't it be replaced by `util-linux-libuuid` since all other recipes depend on it instead of `libuuid` recipe (and both recipes can't live together in the same graph since there is a provides attribute in `util-linux-libuuid`)?

SpaceIm: I've opened https://github.com/conan-io/conan-center-index/pull/20902

SpaceIm: Moreover traits of libuuid were wrong. These undefined references come from a mistake in this CMake patch where you don't properly link libuuid actually (there is no such thing as libuuid_LIBRARY in CMakeDeps)

PR #20703, opened 2023-10-20
Opencolorio 2.3.0

irieger:
Specify library name and version:  **opencolorio/***

Version bump and dependency bump

Based on #19549 which started the work but so far the contributor hasn't answered. As I need the dependency updated and would prefer to directly also use the new version in a library that requires OpenColorIO, I want to move this forward.

---


toge:
@irieger 
Thanks you for contribution!
I will close my PR.

irieger:
> @irieger Thanks you for contribution! I will close my PR.

You're welcome. Didn't want to steal your PR/contribution but just got a bit impatient waiting for anyone to react on your PR. Also let's see what CI says, I think your PR didn't build on the Conan 2 pipeline side (for 1.1.1).

Not sure if it is expected for packages to support Conan 2 for all old versions? (I mean I'd prioritize having the newest version build on Conan 2 and maybe prevent regressions for old packages that build on Conan 2 before, but not enforce old builds?)

irieger:
Hey @conan-io/barbarians
what is the way forward with this? The Conan 2.0 pipeline fails for one version (and the canceled one would fail too). But that would be the same for the current state of the repository, as OpenColorIO before version 2.2.0 required openexr < 3 and openexr is only building on conan 2 from version 3.

Also as I haven't got any meaningful feedback, is a PR normally only looked at when it is passing CI? This would be the problem for quite a few projects I assume where older versions might not be easily portable nor needed for projects using a current Conan, but where maybe for backwards compatibility it would be nice to keep old version? (I don't see the point of removing the old versions besides pleasing the CI?)

valgur (approved):

recipes/opencolorio/all/conandata.yml
```diff
@@ -9,6 +12,16 @@ sources:
     url: "https://github.com/AcademySoftwareFoundation/OpenColorIO/archive/v1.1.1.tar.gz"
     sha256: "c9b5b9def907e1dafb29e37336b702fff22cc6306d445a13b1621b8a754c14c8"
 patches:
+  "2.3.0":
+    - patch_file: "patches/2.3.0-0001-fix-cmake-source-dir-and-targets.patch"
+      patch_description: "use cci package, use PROJECT_BINARY_DIR/PROJECT_SOURCE_DIR"
+      patch_type: "conan"
+    - patch_file: "patches/2.3.0-0002-fix-pystring.patch"
+      patch_description: "fix include path for pystring"
+      patch_type: "conan"
```
franramirez688, line 21: This fix is reported upstream?

franramirez688 (commented):
Hi @irieger 

Thank you so much for this PR. It looks great! 😄 Only a doubt about the patches.

recipes/opencolorio/all/conandata.yml
irieger: Good point. Just about to start the discussion and created a proposed MR: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1900 and will follow up in their Slack channel about those changes.

irieger: There is another MR in the works that should solve the problem as we haven't been the only ones running into problems where I give feedback on now: https://github.com/remia/OpenColorIO/pull/52

irieger: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1901 has now been raised. Tested it locally by adding a zip of the PR branch as a separate version and it build without the patch.

irieger: @franramirez688 Any chance to move forward with this?

The upstream seems to be on a good track btw. and got the first approval.

franramirez688: Hi @irieger 
Sorry for the delay. Thanks a lot for the info 😁 
Then, we should put that info in the conandata.yml:

```suggestion
      patch_type: "portability"
      patch_source: "https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1901"
```

franramirez688 (approved):

jwillikers (approved):

PR #20704, opened 2023-10-20
antlr4: add recipe

toge:
Specify library name and version:  **antlr4/***

It is not an antlr runtime but an antlr parser generator.

---


RubenRBS:
Hi @toge, can you give me some insight into the need for this recipe? What's its use-case/usage? As far as I can see, it's part of a bigger project, but this one is a java program meant to generate parsers, no? Happy to get more info on it! (And thanks for the PR!)

jwillikers (approved):

toge:
@RubenRBS 
Sorry I missed your comments.

You are correct. This recipe makes available the code generator part of antlr4.
Conan has a library, antlr4-cppruntime, to run the C++ code generated by antlr4, but since there is no tool in conan to generate  code to use this, I regret that we cannot create an environment to use antlr4 only with conan. 
For this reason, I thought I would create a recipe to make the code generator portion of antlr4 available.

danimtb (approved):
Thanks!

michalwidera:
> Hi @toge, can you give me some insight into the need for this recipe? What's its use-case/usage? As far as I can see, it's part of a bigger project, but this one is a java program meant to generate parsers, no? Happy to get more info on it! (And thanks for the PR!)

Hi Ruben. I found that this antlr has been merged. I see that antlr4 and java appeard in cache folder. But Your question is still open. I have no idea how to call this tool from conan cache.
Did you saw working example of this ?
I was trying solve this with tricky ways of ${antlr4_BIN_DIRS_DEBUG}/antlr4 but this looks ugly.
This could be death code here.

PR #20706, opened 2023-10-20
[avahi] Use internal strlcpy always

uilianries:
Specify library name and version:  **avahi/0.8**

When trying to build `avahi/0.8` internally using tapaholes we found the follow error:

```
[100%] Linking C executable test_package
/opt/conan/binutils/bin/ld: /home/conan/w/prod-v2/bsr/25159/ddabe/p/b/avahibcaf8397460a4/p/lib/libavahi-common.a(libavahi_common_la-domain.o): in function `avahi_service_name_split':
domain.c:(.text+0xc68): undefined reference to `strlcpy'
```

The [strlcpy](https://linux.die.net/man/3/strlcpy) is a method provided by [libbsd](https://libbsd.freedesktop.org/wiki/), not always available in CCI Docker images, plus, creating inconsistent scenarios where we link a system shared library that is not expected for most of linux distros.

- This PR enforces to use the internal implementation of `strlcpy` always, avoiding linkage errors.
- Expose include folder as root, just like used by [avahi examples](https://github.com/lathiat/avahi/tree/master/examples)
- Consume basic function from avahi to validate its own library, not only the dns_sd.

This PR is required by #20287

----

I tried to build libbsd locally, but it requires a newer glibc version due md5 method support. So we will not have it so soon in CCI. 

---


jwillikers:
@uilianries I think you can just use `tc.configure_args.append("ac_cv_func_strlcpy=no")` to disable this with the build system instead of having to use a patch.

uilianries:
@jwillikers Thank you! I just updated the PR according your suggestion.

jwillikers (approved):

RubenRBS (approved):

jwillikers:
I'm seeing this error on Linux now:

```
/home/jordan/.conan2/p/b/avahi98352b3d5ad69/b/src/avahi-common/domain.c:482:15: error: static declaration of ‘strlcpy’ follows non-static declaration
  482 | static size_t strlcpy(char *dest, const char *src, size_t n) {
      |               ^~~~~~~
In file included from /home/jordan/.conan2/p/b/avahi98352b3d5ad69/b/src/avahi-common/domain.c:24:
/usr/include/string.h:506:15: note: previous declaration of ‘strlcpy’ with type ‘size_t(char * restrict,  const char * restrict,  size_t)’ {aka ‘long unsigned int(char * restrict,  const char * restrict,  long unsigned int)’}
  506 | extern size_t strlcpy (char *__restrict __dest,
```

PR #20721, opened 2023-10-21
[google-cloud-cpp] bump requirements and other fixes

coryan:
Specify library name and version:  **google-cloud-cpp/v2.5.0** and **google-cloud-cpp/v2.12.0**.

Several requirements need version bumps. Where possible I chose version ranges over pinning specific versions.

I had to backport fixes for MSVC. I am only guessing, but it seems the toolchain changed in the CI system and now spelling `||` as `or` results in build errors.

I also disabled `-Werror` because the code in `v2.5.0` has warnings with newer toolchains. This is to be expected, and that is why the project no longer enables `-Werror` by default.

---

### DIFF
--- a/recipes/google-cloud-cpp/2.x/conanfile.py
+++ b/recipes/google-cloud-cpp/2.x/conanfile.py
@@ -138,19 +138,16 @@ class GoogleCloudCppConan(ConanFile):
         get(self, **self.conan_data["sources"][self.version], destination=self.source_folder, strip_root=True)
 
     def requirements(self):
-        # Updating Protobuf requires updates to the generated code, even within
-        # a minor version.
+        # These must remain pinned in conan index.
         self.requires("protobuf/3.21.12", transitive_headers=True)
-        # Each release of Abseil is a new major version, not guaranteed to be
-        # compatible with previous verssions.
-        self.requires("abseil/[>=20230125.3.0 <20230126]", transitive_headers=True)
-        # The rest are more semver-like and require less pinning.
-        self.requires("grpc/[>=1.50]", transitive_headers=True)
-        self.requires("nlohmann_json/[>=3.10 <4]")
-        self.requires("crc32c/[>=1.1 <2]")
-        self.requires("libcurl/[>=7.47 <9]")
+        self.requires("abseil/20230125.3", transitive_headers=True)
+        self.requires("grpc/1.54.3", transitive_headers=True)
+        self.requires("nlohmann_json/3.11.2")
+        self.requires("crc32c/1.1.2")
+        # The rest require less pinning.
+        self.requires("libcurl/[>=7.78 <9]")
         self.requires("openssl/[>=1.1 <4]")
-        self.requires("zlib/[>=1.2 <2]")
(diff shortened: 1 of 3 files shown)


RubenRBS:
Hi @coryan sorry for the delay, the PR message made me think it would be automatically merged and I didn't check the contents until now (and realized it wouldn't, so sorry to keep you waiting!)

recipes/google-cloud-cpp/2.x/conanfile.py
```diff
@@ -138,23 +138,29 @@ def source(self):
...
+        # Updating Protobuf requires updates to the generated code, even within
+        # a minor version.
+        self.requires("protobuf/3.21.12", transitive_headers=True)
+        # Each release of Abseil is a new major version, not guaranteed to be
+        # compatible with previous verssions.
+        self.requires("abseil/[>=20230125.3.0 <20230126]", transitive_headers=True)
+        # The rest are more semver-like and require less pinning.
+        self.requires("grpc/[>=1.50]", transitive_headers=True)
+        self.requires("nlohmann_json/[>=3.10 <4]")
+        self.requires("crc32c/[>=1.1 <2]")
```
RubenRBS, line 150: We appreciate the comments for these ones, but besides libcurl, openssl and zlib, we're not _quite yet_ ready to allow general semver usage on libraries (But we're close to it!) - For now these should be pinned

recipes/google-cloud-cpp/2.x/conanfile.py
```diff
@@ -138,23 +138,29 @@ def source(self):
...
+        self.requires("libcurl/[>=7.47 <9]")
         self.requires("openssl/[>=1.1 <4]")
-        self.requires("zlib/1.2.13")
+        self.requires("zlib/[>=1.2 <2]")
```
RubenRBS, line 153: Conan 2.0.14 will improve the situation here with https://github.com/conan-io/conan/pull/14912 allowing to calculate range intersections of missmatching but compatible ranges, but it's still better if we have as unified set of ranges for libraries as possible - see here for the current ones https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges (and additionaly, zlib's is `[>=1.2.11 <2]`)

RubenRBS (changes requested):
Thanks! The backports look good, but the verison ranges need a bit more time from our side (But I still appreciate the comments on those, very informative)

recipes/google-cloud-cpp/2.x/conanfile.py
coryan: Done.

recipes/google-cloud-cpp/2.x/conanfile.py
coryan: Ack.  Done.

coryan:
Looks like something is broken in the CI?  Please let me know if I should take some action.

RubenRBS:
Let's see if it it builds now

RubenRBS (approved):

uilianries (approved):
LGTM

PR #20732, opened 2023-10-22
pulseaudio: switch to Meson, add new versions, rename CMake targets

valgur:
Follow-up for #20275 to have an actual PR to merge.

`pulseaudio` have had Meson build system support for a while and the latest versions have dropped Autotools altogether. This PR updates the recipe accordingly.

I also noticed that the recipe does not explicitly set the CMake file and target names and I did not see a reference to an existing convention either. So I renamed the targets based on the CMake modules used by KDE at https://api.kde.org/ecm/find-module/FindPulseAudio.html, which seems authoritative enough to me.
@ericLemanissier, I saw you left a comment in the recipe regarding generated CMake variables. Does the above change look reasonable to you?

TODO:

- [ ] I have not check all of the options and dependencies used by the project's [`meson.build`](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/blob/v16.1/meson.build). Only made minimal tweaks to get the existing functionality in the recipe to work.
- [ ] The Meson-based build system should support Windows as well. Needs testing.


SpaceIm:
> I also noticed that the recipe does not explicitly set the CMake file and target names and I did not see a reference to an existing convention either. So I renamed the targets based on the CMake modules used by KDE at https://api.kde.org/ecm/find-module/FindPulseAudio.html

> which seems authoritative enough to me.

It's not.

valgur:
> It's not.
> 
> There are 2 authoritative sources for a library:
> 
> * what is provided by a regular CMake installation (basically https://cmake.org/cmake/help/latest/manual/cmake-modules.7.html)
> * what is provided by regular installation of this library itself (a CMake config file usually)
> 
> Everything else is confidential and shouldn't be used as a reference.

I kind of agree, but IMHO the goal of the packages provided by CCI should be to work as frictionlessly as possibly with existing code bases and environments. Since PulseAudio is most likely used via a system package manager, I think matching the naming conventions used by Ubuntu and Debian would be quite reasonable. Both of them include a `/usr/lib/x86_64-linux-gnu/cmake/PulseAudio/PulseAudioConfig.cmake` CMake config file (https://packages.ubuntu.com/mantic/amd64/libpulse-dev/filelist) with the following content:

```cmake
set(PULSEAUDIO_FOUND TRUE)

set(PULSEAUDIO_VERSION_MAJOR 16)
set(PULSEAUDIO_VERSION_MINOR 1)
set(PULSEAUDIO_VERSION 16.1)
set(PULSEAUDIO_VERSION_STRING "16.1")

find_path(PULSEAUDIO_INCLUDE_DIR pulse/pulseaudio.h HINTS "/usr/include")
find_library(PULSEAUDIO_LIBRARY NAMES pulse libpulse HINTS "/usr/lib/x86_64-linux-gnu")
find_library(PULSEAUDIO_MAINLOOP_LIBRARY NAMES pulse-mainloop-glib libpulse-mainloop-glib HINTS "/usr/lib/x86_64-linux-gnu")
```

With that in mind I still lean in favor of renaming it to `PulseAudio` and also additionally exporting the uppercase CMake variables used in the above config file.

SpaceIm:
It turns out that there is a CMake config file upstream: https://github.com/pulseaudio/pulseaudio/blob/v16.99.1/PulseAudioConfig.cmake.in#L1-L12

So I suggest to not define `cmake_find_mode`, because there is no Find file, and to not define `cmake_target_name`, because there is no cmake target, so it could be any arbitrary name (why not `pulseaudio::pulseaudio` instead of `PulseAudio::PulseAudio`, so stick to default CMakeDeps name `pulseaudio::pulseaudio` since no one is authoritative).

Moreover you'll have to create these custom CMake vars, because they won't be defined out of the box by CMakeDeps:

- `PULSEAUDIO_INCLUDE_DIR`
- `PULSEAUDIO_LIBRARY`
- `PULSEAUDIO_VERSION_MAJOR`
- `PULSEAUDIO_VERSION_MINOR`
- `PULSEAUDIO_VERSION_STRING`

valgur:
Closing in favor of #22441.

PR #20737, opened 2023-10-22
escape: add recipe

toge:
Specify library name and version:  **escape/***

escape is one of the termox(https://github.com/a-n-t-h-o-n-y/TermOx)'s dependency libraries.

---


recipes/escape/all/conanfile.py
```diff
@@ -0,0 +1,78 @@
...
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
```
uilianries, line 55:
```suggestion
            )
        if self.setting.build_type == "Debug":
            raise ConanInvalidConfiguration(f"{self.ref} does not support debug installation.")
```

The upstream only installs Release as build type: https://github.com/a-n-t-h-o-n-y/Escape/blob/main/CMakeLists.txt#L58

I don't think you need to patch it, instead, just raise an error. I'll open a PR to the upstream asking to change it.

toge:
@uilianries 
Now CI is passed.
Could you review it again?

PR #20036, opened 2023-09-20
octomap: add version 1.9.8

valgur:

PR #20748, opened 2023-10-23
octomap: add a missing patch to the latest version

valgur:
> Thanks, but you have forgotten a patch which was in previous versions, allowing to build with C++20. This patch (https://github.com/OctoMap/octomap/pull/394) has been merged upstream yes, but octomap 1.9.8 has been released 1 year before this merge.

_Originally posted by @SpaceIm in https://github.com/conan-io/conan-center-index/issues/20036#issuecomment-1741878489_


RubenRBS (approved):
Damn, good catch @SpaceIm :)

jwillikers (approved):

SpaceIm:
FYI, it was not necessary to create an new patch (and not necessary to add these extra patches for octovis which is not build in octomap recipe), the one of 1.9.6 can be reused in 1.9.8

PR #20752, opened 2023-10-23
libxmlpp: relocatable shared lib for macos + revert to default cmake target name

SpaceIm:
- Set install_name_dir to `@rpath` for shared lib macOS, so that it's relocatable.
- There is no CMake config file upstream, so there is no reason to do crazy things like to invent some random CMake target name. Moreover there is only one .pc file installed: `libxml++-<version>.pc`.
So `package_info()` can be simplified.
- rename base recipe folder from 2.x.x to all (2.x.x makes little sense, there are both 2.42.1 and 5.0.1 in this folder...)
- add package_type

---


RubenRBS (approved):
Oh my, this looks so much better now!

valgur (approved):

jwillikers (approved):

PR #20788, opened 2023-10-27
[sail] Update to 0.9.0 and remove rc2 as incompatible

HappySeaFox:
Specify library name and version:  **sail/0.9.0**

0.9.0 release. 0.9.0-rc2 has been deprecated and removed.

---


HappySeaFox:
@jwillikers Hi! Could you please review?

recipes/sail/all/conanfile.py
```diff
@@ -18,25 +18,21 @@ class SAILConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "thread_safe": [True, False],
-        "with_avif": [True, False],
```
uilianries, line 21: It's not clear what's high/medium/low priority, any option should be easy to understand, that's why we follow the upstream options, or try to add something known. I would ask to revert it and keep the library name instead. In case some of them are not compatible to be built together, then you can use `validate()` to check and raise an `ConanInvalidConfiguration`.

HappySeaFox: Hi. Low/high/highest etc. are just sets of codes as implemented in upstream. It's much easier to maintain codecs in sets as SAIL gets more and more of them, and I don't want to end up having 25 different `with_[codec]` options. SAIL still supports enabling or disabling individual codecs along with sets. Please share what you think about it. If you think individual `with_[codec]` options must be kept along with sets, I will update the recipe.

uilianries: I'll talk with Conan team about it. Indeed is much easier having a set of codecs than each codec as an option. This change is a breaking change per say, because changes a bunch of options.

HappySeaFox: Added `_codecs` suffix to the priorities to make priorities a little bit more clear.

recipes/sail/all/conanfile.py
```diff
@@ -51,23 +77,34 @@ def configure(self):
...
+        del self.info.options.with_gif
+        del self.info.options.with_jpeg2000
+        del self.info.options.with_jpeg
+        del self.info.options.with_png
+        del self.info.options.with_tiff
+        del self.info.options.with_webp
+
+    def validate(self):
+        for option_name in ["with_avif", "with_gif", "with_jpeg2000", "with_jpeg", "with_png", "with_tiff", "with_webp"]:
+            if getattr(self.options, option_name) != "deprecated":
```
uilianries, line 105:
```suggestion
            if self.options.get_safe(option_name, "deprecated") != "deprecated":
```

recipes/sail/all/conanfile.py
```diff
@@ -18,27 +18,53 @@ class SAILConan(ConanFile):
...
+        "with_jpeg2000": "Deprecated",
+        "with_jpeg": "Deprecated",
+        "with_png": "Deprecated",
+        "with_tiff": "Deprecated",
+        "with_webp": "Deprecated",
+        "with_highest_priority_codecs": "Enable highest priority codecs",
+        "with_high_priority_codecs": "Enable high priority codecs",
+        "with_medium_priority_codecs": "Enable medium priority codecs",
+        "with_low_priority_codecs": "Enable low priority codecs",
+        "with_lowest_priority_codecs": "Enable lowest priority codecs",
```
uilianries, line 64:
```suggestion
        "with_highest_priority_codecs": "Enable highest priority codecs",
        "with_high_priority_codecs": "Enable high priority codecs",
        "with_medium_priority_codecs": "Enable medium priority codecs",
        "with_low_priority_codecs": "Enable low priority codecs",
        "with_lowest_priority_codecs": "Enable lowest priority codecs",
```

@HappySeaFox Do see a better way to explain it? Like, which codec could be inserted in each option, or maybe, what refers a specific priority? I searched in https://github.com/HappySeaFox/sail/blob/master/BUILDING.md#cmake-options-overview, but TBH, I didn't find any clue.

HappySeaFox: Hi! Yes, every codec has explicit priority defined. See https://github.com/HappySeaFox/sail/blob/3bddd3f50a710cbf000162de555ef20e4e9a2d0a/src/sail-codecs/CMakeLists.txt#L7, https://github.com/HappySeaFox/sail/blob/3bddd3f50a710cbf000162de555ef20e4e9a2d0a/src/sail-codecs/avif/avif.codec.info.in#L6 as an example

uilianries: Thank you! I just updated the list, according the other codecs.

uilianries (approved):
LGTM

jwillikers (approved):

PR #20791, opened 2023-10-27
Add ohPipeline

DoomHammer:
Specify library name and version:  **ohpipeline/1.39.1000**

This library is required to build audio products based on the OpenHome protocol.

---

- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated. <- I'm not sure how to make them work with Conan 2.0 as the 1.0 instructions don't work.


DoomHammer:
@conan-io/barbarians any chance to use a more recent CMake?

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,99 @@
...
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", src=self.recipe_folder, dst=os.path.join(self.export_sources_folder, "src"))
+        export_conandata_patches(self)
+
```
uilianries, line 31:
```suggestion

    def build_requirements(self):
        self.tool_requires("cmake/[>3.XX <4]")

```

Let's install cmake directly as tool requirement. Don't worry, users still can manage it by their profile.

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,110 @@
...
+        self.requires("faac/1.30", transitive_headers=True, transitive_libs=True)
+        self.requires("flac/1.4.2", transitive_headers=True, transitive_libs=True)
+        self.requires("ogg/1.3.5", transitive_headers=True, transitive_libs=True)
+        self.requires("vorbis/1.3.7", transitive_headers=True, transitive_libs=True)
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
```
uilianries, line 54:
```suggestion
        tc = CMakeToolchain(self)
        # INFO: AacFdkBase.cpp uses #include <aacdecoder_lib.h>
        libfdk_include_dir = os.path.join(self.dependencies["libfdk_aac"].cpp_info.includedirs[0], "fdk-aac")
        tc.variables["CMAKE_CXX_FLAGS"] = tc.variables.get("CMAKE_CXX_FLAGS", "") + " -I{}".format(libfdk_include_dir)

```

fishy but works, let's pass it, so we no longer need a patch.

recipes/ohpipeline/all/patches/0003-ohpipeline.patch
```diff
@@ -0,0 +1,57 @@
...
+ 
+ using namespace OpenHome;
+@@ -391,7 +392,9 @@ Track* Shuffler::NextTrackRef(TUint aId)
+                 else if (index == iShuffleList.size()-1) {
+                     // we've run through the entire list
+                     // prefer re-shuffling over repeating the order of tracks if we play again
+-                    std::random_shuffle(iShuffleList.begin(), iShuffleList.end());
++                    std::random_device rd;
++                    std::mt19937 g(rd());
++                    std::shuffle(iShuffleList.begin(), iShuffleList.end(), g);
```
uilianries, line 43: The old warrior `std::random_shuffle` was deprecated in C++14 and removed in C++17, but the upstream did change it, right? Could please open a PR to the upstream with this change? Usually we do not accept this kind of portability, because the upstream is not C++17 tested as well. I still prefer keeping C++17 not supported with `check_max_cppstd` method, e.g https://github.com/conan-io/conan-center-index/blob/master/recipes/exiv2/all/conanfile.py#L108

uilianries (commented):
Thank you for your PR!! Please, take a look on my review 😄

recipes/ohpipeline/all/patches/0001-ohpipeline.patch
```diff
@@ -0,0 +1,26 @@
+diff --git a/OpenHome/Media/Codec/AacFdkBase.cpp b/OpenHome/Media/Codec/AacFdkBase.cpp
+index 65f45365..e870cfe9 100644
+--- a/OpenHome/Media/Codec/AacFdkBase.cpp
++++ b/OpenHome/Media/Codec/AacFdkBase.cpp
+@@ -7,7 +7,7 @@
+ #include <string.h>
+ 
+ extern "C" {
+-#include <aacdecoder_lib.h>
++#include <fdk-aac/aacdecoder_lib.h>
```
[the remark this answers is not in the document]
DoomHammer: How?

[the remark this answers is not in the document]
DoomHammer: Ah, I see it now :)

recipes/ohpipeline/all/CMakeLists.txt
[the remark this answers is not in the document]
DoomHammer: I will send it. Unfortunately, I don't have good experience with sending patches upstream with this org but at least we'll have it there for documentation purposes :)

Do you think it'd make sense to split the main CMakeLists.txt into several smaller ones?

As for Waf: I tried using it form building a Conan package, but creating CMake turned out to be easier...

recipes/ohpipeline/all/patches/0003-ohpipeline.patch
DoomHammer: The upstream uses C++11. Clang on Mac uses 14 by default, I think and that's why the change was required.

Changes CMake to use C++11 so this should no longer be necessary.

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", src=self.recipe_folder, dst=os.path.join(self.export_sources_folder, "src"))
+        export_conandata_patches(self)
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>3.23 <4]")
+
+    def requirements(self):
+        self.requires("ohnet/1.36.5182", transitive_headers=True, transitive_libs=True)
```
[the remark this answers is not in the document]
DoomHammer: Removed all transitives other than ohNet

recipes/ohpipeline/all/CMakeLists.txt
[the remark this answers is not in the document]
uilianries: Let's keep a single cmakefile for now. The current one is large, but simple to read. I totally understand trying to use another build tool that is not so supported by Conan makes life harder. There are other PRs that we accepted a custom cmake file to build the project.

uilianries:
> Ok, now I have some fPIC woes. Any idea how to approach those, @uilianries ?

I'm curious if it was because that workaround involving cmake_cxx_flags or is something related the the cmakefile. I'm not on my computer now, but I'll build it locally tomorrow.

DoomHammer:
May this be caused by improper fPIC handling in ohNet?

uilianries:
@DoomHammer Sorry the delay, I made a PR to your branch here: https://github.com/DoomHammer/conan-center-index/pull/1

Please, consider it as a fix for the current error. The fPIC error is occurring because all libraries are configured to be built as static, but Conan build both static and dynamic.

DoomHammer:
@uilianries do you think ohNet library should also be updated regarding the static changes?

DoomHammer:
It's very hacky but it works. @uilianries @memsharded any ideas if this could be improved?

recipes/ohpipeline/all/CMakeLists.txt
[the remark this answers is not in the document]
valgur: I would recommend forking the library or at least creating a separate repo for the CMake project files.

valgur (changes requested):
The CMakeLists.txt is too complex to be maintained by CCI. The project should be forked or at least the CMake files maintained in a separate repo.

DoomHammer:
@valgur is it better now?

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+        self.requires("openssl/3.2.0")
+        self.requires("libmad/0.15.1b")
+        self.requires("alac/cci.20121212")
+        self.requires("libfdk_aac/2.0.2")
+        self.requires("faac/1.30")
+        self.requires("flac/1.4.2")
```
valgur, line 52:
```suggestion
        self.requires("openssl/[>=1.1 <4]")
        self.requires("libmad/0.15.1b")
        self.requires("alac/cci.20121212")
        self.requires("libfdk_aac/2.0.3")
        self.requires("faac/1.30")
        self.requires("flac/1.4.3")
```

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+            raise ConanInvalidConfiguration(f"{self.ref} doesn't support shared")
```
valgur, line 36: This check needs to be moved to `validate()`.

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", src=self.recipe_folder, dst=os.path.join(self.export_sources_folder, "src"))
```
valgur, line 39:
```suggestion
```

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+        export_conandata_patches(self)
```
valgur, line 40: Can remove `export_conandata_patches(self)` and `apply_conandata_patches(self)` as well.

recipes/ohpipeline/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "7":
+            raise ConanInvalidConfiguration(f"{self.ref} doesn't support GCC < 7")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
```
valgur, line 64:
```suggestion
        venv = VirtualBuildEnv(self)
        venv.generate()

        tc = CMakeToolchain(self)
```
Required for the CMake tool_requires.

valgur:
@DoomHammer I opened a PR with all of my recipe improvement suggestions: https://github.com/DoomHammer/conan-center-index/pull/2

Thank you for moving the CMakeLists.txt into its own repo! The recipe looks much better now.

DoomHammer:
@valgur do you see the entire error message? I cannot deduce what's the problem based on what's presented here and I also cannot reproduce the problem locally.

RubenRBS:
Hi @DoomHammer - I've restarted the PR's CI - We had an outagge with one of the OSx agents that caused some builds to fail without logs. It's now fixed so let's see if this builds now :) Thanks a lot for your patience!

valgur:
@DoomHammer The macOS Debug (but not Release) build fails with
```
In file included from /Users/jenkins/workspace/prod-v1/bsr@2/26380/effdd/.conan/data/ohpipeline/1.139.1000/_/_/build/2a909c6f087d7d10f9d1a2e759cb2f39f2dcdb77/src/OpenHome/Web/ConfigUi/ConfigUiMediaPlayer.cpp:8:
/Users/jenkins/workspace/prod-v1/bsr@2/26380/effdd/.conan/data/ohpipeline/1.139.1000/_/_/build/2a909c6f087d7d10f9d1a2e759cb2f39f2dcdb77/src/OpenHome/Av/Qobuz/Qobuz.h:15:10: fatal error: 'Generated/CpAvOpenhomeOrgPlaylist1.h' file not found
#include <Generated/CpAvOpenhomeOrgPlaylist1.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```
If you can't figure out a fix or a workaround for this, you can probably just raise a ConanInvalidConfiguration exception for Macos + Debug.

DoomHammer:
MacOS + Debug works for me locally albeit with a different clang version. But what we see here is a lack of file that's used both in Debug and Release. So this is extremely weird. A race condition maybe?

DoomHammer:
Which CMake generator does the CI use? Is it Makefile or Ninja or something else?

valgur:
It uses the default CMake generator - Unix Makefiles.
Did you try with Conan v1 as well?

DoomHammer:
Yes, Conan v1 seems to work for me. I found a possible fix, though, so I'm trying it now.

DoomHammer:
Nope, I only made that worse...

DoomHammer:
OMG, we made it 😱

recipes/ohpipeline/all/patches/001-missing-include.patch
valgur, line 1: Please submit it to the upstream project as well, if you already haven't.

valgur (approved):
Nice! Thanks!

recipes/ohpipeline/all/patches/001-missing-include.patch
DoomHammer: Exhibit A: https://github.com/openhome/ohPipeline/pull/8

DoomHammer: Also: I see this has been fixed in the more recent versions, which I want to handle after this one is merged. It's been lingering way to long already :P

recipes/ohpipeline/all/conandata.yml
```diff
@@ -0,0 +1,13 @@
+sources:
+  "1.139.1000":
+    source:
+      url: "https://github.com/openhome/ohPipeline/archive/refs/tags/ohMediaPlayer_1.139.1000.tar.gz"
+      sha256: "070eac6d974bbe520d0b292df1ca4a72160a2b607a60dabdaa0971c3ef96e5e4"
+    cmake:
+      url: "https://raw.githubusercontent.com/merakiacoustic/ohPipeline/ohMediaPlayer_1.139.1000_meraki2/CMakeLists.txt"
+      sha256: "f9e11d4adb8426a8375dae02e3b641d66efc6ef3abd6f96cb81f6e85352083a3"
```
danimtb, line 8: I know I might be a bit late for the review but think we should not incorporate new build systems to a 3rd party library as there might be errors or subtle behavior changes for the end user. However, I understand the effort here and the usage of an external fork-provided CMakeLists. For future iterations of this recipe, let's try to use the original waf build system and stick to only the sources and instructions provided by the library. Thanks!

DoomHammer: Making the original work with Conan probably requires more code than just using CMake but ends up being less flexible...

DoomHammer: True. But it would be nightmare mostly for me, either way ;) The current version of the library doesn't even compile unless you have non-free SDK, so I'll need to patch it to handle this as an option.

PR #20792, opened 2023-10-27
libwebm: Update to 1.0.0.31 to fix C++20 builds.

simoncent:
Specify library name and version:  **libwebm/1.0.0.31**

The library makes use of some deprecated features which were removed in the C++20 spec. Some compilers are now getting around to removing these, for example older versions of this library no longer build with XCode 15 when targeting C++20. 

This update fixes the build when targeting C++20, without breaking it when targeting older spec versions.

---


uilianries (approved):
LGTM

jwillikers (approved):

PR #20798, opened 2023-10-28
openssl 1.x: cleanup recipe after removal of versions < 1.1.1 in #20441

SpaceIm:
https://github.com/conan-io/conan-center-index/pull/20441 has removed versions < 1.1.1 but did not remove all the logic of these versions in recipe.

---

### DIFF
--- a/recipes/openssl/1.x.x/conanfile.py
+++ b/recipes/openssl/1.x.x/conanfile.py
@@ -88,7 +88,7 @@ class OpenSSLConan(ConanFile):
         "no_tls1": [True, False],
         "capieng_dialog": [True, False],
         "enable_capieng": [True, False],
-        "openssldir": ["ANY", None]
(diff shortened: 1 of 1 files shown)


uilianries (approved):
LGTM. Nice, much clear now!

recipes/openssl/1.x.x/conanfile.py
```diff
@@ -85,7 +32,6 @@ class OpenSSLConan(ConanFile):
     settings = "os", "arch", "compiler", "build_type"
     options = {
         "no_threads": [True, False],
-        "no_zlib": [True, False],
```
Croydon, line 88: In the past, we did not remove options instantly, but printed a warning first. What is the current policy on that?

SpaceIm: this option is already removed for versions 1.1.1 in config_options

PR #20800, opened 2023-10-28
iconfontcppheaders: create package

technoyes:
Created a IconFontCppHeaders package from https://github.com/juliettef/IconFontCppHeaders. This header only library allows easy use of a number of popular icon fonts.

I am not the original author IconFontCppHeaders. I just packaged it for Conan.

There are no released versions of the upstream library, so the Conan package is just `cci.20231026` from the current top commit of the upstream Git repo.

Conan package name is `iconfontcppheaders/cci.20231026`.

This  is my first Conan package. I have tried to follow the guidelines in the Conan documentation, but I have probably gotten some things wrong. If so, please direct me to the appropriate documentation or example packages where things are done "right" and I'll amend the PR as needed.

---

- [X] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6). _Using 2.0.13_
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated. _The Conan Center Hook does not work with v2 yet so unable to do this._


technoyes:
Fixed all errors found in the CI pipeline.

Ready for review.

czoido:
Hi @technoyes,
Thanks a lot for the contribution. In general, you can use the templates in the docs of the repo as an starting point for adding packages, like this one for header-only packages: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/header_only/all/conanfile.py

recipes/iconfontcppheaders/all/conanfile.py
```diff
@@ -0,0 +1,40 @@
...
+class FireHppConan(ConanFile):
+    name = "iconfontcppheaders"
+    description = "Headers for icon fonts Font Awesome, Fork Awesome, Google Material Design, Kenney game icons, Fontaudio, Codicons, Pictogrammers Material Design icons."
+    license = "Zlib"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/juliettef/IconFontCppHeaders"
+    topics = ("icons", "fonts", "icon-fonts", "header-only")
+
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
```
czoido, line 19:
```suggestion
    settings = "os", "arch", "compiler", "build_type"
    no_copy_source = True
```

recipes/iconfontcppheaders/all/conanfile.py
```diff
@@ -0,0 +1,40 @@
...
+    generators = "CMakeDeps", "CMakeToolchain"
+
+    def layout(self):
+        cmake_layout(self, src_folder='src')
```
czoido, line 23:
```suggestion
    def layout(self):
        basic_layout(self, src_folder="src")
```

recipes/iconfontcppheaders/all/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+cmake_minimum_required(VERSION 3.1.2)
```
czoido, line 1: no need to add this file here

recipes/iconfontcppheaders/all/conanfile.py
```diff
@@ -0,0 +1,40 @@
...
+    def package_id(self):
+        self.info.clear()
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def package(self):
+        copy(self, "licence.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        copy(self, "*.h", self.source_folder, os.path.join(self.package_folder, "include", "IconFontCppHeaders"))
+
```
czoido, line 34:
```suggestion
        copy(self, "*.h", self.source_folder, os.path.join(self.package_folder, "include"))

```
Looks like in general this headers are included like: 
```
#include <IconsFontAwesome6.h>
```
and not like
```
#include <IconFontCppHeaders/IconsFontAwesome6.h>
```
Is that right?

recipes/iconfontcppheaders/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,8 @@
+#include <IconFontCppHeaders/IconsFontAwesome6.h>
```
czoido, line 1:
```suggestion
#include <IconFontCppHeaders/IconsFontAwesome6.h>
```
Related with the comment I made about `copy(self, "*.h", self.source_folder, os.path.join(self.package_folder, "include", "IconFontCppHeaders"))` but I'm not completely sure about that

recipes/iconfontcppheaders/all/conanfile.py
```diff
@@ -0,0 +1,40 @@
...
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", self.recipe_folder, self.export_sources_folder)
```
czoido, line 36:
```suggestion
```

technoyes:
Thank you for the detailed feedback @czoido ! I did start with the header-only template, but I think I wandered off the wrong path. I thought I had to add a CMakeList.txt in order for the package (and tests) to be buildable from a CMake consumer, so I switched from basic_layout to cmake_layout. That in turn made me think I was "patching" and thus had to remove no_copy_source = True...

I think putting the include files under sub-directory looks nicer - but I think you are right that we shouldn't since upstream does not place them in any sub-directory. Better to keep aligned with upstream than invent something new.

In short, great feedback! I applied it all and tested with my project locally again.

Thanks for re-merging master just now to make it build again!

czoido:
> Thank you for the detailed feedback @czoido ! I did start with the header-only template, but I think I wandered off the wrong path. I thought I had to add a CMakeLists.txt in order for the package (and tests) to be buildable from a CMake consumer, so I switched from basic_layout to cmake_layout. That in turn made me think I was "patching" and thus had to remove no_copy_source = True...
> 
> I think putting the include files under sub-directory looks nicer - but I think you are right that we shouldn't since upstream does not place them in any sub-directory. Better to keep aligned with upstream than invent something new.
> 
> In short, great feedback! I applied it all and tested with my project locally again.
> 
> Thanks for re-merging master just now to make it build again!

Thanks to you! For the reference, there's a small section in the conan docs too about this topic: https://docs.conan.io/1/howtos/header_only.html

czoido (approved):

jwillikers (approved):

technoyes:
Thank you for reviewing @czoido  and @jwillikers. One less "vendored in" dependency to worry about.

I really like how using Conan simplifies dependency management. A worthy cousin to pip/yarn in py/js worlds for sure! A bit more complex to use due to the inherent complexity of C++, but not too bad once you get the hang of it.

PR #20284, opened 2023-10-02
cpptrace: Add recipe

jeremy-rifkin:
Specify library name and version:  **cpptrace/0.2**

I'm the author of [cpptrace](https://github.com/jeremy-rifkin/cpptrace), which is a dependency for [libassert](https://github.com/jeremy-rifkin/libassert) v1.2.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20844, opened 2023-10-30
libassert: add 1.2.1

jeremy-rifkin:
Specify library name and version:  **libassert/1.2.1**

I'm the author of libassert. Now that #20284 has been merged (and huge thanks to all the help!) libassert 1.2.1 can be added.

~~Apologies for the large PR. Because libassert 1.2.1 builds very differently, mainly because stack traces are now handled with cpptrace, I think it makes sense to separate the recipes for pre-cpptrace / post-cpptrace. Happy to change this if it's not the preferred way to go about this.~~

---


czoido:
Hi @jeremy-rifkin,
Thanks a lot for the contribution. I was thinking that maybe using the version of the library in the recipe, instead of separating recipes in different folders, would be cleaner. That's something more usual in conan recipes that separating into folders by version (that usually happens for different major versions and if the diff between conanfiles is really huge)

jeremy-rifkin:
Hi @czoido thanks so much for taking a look. I'd prefer to not split into two different recipes if possible, just because it's the same library. Looking at the two conan files now, I think the diff isn't too big in the end so I can probably just merge everything back into an `all/` folder.

jeremy-rifkin:
Libassert didn't previously list libm in the system libraries. I can add that if needed. Otherwise should be good to go for a review.

SSE4 (approved):

recipes/libassert/all/conanfile.py
```diff
@@ -92,6 +105,26 @@ def package(self):
...
+        if Version(self.version) >= Version("1.2.1"):
+            # For some reason after conan installs the include is #include <assert/assert/assert.hpp> which is not what
+            # should happen
+            copy(
+                self,
+                "assert.hpp",
+                src=os.path.join(self.package_folder, "include/assert/assert/"),
+                dst=os.path.join(self.package_folder, "include/assert/"),
+                keep_path=False
+            )
```
czoido, line 118: this could be avoided doing:

```
self.cpp_info.includedirs = [os.path.join("include", "assert")]
```
in the package_info

recipes/libassert/all/conanfile.py
```diff
@@ -73,11 +75,22 @@ def export_sources(self):
...
+        else:
+            tc = CMakeToolchain(self)
+            if is_msvc(self):
+                tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+            if not self.options.shared:
+                tc.variables["ASSERT_STATIC"] = True
+            tc.variables["ASSERT_USE_EXTERNAL_CPPTRACE"] = True
+            tc.generate()
+            tc = CMakeDeps(self)
+            tc.generate()
```
czoido, line 93: Maybe we can simplify this a bit?
```suggestion
    def generate(self):
        tc = CMakeToolchain(self)

        if is_msvc(self):
            tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)

        if Version(self.version) >= Version("1.2.1"):
            if not self.options.shared:
                tc.variables["ASSERT_STATIC"] = True
            tc.variables["ASSERT_USE_EXTERNAL_CPPTRACE"] = True
            deps = CMakeDeps(self)
            deps.generate()

        tc.generate()
```

jeremy-rifkin:
Thank you so much for taking a look, pushed the changes now

jeremy-rifkin:
Hi, is anything else needed from me to get this merged?

czoido:
> Hi, is anything else needed from me to get this merged?

Hi @jeremy-rifkin, sorry for the delay. I think there are a couple of things we can improve, let me add some comments and I'll push the changes.

recipes/libassert/all/conanfile.py
```diff
@@ -92,6 +102,17 @@ def package(self):
              src=self.source_folder)
         cmake = CMake(self)
         cmake.install()
+        
+        if Version(self.version) >= Version("1.2.1"):
```
czoido, line 106: I think we could do this unconditionally, looks like the test_package also to avoid moving the dll's in the test_package

recipes/libassert/all/conanfile.py
```diff
@@ -105,14 +126,21 @@ def package_info(self):
 
         self.cpp_info.set_property("cmake_file_name", "assert")
         self.cpp_info.set_property("cmake_target_name", "assert::assert")
+        
+        # For some reason after conan installs the include is #include <assert/assert/assert.hpp> which is not what
+        # should happen
+        if Version(self.version) >= Version("1.2.1"):
+            self.cpp_info.includedirs = [os.path.join("include", "assert")]
```
czoido, line 133: Maybe better we append this folder unconditionally and we will not break consumers of the previous versions of the package. Note that this also could be managed patching the original CMakeLists.txt
```suggestion
        self.cpp_info.includedirs.append(os.path.join("include", "assert"))
```

recipes/libassert/all/test_package/CMakeLists.txt
```diff
@@ -9,3 +9,8 @@ add_executable(${PROJECT_NAME} test_package.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE assert::assert)
 # In case the target project need a specific C++ standard
 target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
+
+option(CONAN_ASSERT_ASSERT_ASSERT "Controls old include vs new include" OFF)
+if(CONAN_ASSERT_ASSERT_ASSERT)
+  target_compile_definitions(${PROJECT_NAME} PRIVATE CONAN_ASSERT_ASSERT_ASSERT)
+endif()
```
czoido, line 16:
```suggestion
```
This can be removed if we append the folder in the recipe

recipes/libassert/all/test_package/conanfile.py
```diff
@@ -17,12 +18,18 @@ def layout(self):
         cmake_layout(self)
 
     def generate(self):
-        for dep in self.dependencies.values():
-            copy(self, "*.dll", dep.cpp_info.libdirs[0], self.build_folder)
+        require_version = re.split('[@#]', self.tested_reference_str)[0].split("/", 1)[1]
+        if Version(require_version) < Version("1.2.1"):
+            for dep in self.dependencies.values():
+                copy(self, "*.dll", dep.cpp_info.libdirs[0], self.build_folder)
```
czoido, line 24: we can get rid of this if we copy unconditionally the dlls in the recipe

recipes/libassert/all/test_v1_package/conanfile.py
```diff
@@ -13,9 +12,6 @@ def build(self):
         cmake.configure()
         cmake.build()
 
-    def imports(self):
```
czoido, line 16: removing this as this should be handled by the run_environment environment in Conan 1.X with legacy generators

recipes/libassert/all/conanfile.py
jeremy-rifkin: Thanks! I had been worried about breaking anything but this seems like a good way to go.

jeremy-rifkin:
Thanks! The changes look really good.

czoido (approved):

jwillikers (approved):

jeremy-rifkin:
Thanks all!

PR #20850, opened 2023-10-31
openexr: Add some missing #include <cstdint> definitions for newer gcc

irieger:
Specify library name and version:  **openexr/2.x**

Newer gcc versions like 13.2.1 I use on my machine fail when cstdint isn't included properly which previous versions ignored. This fixes the build of all openexr/2.* packages which previously failed me and made me assume openexr/2.* isn't update for conan 2.

---


irieger:
Tried to fix the warnings from CI in https://github.com/conan-io/conan-center-index/pull/20850/commits/1de0bf166b36a067c76f795338c6a2235b11783c but that does lead to failed package builds.

uilianries (changes requested):
@irieger Thank you for your fix, but as far as I see it's an upstream problem, not a recipe limitation. Please, open a PR directly to the upstream first, otherwise, we will be new maintainers of theses patches and will have no solution for next versions too.

irieger:
> @irieger Thank you for your fix, but as far as I see it's an upstream problem, not a recipe limitation. Please, open a PR directly to the upstream first, otherwise, we will be new maintainers of theses patches and will have no solution for next versions too.

Upstream has this fixed. Just added the latest version of the v2 series (2.5.9) which doesn't require this patch anymore on my side.

uilianries:
> Upstream has this fixed. Just added the latest version of the v2 series (2.5.9) which doesn't require this patch anymore on my side.

@irieger Could you please point where? PR or commit? Plus, could update the conandata.yml patch_source and the category to portability.

Patch fields are documented here: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conandata_yml_format.md#patches-fields

Adding patch_source will help us to track that change.

irieger:
> @irieger Could you please point where? PR or commit? Plus, could update the conandata.yml patch_source and the category to portability.
This is the commit in the Release branch for 2.5 legacy releases: https://github.com/AcademySoftwareFoundation/openexr/commit/310ae8600

Added this and a change to type portability in the latest commit.

irieger:
@uilianries I did the requested changes regarding the patch and everything builds. Only thing is that github-actions reports warnings regarding missing system libs even for older versions that are untouched. Not sure if that can be ignored but I had a commit trying to fix it that broke the builds.

> post_package(): WARN: [MISSING SYSTEM LIBS (KB-H043)] Library './lib/libIlmImf.so' links to system library 'm' but it is not in cpp_info.system_libs.
And so on

uilianries:
>  I did the requested changes regarding the patch and everything builds. Only thing is that github-actions reports warnings regarding missing system libs even for older versions that are untouched. Not sure if that can be ignored but I had a commit trying to fix it that broke the builds.

@irieger It can be ignored for now, no problem.

uilianries (approved):
LGTM

jwillikers (approved):

PR #20851, opened 2023-10-31
freealut: add recipe

lucaskdc:
Specify library name and version:  **freealut/1.1.0**

Since many applications that use openal used alut together, and currently the conan center index contains only the recipe for the OpenAl Soft, thir PR adds the recipe for freealut.

---


recipes/freealut/all/conanfile.py
```diff
@@ -0,0 +1,58 @@
...
+    }
+    default_options = {
+        "shared": True
+    }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def configure(self):
+        self.options["openal-soft"].shared = self.options.shared
```
valgur, line 28: Add a check for this in `validate()` as well. Options set in `configure()` can be overriden by the user.

recipes/freealut/all/conanfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("openal-soft/[>=1.1]", visible=True, transitive_headers=True, transitive_libs=True)
```
valgur, line 34: CCI currently allows version ranges to be used for only a small subset of packages. You will probably have to replace this with a specific version and a comment instead.
`visible=True` is the default and can be dropped.

recipes/freealut/all/conanfile.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def package(self):
+        copy(self, "COPYING", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rm(self, "freealut-config", os.path.join(self.package_folder, "bin"))
+
+    def package_info(self):
+        self.cpp_info.libs = collect_libs(self)
```
valgur, line 57:
```suggestion
        self.cpp_info.libs = ["alut"]
```

recipes/freealut/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.1)
```
valgur, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15)
```
Conan v2 requires CMake 3.15+.

recipes/freealut/all/test_v1_package/conanfile.py
valgur, line 1: `test_v1_package` is no longer required or recommended for new recipes.

recipes/freealut/all/conanfile.py
```diff
@@ -0,0 +1,58 @@
...
+    homepage = "https://directory.fsf.org/wiki/Freealut"
+    license = "LGPL-2.0-or-later"
+
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True]
+    }
+    default_options = {
+        "shared": True
+    }
```
valgur, line 22:
```suggestion
    package_type = "shared-library"
    settings = "os", "arch", "compiler", "build_type"
```

valgur (changes requested):
Looks good overall but should definitely set the `package_type = "shared-library"` and remove the `shared` option.

luizfeldmann:
> Looks good overall but should definitely set the `package_type = "shared-library"` and remove the `shared` option.

We need the static build for vxWorks

recipes/freealut/all/conanfile.py
lucaskdc: Done

recipes/freealut/all/test_v1_package/conanfile.py
lucaskdc: But the v2 pipeline only runs if the v1 is succeed. What to do then?

recipes/freealut/all/conanfile.py
lucaskdc: Since the most recipes build as static by default (including the openal-soft), i did like the grpc recipe. Allow it to build static, but if the freealut is built as shared lib it forces to use shared requirements to avoid binary duplication.

recipes/freealut/all/conanfile.py
lucaskdc: Done

lucaskdc:
> Looks good overall but should definitely set the `package_type = "shared-library"` and remove the `shared` option.

Since the most recipes build as static by default (including the openal-soft), i changed it to do like the grpc recipe. It allows to build as static, but if the freealut is built as shared lib it forces to use requirements as shared too to avoid binary duplication.

lucaskdc:
> > Looks good overall but should definitely set the `package_type = "shared-library"` and remove the `shared` option.
> 
> We need the static build for vxWorks

probably openal-soft won't build for vxworks, but i left the static option there just in case someone uses it linux.

recipes/freealut/all/conanfile.py
lucaskdc: Done

lucaskdc:
@valgur thanks very much for your comments.
I guess this PR is worth reviewing again now.

recipes/freealut/all/conandata.yml
```diff
@@ -0,0 +1,9 @@
+sources:
+  "1.1.0":
+    url: "http://ftp.debian.org/debian/pool/main/f/freealut/freealut_1.1.0.orig.tar.gz"
+    sha256: "60d1ea8779471bb851b89b49ce44eecb78e46265be1a6e9320a28b100c8df44f"
+patches:
+  "1.1.0":
+    - patch_file: "patches/freealut-1.1.0-conan-build.patch"
+      patch_description: "Change CMakeLists to get OpenAL from conan and allow multiplatform build as static lib"
```
valgur, line 8: Not mandatory by any means, but I would split the CMakeLists.txt and the header modifications into two separate patches. Makes it easier to re-use the patches in case of a version bump.

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
```diff
@@ -0,0 +1,184 @@
...
+-INCLUDE_DIRECTORIES(${Alut_SOURCE_DIR}/include)
++INCLUDE_DIRECTORIES(${CMAKE_SOURCE_DIR}/include)
+ 
+ SET(ADD_WFLAGS "")
+ SET(ADD_CFLAGS "")
+ SET(ADD_LDFLAGS "")
+ SET(ADD_LIBS "")
+ 
+-OPTION(BUILD_STATIC    "build static library too"   ON)
++option(BUILD_SHARED_LIBS "Build using shared libraries" ON)
```
valgur, line 48: Preferably, don't add this as an option (which converts it into a cache variable). That way you don't need to explicitly set it as a `tc.cache_variable`. CMakeToolchain sets it by default, but as a plain `tc.variable`.

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
```diff
@@ -0,0 +1,184 @@
...
+-ENDIF(BUILD_STATIC)
+-
+-
+-ADD_LIBRARY(alut SHARED ${ALUT_SOURCES})
++ADD_LIBRARY(alut ${ALUT_SOURCES})
+ SET_TARGET_PROPERTIES(alut PROPERTIES VERSION ${VERSION} SOVERSION ${MAJOR_VERSION})
++set_target_properties(alut PROPERTIES PUBLIC_HEADER "include/AL/alut.h")
++if (NOT BUILD_SHARED_LIBS)
++  target_compile_definitions(alut PUBLIC ALUT_LIBTYPE_STATIC)
++endif()
```
valgur, line 107: Not needed. `add_library()` uses the value of `BUILD_SHARED_LIBS` automatically to set `STATIC`/`SHARED`.

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
```diff
@@ -0,0 +1,184 @@
...
+@@ -170,8 +146,10 @@ CONFIGURE_FILE(${CMAKE_SOURCE_DIR}/admin/pkgconfig/freealut.pc.in
+                ${CMAKE_BINARY_DIR}/admin/pkgconfig/freealut.pc @ONLY)
+ CONFIGURE_FILE(${CMAKE_SOURCE_DIR}/admin/pkgconfig/freealut-config.in
+                ${CMAKE_BINARY_DIR}/admin/pkgconfig/freealut-config @ONLY)
+-INSTALL_PROGRAMS(/bin FILES admin/pkgconfig/freealut-config)
+-INSTALL_FILES(/lib/pkgconfig FILES admin/pkgconfig/freealut.pc)
++install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/admin/pkgconfig/freealut-config 
++    DESTINATION ${CMAKE_INSTALL_BINDIR})
++install(FILES ${CMAKE_CURRENT_BINARY_DIR}/admin/pkgconfig/freealut.pc
++    DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
```
valgur, line 129: You can drop the pkgconfig installation as that needs to be removed later in the recipe anyway.

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
```diff
@@ -0,0 +1,184 @@
...
+ 
+     FOREACH(TEST ${TESTS})
+       SET(TEST_SRC test_suite/test_${TEST}.c)
+-      SET_SOURCE_FILES_PROPERTIES(${TEST_SRC} PROPERTIES COMPILE_FLAGS "-Wno-deprecated-declarations")
++      if(CMAKE_COMPILER_IS_GNUCC)
++          SET_SOURCE_FILES_PROPERTIES(${TEST_SRC} PROPERTIES COMPILE_FLAGS "-Wno-deprecated-declarations")
++      endif()
+       ADD_EXECUTABLE(test_${TEST} ${TEST_SRC})
+       TARGET_LINK_LIBRARIES(test_${TEST} ${OPENAL_LIB} ${ADD_LIBS} alut)
+     ENDFOREACH(TEST)
```
valgur, line 143:
```suggestion
```
The recipe is not running tests anyway.

valgur (approved):

luizfeldmann (approved):

recipes/freealut/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, copy, get, rmdir, rm
+from conan.errors import ConanInvalidConfiguration
+import os
+
+required_conan_version = ">=2.0"
```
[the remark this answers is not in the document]
lucaskdc: The last commit is >=1.61.0

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
lucaskdc: this define is being used in the alut.h to remove the dllexport/dllimport

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
lucaskdc: done

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
lucaskdc: Since this PR: https://github.com/conan-io/conan/pull/12401 the conan toolchain sets it as a cache_variable, so it might be usefull if someone wants to build from source with option shared=false and then with option shared=true.

recipes/freealut/all/conandata.yml
lucaskdc: splitted

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
lucaskdc: dropped

recipes/freealut/all/patches/freealut-1.1.0-conan-build.patch
lucaskdc: still, no BUILD_SHARED_LIBS as an cmake option is not needed, as the toolchain sets it

uilianries (commented):
Saudações @lucaskdc e obrigado pelo seu PR!

I sent you a PR as my code review, due several changes needed:

https://github.com/lucaskdc/conan-center-index/pull/1

We try to avoid as much as possible adding patches to our packages, because as soon they are merged, we become the maintainers, more info: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#policy-on-patches

I only tested that PR in Linux, but in case you are having trouble in Windows, to generate shared libraries, I would suggest opening an issue to the upstream instead. As gambiarra, you can use `WINDOWS_EXPORT_ALL_SYMBOLS`, e.g.

```python
def generate(self):
    tc = CMakeToolchain(self)
    tc.variables['WINDOWS_EXPORT_ALL_SYMBOLS'] = self.options.shared
    ...
```

Regards.

lucaskdc:
> Saudações @lucaskdc e obrigado pelo seu PR!
> 
> I sent you a PR as my code review, due several changes needed:
> 
> [lucaskdc#1](https://github.com/lucaskdc/conan-center-index/pull/1)
> 
> We try to avoid as much as possible adding patches to our packages, because as soon they are merged, we become the maintainers, more info: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#policy-on-patches
> 
> I only tested that PR in Linux, but in case you are having trouble in Windows, to generate shared libraries, I would suggest opening an issue to the upstream instead. As gambiarra, you can use `WINDOWS_EXPORT_ALL_SYMBOLS`, e.g.
> 
> ```python
> def generate(self):
>     tc = CMakeToolchain(self)
>     tc.variables['WINDOWS_EXPORT_ALL_SYMBOLS'] = self.options.shared
>     ...
> ```
> 
> https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html
> 
> Regards.

@uilianries Agradeço vossa bela contribuição.
Regarding the build for Windows, I had to change a few lines to allow it build as a shared lib for Windows. It seems too much effort to keep that cmakelists file and make it build as a static lib for Windows due linking issues of OpenAl. Also, there is no official repository of freealut anymore. There are a few forks around there, like (https://github.com/vancegroup/freealut) but they aren't making new releases.

I merged your contributions and added a few lines to fix the build for Windows as a shared lib, and added a validation condition for block build as static lib for Windows.

I hope it will be enough since the old Alut SDK distributed only as a shared lib. Also, it's harder to make proprietary software be LGPL compatible linking the LGPL lib as a static one, in this way the most common use case it use it as a shared lib in Windows.

lucaskdc:
I removed support for static lib in macOS since it's taking me much time and I don't see any demand to use as static lib in macOS. Without a lot of cmake patches I don't see a clean way to fix that.

uilianries (commented):
I'm able to build mixed shared-static library using freealut, and it's a common rule for ConanCenterIndex, unless there is a real restriction, like in CMakeLists.txt there is a `SHARED IMPORTED` declared, which is not the case.

For Conan 2.x CI builds, we run all shared and all static, only Conan 1.x mixes those libraries.

I understand the restriction over LGPL usage, but is something not managed by Conan directly, the user should be free to configure it instead.

luizfeldmann:
Hi @lucaskdc any progress here ?

danimtb (approved):

luizfeldmann:
@valgur @franramirez688 @uilianries
We request your review please.

jwillikers (approved):

valgur (approved):

PR #20856, opened 2023-10-31
Minor clean up of template packages

jwillikers:
Alphabetize things.
Use `_min_cppstd` and `_compilers_minimum_version` consistently.
Omit keywords from the `copy` command and use the same order of the arguments across all calls.
Use the `LANGUAGES` keyword in CMake project commands.
Bump the Meson version in the Meson test package.

---


docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -41,6 +42,21 @@ class PackageConan(ConanFile):
         "with_foobar": True,
     }
 
+    @property
+    def _min_cppstd(self):
+        return 11
```
uilianries, line 47:
```suggestion
        return 14
```
I guess these compilers are C++14, not C++11

jwillikers: Nice! I've updated the other usages to be consistent.

SSE4 (approved):

uilianries (approved):
LGTM

PR #20866, opened 2023-10-31
fastgltf: Update to 0.6.1

spnda:
Specify library name and version:  **fastgltf/0.6.0**

Updates fastgltf to 0.6.0. Also I've seen spnda/fastgltf#22. Would adding an option like `FASTGLTF_FIND_SIMDJSON` be appropriate, which if set calls `find_package` so that these patches can be avoided? That would be for the next release, though I think.

### DIFF
--- a/recipes/fastgltf/all/conanfile.py
+++ b/recipes/fastgltf/all/conanfile.py
@@ -55,6 +55,10 @@ class fastgltf(ConanFile):
         if self.settings.os == "Windows":
             del self.options.fPIC
 
+        if Version(self.version) <= "0.6.0":
+            del self.options.disable_custom_memory_pool
+            del self.options.use_64bit_float
+
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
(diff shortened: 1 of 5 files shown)


spnda:
I don't understand why I'm getting errors that `std::pmr::string` can't be found on the Clang ones that seem to be using libstdc++ 10 (the headers with std::string are in the 10.3 directory in the log). I'm compiling as C++17 and GCC 10.3 should support this, as [this commit](https://github.com/gcc-mirror/gcc/commit/1fc9d0b0e4657b29ff140fa03dca1711aa5a1198) introducing the `pmr` namespace aliases was merged into GCC 9.1.

The [tests using GCC 11](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/20866/3-linux-gcc/fastgltf/0.6.1//summary.json) all succeed. So it's some issue with the older version of libstdc++. I run tests on CI myself with GCC 9 and 10 which all succeed, though I think those are running on libstdc++ 13. So, what can I do about this?

recipes/fastgltf/all/conanfile.py
```diff
@@ -23,11 +23,15 @@ class fastgltf(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "enable_small_vector": [True, False],
+        "disable_custom_memory_pool": [True, False],
+        "use_64bit_float": [True, False],
```
uilianries, line 27: There options are not available for 0.4 and 0.5:

- https://github.com/spnda/fastgltf/blob/v0.6.0/CMakeLists.txt#L16
- https://github.com/spnda/fastgltf/blob/v0.5.0/CMakeLists.txt#L15
- https://github.com/spnda/fastgltf/blob/v0.4.0/CMakeLists.txt#L15

Please, remove them in the `config_options` like:

```
if Version(self.version) <= "0.6.0":
    del self.options.disable_custom_memory_pool
    del self.options.use_64bit_float
```

Then, you will need to get those options using safe method:

```
if self.options.get_safe("disable_custom_memory_pool"):
    tc.variables["FASTGLTF_DISABLE_CUSTOM_MEMORY_POOL"] = True
```

uilianries (changes requested):
Thank for your contribution. Those new options, are only available since 0.6.0, so they are not affect the package ID for older versions.

uilianries (approved):
LGTM

franramirez688 (approved):

PR #19298, opened 2023-08-19
gdal: migrate to Conan v2

valgur:
- Resolves #20843
- Resolves #19041
- Resolves #17984
- Resolves #17830
- Resolves #16935
- Resolves #16479
- Resolves #15660
- Resolves #14166
- Resolves #13801

PR #20872, opened 2023-11-01
gdal/pre_3.5.0: clean up build configuration, bump deps

valgur:
Split off the changes to the older recipe from #19298 to keep the PR size more manageable.


uilianries:
I just started an internal build to generate the missing libgeotiff/1.7.1

SpaceIm:
What's the rational for these changes?

SpaceIm:
I also want to mention that you won't be able to pass msvc v2 pipeline due to https://github.com/conan-io/conan/issues/13603 (it comes from sqlite3).

SpaceIm:
Please don't refactor configure block of autotools, I've written this recipe and I've tested many many combinations, far more than conancenter, the risk of breaking is too high, and I don't see any interest for this refactoring, it's just cosmetic.

valgur:
@SpaceIm Thanks a lot for filling me in on the details about the quirks in the NMake configuration. It looked quite incorrect / broken in places and it was not clear if it was intentional or not. A short comment in the recipe would have been very helpful.

I don't quite agree the reduction boilerplate not being worth it, at least for the Autotools part. It makes it easier to follow and verify thanks to being more predictable and discarding irrelevant details. But yeah, it's not a good idea to do it after the fact.

The pre-3.5.0 recipe is more or less deprecated after merging #19298 soon-ish, so this PR is not worth for that reason alone. Closing.

Issue #1789, opened 2020-05-31
[question] Require multiple libraries from the same recipe with different options (fftw)


gmgunter:
The FFTW library can be configured to support FFTs with [single, double, or long double precision](http://www.fftw.org/fftw3_doc/Precision.html#Precision), generating a separate library object for each precision. Applications that wish to support FFTs with multiple floating-point precisions must link to multiple FFTW libraries. 

The current [**fftw/3.3.8** recipe allows specifying `precision` as an option](https://github.com/conan-io/conan-center-index/blob/c9c9a1d28aebe54ea207cae2b23becafde713b16/recipes/fftw/all/conanfile.py#L17) but doesn't supply an option to choose multiple precisions. 

In my `conanfile.txt` file, I want to require fftw and specify multiple values for the `precision` option, e.g.:
```
[requires]
fftw/3.3.8

[generators]
cmake

[options]
fftw:precision=float,double
```
Is this possible? If so, would it generate a single cmake target with both libraries or two different cmake targets?  If not, is there some other way to achieve this with conan?

jgsogo:
Hi, @gmgunter!

Quite interesting use-case! Indeed, it says _"You can link to the different-precision libraries simultaneously."_, so these shouldn't be mutually exclusive options.

Conan doesn't support two instances with a different configurations of the same package in the same context. I mean, what you are suggesting is not possible and it is something very core to Conan so it is unlikely that we can change it in the client.

This should be fixed in the recipe, these options should be provided as flags (names can be improved)

```python 
options = {
    'precision_double': [True, False],
    'precision_single': [True, False],
    'precision_longdouble': [True, False],
}
```

we should probably activate all of them in ConanCenter and then, we the help of components the consumer will be able to choose which ones to use/link.

Will you submit the PR?

Thanks!

gmgunter:
I made an initial attempt at implementing the necessary changes. However, I ran into some frustrating roadblocks that I hadn't anticipated. I don't think I have the patience right now to navigate fftw's clunky build system and follow through with a PR, but I'll share what I've learned in case someone else is brave enough.

[...]

PR #20888, opened 2023-11-02
[fftw] support multiple precisions

SSE4:
Specify library name and version:  **fftw/all**

closes: #1789

---


jwillikers (approved):

recipes/fftw/all/conanfile.py
```diff
@@ -70,64 +87,93 @@ def generate(self):
...
         tc.generate()
 
+    @property
+    def build_folder(self):
+        bf = super().build_folder
+        return os.path.join(bf, self._current_precision) if self._current_precision else bf
+
+    @property
+    def _all_precisions(self):
+        return [p for p in ALL if self.options.precision == p or self.options.get_safe(f"precision_{p}")]
```
uilianries, line 103:
```suggestion
        return [p for p in ALL if self.options.get_safe(f"precision_{p}")]
```

recipes/fftw/all/conanfile.py
```diff
@@ -70,64 +87,93 @@ def generate(self):
...
+            component.names["cmake_find_package_multi"] = cmake_target_name
+            component.set_property("cmake_target_name", f"{cmake_namespace}::{cmake_target_name}")
+            component.set_property("pkg_config_name", pkgconfig_name)
+
+    def package_id(self):
+        if self.info.options.precision != 'deprecated':
+            self.info.options.precision_single = self.info.options.precision == SINGLE
+            self.info.options.precision_double = self.info.options.precision == DOUBLE
+            self.info.options.precision_longdouble = self.info.options.precision == LONGDOUBLE
+            self.info.options.precision_quad = self.info.options.precision == QUAD
```
uilianries, line 169:
```suggestion
```

recipes/fftw/all/conanfile.py
```diff
@@ -70,64 +87,93 @@ def generate(self):
...
-        cmake.configure()
-        cmake.build()
+        for self._current_precision in self._all_precisions:
+            cmake = CMake(self)
+            cli_args = [
+                f"-DENABLE_FLOAT={on_off(self._current_precision == SINGLE)}",
+                f"-DENABLE_LONG_DOUBLE={on_off(self._current_precision == LONGDOUBLE)}",
+                f"-DENABLE_QUAD_PRECISION={on_off(self._current_precision == QUAD)}"
+            ]
+            cmake.configure(cli_args=cli_args)
```
uilianries, line 117:
```suggestion
            variables = {
                "ENABLE_FLOAT": on_off(self._current_precision == SINGLE),
                "ENABLE_LONG_DOUBLE": on_off(self._current_precision == LONGDOUBLE),
                "EENABLE_QUAD_PRECISION": on_off(self._current_precision == QUAD)
            }
            cmake.configure(variables=variables)
```

You can let CMake helper do the parsing job instead.

https://docs.conan.io/2/reference/tools/cmake/cmake.html#conan.tools.cmake.cmake.CMake.configure

uilianries (commented):
@SSE4 Thank you your hotfix!

I would ask to deprecate de `precision` option in fact, like documented in https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#can-i-remove-an-option-from-a-recipe

The risk of breaking users is counted too, but documented in https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#how-to-protect-my-project-from-breaking-changes-in-recipes how to avoid it.

SSE4:
> I would ask to deprecate de `precision` option in fact, like documented in https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#can-i-remove-an-option-from-a-recipe

sorry, I don't understand. I did exactly like documented here - https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#can-i-remove-an-option-from-a-recipe

for instance:
```
options = {"foobar": [True, False, "deprecated"]}
default_options = {"foobar": "deprecated"}
```
you can see:
https://github.com/conan-io/conan-center-index/pull/20888/files#diff-a2da4addc10b5279e914aa8b033195377d13a181faa3a2b3e8e5920796697bd4R29
https://github.com/conan-io/conan-center-index/pull/20888/files#diff-a2da4addc10b5279e914aa8b033195377d13a181faa3a2b3e8e5920796697bd4R42

and
```
def configure(self):
    if self.options.foobar != "deprecated":
        self.output.warning("foobar option is deprecated, do not use anymore.")
```
you can see:
https://github.com/conan-io/conan-center-index/pull/20888/files#diff-a2da4addc10b5279e914aa8b033195377d13a181faa3a2b3e8e5920796697bd4R68-R69

and finally:
```
def package_id(self):
    del self.info.options.foobar
```
you can see:
https://github.com/conan-io/conan-center-index/pull/20888/files#diff-a2da4addc10b5279e914aa8b033195377d13a181faa3a2b3e8e5920796697bd4R170

so it's done literally line by line as described in documentation you linked. am I missing your point?

recipes/fftw/all/conanfile.py
SSE4: this will result in weird package IDs, as far as I understand.
the idea is specifying `precision = single` works the same as specifying solely `precision_single=True` and all other precision options to False, and so on. binaries produced in these cases should be identical, therefore, package IDs should be computed identically as well.

recipes/fftw/all/conanfile.py
SSE4: that will not produce the same effect - specifying `precision=deprecated` will produce package with no libraries at all (=useless package).

recipes/fftw/all/conanfile.py
uilianries: Keeping two different options doing the same thing does not sounds correct, they will affect the package ID, but result in the same. If we need to deprecate the precision option, than it should be removed from the package ID. It will not result in weird package IDs, but they will follow the new options. Still, people can use lockfiles to avoid breaking.

SSE4: > they will affect the package ID, but result in the same

right now (with my changes) package ID is the same in both situations

>  than it should be removed from the package ID

it's already removed from the package ID, see https://github.com/conan-io/conan-center-index/pull/20888/files#diff-a2da4addc10b5279e914aa8b033195377d13a181faa3a2b3e8e5920796697bd4R170

> It will not result in weird package IDs

with changes your suggests `precision = deprecated` acts as nothing build, means package with no binaries. that's useless package and bad package ID.

uilianries: My point is not using the same package ID, but keeping the recipe as simple as possible. Please, remove this piece that I suggested, it only increases complexity.

SSE4: I can try to remove, let's see, but package produced with `precision != deprecated` will be completely useless (I am not sure if it's okay for consumers).

jwillikers (approved):

uilianries (approved):
After talking this issue about accepting multiple precision at same time, we are inclined to accept this change.

It will require extra PRs (I'll work on this), updating other packages affected by fftw and generating those expected config cmake files.

SpaceIm:
Is there a risk of ODR violation?

uilianries:
@SpaceIm I don't think it will be a problem: http://www.fftw.org/fftw3_doc/Precision.html#Precision

SpaceIm:
This PR seems to break build of fftw with Ninja:

```
fftw/3.3.10: RUN: cmake -G "Ninja" -DCMAKE_TOOLCHAIN_FILE="/Users/spaceim/.conan2/p/b/fftw154fc10fc776c/b/build/Release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="/Users/spaceim/.conan2/p/b/fftw154fc10fc776c/p" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" -DENABLE_FLOAT="OFF" -DENABLE_LONG_DOUBLE="OFF" -DEENABLE_QUAD_PRECISION="OFF" "/Users/spaceim/.conan2/p/b/fftw154fc10fc776c/b/src"
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.


-- Using Conan toolchain: /Users/spaceim/.conan2/p/b/fftw154fc10fc776c/b/build/Release/generators/conan_toolchain.cmake
-- Conan toolchain: Setting BUILD_SHARED_LIBS = ON
-- The C compiler identification is AppleClang 15.0.0.15000040
-- The CXX compiler identification is AppleClang 15.0.0.15000040
CMake Error at /Users/spaceim/.conan2/p/b/fftw154fc10fc776c/b/build/Release/double/CMakeFiles/CMakeScratch/TryCompile-IJW3BF/CMakeLists.txt:5 (project):
  Running

   '/usr/local/bin/ninja' '--version'

  failed with:

   Subprocess killed


CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error at /Applications/CMake.app/Contents/share/cmake-3.27/Modules/CMakeDetermineCompilerABI.cmake:57 (try_compile):
  Failed to configure test project build system.
  /Applications/CMake.app/Contents/share/cmake-3.27/Modules/CMakeTestCCompiler.cmake:26 (CMAKE_DETERMINE_COMPILER_ABI)
  CMakeLists.txt:7 (project)


-- Configuring incomplete, errors occurred!

fftw/3.3.10: ERROR:
Package '83745d8b39ed8175513ab49661917229f7721e89' build failed
```

recipes/fftw/all/conanfile.py
```diff
@@ -70,64 +87,88 @@ def generate(self):
...
         apply_conandata_patches(self)
-        cmake = CMake(self)
-        cmake.configure()
-        cmake.build()
+        for self._current_precision in self._all_precisions:
+            cmake = CMake(self)
+            variables = {
+                "ENABLE_FLOAT": on_off(self._current_precision == SINGLE),
+                "ENABLE_LONG_DOUBLE": on_off(self._current_precision == LONGDOUBLE),
+                "EENABLE_QUAD_PRECISION": on_off(self._current_precision == QUAD)
```
SpaceIm, line 115: Typo here? CMake tells me that `EENABLE_QUAD_PRECISION` is not used. Shouldn't it be `ENABLE_QUAD_PRECISION`?

recipes/fftw/all/conanfile.py
```diff
@@ -70,64 +87,88 @@ def generate(self):
...
+            rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
 
     def package_info(self):
-        prec_suffix = self._prec_suffix[str(self.options.precision)]
-        cmake_config_name = "FFTW3" + prec_suffix
-        cmake_namespace = "FFTW3"
-        cmake_target_name = "fftw3" + prec_suffix
-        pkgconfig_name = "fftw3" + prec_suffix
-        lib_name = "fftw3" + prec_suffix
+        cmake_config_name = cmake_namespace = "FFTW3"
```
SpaceIm, line 129: It's a regression here, there should be a FIXME at least. Each lib has a different CMake config file, and it can't be modeled by conan with this change in fftw.

SpaceIm: Why not just create one recipe per fftw flavor?

uilianries: @SpaceIm Thank you for your thoughts! Indeed we discussed about the a proxy recipe option with separated recipes, but we opted for a single recipe and a next PR creating those separated CMake config files. You see, other package managers follow the same approach, is a bit messy, but should work.

SpaceIm: > and a next PR creating those separated CMake config files

How? Do you plan to address https://github.com/conan-io/conan/issues/9000?

uilianries: @SpaceIm That feature will require more time to be done. The PR in question is a workaround for this recipe.

SpaceIm: Like exposing a custom cmake config file which would include expected config file generated by cmakedeps?

uilianries: That's it!

SSE4: @uilianries can we add more examples to the docs on how to properly model CMake properties?

for example, while working on this recipe, I was unable to find any guidance on how to generate config files with different file names per targets.

recipes/fftw/all/conanfile.py
SSE4: yes, typo :(

PR #15556, opened 2023-01-30
[poc] Add lapack/3.10.1

samuel-emrys:
Specify library name and version:  **lapack/3.10.1**

Initial proof of concept lapack recipe implementation.

This is a draft to illustrate the requirement to model internal gcc libraries, as presented in #15128. This is a blocker for this recipe.

An additional blocker for this recipe is that it will fail to build when the user profile defines `CC`, `CXX` or `tools.build:compiler_executables`. This results in the following error:

```
lapack/3.10.1@user/testing: CMake command: cmake -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE="/home/user/.conan/data/lapack/3.10.1/user/testing/build/fd2230a7a63c79d2e75d3f38ce3ae3cd07a0705c/build/Release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PRE[5762/46341]
owlings/.conan/data/lapack/3.10.1/user/testing/package/fd2230a7a63c79d2e75d3f38ce3ae3cd07a0705c" -DCMAKE_GNUtoMS="OFF" -DBUILD_TESTING="OFF" -DBUILD_SHARED_LIBS="ON" -DLAPACKE="ON" -DCBLAS="ON" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "/home/user/.
conan/data/lapack/3.10.1/user/testing/build/fd2230a7a63c79d2e75d3f38ce3ae3cd07a0705c/."                                                                                                                                                                                              
-- Using Conan toolchain: /home/user/.conan/data/lapack/3.10.1/user/testing/build/fd2230a7a63c79d2e75d3f38ce3ae3cd07a0705c/build/Release/generators/conan_toolchain.cmake                                                                                                        
-- Conan toolchain: Setting BUILD_SHARED_LIBS = ON                                                                                                                                                                                                                                       
-- The Fortran compiler identification is GNU 12.2.0                                                                                                                                                                                                                                     
-- The C compiler identification is Clang 15.0.7                                                                                                                                                                                                                                         
-- Detecting Fortran compiler ABI info                                                                                                                                                                                                                                                   
-- Detecting Fortran compiler ABI info - done                                                                                                                                                                                                                                            
[...]

PR #20894, opened 2023-11-02
[openblas] Make build_lapack=True default for >= 0.3.21

### LINKED ISSUE #7361: [package] openblas/0.3.17: Should use option build_lapack=True by default
### Summary

openblas currently builds with `build_lapack=False`. [This comment](https://github.com/conan-io/conan-center-index/issues/4696#issuecomment-897002205) suggests that `gfortran` is now available in Conan Docker Tools (so presumably the CI), and so building LAPACK should now be possible. Previous discussion regarding the decision to set `build_lapack=False` as the default can be found [here](https://github.com/conan-io/conan-center-index/pull/705#discussion_r374627398), but essentially it comes down to the fact that it couldn't be built on the CI.

This will facilitate the creation of a useful armadillo package, as per https://github.com/conan-io/conan-center-index/pull/7334. Without LAPACK, armadillo can only operate using a subset of features, and CCI requires that the default configuration for a package uses only default configurations for its dependencies.

samuel-emrys:
Version 0.3.21 of openblas introduced the ability to build lapack from C. Make this default to enable CCI recipes to use lapack when depending on openblas.

Additionally, fix an inconsistency in version constraint. Review was provided in https://github.com/conan-io/conan-center-index/pull/19808 to constrain it to 0.3.24 as this was the latest version available in CCI. However, functionality to build lapack was added in 0.3.21, so it doesn't really make sense to exclude versions between 0.3.21 and 0.3.24 from building lapack should they be added at a later date.

Closes #7361

Specify library name and version:  **openblas/0.3.24**

---

### DIFF
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -46,12 +46,12 @@ class OpenblasConan(ConanFile):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+        if Version(self.version) >= "0.3.21":
+            self.options.build_lapack = True
 
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        if Version(self.version) >= "0.3.21":
-            self.options.build_lapack = True
 
     def validate(self):
         if hasattr(self, "settings_build") and cross_building(self, skip_x64_x86=True):
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -50,6 +50,8 @@ class OpenblasConan(ConanFile):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+        if Version(self.version) >= "0.3.21":
+            self.options.build_lapack = True
 
     def validate(self):
         if hasattr(self, "settings_build") and cross_building(self, skip_x64_x86=True):
@@ -87,7 +89,7 @@ class OpenblasConan(ConanFile):
         # This checks explicit user-specified fortran compiler
         if self.options.build_lapack:
             if not self._fortran_compiler:
-                if Version(self.version) < "0.3.24":
+                if Version(self.version) < "0.3.21":
                     self.output.warning(
                         "Building with LAPACK support requires a Fortran compiler.")
                 else:
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -171,7 +171,7 @@ endif()"""
             self.cpp_info.components["openblas_component"].system_libs.append("m")
             if self.options.use_thread:
                 self.cpp_info.components["openblas_component"].system_libs.append("pthread")
-            if self.options.build_lapack:
+            if self.options.build_lapack and self._fortran_compiler:
                 self.cpp_info.components["openblas_component"].system_libs.append("gfortran")
 
         self.output.info(
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -111,13 +111,9 @@ class OpenblasConan(ConanFile):
         # This is a workaround to add the libm dependency on linux,
         # which is required to successfully compile on older gcc versions.
         tc.cache_variables["ANDROID"] = self.settings.os in ["Linux", "Android"]
-
-        if self.settings.arch == "armv8":
-            # OpenBLAS fails to detect the appropriate architecture for armv8, so help it out
-            tc.variables["TARGET"] = "ARMV8"
         tc.generate()
 
-    def build(self):
+    def _patch_sources(self):
         if Version(self.version) < "0.3.21":
             if Version(self.version) >= "0.3.12":
                 search = """message(STATUS "No Fortran compiler found, can build only BLAS but not LAPACK")"""
@@ -142,6 +138,20 @@ endif()"""
                 search,
                 replace,
             )
+
+        if Version(self.version) < "0.3.24":
+            # OpenBLAS fails to detect the appropriate target architecture for armv8, so help it out.
+            # This will ensure ARM64 is set, which will set TARGET=ARMV8 appropriately.
+            # This is a portability hotfix of https://github.com/OpenMathLib/OpenBLAS/pull/4142
+            replace_in_file(
+                self,
+                os.path.join(self.source_folder, "cmake", "system_check.cmake"),
+                'elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")',
+                'elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*|armv8.*)")'
+            )
+
+    def build(self):
+        self._patch_sources()
         cmake = self._configure_cmake()
         cmake.build()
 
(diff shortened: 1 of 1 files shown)


samuel-emrys:
@uilianries any idea what's going on with CI here?

jwillikers (approved):

samuel-emrys:
> @samuel-emrys Could you please, show a link or document proving that support, just in case?

This PR only makes the lapack build default. Support for building lapack from C was [approved by yourself](https://github.com/conan-io/conan-center-index/pull/19808#pullrequestreview-1710109471) in #19808:

> Okay, it seems be following https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.21

From the v0.3.21 release notes:

> when no Fortran compiler is available, OpenBLAS builds will now automatically
build LAPACK from an f2c-converted copy of LAPACK 3.9.0 unless the NO_LAPACK option
is specified (more recent releases make too heavy use of Fortran90+ features to be easily convertible to C)

uilianries (changes requested):
I'm not confident accepting this PR now, because there is no fortran support when building this package in CCI:

https://c3i.jfrog.io/c3i/misc/logs/pr/20894/3-linux-gcc/openblas/0.3.24//729e2486182de25560f46a58cc644fb2e1eb8b2f-build.txt

The version 0.3.24 will build with lapack, but no fortran is available, so the generated package is not good, someone will open an issue complaining about this lack of support for sure.

I'll check with Conan team to include fortran to Linux docker images, at least, we could solve it partially.

samuel-emrys:
> I'm not confident accepting this PR now, because there is no fortran support when building this package in CCI:
> 
> https://c3i.jfrog.io/c3i/misc/logs/pr/20894/3-linux-gcc/openblas/0.3.24//729e2486182de25560f46a58cc644fb2e1eb8b2f-build.txt
> 
> The version 0.3.24 will build with lapack, but no fortran is available, so the generated package is not good, someone will open an issue complaining about this lack of support for sure.


@uilianries I'm not sure I follow what you mean. What are you drawing attention to in those logs? All that related to fortran that I could find was this:

> openblas/0.3.24: Building LAPACK without Fortran compiler

which is the point of this PR. We can remove any dependency on fortran by building lapack from C sources. I'm not sure why anybody would complain, as it's reasonably easy to trigger a fortran build by defining the `fortran` key in `compiler_executables` if desired as per https://github.com/conan-io/conan-center-index/blob/e35433047da05a6210a7daf2bd232210d431f70c/recipes/openblas/all/conanfile.py#L40-L44

> I'll check with Conan team to include fortran to Linux docker images, at least, we could solve it partially.

#19808 removed the dependence on fortran so that we can build lapack with a C++ only toolchain - this seems to be the most desirable solution to me - what am I missing? `-lgfortran` is only added when the user specifies a value for `compiler_executables={"fortran": "/path/to/gfortran"}`, which would not be default for openblas >= 0.3.21 since this is not included in the conan profiles.

Lack of lapack support on CCI due to lack of a fortran compiler has been an ongoing issue for years now (https://github.com/conan-io/conan-center-index/pull/5260, https://github.com/conan-io/conan-center-index/pull/15556). I would really rather solve this with the simplest solution we have which is to build it from C. It's not perfect - it version locks lapack to 3.9.0, but it gives CCI minimum viability uniformly across all operating systems. Other math packages only provide limited utility without lapack, such as armadillo.

If we can explore adding fortran compilers to conan CI infrastructure to add a proper lapack recipe or build it here, that would be good as well, but I strongly suggest we should go for the quick win here.

samuel-emrys:
@uilianries can you engage on this more/elaborate on your reasoning? it doesn't make a lot of sense to me and having lapack available is a critical requirement for math libraries in CCI.

cc @RubenRBS @memsharded

uilianries:
@samuel-emrys indeed is a problem also related to the CI. I'll take a look to provide fortran via Docker images, so it should be able to build anything. Some images are prepared already (conanio/gcc11-ubuntu16.04), but Clang images need and extra build (checking to use llvm flang). For legacy images will be more complicated, because some of them are using ubuntu-toolchain-r or llvm repository and fortran is not always available.

samuel-emrys:
Adding fortran support to conan ci's is a good idea and is something that should be progressed independently of this PR, but it does open a can of worms with respect to binary compatibility - conan doesn't currently have a way to describe the ABI or runtime requirements that building a package with ifort or gfortran implies.

Building the fortran components from sources that have been converted to C really is the best way forward here in my opinion, because none of those ABI considerations will factor in here - it becomes a plain C library.

This is what we want as default wherever possible until conan has the features to support polyglot considerations.

samuel-emrys:
@uilianries @RubenRBS what's the best way to have a conversation about this and or an alternate pathway to getting lapack/other fortran libs on cci? Here or slack? I don't understand the rationale for not accepting this - this seems to be the best solution for conan because it bypasses the fortran requirement entirely, and this has been an ongoing issue for **years** (https://github.com/conan-io/conan-center-index/pull/5260, https://github.com/conan-io/conan-center-index/pull/15556) - getting a resolution here would be good and would enable a number of scientific computing libraries to work unimpeded.

// cc @memsharded

samuel-emrys:
Just to be clear in case it was missed above, the reason I'm confused is that from the [release notes](https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.21):

> when no Fortran compiler is available, OpenBLAS builds will now automatically
build LAPACK from an f2c-converted copy of LAPACK 3.9.0 unless the NO_LAPACK option
is specified (more recent releases make too heavy use of Fortran90+ features to be easily convertible to C)

This means that the sources have been converted from fortran to c, so this change triggers a build from the ***C sources*** meaning that there's no dependency on fortran at all here - this is essentially a C library, which sidesteps the problem of using fortran at all. I would imagine is the easiest path forward for conan-center-index.

recipes/openblas/all/conanfile.py
```diff
@@ -46,6 +46,8 @@ def _fortran_compiler(self):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+        if Version(self.version) >= "0.3.21":
```
uilianries, line 49:
```suggestion
        # INFO: When no Fortran compiler is available, OpenBLAS builds LAPACK from an f2c-converted copy of LAPACK unless the NO_LAPACK option is specified
```

I would include a note, just to be clear about it.

samuel-emrys:
These failures are as a result of https://github.com/conan-io/conan-center-index/commit/d34e4a1870cd0790a3ee791936ab7fee7d24f0ed, which introduces a native Macos armv8 runner to the c3i build matrix. armv8 builds of openblas have to date, only been explicitly disallowed for cross builds, and so an "invalid" configuration has made it through here and is causing failures.

I've attempted to add armv8 compatibility in this PR (by hotfixing https://github.com/OpenMathLib/OpenBLAS/pull/4142
), but if this doesn't work, I'll exclude native armv8 builds from the valid build matrix and leave the improvement for a future PR.

This also relates to https://github.com/conan-io/conan-center-index/pull/21485

samuel-emrys:
I've reverted the changes to add armv8 compatibility as they weren't immediately working (and they're beyond the scope of this PR), and I've added an additional exclusion for `armv8` to the `validate()` method. This can be followed up in https://github.com/conan-io/conan-center-index/pull/21485 which aims to add this compatibility.

uilianries (approved):
LGTM

RubenRBS (approved):

PR #20897, opened 2023-11-02
onetbb: remove shared option + fix tbbmaloc & tbbproxy logic

SpaceIm:
- regarding removal of shared option: see https://github.com/conan-io/conan/issues/15040#issuecomment-1790753531
- for tbbmaloc & tbbproxy, well current logic was super weird. It comes from https://github.com/conan-io/conan-center-index/pull/17311, but it doesn't really make sense. These options didn't exist before 2021.5.0 or 2021.6.0, so they have to be removed in config_option, that's all.

---

### DIFF
--- a/recipes/onetbb/all/conanfile.py
+++ b/recipes/onetbb/all/conanfile.py
@@ -21,20 +21,15 @@ class OneTBBConan(ConanFile):
         " programs that take full advantage of multicore performance, that are portable, composable"
         " and have future-proof scalability.")
     topics = ("tbb", "threading", "parallelism", "tbbmalloc")
-    package_type = "library"
-
+    package_type = "shared-library"
     settings = "os", "arch", "compiler", "build_type"
     options = {
-        "shared": [True, False],
-        "fPIC": [True, False],
         "tbbmalloc": [True, False],
         "tbbproxy": [True, False],
         "tbbbind": [True, False],
         "interprocedural_optimization": [True, False],
     }
     default_options = {
-        "shared": True,
-        "fPIC": True,
         "tbbmalloc": True,
         "tbbproxy": True,
         "tbbbind": True,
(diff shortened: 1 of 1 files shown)


valgur (approved):
Thanks! I really hope this fixes the issue.

SpaceIm:
Hum, another option I don't like in this recipe: `interprocedural_optimization`. It shouldn't be an option. There shouldn't be option doing babysitting of compiler flags. And LTO must not be enabled by default (well in shared lib it's harmless but still).

SpaceIm:
@danimtb @uilianries @RubenRBS I'm not even sure it will fix all issues related to missing packages in v2 pipeline. Since onetbb has dependencies, if I remove shared option of onetbb, it will build onetbb flavor with static dependencies only? So if a downstream recipe has a shared option, it will break with shared=True since conancenter won't have the flavor of onetbb with hwloc shared=True, correct?

jwillikers (approved):

ilya-lavrenov (approved):

uilianries (changes requested):
Removing the shared option sounds good to me.

I'll block to discuss first with Conan team, in case of side-effects.

We still could list all affected packages and run an internal CI build to regenerate after merging it, but still need to check what could happen with CI composition, like `*/*:shared=True/False` and even mixed (Conan 1.x) for future PRs.

uilianries:
After discussing with Conan team about this situation, indeed the `hwloc` will cause trouble due Conan 2.x CCI build configuration. So our better shot now is restricting `hwloc` to a shared library only.

I opened the PR https://github.com/conan-io/conan-center-index/pull/20976/. That PR should be merged first, then we can update this one and merge it too. 

Thank you @SpaceIm for not only opening the PR, but also sharing your investigation.

fschoenm:
@uilianries Can this be merged now or is there something left to do?

SpaceIm:
I've triggered a new build, now that https://github.com/conan-io/conan-center-index/pull/20976 has been merged.

uilianries (approved):

danimtb (approved):

PR #20903, opened 2023-11-02
cthash: relax compiler version check

SpaceIm:
Most folks won't define something like compiler.version=12.2, but compiler.version=12, so relax version check.

Do not define msvc min version since it raises anyway for msvc, it's misleading.

---


jwillikers (approved):

uilianries (approved):
LGTM

PR #20904, opened 2023-11-02
[highfive] Bump version to 2.8.0

jjcasmar:
Specify library name and version:  **highfive/2.8.0**

---

- [X ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [X ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [X ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

### DIFF
--- a/recipes/highfive/all/conanfile.py
+++ b/recipes/highfive/all/conanfile.py
@@ -22,12 +22,14 @@ class HighFiveConan(ConanFile):
         "with_eigen": [True, False],
         "with_xtensor": [True, False],
         "with_opencv": [True, False],
+        "with_static_hdf5": [True, False],
     }
     default_options = {
         "with_boost": True,
         "with_eigen": True,
         "with_xtensor": True,
         "with_opencv": False,
+        "with_static_hdf5": False,
     }
 
     def layout(self):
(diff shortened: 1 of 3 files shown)


franramirez688 (approved):

recipes/highfive/all/conanfile.py
```diff
@@ -65,6 +67,7 @@ def generate(self):
         tc.variables["HIGHFIVE_EXAMPLES"] = False
         tc.variables["HIGHFIVE_BUILD_DOCS"] = False
         tc.variables["HIGHFIVE_USE_INSTALL_DEPS"] = False
+        tc.variables["HIGHFIVE_STATIC_HDF5"] = self.options.with_static_hdf5
```
jwillikers, line 70: Shouldn't this be true if `hdf5` is built statically?

```suggestion
        tc.variables["HIGHFIVE_STATIC_HDF5"] = not self.dependencies["hdf5"].options.shared
```

jwillikers (changes requested):

recipes/highfive/all/conanfile.py
jjcasmar: I guess it depends on how we want to drive the static version of hdf5. According to 2.8.0, there exist an option in CMake to select to link to a static version of HDF5. I think this should be reflected as an options for the HighFive package. 

We should set `self.dependencies["hdf5"].options.shared = self.options.with_static_hdf5` though.

jwillikers: That seems a little backwards to me, as consumers probably expect that setting `hdf5/*:shared=False` to link statically.

jjcasmar: Fair enough. I will change it

SSE4 (approved):

PR #20905, opened 2023-11-02
freeglut: Add support for Wayland

jwillikers:
Specify library name and version:  **freeglut/***

Adds the `with_wayland` option to build with support for Wayland.
Both Wayland and X11 support cannot currently be enabled simultaneously for freeglut.
It's weird.

Includes some patches to improve how the dependencies are picked up to avoid pulling in dependencies that aren't provided by Conan.

---


recipes/freeglut/all/conanfile.py
```diff
@@ -24,33 +25,79 @@ class freeglutConan(ConanFile):
...
         "print_warnings_at_runtime": [True, False],
         "replace_glut": [True, False],
+        "with_wayland": [True, False],
+
     }
     default_options = {
         "shared": False,
         "fPIC": True,
-        "gles": False,
+        "gles": True,
```
uilianries, line 34: What's the default value in the upstream? Why can not we keep False?

jwillikers: The default is enabled for Android and disabled otherwise. See [here](https://github.com/FreeGLUTProject/freeglut/blob/c0e9657c1e44ce4ae7091f317215fa2c48541643/CMakeLists.txt#L63-L68). I've updated it to do likewise. I'd figured that enabling GLES support by default would reduce the likelihood of needing to rebuild the binaries for cases that required it.

uilianries (approved):
LGTM. Thank you for sending these patches to the upstream!

RubenRBS (approved):

Issue #20365, opened 2023-10-05
[package] openimageio/2.4.7.1: upstream sha256 has changed


svenpilz:
### Description

Building `openimageio/2.4.7.1` fails with 

```
files.get(
	ConanException: sha256 signature failed for 'v2.4.7.1.tar.gz' file. 
 Provided signature: fd298f71e44c6776863db4b37c4a1388dba0d2eb37378afea95ab07a7cd6ecd4  
 Computed signature: a3dc6fdb3693eb5f1e22191e41c05800a4944f3c76daffe90bd203f956180126
```

Manually downloading and checking the signature confirms the change.

### Package and Environment Details

* Operating System+version: **Linux Ubuntu 20.04**, **Arch**, **Windows**
* Conan version: **conan 1.59.0**

### Conan profile

Fails for different profiles and operating systems.

### Steps to reproduce

conan create openimageio/all "2.4.7.1@"

### Logs

<details><summary>Click to expand log</summary>

```
files.get(
	ConanException: sha256 signature failed for 'v2.4.7.1.tar.gz' file. 
 Provided signature: fd298f71e44c6776863db4b37c4a1388dba0d2eb37378afea95ab07a7cd6ecd4  
 Computed signature: a3dc6fdb3693eb5f1e22191e41c05800a4944f3c76daffe90bd203f956180126
```

</details>

svenpilz:
Maybe because they moved to https://github.com/AcademySoftwareFoundation/OpenImageIO?

RubenRBS:
Hi! Thanks a lot for your report and for checking with the authors for a clarification, once we know more, we'll happily modify it if they confirm it's not unexpected :)

svenpilz:
They think this maybe related to renaming/moving the repository, but cannot test this as the old archives are no longer available on GitHub.

The base folder in the archive might be named after the repository and thus it generated a different archive now. Which they did not anticipate as a side effect.

See https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4003#issuecomment-1750002206.

svenpilz:
As a side note, I suppose the issue might happen with other repositories in https://github.com/AcademySoftwareFoundation as well (if they moved recently).

svenpilz:
@RubenRBS, would you call this a reasonable explanation, or does that scream supply-chain issue and changing the expected signatures would a no-go at the moment?

RubenRBS:
[...]

PR #20910, opened 2023-11-03
openimageio: Adapt to moved upstream repository.

svenpilz:
URLs and checksums changed because the repository was moved to a new location.

GitHub exports the sources in tarballs containing a root folder named like the repository, thus the checksums changed.

See:
* https://github.com/conan-io/conan-center-index/issues/20365
* https://github.com/AcademySoftwareFoundation/OpenImageIO/issues/4003

---

### DIFF
--- a/recipes/openimageio/all/conandata.yml
+++ b/recipes/openimageio/all/conandata.yml
@@ -1,16 +1,16 @@
 sources:
   "2.2.7.0":
-    url: "https://github.com/OpenImageIO/oiio/archive/Release-2.2.7.0.tar.gz"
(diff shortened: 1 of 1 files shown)


svenpilz:
The build failure seems to come form `opencolorio` not finding its `tinyxml` dependency?

uilianries:
> The build failure seems to come form opencolorio not finding its tinyxml dependency?

Probably related to legacy cmake generator ... Need to check the CMakeLists.txt from upstream how is including tinyxml ....

svenpilz:
What would be the way to go here?

With current `master` the whole package is broken and it seems this also needs some maintenance that would happen with the Conan 2 port anyway, right?

RubenRBS:
This was fixed when porting the recipe to v2, sorry about the delay!

PR #20911, opened 2023-11-03
tcl: fix MinGW and properly split build and install in autotools branch

SpaceIm:
I've tried to build tcl with MinGW, and there was several issues:
- Value passed to `build_script_folder` of `Autotools.configure()` was an absolute path, while it should be a relative path. It leads to a duplicate root symbol `//` for *nix platforms, which is not a big deal, but for windows you get something like `/c/c` which doesn't exist.
- Once fixed, there was another error because custom `prefix` passed to `Autotoolschain` was not converted with unix_path, leading very quickly to a configure error since paths separators were stripped out. I've fixed this by simply removing this custom `prefix`, it doesn't play well to mix up `prefix` and `DESTDIR` with the same absolute path anyway (it would have lead to other issues for mingw also, like files installed in weird locations).
- Once fixed, there was another issue that all files of binaries target were not created before other targets depending on these files. This is why now I enforce build of `binaries` target before build of `all`.
I've also properly split build and install. In Autotools branch, everything was built in `package()` method instead of `build()`.

Now it builds fine with MinGW.

---

### DIFF
--- a/recipes/tcl/all/conanfile.py
+++ b/recipes/tcl/all/conanfile.py
@@ -204,14 +204,8 @@ class TclConan(ConanFile):
             fix_apple_shared_install_name(self)
 
         # Relocatable tclConfig.sh
-        # Is it really required to keep this file?
         tclConfigShPath = os.path.join(self.package_folder, "lib", "tclConfig.sh")
-        ## Remove references to build folder
-        build_folder = self.build_folder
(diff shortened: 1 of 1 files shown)


jwillikers (approved):

SpaceIm:
@RubenRBS @uilianries could you review please?

danimtb (approved):

PR #20912, opened 2023-11-03
[OneTBB] Deprecate support for shared option

uilianries:
Specify library name and version:  **onetbb/2021.x**

The OneTBB package uses `shared=True` by default because the upstream does not encourage of using is as a static library. The recipe still offers the option, but add a big warning or an InvalidConfiguration.

This PR proposes deprecating it not only because to follow upstream recommendation and avoid unexpected errors with static library, but also to remedy current missing dependencies error when consuming OneTBB.

So only shared/dynamic libraries will be provided.

---

### DIFF
--- a/recipes/onetbb/all/conanfile.py
+++ b/recipes/onetbb/all/conanfile.py
@@ -25,7 +25,7 @@ class OneTBBConan(ConanFile):
 
     settings = "os", "arch", "compiler", "build_type"
     options = {
-        "shared": [True, False],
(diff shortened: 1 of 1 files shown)


SpaceIm:
This option can't be deprecated with this mechanism, because c3i will force shared option anyway. There is https://github.com/conan-io/conan-center-index/pull/20897 already?

uilianries:
@SpaceIm sorry, I didn't see your PR. Closing this one in favor of #20897

PR #20913, opened 2023-11-03
glfw: Add support for Wayland

jwillikers:
Adds with_wayland option to enable Wayland support in GLFW. In the current release, there is no option to toggle X11 support. The current master branch allows choosing between X11, Wayland, or both. I include the logic for enabling X11 / Wayland for the next release. I can remove it of course, but figured it might save future work.

Update vulkan-loader.
Add patch to workaround issues with ECM not generating sources from Wayland Protocols.

---


recipes/glfw/all/conanfile.py
```diff
@@ -23,49 +27,84 @@ class GlfwConan(ConanFile):
...
         "vulkan_static": [True, False],
+        "with_x11": [True, False],
+        "with_wayland": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "vulkan_static": False,
+        "with_x11": True,
+        "with_wayland": True,
```
uilianries, line 38: Is it possible to use both Wayland and X11 at same time?

recipes/glfw/all/patches/3.3.2-0003-Wayland-Remove-extra-cmake-modules-dependency.patch
```diff
@@ -0,0 +1,122 @@
...
+ #--------------------------------------------------------------------
+ if (_GLFW_WAYLAND)
+-    find_package(ECM REQUIRED NO_MODULE)
+-    list(APPEND CMAKE_MODULE_PATH "${ECM_MODULE_PATH}")
+ 
+-    find_package(Wayland REQUIRED Client Cursor Egl)
+-    find_package(WaylandScanner REQUIRED)
+-    find_package(WaylandProtocols 1.15 REQUIRED)
++    include(FindPkgConfig)
++    pkg_check_modules(Wayland REQUIRED
```
uilianries, line 27: What's the problem by using CMakeDeps files?

recipes/glfw/all/patches/3.3.2-0003-Wayland-Remove-extra-cmake-modules-dependency.patch
```diff
@@ -0,0 +1,122 @@
...
+-        "${WAYLAND_PROTOCOLS_PKGDATADIR}/unstable/idle-inhibit/idle-inhibit-unstable-v1.xml"
+-        BASENAME idle-inhibit-unstable-v1)
+ elseif (_GLFW_OSMESA)
+     set(glfw_HEADERS ${common_HEADERS} null_platform.h null_joystick.h
+                      posix_time.h posix_thread.h osmesa_context.h)
+@@ -62,6 +37,48 @@ elseif (_GLFW_OSMESA)
+                      null_joystick.c posix_time.c posix_thread.c osmesa_context.c)
+ endif()
+ 
++if (_GLFW_WAYLAND)
```
uilianries, line 75: This part of the patch looks something more the applied directly to the upstream than a Conan patch.

recipes/glfw/all/patches/3.3.3-0002-Use-dependencies-from-Conan.patch
```diff
@@ -0,0 +1,38 @@
...
++    pkg_check_modules(Wayland REQUIRED IMPORTED_TARGET
+         wayland-client>=0.2.7
+         wayland-cursor>=0.2.7
+         wayland-egl>=0.2.7
+@@ -254,11 +254,7 @@ if (_GLFW_WAYLAND)
+ 
+     list(APPEND glfw_PKG_DEPS "wayland-client")
+ 
+-    list(APPEND glfw_INCLUDE_DIRS "${Wayland_INCLUDE_DIRS}")
+-    list(APPEND glfw_LIBRARIES "${Wayland_LIBRARIES}" "${CMAKE_THREAD_LIBS_INIT}")
```
uilianries, line 28: It's possible to manipulate the CMakeDeps cmake file names. Does it help, instead of applying a new patch? 

https://docs.conan.io/2/reference/tools/cmake/cmakedeps.html#overwrite-properties-from-the-consumer-side-using-cmakedeps-set-property

uilianries (commented):
Thank you for your PR. I see Wayland is indeed supported by GLFW, but it requires tons of patches. I'm curious whether could be used set_property to manipulate names, instead of all those patches.

recipes/glfw/all/patches/3.3.3-0002-Use-dependencies-from-Conan.patch
jwillikers: I forgot about that. I'll definitely give that a try.

recipes/glfw/all/patches/3.3.2-0003-Wayland-Remove-extra-cmake-modules-dependency.patch
jwillikers: You're right, this fix was pulled from the upstream project, but is absolutely required to build with Wayland. I'll redo the patches based on your other comments and make it so.

uilianries: Thank you!

recipes/glfw/all/patches/3.3.3-0002-Use-dependencies-from-Conan.patch
jwillikers: Yep, that allows me to drop an entire patch by tweaking the xkbcommon CMake config file name to be `XKBCommon`. This is done now.

recipes/glfw/all/conanfile.py
jwillikers: Yes, it is and in fact up to version `3.3.8`, `X11` support can't be disabled. However, upstream changes now allow selectively choosing one, the other, or both as desired. I've done some work to prepare for the next version ahead of time to make it easier to update the recipe in the future.

recipes/glfw/all/patches/3.3.2-0003-Wayland-Remove-extra-cmake-modules-dependency.patch
jwillikers: This particular patch was pulled from the upstream project to remove the dependency on the extra-cmake-modules, which doesn't actually work correctly when used with the Conan package. I think it's easier to actually leave this part of the patch intact rather than attempting to minimize the size of the patch, especially since it would require extra work to define things like they are in extra-cmake-modules. In particular, we'd have to define `WAYLAND_PROTOCOLS_PKGDATADIR` ourselves, whereas the pkg-config uses a standard variable already provided by our pkg-config file for the wayland package.

recipes/glfw/all/patches/3.3.2-0003-Wayland-Remove-extra-cmake-modules-dependency.patch
jwillikers: I've indicated the additional patch as `upstream` instead of `portability` to try and make the patch source and scope clearer.

recipes/glfw/all/conandata.yml
```diff
@@ -21,6 +21,36 @@ sources:
     url: "https://github.com/glfw/glfw/releases/download/3.3.2/glfw-3.3.2.zip"
     sha256: "08a33a512f29d7dbf78eab39bd7858576adcc95228c9efe8e4bc5f0f3261efc7"
 patches:
+  "3.3.8":
+    - patch_file: "patches/3.3.5-0001-Wayland-Remove-extra-cmake-modules-dependency.patch"
+      patch_description: "Remove dependency on extra-cmake-modules to fix generation of Wayland Protocols"
+      patch_type: "upstream"
```
uilianries, line 27:
```suggestion
      patch_type: "official"
```
Upstream is an invalid name:

https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conandata_yml_format.md#patch_type

Please, don't forget to update others.

jwillikers: Thanks!

uilianries (approved):

RubenRBS (approved):
Thanks a lot :)

SpaceIm:
@jwillikers @uilianries @RubenRBS 

This PR breaks all platforms except Linux and FreeBSD by adding xorg if glfw version <= 3.3.8, see all Invalid Configurations in c3i logs

recipes/glfw/all/conanfile.py
```diff
@@ -23,49 +27,85 @@ class GlfwConan(ConanFile):
...
         self.requires("opengl/system")
         if self.options.vulkan_static:
-            self.requires("vulkan-loader/1.3.239.0")
-        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.requires("vulkan-loader/1.3.243.0")
+        if self.options.get_safe("with_x11") or self.version <= Version("3.3.8"):
             self.requires("xorg/system")
+        if self.options.get_safe("with_wayland"):
+            self.requires("wayland/1.22.0")
+            self.requires("wayland-protocols/1.32")
```
SpaceIm, line 76: why wayland-protocols is in requirements?

recipes/glfw/all/conanfile.py
```diff
@@ -23,49 +27,85 @@ class GlfwConan(ConanFile):
...
+            self.requires("xkbcommon/1.6.0")
+
+    def validate(self):
+        if self.options.get_safe("with_wayland") and not self.dependencies["xkbcommon"].options.with_wayland:
+            raise ConanInvalidConfiguration(f"{self.ref} requires the with_wayland option in xkbcommon to be enabled when the with_wayland option is enabled")
+
+    def build_requirements(self):
+        if self.options.get_safe("with_wayland"):
+            self.tool_requires("wayland/<host_version>")
```
SpaceIm, line 85: required_conan_version sould be updated accordingly

recipes/glfw/all/conanfile.py
```diff
@@ -23,49 +27,85 @@ class GlfwConan(ConanFile):
...
-        tc.variables["GLFW_BUILD_EXAMPLES"] = False
-        tc.variables["GLFW_BUILD_TESTS"] = False
-        tc.variables["GLFW_BUILD_DOCS"] = False
-        tc.variables["GLFW_INSTALL"] = True
+        tc.cache_variables["GLFW_BUILD_DOCS"] = False
+        tc.cache_variables["GLFW_BUILD_EXAMPLES"] = False
+        tc.cache_variables["GLFW_BUILD_TESTS"] = False
+        tc.cache_variables["GLFW_BUILD_X11"] = self.options.get_safe("with_x11")
+        tc.cache_variables["GLFW_INSTALL"] = True
+        tc.variables["GLFW_USE_WAYLAND" if self.version <= "3.3.8" else "GLFW_BUILD_WAYLAND"] = self.options.get_safe("with_wayland")
```
ericLemanissier, line 97: `self.version <= "3.3.8"`
This code compares strings, not versions. `Version(...)` must be added

recipes/glfw/all/conanfile.py
```diff
@@ -23,49 +27,85 @@ class GlfwConan(ConanFile):
...
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
         self.requires("opengl/system")
         if self.options.vulkan_static:
-            self.requires("vulkan-loader/1.3.239.0")
-        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.requires("vulkan-loader/1.3.243.0")
+        if self.options.get_safe("with_x11") or self.version <= Version("3.3.8"):
```
sopvop, line 72: This made it impossible to build glfw on windows, because it always requires xorg

jwillikers:
PR #21198 created.

recipes/glfw/all/conanfile.py
jwillikers: Nope. Fixed in #21198.

recipes/glfw/all/conanfile.py
jwillikers: Yep, fixed in #21198.

PR #20920, opened 2023-11-03
duckdb: do not remove import lib during packaging if mingw shared

SpaceIm:
Any logic which tries to naively remove  static lib with pattern `*.a` when we want to keep shared only is broken for MinGW, since import libs have extension `.dll.a`

---


SpaceIm:
I've also removed `0.6.0-0002-include-stdlib.patch` which is no-op (but with warnings) for versions >= 0.7.0 since it has been fixed upstream.

uilianries (approved):

PR #19718, opened 2023-09-10
libarchive: add v3.7.1, update requirements

robomics:
Specify library name and version:  **libarchive/all**

- Add libarchive v3.7.1
- Use version range for zlib requirement
- Fix typo in PCRE requirement

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #20923, opened 2023-11-03
libarchive: fix mingw shared

SpaceIm:
Fix regression of https://github.com/conan-io/conan-center-index/pull/19718

So we build either shared or static in 3.7.1 like all other previous versions, and avoid this removal of .a files if shared which breaks mingw shared since import lib is .dll.a

---

### DIFF
--- a/recipes/libarchive/all/patches/0003-3.7.1-cmake.patch
+++ b/recipes/libarchive/all/patches/0003-3.7.1-cmake.patch
@@ -56,3 +56,36 @@ index 6849ce40..8d5b6018 100644
    FIND_PACKAGE(OpenSSL)
    IF(OPENSSL_FOUND)
      SET(HAVE_LIBCRYPTO 1)
+diff --git a/libarchive/CMakeLists.txt b/libarchive/CMakeLists.txt
+index f7fdfb68..9d2915ac 100644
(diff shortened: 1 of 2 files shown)


jwillikers (approved):

recipes/libarchive/all/patches/0003-3.7.1-cmake.patch
```diff
@@ -56,3 +56,36 @@ index 6849ce40..8d5b6018 100644
    FIND_PACKAGE(OpenSSL)
    IF(OPENSSL_FOUND)
      SET(HAVE_LIBCRYPTO 1)
+diff --git a/libarchive/CMakeLists.txt b/libarchive/CMakeLists.txt
```
uilianries, line 59: Please push this change to the upstream via a PR. I still prefer filtering files into recipes than starting to maintain new patches.

https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#policy-on-patches

SpaceIm: That's not possible to push this change. static lib must be built when you build executables (we disable these targets).

uilianries (approved):

PR #20953, opened 2023-11-06
mesa-glu: Add recipe

jwillikers:
Add a recipe for Mesa's GLU library.

Specify library name and version:  **mesa-glu/9.0.3**

---


recipes/mesa-glu/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+class MesaGluConan(ConanFile):
+    name = "mesa-glu"
+    description = "Mesa's implementation of the OpenGL utility library"
+    license = ("SGI-B-1.1", "SGI-B-2.0", "MIT")
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://www.mesa3d.org/"
+    topics = ("gl", "glu", "mesa", "opengl")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    provides = "libGLU"
```
uilianries, line 24:
```suggestion
```
We dont have such package, neither an open PR. Still, we will lowecase always.

uilianries (commented):
Looking good, just a minor question about `provides` attribute.

recipes/mesa-glu/all/conanfile.py
jwillikers: But `glu/system` provides the same library I'm pretty sure, at least on Linux.

uilianries: Then is should be `glu` instead:

https://docs.conan.io/2/reference/conanfile/attributes.html#provides

recipes/mesa-glu/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+class MesaGluConan(ConanFile):
+    name = "mesa-glu"
+    description = "Mesa's implementation of the OpenGL utility library"
+    license = ("SGI-B-1.1", "SGI-B-2.0", "MIT")
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://www.mesa3d.org/"
+    topics = ("gl", "glu", "mesa", "opengl")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    provides = "libglu"
```
uilianries, line 24:
```suggestion
    provides = "glu"
```

Provides uses the package name, or, in case other package provide too.

recipes/mesa-glu/all/conanfile.py
jwillikers: Got it. I've committed your suggestion.

uilianries (approved):
LGTM

RubenRBS (approved):
Thanks!

arnoldad:
@jwillikers Hello, forgive the naive question, but how does the mesa-glu project differ from mesa itself (such as the release here: https://docs.mesa3d.org/relnotes/24.0.2.html) I came here looking to see if Conan could mange a dependency our project has on mesa, and it really wasn't clear in the git repo for glu (https://gitlab.freedesktop.org/mesa/glu) what exactly is the intention, and there is no readme, etc. Is this perhaps a legacy version of mesa?

jwillikers:
> @jwillikers Hello, forgive the naive question, but how does the mesa-glu project differ from mesa itself (such as the release here: https://docs.mesa3d.org/relnotes/24.0.2.html) I came here looking to see if Conan could mange a dependency our project has on mesa, and it really wasn't clear in the git repo for glu (https://gitlab.freedesktop.org/mesa/glu) what exactly is the intention, and there is no readme, etc. Is this perhaps a legacy version of mesa?

Hi @arnoldad, no problem. `mesa-glu` is Mesa's implementation of the GLU, i.e. GL Utility library, which is a rather small library that was split out from Mesa proper. I would guess that if you depend on Mesa, it's very likely not what you need. Depending on Mesa though is pretty since it provides a lot of different things, do you know if your project depends on a specific library like `libgbm`, a particular graphics acceleration API like OpenCL, OpenGL, OpenGLES, or Vulkan,  or a software GL implementation like LLVMPipe?

arnoldad:
@jwillikers Thanks for taking the time to reply. It looks like the project I’ve adopted uses mesa’s openGL implementation to do things such as render groups of mesh files. So it seems that this is not what I am after - but have found another Conan index issue with a request to create a package for mesa proper - I will follow up there. Thanks again!

PR #20962, opened 2023-11-06
libdicom: add a new recipe

valgur:
Adds libdicom: https://github.com/ImagingDataCommons/libdicom

[Packaging status](https://repology.org/project/libdicom/versions)

### DIFF
--- a/recipes/libdicom/all/conanfile.py
+++ b/recipes/libdicom/all/conanfile.py
@@ -79,5 +79,5 @@ class LibdicomConan(ConanFile):
         fix_apple_shared_install_name(self)
 
     def package_info(self):
-        self.cpp_info.libs = ["libdicom"]
+        self.cpp_info.libs = ["dicom"]
         self.cpp_info.set_property("pkg_config_name", "libdicom")
--- a/recipes/libdicom/all/conanfile.py
+++ b/recipes/libdicom/all/conanfile.py
@@ -81,3 +81,5 @@ class LibdicomConan(ConanFile):
     def package_info(self):
         self.cpp_info.libs = ["dicom"]
         self.cpp_info.set_property("pkg_config_name", "libdicom")
+        if not self.options.shared:
+            self.cpp_info.defines = ["DCM_STATIC"]
--- a/recipes/libdicom/all/conanfile.py
+++ b/recipes/libdicom/all/conanfile.py
@@ -1,0 +1,83 @@
+import os
+
+from conan import ConanFile
+from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import copy, get, rm, rmdir
+from conan.tools.gnu import PkgConfigDeps
+from conan.tools.layout import basic_layout
+from conan.tools.meson import Meson, MesonToolchain
+
+required_conan_version = ">=1.53.0"
+
+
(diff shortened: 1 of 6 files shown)


valgur:
Requires #20964.

jwillikers (approved):

recipes/libdicom/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+        meson.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
+        fix_apple_shared_install_name(self)
+
+    def package_info(self):
+        self.cpp_info.libs = ["dicom"]
+        self.cpp_info.set_property("pkg_config_name", "libdicom")
```
uilianries, line 83:
```suggestion
        self.cpp_info.set_property("pkg_config_name", "libdicom")
        if not self.options.shared:
            self.cpp_info.defines = ["DCM_STATIC"]
```

https://github.com/ImagingDataCommons/libdicom/blob/v1.0.5/include/dicom/dicom.h#L11

valgur: Good catch, thanks!

franramirez688: @uilianries  @valgur  The `meson.build` is setting it by getting the `default_library` option. Perhaps, it'd be better to add it via MesonToolchain:

```python
        tc = MesonToolchain(self)
        tc.project_options["tests"] = "false"
        if not self.options.shared:
             tc.project_options["default_library"] = "static"
        tc.generate()
```

recipes/libdicom/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+        if not self.options.shared:
+            self.cpp_info.defines = ["DCM_STATIC"]
```
franramirez688, line 85: Sorry, I just double-checked it, and MesonToolchain is already defining the `default_library` in the built-in options section, so we can remove that part:

```suggestion
```

recipes/libdicom/all/conanfile.py
franramirez688: Follow-up: https://github.com/conan-io/conan-center-index/pull/20962/files#r1392573896

recipes/libdicom/all/conanfile.py
valgur: Sorry, I'm confused. This exported define is not really related to MesonToolchain?

franramirez688: Sorry, yes, it's related, and it's already defined by the `default_library` in the `conan_meson_xxxx.ini` within the `[built-in options]` section. See https://github.com/ImagingDataCommons/libdicom/blob/ba0e6d6f1fa08d69a4582b38817ca7ec699705e5/meson.build#L62C46-L62C61

franramirez688: Discussed with @uilianries Keeping it to prevent missing symbols.

franramirez688 (approved):

jwillikers (approved):

PR #20966, opened 2023-11-06
tpl: add a new recipe

valgur:
Adds tpl: https://troydhanson.github.io/tpl/

[Packaging status](https://repology.org/project/tpl/versions)

### DIFF
--- a/recipes/tpl/all/conandata.yml
+++ b/recipes/tpl/all/conandata.yml
@@ -1,0 +1,5 @@
+sources:
+  "cci.20210302":
+    url: "https://github.com/troydhanson/tpl/archive/f8138ad393f4b1985c916029ab6d703e4e7a1c4c.tar.gz"
+    sha256: "736dcd98bc92178a9cfaeaf8d7f9a67aa49c5ee36b9f76286e055a8a97b640dd"
+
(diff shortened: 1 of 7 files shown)


recipes/tpl/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "1.6.1.cci.20210302":
```
danimtb, line 2: the new convention for the cci version format is still not in place as commented here https://github.com/conan-io/conan-center-index/pull/24411#pullrequestreview-2138128755

Basically, we would need to drop support of conan 1 first so we can have it as an official version format to be fully compatible with version ranges. Sorry if it was not clear
```suggestion
  "cci.20210302":
```

recipes/tpl/config.yml
```diff
@@ -0,0 +1,3 @@
+versions:
+  "1.6.1.cci.20210302":
```
danimtb, line 2:
```suggestion
  "cci.20210302":
```

valgur:
.

valgur:
You win this time github-actions bot. The recipe is clearly too complex and not worth the effort.

PR #20979, opened 2023-11-07
'python2 --version' sends output to stderr, redirect to stdout to cat…

wdobbe:
Specify library name and version:  **qt/5.15.11**

The qt5 qtwebengine module still needs python2 to build it. The qt5 conanfile.py checks if a valid python2 version is installed by running 'python2 --version'.
However the python2 executable sends the version output to stderr (python3 sends it to stdout). As a result the stdout output of the self.run(...) command is empty and results in an error:

```
qt/5.15.11: RUN: "/usr/bin/python2" --version
Python 2.7.18

ERROR: qt/5.15.11: Error in validate_build() method, line 156
        verstr = mybuf.getvalue().strip().split("Python ")[1]
        IndexError: list index out of range
```
Because self.run(...) doesn't seem to have an argument to catch the stderr output I changed the 'python2 --version' command to send stderr output to stdout.

Tested on Linux with conan 2.0.13.

---


ericLemanissier (approved):

recipes/qt/5.x.x/conanfile.py
```diff
@@ -151,7 +151,7 @@ def validate_build(self):
             # In any case, check its actual version for compatibility
             from six import StringIO  # Python 2 and 3 compatible
             mybuf = StringIO()
-            cmd_v = f"\"{python_exe}\" --version"
+            cmd_v = f"\"{python_exe}\" --version 2>&1"
```
RubenRBS, line 154: This won't work on Windows - You might need to use `subprocess` to execute the command directly and get stderr from there

I'll look into allowing Conan's run method to also output stderr just like what is done with stdout

RubenRBS (changes requested):
Thanks! I don't think this will work on Windows though, but I've proposed a workaround

recipes/qt/5.x.x/conanfile.py
wdobbe: Hi Ruben,

Thanks for reviewing.
Before submitting I tested the "`python --version 2>&1`" command in a Windows command prompt and it worked.
A colleague of mine is currently building qt 5.15.11 on Windows and with the above patched he has no errors yet. Before the patch the build would quit almost immediately because of the broken Python2 version check.
So I am quite confident this solution will work also on Windows. 

Of course it would still be nice if stderr could be re-directed as well.

uilianries: @wdobbe Please, let's avoid workarounds and follow Ruben's approach, the `subprocess` is specially designed for this case, it's cross-platform compatibly, has fine-grained control over pipes, is more secure (avoids injections), and much better for error handling in case of an exception.

wdobbe: Changed as requested and tested with Conan 2.13.0 on Linux x86_64 with:
`conan create . --version=5.15.11 -o qt/*:qtwebengine=True -o sqlite3/*:shared=True -o qt/*:shared=True -o qt/*:gui=True -o qt/*:qtdeclarative=True -o qt/*:qtlocation=True -o qt/*:qtwebchannel=True --build=missing`

ericLemanissier (approved):

valgur:
The missing `libmysqlclient` binary should now be available after #21582 was merged.

wdobbe:
@valgur Yes, thanks. I already did a rebase and push to re-trigger a build for PR #21026 (also qt/5.x) but it now results in the following error:
```
opengl/system: WARN: System requirements: 'libgl-dev' are missing but can't install because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'
opengl/system: WARN: System requirements: 'libgl1-mesa-dev' are missing but can't install because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'
opengl/system: ERROR: while executing system_requirements(): None of the installs for the package substitutes succeeded.
```
Comment from Eric Lemanissier:

>  I guess conan team removed tools.system.package_manager configuration from C3I, you should see with the repo admins

tsondergaard:
> @valgur Yes, thanks. I already did a rebase and push to re-trigger a build for PR #21026 (also qt/5.x) but it now results in the following error:
> 
> ```
> opengl/system: WARN: System requirements: 'libgl-dev' are missing but can't install because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'
> opengl/system: WARN: System requirements: 'libgl1-mesa-dev' are missing but can't install because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'
> opengl/system: ERROR: while executing system_requirements(): None of the installs for the package substitutes succeeded.
> ```
> 
> Comment from Eric Lemanissier:
> 
> > I guess conan team removed tools.system.package_manager configuration from C3I, you should see with the repo admins

Looking at #21943 it appears it ran into the same problem and it has been fixed now. @valgur, could you trigger a rebuild?

ericLemanissier (approved):

recipes/qt/5.x.x/conanfile.py
```diff
@@ -148,12 +149,14 @@ def validate_build(self):
...
-            self.run(cmd_v, mybuf)
-            verstr = mybuf.getvalue().strip().split("Python ")[1]
+            # In any case, check its actual version for compatibility.
+            # Note that python2 exe sends its version to stderr instead of stdout.
+            completed = subprocess.run( [python_exe, '--version'], stderr=subprocess.PIPE,
+                                            timeout=5 )
+            if completed.returncode != 0:
+                msg = ("Found python2 executable %s but could not run it" % python_exe)
+                raise ConanInvalidConfiguration(msg)
+            verstr = completed.stderr.decode('utf-8').split("Python ")[1]
```
valgur, line 159:
```suggestion
            verstr = completed.stderr.decode('utf-8').split("Python ")[1].strip()
```
Just to be safe.

valgur (approved):

uilianries:
There is another important point about this issue, it targets Python 2, which is EOL, and Conan 2.x. The Conan 2.x explicitly requires Python 3.6: https://github.com/conan-io/tribe/blob/main/design/003-codebase-python.md

ericLemanissier:
sure, but in qt5 [qtwebengine requires python 2](https://github.com/qt/qtwebengine/blob/v5.15.16-lts/src/buildtools/configure.json#L398), because the chromium version used under the hood requires python 2.
The part of the recipe modified by this PR raises an exception with a clear message so that the consumer has the information. Otherwise qt itself just disables qtwebengine and continue the build with a simple warning.

jcar87:
Hi @wdobbe - thank you for your contribution and please accept my apologies for the delay in ensuring this PR went through CI validation.

As part of reviewing a different PR, I've  proposed a simpler approach that allows us to still call `self.run`:
https://github.com/conan-io/conan-center-index/pull/24768

I'll make sure this is reviewed and merged this week.

jcar87:
Closing via https://github.com/conan-io/conan-center-index/pull/24768.

Apologies it took this long to fix this.

PR #20984, opened 2023-11-07
fix exiv2 provides

memsharded:
Specify library name and version:  **exiv2**

Follow up from https://github.com/conan-io/conan/issues/15065, fixing ``provides`` definition

---

### DIFF
--- a/recipes/exiv2/all/conanfile.py
+++ b/recipes/exiv2/all/conanfile.py
@@ -44,8 +44,6 @@ class Exiv2Conan(ConanFile):
         "win_unicode": False,
     }
 
-    provides = []
-
     def export_sources(self):
         export_conandata_patches(self)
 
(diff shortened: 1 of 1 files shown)


recipes/exiv2/all/conanfile.py
```diff
@@ -44,8 +44,6 @@ class Exiv2Conan(ConanFile):
         "win_unicode": False,
     }
 
-    provides = []
```
Nekto89, line 47: is there any difference between having "provides = None" and not using it explicitly?

uilianries: Yes, in conan client code, when checking dependencies graph, is expected `None` is case providing nothing, using list it will result in an undesirable behavior:

https://github.com/conan-io/conan/blob/9d48c916ca51713945812bf8b5668ddcd9a303c5/conans/client/graph/provides.py#L16

RubenRBS (approved):

PR #20991, opened 2023-11-08
rust: new recipe with pre-built binaries

valgur:
Adds Rust toolchain. Mostly for [`librsvg`](https://gitlab.gnome.org/GNOME/librsvg).

Rust is quite straightforward to build from sources, but due to being based on LLVM, the memory consumption in C3I will almost certainly exceed its limits. It also builds slowly: 20+ minutes locally and probably 1.5 h or more in CI.


uilianries:
The [librsvg](https://gitlab.gnome.org/GNOME/librsvg) has C files, that can be built without Rust. I'll add to discuss with Conan team about this PR, as Rust is not supported by Conan. 

@valgur Do really need Rust for a project, or only due librsvg?

valgur:
> The [librsvg](https://gitlab.gnome.org/GNOME/librsvg) has C files, that can be built without Rust. I'll add to discuss with Conan team about this PR, as Rust is not supported by Conan. 
> 
> @valgur Do really need Rust for a project, or only due librsvg? 

Are you sure? The C portion of the library is just a wrapper around the Rust library, as far as I can tell. See also the related librsvg PR: #21005

I don't have any immediate need for Rust besides librsvg, but I think it's a common and important enough library to warrant Rust support in CCI.

As a build tool dependency, Rust is very well-behaved and comparable to CMake or Ninja. Binaries produced by Rust link statically against Rust's libstd by default. The only minor issues I can think of are (1) the size of the Rust package (300 MB) which can become a nuisance if multiple different versions are kept in the cache, (2) Rust caches its Cargo packages inside the home directory, which can be unexpected and unwanted, (3) cross-compiling might be difficult.

uilianries (changes requested):
Thank you for submitting your pull request. After discussing it with the Conan team, we have reservations about accepting it in the short to medium term. This hesitation stems from potential issues arising after the implementation is integrated into the Conan Center.

The Rust toolkit provides libraries that may have runtime dependencies, and these dependencies are currently not adequately tracked by the package settings due to the existing design. This fragility could lead to misconfigurations with the package ID, posing challenges to the overall stability and reliability of the users.

valgur:
> The Rust toolkit provides libraries that may have runtime dependencies, and these dependencies are currently not adequately tracked by the package settings due to the existing design. This fragility could lead to misconfigurations with the package ID, posing challenges to the overall stability and reliability of the users.

@uilianries That's a valid concern, for sure, but shouldn't the potential runtime dependency issues be looked at the consuming recipes case-by-case, instead? As I mentioned, the default in Rust is to link its dependencies statically, after all.

jcar87:
Closing echoing @uilianries message. This is not something we are prioritising right now. I would suggest to open an issue so that other uses can articulate their needs for a rust recipe, which we may considered in the future.

valgur:
@jcar87 This is a fully functional recipe and a perfectly fine location for discussion you suggested, I think. Could you please keep it open for the time being until external toolchain support is on you radar?

PR #20999, opened 2023-11-08
fix(openssl): remove renaming of debug files within the recipe

Sil3ntStorm:
Specify library name and version:  **openssl/***

Removes the "d" suffix which is only ever introduced in conan, without there being any need in a conan environment as different builds are always in different directories.

---

### DIFF
--- a/recipes/openssl/1.x.x/conanfile.py
+++ b/recipes/openssl/1.x.x/conanfile.py
@@ -539,10 +539,6 @@ class OpenSSLConan(ConanFile):
             with chdir(self, self.source_folder):
                 self.run(f"nmake -f Makefile install_sw DESTDIR={self.package_folder}")
             rm(self, "*.pdb", self.package_folder, recursive=True)
-            if self.settings.build_type == "Debug":
-                with chdir(self, os.path.join(self.package_folder, "lib")):
-                    rename(self, "libssl.lib", "libssld.lib")
-                    rename(self, "libcrypto.lib", "libcryptod.lib")
         else:
             autotools = Autotools(self)
             with chdir(self, self.source_folder):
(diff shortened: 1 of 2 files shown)


SpaceIm:
Was there not already a PR for this?

Just for reviewers, be aware that several recipes have fragile workaround for this d postfix introduced by conan, and it may break if it's removed, like librdkafka since a recent PR: https://github.com/conan-io/conan-center-index/pull/17894#discussion_r1382124697

Sil3ntStorm:
There was #10939 but it never got anywhere and the recipe had changed to much. So this is the current version superseding the previous one.

uilianries:
@Sil3ntStorm Could you please explain the case that you are trying to fix and is blocking/breaking you? I'm not confident of changing OpenSSL in case there is no real blocker involved, this recipe affects many others packages in Conan Center.

jwillikers (approved):

Croydon (approved):
In theory we never break end-consumers because everyone is using revisions anyway 😅 

And if they don't use revisions then they get breaks all the time from CCI.

This is a change of OpenSSL that we should never have added. It changes the outcome from what upstream intended and the current state might actual be broken for some consumers.

I would rather fix those theoretical cases than avoiding some fixing of other recipes.

I did a quick search for `libssl` and `libcrypto` in all recipes, looked at the results and and it looks like only `librdkafka` needs updating

danimtb (approved):

jcar87:
> In theory we never break end-consumers because everyone is using revisions anyway 😅
> And if they don't use revisions then they get breaks all the time from CCI.

Indeed. this PR is likely to have issues for Conan 1.x users that do not have revisions enabled. If they don't have revisions enabled and already have Debug binaries for this library, the package info() will be wrong and they won't be able to link.

Even with revisions enabled, downstream consumers are only rebuilt if they use the static library, but won't be rebuilt if they depend on the shared variant.
 If changing the `.lib` filename causes changes to the dll name that  will be embedded in a downstream DLL consumer - then this is a breaking change even in that scenario, that would require a rebuild of the consumer package.

Changing the name of the library that gets embedded in the downstream consumers (the name of the `.dll` or the portion of the library name that gets embedded in the `DT_NEEDED` or equivalent in Lunux or mac, including the SONAMe) is akin to a breaking ABI change, but Conan cannot know this from the version alone (especially if only the recipe revision changed). This is why these things should never be modified.

> This is a change of OpenSSL that we should never have added. It changes the outcome from what upstream intended and the current state might actual be broken for some consumers.

I fully agree. We are going to be more careful and reject any changes like this in future recipes. I've noticed some contributors renaming libraries because it then makes for a "cleaner" implementation of the `libs` in `package_info()` - however, the library filenames should *not* be changed. Arguably, handling this conditionally in the `package_info()` is equivalent complexity/lines of code. We are going to be more careful as reviews.

> I would rather fix those theoretical cases than avoiding some fixing of other recipes.

Agree !

jcar87 (changes requested):
Agree with the changes, but this may actually cause breakages - we are evaluating the risk before merging this

jcar87:
Update: have verified that the DLL names are the same regardless of the name of the import library (.lib). So any issues with revisions would be limited to Conan 1.x users who update the recipe without rebuilding the package, if such package exists.
Will re-run the build to generate the binaries for the most recent versions before merging.

jcar87 (approved):

PR #21000, opened 2023-11-08
hdf5: add version 1.14.3

KingKili:
Specify library name and version:  **hdf5/1.14.3**

This PR adds the new version of hdf5, 1.14.3. Furthermore, it also adds the ability to cross compile, if the Version is 1.14.3 and QEMU is specified in the conf.
QEMU will be set as CMAKE_CROSSCOMPILING_EMULATOR, allowing hdf5 to detect the necessary settings for the target platform.

---


recipes/hdf5/all/conanfile.py
```diff
@@ -46,6 +46,12 @@ class Hdf5Conan(ConanFile):
         "enable_unsupported": False
     }
 
+    def get_crossbuild_emulator(self):
+        for (k, v) in self.conf.items():
+            if k == "tools.qemu:path":
```
RubenRBS, line 51: Where does this conf name come from? There's nothing like it in the client - And it won't let recipes create new `tool.` confs not built-in.

I'm guessing you are looking for something like `self.conf.get("user.qemu:path")` or similar in this case :)

RubenRBS (commented):
Thanks a lot for your contibution! I have some concerns about the conf, but let's address that later, for now I have a question about its name

recipes/hdf5/all/conanfile.py
KingKili: As was unable to find an official qemu configuration setting. Therefore, I decided to define my own name. This was the best thing I could think off at the time🤷‍♂️.
But you preferer, I can certainly switch to `self.conf.get("user.qemu:path")`.

KingKili: @RubenRBS Should I go back to my iterating over conf items approach. Or is there another API I am missing?

recipes/hdf5/all/conanfile.py
```diff
@@ -118,6 +120,10 @@ def generate(self):
         cmakedeps.generate()
 
         tc = CMakeToolchain(self)
+        if not can_run(self):
+            c = self.conf.get("user.qemu:path")
```
Nekto89, line 124: you can inject CMAKE_CROSSCOMPILING_EMULATOR without changing recipe through user_toolchain

in profile:
[conf]
hdf5*:tools.build.cross_building:can_run=true
tools.cmake.cmaketoolchain:user_toolchain=[".....path_to_toolchain.cmake"]

in toolchain:
set(CMAKE_CROSSCOMPILING_EMULATOR ...

Or with more modern cmake (3.28+) you can even use environment variable through profile instead of toolchain

KingKili: Great Idea, unfortunately in the recipe will be still required, since it is currently hardwired to fail in a cross-build environment ([conanfile.py](https://github.com/conan-io/conan-center-index/blob/35d081612727c323cf5b8b698fe68dcb413353da/recipes/hdf5/all/conanfile.py#L88-L90)).
To solve this issue, we have three options in my eyes:
1. We could either blindly allow cross compilation and accept the fact the builds will fail when crossbuilding.
2. With you proposed alternative, we must check if CMAKE_CROSSCOMPILING_EMULATOR is set in the toolchain file (At least in my case is CMake 3.28 not available :-/). 
3. Like it is currently implemented, we add an option to the recipe.
What is the best option moving forward in your opinion?

KingKili:
Ping? Does anyone know how I should continue?

recipes/hdf5/all/conanfile.py
Nekto89: > Great Idea, unfortunately in the recipe will be still required, since it is currently hardwired to fail in a cross-build environment ([conanfile.py](https://github.com/conan-io/conan-center-index/blob/35d081612727c323cf5b8b698fe68dcb413353da/recipes/hdf5/all/conanfile.py#L88-L90)). To solve this issue, we have three options in my eyes:
> 
>     1. We could either blindly allow cross compilation and accept the fact the builds will fail when crossbuilding.
> 
>     2. With you proposed alternative, we must check if CMAKE_CROSSCOMPILING_EMULATOR is set in the toolchain file (At least in my case is CMake 3.28 not available :-/).
> 
>     3. Like it is currently implemented, we add an option to the recipe.
> 
> 
> What is the best option moving forward in your opinion?

With hdf5*:tools.build.cross_building:can_run=true it will not fail in this check

Nekto89: I've just shared how I am currently using this recipe with emscripten (hdf5 1.14.1).

KingKili: Sorry for bothering you again. I cannot get your approach to work with a conanfile.py. 
Currently I use:
```Python
 def requirements(self):
        self.requires("hdf5/1.14.3")
        self.options["hdf5:tools.options.cross_building"].can_run = True
```
Unfortunately, this does not change anything, and the build still fails with:
`hdf5/1.14.3: Invalid ID: Current recipe doesn't support cross-building (yet)`

Do you have any idea what I am missing?

PR #21005, opened 2023-11-08
librsvg: add a new recipe

### LINKED ISSUE #11392: [request] librsvg/2.54
### Package Details
  * Package Name/Version: **librsvg/2.54**
  * Website: **https://gitlab.gnome.org/GNOME/librsvg**
  * Source code: **https://gitlab.gnome.org/GNOME/librsvg**

Can preview and convert svg to image, used by gimp

valgur:
Adds librsvg: https://gitlab.gnome.org/GNOME/librsvg

Depends on Rust: #20991

- Resolves #11392

[Packaging status](https://repology.org/project/librsvg/versions)

### DIFF
--- a/recipes/librsvg/all/conandata.yml
+++ b/recipes/librsvg/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "2.57.0":
+    url: "https://download.gnome.org/sources/librsvg/2.57/librsvg-2.57.0.tar.xz"
+    sha256: "335fe2e0c2cbf1b7bf0668651224a23e135451f0b1793cd813649be2bffa74e8"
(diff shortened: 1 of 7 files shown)


valgur:
Unfortunately, this recipe is currently seriously broken due to the shortcomings of the [`system_deps`](https://docs.rs/system-deps/latest/system_deps/) Cargo crate used to wrap "system" dependencies like Pango, Cairo and all other dependencies provided by Conan in the Rust build. The `system_deps` crate adds all libdir paths to the linker flags simultaneously, which causes the .so files from `/usr/lib/...` to be picked up by the linker due to `xorg/system` and `opengl/system`. 
I managed to work around it in the project by heavily patching the `build.rs` files, but it is also affects all transitive deps like `cairo-sys-rs`, etc.

Would have to collaborate with `system_deps` to get this fixed more generally or find a way to inject a patched version of it for all projects making use of it, which is far from ideal. :slightly_frowning_face:

valgur:
.

valgur:
Regardless of rustc/cargo being available on CCI, this recipe is currently not viable due to `system_deps` crate not handling relocated libraries correctly.

PR #21026, opened 2023-11-09
Fix assembly error in qt5 webengine module with binutils >= 2.41

wdobbe:
Specify library name and version:  **qt/5.15.11**

The build of the qt5 qtwebengine module fails during the compilation of ffmpeg that comes with the chromium engine:
`3rdparty/chromium/third_party/ffmpeg/libavcodec/x86/mathops.h:125: Error: operand type mismatch for `shr'`

In qt6 this has been fixed upstream (see https://bugreports.qt.io/browse/QTBUG-116649).
Add this patch also to qt5.

Patch tested with Conan 2.0.13 on Linux x86_64.

---


recipes/qt/5.x.x/conandata.yml
```diff
@@ -108,6 +108,11 @@ patches:
       patch_description: "Fix usage of memory_resource with apple-clang>=15 and deployment target of macOS < 14"
       patch_type: "portability"
       patch_source: "https://codereview.qt-project.org/c/qt/qtbase/+/482392"
+    - patch_file: "patches/ffmpeg-x86-optimization.patch"
+      base_path: "qt5/qtwebengine"
+      patch_description: "Fix build error with binutils >= 2.41"
+      patch_type: "portability"
+      patch_source: "https://packages.ubuntu.com/source/focal-updates/qtwebengine-opensource-src"
```
jwillikers, line 115: This should probably point to the upstream bug report instead. 

```suggestion
      patch_source: "https://trac.ffmpeg.org/ticket/10405"
```

jwillikers (approved):

RubenRBS (approved):
Thanks a lot for your contribution and patience :)

RubenRBS (changes requested):
Blocking for a sec until I check one last thing, sorry about the noise!

RubenRBS (commented):
Quick question:
- Why is it not applied to all 5.x versions?
- Can the base_path be changed so that the patch itself can be the same used in upstream?

wdobbe:
>     * Why is it not applied to all 5.x versions?
Because I have no time to check the patch with all older qt5 versions, qtwebengine takes forever to build.
CCI builds also don't build the qtwebengine module, I suppose for the same reason (build resources). So I didn't want to add a patch without knowing if it will apply correctly.
Qt5 has been end-of-life for some years and the opensource version is not maintained anymore by the Qt Company. [KDE took over maintenance](https://ubunlog.com/en/kde-toma-las-riendas-del-mantenimiento-de-la-rama-publica-de-qt-5-15) but they also maintain only qt 5.15.
> 
>     * Can the base_path be changed so that the patch itself can be the same used in upstream?
Yes. The patch I added was taken from the Debian Qt package. I will change it then to the original ffmpeg patch even though I don't see the benefit.

valgur (approved):

jcar87:
Thanks @wdobbe  for your contribution and apologies it has taken long for us to review this. I'm preparing a PR with fixes to qt6 webengine for all platforms, and will cherry pick this and include this, preserving your authorship.

A couple of points: 
Qt 5.15 is still maintained by the Qt company until 2025: https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders - although for commercial customers. It has been observed that a year after their commercial release, the patch releases are made open source.

The most recent qt 5.15 relases (including qt 5.15.14 which the latest we maintin), are based on Chromium version 87 (see https://wiki.qt.io/QtWebEngine/ChromiumVersions)

The patch in question https://bugreports.qt.io/browse/QTBUG-116649 has been backported to their 87 branch as per https://codereview.qt-project.org/c/qt/qtwebengine-chromium/+/537257 - so this is enough for us to confirm that Qt has backported the fix to the chromium version in the 5.15.x branch. We are a bit more careful/skeptical when patches from Qt 6.x are backported when there is no indication that Qt themselves would do it - we have no intention to maintain Qt5 beyond the official open source releases - although in this case it is clear it is maintained and likely to be part of future 5.15.x releases.

jcar87:
Fixed in https://github.com/conan-io/conan-center-index/pull/24826
(preserving the authorship in of this PR)

Issue #21048, opened 2023-11-10
[package] rocksdb/6.27.3: Compilation failure under gcc-13 due to missing #include <cstdint>


Bronek:
## Summary

Package rocksdb/6.27.3 fail to compile under gcc-13 due to missing `#include <cstdint>` in several files. This include was not needed in user code when compiling with older gcc versions because of indirect includes  ( https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes ). The problem has been fixed in rocksdb 8 and subsequent versions https://github.com/facebook/rocksdb/pull/11118 , however this version is not available in conan repository and there are no backport available for older versions.

## Possible workarounds

* Upgrade to version rocksdb/8.x.y . Downsides:
   - not currently available in conan
   - migration effort to newer rocksdb API version
   - extensive re-certification/re-testing effort after migration to newer rocksdb version (due to changes required by migration)
* Patch system `/usr/include/c++/13/string` to indirectly `#include <cstdint>`, bringing back old gcc-12 behaviour in this respect . Downsides
   - not portable
   - limited applicability
* Magic ?

### Package and Environment Details

* Package Name/Version: **rocksdb/6.29.5**
* Operating System+version: **Fedora 39**
* Compiler+version: **GCC 13**
* Conan version: **conan 1.61.0**
* Python version: **Python 3.11.6**

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=armv8
arch_build=armv8
compiler=gcc
compiler.version=13
compiler.libcxx=libstdc++11
build_type=Release
compiler.cppstd=20
[options]
[build_requires]
[env]
[conf]
tools.build:compiler_executables={'c': '/usr/bin/gcc', 'cpp': '/usr/bin/g++'}

### Steps to reproduce

```
cd recipes/rocksdb
conan create all/conanfile.py rocksdb/6.29.5@ -pr:b=default -pr:h=default
```

### Logs

<details><summary>Click to expand log</summary>

```
[HOOK - conan-center.py] pre_export(): [DEPRECATED GLOBAL CPPSTD (KB-H001)] OK
[HOOK - conan-center.py] pre_export(): [REFERENCE LOWERCASE (KB-H002)] OK
[HOOK - conan-center.py] pre_export(): [RECIPE METADATA (KB-H003)] OK
[HOOK - conan-center.py] pre_export(): [HEADER_ONLY, NO COPY SOURCE (KB-H005)] OK
[...]

PR #21049, opened 2023-11-10
Backport gcc-13 fix from rocksdb/pull/11118

Bronek:
Specify library name and version:  **rocksdb/6.29.5**

This fixes https://github.com/conan-io/conan-center-index/issues/21048 by backporting upstream fix https://github.com/facebook/rocksdb/pull/11118 into a new `0002-Add-include-cstdint-for-gcc-13.patch` file . The file modifies 3 instead of original 2 files. The extra file is `include/rocksdb/utilities/checkpoint.h` - due to changes in rocksdb API, this file relies on cstdint types in version 6, but not in version 8 (where https://github.com/facebook/rocksdb/pull/11118 was applied).

---


Bronek:
Reviewers, please kindly indicate **if**:
* equivalent patches should be prepared for older rocksdb versions
* extra field `patch_type` should be added (this is `portability` for gcc 13)
* extra field `patch_source` should be added (would point to https://github.com/facebook/rocksdb/pull/11118 )

Bronek:
@conan-io/barbarians please note the v2 build is held by https://github.com/conan-io/conan-center-index/pull/17186

Bronek:
> Hi @Bronek thanks a lot for your work in first reporting the issue and taking the time to patch it accordingly, we really appreciate it :)
> 
> As for your questions:
> 
> > equivalent patches should be prepared for older rocksdb versions:
> 
> If this is something that might bite somebody else, and if you want to, then yes! We're more than happy to accept the extra contribution!

Thanks ! Turns out the patch which I am backporting (with minor modification explained at top) fits only two version 6.29 and 6.27 . For older versions the patch would need to look very different (it would be more intrusive), and so I'm afraid it might not pass the patching policies. I would be happy to prepare it, if there's a demand, but under different PR as it won't be straight backporting anymore.

> We're not enforcing them, but yes, they are super useful for future maintenance, where they will make it easier to check if they are still relevant and how

I tried doing that, but I think it would need to be added in https://github.com/conan-io/conan-center-index/pull/17186 which adds conan 2 support.

So it seems like this PR will not receive further changes, unless explicitly requested.

Bronek:
> > equivalent patches should be prepared for older rocksdb versions:
> 
> If this is something that might bite somebody else, and if you want to, then yes! We're more than happy to accept the extra contribution!

Thanks ! Turns out the patch which I am backporting (with minor modification explained at top) fits only two version 6.29 and 6.27 . For older versions the patch would need to look very different (it would be more intrusive), and so I'm afraid it might not pass the patching policies. I would be happy to prepare it, if there's a demand, but under different PR as it won't be straight backporting anymore.

> We're not enforcing them, but yes, they are super useful for future maintenance, where they will make it easier to check if they are still relevant and how

danimtb (approved):

ericLemanissier (approved):

PR #21052, opened 2023-11-10
crashpad: migrate to Conan v2

valgur:
Requires
- #18847


memsharded:
Superseded in favor of https://github.com/conan-io/conan-center-index/pull/24320.

That PR is doing the most "basic" migration, just doing the very minimum to make it work with Conan 2 and Conan 1.
There would be some missing aspects included in this PR that are not included, but the priority is to make it build and work incrementally from there, for example:
- Adding latest 2023 version
- Adding the ``shared`` option to allow crashpad to be shared library
- Cleaner separation of generate and build methods

PR #21058, opened 2023-11-10
[ginkgo] add version 1.7.0

upsj:
ginkgo/1.7.0

---

- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated. (I've tried installing the hooks, but it seems that the documentation is stuck halfway between v1.x and v2.x and referring to a command that no longer exists, namely `conan config set hooks.conan-center`)


recipes/ginkgo/all/conandata.yml
```diff
@@ -1,4 +1,13 @@
 sources:
+  "1.7.0":
+    url: "https://github.com/ginkgo-project/ginkgo/archive/v1.7.0.tar.gz"
+    sha256: "f4b362bcb046bc53fbe2e578662b939222d0c44b96449101829e73ecce02bcb3"
+  "1.6.0":
+    url: "https://github.com/ginkgo-project/ginkgo/archive/v1.6.0.tar.gz"
+    sha256: "c59f574090eac212d272c57f0858db0c13ef41ce5547e38331a74c85ef7fba99"
+  "1.5.0":
+    url: "https://github.com/ginkgo-project/ginkgo/archive/v1.5.0.tar.gz"
+    sha256: "9b24ce0a17ebdf08aa65e87af2d81d6d1e2c94ce317bb26cd32ce47dea8b4bad"
```
uilianries, line 10: Do you really need all these versions? I would suggest only adding the latest in case you don't need these older.

recipes/ginkgo/all/conanfile.py
```diff
@@ -109,7 +120,12 @@ def build(self):
         cmake.build()
 
     def package(self):
-        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        copy(
```
uilianries, line 123: Please, avoid changing the code format. It's not really mandatory following pep8, so it would cause only more code diff to review.

uilianries (commented):
Thank you for PR!

It seems like since version 1.6.0 some things changed and you will need to update some parts:

- Now it needs CMake >= 3.16, so you need to add tool_requires:

```python
def build_requirements(self):
    tool_requires(self, "cmake/[3.16 <4]")
```

- There is an error when building with Clang 12: `timer.cpp:127:12: error: no matching function for call to 'duration_cast'`

My guess is the upstream's CMakeLists.txt did not configure the required C++ standard. Need to verify it, and if confirmed, could you please open an issue the upstream reporting the problem? Otherwise, could be something the Conan recipe.

upsj:
> There is an error when building with Clang 12: timer.cpp:127:12: error: no matching function for call to 'duration_cast'

The culprint is libc++, I might need to disable it for now until we can patch it.

recipes/ginkgo/all/conanfile.py
upsj: I will maintain this package for the forseeable future, and my editor will try to format this every time I make a change, so not formatting would make my job much harder.

recipes/ginkgo/all/conandata.yml
upsj: no, I removed them

upsj:
Not sure what to make of the warnings, I'm kind of baffled a basic system library like `libm` needs to be listed.

uilianries:
> Not sure what to make of the warnings, I'm kind of baffled a basic system library like `libm` needs to be listed.

Don't worry about. The CI runs a `ldd` to check for system libraries that not listed in the `cpp_info.system_libs`. Sometimes, it causes false-positives.

recipes/ginkgo/all/conanfile.py
uilianries: But I'll insist, because formatting will result only in more time consumption for reviewers, is harder to review and will not change any behavior. Your PR only adds a new version, usually, this kind of PR consumes less than 10 lines of code in total. As you are probably using an IDE, it should easily configurable. So please, consider it for future PRs.

recipes/ginkgo/all/conanfile.py
```diff
@@ -74,13 +83,27 @@ def loose_lt_semver(v1, v2):
...
+        is_windows_shared = self.options.shared and is_msvc(self)
+
+        if (
+            is_windows_shared
+            and self.settings.build_type == "Debug"
+            and Version(self.version) >= "1.7.0"
+        ):
+            raise ConanInvalidConfiguration(
+                "Ginkgo >= 1.7.0 cannot be built in shared debug mode on Windows"
+            )
```
uilianries, line 105: What kind of error it produces? Is it something disable by the upstream or only affecting this new version?

uilianries (commented):
Looking good, only a question related to windows restriction.

recipes/ginkgo/all/conanfile.py
upsj:
```
LINK : fatal error LNK1189: library limit of 65535 objects exceeded [C:\J2\w\prod-v1\bsr@2\51653\accad\.conan\data\ginkgo\1.7.0\_\_\build\4f1710918aa542fccb5a54d7bd712e4b0750b50d\build\core\ginkgo.vcxproj]
ginkgo/1.7.0: 
```
see https://github.com/conan-io/conan-center-index/pull/21058#issuecomment-1812003537

recipes/ginkgo/all/conanfile.py
upsj: I'll oblige, but I'm still surprised a project of this size does not have consistent formatting requirements.

recipes/ginkgo/all/conanfile.py
upsj: This is the same approach we take for MSVC in vcpkg, where we only allow static builds altogether.

uilianries (approved):
LGTM

jwillikers (approved):

RubenRBS (approved):
Thanks! I've now restarted the v2 run, it was a transient gh network error

uilianries:
I just rebased this PR, to restart the entire CI.

recipes/ginkgo/all/conanfile.py
```diff
@@ -57,6 +57,9 @@ def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
 
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.16 <4]")
```
SpaceIm, line 61: only since 1.7.0, previous versions must not undergo specificities of other versions

PR #19739, opened 2023-09-11
liburing: Update recipe for version 2.3

thejohnfreeman:
Recent versions of Facebook's folly [depend](https://github.com/facebook/folly/issues/1913) on liburing >= 2.3.

Linux is on kernel major version 6, for which there are no header packages in Conan Center. I was able to build by just removing the requirement.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21061, opened 2023-11-10
liburing: Add 2.3+2.4 and update to Conan 2

Ahajha:
Specify library name and version:  **liburing/2.3**

### DIFF
--- a/recipes/liburing/all/conanfile.py
+++ b/recipes/liburing/all/conanfile.py
@@ -104,7 +104,7 @@ class Liburing(ConanFile):
             # FIXME: The liburing.so symlink ends up broken. Remove and replace.
             symlinks.remove_broken_symlinks(self, self.package_folder)
             source_file_suffix = "1" if self.version < Version("2.0") else "2"
-            os.symlink(src=os.path.join(self.package_folder, "lib", f"liburing.so.{source_file_suffix}"),
+            os.symlink(src=f"liburing.so.{source_file_suffix}",
                        dst=os.path.join(self.package_folder, "lib", "liburing.so"))
 
     def package_info(self):
--- a/recipes/liburing/all/conanfile.py
+++ b/recipes/liburing/all/conanfile.py
@@ -1,5 +1,5 @@
 from conan import ConanFile
-from conan.tools.files import get, rmdir, copy, apply_conandata_patches, chdir, export_conandata_patches, symlinks
+from conan.tools.files import get, rmdir, copy, apply_conandata_patches, chdir, export_conandata_patches, rm
 from conan.tools.scm import Version
 from conan.tools.gnu import Autotools, AutotoolsToolchain
 from conan.tools.layout import basic_layout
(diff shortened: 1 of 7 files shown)


Ahajha:
This is a continuation of #19739

Ahajha:
Locally, still figuring out why the --nolibc option is failing, and shared library installs seem to be broken. Will figure out if this is a regression or not.

Ahajha:
For some reason, with the new changes `liburing.so`, a symlink, ends up broken. For now, I have a fix that just removes it and makes a new symlink.

recipes/liburing/all/conanfile.py
```diff
@@ -60,49 +56,57 @@ def validate(self):
...
-            patch(self, **data)
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = AutotoolsToolchain(self)
+        def whether(condition):
+            return "" if condition else None
+        tc.update_configure_args({
+            "--nolibc": whether(not self.options.get_safe("with_libc", default=True)),
```
franramirez688, line 70: IMO, I believe that it'd be better to simplify it, remove that function which is only used once, and make it explicit:

```suggestion
            "--nolibc": None if self.options.get_safe("with_libc") else "",
```

recipes/liburing/all/conanfile.py
```diff
@@ -60,49 +56,57 @@ def validate(self):
...
-    def _patch_sources(self):
-        for data in self.conan_data.get("patches", {}).get(self.version, []):
-            patch(self, **data)
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = AutotoolsToolchain(self)
+        def whether(condition):
+            return "" if condition else None
```
franramirez688, line 68:
```suggestion
```

recipes/liburing/all/conanfile.py
```diff
@@ -60,49 +56,57 @@ def validate(self):
...
         rmdir(self, os.path.join(self.package_folder, "man"))
 
         if self.options.shared:
             os.remove(os.path.join(self.package_folder, "lib", "liburing.a"))
+        
+            # FIXME: The liburing.so symlink ends up broken. Remove and replace.
+            symlinks.remove_broken_symlinks(self, self.package_folder)
+            source_file_suffix = "1" if self.version < Version("2.0") else "2"
+            os.symlink(src=f"liburing.so.{source_file_suffix}",
+                       dst=os.path.join(self.package_folder, "lib", "liburing.so"))
```
franramirez688, line 108: I think that this could be solved by adding this line in the configure step (`tc.update_configure_args`):

```python
     "--libdevdir": "${prefix}/lib"  # pointing to libdir
```
Let's try that. If it works, we can remove all these lines.

franramirez688 (commented):
Hi @Ahajha

Thanks a lot for the contribution. It's looking great! Several comments before merging the PR 😁

recipes/liburing/all/conanfile.py
Ahajha: Had to change this slightly, the `default=True` is important, since when the option doesn't exist we need to make sure to not pass the flag at all.

recipes/liburing/all/conanfile.py
Ahajha: Done

recipes/liburing/all/conanfile.py
Ahajha: That seems to work, thanks!

franramirez688 (approved):

jwillikers (approved):

PR #21064, opened 2023-11-11
tinyalsa: conan2

Ahajha:
Specify library name and version:  **tinyalsa/all**


Ahajha:
Looking at changes in tinyalsa, 2.0.0 seems to be just bugfixes from 1.1.1. For ease of converting the recipe, it now uses the CMake file provided in 2.0.0. Hopefully the changes are small enough and enough time has passed that anyone using this package would be willing to upgrade.

uilianries (changes requested):
@Ahajha Please, do not remove a major version from `conandata.yml`, it may break users: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#removing-old-versions

Ahajha:
> Please, do not remove a major version from `conandata.yml`, it may break users

I originally thought that the changes between 1.1.1 and 2.0.0 were small, it seems I somehow overlooked a lot of changes. I did feel odd about it, I can go back and re-add 1.1.1, unfortunately I can't use CMake for that version :(

uilianries:
> I originally thought that the changes between 1.1.1 and 2.0.0 were small, it seems I somehow overlooked a lot of changes. I did feel odd about it, I can go back and re-add 1.1.1, unfortunately I can't use CMake for that version :(

@Ahajha you can split the recipe in separated folders like 1.x and 2.x. For real example: https://github.com/conan-io/conan-center-index/tree/master/recipes/mosquitto

Ahajha:
> @Ahajha you can split the recipe in separated folders like 1.x and 2.x. For real example: https://github.com/conan-io/conan-center-index/tree/master/recipes/mosquitto

I have both using the Makefile for now, I'm wondering if the consistency is better or the simplicity/"easy correctness" of CMake for 2.0.0 is better. It's working as is, I think it's probably okay.

recipes/tinyalsa/all/conandata.yml
```diff
@@ -8,4 +8,14 @@ sources:
...
     - patch_file: "patches/tinyalsa-2.0.0-strncpy-warning.patch"
-      base_path: "source_subfolder"
+    - patch_file: "patches/tinyalsa-2.0.0-fix-hardcoded-gcc.patch"
+      patch_description: "Fix hardcoded references to GCC"
+      patch_type: "portability"
+  "1.1.1":
+    - patch_file: "patches/tinyalsa-1.1.1-fix-make-clang-build.patch"
+      patch_description: "fix clang build using make"
+      patch_type: "backport"
+      patch_source: "based on https://github.com/tinyalsa/tinyalsa/commit/1c13f7cbe1ebee24c2b62d5ba16f2702882da49a"
```
uilianries, line 18:
```suggestion
      patch_source: "https://github.com/tinyalsa/tinyalsa/commit/1c13f7cbe1ebee24c2b62d5ba16f2702882da49a"
```


Patch source should the URL only, so we can parse it better. You can add extra details in the PR description.

recipes/tinyalsa/all/conanfile.py
```diff
@@ -1,68 +1,70 @@
...
 class TinyAlsaConan(ConanFile):
     name = "tinyalsa"
+    description = "A small library to interface with ALSA in the Linux kernel"
     license = "BSD-3-Clause"
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/tinyalsa/tinyalsa"
     topics = ("tiny", "alsa", "sound", "audio", "tinyalsa")
-    description = "A small library to interface with ALSA in the Linux kernel"
-    exports_sources = ["patches/*",]
-    options = {"shared": [True, False], "with_utils": [True, False]}
```
uilianries, line 15: Why did you remove `with_utils` options? Is it not available for 1.1? Please, consider not removing an option, it affects directly the package ID: https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#can-i-remove-an-option-from-a-recipe

uilianries (commented):
The patches are okay, as more or less they are integrated already to the upstream and the latest release is from 2021.

recipes/tinyalsa/all/conanfile.py
uilianries: Please, revert that change, people still can decide if they want or not executables installed too.

Ahajha: I'll revert the change, though I'm still confused as to what utility it provides. The executables are tiny (combined total of 160 Kb on my machine). Removing them doesn't save any significant amount of time or space, and keeping them shouldn't break anything, the _only_ possible thing is the environment variables.

RubenRBS (approved):

PR #21079, opened 2023-11-12
nodejs: update recipe for Conan v2

valgur:


tsondergaard:
@valgur, My PR, #22169, fails the CI build because of the following conan-center hook 

```
[HOOK - conan-center.py] pre_export(): ERROR: [NO BUILD SYSTEM FUNCTIONS (KB-H061)] /home/conan/w/prod-v1_cci_PR-22169/recipes/nodejs/all/conanfile.py:47 Build system dependent functions detected. (Use of platform is forbidden in validate) (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H061-NO-BUILD-SYSTEM-FUNCTIONS) 
```

I see that you have moved the platform check to the test package in this PR. Rather than improve my small PR I was wondering if steps could be taken to improve the odds of this PR receiving attention so it could be merged for the benefit of all. I hope I am not too forward here, but do you think it would help if the following changes were made to this PR to make it quick for reviewers to respond to?

* Update PR summary and body from just "." to explain what is fixed in this PR - such as the recipe not working on Fedora and RHEL due to the glibc check
* Interactively rebase branch and create a cleaner history of the changes
* Change the git commit comments to include motivation for each change instead of just saying what was changed

I'm going to drop #22169 as yours is way better.

valgur:
@tsondergaard I referenced your PR in the description (thank you for the detailed description of the problem!). I don't think a git rebase is worthwhile since the changes get squashed anyway on merge and the PR description is generally sufficient to provide context for the changes.

recipes/nodejs/all/conanfile.py
```diff
@@ -1,74 +1,66 @@
...
 
-    @property
-    def _source_subfolder(self):
-        return os.path.join(self.source_folder, "source_subfolder")
+    def layout(self):
+        pass
+
+    def package_id(self):
+        del self.info.settings.compiler
+        del self.info.settings.build_type
```
tsondergaard, line 28: Can you explain to me why it is necessary to provide this `package_id` function?

recipes/nodejs/all/conanfile.py
```diff
@@ -1,74 +1,66 @@
...
-    settings = "os", "arch", "compiler"
+    settings = "os", "arch", "compiler", "build_type"
     no_copy_source = True
     short_paths = True
 
-    @property
-    def _source_subfolder(self):
-        return os.path.join(self.source_folder, "source_subfolder")
+    def layout(self):
+        pass
```
tsondergaard, line 24: Why do you need this no-op definition of `layout()`?

recipes/nodejs/all/conanfile.py
```diff
@@ -1,74 +1,66 @@
...
-        if not self.version in self.conan_data["sources"] or \
-           not str(self.settings.os) in self.conan_data["sources"][self.version] or \
-           not self._nodejs_arch in self.conan_data["sources"][self.version][str(self.settings.os)]:
+        if not self._dl_info:
             raise ConanInvalidConfiguration("Binaries for this combination of architecture/version/os not available")
 
-        if Version(self.version) >= "18.0.0":
-            if str(self.settings.os) == "Linux":
-                if Version(self._glibc_version) < '2.27':
-                    raise ConanInvalidConfiguration("Binaries for this combination of architecture/version/os not available")
```
tsondergaard, line 56: You've moved this check to test_package. What is the reason for this move?

tsondergaard (commented):
I've added some questions to things that I don't understand in this merge request. Most likely the answers will be obvious to people more familiar with conan than me.

recipes/nodejs/all/conanfile.py
valgur: `package_id()` determines which settings and options have an effect on the computed hash that is used as the package ID. The binary package should be compatible irregardless of the C++ compiler and build type used, so these are removed.

recipes/nodejs/all/conanfile.py
valgur: To disable a linter warning / convention.

tsondergaard (approved):

RubenRBS (approved):

jwillikers (approved):

PR #21098, opened 2023-11-13
Update RESTinio to version 0.7.0

ngrodzitski:
Specify library name and version:  **RESTinio/0.7.0**

---

### DIFF
--- a/recipes/restinio/config.yml
+++ b/recipes/restinio/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "0.7.0":
+    folder: "v0.7"
   "0.6.19":
     folder: all
   "0.6.18":
--- a/recipes/restinio/v0.7/conandata.yml
+++ b/recipes/restinio/v0.7/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "0.7.0":
+    url: "https://github.com/Stiffstream/restinio/archive/v.0.7.0.tar.gz"
+    sha256: "1d6ad5812c02ee42342ed0ed78678a4e37c1cfe87f2f7c32c2862f825062feaa"
--- a/recipes/restinio/v0.7/conanfile.py
+++ b/recipes/restinio/v0.7/conanfile.py
@@ -1,0 +1,102 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import get, copy
+from conan.tools.layout import basic_layout
+from conan.tools.scm import Version
+from conan.errors import ConanInvalidConfiguration
+import os
+
+required_conan_version = ">=1.52.0"
+
+
+class RestinioConan(ConanFile):
+    name = "restinio"
+    license = "BSD-3-Clause"
+    homepage = "https://github.com/Stiffstream/restinio"
+    url = "https://github.com/conan-io/conan-center-index"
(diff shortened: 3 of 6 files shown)


RubenRBS (approved):
Usually if the recipes do not diverge too much we usually prefer to keep them as one, this would be in the limit - thanks a lot for taking the time to fix it for the new version, really appreciated! :)

recipes/restinio/all/conanfile.py
```diff
@@ -64,51 +59,44 @@ def package_id(self):
...
+
+    def build(self):
+        pass
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def package(self):
+        copy(self, "*.*pp", src=os.path.join(self.source_folder, "dev", "restinio"), dst=os.path.join(self.package_folder, "include", "restinio"))
         copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
```
RubenRBS, line 94: Is CMAKE.install() not suitable anymore?

recipes/restinio/all/conanfile.py
```diff
@@ -64,51 +59,44 @@ def package_id(self):
...
+            raise ConanInvalidConfiguration("%s requires a compiler that supports at least C++%s" % (self.name, minimal_cpp_standard))
+
+    def build(self):
+        pass
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def package(self):
+        copy(self, "*.*pp", src=os.path.join(self.source_folder, "dev", "restinio"), dst=os.path.join(self.package_folder, "include", "restinio"))
```
RubenRBS, line 93: is this packaging both cpp and hpp files as a header-library?

RubenRBS (commented):
Thanks! I have a question regarding the new package() method, but otherwise looks great!

recipes/restinio/all/conanfile.py
ngrodzitski: I could have use it, yes, but it goes with maintenance costs:
* set the tuning variables right (they can be added/removed and their names/values might be changed in future versions) 
* Conan itself may change the cmake-routines (like v1=>v2).

On the other hand RESTinio takes a promise that a set of headers would be enough - so copying sources is a nice way to go without extra cmake machinery.

recipes/restinio/all/conanfile.py
ngrodzitski: > is this packaging both cpp and hpp files as a header-library?

"*.*pp" is to also include "*.ipp" files.

uilianries:
Thank you for checking this breaking change!

It's a bit hard to review it, because is actually changing the previous, and copying the old version too. My suggestion it only adding a new folder (e.g 0.7) and include it in config.yml, so it will generate less noise and a clean review.

jwillikers (approved):

fdgStilla:
We would be interested in using it with Visual2019 but you limit it to Visual2022 (VS17). VS17 is actually equivalent to msvc193 so it weird to limit to VS17 and msvc191. What is the actual limit for this version?
```
            "Visual Studio": "17",
            "msvc": "191"
```

eao197:
I haven't checked v.0.7.0, but I've checked v.0.7.1 right now and it compiles and passes tests under VS2019.

PR #21111, opened 2023-11-14
xtr: migrate to Conan v2

valgur:
Requires
- #21061

### DIFF
--- a/recipes/xtr/all/conanfile.py
+++ b/recipes/xtr/all/conanfile.py
@@ -87,12 +87,14 @@ class XtrConan(ConanFile):
         if minimum_version and Version(self.settings.compiler.version) < minimum_version:
             raise ConanInvalidConfiguration(f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support.")
 
+        if self.settings.arch not in ["x86_64"]:
+            raise ConanInvalidConfiguration(f"Unsupported arch={self.settings.arch}")
         if Version(self.version) < "2.0.0" and str(self.settings.compiler.libcxx) == "libc++":
-            raise ConanInvalidConfiguration(f"Use at least version 2.0.0 for libc++ compatibility")
+            raise ConanInvalidConfiguration("Use at least version 2.0.0 for libc++ compatibility")
         if self.options.get_safe("enable_io_uring_sqpoll") and not self.options.get_safe("enable_io_uring"):
-            raise ConanInvalidConfiguration(f"io_uring must be enabled if io_uring_sqpoll is enabled")
+            raise ConanInvalidConfiguration("io_uring must be enabled if io_uring_sqpoll is enabled")
         if self.options.get_safe("sink_capacity_kb") and not str(self.options.get_safe("sink_capacity_kb")).isdigit():
-            raise ConanInvalidConfiguration(f"The option 'sink_capacity_kb' must be an integer")
+            raise ConanInvalidConfiguration("The option 'sink_capacity_kb' must be an integer")
 
         if Version(self.dependencies["fmt"].ref.version) < 6:
             raise ConanInvalidConfiguration("The version of fmt must be >= 6.0.0")
--- a/recipes/xtr/all/conanfile.py
+++ b/recipes/xtr/all/conanfile.py
@@ -49,16 +49,10 @@ class XtrConan(ConanFile):
         return {
             "gcc": "10",
             "clang": "12",
-            "apple-clang": "11",
-            "msvc": "192",
-            "Visual Studio": "16",
         }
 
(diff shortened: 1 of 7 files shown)


recipes/xtr/all/conanfile.py
```diff
@@ -1,125 +1,159 @@
...
-        if tools.Version(self.version) < "1.0.1":
-            del self.options.sink_capacity_kb
-        if tools.Version(self.version) < "2.0.0":
+        if Version(self.version) < "2.0.0":
             del self.options.enable_io_uring
             del self.options.enable_io_uring_sqpoll
 
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
```
choll, line 50: Surely you'd want fPIC enabled for a shared library?

recipes/xtr/all/conanfile.py
```diff
@@ -1,125 +1,159 @@
...
+            and self.options.get_safe("enable_io_uring")
+        ):
+            self.requires("liburing/2.4")
 
     def validate(self):
-        if self.settings.os not in ("FreeBSD", "Linux"):
-            raise ConanInvalidConfiguration(f"Unsupported os={self.settings.os}")
-        if self.settings.compiler not in ("gcc", "clang"):
-            raise ConanInvalidConfiguration(f"Unsupported compiler={self.settings.compiler}")
-        if self.settings.arch not in ("x86_64", ):
```
choll, line 54: xtr contains some x86 inline assembly so I think it would be better to keep the arch check

recipes/xtr/all/conanfile.py
```diff
@@ -1,125 +1,159 @@
...
-        self.requires("fmt/7.1.3")
+        # INFO: https://github.com/choll/xtr/blob/2.1.0/include/xtr/detail/buffer.hpp#L27
+        self.requires("fmt/10.1.1", transitive_headers=True, transitive_libs=True)
         # Require liburing on any Linux system as a run-time check will be
         # done to detect if the host kernel supports io_uring.
-        if tools.Version(self.version) >= "2.0.0" and self.settings.os == "Linux" and self.options.get_safe("enable_io_uring"):
-            self.requires("liburing/2.1")
+        if (
+            Version(self.version) >= "2.0.0"
+            and self.settings.os in ["Linux", "FreeBSD"]
```
choll, line 62: FreeBSD doesn't have io_uring AFAIK

recipes/xtr/all/conanfile.py
valgur: Yup, this disables the fPIC option, making it non-optional. It's standard recipe boilerplate on CCI: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L61-L67

recipes/xtr/all/conanfile.py
choll: Yeah. It won't work on macOS or Windows either, FYI. It might work on other varieties of Unix but is only supported/tested on Linux and FreeBSD at the moment.

choll: It's just x86_64 at the moment

valgur: Ok, thanks! And thanks for taking the time to review the recipe as well!

choll: np, thanks for updating the recipe. Are you using xtr?

recipes/xtr/config.yml
```diff
@@ -3,9 +3,5 @@ versions:
     folder: all
   "2.0.1":
     folder: all
-  "2.0.0":
-    folder: all
   "1.0.1":
     folder: all
-  "1.0.0":
-    folder: all
```
choll, line 11: What's the reason for deleting versions? I think it would be better to leave them in, to avoid breaking anyone's build.

valgur: This does not really directly affect the availability of these versions, they simply become "unmaintained". Updates to the recipe are published as new revisions, all old revisions are kept intact.

The standard practice on CCI is to keep all major versions, if possible, and a few latest minor versions. Only the latest patch version is typically kept around. This is done to limit the size of the CI builds, mostly, and to reduce maintenance burden if the recipes require a lot of patching.

https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#supported-versions

choll (approved):

choll:
@uilianries @RubenRBS are you able to review? thanks

uilianries (approved):
LGTM

jwillikers (approved):

PR #21130, opened 2023-11-14
pdfium: migrate to Conan v2

valgur:
Bumped the version to the newest one supported by `pdfium-cmake`.
The existing version had compilation issues in its C++17 compatibility library.

### DIFF
--- a/recipes/pdfium/all/conanfile.py
+++ b/recipes/pdfium/all/conanfile.py
@@ -49,10 +49,11 @@ class PdfiumConan(ConanFile):
         self.requires("icu/74.1")
         self.requires("lcms/2.14")
         self.requires("openjpeg/2.5.0")
+        self.requires("zlib/[>=1.2.11 <2]")
         if self.options.with_libjpeg == "libjpeg":
             self.requires("libjpeg/9e")
         elif self.options.with_libjpeg == "libjpeg-turbo":
-            self.requires("libjpeg-turbo/3.0.0")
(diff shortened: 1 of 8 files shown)


jwillikers (approved):

recipes/pdfium/all/conanfile.py
```diff
@@ -27,84 +33,83 @@ class PdfiumConan(ConanFile):
...
     def validate(self):
         if self.settings.compiler.cppstd:
-            tools.check_min_cppstd(self, 14)
+            check_min_cppstd(self, 14)
         minimum_compiler_versions = {
-            "gcc": 8,
-            "Visual Studio": 15,
+            "gcc": "8",
+            "msvc": "191",
+            "Visual Studio": "15",
```
RubenRBS, line 64: Is there no restriction on clang, is the package not compatible with it (Seems like it is based on build logs) or just missing info?

RubenRBS (approved):

recipes/pdfium/all/conanfile.py
valgur: It built fine with Clang 5.0 in Docker. Conan 2 itself failed to install in the old Clang 4.0 container. I guess leaving it unspecified is fine.

valgur: It built fine with Clang 5.0 in Docker. Conan 2 itself failed to install in the old Clang 4.0 container. I guess leaving it unspecified is fine.

PR #21135, opened 2023-11-14
poppler: migrate to Conan v2

valgur:
Worked locally with all options enabled.


RubenRBS:
Feel free to remove 21.07.0, after looking at it with @uilianries it's going to be faster than the alternative of trying to properly support it :)
Just remember to remove the toolchain variables that would no longer be needed after that!

h-2:
Having this package in Conan2 is one of our main blockers, so work on this is very much appreciated!

Btw, does the package also offer the details headers found in operating system packages like `libpoppler-private-dev` on Ubuntu?
Thanks!

h-2:
> @h-2 I didn't check but I would say yes. Most of Conan packages are equivalent of -dev on Ubuntu.

Yeah, but the *normal* ubuntu dev package is `libpoppler-dev`. The one I asked for is `libpoppler-private-dev` which includes things not normally included. Maybe that would be an extra component?

valgur:
.

valgur:
@user4223 Feel free to open a PR at https://github.com/valgur/conan-center-index.

The Poppler recipe had some pretty difficult linker errors in the CI, which seemed to be caused by the order of the linked libraries. There might even be a dependency cycle somewhere in the deps.

Anyway, it's difficult to move forward with this PR if I can't use the CI to catch the failing configurations.

PR #21137, opened 2023-11-14
gdk-pixbuf: bump deps and several fixes

valgur:
This PR contains several fixes and bumping dependencies. Among them:

* Solved pkg-config variables changes introduced by Conan >= 2.0.8 (https://github.com/conan-io/conan/pull/14233)
* Recovered macOS platform. Workarounds for macOS and `shared=True`:
  *  Avoid generating gdk-pixbuf/loaders.cache (meson `custom_target` is failing due to lack of environment variables).
  * Using Conan libgettext instead of system one. Related comments: https://github.com/conan-io/conan-center-index/pull/16266#issuecomment-1501097433
* Solved static linking with libtiff and libpng (reported upstream https://gitlab.gnome.org/GNOME/gdk-pixbuf/-/merge_requests/159)

Related failing PRs: #21126.

PR #21138, opened 2023-11-14
openslide: new recipe

valgur:
Adds OpenSlide: https://github.com/openslide/openslide

An optional dependency for libvips (https://www.libvips.org/).

[Packaging status](https://repology.org/project/openslide/versions)


recipes/openslide/all/test_package/test_package.c
```diff
@@ -0,0 +1,6 @@
+#include <openslide.h>
+
+int main() {
+    openslide_detect_vendor("");
```
bgilbert, line 4: This doesn't really test anything.  You'd need to do `openslide_open("")` instead, with the `OPENSLIDE_DEBUG` environment variable set to `synthetic`, and you'd need to check whether the open succeeded.  `meson test` will do all of this for you, and I'd recommend using that rather than hand-rolling your own test.  If you want to avoid running tests at build time, you could run `OPENSLIDE_DEBUG=synthetic slidetool prop list ""` instead.

recipes/openslide/all/conanfile.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("cairo/1.18.0")
+        self.requires("gdk-pixbuf/2.42.10")
+        self.requires("glib/2.78.1", force=True)  # FIXME
+        self.requires("libdicom/1.0.5")
+        self.requires("libjpeg/9e")
```
bgilbert, line 51: libjpeg-turbo is faster and widely used.  I'd recommend using it instead.

bgilbert (commented):
Thanks for doing this!

recipes/openslide/all/test_package/test_package.c
valgur: > This doesn't really test anything.

As intended. :smile: `test_package` exists only to confirm that compilation, linking and dynamic linking during execution actually work.

As for unit tests - CCI unfortunately does not allow running actual tests as part of the build since it would be too taxing for the CI given the quite large build matrix.

recipes/openslide/all/conanfile.py
valgur: Would love to, but due to dependency conflicts CCI has to stick to one variant of the library and has settled on `libjpeg` as the default. I personally hope that this is something that could be changed after the Conan 2.0 migration effort is complete.

I added an option for the choice of `libjpeg` variant at least.

valgur:
> Thanks for doing this!

@bgilbert No problem, thanks for dropping by and adding your review!

recipes/openslide/all/conanfile.py
```diff
@@ -58,6 +60,10 @@ def requirements(self):
         self.requires("sqlite3/3.44.0")
         self.requires("zlib/[>=1.2.11 <2]")
 
+    def validate(self):
+        if is_msvc(self):
+            raise ConanInvalidConfiguration(f"OpenSlide requires GNU C++ extensions support and is not compatible with MSVC")
```
bgilbert, line 72: Nit: C, not C++.

recipes/openslide/all/test_package/test_package.c
bgilbert: Ah, got it.  If the rules don't absolutely forbid it, you might still consider running the built-in smoke test though.  Its purpose is exactly to check that each of the dependencies is linked correctly and can read a built-in 16x16 test image.  `meson test` runs _only_ that test right now (most of our tests require a lot of slide data), but if you want to be future-proof, you could avoid `meson test` and directly open the zero-length filename as mentioned above.  On my machine, invoking it via `slidetool prop list ""` takes less than 30 ms of CPU time.

recipes/openslide/all/conanfile.py
valgur: Thanks!

recipes/openslide/all/conanfile.py
```diff
@@ -0,0 +1,131 @@
...
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import copy, get, rm, rmdir
+from conan.tools.gnu import PkgConfigDeps
+from conan.tools.layout import basic_layout
+from conan.tools.meson import Meson, MesonToolchain
+from conan.tools.microsoft import is_msvc
+
+required_conan_version = ">=1.53.0"
```
RubenRBS, line 13: Will need >1.60 <2 || >=2.0.5 fopr this one as it's using the `<host_version>` syntax

recipes/openslide/all/conanfile.py
```diff
@@ -0,0 +1,131 @@
...
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("cairo/1.18.0")
+        self.requires("gdk-pixbuf/2.42.10")
+        self.requires("glib/2.78.1", force=True)  # FIXME
```
RubenRBS, line 55: Why are these needed?

RubenRBS (commented):
Thanks! Rebuilding the missing dependency now

recipes/openslide/all/conanfile.py
valgur: It's blocked by #21137.

valgur:
Blocked by 
- #23143

AbrilRBS (approved):
Thanks a lot for your efforts and patience, this looks great to go now :)

recipes/openslide/all/conanfile.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def package_info(self):
+        self.cpp_info.libs = ["openslide"]
+        self.cpp_info.includedirs.append(os.path.join("include", "openslide"))
+
+        if self.settings.os == "Macos":
+            self.cpp_info.frameworks = ["CoreText"]
+
+        self.cpp_info.requires = [
+            "cairo::cairo_",
```
AbrilRBS, line 120: (Thanks @czoido for spotting this!) Why is the virtual main component used here instead of the whole `cairo::cairo` global one?

valgur: That's done intentionally to match the component used by the project. It also avoids overlinking and unnecessary transitive dependencies, somewhat.

AbrilRBS (approved):

danimtb (approved):

PR #21140, opened 2023-11-14
libtiff: add LERC support, move C++ lib into separate component

valgur:


recipes/libtiff/all/test_package/CMakeLists.txt
```diff
@@ -1,7 +1,11 @@
...
+cmake_minimum_required(VERSION 3.15)
+project(test_package C CXX)
 
 find_package(TIFF REQUIRED)
 
-add_executable(${PROJECT_NAME} test_package.c)
-target_link_libraries(${PROJECT_NAME} PRIVATE TIFF::TIFF)
+add_executable(${PROJECT_NAME}_c test_package.c)
+target_link_libraries(${PROJECT_NAME}_c PRIVATE TIFF::TIFF)
+set_target_properties(${PROJECT_NAME}_c PROPERTIES LINKER_LANGUAGE CXX)
```
SpaceIm, line 8:
```suggestion
```

It shouldn't be needed, otherwise there is an issue.

valgur: Oh, thanks! That's a leftover from debugging. Did not intend to leave it in.

recipes/libtiff/all/conanfile.py
```diff
@@ -37,6 +40,7 @@ class LibtiffConan(ConanFile):
         "lzma": True,
         "jpeg": "libjpeg",
         "zlib": True,
+        "lerc": True,
```
SpaceIm, line 43: honestly it could be disabled by default

recipes/libtiff/all/conanfile.py
```diff
@@ -121,14 +127,16 @@ def generate(self):
             deps.set_property("xz_utils", "cmake_target_name", "liblzma::liblzma")
             deps.set_property("libdeflate", "cmake_file_name", "Deflate")
             deps.set_property("libdeflate", "cmake_target_name", "Deflate::Deflate")
+        deps.set_property("lerc", "cmake_target_name", "LERC::LERC")
```
SpaceIm, line 130: Does it hard fail if with_lerc=False?

valgur: No, it's not strict about the target names.

recipes/libtiff/all/conanfile.py
valgur: Yeah, I it's a quite uncommon compression format. Would allow the PR to pass without waiting for the LERC fix to get merged either.

valgur:
.

PR #21164, opened 2023-11-16
aws-cdi-sdk: migrate to Conan v2

valgur:
Replaced the Makefile in the project with an equivalent CMakeLists.txt: https://github.com/aws/aws-cdi-sdk/blob/v2.4.1/Makefile
That turned out to take less effort than patching around and configuring the Makefile. This should also enable Windows and macOS support for the recipe.


recipes/aws-cdi-sdk/all/conanfile.py
```diff
@@ -1,144 +1,121 @@
...
+        c_sdk.set_property("pkg_config_name", "aws-cdi-sdk")
+        c_sdk.libs = ["cdisdk"]
+        c_sdk.requires = ["aws-cpp-sdk-cdi"]
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            c_sdk.defines = ["_LINUX"]
+        libcxx = stdcpp_library(self)
+        if libcxx:
+            c_sdk.system_libs.append(libcxx)
+
         # TODO: to remove in conan v2 once cmake_find_package_* generators removed
```
perseoGI, line 112: As you are dropping test_v1_package, you could also remove conan v1 related stuff!

valgur: Does Conan v2 automatically add libcxx to the linker flags? Is this mentioned somewhere in the docs as well?

jcar87: @perseoGI  - please remember that in line with the announcement https://github.com/conan-io/conan-center-index/discussions/25461 and previous discussions - we are asking contributors to not remove old logic, even when it seems trivial - literally every PR that is currently open and to be opened could have something like this - and we want to ensure PRs are as minimal as possible to make them easier to review, given the backlog. 

We also don't want to burden contributors with maintenance that we can do ourselves.  - the same goes with `test_v1_package` which we will remove at a later date all at once. 

As a side note: a lot of those "remove once the recipe is conan v2 only" are actually wrong and misleading - I've already seen some recipes were there's lines that _are_ needed for conan 2, right below that comment - so it's not as clear cut as it seems to be - and it's probably better to have automation for this to do it as safely as possible.  Also, I'd say we don't want a contributor to turn a perfectly valid and CI-passing PR into something that is broken because they removed the wrong line on our instruction.

valgur:
.

valgur:
fine, I'll close it soon enough

PR #21166, opened 2023-11-16
Remove missing users from authorized_users.yml

RubenRBS:
Remove usernames that no longer exist in Github


uilianries (approved):
I checked locally, via Github API, and these removed names no longer exist in Github.

jwillikers (approved):

PR #21179, opened 2023-11-17
Add bmulti library

correaa:
Specify library name and version:  **b-multi/0.80.1**

This is an attempt to the add the Multi library to Conan.
https://gitlab.com/correaa/boost-multi"

---


uilianries:
@correaa Thank you for your first contribution!

Please consider the follow points:

- In order to send contributions, you need to sign the CLA commented above: https://github.com/conan-io/conan-center-index/pull/21179#issuecomment-1815808736
- To be able to run PRs in ConanCenterIndex CI service, you first need to request access in the issue #4, as commented above: https://github.com/conan-io/conan-center-index/pull/21179#issuecomment-1815811172
- Your recipe is trying to export local source files, than build a package. This approach is not allowed in ConanCenterIndex, instead, you need to download the source from a server, than, build it. We have some templates that can be used instead of creating a new recipe from zero: https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates

Please, read the contribution guide to get more information: https://github.com/conan-io/conan-center-index/tree/master/docs/adding_packages

correaa:
done. 

Regarding exporting local sources, I don’t understand what it means. 
i believe my submission follows the header only template.
Am I missing something?

correaa:
got it. i had a typo in the filename and i had to remove version number. 

anything else?

correaa:
I see "403 Forbidden" when I try to look at the detail of the failed steps, such as `ci3/conan-v2/pr-merge`.

uilianries:
@correaa The output log is only available by the Bot comment above: https://github.com/conan-io/conan-center-index/pull/21179#issuecomment-1817326326

Those links listed on Gthub Check Status are not available. 


The bot indicated that you were not in the authorized list yet, but you should be listed there now. I'll restart the CI build, it should run good now.

uilianries (commented):
@correaa Thank you for your PR! Please, add a test package that validates your recipe is really packaging your headers.

You dont need to write from scratch, there is a good template that you can re-use: https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates/header_only

recipes/b-multi/all/conanfile.py
```diff
@@ -0,0 +1,28 @@
...
+        "array",
+        "multidimensional",
+        "library",
+    )
+    license = "Boost"
+    url = "https://gitlab.com/correaa/boost-multi"
+    # No settings/options are necessary, this is header only
+    exports_sources = "include/*"
+    no_copy_source = True
+
```
uilianries, line 19:
```suggestion

	@property
    def _min_cppstd(self):
        return 17

    @property
    def _compilers_minimum_version(self):
        return {"gcc": "8",
                "clang": "7",
                "apple-clang": "12",
                "Visual Studio": "16",
                "msvc": "192"}

    def layout(self):
        basic_layout(self, src_folder="src")

    def validate(self):
        if self.settings.compiler.cppstd:
            check_min_cppstd(self, self._min_cppstd)
        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
            raise ConanInvalidConfiguration(
                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
            )

    def source(self):
        get(self, **self.conan_data["sources"][self.version], strip_root=True)

```

- Adjust the source folder to follow Conan folder layout
- Use gitlab url to download the sources
- The project requires C++17 (https://gitlab.com/correaa/boost-multi/-/blob/master/CMakeLists.txt?ref_type=heads#L58)

recipes/b-multi/all/conanfile.py
```diff
@@ -0,0 +1,28 @@
...
+    def package(self):
+        # This will also copy the "include" folder
+        copy(self, "*.hpp", self.source_folder, self.package_folder)
```
uilianries, line 22:
```suggestion
        copy(self, "*.hpp", os.path.join(self.source_folder, "include"), os.path.join(self.package_folder, "include"))
```
Avoid copying unnecessary files in the future, like from test and example folders.

recipes/b-multi/all/conanfile.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def package_info(self):
+        # For header-only packages, libdirs and bindirs are not used
+        # so it's recommended to set those as empty.
```
uilianries, line 26:
```suggestion
```
Comments from templates are not needed here.

recipes/b-multi/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+cmake_minimum_required(VERSION 3.15)
+project(test_package LANGUAGES C) # if the project is pure C
+# project(test_package LANGUAGES CXX) # if the project uses c++
```
uilianries, line 3:
```suggestion
project(test_package LANGUAGES CXX)
```
As your project is C++

recipes/b-multi/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+
+find_package(package REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} test_package.cpp)
+# don't link to ${CONAN_LIBS} or CONAN_PKG::package
+target_link_libraries(${PROJECT_NAME} PRIVATE package::package)
+# In case the target project need a specific C++ standard
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
```
uilianries, line 11:
```suggestion
find_package(multi REQUIRED CONFIG)

add_executable(${PROJECT_NAME} test_package.cpp)
target_link_libraries(${PROJECT_NAME} PRIVATE multi::multi)
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
```

Use the cmake target from the upstream and follow the requires c++ standard listed in the cmakelists in the upstream too.

recipes/b-multi/all/conanfile.py
```diff
@@ -0,0 +1,28 @@
...
+        self.cpp_info.bindirs = []
+        self.cpp_info.libdirs = []
```
uilianries, line 28:
```suggestion
        self.cpp_info.set_property("cmake_file_name", "multi")
        self.cpp_info.set_property("cmake_target_name", "multi::multi")

        self.cpp_info.bindirs = []
        self.cpp_info.libdirs = []
```
use the same cmake targets as specified by the upstream: https://gitlab.com/correaa/boost-multi/-/blob/master/CMakeLists.txt?ref_type=heads#L82

recipes/b-multi/all/test_package/conanfile.py
```diff
@@ -0,0 +1,27 @@
+from conan import ConanFile
+from conan.tools.build import can_run
+from conan.tools.cmake import cmake_layout, CMake
+import os
+
+
+# It will become the standard on Conan 2.x
```
uilianries, line 7:
```suggestion
```
Template comments can be removed.

recipes/b-multi/all/conanfile.py
```diff
@@ -0,0 +1,28 @@
+from conan import ConanFile
+from conan.tools.files import copy
```
uilianries, line 2:
```suggestion
from conan.tools.files import get, copy
from conan.tools.build import check_min_cppstd
from conan.tools.layout import basic_layout
from conan.tools.scm import Version
from conan.errors import ConanInvalidConfiguration
import os
```

To fix other suggestions, you will need to update your imports too.

uilianries (commented):
Thank you for updating your PR. I suggested some important changes.

recipes/b-multi/all/conanfile.py
```diff
@@ -0,0 +1,28 @@
...
+    description = "Multidimensional array access to contiguous or regularly contiguous memory. (Not an official Boost library)"
+    topics = (
+        "array",
+        "multidimensional",
+        "library",
+    )
+    license = "Boost"
+    url = "https://gitlab.com/correaa/boost-multi"
+    # No settings/options are necessary, this is header only
+    exports_sources = "include/*"
```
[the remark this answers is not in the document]
correaa: Is "header-library" smart enough to deduce that the sources are in `include/`? that's great.

recipes/b-multi/all/conanfile.py
correaa: I appreciate the fixes, I couldn't have come up with them myself at this stage.

recipes/b-multi/all/conanfile.py
correaa: I see now where `include/` appears.

recipes/b-multi/all/test_package/CMakeLists.txt
correaa: Wouldn't `cxx_std_17` be inherited from the `multi` package?

recipes/b-multi/all/conanfile.py
correaa: Thank you for referencing back to my own CMakeLists.txt . Please let me know if you have suggestions to improve my CMakeLists.txt to make it more compatible with Conan.

uilianries:
@correaa Hello again, after talking about this PR, we agreed that you can use `boost-multi` as package name. In case you have intention to submit to Boost, and in case accepted in the future, we just mark this recipe as deprecated, so people will use Boost instead.

The thing is, we had problems with boost-xxx packages, but when was an official package distribute, resulting in a collision when someone install boost and the boost-xxx. It's not your case, because your project is not official.

uilianries:
> Am I in the right track? I am talking about the failing test.

No problem, let's rename, then I'll take a look in case of failure. So far, the recipe is looking good.

recipes/boost-multi/all/conanfile.py
```diff
@@ -0,0 +1,59 @@
...
+    homepage = "https://gitlab.com/correaa/boost-multi"
+    description = "Multidimensional array access to contiguous or regularly contiguous memory. (Not an official Boost library)"
+    topics = (
+        "array",
+        "multidimensional",
+        "library",
+    )
+    license = "BSL-1.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    package_type = "header-library"
```
uilianries, line 21:
```suggestion
    package_type = "header-library"
    settings = "os", "arch", "compiler", "build_type"
```

Settings is missing

recipes/boost-multi/all/conanfile.py
```diff
@@ -0,0 +1,59 @@
...
+    def _compilers_minimum_version(self):
+        return {"gcc": "8",
+                "clang": "7",
+                "apple-clang": "12",
+                "Visual Studio": "16",
+                "msvc": "192"}
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
```
uilianries, line 38:
```suggestion

    def package_id(self):
        self.info.clear()

```

So we remove settings from the package ID.

recipes/boost-multi/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,9 @@
+#include "multi/array.hpp"
```
uilianries, line 1:
```suggestion
#include <iostream>
#include "multi/array.hpp"
```

recipes/boost-multi/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "0.80.1":
+    url: https://gitlab.com/correaa/boost-multi/-/archive/v0.80.1/boost-multi-v0.80.1.tar.gz
+    sha256: 7d1ebee123783e0fadfb58a0f55309148d63295fcb4b9c5442d851e3578cad1e
```
uilianries, line 4:
```suggestion
  "0.83.0":
    url: "https://gitlab.com/correaa/boost-multi/-/archive/v0.83.0/boost-multi-v0.83.0.tar.gz"
    sha256: "52324a01e25fd66a32c8bf88f41940ba6364fd218cfab53b7ec5c63115dea1bf"
```

I would recommend updating to 0.83.0, so the LICENSE file will be available.

recipes/boost-multi/config.yml
```diff
@@ -0,0 +1,3 @@
+versions:
+  "0.80.1":
```
uilianries, line 2:
```suggestion
  "0.83.0":
```

correaa:
Is it possible to have latest version of the library? is it not recommended? Ideally, I think the best is to have at least one named (versioned) version and latest (from master branch).

Is that ok?

uilianries:
@correaa You are the author, tell me you 😉 

There is recommendation about version in CCI: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conanfile_attributes.md#version

The recommendation is adding only the latest version, then add old versions in case someone needs and open an issue asking. Each version could generate +100 packages, so is important save some resource from CI when possible.

Consuming directly from master is allowed for those cases when the project is quiet for a long period, or, the project does not provide tags/releases. 

In your case, I would recommend providing the latest version. In case you have something more to be added from master, you could generate a patch version in your project, then use it.

correaa:
So a way to do this would be to always have a version (branch) called latest (e.g. manually maintained in github/gitlab) and add only that one to Conan?

uilianries:
> So a way to do this would be to always have a version (branch) called latest that is quasi in sync with master (e.g. manually maintained in github/gitlab) and add only that one to Conan?

Not mandatory, but preferable. 

> After that what would remain to be done?

Just update `config.yml` and `conandata.yml` with the new release. The CI should re-build the PR automatically.

correaa:
On the contrary, thank you for finishing this PR and sorry for dropping it before.

Is it possible/recommended to install the current master instead or in addition to 0.86?

what is the best way to test in my own CI if the conan package is
working? (I assume that conan doesn't test that the library is well installed, does it?)

uilianries:
@correaa Thank you for your questions. 

> Is it possible/recommended to install the current master instead or in addition to 0.86?

In ConanCenterIndex we prefer always using the released versions, as they are controlled based on tags. Still, there are exceptions for those projects that do not provide a release for a long period or are barely maintained, both are not your case, as your project looks active and healthy. So it would not be possible in ConanCenterIndex.
 
> what is the best way to test in my own CI if the conan package is working?

The recipe in ConanCenterIndex right now for your project is really simple, as it only copies headers and license to the package and only build a small test package. You always can open a new issue asking for a version bump, as you have a new release, or even open a PR directly. You really do not need to predict the recipe for your project, but it's very kind, as most of cases we follow the upstream only.

Still, if you wish to have a recipe pointing to the master branch, you could add a conanfile.py directly to your repository, and it should build the current branch. I could send a PR later to your project if you want.

The Conan documentation shows how to build a Conan recipe when the source code is on the same place: https://docs.conan.io/2/tutorial/creating_packages/create_your_first_package.html

> I assume that conan doesn't test that the library is well installed, does it?

It does, the [test_package](https://docs.conan.io/2/tutorial/creating_packages/test_conan_packages.html) is designed to validate the packaged headers, and libraries and check the linkage. It's not designed to run unit tests or feature validation, as it should be very minimal.

The current test package in this PR imports headers, these are packaged by Conan. The license file, copied from your project, is in the `licenses` folder. We do not test the license file in the test package, but we have a hook in the CI that does it just after packaging the source files. 


I hope I could clarify to you how the packaging project process works in ConanCenterIndex.

correaa:
All this is great. Thank you. 

I will ask if any of my users want to use the library through Conan. 

LGTM, please merge when ready.

correaa:
I though the next step was to merge this. Do I need to do something before that?

correaa:
Please merge if you think it is appropriate.

RazielXYZ (approved):

correaa:
Hi, this library, called Multi, might be soon accepted in Boost. I don't know what I did in the past here (but it caught me in the middle of the transition to Conan 2)

*Can anybody help finishing this an including this library in Conan 2?*

the library has two synchronized repos:

https://gitlab.com/correaa/boost-multi

and (preferred now)

https://github.com/correaa/boost-multi

correaa:
Since I didn't understand what I was going on and didn't understand what was stopping this PR, I started over here: https://github.com/conan-io/conan-center-index/pull/29880

I am more confident in the new version, please let me know how it goes.

PS: The library is now "conditional accepted" in Boost

PR #21153, opened 2023-11-15
isl: migrate to Conan v2

valgur:
Minor additions to #14916. Notably, enabled `imath` support.

- Resolves #14916.
- Resolves #14915.

PR #15128, opened 2023-01-06
[gcc] Model internal libraries

samuel-emrys:
This PR provides an initial model of gcc's internal libraries to allow the recipe to be used as a runtime requirement rather than only as a tool requirement. This is necessary for packages such as lapack, which provides fortran libraries liblapack and libblas, which have a runtime dependency on libgfortran.

I've made an initial attempt to model the relationships between these libraries so that they are included and in the right order in any linker commands. This implementation is likely only reasonably accurate for the shared libraries since `ldd` is available to show the relationships the shared libraries have. The static libraries are somewhat harder to model and has been excluded as out of scope for this pull request, though targets are still provided.

Closes #15099

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

Issue #15099, opened 2023-01-04
[package] gcc recipe doesn't export library metdata


samuel-emrys:
### Description

The current gcc recipe is predominantly targetted at providing executables, however it also provides the following libraries:

* libatomic
* libitm
* libstdc++
* libubsan
* libssp
* libasan
* libtsan
* libgfortran
* liblsan
* libquadmath
* libgomp
* libgcc_s

and a few others that require some more research to determine whether they're worth modelling. This ticket is to model the libraries distributed with gcc as components so that they can be appropriately linked against when required.

An example of when this is required is for an upcoming lapack recipe, which requires linkage against libgfortran. Without this change, it is not possible to prevent this linkage from happening against the system libraries, if they exist at all, which if they don't, would just fail to link.

### Package and Environment Details

NA

### Conan profile

NA

### Steps to reproduce

NA

### Logs

NA

jcar87:
hi @samuel-emrys, thanks for raising this issue.

> An example of when this is required is for an upcoming lapack recipe, which requires linkage against libgfortran. Without this change, it is not possible to prevent this linkage from happening against the system libraries, if they exist at all, which if they don't, would just fail to link.

Would you be able to provide an example where this does not work as expected? If I'm not mistaken, a consumer using gcc from the recipe, if performing linking via gcc itself, should already be able to link against the libgfortran from the package, rather than a system one, without any additional actions. Is this not the case?

samuel-emrys:
Hey @jcar87 yes, linking works fine when invoking the compiler, so there are no problems when building. The problem is modelling the library as a dependency from the consumer perspective, or ensuring that the path to the library is exported appropriately/libraries are linked to in the right order when linking static libs. To be explicit about this, without modelling the libraries within the `gcc` package, one has to add `gfortran` to the `system_libs` list. This will append `-lgfortran` to the link command, and it may resolve to the system version dependening on what the environment is. There are a couple of problems with this:

* This is also semantically inaccurate - the intent here is not to link against the system library, but what's been provided by the `gcc` recipe.
* The consumer shouldn't need to care about libgfortran. This is a dependency of `blas` or `lapack` and should be added automatically when either of those is used.

To illustrate this, by modelling the internal libraries explicitly in the conan recipe, consumer recipes are able to specify the specific dependency:

```python
    def package_info(self):
        self.cpp_info.components["blas"].set_property("cmake_target_name", "BLAS::BLAS")
[...]

PR #21193, opened 2023-11-17
gcc: model exported libraries, better mirrors, more specific license

valgur:
Continues the changes from #15128 to model the installed runtime libraries correctly. The compilers cannot be used reliably unless `self.requires("gcc/")` with `self.cpp_info.requires.append("gcc::gfortran")` is used in addition to `self.tool_requires("gcc/")`.

Set the `package_type` to `library` for the runtime libraries to be usable as `self.requires()`. See https://github.com/conan-io/conan/issues/15306 for more context.

The current URLs for the sources were extremely slow - 10 minutes to download 140 MB. Replaced the .tar.gz with tar.xz for 80 MB instead of 140 MB size and picked the fastest 2 mirrors out of 25 tested. Download takes less than 10 seconds now.

Made the license ID more specific by including the exception for runtimes:
https://spdx.org/licenses/GCC-exception-3.1.html
https://gcc.gnu.org/onlinedocs/libstdc++/manual/license.html

Requires #21153 for Conan 2.0 pipeline to pass.


samuel-emrys:
The changes you've pulled in from my PR in https://github.com/conan-io/conan-center-index/pull/15128 exceed the scope of what you've defined for this PR. Given that there seems to be some reluctance to adopt the internal library model for the gcc recipe, my suggestion would be to remove these changes from this PR if you want it to be accepted.

valgur:
> The changes you've pulled in from my PR in #15128 exceed the scope of what you've defined for this PR. Given that there seems to be some reluctance to adopt the internal library model for the gcc recipe, my suggestion would be to remove these changes from this PR if you want it to be accepted.

@samuel-emrys Yes, I actually care more about that change than the other minor ones, really.

`gfortran` is not really usable via a `self.tool_requires("gcc/")` unless `self.requires("gcc/")` with `self.cpp_info.requires.append("gcc::gfortran")` is also used. I've tried multiple approaches with `self.cpp_info.system_libs` and what not, and this is the only one that really works.

jcar87:
Closing this, as per the reasons already expressed here: https://github.com/conan-io/conan-center-index/issues/15099#issuecomment-1782681777

Currently we don't have a package type that can correctly and comprehensively model libraries within a toolchain. In most circumstances, libraries within a toolchain like GCC don't need to be modelled: the compiler knows where to look for those libraries, and a "relocatable" gcc would know, as a fallback, where to look for libraries, or would already implicitly link them (e.g. like glibc, libstdc++, etc).

It is true that the `gcc` recipe in Conan center is unmaintained - at the moment we are not prioritising this because the vast majority of our users use a GCC provided by their system or SDK. If there are specific use cases that could be satisfied by an improved recipe, please open an issue where this can be discussed and other users can articulate their needs, and we will consider this in the future.

valgur:
@jcar87 Thanks for looking into it. I agree with all of your points.

My real motivation in this PR was to get some kind of Fortran support going on CCI, which is required for quite a few numerical libraries. A full GCC toolchain with C/C++ support is more of a liability in that case anyway as it will interfere with the compiler configured in the profile. That's why I opened #23334. I invite you to take a closer look at it if you get a chance.

PR #21196, opened 2023-11-17
qwt: use a more specific license identifier

valgur:
Adds https://spdx.org/licenses/Qwt-exception-1.0.html

The full license: https://sourceforge.net/p/qwt/git/ci/develop/tree/COPYING


recipes/qwt/all/conanfile.py
```diff
@@ -12,7 +12,7 @@
 
 class QwtConan(ConanFile):
     name = "qwt"
-    license = "LGPL-2.1-or-later"
+    license = "LGPL-2.1+ WITH Qwt-exception-1.0"
```
jwillikers, line 15:
```suggestion
    license = "LGPL-2.1-or-later WITH Qwt-exception-1.0"
```

jwillikers (changes requested):

recipes/qwt/all/conanfile.py
jwillikers: Just that updated SPDX license name for the LGPL here.

jwillikers (approved):

PR #21202, opened 2023-11-17
libvpx: add 1.13.1 + remove shared option and set package_type to static-library if Windows instead of raising for shared=True

SpaceIm:
`libvpx` is a dependency of `ffmpeg` (optional but enabled by default), therefore v2 pipeline can't build `ffmpeg` shared for msvc because `libvpx` with `shared=True` raises for Windows. And as a consequence, it also prevents v2 pipeline to build shared flavor of downstream packages of `ffmpeg`, like `opencv`.
Therefore this PR removes shared option of `libvpx` if Windows, so that v2 pipeline can build downstream packages with `-o *:shared=True`.

It's the same rational as https://github.com/conan-io/conan-center-index/pull/17228 or https://github.com/conan-io/conan-center-index/pull/21144

---

### DIFF
--- a/recipes/libvpx/all/conandata.yml
+++ b/recipes/libvpx/all/conandata.yml
@@ -26,6 +26,7 @@ patches:
     - patch_file: "patches/0001-extended-support-1.9.0.patch"
       patch_type: "portability"
       patch_description: "Add support for more compilers"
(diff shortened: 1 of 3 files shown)


valgur (approved):

SpaceIm:
Actually I see that it msys2 is able to build shared lib, and I've found some logic about shared msvc, so there might be a way.

jwillikers (approved):

uilianries (approved):
LGTM

PR #21238, opened 2023-11-19
brunsli: allow shared brotli dependency

valgur:


valgur:
The project itself hard-codes the builds to be static, but I patched around that in the original PR and forgot to re-enable it in the recipe.

jwillikers (approved):

valgur:
@RubenRBS Ping.

ericLemanissier (approved):

AbrilRBS (approved):

AbrilRBS:
My main fear with this is that we could become maintainers of "using shared brotli" if upstream does not really wants to support that - I'm initially in favour, but let me quickly double check with the team, sorry for the delay!

jcar87:
Seems like the upstream project allows shared since version 1.1.0: https://github.com/google/brotli/commit/641bec0e30bea648b3da1cd90fc6b44deb429f71 
(unsure how that correlates with the conan center index).

in general please add context to the PR description when removing these checks - they are (usually) there for a very valid reason, and without additional context we don't have information to determine whether we risk bringing back old problems or not.

PR #21247, opened 2023-11-19
libhal: Correct 2.0.1 dependency on conan center

kammce:
Specify library name and version:  **libhal/2.0.0**

Boost does not work as a dependency for baremetal platforms. Boost-leaf must be used for `libhal/[^2.0.0]` 

---

### DIFF
--- a/recipes/libhal/all/conandata.yml
+++ b/recipes/libhal/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
   "2.0.1":
     url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.1.tar.gz"
-    sha256: "TBD"
+    sha256: "b3667b8333f18ff61c394a846c7220d5d692097db4b2e1ca02b0cfb392be8469"
--- a/recipes/libhal/all/conandata.yml
+++ b/recipes/libhal/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
   "2.0.1":
     url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.1.tar.gz"
-    sha256: "ba05f9d2172f7afbf816ab18f2dcd7c8cc4452df2b268a832e995a0321c1d2f9"
+    sha256: "TBD"
--- a/recipes/libhal/all/conandata.yml
+++ b/recipes/libhal/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
   "2.0.1":
     url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.1.tar.gz"
-    sha256: "b3667b8333f18ff61c394a846c7220d5d692097db4b2e1ca02b0cfb392be8469"
+    sha256: "ba05f9d2172f7afbf816ab18f2dcd7c8cc4452df2b268a832e995a0321c1d2f9"
+  "2.0.2":
+    url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.2.tar.gz"
+    sha256: "bb69fffbff58ac9a91f71636422d81a4426fe70c3b47ca9b07c87fb074c989dc"
--- a/recipes/libhal/all/conandata.yml
+++ b/recipes/libhal/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "2.0.1":
-    url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.1.tar.gz"
-    sha256: "ba05f9d2172f7afbf816ab18f2dcd7c8cc4452df2b268a832e995a0321c1d2f9"
   "2.0.2":
     url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.2.tar.gz"
     sha256: "bb69fffbff58ac9a91f71636422d81a4426fe70c3b47ca9b07c87fb074c989dc"
+  "2.0.1":
+    url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.1.tar.gz"
+    sha256: "ba05f9d2172f7afbf816ab18f2dcd7c8cc4452df2b268a832e995a0321c1d2f9"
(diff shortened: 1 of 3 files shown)


jwillikers (approved):

kammce:
Actually hold off on pulling this PR in. I'm going to change the dependency to just boost header only to make this easier.

kammce:
Okay this can be submitted now. So long as CI works.

RubenRBS:
Hi @kammce quick question: What was the cause behind the sha256 change?

kammce:
> Hi @kammce quick question: What was the cause behind the sha256 change?

This is the commit that I used to patch the 2.0.1 version https://github.com/libhal/libhal/commit/d33fb4e31a7bc84e2036cf0e81b3ed571a3cac34.

kammce:
Rather than change the version, I added a new one and recovered the old one.

recipes/libhal/all/conanfile.py
```diff
@@ -36,14 +36,26 @@ def _compilers_minimum_version(self):
...
     def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.83.0", transitive_headers=True)
+        # NOTE: from the author, kammce, to the conan team. boost-leaf is tied
+        # to the 2.0.1 and must be kept for it to work correctly.
+        version = Version(self.version)
+        if version == "2.0.1":
+            self.requires("boost-leaf/1.81.0",)
```
uilianries, line 54: The `boost-leaf` package is deprecated and should be using `boost` instead. The version 1.83.0 (and older) has leaf integrated: https://www.boost.org/doc/libs/1_83_0/libs/leaf/doc/html/index.html Is it not enough?

recipes/libhal/all/conanfile.py
```diff
@@ -36,14 +36,26 @@ def _compilers_minimum_version(self):
...
+        elif version == "2.0.2":
+            self.requires("boost/1.83.0",
+                          transitive_headers=True,
+                          options={ "header_only": True })
```
uilianries, line 58: The need of header_only, should be part of `validate()` too, because any user is able to override it by command line:

```
def validate(self):
    if not self.dependencies["boost"].options.header_only:
        raise ConanInvalidConfiguration(f"{self.ref} requires boost/*:header_only=True due boost::leaf.")
```

recipes/libhal/all/conanfile.py
```diff
@@ -74,18 +86,27 @@ def build(self):
...
         self.cpp_info.bindirs = []
+        self.cpp_info.frameworkdirs = []
         self.cpp_info.libdirs = []
+        self.cpp_info.resdirs = []
+
+        version = Version(self.version)
+        if self._bare_metal and version < "3.0.0":
+            self.cpp_info.defines = [
+                "BOOST_LEAF_EMBEDDED",
+                "BOOST_LEAF_NO_THREADS"
```
uilianries, line 111: Are these defines really provided by libhal, or actually Boost should export it in case of using baremetal?

recipes/libhal/all/conanfile.py
kammce: No because packages 2.0.1 and below use the single header file in its includes. Boost does not include the single header file. So to keep the software the same and building, we must use the deprecated Boost.LEAF package. Or opt to remove the 2.0.1 package from the index. Or changing 2.0.1 to accommodate the boost header only. I think using boost leaf is the better alternative here.

recipes/libhal/all/conanfile.py
kammce: These are provided by libhal for versions below 3.0.0. Not Boost.

kammce: Also, Boost could emit the BOOST_LEAF_EMBEDDED if the OS is baremetal, but the NO_THREADS flag is one of 3 options. Another one is FREERTOS and another allows a hook for TLS support for RTOSs not currently supported directly. So this would have to be another option for the boost library. Note that I plan to drop Boost.LEAF and boost after 3.0.0 in favor of C++ exceptions, so there won't be new dependencies on Boost.LEAF in the future.

recipes/libhal/all/conanfile.py
kammce: Done

recipes/libhal/all/conandata.yml
```diff
@@ -2,3 +2,6 @@ sources:
   "2.0.1":
     url: "https://github.com/libhal/libhal/archive/refs/tags/2.0.1.tar.gz"
     sha256: "ba05f9d2172f7afbf816ab18f2dcd7c8cc4452df2b268a832e995a0321c1d2f9"
+  "2.0.2":
```
jwillikers, line 5: Could you put the newer version above? We try to keep the versions ordered in that way.

recipes/libhal/config.yml
```diff
@@ -1,3 +1,5 @@
 versions:
   "2.0.1":
     folder: "all"
+  "2.0.2":
```
jwillikers, line 4: Same thing about ordering the versions, but looks good otherwise.

jwillikers (approved):

recipes/libhal/all/conanfile.py
uilianries: Unfortunately, using boost-leaf package is not an option, it's deprecated already. We just didn't remove it due CCI policy to avoid breaking users. 

Let's drop 2.0.1 then, it will consume less maintenance and will "enforce" users to use the latest version of libhal instead.

jwillikers (approved):

recipes/libhal/all/conanfile.py
kammce: I thought that was against CCI guidelines to drop packages? I'm all for it though.

uilianries: @kammce we actually don't remove the package from Conan Center, but the version from conandata.yml, so we no longer build new revisions: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#removing-old-versions

uilianries: @kammce after talking with @RubenRBS, is clear supporting boost-leaf for your recipe is the best case, I didn't understand that libhal was not designed to support Boost itself. So, boost-leaf is a risk that we can assume, but need to be clear for users, with a `self.output.warning` message. Rubén should talk to you via Slack to align better. Thank you for bringing more details about your case.

RubenRBS (approved):
Thanks a lot for your patience <3

PR #21268, opened 2023-11-20
cppserver: remove legacy conan v1 output.warn which breaks conan v2 with "unknown compilers"

SpaceIm:

Maintainer edits:
* `cppcommon/*`: correctly reflect that fmt is a `transitive_libs=True` dependency - including certain headers from cppcommon (as included in cppserver) cause the consumer to reference fmt symbols
* stop maintaining older versions of cppserver 
* assume `compiler.cppstd` is defined and use `check_min_cppstd()`
* reflect transitive headers and libs for dependencies in `cppserver`
* remove `test_v1_package`


jcar87 (approved):
Thanks!

PR #21271, opened 2023-11-20
arbiter: new recipe

valgur:
Adds arbiter: https://github.com/connormanning/arbiter

Arbiter provides simple/fast/thread-safe C++ access to filesystem, HTTP, S3, and Dropbox resources in a uniform way. It is designed to be extendible, so other resource types may also be abstracted.

Adding it mostly to unvendor it in the PDAL recipe.
The CMake setup in the project was not great. I found it easier to rewrite it entirely instead of patching.

[Packaging status](https://repology.org/project/arbiter/versions)

### DIFF
--- a/recipes/arbiter/all/conanfile.py
+++ b/recipes/arbiter/all/conanfile.py
@@ -1,0 +1,105 @@
+import os
+
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, rm, export_conandata_patches, apply_conandata_patches, rmdir, save
+
+required_conan_version = ">=1.53.0"
+
+
+class PackageConan(ConanFile):
+    name = "arbiter"
+    description = "Uniform access to the filesystem, HTTP, S3, GCS, Dropbox, etc."
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/connormanning/arbiter"
+    topics = ("filesystem", "http", "s3", "gcs", "dropbox")
+
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 11
+
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", self.recipe_folder, self.export_sources_folder)
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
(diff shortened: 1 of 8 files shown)


RubenRBS:
@valgur just a heads-up. If you see that I've assigned myself to any of your PRs that are failing and then you get the PR to build, feel free to ping me so I can go back to the PR sooner than if you just wait for me to make the rounds! :)

recipes/arbiter/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("libcurl/[>=7.78.0 <9]")
+        self.requires("openssl/[>=1.1 <4]")
+        self.requires("zlib/[>=1.2.11 <2]")
+        self.requires("nlohmann_json/3.11.2", transitive_headers=True)
```
trns1997, line 52: Could we upgrade to the below version to prevent dependency mismatch error
```
diff --git a/recipes/arbiter/all/conanfile.py b/recipes/arbiter/all/conanfile.py
index 5045c10f6..eb44c83d5 100644
--- a/recipes/arbiter/all/conanfile.py
+++ b/recipes/arbiter/all/conanfile.py
@@ -49,7 +49,7 @@ class PackageConan(ConanFile):
         self.requires("libcurl/[>=7.78.0 <9]")
-        self.requires("nlohmann_json/3.11.2", transitive_headers=True)
+        self.requires("nlohmann_json/3.11.3", transitive_headers=True)
         self.requires("rapidxml/1.13", transitive_headers=True)
 
     def validate(self):

```

trns1997: @valgur any updates regarding this?

RubenRBS, line 52:
```suggestion
        self.requires("nlohmann_json/3.11.3", transitive_headers=True)
```

trns1997 (approved):

trns1997:
@RubenRBS @valgur it would be great to merge this along with gdal (https://github.com/conan-io/conan-center-index/pull/19298) that way we can patch up pdal :)

uilianries:
I'll send the cmake to the upstream and try to keep a separated option for vendors.

valgur:
@uilianries Can we perhaps merge this as is and open another PR once upstream CMakeLists.txt has been updated? It's blocking #21296.

uilianries:
@valgur I'll still prefer patching than pushing a new CMakeList.txt as we can miss some behavior with upstream's cmake. But don't worry, I'll send you a PR with the patch.

uilianries:
@valgur Thank you for your PR! I sent yesterday a PR to the upstream: https://github.com/connormanning/arbiter/pull/51

The author commented about Arbiter be a vendorized project for PDAL and Entwine only, not a distributed module as we usually package: https://github.com/connormanning/arbiter/pull/51#issuecomment-1957742813

So, packaging it will demand an effort for a library that will not be consumed as a separated package, only will be used by PDAL. You needed to add an extra CMakeLists to make it des-vendorized, which cost even more maintenance from us.

After talking to Conan team about this situation, we see as keeping Arbiter vendorized in PDAL will be much simpler and should not create trouble for us. In case someone opens an issue asking for this project as a Conan package, we can re-visit this topic, but for now I'm closing this PR, so we can focus in PDAL only.

PR #21274, opened 2023-11-20
stlab: several fixes

SpaceIm:
- Remove legacy conan v1 `self.output.warn`, otherwise it fails for "unknown compilers
- Remove noisy messages in `configure()` (these messages are already displayed by CMake during build, and messages in `configure()` are really noisy in a `conan install`)
- Add upper bound to cmake version range in build requirements
- Remove `settings.build_type` from package id. I think `settings.arch` could also be removed by I'm not sure.
- Set `cpp_info.bindirs` & `cpp_info.libdirs` to empty list

---

### DIFF
--- a/recipes/stlab/all/conanfile.py
+++ b/recipes/stlab/all/conanfile.py
@@ -62,7 +62,7 @@ class Stlab(ConanFile):
 
     def requirements(self):
         if self.options.with_boost:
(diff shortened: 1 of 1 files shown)


SpaceIm:
@uilianries is it possible to fix hook KB-H014? It doesn't check for package_type.

PR #19882, opened 2023-09-16
osmanip: add package_type + modernize a little bit more

SpaceIm:
- add package_type
- use self.settings/self.compiler in validate() instead of self.info
- consistent value between Visual Studio & msvc in _compiler_required_cpp17()
- bump boost & arsenalgear

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #16264, opened 2023-02-24
arsenalgear: add version 2.1.0

toge:
Specify library name and version:  **arsenalgear/2.1.0**

- add version 2.1.0
  - 2.1.0 provides official CMakeLists.txt

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21282, opened 2023-11-20
arsenalgear: restore shared option + allow msvc since 2.1.0 + fix CMakeLists of 2.1.0 to avoid building and packaging doctest

SpaceIm:
- Restore shared option, it has been removed by https://github.com/conan-io/conan-center-index/pull/16264, but I think it was a mistake. Moreover removing shared option leads to missing arsenalgear packages for osmanip (see https://github.com/conan-io/conan-center-index/pull/19882#issuecomment-1819891094 & https://github.com/conan-io/conan-center-index/issues/21283)
- Allow msvc since 2.1.0, it seems to build fine (I've tested with 2.0.1, and it didn't pass, so only since 2.1.0)
- We use upstream CMakeLists since 2.1.0, but it has several issues (build & download doctest, pollutes install dir with doctest headers, doesn't honor compiler.cppstd, don't allow to build shared while it works just fine). So I've added a patch and submitted a PR upstream: https://github.com/JustWhit3/arsenalgear-cpp/pull/4
- I've also added cmake names explicitly since a CMake config file is installed upstream since 2.1.0, see:
  https://github.com/JustWhit3/arsenalgear-cpp/blob/v2.1.0/CMakeLists.txt#L30
  https://github.com/JustWhit3/arsenalgear-cpp/blob/v2.1.0/CMakeLists.txt#L66
  https://github.com/JustWhit3/arsenalgear-cpp/blob/v2.1.0/CMakeLists.txt#L87-L91

---

### DIFF
--- a/recipes/arsenalgear/all/conandata.yml
+++ b/recipes/arsenalgear/all/conandata.yml
@@ -11,5 +11,5 @@ sources:
 patches:
   "2.1.0":
     - patch_file: "patches/2.1.0-0001-fix-cmake.patch"
-      patch_description: "CMake: honor compiler.cppstd, and avoid to download and build doctest"
+      patch_description: "CMake: allow shared, honor compiler.cppstd, avoid to download and build doctest"
       patch_type: "conan"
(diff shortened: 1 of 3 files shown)


SpaceIm:
I've submitted https://github.com/JustWhit3/arsenalgear-cpp/pull/4 upstream.

jwillikers (approved):

uilianries (approved):
LGTM

PR #21303, opened 2023-11-21
boost-wintls: Add new recipe

Nerixyz:
Specify library name and version:  **boost-wintls/0.9.8**

I'm adding [boost-wintls](https://github.com/laudrup/boost-wintls) which is a native Windows TLS wrapper for `boost::asio`, similar to [certify](http://github.com/djarek/certify) (already on conan). In fact, `boost-wintls` is the preferred TLS wrapper as djarek, the author of certify, [explains here](https://github.com/microsoft/vcpkg/pull/16783#issuecomment-941705540).

I've overwritten the default CMake target name to ease compatibility with vcpkg and git-submodule setups, where the target is called `boost-wintls`. On vcpkg, the package is called `bext-wintls` since they provide each boost module as `boost-<module>` and this isn't an official boost module, but I think it's better to call it `boost-wintls` here.

It only makes sense to build the package on Windows, but I'm not sure how to encode that other than through the check in `validate`.

Technically, this package only depends on `Boost::headers` (for ASIO) should this be encoded in `requirements` and if so, how?

---

### DIFF
--- a/recipes/boost-wintls/all/conanfile.py
+++ b/recipes/boost-wintls/all/conanfile.py
@@ -1,0 +1,111 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.layout import basic_layout
+from conan.tools.files import (
+    apply_conandata_patches,
+    copy,
+    export_conandata_patches,
+    get,
+)
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class BoostWinTLS(ConanFile):
+    name = "boost-wintls"
+    description = "Native Windows TLS stream wrapper for use with boost::asio"
+    license = "BSL-1.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://wintls.dev"
+    topics = (
+        "header-only",
+        "windows",
+        "tls",
+        "ssl",
+        "networking",
+        "boost",
+        "asio",
+        "sspi",
+        "schannel",
+    )
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {}
+    default_options = {}
+
+    @property
+    def _min_cppstd(self):
+        return 14
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "clang": "12",
+            "gcc": "7",
+            "msvc": "192",
+            "Visual Studio": "16",
+        }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("boost/[>=1.75 <2.0]")
+
+    def package_id(self):
+        self.info.clear()
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(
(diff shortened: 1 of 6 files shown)


uilianries (changes requested):
Thank you for your contribution and intention of adding this project to Conan Center Index.

Unfortunately, we can not accept it due CCI policy: https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#can-we-add-package-which-are-parts-of-bigger-projects-like-boost

We had past incidents with other project that are part of boost now, like duplication in the dependencies graph.

uilianries:
What can be done is following the convention: https://github.com/conan-io/conan-center-index/blob/master/docs/faqs.md#can-i-add-my-project-which-i-will-submit-to-boost

So, this project should be named as `laudrup-wintls`.

Nerixyz:
> So, this project should be named as `laudrup-wintls`.

Should the CMake target be renamed as well?

uilianries:
@Nerixyz I talked to Conan team about this PR and after checking the upstream, everything uses uses the Boost name, including headers and Cmake targets. However, as boost-wintls is not part of official Boost, it should be a problem accepting using the namespace boost. We had past problem with boost-leaf, which is in Boost too.

So please, rename (again) the folder and package name to `boost-wintls`.

I'll update the FAQs be to clear about it. Sorry the confusion.

uilianries (commented):
Thank you for your PR! Here are some suggestions.

recipes/boost-wintls/all/conanfile.py
```diff
@@ -0,0 +1,91 @@
...
+    settings = "os", "arch", "compiler", "build_type"
+
+    @property
+    def _min_cppstd(self):
+        return 14
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "clang": "12",
```
uilianries, line 40:
```suggestion
            "clang": "7",
            "apple-clang": "10",
```

The Clang 7 supports C++14 already: https://en.cppreference.com/w/cpp/compiler_support/14

Apple-Clang 10 supports C++14 too: https://developer.apple.com/xcode/cpp/#c++14

valgur (changes requested):
LGTM, but the minimum Clang versions should probably be lowered, as mentioned by uilianries.

Nerixyz:
The project was renamed to `wintls` (at least the headers and CMake) and got support for standalone ASIO. I'll wait for the next release and add the proper recipe then. How would I model support for both Boost.Asio and standalone ASIO?

valgur:
@Nerixyz You can add add an option `"asio": ["boost", "standalone"]` and then select the package accordingly in `self.requirements()`, perhaps?

uilianries:
@Nerixyz Hello! I just pushed some changes to this PR, including:

- Bump the WinTLS version to 0.9.9
- Requiring Conan 2.1 as the minimum version.
- Add option for asio backend
- Export defines when using standalone asio instead of boost
- Simplified the test package to work requiring only WinTLS, not its backend.


Using [replace_requires](https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires) would be another option instead of adding a new option for asio backend, but I still see it as a valid approach, because it needs to export a compiler definition and point boost::headers and using boost as backend.

Please, feel free to comment about my changes and give your feedback. Regards.

uilianries (approved):
LGTM

Nerixyz (commented):
Thank you so much for taking the time and updating this. I completely forgot about this PR (probably because I was waiting on TLS 1.3). Using `boost-wintls` as the name for the recipe is fine imo.

PR #21305, opened 2023-11-22
jinja2cpp: add new version 1.3.2

toge:
Specify library name and version:  **jinja2cpp/1.3.2**

jinja2cpp/1.3.2 requires Boost::json.
But Boost::json is not avaiable in apple-clang.
https://github.com/conan-io/conan-center-index/pull/5246/files#r616257192

In this PR, I will drop apple-clang support for jinja2cpp/1.3.2.
I'm trying to relax apple-clang condition for boost.

---


recipes/jinja2cpp/all/patches/fix-cmake-1.3.1.patch
```diff
@@ -0,0 +1,54 @@
...
+     struct SelectMemberPtr;
+diff --git a/thirdparty/CMakeLists.txt b/thirdparty/CMakeLists.txt
+index fc8f51d..49540d3 100644
+--- a/thirdparty/CMakeLists.txt
++++ b/thirdparty/CMakeLists.txt
+@@ -70,8 +70,8 @@ endif()
+ 
+ if (NOT DEFINED JINJA2_PRIVATE_LIBS_INT)
+     set(JINJA2CPP_PRIVATE_LIBS ${JINJA2CPP_PRIVATE_LIBS}
+-        Boost::variant Boost::filesystem Boost::algorithm Boost::lexical_cast Boost::json
```
rmorozov, line 48: @toge , would you mind telling me why we removing boost::regex, boost::json?

boost::regex seems to significantly speed ups template loading?

toge: @rmorozov 
Oh, it's my mistake.
Thank you  for your review!

rmorozov: I have added special option to choose which regex to use - default one is boost regex, but if it is not suitable feel free to override and use std regex as it was earlier

https://github.com/jinja2cpp/Jinja2Cpp/blob/4393211baa8ba69437d070be3c9fcba59aad6793/CMakeLists.txt#L15

recipes/jinja2cpp/all/conanfile.py
```diff
@@ -76,6 +76,12 @@ def validate(self):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
+        if Version(self.version) >= "1.3.1" and self.settings.compiler == "apple-clang":
```
rmorozov, line 79: I have a special switch for this case
https://github.com/jinja2cpp/Jinja2Cpp/blob/master/CMakeLists.txt#L18

for apple-clang there can be -DJINJA2CPP_WITH_JSON_BINDINGS=rapid, which doesn't require boost::json

toge: @rmorozov 
Thanks a lot!
I will fix this issue.

toge: @rmorozov 
I tried to compile jinja2cpp/1.3.1 with apple-clang with INJA2CPP_WITH_JSON_BINDINGS=rapid.
But I get the same compile error due to Boost::json.
This is because CMakeLists.txt in jinja2cpp/1.3.1 tries to link Boost::json despite the value of INJA2CPP_WITH_JSON_BINDINGS.
I will try to make jinja2cpp/1.3.1 not compatible with apple-clang in this PR, because I think it is necessary to fix CMakeLists.txt or conan's boost recipe.

toge:
@RubenRBS 
Cloud you review it?

recipes/jinja2cpp/all/patches/fix-cmake-1.3.1.patch
```diff
@@ -0,0 +1,52 @@
...
+     template<typename CharT>
+     constexpr auto GetValueStr() const
+     {
+         constexpr auto memPtr = SelectMemberPtr<CharT, &MultiStringLiteral::charValue, &MultiStringLiteral::wcharValue>::GetPtr();
+         return std::basic_string<CharT>(this->*memPtr);
+     }
++#endif
+ 
+     template<typename CharT, const char* MultiStringLiteral::*, const wchar_t* MultiStringLiteral::*>
+     struct SelectMemberPtr;
```
valgur, line 39: This should be a separate patch. Submitted upstream as well, if possible.

valgur (changes requested):

RubenRBS (commented):
Sorry for the delay @toge, I'll get to this review early next monday, this PR fell thru the cracks and I just now found it again :)

uilianries (commented):
Hello and thank you for bumping this recipe.

Boost regex is not mandatory in 1.3.2, but used by default due its speed:

https://github.com/jinja2cpp/Jinja2Cpp/compare/1.2.1...1.3.2#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR20

https://github.com/jinja2cpp/Jinja2Cpp/compare/1.2.1...1.3.2#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R277

I would suggest adding a logic in generate to follow that support:

```
def generate(self):
    tc = CMakeToolchain(self)
    if Version(self.version) >= "1.3.2":
        tc.cache_variables["JINJA2CPP_USE_REGEX"] = "std" if self.dependencies["boost"].options.without_regex else "boost"
```

toge:
@uilianries 
Sorry for late response.

Your suggested code doesn't work well.
```
CMake Error at CMakeLists.txt:188 (target_link_libraries):
  Target "jinja2cpp" links to:

    Boost::regex

    * An ALIAS target is missing.
```

`Boost::regex` is always required in external_boost_deps.cmake.
https://github.com/jinja2cpp/Jinja2Cpp/blob/1.3.2/thirdparty/external_boost_deps.cmake#L56

It may be a bug.

I added `with_regex` option to select regex library.
Could you review it again?

uilianries:
@toge thank you for pinging. I'll take a look tomorrow 😁

uilianries (approved):
LGTM. Thank you @toge!

AbrilRBS (approved):

PR #21323, opened 2023-11-23
duckx: add recipe

toge:
Specify library name and version:  **duckx/***

---

### DIFF
--- a/recipes/duckx/all/conanfile.py
+++ b/recipes/duckx/all/conanfile.py
@@ -1,0 +1,72 @@
+from conan import ConanFile
+from conan.tools.files import apply_conandata_patches, export_conandata_patches, get, rmdir
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+import os
+
+required_conan_version = ">=1.53.0"
+
+class DuckxConan(ConanFile):
+    name = "duckx"
+    description = " C++ library for creating and updating Microsoft Word (.docx) files."
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/amiremohamadi/DuckX/"
+    topics = ("docx", "docx-files", "office")
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": False,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 11
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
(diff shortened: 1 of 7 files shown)


recipes/duckx/all/conanfile.py
```diff
@@ -0,0 +1,75 @@
...
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.options["pugixml"].header_only = True
```
RubenRBS, line 39: Let me double-check, but I think the better approach in CCI is to set
```
default_options = {
    "pugixml/*:header_only": True
}
```

and have a `validate()` check to ensure that the option is properly set

RubenRBS (commented):
Thanks! One minor note about the `header_only` handling of pugixml, but let me double check that I'm not making things up!

recipes/duckx/all/conanfile.py
toge: @RubenRBS 
Thank you for your review.
I update this PR. Please review it again.

recipes/duckx/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def requirements(self):
+        self.requires("pugixml/1.14", transitive_headers=True)
+        self.requires("kuba-zip/0.2.6", transitive_headers=True)
+
+    def validate(self):
+        if self.info.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+        if not self.dependencies["pugixml"].options.header_only:
+            raise ConanInvalidConfiguration(f"{self.ref} requires header_only spdlog.")
```
RazielXYZ, line 54:
```suggestion
            raise ConanInvalidConfiguration(f"{self.ref} requires header_only pugixml.")
```

recipes/duckx/all/conanfile.py
```diff
@@ -0,0 +1,83 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("pugixml/1.14", transitive_headers=True)
+        self.requires("kuba-zip/0.2.6", transitive_headers=True)
```
RazielXYZ, line 48:
```suggestion
        self.requires("kuba-zip/0.3.1", transitive_headers=True)
```
Maybe we can try and bump this while we're at it

recipes/duckx/all/conanfile.py
RazielXYZ: It is a bit problematic that this *requires* header_only pugixml and breaks otherwise, but I'm not sure what it would take to fix it so it works with both

RazielXYZ (approved):
Looks good!

RubenRBS (approved):

PR #21324, opened 2023-11-23
cantera: Add new recipe

NelDav:
Specify library name and version:  **cantera/3.0.0**

Recently, I created a conan package of cantera and pushed it to a local conan repository of my company.
I thought it might be usefull for other people if cantera is accessible via conancenter.
Because of that, I contacted the authors of cantera and asked them if it would be ok if I publish cantera at conancenter.
They are fine with that and now I want to add cantera to conancenter.

---


recipes/cantera/all/FindCantera.cmake
uilianries, line 1: Please, do not include custom cmake files in case the upstream does not provide one. Instead, use `CMakeDeps`. Please, remove this file.

recipes/cantera/all/conanfile.py
```diff
@@ -0,0 +1,117 @@
...
+        
+        self.scons_extra_lib_dirs = \
+            self.dependencies["fmt"].cpp_info.libdirs +\
+            self.dependencies["yaml-cpp"].cpp_info.libdirs +\
+            self.dependencies["eigen"].cpp_info.libdirs
+        
+        self.scons_sundials_include = self.dependencies["sundials"].cpp_info.includedirs[0]
+        self.scons_sundials_libdir = self.dependencies["sundials"].cpp_info.libdirs[0]
+        
+        self.scons_boost_inc_dir = self.dependencies["boost"].cpp_info.includedirs[0]
```
uilianries, line 67: I didn't locally, but have you tried SConsDeps? https://docs.conan.io/2/reference/tools/scons.html

recipes/cantera/all/conanfile.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def layout(self):
+        self.folders.source = "."
```
uilianries, line 70: Is it not possible to use `basic_layout`? https://docs.conan.io/2/reference/tools/layout.html

Please, use basic_layout, so we can predict the folders layout.

uilianries (commented):
Thank you for your first contribution!

recipes/cantera/all/test_v1_package/CMakeLists.txt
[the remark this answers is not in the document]
NelDav: Ah thanks, I was struggling how to run the reciep for `test_v1_package`, so that is pretty nice. :)

recipes/cantera/all/FindCantera.cmake
NelDav: Sure, that is a relict form my first implementation. I just didn't removed it so far. I will do that as soon as possible.

recipes/cantera/all/conanfile.py
NelDav: I did, but I have to admit, that I do not know SCons well and therefore do not know, how to ue the `SConscript_conandeps` file generated by `SConsDeps`. I think In order to do so, I have to modify the `SConstruct` file but I do not want to modify the original file.

NelDav:
Additionally, suddenly the test_package cannot find `#include <fmt/format.h>` anymore. Last friday, that wasn't a problem. However now this problem occurs in the pipeline as well as locally.

NelDav:
> Thank you for your first contribution! 

Thanks for your first review.

uilianries:
@NelDav Please, take a look on serf recipe to have a reference about SCons with Conan: https://github.com/conan-io/conan-center-index/blob/master/recipes/serf/all/conanfile.py

NelDav:
> @NelDav Please, take a look on serf recipe to have a reference about SCons with Conan: https://github.com/conan-io/conan-center-index/blob/master/recipes/serf/all/conanfile.py

Thanks for the example. As far as I see, the main difference is that the `generate` and `build` parts are seperated more strictly, which is nice. What I do not understand, is the benefit of storing the arguments in a seperated file. Why is it not sufficient to simply store them in a member variable and access this variable from the `build` function?

NelDav:
[#KB-H012](https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H012-PACKAGE-LICENSE) say, that the package should contain a directory named `license`, where the license file is located. On the other hand, the [adding package](https://github.com/conan-io/conan-center-index/tree/master/docs/adding_packages#two-creating-a-package) documentation says:
> In ConanCenter, our belief is recipes should always match upstream, in other words, what the original author(s) intended.

Cantera stores the license file in a directory called `doc` or `documentation`, I forgot the exact wording.
So, shall I rename the directory or is there another workaround to fix the corresponding hook?

NelDav:
@uilianries 

Ok, I want to come back to my question about C++17 but want to formulate it a little bit more general now.
Some of the old CI jobs do not work. For example the the one using gcc5 is [not able to locate eigen](https://github.com/conan-io/conan-center-index/pull/21324#issuecomment-1829890361). I do not know what scons does when calling `CheckCXXHeader` but I think it is a bug somwehre in this function call. However, I think it does not realy matter since this job will not compile either. GCC5 does not supports the c++ `filesystem` library.

So, how to controll what jobs are execute, which jobs are allowed to fail, what jobs aren't and can I controll that?

uilianries:
> Cantera stores the license file in a directory called `doc` or `documentation`, I forgot the exact wording. So, shall I rename the directory or is there another workaround to fix the corresponding hook?

@NelDav It's not like that, you can just copy from source folder to the license folder directly: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L134

About the `doc` folder, it should be removed in case it does not contain a configuration file that is really required to use the package.

Thank you for asking

uilianries:
> So, how to controll what jobs are execute, which jobs are allowed to fail, what jobs aren't and can I controll that?

You still can control which compiler version and cppstd version are supported. For real example:

https://github.com/conan-io/conan-center-index/blob/590955cdaabfe730afcc6c7b4d5bc1e8f8967dea/recipes/osmanip/all/conanfile.py#L48

NelDav:
> It's not like that, you can just copy from source folder to the license folder directly: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L134
> 
> About the `doc` folder, it should be removed in case it does not contain a configuration file that is really required to use the package.

I think there was a missunderstanding. During the cantera build, the license file is automatically generated. When calling `scons install`, this license file is automatically placed into the package folder. But, it is not placed into a directory named `license`, but `doc`.
However, I think simply renaming this folder would be the best solution.

NelDav:
Currently I have the problem, that the CI faile, but @conan-center-bot  does not show me any recent results.
Additionally, before certain steps peformed by the ci were displayed in the CI summary, now only the upper level steps `c3i/conan-v2/pr-merge` and `continuuous-integration/jenkins/pr-merge` are displayed. When I click on Detail, I get http error 403

uilianries:
> Currently I have the problem, that the CI fails, but @conan-center-bot does not show me any recent results. Additionally, before certain steps peformed by the ci were displayed in the CI summary, now only the upper level steps `c3i/conan-v2/pr-merge` and `continuuous-integration/jenkins/pr-merge` are displayed. When I click on Detail, I get http error 403

Sorry hear that. The `c3i/conan-v2/pr-merge` and `continuuous-integration/jenkins/pr-merge` and indeed not allowed. The bot should post a comment in the PR with the build logs instead.

I just started a new build, hope this time works. The error that I found in the log, was related to an internal error when running git diff command.

NelDav:
Hi @uilianries, currently I have some problems regarding debug builds.
Cantera does not support linking against `fmt` and `yaml-cpp` which were build in debug mode.

Because of that, I would like to use release build for the dependencies.
I am wondering if this is somehow possible.
Annother solution would be to deactivate debug build for cantera.

uilianries:
@NelDav You can raise a ConanInvalidConfiguration for Debug with a helpful message, like:

"{self.ref} recipe does not support debug build type. Contributions are welcome."

NelDav:
Thanks, thats what I meant when I said

> Annother solution would be to deactivate debug build for cantera.

I just chose this solution for now.

NelDav:
Hi @uilianries, again, I have some questions.
I have to admit, this is much more effort than I imagined.
However, the `clang` build is failing because the environment variable `$CC` is defined in a strange way.
I think this variable should contain the path to the compiler right or the compiler name right? Something like it is shown in https://docs.conan.io/1/reference/profiles.html#profile-includes

In my case, `$CC` is simply `cc`, which is not known by canteras `SConstruct` file.
I am not an expert for stuff like this, so I am not sure if the value `cc` is correct, but I think it looks strange.
I would expect `/<path>/clang` or `clang`.

Here is how cantera checks for clang (you can see, that they expecte that `clang` is somewhere in th `$CC` string):
grafik

NelDav:
Seccond problem is, that on macos, sundials is not found.
I checked the way, how cantera searches for for sundials, and it is quite straight forward.
They simply check if the `sundials_cvodes` library exists in the stated folder and if a certain header file exists.
If this is the case, scons tries to call some specific functin. If any of those function can be called, sundials was found.

However, I am not able to check the macos build in more detail because I do not have access to macos.
It would be amazing, if you know somebody who has access to macos and enough time to check this problem.

NelDav:
> In my case, $CC is simply cc, which is not known by canteras SConstruct file.
I am not an expert for stuff like this, so I am not sure if the value cc is correct, but I think it looks strange.
I would expect /<path>/clang or clang.

I read the documentation of GNU make and it looks like `cc` is the default value of $CC

uilianries:
> Hi @uilianries, again, I have some questions. I have to admit, this is much more effort than I imagined. However, the `clang` build is failing because the environment variable `$CC` is defined in a strange way. I think this variable should contain the path to the compiler or the compiler name right? Something like it is shown in [docs.conan.io/1/reference/profiles.html#profile-includes](https://docs.conan.io/1/reference/profiles.html#profile-includes)
> 
> In my case, `$CC` is simply `cc`, which is not known by canteras `SConstruct` file. I am not an expert for stuff like this, so I am not sure if the value `cc` is correct, but I think it looks strange. I would expect `/<path>/clang` or `clang`.
> 
> Here is how cantera checks for clang (you can see, that they expecte that `clang` is somewhere in th `$CC` string):
> 
> ```python
> elif "clang" in env.subst("$CC"):
>     config.select("clang")
> 
> else:
>     logger.error(f"Unrecognized C compiler {env['CC']!r}")
>     sys.exit(1)
> ```

The `cc` is configured for each docker image used in ConanCenterIndex, it should follow the compiler that's expected by the docker image name. So, those images with clang as main compiler, `cc` should point to clang as well.

As you are using environment variables, you will need set them, just in case when someone installs the package, don't get it messed with the system. You could something like:

```
def generate(self):
    # Only for Linux/FreeBSD, need to check for Apple and Windows
    cxx_compiler = {"clang": "clang++", "gcc": "g++"}.get(self.settings.compiler)
    tc = AutotoolsToolchain(self)
    env = tc.environment()
    env.define("CC", str(self.settings.compiler))
    env.define("CXX", cxx_compiler)
    envvars = env.vars(self)
    envvars.save_script("conan_compiler_envvars")
```

This will generate a script named `conan_compiler_envvars` that will be loaded automatically before executing `run(self, ...)`. It should fill not only those environment variables like `CC` and `CXX`, but also `CXXFLAGS`, `CFLAGS` and `LDFLAGS` as well.

uilianries:
> Seccond problem is, that on macos, sundials is not found. I checked the way, how cantera searches for for sundials, and it is quite straight forward. They simply check if the `sundials_cvodes` library exists in the stated folder and if a certain header file exists. If this is the case, scons tries to call some specific functin. If any of those function can be called, sundials was found.
> 
> However, I am not able to check the macos build in more detail because I do not have access to macos. It would be amazing, if you know somebody who has access to macos and enough time to check this problem.

But how this scons script searches it? Could it be a pkgconfig file instead? We have the [PkgConfigDeps](https://docs.conan.io/2/reference/tools/gnu/pkgconfigdeps.html) for instance.

Otherwise, you could use [AutotoolsDeps](https://docs.conan.io/2/reference/tools/gnu/autotoolsdeps.html) generator to extract where libraries are installed.

NelDav:
> This will generate a script named conan_compiler_envvars that will be loaded automatically before executing run(self, ...). It should fill not only those environment variables like CC and CXX, but also CXXFLAGS, CFLAGS and LDFLAGS as well.

Thanks, I am wondering, if it would be better to use [SConsDeps](https://docs.conan.io/2/reference/tools/scons.html) instead, this generator will also set `CXX`, but the documentation does not tell anything about `CC`.
I tried that, but for some reason, conan cannot find the `conan.tools.scons` module.

NelDav:
> But how this scons script searches it? 

So finding the location, where the library is installed might not be the problem.
I am retriving the location inside the conan file like this: `sundials_info = self.dependencies["sundials"].cpp_info`

The include and lib directory is passed to scons.
I think the problem is, that the name of the lib file (or what ever files apple uses) differs from the one that cantera expects.
But as told, I am not sure, since I am not able debug.

NelDav:
@conan-io/barbarians I tested clang (14 and 13) on my system and the test works totally fine. But, as you can see in https://github.com/conan-io/conan-center-index/pull/21324#issuecomment-1912116858, the pipeleine is not able to link cantera. Do you have anny suggestion how to debug?

NelDav:
> @conan-io/barbarianse (@uilianries) I tested clang (14 and 13) on my system and the test works totally fine. But, as you can see in [#21324 (comment)](https://github.com/conan-io/conan-center-index/pull/21324#issuecomment-1912116858), the pipeleine is not able to link cantera. Do you have anny suggestion how to debug?

@uilianries, please I do not know how to proceed. Is it possible to debug the problem on the same system, the CI uses? For example by using Docker or something similar?

uilianries:
Hello @NelDav You can use the Docker image for Clang 13 that's used by the CI

https://hub.docker.com/r/conanio/clang13-ubuntu16.04-jenkins/tags

`docker pull conanio/clang13-ubuntu16.04:1.62.0`

Then you can use it directly you branch:

```
cd conan-center-index/recipes/cantera
docker run --rm -ti -v ${PWD}:/home/conan/project --platform=linux/amd64 conanio/clang13-ubuntu16.04:1.62.0
# In the container
cd project/
conan create all/ cantera/3.0.0@
```

NelDav:
Thank you very much @uilianries, that was very helpfull.
I will debug now :)

NelDav:
Hey @uilianries, most builds are working now, but I still have some problems. For example the Visual Studio build fails with some linker errors. On my local machine, the build works quite fine. The warning at the end of the test build looks like cmake is not able to find boost. I am wondering, what cmake version is installed on this image? In older versions, cmake did not use the `<PackageName>_ROOT`.

PR #21325, opened 2023-11-23
[pcl] Hotfix: export qhull define

### LINKED ISSUE #21290: pcl:with_qhull=True does not set HAVE_QHULL
### What is your question?

Hi,

I've build pcl/1.13.1 with apple-clang 15.0 and pcl:with_qhull=True because I need pcl::ConcaveHull in my application. 
The absence of setting HAVE_QHULL prevents the include or linking with concave_hull.

My conanfile:
```
[requires]
    pcl/1.13.1

[generators]
    CMakeDeps
    CMakeToolchain
```
pcl:with_qhull is default True and it doesn't change if I add this to the options or not. 

In pcl_config.h.in:
`#cmakedefine HAVE_QHULL 1`

This should result in setting HAVE_QHULL in pcl_config.h, but:
`/* #undef HAVE_QHULL */`

Is there an option to set HAVE_QHULL?

uilianries:
Specify library name and version:  **qhull/1.13.1**

The exported `pcl_config.h` is not following the requested package options:

https://github.com/PointCloudLibrary/pcl/blob/pcl-1.13.1/pcl_config.h.in#L41

Let's export those defines directly. 

fixes #21290

/cc @DevlBert

---

### DIFF
--- a/recipes/pcl/all/conanfile.py
+++ b/recipes/pcl/all/conanfile.py
@@ -574,6 +574,13 @@ class PclConan(ConanFile):
         if self.settings.os == "Windows":
             common.system_libs.append("ws2_32")
 
+        if self.options.with_qhull:
+            self.cpp_info.defines.append('HAVE_QHULL=1')
+        if self.options.with_png:
+            self.cpp_info.defines.append('HAVE_PNG=1')
+        if self.options.with_cuda:
+            self.cpp_info.defines.append('HAVE_CUDA=1')
+
         # TODO: Legacy, to be removed on Conan 2.0
         self.cpp_info.names["cmake_find_package"] = "PCL"
         self.cpp_info.names["cmake_find_package_multi"] = "PCL"


RubenRBS (approved):

ericLemanissier (approved):

SpaceIm:
```
The exported pcl_config.h is not following the requested package options:

https://github.com/PointCloudLibrary/pcl/blob/pcl-1.13.1/pcl_config.h.in#L41
```

Why? Shouldn't it work?

SpaceIm:
Looking at upstream build files, it's a bug of pcl recipe (because we don't use their custom Find modules where all these `HAVE_*` are defined and then catched during configuration of `pcl_config.h.in`).
So conan patches should be fixed (or we could just inject these CMake variables through CMakeToolchain), instead of exposing these interface definitions.

PR #18864, opened 2023-07-24
coin-cbc: migrate to Conan v2

valgur:

PR #21328, opened 2023-11-23
coin-utils: explicitly disable blas and lapack, add CMakeDeps test

valgur:
BLAS and LAPACK were getting picked up from the system automatically otherwise. This breaks downstream package builds.

Also added a separate test_package test against CMakeDeps. Not that important here, but the `coin-cbc` package is having severe issues with `PkgConfigDeps` (#18864), but not `CMakeDeps`, so I thought it would not hurt to include it for all COIN packages.


jwillikers (approved):

EstebanDugueperoux2 (approved):

franramirez688 (approved):

franramirez688:
> Hi @franramirez688 I need this PR for my project. What it remains to do on thi PR?
> 
> Regards.

So sorry for the late reply, I completely forgot about it 😭

PR #21341, opened 2023-11-23
libecwj2: new recipe

valgur:
Adds `libecwj2`: https://trac.osgeo.org/gdal/wiki/ECW

Source code mirror: https://github.com/meteogrid/libecwj2

`libecwj2` is a legacy version of the ERDAS ECW/JP2 SDK, which provides support for ECW and JPEG 2000 image formats common in remote sensing. The 3.3 version is the last open source (but not OSI-compatible) version of this library, released in 2006. The licensing allows modifications only through patching, which is why it has not been forked and the original source archive is still used as the basis.

Used by GDAL for its ECW plugin.

[Packaging status](https://repology.org/project/libecwj2/versions)


RubenRBS:
Thanks! Will double check that the license works for us - but nevertheless thanks for taking the time to create the recipe, really appreciated :)

jwillikers (approved):

jwillikers (approved):

valgur:
Needs work, still. Had some compilation errors when used as a GDAL dependency.

PR #21350, opened 2023-11-24
doxygen: Drop compiler and build type from package ID

kyllingstad:
Specify library name and version:  **doxygen/\***

Doxygen is a documentation generator. The compiler and build type used for building client code should be irrelevant.

---


uilianries (changes requested):
Unfortunately we can't do that: https://github.com/conan-io/conan-center-index/pull/20134#issuecomment-1739386435

We need to keep Doxygen with full settings, otherwise we will have problems when building with older glibc versions.

kyllingstad:
Ok, I understand. Thanks for the response.

PR #21354, opened 2023-11-24
Add recipe for sqlite3mc

utelle:
Specify library name and version:  **sqlite3mc/1.8.0**

I'm the author of the project [SQLite3 Multiple Ciphers](https://github.com/utelle/SQLite3MultipleCiphers) - **sqlite3mc** in short. The project adds an encryption extension to the [SQLite](https://sqlite.org) library allowing to transparently encrypt the database files. As the name suggests the extension supports several common cipher schemes (used by other Open Source projects):

- [wxSQLite3](https://github.com/utelle/wxsqlite3): AES 128 Bit CBC - No HMAC
- [wxSQLite3](https://github.com/utelle/wxsqlite3): AES 256 Bit CBC - No HMAC
- [sqleet](https://github.com/resilar/sqleet): ChaCha20 - Poly1305 HMAC
- [SQLCipher](https://www.zetetic.net/sqlcipher/): AES 256 Bit CBC - SHA1/SHA256/SHA512 HMAC
- [System.Data.SQLite](http://system.data.sqlite.org/): RC4
- [Ascon](https://ascon.iaik.tugraz.at/): Ascon-128 v1.2 (Lightweight Authenticated Encryption)

The actual cipher scheme used for database encryption can be chosen at runtime.

I was asked by a user of my project, @Myroendan, whether a Conan recipe could be provided. So, here it is.

---


recipes/sqlite3mc/all/conandata.yml
```diff
@@ -0,0 +1,5 @@
+sources:
+  "1.8.0":
+    url: "https://github.com/utelle/SQLite3MultipleCiphers/archive/refs/tags/v1.8.0.tar.gz"
+    sha256: "13D9B939BEF7C7371D58A3874F83B18CF330EB2171205B3680ACDDB2215BE0E5"
+    strip_root: true
```
SpaceIm, line 5:
```suggestion
```

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,226 @@
...
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version])
```
SpaceIm, line 128:
```suggestion
        get(self, **self.conan_data["sources"][self.version], strip_root=True)
```

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.options.require_zlib:
+            self.requires("zlib/[>=1.2.9]")
```
SpaceIm, line 135:
```suggestion
            self.requires("zlib/[>=1.2.11 <2]")
```

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,226 @@
...
+        if self.options.with_icu:
+            self.requires("icu/[>=67.1]")
```
SpaceIm, line 137:
```suggestion
            self.requires("icu/74.1")
```

recipes/sqlite3mc/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+cmake_minimum_required(VERSION 3.24.0)
```
SpaceIm, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15)
```

recipes/sqlite3mc/all/test_package/conanfile.py
```diff
@@ -0,0 +1,28 @@
...
+from conan.tools.build import can_run
+
+
+class sqlite3mcTest(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps", "CMakeToolchain"
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.24 <4]")
+
```
SpaceIm, line 13:
```suggestion
```

recipes/sqlite3mc/all/test_package/conanfile.py
```diff
@@ -0,0 +1,28 @@
+import os
+from conan import ConanFile
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.build import can_run
+
+
+class sqlite3mcTest(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps", "CMakeToolchain"
```
SpaceIm, line 9:
```suggestion
    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv"
    test_type = "explicit"
```

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,226 @@
+import os
+from conan import ConanFile
+from conan.tools.cmake import cmake_layout, CMakeDeps, CMakeToolchain, CMake
+from conan.tools.files import get, copy
+
+required_conan_version = ">=1.56.0 <2 || >=2.0.6"
```
SpaceIm, line 6:
```suggestion
required_conan_version = ">=1.53.0"
```

SpaceIm (changes requested):

recipes/sqlite3mc/all/conanfile.py
SpaceIm: see https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,225 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.options.require_zlib:
+            self.requires("zlib/[>=1.2.11 <2]")
+        if self.options.with_icu:
+            self.requires("icu/[74.1]")
```
SpaceIm, line 136:
```suggestion
            self.requires("icu/74.1")
```

RubenRBS (commented):
Hi @utelle thanks a lot for taking the time to contribute the recipe for your library, we're always happy to get author's involvment with their recipes as that usually makes the resulting packages much better

We'll now take a look at the recipe and try to build some permutation of some of the options you've added for the library, thanks a lot for your patience :)

RubenRBS:
Builds look good on my side, will properly review once the CI finishes this run, thanks a lot for your patience! :)

SpaceIm (approved):

utelle:
Over 2 months have passed ... and I see no progress at all. Is there anything I can or should do from my side?

recipes/sqlite3mc/all/test_package/main.c
```diff
@@ -0,0 +1,73 @@
+#include <sqlite3mc/sqlite3mc.h>
+#include <stdio.h>
+#include <string.h>
+
+int main(int argc, char *argv[])
+{
+    sqlite3 *db;
```
RubenRBS, line 7: Hi! Could we get a simplified version of this? It's not in CCI's policy to have extensive tests in the test_package, it should be as minimal as possible while testing the correct _package_ of the library :)

RubenRBS (changes requested):
Hi @utelle thanks a lot for your contribution, and sorry about the delay, we appreciate your patience a lot, some PRs get longer to get reviewed than other, sorry about that :)

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,225 @@
...
+        cmake = CMake(self)
+        cmake.install()
+        copy(self, "LICENSE*", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"), keep_path=False)
+
+    def package_info(self):
+        if self.options.shared:
+            self.cpp_info.libs = ["sqlite3mc"]
+        else:
+            self.cpp_info.libs = ["sqlite3mc_static"]
+        if self.settings.os in ("Linux", "Macos"):
```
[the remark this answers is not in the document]
utelle: Could you please explain, why `"Macos"` should be replaced by `"FreeBSD"`? IMHO this can't be correct, because

1) the libraries `pthread`, `dl`, and `m` are required for `Macos`, too, and 
2) in line 224 `self.settings.os` is explicitly checked for `"Macos"` again, because an additional attribute is required under `Macos`.

I could understand, if you would want the following change:
```python
 if self.settings.os in ("Linux", "FreeBSD", "Macos"):
```

recipes/sqlite3mc/all/conanfile.py
```diff
@@ -0,0 +1,225 @@
...
+        tc.variables["SQLITE_ENABLE_FTS3"] = self.options.enable_fts3
+        tc.variables["SQLITE_ENABLE_FTS3_PARENTHESIS"] = self.options.enable_fts3_paranthesis
+        tc.variables["SQLITE_ENABLE_FTS4"] = self.options.enable_fts4
+        tc.variables["SQLITE_ENABLE_FTS5"] = self.options.enable_fts5
+
+        tc.variables["SQLITE_ENABLE_CARRAY"] = self.options.enable_carray
+        tc.variables["SQLITE_ENABLE_CSV"] = self.options.enable_csv
+        tc.variables["SQLITE_ENABLE_EXTFUNC"] = self.options.enable_extfunc
+        tc.variables["SQLITE_ENABLE_GEOPOLY"] = self.options.enable_geopoly
+        tc.variables["SQLITE_ENABLE_JSON1"] = self.options.enable_json1
```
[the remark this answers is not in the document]
utelle: > I tried to find an usage for this option, but I failed to find it. I see it in the CMakeLists, but I don't see it in the code, I mean, a required definition like `ifdef`:
> 
> https://github.com/search?q=repo%3Autelle%2FSQLite3MultipleCiphers%20SQLITE_ENABLE_JSON1&type=code
> 
> Could you please explain a bit more how does it work, same to other options?

Several SQLite extensions can be included optionally at compile time using `#ifdef` statements. Mostly this happens in the main C file [sqlite3mc.c](https://github.com/utelle/SQLite3MultipleCiphers/blob/main/src/sqlite3mc.c) (like for `SQLITE_ENABLE_CSV` and `SQLITE_ENABLE_EXTFUNC`). 

But there are also extension that are already included in the original (resp patched) SQLite source amalgamation C file [sqlite3patched.c](https://github.com/utelle/SQLite3MultipleCiphers/blob/main/src/sqlite3patched.c) (like for `SQLITE_ENABLE_GEOPOLY`).

The symbol `SQLITE_ENABLE_JSON1` is different. This symbol does indeed not exist any more. In prior SQLite versions it was available, but in recent versions the JSON extension is included per default, but it could be explicitly excluded by specifying the symbol `SQLITE_OMIT_JSON`.

For the original project [**SQLite3 Multiple Ciphers**](https://github.com/utelle/SQLite3MultipleCiphers) I decided to expose many common SQLite configuration options to give users a high degree of freedom to choose the options and extensions most appropriate for their project. Per default many extensions are enabled, but in some projects there exist memory restrictions, so that users want to be able to disable unneeded extensions.

recipes/sqlite3mc/all/test_package/main.c
utelle: > Hi! Could we get a simplified version of this? It's not in CCI's policy to have extensive tests in the test_package, it should be as minimal as possible while testing the correct _package_ of the library :)

Well, this **is** the most simple test I could think of. To verify that the package was build correctly it is necessary to create an encrypted database with some content and to access it with the right passphrase and a wrong passphrase afterwards. 

If that test is **too complex**, we don't need any test at all.

utelle: I removed the configuration option `SQLITE_ENABLE_JSON1`, because this option is no longer used by SQLite. And I added `FreeBSD` to the os list (instead of replacing `Macos` by `FreeBSD`), because this seems to be the only logical modification.

Regarding the "extensive" test I have no idea what you expect. What could and/or should be tested, if not the proper build of the package?

Myroendan: Hi,
we could remove the error handlings from the test,  use logical OR on the result during operations, and check it once in the end. That would make 35-40 lines shorter. Would that work?

utelle: > we could remove the error handlings from the test, use logical OR on the result during operations, and check it once in the end. That would make it 35-40 lines shorter. Would that work?

It is certainly possible to reduce the number of lines of code by improving the error handling code (for example, by introducing a small function that checks the return code and emits an error message if necessary), but I'm not sure if that is what @RubenRBS intends.

recipes/sqlite3mc/all/conanfile.py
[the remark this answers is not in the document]
valgur: I see that these are indeed linked against by the project on macOS: https://github.com/utelle/SQLite3MultipleCiphers/blob/7e8ecba9a82ea1336b118a21ccb733f4ed292efa/CMakeLists.txt#L262-L266

Excuse the dumb question (I'm not that familiar with the platform): do `m` and `dl` even exist on macOS? It's the first time I'm seeing these being linked against outside of Linux/FreeBSD. As far as I know, these are only provided by glibc and similar libc implementations, which is not available on macOS.

valgur (approved):

recipes/sqlite3mc/all/conanfile.py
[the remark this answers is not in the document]
utelle: > I see that these are indeed linked against by the project on macOS: https://github.com/utelle/SQLite3MultipleCiphers/blob/7e8ecba9a82ea1336b118a21ccb733f4ed292efa/CMakeLists.txt#L262-L266
> 
> Excuse the dumb question (I'm not that familiar with the platform): do `m` and `dl` even exist on macOS? It's the first time I'm seeing these being linked against outside of Linux/FreeBSD. As far as I know, these are only provided by glibc and similar libc implementations, which is not available on macOS.

Actually, I'm not very familiar with macOS either. However, the GitHub CI action (for example [this run](https://github.com/utelle/SQLite3MultipleCiphers/actions/runs/7723062958/job/21052394677)) runs without errors on macOS, and up to now I haven't received any complaints from users of my component.

utelle:
@RubenRBS 
> Hi @utelle thanks a lot for your contribution, and sorry about the delay, we appreciate your patience a lot, some PRs get longer to get reviewed than other, sorry about that :)

Another  month went by ... and my latest modifications and comments didn't get any helpful response. I have to admit that I gradually loose interest in this endeavor.

RubenRBS (approved):

RubenRBS:
Sorry about this, I know that feeling ignored after taking the time to contribute a PR with all the effort it requires can be discouraging - nothing further from the truth, you contribution is truly appreciated.

Sometimes some PRs fall thru the cracks due to the sheer volume of PRs that CCI recives - we're working on systems to be able to better handle that load, so that things like this don't happen again, but that'd still take some time

I've approved to unblock this and will ask the team to send a follow-up pr if necessary to get the final touches if any are needed, but again, sorry about the delay :)

utelle:
> I know that feeling ignored after taking the time to contribute a PR with all the effort it requires can be discouraging

Well, I think I really showed a lot of patience while the PR was queued for processing, but once review(s) began I had expected a somewhat more continuous process. Working on a PR with long breaks inbetween requires more time and energy on both sides to pick up the loose ends where they were left off.

> Sometimes some PRs fall thru the cracks due to the sheer volume of PRs that CCI recives - we're working on systems to be able to better handle that load, so that things like this don't happen again, but that'd still take some time

I fully understand that it is difficult to handle such a high volume of PRs, especially as the work is mostly done by volunteers in their sparetime. And of course it is important to carefully check the quality of new recipes to give Conan users a pleasant experience.

> I've approved to unblock this and will ask the team to send a follow-up pr if necessary to get the final touches if any are needed, but again, sorry about the delay :)

Thanks.

In this context I have a question: how should updates of the recipe be handled? Actually, the current version of _SQLite3 Multiple Ciphers_ is already 1.8.4. Which versions should be added to the Conan recipe? All intermediate versions? Or just the latest? Should old versions be dropped from the recipe? Or kept forever?

PR #21386, opened 2023-11-26
base64: modernize more + avoid to build tests and warnings as errors in CMake build + fix cmake target name

SpaceIm:
- add package_type
- remove shared option before 0.5.0 instead of raising if shared
- fix cmake target name: it's `aklomp::base64` (see https://github.com/aklomp/base64/blob/v0.5.1/CMakeLists.txt#L285)
- use rm_safe
- no warnings as errors
- don't build tests in CMake build
- reproducible build (disable OpenMP, otherwise it may or may not use OpenMP, an option could be added in another PR)
- don't build cli (it's not super portable, and it was not built in previous versions)

---


uilianries (approved):
LGTM

toge (approved):

Issue #21357, opened 2023-11-25
[request] cpython/*


lmascelli:
### Package Name/Version

cpython/*

### Changelog

https://github.com/conan-io/conan-center-index/tree/master/recipes/cpython

### Context about the new update

The current recipe seams to be incompatible with conan 2.0

Issue #20695, opened 2023-10-19
[package] cpython/3.10.0: Issue with libuuid


hiteshdasikamw:
### Description

I have a conan project with the following dependencies
`    requires = [
        "gtest/cci.20210126", 
        "boost/1.78.0",
        "libunwind/1.6.2",
        "batteries/0.5.17@tastolfi+batteries/stable",
        "cli11/2.2.0",
        "protobuf/3.20.0",
        "lua/5.4.4",
        "libzip/1.8.0",
        "xz_utils/5.4.2",
        "zlib/1.2.13",
        "openssl/3.1.3",
    ]`

The moment I include "cpython/3.10.0" , I end up getting the below error:

ERROR: At least two recipes provides the same functionality:
 - 'libuuid' provided by 'libuuid/1.0.3', 'util-linux-libuuid/2.39'

Below is my linux distro info

Distributor ID:	Debian
Description:	Debian GNU/Linux 11 (bullseye)
Release:	11
Codename:	bullseye

### Package and Environment Details

* Package Name/Version: **cpython/3.10.0**
* Operating System+version: **Debian 11**
* Compiler+version: **GCC 10.2**
* Conan version: **conan 1.60.0**
* Python version: **Python 3.9.2**

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=10
compiler.libcxx=libstdc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

make install

### Logs

<details><summary>Click to expand log</summary>

```
WARN: boost/1.78.0: requirement zlib/[>=1.2.11 <2] overridden by mps_universal_runtime/0.2.1 to zlib/1.2.13 
WARN: libunwind/1.6.2: requirement xz_utils/5.4.4 overridden by mps_universal_runtime/0.2.1 to xz_utils/5.4.2 
WARN: libunwind/1.6.2: requirement zlib/[>=1.2.11 <2] overridden by mps_universal_runtime/0.2.1 to zlib/1.2.13 
WARN: protobuf/3.20.0: requirement zlib/[>=1.2.11 <2] overridden by mps_universal_runtime/0.2.1 to zlib/1.2.13 
WARN: libzip/1.8.0: requirement zlib/[>=1.2.11 <2] overridden by mps_universal_runtime/0.2.1 to zlib/1.2.13 
[...]

Issue #19374, opened 2023-08-23
[package] cpython/3.10.0: can't build with MSVC


AndreyMlashkin:
### Description

Can't build cpython/3.10.0 with MSVC

### Package and Environment Details

* Package Name/Version: **cpython/3.10.0**
* Operating System+version: **Windows 11**
* Compiler+version: **MSVC 19**
* Conan version: **conan 1.60.2**

### Conan profile

[settings]
os=Windows
os_build=Windows
arch=x86_64
arch_build=x86_64
compiler=Visual Studio
compiler.version=17
build_type=Release
[options]
cpython:shared=True
[build_requires]
[env]

### Steps to reproduce

conan install cpython/3.10.0@ --build missing -u -r conan-center

### Logs

<details><summary>Click to expand log</summary>

```
conan install cpython/3.10.0@ --build missing -u -r conan-center
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=17
os=Windows
os_build=Windows
[options]
cpython:shared=True
[build_requires]
[env]

WARN: tcl/8.6.10: requirement zlib/1.2.13 overridden by tk/8.6.10 to zlib/1.2.11
bzip2/1.0.8: WARN: Can't update, no package in remote
expat/2.4.1: WARN: Can't update, no package in remote
libffi/3.2.1: WARN: Can't update, no package in remote
mpdecimal/2.5.0: WARN: Can't update, no package in remote
openssl/1.1.1l: WARN: Can't update, no package in remote
sqlite3/3.36.0: WARN: Can't update, no package in remote
xz_utils/5.2.5: WARN: Can't update, no package in remote
zlib/1.2.11: WARN: Can't update, no package in remote
tcl/8.6.10: WARN: Can't update, no package in remote
tk/8.6.10: WARN: Can't update, no package in remote
cpython/3.10.0: WARN: Package binary is corrupted, removing: 393bb3b4e72af5ae35fc5b9a0d21443d6354a8ac
Installing package: cpython/3.10.0
Requirements
    bzip2/1.0.8 from 'conan-center' - Cache
    cpython/3.10.0 from 'conan-center' - Cache
    expat/2.4.1 from 'conan-center' - Cache
    libffi/3.2.1 from 'conan-center' - Cache
    mpdecimal/2.5.0 from 'conan-center' - Cache
    openssl/1.1.1l from 'conan-center' - Cache
[...]

PR #21387, opened 2023-11-26
CPython: Conan 2.0

Ahajha:

I will be leaving the versions as-is in this PR, as changing them tends to break things, and I want to keep the scope small. A separate PR for that is here: https://github.com/conan-io/conan-center-index/pull/22599 (depends on this PR)

I believe there is some value in having a 2.7 recipe available, but probably not worth it for the CCI maintainers to deal with. If specifically needed, there is a mostly working 2.7 (and 3.7) version available a few commits back from the end of this branch.

Previously, the packages were not 100% relocatable on *nix. For now I always set `PYTHONHOME` in the test packages, not sure if this is a perfect idea but for now it seems to be a reasonably not-hacky solution.

Short paths are enabled. Lots of files in __pycache__ directories are packaged - this is intentional, as they are compiled (bytecode/JIT) versions of all the standard library modules.

Known intermittent issues:
1. Sometimes, when packaging on the Mac CI, there will be an error relating to a file already existing. If I had to guess the frequency, probably around 25% of the time this happens. I haven't been able to reliably reproduce it.

There are many CPython issues/PRs out right now:

Resolves https://github.com/conan-io/conan-center-index/issues/21357  
Resolves https://github.com/conan-io/conan-center-index/issues/20695  
Resolves https://github.com/conan-io/conan-center-index/issues/19374  
Resolves https://github.com/conan-io/conan-center-index/issues/10808  
Resolves https://github.com/conan-io/conan-center-index/issues/12373 (no project upgrade occurs)  
Resolves https://github.com/conan-io/conan-center-index/issues/8198  
Resolves https://github.com/conan-io/conan-center-index/issues/9890 (was only broken on Linux/Mac. Some other with_xxx options don't work, but that's a separate issue)

Unsure if it fixes this, I test locally on an M2 with apple-clang 15 so similar but not exact: https://github.com/conan-io/conan-center-index/issues/13900
This should be able to be closed after this, since this is the last recipe using the old `libuuid` recipe: https://github.com/conan-io/conan-center-index/issues/19084
This is a problem with CPython being static by default. In future versions, CPython doesn't even properly support static mode, but it's still annoying that the default configuration is invalid: https://github.com/conan-io/conan-center-index/issues/9333
Unsure, seems like a CPython problem itself: https://github.com/conan-io/conan-center-index/issues/8608

This PR is based on the following, and should no longer be needed:
Closes https://github.com/conan-io/conan-center-index/pull/18064  

There was also an  attempt to fix the libuuid issue:
Closes https://github.com/conan-io/conan-center-index/pull/19547

---


RubenRBS (commented):
Quick comment about the patches

recipes/cpython/all/conandata.yml
```diff
@@ -16,57 +16,32 @@ sources:
...
-      base_path: "source_subfolder"
-    - patch_file: "patches/3.9.7-0002-_msi-vcxproj.patch"
-      base_path: "source_subfolder"
-    - patch_file: "patches/3.10.0-0003-_ctypes-ffi.patch"
-      base_path: "source_subfolder"
-    - patch_file: "patches/3.10.0-0004-setup.py-pass-CFLAGS-CPPFLAGS.patch"
-      base_path: "source_subfolder"
-    - patch_file: "patches/3.10.0-0005-disable-macos-tcltk.patch"
-      base_path: "source_subfolder"
+    - patch_file: "patches/3.10/3.10.0-0001-msvc.patch"
```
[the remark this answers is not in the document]
Ahajha: I do plan on removing the MSVC patches in a future PR, but for now I can document all of them.

Ahajha:
The linux configure error on v2 seems to be caused by this PR: https://github.com/conan-io/conan-center-index/pull/22081

My guess is that any dependencies that use xorg will need to be rebuilt. Curious why v1 still works.

Ahajha:
Current blocker: Some issues with tcl and tk on Mac. Will link relevant PRs as they are made.

recipes/cpython/all/conandata.yml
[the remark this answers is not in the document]
Ahajha: Finally got around to adding these. A few notes on some of the patches, some are irrelevant upstream as the relevant sections have been removed for one reason or another:
`patches/3.9/3.9.7-0002-_msi-vcxproj.patch`: `msilib` is removed in 3.13: https://github.com/python/cpython/pull/104911
`patches/3.10/3.10.0-0004-setup.py-pass-CFLAGS-CPPFLAGS.patch`: The setup.py script is entirely removed in 3.12
`patches/3.10/3.10.0-0005-disable-macos-tcltk.patch`: This is in setup.py, I think the relevant section is removed in 3.11, though again regardless the whole script is removed in 3.12

uilianries:
Just started an internal CI build to generate the missing binary tk/8.6.10

RubenRBS (commented):
_Wow_, it's finally here. I want to congratulate you for taking the time and effort to contribute this, it truly is appreciated ❤️ 

I've asked @uilianries to help me review the PR this week, but from I can see this looks great :)

jellespijker:
I couldn't agree more. Great effort and work on this PR. I will make sure that you guys get mentioned as contributors on the next UltiMaker Cura 5.8 release which will be possible due to your hard work.

Ahajha:
TODO: Looks like Python 2.x breaks on the latest xz_utils.

Ahajha:
Grrr... Not sure but it looks like the recipe has magically re-broken itself in CI

jwillikers:
> Grrr... Not sure but it looks like the recipe has magically re-broken itself in CI

Looks like an issue with long paths on Windows to me:

```
LINK : fatal error LNK1104: cannot open file 'C:\J\workspace\prod-v1\bsr@3\cci-0fb77bd6\recipes\cpython\all\test_v1_package\build\7595ede32249a44bc554fbfc571f6a1af711d216\temp.win-amd64-3.10-pydebug\Debug\J\workspace\prod-v1\bsr@3\cci-0fb77bd6\recipes\cpython\all\test_package\py3\spam_d.cp310-win_amd64.lib'
```

recipes/cpython/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+    @property
+    def cmake_arg(self):
+        return ";".join(self._cmake_lut[a] for a in (self.debug, self.pymalloc, self.unicode))
+
+
+class TestPackageConan(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "cmake"
```
jwillikers, line 34:
```suggestion
    generators = "cmake"
    short_paths = True
```

Ahajha: I'll try this in a second, but I think the underlying issue is that for some reason it's doubling the path within the test package folder. But this seems like a good enough solution if I can't figure out why that's happening.

jwillikers:
Fyi, using Conan V2 the test package fails to build for me when testing the SSL module for Python 3.10.0. Same error trying with Python 3.12.2 from your other branch as well.

```
cpython/3.10.0 (test package): Module worked as expected
cpython/3.10.0 (test package): RUN: /home/jordan/.conan2/p/b/cpyth43a849bd4dab6/p/bin/python3.10 /var/home/jordan/Projects/conan-center-index/recipes/cpython/all/test_package/test_package.py -b /var/home/jordan/Projects/conan-center-index/recipes/cpython/all/test_package/build/x86_64-gcc/Debug -t ssl
testing ssl
Traceback (most recent call last):
  File "/var/home/jordan/Projects/conan-center-index/recipes/cpython/all/test_package/test_package.py", line 226, in <module>
    main()
  File "/var/home/jordan/Projects/conan-center-index/recipes/cpython/all/test_package/test_package.py", line 222, in main
    ALL_TESTS[ns.test_module]()
  File "/var/home/jordan/Projects/conan-center-index/recipes/cpython/all/test_package/test_package.py", line 22, in inner_fn
    fn()
  File "/var/home/jordan/Projects/conan-center-index/recipes/cpython/all/test_package/test_package.py", line 211, in test_ssl
    default_context = ssl.create_default_context()
  File "/home/jordan/.conan2/p/b/cpyth43a849bd4dab6/p/lib/python3.10/ssl.py", line 756, in create_default_context
    context = SSLContext(PROTOCOL_TLS_CLIENT)
  File "/home/jordan/.conan2/p/b/cpyth43a849bd4dab6/p/lib/python3.10/ssl.py", line 495, in __new__
    self = _SSLContext.__new__(cls, protocol)
ssl.SSLError: [CONF: MODULE_INITIALIZATION_ERROR] module initialization error (_ssl.c:3051)

cpython/3.10.0 (test package): WARN: Module 'ssl' does not work, but should have worked
ERROR: cpython/3.10.0 (test package): Error in test() method, line 197
	self._test_module("ssl", True)
while calling '_test_module', line 154
	self.run(f"{self._python} {self.source_folder}/test_package.py -b {self.build_folder} -t {module}", env="conanrun")
```

I'm on Fedora. Perhaps it would have something to do with that.

jwillikers:
> Fyi, using Conan V2 the test package fails to build for me when testing the SSL

Aha, found the problem: https://projects.blender.org/blender/blender/issues/114452#issuecomment-1059165

Setting `OPENSSL_CONF=/dev/null` does indeed solve the problem.

recipes/cpython/all/test_package/conanfile.py
```diff
@@ -18,175 +25,200 @@ def __init__(self, debug, pymalloc, unicode):
...
+                self._test_module("sqlite3", self._cpython_option("with_sqlite3"))
                 self._test_module("decimal", True)
                 self._test_module("ctypes", True)
-
-            if tools.is_apple_os(self.settings.os) and not self.options["cpython"].shared:
-                self.output.info("Not testing the module, because these seem not to work on apple when cpython is built as a static library")
+                skip_ssl_test = is_msvc(self) and self._py_version < "3.8" and self._cpython_option("shared")
+                if not skip_ssl_test:
+                    # Unsure cause of failure in this oddly specific combo, but these versions are EOL so not concerned with fixing.
+                    self._test_module("ssl", True)
```
jwillikers, line 197:
```suggestion
                    env = Environment()
                    if self.settings.os != "Windows":
                        env.define_path("OPENSSL_CONF", os.path.join(os.sep, "dev", "null"))
                    with env.vars(self).apply():
                        self._test_module("ssl", True)
```

recipes/cpython/all/test_package/conanfile.py
```diff
@@ -1,10 +1,17 @@
...
-import re
-import shutil
+from io import StringIO
 
+from conan import ConanFile, conan_version
+from conan.errors import ConanException
+from conan.tools.apple import is_apple_os
+from conan.tools.build import can_run
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.env import VirtualRunEnv
```
jwillikers, line 9:
```suggestion
from conan.tools.env import Environment, VirtualRunEnv
```

jwillikers (changes requested):
This fixes the issue with the SSL module test for me on Fedora.

recipes/cpython/all/test_package/conanfile.py
Ahajha: Will also apply this shortly, just getting to the other issue.

recipes/cpython/all/test_v1_package/conanfile.py
Ahajha: Looks like this is (was?) a known issue: https://github.com/pypa/pip/issues/3055

Maybe this could be fixed by setting Windows's temp dir to a shorter path, I'll ask someone if that's possible in CI.

Ahajha: And also it looks like it's intermittent, which is annoying.

Ahajha: Unfortunately short_paths doesn't work since this is in a local build folder. I added a bandaid fix: https://github.com/conan-io/conan-center-index/pull/21387/commits/73e25ce7154d8eadc4821e3f5fcd9eb315716a87

Essentially shorten the directory slightly (as much as we can control) to buy us a few more characters before hitting the max path limit.

jwillikers (approved):

uilianries:
@Ahajha Thank you for your great effort on this PR, CPython is a challenge. 

I see you are still supporting Python 2.7, which is no longer supported, including for bugfixes. Such supports increases the recipe complexity, which is really bad for maintenance. Plus, right now there is no usage for cpython 2.x in ConanCenterIndex.

I would ask to drop that support, it would simplify the recipe by a lot, and will avoid future patches for a dead version.

Ahajha:
@uilianries I spoke with Ruben about my concerns with EOL versions, specifically 2.7. I think there could be value in having these recipes available _somewhere_ (My thoughts are in the PR description), but maybe CCI isn't the place. The maintenance burden actually isn't that high, build system-wise it's fairly similar to other versions.

I could split 2.x into a subdir, or maybe just remove it outright since I believe the branch will stick around here, and the commits on this branch haven't been squashed. So if someone needs that version, they can just dig up the commit. How does that sound?

Additionally, if we're removing versions because they're EOL, should I go ahead and drop 3.7 as well?

uilianries:
@Ahajha Thank you for your considerations and quick answer!

> I could split 2.x into a subdir, or maybe just remove it outright since I believe the branch will stick around here, and the commits on this branch haven't been squashed. So if someone needs that version, they can just dig up the commit. How does that sound?

The PR indeed will be available always, so we still can bring 2.7 in case someone really needs and open an issue specifying the case. So keeping only 3.x for now looks safer, less work, less maintenance. 

> Additionally, if we're removing versions because they're EOL, should I go ahead and drop 3.7 as well?

Yes, please. Good catch about it. https://devguide.python.org/versions/

Ahajha:
Should be good to go now, I think I got all of the pre-3.8 logic.

jwillikers (approved):

recipes/cpython/all/conanfile.py
```diff
@@ -53,286 +56,294 @@ class CPythonConan(ConanFile):
...
-            conf_args.append("--without-gcc")
+        if Version(self.version) >= "3.10":
+            tc.configure_args.append("--disable-test-modules")
+        if self.options.get_safe("with_sqlite3"):
+            tc.configure_args.append("--enable-loadable-sqlite-extensions={}".format(
+                yes_no(not self.dependencies["sqlite3"].options.omit_load_extension)
+            ))
+        if self.settings.compiler == "intel-cc":
+            tc.configure_args.append("--with-icc")
+        if os.environ.get("CC") or self.settings.compiler != "gcc":
```
uilianries, line 203:
```suggestion
        if self.settings.compiler != "gcc":
```

Avoid using external environment variables, it will result in not reproducible scenarios.

recipes/cpython/all/conanfile.py
```diff
@@ -53,286 +56,294 @@ class CPythonConan(ConanFile):
...
+        if not is_apple_os(self):
+            tc.extra_ldflags.append('-Wl,--as-needed')
+
+        tc.generate()
+
+        deps = AutotoolsDeps(self)
+        deps.generate()
+
+    def generate(self):
+        VirtualRunEnv(self).generate(scope="build")
```
uilianries, line 230: Which requirement do you need here, that requires VirtualRunEnv?

recipes/cpython/all/conanfile.py
```diff
@@ -341,135 +352,110 @@ def project_build(name):
...
         else:
-            autotools = self._configure_autotools()
+            autotools = Autotools(self)
+            # For debugging configure errors
+            try:
+                autotools.configure()
+            except ConanException:
+                with open(os.path.join(self.build_folder, "config.log"), 'r') as f:
+                    self.output.info(f.read())
+                raise
```
uilianries, line 417:
```suggestion
            autotools.configure()
```

Let's keep it simple, please. Who is working on the recipe is able to access the build folder and read any file.

recipes/cpython/all/conanfile.py
```diff
@@ -481,62 +467,80 @@ def _msvc_package_layout(self):
...
+            rm(self, "vcruntime*", os.path.join(self.package_folder, "bin"), recursive=True)
         else:
-            autotools = self._configure_autotools()
-            autotools.install()
-            tools.rmdir(os.path.join(self.package_folder, "lib", "pkgconfig"))
-            tools.rmdir(os.path.join(self.package_folder, "share"))
+            if (os.path.isfile(os.path.join(self.package_folder, "lib"))):
+                # FIXME not sure where this file comes from
+                self.output.info(f"{os.path.join(self.package_folder, 'lib')} exists, but it shouldn't.")
+                rm(self, "lib", self.package_folder)
```
uilianries, line 537:
```suggestion
            # FIXME not sure where this file comes from
            rm(self, "lib", self.package_folder)
```
Let's simplify: The `rm` only removes files, and does not raise error for file not found.

recipes/cpython/all/conanfile.py
```diff
@@ -481,62 +467,80 @@ def _msvc_package_layout(self):
...
+            autotools = Autotools(self)
+            # FIXME: Autotools.install() always adds DESTDIR, we don't want this argument.
+            # Use .make() directly instead
+            autotools.make(target="altinstall")
```
uilianries, line 541:
```suggestion
            autotools.install(args=["DESTDIR="])
```

https://docs.conan.io/2/reference/tools/gnu/autotools.html#conan.tools.gnu.autotools.Autotools.install

recipes/cpython/all/conanfile.py
```diff
@@ -692,47 +684,45 @@ def package_info(self):
...
+            self.cpp_info.components["_hidden"].includedirs = []
             self.cpp_info.components["_hidden"].libdirs = []
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                self.cpp_info.components["_hidden"].system_libs.append("nsl")
 
         if self.options.env_vars:
             bindir = os.path.join(self.package_folder, "bin")
-            self.output.info("Appending PATH environment variable: {}".format(bindir))
+            self.output.info(f"Appending PATH environment variable: {bindir}")
             self.env_info.PATH.append(bindir)
```
uilianries, line 699:
```suggestion
            self.env_info.PATH.append(bindir)
            self.buildenv_info.prepend_path("PATH", bindir)
            self.runenv_info.prepend_path("PATH", bindir)
```

Env info is Conan 1.x and will be deprecated in the future. In Conan 2.x you should use runenv or buildenv.For Python I would say both cases are valid because I see python scripts during build time, only to generate C/C++ code, but also I saw some packaged tools that are python scripts too.

recipes/cpython/all/conanfile.py
```diff
@@ -692,47 +684,45 @@ def package_info(self):
...
-            python_root = ""
-        else:
-            python_root = self.package_folder
-            if self.options.env_vars:
-                self.output.info("Setting PYTHON_ROOT environment variable: {}".format(python_root))
-                self.env_info.PYTHON_ROOT = python_root
+        python_root = self.package_folder
+        if self.options.env_vars:
+            self.output.info(f"Setting PYTHON_ROOT environment variable: {python_root}")
+            self.env_info.PYTHON_ROOT = python_root
```
uilianries, line 726:
```suggestion
            self.env_info.PYTHON_ROOT = python_root
            self.buildenv_info.define_path("PYTHON_ROOT", python_root)
```

uilianries (commented):
I reviewed the PR now, it's looking good, only minor details to be checked.

recipes/cpython/all/conanfile.py
```diff
@@ -53,286 +56,294 @@ class CPythonConan(ConanFile):
...
-        return "source_subfolder"
-
-    @property
-    def _version_number_only(self):
-        return re.match(r"^([0-9.]+)", self.version).group(1)
-
-    @property
-    def _version_tuple(self):
-        return tuple(self._version_number_only.split("."))
+    short_paths = True
```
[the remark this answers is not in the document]
Ahajha: Python packages compiled versions of all of the modules shipped in the stdlib in a `__pycache__` folder. These paths end up being very long, you can look at some of the hooks in the earlier runs of the PR to see such file names (`./lib/python3.10/site-packages/setuptools/_distutils/__pycache__/versionpredicate.cpython-310.pyc`, for example).

It would be a bad idea to remove these, as I believe they would just get regenerated when they are used, breaking cache immutability.

recipes/cpython/all/conanfile.py
Ahajha: This was here prior, I'll double check that it doesn't break anything if used.

recipes/cpython/all/conanfile.py
Ahajha: The build runs Python once it's done to do some sanity checks, so we need access to the run environment in the build.

recipes/cpython/all/conanfile.py
Ahajha: Fair enough, this was mainly to debug the xvmc issue in CCI.

recipes/cpython/all/conanfile.py
Ahajha: I might remove this hack altogether, I don't know if it ever actually worked. I'll check through some logs to see.

recipes/cpython/all/conanfile.py
Ahajha: I'll double check this, I remember this took us a while to figure out so I want to make sure it doesn't break. Also not sure if we should be using `install` or `altinstall` here, `altinstall` I believe is intended for use when there are multiple Python installations, which is technically the case here.

recipes/cpython/all/conanfile.py
Ahajha: I think I can actually remove this condition entirely, seems that the flag was removed in 3.8: https://github.com/python/cpython/blob/main/Misc%2FNEWS.d%2F3.8.0a1.rst#L7485

uilianries: Much better, please, remove then. Thank you for checking it!

recipes/cpython/all/conanfile.py
Ahajha: It seems intermittent, and actually hasn't happened recently (only checked some recent logs, not going through all of them). I think it's best to just remove it, if it comes up we can just rerun the job.

recipes/cpython/all/conanfile.py
Ahajha: Seems to work!

recipes/cpython/all/conanfile.py
Ahajha: Done, removed the whole condition.

recipes/cpython/all/conanfile.py
Ahajha: Done

recipes/cpython/all/conanfile.py
Ahajha: I added these for all the existing variables, as well as a note regarding the old version's removal at some point.

Ahajha:
@uilianries I believe I addressed all the review points.

My only (very minor) concern left is regarding the environment variables, I think we may be able to simplify one or more of:
1. The variables themselves
2. The conf info
3. The conditions behind them and their related conf info being defined (should we use an env_vars option, or have them be unconditionally defined? I don't know if other packages do this) (others are conditional on Windows or Apple, maybe just have them be unconditional since I don't think the conditions are actually correct)
4. The usage of both the env vars and the conf info (for example, we no longer need to specify the full path to the python executable, we can just run `python` and it will work)

It works for now, so I think it's fine to leave as it, but maybe it's worth a revisit. I can add it to my TODO issue (that I believe you saw) if you agree.

recipes/cpython/all/conanfile.py
Ahajha: Done

jwillikers (approved):

RubenRBS (approved):

jwillikers:
image

Ahajha:
Thanks to everyone involved for helping this along! I learned a lot from working on this PR for sure :)

ilatypov:
`python -m pip` fails missing pyexpat, socket on Windows. Sigh.  The original submission took pains to use conan's dependencies for Python 2.  I don't know if that worked (I cannot even download cpython 2.7.18 because it is not compatible with the new Conan API).  Am I doing the right thing?  If I do, could that be added as a smoke test (currently failing)?

I guess the use of cpython from conan may have a limited user base because using conan implies using Python.  But it's needed for bootstrapping in the corporate requirements of using approved binary repositories (such as Conan Center).  And of course the conan way of building C Python could eventually substitute the less-straightforward build systems, enabling C Python development self-bootstrapping.

```
bin\python -m pip
Traceback (most recent call last):
[...]
  File "C:\Users\XXX\cpython\bin\lib\site-packages\pip\_vendor\pkg_resources\__init__.py", line 35, in <module>
    import plistlib
  File "C:\Users\XXX\cpython\bin\Lib\plistlib.py", line 65, in <module>
    from xml.parsers.expat import ParserCreate
  File "C:\Users\XXX\cpython\bin\Lib\xml\parsers\expat.py", line 4, in <module>
    from pyexpat import *
ModuleNotFoundError: No module named 'pyexpat'
```

> For visual studio, the patches are needed to use conan packages instead of python binaries.
See [this repo](https://github.com/python/cpython-source-deps) and [this repo](https://github.com/python/cpython-bin-deps).

> For python 2.7, I had to apply some fixes to make it build with more recent visual studios.
These patches will never hit upstream, as 2.7 is dead.

_Originally posted by @madebr in https://github.com/conan-io/conan-center-index/issues/1510#issuecomment-974257101_

Ahajha:
@ilatypov There was discussion about leaving Python 2 support, but we ended up dropping it. That said, I'm more than willing to re-add it if you think there's a use case for it in Conan Center.

Regarding the failing test, do you know if this ever worked?

Also regarding "The original submission took pains to use conan's dependencies for Python 2" - Is this just saying that Python 2 was originally supported, or that it used Conan dependencies? If the latter, I don't think anything has changed - the recipe still goes to great lengths to use Conan's dependencies.

I've seen slow but steady adoption of the cpython package within ConanCenter, mainly in places where Conan's minimum Python version isn't enough. When I started this PR (branching off of a few others), I had a use case of easily switching Python versions within a multi-repo application as an alternative to essentially requiring some containerization.

All that said, I'm happy to look into the failing test and re-add Python 2.

ilatypov:
I hijacked this thread that showed amazing collaboration.  I did not have concerns about resurrecting conan's Python 2 and, unfortunately, I don't know if the smoke test would work with conan's Python 2.  I am concerned that Python 3 was failing where it could be used for setting up access to any other package through the trust to the binaries published in Conan Center.

Ahajha:
I'm a little confused as to your exact problem, but I have a guess as to the problem:

The `_supports_modules` attribute is representative of a Python limitation, not a Conan one. If statically linking Python on Windows, no extension modules can be built, so perhaps that's why you're seeing this issue? You could set `cpython/*:shared=True` if you haven't already, perhaps that's what you're looking for?

ilatypov:
I am not linking, just using Python from Conan Center.  It's smoke test above (`-m pip`) is failing.

Ahajha:
Let me clarify - Regardless of what you're doing with the package, on Windows, if `cpython/*:shared=False`, which is the default, then the extension modules (such as expat) will not be available. So setting `cpython/*:shared=True` I think will fix your issue.

ilatypov:
It worked for 3.8.12 here, thank you.  I am new to conan, so I forgot that Conan Center is not the ultimate goal of the project, but only its optimization.  Building from source took only minutes and worked great.
```
PS> conan install --requires=cpython/3.8.12 -o shared=True -u --build=missing -r opentext -vvv --deployer=full_deploy
[...]
PS> .\full_deploy\host\cpython\3.8.12\Release\x86_64\bin\python.exe -m pip
Usage:
  C:\Users\XXX\full_deploy\host\cpython\3.8.12\Release\x86_64\bin\python.exe -m pip <command> [options]

Commands:
[...]
```

PR #18642, opened 2023-07-19
libidn2: migrate to Conan v2, add shared MSVC build

valgur:

PR #21389, opened 2023-11-26
getdns: migrate to Conan v2,  bump to v1.7.3

valgur:
- Requires #18642


jwillikers (approved):

uilianries:
I did some changes over this PR, including:

- Replaced cmake wrapper by CMAKE_PROJECT_INCLUDE to include external dependencies
- Simplified test package to avoid fetching the network
- Remove option for tls: GnuTLS is only supported on Linux and needs libnuma to work. Checking other package managers, only Gentoo is supporting it, and still, has a separated package for gnutls-numa
- Remove stub option and configured to True always. Other package managers are using True as well, and CCI has no libunbound available
- Enable libidn2 by default, following the upstream
- Injected compiler flag to ignore pointer error on Clang (reported to the upstream)
- Imported libidn2 definition when building on Windows to avoid linkage error.

uilianries (dismissed):
LGTM

valgur:
LGTM as well. Thanks!

AbrilRBS (approved):

PR #21425, opened 2023-11-27
Add missing requirement for libhal

kammce:
libhal 2.Y.Z requires tl-function-ref as well as boost-leaf.

Specify library name and version:  **libhal/***

---

### DIFF
--- a/recipes/libhal/all/conanfile.py
+++ b/recipes/libhal/all/conanfile.py
@@ -44,6 +44,7 @@ class LibHALConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
+        self.requires("tl-function-ref/1.0.0")
         # NOTE from the author, kammce, and CCI maintainers:
         # although boost-leaf is deprecated, we've kept it for 2.x versions,
         # don't update it as upstream code won't work with boost itself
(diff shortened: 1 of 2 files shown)


RubenRBS (dismissed):

RubenRBS:
Thanks! Any reason this was not caught by the test package originally?

kammce:
> Thanks! Any reason this was not caught by the test package originally?

My test package just tests that you can bring in basic libraries, but doesn't check all of them. Where as my unit tests, which would catch this, were not brought over in the previous upgrade of libhal. So I missed this in the last review. 

Should I add the code to perform the unit tests in this recipe? I can also update the test package to ensure that this part of the library is checked.

kammce:
Updated the test package to depend on a header file that depends on tl-reference

RubenRBS (approved):

RubenRBS:
> Should I add the code to perform the unit tests in this recipe?

`test_package`s are not meant to run unittestings, only that the package itself was correctly created. It has its downsides as for example not catching this case earlier, but it's an approach that otherwise cleanly separates concerns and responsabilities for who tests what :)

> I can also update the test package to ensure that this part of the library is checked.

Thanks for doing it!

kammce:
> > Should I add the code to perform the unit tests in this recipe?
> 
> `test_package`s are not meant to run unittestings, only that the package itself was correctly created. It has its downsides as for example not catching this case earlier, but it's an approach that otherwise cleanly separates concerns and responsabilities for who tests what :)
> 
> > I can also update the test package to ensure that this part of the library is checked.
> 
> Thanks for doing it!

Oh I meant to say, that, after we download the sources, we could build unit tests as part of the build. But I'm happy to just have test_package manage this.

toge (approved):

PR #21426, opened 2023-11-27
CPython: Update deps (1.x only)

Ahajha:

Since I was doing a small PR anyways, I wanted to add some of the simple Conan 2.0 migration steps to ease the diff in future PRs.

Current min version is set to 1.53.0, I don't know if the current version actually needs 1.53.0, probably around 1.51, but it will need to be 1.53.0 eventually anyways.

---


Ahajha:
@RubenRBS :)

recipes/cpython/all/conanfile.py
```diff
@@ -170,42 +177,42 @@ def source(self):
...
+            elif Version(self._version_number_only) < "3.10":
                 self.requires("mpdecimal/2.5.0")
             else:
                 self.requires("mpdecimal/2.5.0")  # FIXME: no 2.5.1 to troubleshoot apple
         if self.settings.os != "Windows":
-            if not tools.is_apple_os(self.settings.os):
-                self.requires("libuuid/1.0.3")
-            self.requires("libxcrypt/4.4.25")
+            if not is_apple_os(self):
+                self.requires("util-linux-libuuid/2.39.2", force=True)
```
RubenRBS, line 197: Checking the PR now before merging, why is force needed here?

RubenRBS (commented):
Thanks! Now that the ncruses recipe has been merged, we can finally focus on this PR :)

I have only 1 question about the usage of the force traitm otherwise looks good!

recipes/cpython/all/conanfile.py
Ahajha: It's getting a version conflict with another dependency that uses util-linux-libuuid, in 2.x this doesn't seem to force the version? Is there a better way to do this?

Ahajha:
I see CPython 2.x is failing, I'll try briefly to get it working again but I don't think it's worth too much effort currently.

RubenRBS:
We can merge this as-is on our end, no worries about that, and work on v2 compat at a later point :)

Ahajha:
> We can merge this as-is on our end, no worries about that, and work on v2 compat at a later point :)

I mean CPython 2.7.18, not Conan 2 :) I'll try briefly to see if I can find the cause, if not, I believe there was separate discussion about dropping support for it, at the very least it's low priority. It's also causing other packages to not be built, which I imagine we want.

recipes/cpython/all/conanfile.py
samuel-emrys: @Ahajha which dependency is it conflicting with? I can't see anything in the logs - perhaps you need to update your local cache?

The normal resolution to this kind of issue is to bump the `util-linux-libuuid` version on the other dependency so that both are at the most recent version.

Ahajha: Ah, looks like it was a stale cache. It came from `fontconfig`, which comes from `tk`. I'll remove it.

Ahajha:
Closing this in favor of #21387

PR #21433, opened 2023-11-28
xmlrpc-epi: add recipe

### LINKED ISSUE #21388: [request] xmlrpc-epi/0.54.2
### Package Name/Version

xmlrpc-epi/0.54.2

### Webpage

https://xmlrpc-epi.sourceforge.net/

### Source code

https://sourceforge.net/projects/xmlrpc-epi/files/

From its homepage, xmlrpc-epi is described as...
> ...an implementation of the [xmlrpc protocol](http://www.xmlrpc.org/) in C.

This library is fairly old but linux distros such as Debian and Fedora keep a version alive and available in their main repositories replete with minor patches.

bennettgoble:
Specify library name and version:  **xmlrpc-epi/0.54.2**

This changeset adds a Conan recipe for the [xmlrpc-epi](https://xmlrpc-epi.sourceforge.net/) C library. While its upstream has not been updated since 2011 it still receives security patches from linux distros such as Debian.

The library is used by the [Second Life client](https://github.com/secondlife/viewer) and other software maintained by [Linden Lab](https://lindenlab.com/).

Thank you for the consideration, please let me know if anything needs to change!

Closes #21388

---

- [x] ~I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.~ _Skipped: Not compatible with conan2?_

### DIFF
--- a/recipes/xmlrpc-epi/all/CMakeLists.txt
+++ b/recipes/xmlrpc-epi/all/CMakeLists.txt
@@ -1,0 +1,27 @@
+cmake_minimum_required(VERSION 3.12)
+
+project(xmlrpc-epi C)
+
+find_package(EXPAT REQUIRED)
+
+if(MSVC AND BUILD_SHARED_LIBS)
+    set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
+endif()
+
+file(GLOB HEADERS "src/*.h")
+file(GLOB SOURCES "src/*.c")
+
+add_library(xmlrpc-epi ${SOURCES} ${HEADERS})
+
+target_include_directories(xmlrpc-epi PUBLIC include)
+
+set(LINK_LIBRARIES expat::expat)
+if (WIN32 OR APPLE)
+    find_package(ICONV REQUIRED)
+    list(APPEND LINK_LIBRARIES Iconv::Iconv)
+endif (WIN32)
+
+target_link_libraries(xmlrpc-epi ${LINK_LIBRARIES})
+
+install(TARGETS xmlrpc-epi)
+install(FILES ${HEADERS} DESTINATION include)
--- a/recipes/xmlrpc-epi/all/CMakeLists.txt
+++ b/recipes/xmlrpc-epi/all/CMakeLists.txt
@@ -8,12 +8,22 @@ if(MSVC AND BUILD_SHARED_LIBS)
     set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
 endif()
 
-file(GLOB HEADERS "src/*.h")
 file(GLOB SOURCES "src/*.c")
+file(GLOB PRIVATE_HEADERS "src‌/*_private.h")
+set(PUBLIC_HEADERS
+    "src/base64.h"
+    "src/encodings.h"
+    "src/queue.h"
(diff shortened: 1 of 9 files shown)


bennettgoble:
I've switched this PR to draft while I investigate windows build failures further.

recipes/xmlrpc-epi/all/CMakeLists.txt
uilianries, line 1: I see the original project uses autotools. Could you use it instead? It would be nice avoiding it, otherwise, we would become the maintainers.

recipes/xmlrpc-epi/all/conandata.yml
```diff
@@ -0,0 +1,13 @@
...
+  "0.54.2":
+    url: "https://downloads.sourceforge.net/project/xmlrpc-epi/xmlrpc-epi-base/0.54.2/xmlrpc-epi-0.54.2.tar.bz2"
+    sha256: "c74ef8fb680b140890138a82f37619714b67f69025a775b9ba2009d62cded0b8"
+patches:
+  "0.54.2":
+    - patch_file: "patches/0001-debian.patch"
+      patch_type: "vulnerability"
+      patch_description: "CVE-2016-6296: Heap buffer overflow vulnerability in simplestring_addn"
+      patch_source: https://sources.debian.org/src/xmlrpc-epi/0.54.2-2/debian/patches/debian.patch/
+    - patch_file: "patches/0002-msvc.patch"
```
uilianries, line 11: Do you have a reference about this patch in the upstream

recipes/xmlrpc-epi/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        tc.generate()
+        deps = CMakeDeps(self)
+        deps.generate()
+
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+        if self.settings.build_type == "Debug":
+            # Remove inlines from source files so that we can link the debug build
+            replace_in_file(self, os.path.join(self.source_folder, "src/xml_to_soap.c"), "inline", "")
+            replace_in_file(self, os.path.join(self.source_folder, "src/xmlrpc_introspection.c"), "inline", "")
```
uilianries, line 64: Please, do not enforce a behavior that's not expected by the upstream. In case it's not supported, only use `validate()` to raise `ConanInvalidConfiguration`

uilianries (commented):
Thank you for your pull request, please, consider this review.

recipes/xmlrpc-epi/all/CMakeLists.txt
bennettgoble: @uilianries I've made a version of this branch which uses the original autotools. Unfortunately the autotools configuration cannot build on windows without some rather tedious workarounds due to https://github.com/conan-io/conan/issues/12784.

The original project has some very old visual studio project files from ~2010 which can be upgraded with `devenv` and then modified in the conanfile (inject architectures, etc.) however my concern is that it seems like maintaining all these workarounds is more work than using a short cmake file for this small c library.

With that context what would be the preferred route?

bennettgoble:
Closing this for now, at least until https://github.com/conan-io/conan/issues/12784 is resolved. Thanks for the review.

PR #21436, opened 2023-11-28
lksctp-tools: add recipe

temap:
Specify library name and version:  **lksctp-tools/1.0.19**

---

### DIFF
--- a/recipes/lksctp-tools/all/conanfile.py
+++ b/recipes/lksctp-tools/all/conanfile.py
@@ -1,0 +1,75 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.files import get, chdir, copy, rm, rmdir
+from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.layout import basic_layout
+import os
+
+required_conan_version = ">=1.53.0"
+
+
+class LKSCTPToolsConan(ConanFile):
+    name = "lksctp-tools"
+    settings = "os", "arch", "compiler", "build_type"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/sctp/lksctp-tools"
+    description = "The Linux Kernel Stream Control Transmission Protocol (lksctp) project"
+    package_type = "library"
+    topics = ("sctp")
+    license = ("GPL-2.0", "LGPL-2.1")
+
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def validate(self):
+        if self.settings.os != "Linux":
+            raise ConanInvalidConfiguration(f"{self.ref} is only available on Linux")
+
+    def layout(self):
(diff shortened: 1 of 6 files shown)


RubenRBS (commented):
Hi! Thanks a lot for your contribution. The recipe itself looks good :) - taking care of the hook error should get the PR most of the way there!

temap:
> Hi! Thanks a lot for your contribution. The recipe itself looks good :) - taking care of the hook error should get the PR most of the way there!

Hi, all issues were resolved. Thanks!

samuel-emrys (approved):

recipes/lksctp-tools/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+
+class LKSCTPToolsConan(ConanFile):
+    name = "lksctp-tools"
+    settings = "os", "arch", "compiler", "build_type"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/sctp/lksctp-tools"
+    description = "The Linux Kernel Stream Control Transmission Protocol (lksctp) project"
+    package_type = "library"
+    topics = ("sctp")
+    license = "GPL-2.0, LGPL-2.1"
```
uilianries, line 19:
```suggestion
    license = ("GPL-2.0", "LGPL-2.1")
```

Each license should be a separated string, so we can create a list.

recipes/lksctp-tools/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = AutotoolsToolchain(self)
+        yes_no = lambda v: "yes" if v else "no"
+        tc.configure_args.extend([
+            "--disable-tests",
+            "--enable-static={}".format(yes_no(not self.options.shared)),
```
uilianries, line 52:
```suggestion
        tc.configure_args.extend([
            "--disable-tests",
```

--enable-static, --enable-shared, --disable-static and --disable-shared are not needed as you are using AutotoolsToolchain: https://docs.conan.io/2/reference/tools/gnu/autotoolstoolchain.html#customization

recipes/lksctp-tools/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+            rmdir(self, os.path.join(self.package_folder, "bin"))
+            rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+            rmdir(self, os.path.join(self.package_folder, "lib", "lksctp-tools"))
+            rmdir(self, os.path.join(self.package_folder, "share"))
+            rm(self, "*.la", os.path.join(self.package_folder, "lib"))
+
+        copy(self, pattern="COPYING", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+
+    def package_info(self):
+        self.cpp_info.libs = collect_libs(self)
```
uilianries, line 77:
```suggestion
        self.cpp_info.libs = collect_libs(self)
```
Could you please use the expected libraries names instead of `collect_libs`? We try to avoid collect libs because we have no guess about what expect, and in case providing multiple libraries, collect_libs does not respect the link order.

uilianries (commented):
Looking good, only small changes.

recipes/lksctp-tools/all/conanfile.py
temap: @uilianries I removed the `collect_libs`

uilianries (approved):
LGTM

RubenRBS (approved):

PR #21448, opened 2023-11-28
[docs] Allow boost namespace for non official projects

uilianries:
After talking to Conan team about a non official boost library, having no intention to be integrated to official boost, and finding that even cmake targets and headers are using the name boost, it would be better allow these specific cases to use the project name, including the boost prefix as package name. Otherwise, it would require more changes, like the cmake target and cmake name.

In case the project changes, and want to be part to boost, then, it should be marked as `deprecated` as soon as becomes official in Boost. 

Related to https://github.com/conan-io/conan-center-index/pull/21303

---


uilianries:
I'll not fix the dead link now, but I opened the issue to do it later: https://github.com/conan-io/conan-center-index/issues/21449

valgur (approved):

RubenRBS (approved):

PR #21059, opened 2023-11-10
vulkan-utility-libraries: add new recipe

valgur:
https://github.com/KhronosGroup/Vulkan-Utility-Libraries

Required for newer versions of `vulkan-validationlayers`.

PR #21495, opened 2023-11-30
vulkan-validationlayers: update versions, improve package_type and test_package

valgur:
The package really only exports a `VkLayer_khronos_validation` runtime library and a .json manifest file, especially so after the utility library has been split off into `vulkan-utility-libraries` (#21059).

I adjusted the `package_type` accordingly to `application` and updated `test_package` to test the correct thing.

I intended to update the list of versions to match the other recently updated Vulkan and SPIRV packages (meaning only v1.3.239+ SDK versions), but:
- 1.3.261.1 was not available in `vulkan-loader` for `test_package`.
- 1.3.250.0 was not available in `spirv-headers` and `spirv-tools`.
- 1.3.239.0 failed to load the validation layers in `test_package` and I'm not familiar enough with Vulkan to debug it.
- 1.3.268.0 requires #21059.


neobrain:
FTR this has served me well while debugging some Vulkan issues under macOS with MoltenVK. Happy to report all worked fine, thanks!

valgur:
@uilianries

uilianries:
@valgur Please, let's focus on #21059 first. I had a further review yesterday and I may continue today.

PR #21512, opened 2023-11-30
proj: add v9.4.0, drop old versions

### LINKED ISSUE #23482: PROJ 9.4.0
### What is your question?

Hi!
I would like to know when PROJ 9.4.0 will be available in Conancenter
Thanks,

valgur:
* Resolves #23482.


recipes/proj/all/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "9.3.1":
+    url: "https://github.com/OSGeo/PROJ/archive/9.3.0.tar.gz"
+    sha256: "f48c334eaf56c38d681bcfa37f188f422a562f45a66e6e646a79b4249641ecdb"
```
ErixenCruz, line 4: Labeled 9.3.1, but this is just 9.3.0 copy and pasted. should be 
```yaml
  "9.3.1":
    url: "https://github.com/OSGeo/PROJ/releases/download/9.3.1/proj-9.3.1.tar.gz"
    sha256: "b0f919cb9e1f42f803a3e616c2b63a78e4d81ecfaed80978d570d3a5e29d10bc"
```
The cmake targets patch from 9.2.1 still applies luckily.

I was going to make a PR just for 9.3.1 because it has some bug fixes that we would like to use. Should we just wait for this PR to get in?

valgur: Ugh. Thanks!
Yes, feel free to create a separate PR for the version bump. It might be possible to auto-merge it if it does not require any new patches.

SpaceIm:
Sorry I don't see the point of this modification of patches. I prefer to have patches per version instead of an external cmake file injecting to some magic line in upstream CMakeLists. It's hard to reason about CMake logic with this kind of external cmake file.
Currently we just have to maintain a small patch, is it really a big deal? Why not just submit a PR upstream (by defining `SQLite::SQLite3` in their custom Findsqlite?), so that we won't need a patch anymore.

jwillikers (approved):

RubenRBS (approved):
Thanks!

recipes/proj/all/conanfile.py
```diff
@@ -109,6 +110,7 @@ def generate(self):
             # Workaround for: https://github.com/conan-io/conan/issues/13560
             libdirs_host = [l for dependency in self.dependencies.host.values() for l in dependency.cpp_info.aggregated_components().libdirs]
             tc.variables["CMAKE_BUILD_RPATH"] = ";".join(libdirs_host)
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
SpaceIm, line 113: I'm curious, why is it needed?

perseoGI:
Hi @valgur, thanks for your contribution!
PROJ 9.4.1 was released 5 days ago: https://github.com/OSGeo/PROJ/compare/9.4.0...9.4.1
Maybe we can use this PR to add the latest.

recipes/proj/all/conanfile.py
RubenRBS: Afaik Valgur has been adding these in case upstream ever adds an option before project etc that can't be then controlled from Conan in old Cmake versions

SpaceIm: It doesn't work like this. Policy doesn't depend on CMake version you are using but on CMake version defined in cmake_minimum_required().
This line is very likely useless.

SpaceIm: Looking at conancenter history, I see this policy injected more & more quite randomly without rational. It's quite misleading.

valgur:
PROJ 9.3.0 and older set
cmake_minimum_required(VERSION 3.9)
which in turn sets CMP0077 to OLD by default. This line ensures that
tc.variables[] works as expected in all cases.

valgur:
> This line is very likely useless (or if it is planned to drop this logic
in all recipes, it should live in conan client code I guess, but again it's
useless for properly formed CMakeLists since conan toolchain translates
tc.variables to CACHE variables, therefore CMP0077 doesn't matter).

You are misunderstanding the issue being fixed by CMP0077 in some important
ways. When it’s set to OLD, all set(CACHE) and option() commands in
CMakeLists.txt will overwrite the value previously set by CMakeToolchain,
causing them to be ineffective. There is no such thing as a properly-formed
CMakeLists.txt in this case - only the value of this policy matters.

Believe me, this is a very real issue or at least a potential pitfall in
many recipes. I encourage you to test it yourself, if you’re having doubts.

I have been adding it for all recipes where CMake policy version is
configured to 3.11 or older, just in case. Even if these projects might not
be setting CACHE vars in the current version, they easily can switch to
using them in later versions, which is very easy to miss.

I agree that this would be better handled by the Conan client, but I
suppose setting it by default for all projects might cause unexpected
regressions in unrelated places, making it a bit risky. And even then it
would take a while for the change to become effective for CCI.

perseoGI (dismissed):
LGTM

recipes/proj/all/conanfile.py
```diff
@@ -61,13 +60,15 @@ def layout(self):
 
     def requirements(self):
         self.requires("nlohmann_json/3.11.3")
-        self.requires("sqlite3/3.44.2")
+        self.requires("sqlite3/3.45.3")
```
perseoGI, line 63:
```suggestion
        self.requires("sqlite3/3.44.2")
```

perseoGI (approved):
LGTM

recipes/proj/all/conanfile.py
jwillikers: I agree with @SpaceIm here.

RubenRBS (approved):

PR #21520, opened 2023-11-30
modern-cpp-kafka: add package_type + reset cmake names to default value

SpaceIm:
- add package_type
- do not invent unofficial target name (there is no config file upstream, and the target name representing interface lib in their CMakeLists is not even the same as suggested in this recipe)
- empty cpp_info.bindirs & libdirs
- bump librdkafka

---


RubenRBS (approved):
Approving this, I think we all agree that we should follow upstream decisions (or lack thereof), and in this case it means to remove those lines, thanks a lot!

PR #21522, opened 2023-11-30
diligent-core: migrate to Conan v2

valgur:
Requires
- #21495


MikeLankamp:
Hi @valgur, what's still needed to merge this? I'm trying to using diligent-core/2.5.1 with Conan 2, but it's failing because it has a Conan2-incompatible cmake/2.24.2 tool_requires, which I can't override. This PR seems to fix all of that.

Issue #21523, opened 2023-11-30
[question] raylib 4.0.0 cannot be built


robiarth:
### What is your question?

Hi, i'm trying to compile a simple test project with raylib on windows with Conan 2.0, and I have some problems with the build

Here is my error :
[build error raylib.txt](https://github.com/conan-io/conan-center-index/files/13515477/build.error.raylib.txt)

Do you have any solution ?

SpaceIm:
Try https://github.com/conan-io/conan-center-index/pull/21524

PR #21524, opened 2023-11-30
raylib: fix build with mingw

SpaceIm:
This `#define GLFW_EXPOSE_NATIVE_WIN32` must not be removed in android patch, it doesn't make sense to remove this line since it's in `#if defined(_WIN32)` branch which has nothing to do with Android.

closes https://github.com/conan-io/conan-center-index/issues/21523

---

### DIFF
--- a/recipes/raylib/all/conandata.yml
+++ b/recipes/raylib/all/conandata.yml
@@ -8,6 +8,8 @@ sources:
 patches:
   "4.0.0":
     - patch_file: "patches/4.0.0-0001-enable-cmake-build-android.patch"
+    - patch_file: "patches/4.0.0-0002-fix-glfw-windows.patch"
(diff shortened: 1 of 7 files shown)


SpaceIm:
Well actually it doesn't help. Must this patch should be extracted in another patch because it's unrelated to Android.

SpaceIm:
It may be fixed by https://github.com/raysan5/raylib/commit/7bb8ffc29e5c31edf2875ac7aaca38316cfe45e5

SpaceIm:
Ok, it works now, I've tested on Windows with Visual Studio 2022 and MinGW GCC 13.2.0

RubenRBS (approved):
Thanks a lot for double checking :)

PR #21532, opened 2023-12-01
ios-cmake: migrate to Conan v2, improve test_package

valgur:


recipes/ios-cmake/all/conanfile.py
```diff
@@ -1,26 +1,48 @@
...
         "enable_visibility": [True, False],
         "enable_strict_try_compile": [True, False],
-        "toolchain_target": ["auto", "OS", "OS64", "OS64COMBINED",
-                       "SIMULATOR", "SIMULATOR64", "SIMULATORARM64",
-                       "TVOS", "TVOSCOMBINED",
-                       "SIMULATOR_TVOS", "WATCHOS",
-                       "WATCHOSCOMBINED", "SIMULATOR_WATCHOS",
-                       "MAC", "MAC_ARM64", "MAC_CATALYST", "MAC_CATALYST_ARM64"]
+        "toolchain_target": [
+            "auto",
```
AbrilRBS, line 28: I think we can use the fact that we're migrating this to Conan 2 to remove the auto value altogether so we don't need to check for it later? wdyt @valgur?

recipes/ios-cmake/all/conanfile.py
```diff
@@ -29,121 +51,126 @@ class IosCMakeConan(ConanFile):
...
         self._chmod_plus_x(os.path.join(self.package_folder, "bin", "cmake-wrapper"))
 
-        self.copy("LICENSE.md", dst="licenses", src=self._source_subfolder, keep_path=False)
+        copy(self, "LICENSE.md",
+             dst=os.path.join(self.package_folder, "licenses"),
+             src=self.source_folder,
+             keep_path=False)
         # satisfy KB-H014 (header_only recipes require headers)
-        tools.save(os.path.join(self.package_folder, "include", "dummy_header.h"), "\n")
+        save(self, os.path.join(self.package_folder, "include", "dummy_header.h"), "\n")
```
AbrilRBS, line 119: This is `package_type = "build-script"`, this hook workaround shoulnd't be needed right?

AbrilRBS (commented):
Some minor comments, otherwise I like how this looks :)

recipes/ios-cmake/all/conanfile.py
valgur: I don't think that would work. The logic in `config_options()` sets the default value to an appropriate one based on `self.settings.os` and falls back to `auto` if it cannot be determined. The `self._default_toolchain_target` property does not cover several Apple OS-s, notably `Macos`.

AbrilRBS: Oh so then the idea is that if it's not able to be guessed, we fail later in the validate if it's still set to auto? My only concern is that then running `conan create ... -o="&:toolchain_target=auto"` will always fail, the CLI value is overriding the decision in config_options

AbrilRBS: Oh I see - I've pushed a minor change to let users pass auto and have it be still auto-guessed, let me know if it works for you :)

AbrilRBS (approved):

recipes/ios-cmake/all/conanfile.py
valgur: The `build-scripts` package type is super-buggy or simply inconsistent with the options handling overall, for some reason. The user-supplied option values are simply ignored and not even validated to be one of the allowed values.

valgur: Oooh... the options actually work, but only if I use `-o:b toolchain_target=...`.

amerry:
@valgur are you still intending to work on this PR? This is package is a sticking point for my work's internal Conan 2 migration, so we can probably pick it up if you don't have time (although it doesn't look like it needs much work to drive it forwards? Maybe just rebasing to set off another CI run?).

valgur:
I'd be happy to help, but honestly, it's probably more likely to be merged if you open a new PR based off this one. I am no longer able to trigger the PRs without the team since to the newly-imposed limit on the number of allowed open PRs, and new PRs tend to get more immediate attention as well, it seems.

PR #21535, opened 2023-12-01
glib/2.78.x: Add patch to support Python 3.12

jwillikers:
Specify library name and version:  **glib/2.78.x**

The distutils module was removed in Python 3.12, requiring at least Meson 1.2.3 and a patch for glib. See the upstream PR [here](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3740).

Requires conan-io/conan-docker-tools#554

---

### DIFF
--- a/recipes/glib/all/conandata.yml
+++ b/recipes/glib/all/conandata.yml
@@ -30,6 +30,11 @@ sources:
     url: "https://download.gnome.org/sources/glib/2.75/glib-2.75.3.tar.xz"
     sha256: "7c517d0aff456c35a039bce8a8df7a08ce95a8285b09d1849f8865f633f7f871"
 patches:
+  "2.78.3":
+    - patch_file: "patches/0001-Switch-from-the-deprecated-distutils-module-to-the-p.patch"
+      patch_type: "portability"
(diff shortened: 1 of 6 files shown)


ericLemanissier (approved):

jwillikers:
Oh yes, actually need to install the Python `packaging` module for this to work...

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/7995808868/attempts/1)

uilianries (changes requested):
Backporting the feature is not a good idea as the upstream no longer support it so there is no guarantee about the behavior. Let's keep trying a solution with packaging first. Thank you for checking this possibility.

stevenwdv:
FYI, this is preventing `glib` from building on our macOS builder, see https://github.com/conan-io/conan-center-index/issues/23029#issuecomment-2017442481

jcar87:
closed in favour of https://github.com/conan-io/conan-center-index/pull/24591  (which backports a glib fix for this that requires no additional dependency)

PR #16098, opened 2023-02-17
protobuf: add version 4.23.1

toge:
Specify library name and version:  **protobuf/4.23.1**

- `protobuf/4.23.1` requires abseil
- `protobuf/4.23.1` requires C++14
- `protobuf/4.23.1` builds internal utf8 library (I can't find the original repository)
- while `protobuf/4.23.1` provides original `protobuf-generate.cmake`, `protobuf-config.cmake` have to be added to `cmake_build_modules`.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21592, opened 2023-12-04
protobuf: add v3.25.1

valgur:


recipes/protobuf/all/conandata.yml
```diff
@@ -1,25 +1,25 @@
 sources:
+  "3.25.1":
+    url: "https://github.com/protocolbuffers/protobuf/archive/refs/tags/v3.25.1.tar.gz"
+    sha256: "1a2affa2fbad568b9895b72e3c7cb1f72a14bf2501fba056c724dc68c249cd0f"
```
coryan, line 4: Protobuf versioning is complicated:

https://protobuf.dev/news/2022-05-06/

The major version for C++ changed with v22:

https://protobuf.dev/news/2022-08-03/

```suggestion
  "4.25.1":
    url: "https://github.com/protocolbuffers/protobuf/archive/refs/tags/v4.25.1.tar.gz"
    sha256: "e5b6dac700b6b319fd3329a84ed1594477d645591d7a484657c84cc8f28e28d8"
```

coryan (commented):
Starting with 3.25.x  µpb development was moved to the Protobuf repository:

https://protobuf.dev/news/2023-09-15/

This adds a number of new libraries that need to be represented as components. Note that this is **not** Protobuf "vendoring" µpb, the same team creates both and plan to only continue development in the Protobuf repository.

Likewise, `utf8` is now part of the Protobuf repository:

https://github.com/protocolbuffers/protobuf/blame/main/third_party/utf8_range/CMakeLists.txt

HTH

recipes/protobuf/all/conandata.yml
valgur: Thanks! This will require quite a bit more work than I anticipated. My only real goal was to get the latest version of gRPC to build.
I noticed that #16098 already includes some of the required changes you mention. I'll revert the changes here and create a separate PR based on them.

valgur: Done: #21622

uilianries (approved):
LGTM, checking current recipes in master, no affected package.

    find recipes -name conanfile.py -exec grep protobuf {} +

toge (approved):

PR #21603, opened 2023-12-05
[docs] Add changelog 05-December-2023

danimtb:
Latest changes introduced in the build service

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,14 @@
 # Changelog
 
+### 05-December-2023 - 16:23 CET
+
+- [fix] Use the correct profile to test a tool_require.
+- [feature] Add armv8 extra flag in profile arch when building in Macos M1.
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):

uilianries:
#### Summary

- When using Mac M1 as a build machine, the build profile should according to avoid mismatch when testing over Mac Intel machines.
- Add support to Mac M1 machines in C3I Jenkins, both Mac Intel and M1 are now supported when building packages.
- The Slack messenger is used to inform internally about build status. In case of failure of this extension, the CI build should not be affected.
- Epochs are no longer required, this is the last step of removing them. When a new configuration is needed (e.g. new compiler version), it will be built in advance and moved to Conan Center.
- When building packages with shared option or without shared option present, packages with shared=True should be built with lowest priority to avoid missing packages when testing.

uilianries (approved):
LGTM

PR #21593, opened 2023-12-04
grpc: add v1.60.0

valgur:
Requires
- #21592

PR #21622, opened 2023-12-05
protobuf: add v4.25.3, v5.26.0

valgur:
Continues directly from
- #16098


AndreyMlashkin:
@uilianries @SSE4 
it looks like a CI issue

coryan:
FWIW, this should be v4.25.2.  Protobuf releases exactly the same code with multiple "versions" depending on the intended language.  In this case v25.2 (the language neutral version), v4.25.2 (the C++ version), and v3.25.2 (Python?).

https://protobuf.dev/support/version-support/

Ahajha:
@valgur I changed to the proper next version, when you get a chance could you update the PR title, along with the description (I've linked two issues that should be closed by this PR). I'll ping someone to get the missing binaries.

Ahajha:
protobuf v26 was just released a few days ago, which includes another major version bump to 5.26.x.

coryan:
Just curious, why keep two 3.21.x versions? and two 3.20.x versions?  And for that matter, why keep all the 3.* versions?

Ahajha:
> Just curious, why keep two 3.21.x versions? and two 3.20.x versions? And for that matter, why keep all the 3.* versions?

There are several dozen recipes that use the 3.x versions, that said, from what I can see all are 3.21.9 or 3.21.12 except libprotobuf-mutator, which uses 3.17.1, but it will be bumped in this PR (#18796), which depends on this one.

I think we can remove all but those two versions then.

grmcdorman:
This is blocking #21593 (gRPC update), which is a security update. What is the status of this change?

Ahajha:
I will mention that there is another PR to add 4.25.0 here: https://github.com/conan-io/conan-center-index/pull/23095

Perhaps there are some elements of both that are useful, so maybe they should be reviewed in tandem.

recipes/protobuf/all/conanfile.py
```diff
@@ -58,23 +79,32 @@ def config_options(self):
...
+            if Version(self.version) >= "3.22.0" and is_msvc(self):
+                self.options["abseil"].shared = True
 
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
         if self.options.with_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
+        if Version(self.version) >= "3.22.0":
```
uilianries, line 91: Please, refer where abseil become to be used.

recipes/protobuf/all/conanfile.py
```diff
@@ -58,23 +79,32 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+            if Version(self.version) >= "3.22.0" and is_msvc(self):
+                self.options["abseil"].shared = True
```
uilianries, line 83: Please, point a reference why is it needed.

recipes/protobuf/all/conanfile.py
```diff
@@ -94,13 +124,13 @@ def generate(self):
...
-            if not runtime:
-                runtime = self.settings.get_safe("compiler.runtime")
-            tc.cache_variables["protobuf_MSVC_STATIC_RUNTIME"] = "MT" in runtime
+            tc.cache_variables["protobuf_MSVC_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
         if is_apple_os(self) and self.options.shared:
             # Workaround against SIP on macOS for consumers while invoking protoc when protobuf lib is shared
             tc.variables["CMAKE_INSTALL_RPATH"] = "@loader_path/../lib"
+        tc.variables["protobuf_ABSL_PROVIDER"] = "package"
+        if not valid_min_cppstd(self, self._min_cppstd):
+            tc.variables["CMAKE_CXX_STANDARD"] = self._min_cppstd
```
uilianries, line 133: Why need to enforce the C++ now? Is C++14 missing in Protobuf CMake?

recipes/protobuf/all/test_package/conanfile.py
uilianries, line 1: Please, explain what is being fixed with this change. Otherwise, revert to the previous form.

recipes/protobuf/all/test_package/macos_make_override.cmake
uilianries, line 1: What is not working now that requires this new file?

recipes/protobuf/all/conanfile.py
valgur: https://github.com/protocolbuffers/protobuf/commit/a9f1ea6371c108876649f27a5940a59cc8594768

recipes/protobuf/all/conanfile.py
valgur: It was added by @toge, so I had to look it up. Appears to be related to https://github.com/protocolbuffers/protobuf/blob/v25.3/cmake/abseil-cpp.cmake#L40-L54

I would leave it to a follow-up PR to fix this unless you can suggest a simple fix/workaround to this.

valgur: Failed with
```
LINK : fatal error LNK1104: cannot open file 'abseil_dll.lib' [C:\J\w\prod\BuildSingleReference@3\s\24cac7\1\build\libprotobuf.vcxproj]
```
in the CI before this change.

recipes/protobuf/all/conanfile.py
valgur: @toge Do you remember why this was needed?

recipes/protobuf/all/conanfile.py
```diff
@@ -48,6 +49,26 @@ def _is_clang_cl(self):
...
+                "gcc": "7.3",
+                "clang": "5",
+                "apple-clang": "10",
+                "Visual Studio": "15",
+                "msvc": "191",
+            }
+        return {
+            "clang": "4",
+            "Visual Studio": "14",
+            "msvc": "190",
```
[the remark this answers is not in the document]
valgur: Is this really necessary? We generally don't even include min compiler checks for C++11 in other recipes at all.

recipes/protobuf/all/test_package/macos_make_override.cmake
valgur: See the comment in test_package conanfile. The fix was copied from the gRPC recipe. Other than that, I've thoroughly forgotten all details except that it fixed a build issue on macOS.

valgur: It's probably necessary for the protoc executable to work correctly from CMakeLists.txt. But since we no longer about things like this anymore on CCI, I'll simply revert it. :slightly_smiling_face:

uilianries: Thank you!

recipes/protobuf/all/conanfile.py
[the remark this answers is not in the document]
uilianries: As far as we support Conan 1.x, then yes.

recipes/protobuf/all/conanfile.py
valgur: Conan v1 fails without it:
```
[  2%] Built target utf8_range
protobuf/4.25.3: 
In file included from /Users/jenkins/workspace/prod-v1/bsr/42822/aaffe/.conan/data/protobuf/4.25.3/_/_/build/73b28dd12161ea11293d28fd8ed2d7d3910330ef/src/third_party/utf8_range/utf8_validity.cc:18:
In file included from /Users/jenkins/workspace/prod-v1/bsr/42822/aaffe/.conan/data/protobuf/4.25.3/_/_/build/73b28dd12161ea11293d28fd8ed2d7d3910330ef/src/third_party/utf8_range/utf8_validity.h:10:
In file included from /Users/jenkins/workspace/prod-v1/bsr/42822/aaffe/.conan/data/abseil/20230802.1/_/_/package/f1a36a2aea2da1148eb4bddd758d8db183c6db83/include/absl/strings/string_view.h:39:
In file included from /Users/jenkins/workspace/prod-v1/bsr/42822/aaffe/.conan/data/abseil/20230802.1/_/_/package/f1a36a2aea2da1148eb4bddd758d8db183c6db83/include/absl/base/attributes.h:37:
In file included from /Users/jenkins/workspace/prod-v1/bsr/42822/aaffe/.conan/data/abseil/20230802.1/_/_/package/f1a36a2aea2da1148eb4bddd758d8db183c6db83/include/absl/base/config.h:86:
/Users/jenkins/workspace/prod-v1/bsr/42822/aaffe/.conan/data/abseil/20230802.1/_/_/package/f1a36a2aea2da1148eb4bddd758d8db183c6db83/include/absl/base/policy_checks.h:79:2: error: "C++ versions less than C++14 are not supported."
#error "C++ versions less than C++14 are not supported."
```

recipes/protobuf/all/conanfile.py
```diff
@@ -48,6 +49,26 @@ def _is_clang_cl(self):
...
+        }
```
uilianries, line 70:
```suggestion

```
Let's remove c++11 check then

recipes/protobuf/all/conanfile.py
uilianries: Okay, they pass explicitly the standard and is also documented to use the cxx standard

- https://github.com/protocolbuffers/protobuf/blob/v4.25.3/.github/workflows/test_cpp.yml#L120
- https://github.com/protocolbuffers/protobuf/blob/v4.25.3/cmake/README.md#c-version

recipes/protobuf/all/conanfile.py
```diff
@@ -58,23 +79,32 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+            if Version(self.version) >= "3.22.0" and is_msvc(self):
```
uilianries, line 82:
```suggestion
            if Version(self.version) >= "3.22.0" and is_msvc(self):
                # Protobuf requires absl::abseil_dll
                # https://github.com/protocolbuffers/protobuf/blob/v25.3/cmake/abseil-cpp.cmake#L40
```

recipes/protobuf/all/conanfile.py
```diff
@@ -94,13 +124,13 @@ def generate(self):
...
-            runtime = msvc_runtime_flag(self)
-            if not runtime:
-                runtime = self.settings.get_safe("compiler.runtime")
-            tc.cache_variables["protobuf_MSVC_STATIC_RUNTIME"] = "MT" in runtime
+            tc.cache_variables["protobuf_MSVC_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
         if is_apple_os(self) and self.options.shared:
             # Workaround against SIP on macOS for consumers while invoking protoc when protobuf lib is shared
             tc.variables["CMAKE_INSTALL_RPATH"] = "@loader_path/../lib"
+        tc.variables["protobuf_ABSL_PROVIDER"] = "package"
+        if not valid_min_cppstd(self, self._min_cppstd):
```
uilianries, line 132:
```suggestion
        if not valid_min_cppstd(self, self._min_cppstd):
            # Protobuf does not pass a C++ standard but requires C++14 (>=3.22)
            # https://github.com/protocolbuffers/protobuf/blob/v4.25.3/cmake/README.md#c-version
```

valgur: Good catch. Thanks!

recipes/protobuf/all/conanfile.py
uilianries: @valgur please, include the comments in the recipe too, Protobuf is hard to track.

valgur: Already did. :slightly_smiling_face:  Thanks for helping out with these.

uilianries: Thank you!

recipes/protobuf/all/conanfile.py
```diff
@@ -191,15 +214,23 @@ def package(self):
         cmake.install()
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
         os.unlink(os.path.join(self.package_folder, self._cmake_install_base_path, "protobuf-config-version.cmake"))
+        os.unlink(os.path.join(self.package_folder, self._cmake_install_base_path, f"protobuf-targets-{str(self.settings.build_type).lower()}.cmake"))
         os.unlink(os.path.join(self.package_folder, self._cmake_install_base_path, "protobuf-targets.cmake"))
-        os.unlink(os.path.join(self.package_folder, self._cmake_install_base_path, "protobuf-targets-{}.cmake".format(str(self.settings.build_type).lower())))
-        rename(self, os.path.join(self.package_folder, self._cmake_install_base_path, "protobuf-config.cmake"),
-                     os.path.join(self.package_folder, self._cmake_install_base_path, "protobuf-generate.cmake"))
+        if Version(self.version) < "3.22.0":
```
uilianries, line 219: What's the proposal of this change? I see the same `protobuf-generate.cmake` being installed with 4.25.3:

```
-- Up-to-date: /Users/uilian/.conan/data/protobuf/4.25.3/_/_/package/799e25793bb4a51e590a0065a670b4676d489bda/lib/cmake/protobuf
-- Installing: /Users/uilian/.conan/data/protobuf/4.25.3/_/_/package/799e25793bb4a51e590a0065a670b4676d489bda/lib/cmake/protobuf/protobuf-module.cmake
-- Installing: /Users/uilian/.conan/data/protobuf/4.25.3/_/_/package/799e25793bb4a51e590a0065a670b4676d489bda/lib/cmake/protobuf/protobuf-generate.cmake
-- Installing: /Users/uilian/.conan/data/protobuf/4.25.3/_/_/package/799e25793bb4a51e590a0065a670b4676d489bda/lib/cmake/protobuf/protobuf-options.cmake
-- Installing: /Users/uilian/.conan/data/protobuf/4.25.3/_/_/package/799e25793bb4a51e590a0065a670b4676d489bda/lib/cmake/protobuf/protobuf-config.cmake
-- Installing: /Users/uilian/.conan/data/protobuf/4.25.3/_/_/package/799e25793bb4a51e590a0065a670b4676d489bda/lib/cmake/protobuf/protobuf-config-version.cmake
```

NoneOfUrBusiness12:
Is there any movement on this issue? Would be really nice to get this through to allow https://github.com/conan-io/conan-center-index/pull/21593 to move forward :)

recipes/protobuf/all/conanfile.py
valgur: I don't know why @toge saw a need for the if/else based on the version, but packaging `protobuf-config.cmake` is prohibited by the linter.

uilianries:
> Is there any movement on this issue? Would be really nice to get this through to allow #21593 to move forward :)

@NoneOfUrBusiness12 Yes, it's under review. Please, take a look past conversation in this PR (few days/hours). Plus, you can customize your Github to notify about any change in this pull request.

valgur:
It's a valid error from the linter, though. We should not be including a
`protobuf-config.cmake` file in the package that could get picked up by
`find_package(protobuf CONFIG)`.

uilianries:
> It's a valid error from the linter, though. We should not be including a `protobuf-config.cmake` file in the package that could get picked up by `find_package(protobuf CONFIG)`.

@valgur My question is why changing from protobuf-generate to protobuf-protoc now.

uilianries:
@valgur Please, revert the test package and keep simplified as before. We applied a specific PR just to simplify that part of this recipe.

valgur:
See the discussion at that simplifying PR for details. It fixed the CI
limitation for build context requirements, everything else was mostly a
consequence of that. There was no binary blob to remove from the test, that
might be a security concern.

Conan basically breaks all of the exported CMake functionality of the
project by default, so it is the responsibility of Conan to also ensure
that it is still working as intended. It’s not a functional test for the
project itself but more of a ”linkage” test for CMake functions.

The hacky workaround for macOS is also highly relevant because it will crop
up in all Conan v2 recipes using protobuf. It’s better to find a suitable
fix here than in the downstream recipes. I have already been bitten by this
on CCI.

saschasc:
@valgur @RubenRBS @uilianries Any progress in this PR?

It would be great to have progress in this issue to be able to upgrade GRPC. The latest available GRPC version currently has two high severity CVEs.

jcar87:
Close as superseded by https://github.com/conan-io/conan-center-index/pull/24154

> Conan basically breaks all of the exported CMake functionality of the
> project by default, so it is the responsibility of Conan to also ensure
> that it is still working as intended. It’s not a functional test for the
> project itself but more of a ”linkage” test for CMake functions.

The upstream project does mostly 3 things in their provided CMake configuration:
1. set up the library targets
2. set up the executable target (`protobuf::protoc`)
3. set up the cmake macro `protobuf_generate` and the legacy `protobuf_generate_cpp` for compatibility

(1) is taken care of `CMakeDeps`, and (3) is expressed in the package info - the logic to generate those functions is bundled inside the package completely unchanged by the recipe. The trickier bit is (2) as CMakeDeps does not currently support defining imported executable targets.

I can see why the logic in the recipe was confusing - for earlier versions of protobuf, the macro logic was in `protobuf-config.cmake`, so the recipe was doing 3 things: prevent that file from attempting to load the `-targets.cmake` and related files, retain the macro/function logic, and inject (2) for the executable targets. This certainly did not stand the test of time. I have refactored this to split (2) to a separate file that is included as a cmake build module instead. Newer versions of protobuf are easier, because `protobuf-generate.cmake` is a separate standalone file - this is still packaged _unchanged_.

In order to test the post-conditions of (2) and (3), we've added additional logic in the test https://github.com/conan-io/conan-center-index/blob/master/recipes/protobuf/all/test_package/CMakeLists.txt#L24-L30 
So we test:
* the library targets
* the executable target is well-defined
* the function is defined
* we invoke the executable (with self.run())

From these, we are satisfied that we are protected against regressions - especially considering that we _do not_ alter the `protobuf_generate` logic in any way, shape or form. We are unlikely to consider reverting to what was before. If time proves us wrong, we will be the first to correct course and fix it.


> The hacky workaround for macOS is also highly relevant because it will crop
> up in all Conan v2 recipes using protobuf. It’s better to find a suitable
> fix here than in the downstream recipes. I have already been bitten by this
> on CCI.

I'm not sure this is actually happening. 
When _other recipes_ use protobuf, this is what happens on CI:
* for Conan 1.x, all dependencies are static, even when a consumer is shared so `DYLD_LIBRARY_PATH` is irrelevant. (there may be few exceptions, if a recipe forces the dependency to be shared and we are operating with one profile).
* for Conan 2, we pass `-o "*:shared=True"` when building a shared library - but this applies to the _host_ profile, as per the conan documentation. The `build` profile, which applies when solving `tool_requires` (for consumers that need `protoc`), will remain with the default which is `shared=False` - so again, `DYLD_LIBRARY_PATH` is unlikely to apply and cause any issues.

Of course, if one forces a cmake project in the test_package to execute an executable from the host context in the test_package, this would be an issue - don't think it's worth it if the executable is already run via `self.run()`.

So given this, I'm not sure whether this is a widespread issue.   We are more concerned about the users and whether they experience this, rather than CI. For CI we know how to address this transparently without altering the recipes. If this becomes a widespread issue for users, we can provide advise or prioritise a fix in the Client.

I would advise that if a PR claims to add new versions, that it adds new versions rather than try to add changes that were already discarded in earlier discussions. By all means, ensuring the test package is robust is a good thing, which can be handled separately to the changes in the recipe.

jcar87:
@saschasc  - new versions of protobuf are already available. We will consider PRs to add new versions of gRPC, but bear in mind that we will try to coordinate also updating consumers of gRPC (like google-cloud-cpp), to avoid version conflicts. I have a feeling finding a version of `abseil` that works with all will be tricky - but I could be wrong.

PR #18810, opened 2023-07-22
daggy: migrate to Conan v2

valgur:

PR #21631, opened 2023-12-06
libssh2: bump mbedtls

valgur:
Also to fix missing binaries in #18810


valgur:
Since `mbedtls` is not used by default, I tested the newest and oldest version locally with it enabled and it built without issues.

jwillikers (approved):

uilianries (approved):

PR #21637, opened 2023-12-06
frugally-deep: add v0.15.27, drop old versions, fix version naming

valgur:
Frugally deep uses `0.15.x-p0` for all of its version names. I assume the `-p0` suffix is meant to effectively be a patch version there, but afaik it gets interpreted as a pre-release version by Conan due to the hyphen. I converted the version names to a more standard semver format without the suffix.

### DIFF
--- a/recipes/frugally-deep/all/conandata.yml
+++ b/recipes/frugally-deep/all/conandata.yml
@@ -1,25 +1,7 @@
 sources:
-  "0.15.27":
+  "0.15.27.0":
     url: "https://github.com/Dobiasd/frugally-deep/archive/v0.15.27-p0.tar.gz"
     sha256: "b2c8aba2b59865f6a6024982aebd7fca69ce0b8b24a74ce55be2dbceb34a54c8"
-  "0.15.25-p0":
+  "0.15.25.0":
     url: "https://github.com/Dobiasd/frugally-deep/archive/v0.15.25-p0.tar.gz"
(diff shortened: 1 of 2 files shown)


uilianries (changes requested):
@valgur Please, revert it. The `p0` is for patch, like you said, but Poco C++ also follows it and use -px. Plus, the project officially uses -p0 even in the release: https://github.com/Dobiasd/frugally-deep/releases

valgur:
@uilianries Buut... this breaks the semantic versioning support in Conan? https://docs.conan.io/2.0/tutorial/versioning/version_ranges.html#semantic-versioning

You would have to use the somewhat obscure `self.requires("frugally-deep/[>0.15 <1, include_prerelease]")` for proper version ranges support. Plus, it's somewhat semantically confusing in isolation.

uilianries:
@valgur I understand your point, but it's out of Conan's scope how the upstream uses their versions. It would be better if you open a ticket to the upstream asking about not using the suffix instead. This project is just another in the middle of others that does not follow the version schematic proposed by Conan's feature.

Dobiasd:
Hi, and sorry for the trouble. I did not know the `-p0` would cause these issues. :pray:

Since I can't even remember why we introduced this suffix back in 2018, I just pushed [a new tag](https://github.com/Dobiasd/frugally-deep/releases/tag/v0.15.29) without such a suffix. :heavy_check_mark:

valgur:
Oh, wow. That was a pleasant surprise. :smile: Thank you, @Dobiasd!

PR #21649, opened 2023-12-06
libsystemd: add v255, v253.14, bump deps

valgur:


recipes/libsystemd/all/conanfile.py
```diff
@@ -162,8 +166,10 @@ def _patch_sources(self):
         apply_conandata_patches(self)
 
         meson_build = os.path.join(self.source_folder, "meson.build")
-        replace_in_file(self, meson_build, "@CONAN_SRC_REL_PATH@",
-                        "'../{}'".format(os.path.basename(self.source_folder)))
+        replace_in_file(self, meson_build, "@CONAN_SRC_REL_PATH@", f"'../{self.source_path.name}'")
+        # Fix an invalid path separator, which breaks with newer Meson versions
```
jwillikers, line 170: Has this been fixed upstream?

jwillikers (approved):
Thanks!

recipes/libsystemd/all/conanfile.py
valgur: No. I opened a PR: https://github.com/systemd/systemd/pull/30359

valgur: I have, of course, but was not familiar with the way systemd handles it. My "fix" is most certainly incorrect, apologies.

The reason I assumed it's a path separator issue was due to the decompressed source archive incorrectly splitting these files up into subdirs. It appears to be a Conan bug/limitation.

jwillikers (approved):

recipes/libsystemd/all/conanfile.py
```diff
@@ -70,14 +71,19 @@ def validate(self):
...
         self.tool_requires("gperf/3.1")
         if not self.conf.get("tools.gnu:pkg_config", check_type=str):
-            self.tool_requires("pkgconf/2.0.3")
+            self.tool_requires("pkgconf/2.1.0")
 
     def source(self):
-        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        # Extract using standard Python tools due to Conan's unzip() not handling backslashes in
+        # 'units/system-systemd\x2dcryptsetup.slice', etc. correctly.
+        download(self, **self.conan_data["sources"][self.version], filename="sources.tar.gz")
```
RubenRBS, line 83: Let's check in the client if anything can be done about fixing this behaviour :)

RubenRBS (approved):
Thanks! Looks good, will check in the client if the workaround can be removed eventually by fixing it in the client

PR #21651, opened 2023-12-06
itk: simplify patching, restore DCMTK support

valgur:
Moves most of the patching logic to the included .cmake file and _patch_sources().
Mostly as a preparation for a version bump to v5.3 or v5.4-rc1, which require further work.


AbrilRBS (commented):
I'm picking this one up now @valgur thanks a lot for the links in the custom cmake, they are quite useful!

Do you have any extra insight that might be useful when trying to add the new 5.4 version in a follow-up PR? Thanks!

valgur:
@AbrilRBS It has been a while and I don't remember a single thing about this PR or recipe anymore, sorry. I'll see if I can get this PR to pass at least, though.

recipes/itk/all/conanfile.py
```diff
@@ -242,6 +233,29 @@ def _patch_sources(self):
...
+            os.path.join(self.source_folder, "Modules", "ThirdParty", "GIFTI", "src", "gifticlib", "gifti_io.h"),
+            os.path.join(self.source_folder, "Modules", "ThirdParty", "NIFTI", "src", "nifti", "znzlib", "znzlib.h"),
+            os.path.join(self.source_folder, "Modules", "ThirdParty", "MetaIO", "src", "MetaIO", "src", "localMetaConfiguration.h"),
+            os.path.join(self.source_folder, "Modules", "ThirdParty", "MINC", "src", "libminc", "nifti", "znzlib.h"),
+            os.path.join(self.source_folder, "Modules", "ThirdParty", "NrrdIO", "src", "NrrdIO", "unteem.pl"),
+            os.path.join(self.source_folder, "Modules", "ThirdParty", "NrrdIO", "src", "NrrdIO", "privateNrrd.h"),
+        ]:
+            replace_in_file(self, path, "itk_zlib.h", "zlib.h")
+        # Truncate some third-party modules that are provided by conan_cmake_project_include.cmake
+        for pkg in ["DCMTK", "DoubleConversion", "GDCM", "Expat", "HDF5", "JPEG", "PNG", "TIFF", "ZLIB"]:
```
uilianries, line 253:
```suggestion
        # Truncate some third-party modules that are provided by conan_cmake_project_include.cmake
        for pkg in ["DCMTK", "DoubleConversion", "GDCM", "Expat", "HDF5", "JPEG", "PNG", "TIFF"]:
```

Not needed to manipulate zlib, all files have a ifdef to detect what should be included. Those that are using itk_zlib.h will point to zlib as well due cmake configuration.

- [InsightSoftwareConsortium/ITK@v5.2.1/Modules/ThirdParty/ZLIB/src/itk_zlib.h.in](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.2.1/Modules/ThirdParty/ZLIB/src/itk_zlib.h.in)
- [InsightSoftwareConsortium/ITK@9eb1359/Modules/IO/GIPL/src/itkGiplImageIO.cxx#L22](https://github.com/InsightSoftwareConsortium/ITK/blob/9eb13596cef6009c026b2ee837ea52ee05d00f11/Modules/IO/GIPL/src/itkGiplImageIO.cxx#L22)

recipes/itk/all/conanfile.py
```diff
@@ -106,38 +106,33 @@ def generate(self):
...
         tc.variables["ITK_SKIP_PATH_LENGTH_CHECKS"] = True
 
-        tc.variables["ITK_USE_SYSTEM_LIBRARIES"] = True
         tc.variables["ITK_USE_SYSTEM_DCMTK"] = True
         tc.variables["ITK_USE_SYSTEM_DOUBLECONVERSION"] = True
         tc.variables["ITK_USE_SYSTEM_EIGEN"] = True
         tc.variables["ITK_USE_SYSTEM_FFTW"] = True
         tc.variables["ITK_USE_SYSTEM_GDCM"] = True
         tc.variables["ITK_USE_SYSTEM_HDF5"] = True
-        tc.variables["ITK_USE_SYSTEM_ICU"] = True
```
uilianries, line 116: Why does ICU is no longer listed? I still see it available for 5.1.2: https://github.com/InsightSoftwareConsortium/ITK/blob/v5.1.2/Modules/ThirdParty/DCMTK/CMakeLists.txt#L11

uilianries (commented):
I'm not really sure about this "simplification". Skipping ITK CMake files requires not double check, but triple check due its complexity and what are are missing. Even vcpkg is not doing, but keeping the original files.

valgur:
> I'm not really sure about this "simplification". Skipping ITK CMake files requires not double check, but triple check due its complexity and what are are missing. Even vcpkg is not doing, but keeping the original files.

It started out reasonable-ish, but after all of the additional required fixes and the added unvendorings, the change became a lot less useful, I agree. I'll probably get rid of it when I'm working on the recipe again.

PR #21660, opened 2023-12-06
libvips: add v8.15.1

valgur:


recipes/libvips/all/conanfile.py
```diff
@@ -28,13 +30,14 @@ class LibvipsConan(ConanFile):
         "cpp": [True, False],
         "introspection": [True, False],
         "vapi": [True, False],
+        "with_archive": [True, False],
         "with_cfitsio": [True, False],
         "with_cgif": [True, False],
         "with_exif": [True, False],
         "with_fftw": [True, False],
         "with_fontconfig": [True, False],
-        "with_gsf": [True, False],
```
RubenRBS, line 36: Any info on the deleted option and not going thru the `deprecated` usual path? Why not also remove `with_imagequant` in this case if both recipes are unsupported?

RubenRBS (commented):
Some questions about `with_imagequant` and `with_gsf`, thanks!

recipes/libvips/all/conanfile.py
valgur: Because `with_gsf` never worked:
```python
        if self.options.with_gsf:
            raise ConanInvalidConfiguration("libgsf recipe not available in conancenter yet")
```
`libgsf` option has been removed in the latest version and keeping the broken option around for the older version did not seem reasonable: https://github.com/libvips/libvips/commit/00e4047ff9a223132adfdf874f723139e399c6f2#diff-76ea9078099f2ee8cc0e760a874dd153f225ba16ad10375c24ab6cd53518f3c4

I should have mentioned this in the PR description, really. Apologies.

RubenRBS (approved):

ericLemanissier (approved):

PR #21664, opened 2023-12-06
sundials: add v7.1.1, v6.7.0, v5.8.0, add optional deps, fix CMake names

valgur:
Adds the latest major versions.

Also:
- Adds all optional dependencies that are available on CCI or in a pending PR.
- Updates the CMake config file name to `SUNDIALS` and CMake component names to `SUNDIALS::<component>` to match the project.
- Dropped the `sundials_` prefix from the Conan components to match the CMake component names.


RubenRBS (commented):
Thanks @valgu! Can you take a look at the hook warnings before merging? Thanks!

jwillikers (approved):

recipes/sundials/all/test_package/test_package_v6.c
```diff
@@ -0,0 +1,18 @@
+#include <nvector/nvector_serial.h>
```
danimtb, line 1: we need to document why do we need different code for the test pacakge depending on the version. At least we should provide a hint in the description of this PR. This might be obvious, but I am seeing this from the eyes of a complete outsider to the library itself. Thanks!

danimtb (changes requested):

recipes/sundials/all/test_package/test_package_v6.c
valgur: So I should add something like this as a comment, I guess?
> SUNDIALS v6.0.0 introduced a new SUNContext object on which all other SUNDIALS objects depend.
https://github.com/LLNL/sundials/blob/main/CHANGELOG.md#changes-to-sundials-in-release-600

valgur: Done.

danimtb (approved):

jwillikers (approved):

AbrilRBS (dismissed):

jwillikers (approved):

PR #21665, opened 2023-12-06
geos: add v3.12.1, simplify patching

valgur:


recipes/geos/all/conanfile.py
```diff
@@ -96,8 +93,14 @@ def generate(self):
         tc.variables["BUILD_GEOSOP"] = self.options.utils
         tc.generate()
 
+    def _patch_sources(self):
+        # Avoid setting CMAKE_BUILD_TYPE default when multi-config generators are used.
+        # https://github.com/libgeos/geos/pull/945
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
+                        "set(CMAKE_BUILD_TYPE ${DEFAULT_BUILD_TYPE})", "", strict=False)
```
RubenRBS, line 100: By having this not be strict, we could get into a situation where the patch is no longer needed but we don't realize. Any reason not to have it?

RubenRBS (commented):
Thanks! Question about the new patch method, otherwise looks good!

recipes/geos/all/conanfile.py
valgur: Fixed.

perseoGI:
Hi @valgur, there is a new release https://github.com/libgeos/geos/releases/tag/3.12.2
Could you update this PR te reflect new changes.
The patch simplification LGTM btw

perseoGI (approved):
LGTM

AbrilRBS (approved):
Thanks!

PR #21697, opened 2023-12-09
mp-units 2.1.0 added, apple-clang compilation enabled, and invalid range-v3 dependency removed

mpusz:
Specify library name and version:  **mp-units/2.1.0**

Additionally:
- apple-clang-15 enabled as it should be able to compile both 2.0.0 and the new version
- range-v3 dependency removed as it was not needed by any of the versions (it was required by clang < 14 which is not supported by the V2 framework)

---


mpusz:
Hi, it seems that gcc-11.1 does not like our code. However, everything should work fine starting from gcc-11.2. Is there a way to use the newer gcc-11 compiler in the Conan CI? Does it mean that we should disallow the gcc-11 altogether in the recipe?

uilianries:
@mpusz Do you know if something related to GCC 11 any patch version? It's not possible to run another GCC version in CCI, it would require tons of effort, because we would need to generate a new docker image, include it to Jenkins, and build ALL recipes in ConanCenterIndex now. 

If is something related to GCC 11, maybe you could just rise a `ConanInvalidConfiguration`. We are running GCC 11.1.0 in CCI:

```
 % docker run --rm --platform=linux/amd64 conanio/gcc11-ubuntu16.04 gcc --version 
gcc (GCC) 11.1.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

mpusz:
gcc-11.1 had a bug that was fixed in gcc-11.2. Is there a way to get a compiler minor version in Conan as well? I think that `self.settings.compiler.version` returns a major number only, right?

recipes/mp-units/2.0.0/conanfile.py
```diff
@@ -81,11 +73,13 @@ def loose_lt_semver(v1, v2):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires at least {compiler} {min_version} ({compiler.version} in use)"
             )
-        
-        # Note that apple-clang and msvc are disabled for now, their C++ 20 implementations are not up to speed
+
+        # Note that apple-clang is disabled for now, its C++ 20 implementation are not up to speed
```
jwillikers, line 77: This comment looks inconsistent with this check. Perhaps this comment can be removed since the message in the error has a similar description.

recipes/mp-units/config.yml
```diff
@@ -3,3 +3,5 @@ versions:
     folder: 0.8.0
   "2.0.0":
     folder: 2.0.0
+  "2.1.0":
+    folder: 2.0.0
```
jwillikers, line 7: If you could order the newer versions at the top that would be great.

jwillikers (approved):

recipes/mp-units/2.0.0/conandata.yml
```diff
@@ -2,3 +2,6 @@ sources:
   "2.0.0":
     url: https://github.com/mpusz/units/archive/v2.0.0.tar.gz
     sha256: c04cda9fdaeca3ae42886552658e8bde3b608a24a4a921a621a5db41ec416e0e
+  "2.1.0":
+    url: https://github.com/mpusz/units/archive/v2.1.0.tar.gz
+    sha256: a42057492f5a8e80442cf13602b97bb48b9c3a408bc91bdb60d86aadc84c95cc
```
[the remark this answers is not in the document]
mpusz: Sure, I was not aware of this convention.

recipes/mp-units/config.yml
mpusz: Done

jwillikers (approved):
Thanks!

mpusz:
@jwillikers and @uilianries, how should I fix `loose_lt_semver` (that I copy-pasted from other 2.0 recipes) to properly fail if gcc-11.1 is used?

jwillikers (approved):

mpusz:
Guys, this has been open for nearly a month now. I did not get the feedback I asked for, so I limited the compiler's version range, and it is still not accepted 😢 

I already got a few questions from customers about why there is no 2.1.0 package on ConanCenter even though it was officially released a long time ago. I am also waiting to publish the official release announcement on our blog until the ConanCenter link starts to work.

Can someone please accept this PR?

uilianries (approved):
LGTM. Sorry the delay.

jcar87 (approved):

PR #18623, opened 2023-07-18
imagemagick: migrate to Conan v2

valgur:
- Probably resolves #15168.

PR #21699, opened 2023-12-09
ImageMagick: switch to CMake, migrate to Conan v2

valgur:
Continues from #10003 by @Cyriuz.

It merges the build scripts from the latest commit at https://github.com/Cyriuz/ImageMagick into the official source archive.

I also exposed all available options and added support for all dependencies available on CCI.


recipes/imagemagick/cmake/conanfile.py
```diff
@@ -0,0 +1,445 @@
...
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
+                        "    find_package(${MAGICK_FIND_NAME} QUIET)\n",
+                        "    find_package(${MAGICK_FIND_NAME} REQUIRED)\n"
+                        "    string(TOUPPER ${MAGICK_FIND_NAME} name_upper)\n"
+                        "    set(${name_upper}_FOUND ${${MAGICK_FIND_NAME}_FOUND})\n")
+        # Fix incorrect .so install location
+        for subdir in ["MagickCore", "MagickWand", "Magick++"]:
+            replace_in_file(self, os.path.join(self.source_folder, subdir, "CMakeLists.txt"),
+                            "LIBRARY DESTINATION bin\n",
+                            "LIBRARY DESTINATION lib\n")
```
valgur, line 308: @Cyriuz Could you fix this minor issue upstream in you repository, please? Only `RUNTIME` libraries are normally installed under `bin`.

Amazing work on the CMake build system, by the way. Thank you so much!

valgur:
@Cyriuz Could you please sign the CLA for this PR?

recipes/imagemagick/cmake/test_package/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+            check("with_rsvg", "rsvg")
+            check("with_tiff", "tiff")
+            check("with_webp", "webp")
+            check("with_wmf", "wmf")
+            check("with_x11", "x")
+            check("with_xml2", "xml")
+            check("with_zlib", "zlib")
+            # zstd delegate info is missing:
+            # https://github.com/ImageMagick/ImageMagick/blob/7.1.1-22/MagickCore/version.c#L218
+            # check("with_zstd", "zstd")
```
valgur, line 86: https://github.com/ImageMagick/ImageMagick/pull/6934

recipes/imagemagick/cmake/conanfile.py
Cyriuz: Will take a look! Thank you for continuing on this!

Cyriuz: I did it! Just a heads up, the latest version I have updated the cmake for is 7.1.1-4, I can see that there's a few changes made in the makefiles between that and 7.1.1-22. It's possible there's tweaks that has to be made for it to be building everything correctly. I don't have time to fix it this week probably so either if you want to take a look and make PRs I'll look at them asap or if we want to be safer lock it to 7.1.1-4 to begin with?

valgur: Opened a PR: https://github.com/Cyriuz/ImageMagick/pull/1

Julianiolo:
@valgur What is the state of this? I tried using it with the newest Imagemagick (7.1.1-32) and it worked? I had to make some changes though: I changed the freetype version to 2.15.0 and disabled fftw and openmp since they did not work. In the test_package the identify test does not work, as it cannot find it, and the check for `gslib` and `ps` did not work (both not in the options?). When disabling these everything seemed to work?

(Built on Windows mvsc)

valgur:
@Julianiolo Thanks for testing the recipe. That's good to know!

However, the custom build system included from the forked repo is a too significant change to be acceptable on CCI, I'm afraid. It should really be merged upstream instead, ultimately.

For CCI, a more realistic approach would be to substitute the custom "configure.exe" executable on Windows with a simpler workaround, I think.

Julianiolo:
Yeah I guess the build situation with imagemagick is a little .... unfortunate at the moment.

About an configure.exe workaround: I guess that could work, but that would probably be the same deal as with cmake now. I'm not sure on how this is handled best. Maybe using the cmake solution isn't that bad of an idea?, it does not seem they will adopt more portable build systems any time soon ...

Julianiolo:
I guess it would maybe be possible to use msys2/mingw gcc, but I don't know how conan would/can handle that, when building with msvc...

Cyriuz:
> However, the custom build system included from the forked repo is a too significant change to be acceptable on CCI, I'm afraid. It should really be merged upstream instead, ultimately.

The big problem is that they won't accept any new build system unless it is used by the community and "proven" to work, so at least when I tried to get this into CCI it was with the intent to do just that and then possibly get it merged upstream later. But yes it would be quite some code to maintain on the recipe level unfortunately.

neobrain:
Thanks for working on this! I've been using this recipe locally and it's working fairly well.

I've encountered two small issues though:
* On macOS, linking fails unless `with_opencl=False`: `"_clBuildProgram", referenced from: _InitializeOpenCL in libMagickCore-7.Q16HDRI.a[55](opencl.c.o)`
* Conan generates CMake targets named `imagemagick::Magick++/MagickCore` from the recipe, but CMake's FindImageMagick module `ImageMagick::Magick++/MagickCore`. I think this can be fixed by setting `self.cpp_info.components["MagickCore"].set_property("cmake_target_name", "ImageMagick::MagickCore")`

Since I copied the recipe locally I can just work around this issues easily, but I figured I'd let you know :)

valgur:
I patched out the MFC dependency for configure.exe on Windows at #18623, so that version should support all platforms now.

valgur:
hello world

Issue #21707, opened 2023-12-11
Package: Improvement of recipe for OpenSSL - Addition of custom default OPENSSL_TLS_SECURITY_LEVEL


jmalopoy:
Latest releases of OpenSSL have modified what ciphers are assigned to the security level 0 ("X509 certificates signed using SHA1 are no longer allowed at security level 1 and above" in v3.0 [https://www.openssl.org/news/cl30.txt](https://www.openssl.org/news/cl30.txt) ) and the default security level defined at build time ("The default SSL/TLS security level has been changed from 1 to 2." in v3.2 [https://www.openssl.org/news/cl32.txt](https://www.openssl.org/news/cl32.txt) ).
Unfortunately not everybody is able to upgrade / adapt to such security enhancements so it would be great if the current recipe provides a setting to set / relax the default security level of the library.

Croydon:
#21708 is merged, can be closed

czoido:
Closing this as looks fixed, please feel free to reopen if you have more issues.

PR #21708, opened 2023-12-11
Improvement of recipe for OpenSSL - Addition of custom default OPENSSL_TLS_SECURITY_LEVEL (issue #21707)

jmalopoy:
openssl/3.2

This is a proposal for improvement request #21707.
It allows to specify a custom value to OPENSSL_TLS_SECURITY_LEVEL at build time.

---

- [x ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).


jmalopoy:
> ### Conan v1 pipeline ❌
> 
> Changes not allowed in build 3:
> 
> ```
> [ceres-solver, openssl, ceres-solver]
> ```
> 
> Only one library can be changed in the same PR.
> ### Conan v2 pipeline ❌
> 
> 
> The v2 pipeline failed. Please, review the errors and note this is required for pull requests to be merged. In case this recipe is still not ported to Conan 2.x, please, ping `@conan-io/barbarians` on the PR and we will help you.
> See details:
> 
> Changes not allowed in build 3:
> 
> ```
> [ceres-solver, openssl, ceres-solver]
> ```
> 
> Only one library can be changed in the same PR.

Hi, 

I got the error about modification of more than one library, but I cannot see any details of the real issue.
As far as I know, this PR is only modifying openssl, nothing about 'ceres-solver' (I have no clue about what it is and where it is being modified). Could you provide any help? (pinging @prince-chrismc but please, let me know who can I address this issue)

recipes/openssl/3.x.x/conanfile.py
uilianries, line 1: I would recommend adding a check in `validate()` for `securitylevel` if it's integer. Otherwise, raise `ConanException(f"{self.ref} option securitylevel must be an integer value.")`

uilianries:
> Hi,

I got the error about modification of more than one library, but I cannot see any details of the real issue.
As far as I know, this PR is only modifying openssl, nothing about 'ceres-solver' (I have no clue about what it is and where it is being modified). Could you provide any help? (pinging @prince-chrismc but please, let me know who can I address this issue)

Hello @jmalopoy sorry about that error, it's an internal CI error due concurrent CI jobs. I'll restart the CI build. Also, please, use `@conan-io/barbarians` to ping Conan team when something weird occurs or you need extra support.

recipes/openssl/3.x.x/conanfile.py
SpaceIm: Possible values are 0, 1, 2, 3, 4, 5. Default value is 1 in OpenSSL. Why not just provide this range instead of ANY?

https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_security_level.html

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -85,11 +85,13 @@ class OpenSSLConan(ConanFile):
         "no_whirlpool": [True, False],
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
+        "securitylevel": [None, "ANY"],
```
SpaceIm, line 88: Possible values are 0, 1, 2, 3, 4, 5. Default value is 1 in OpenSSL. Why not just provide this range instead of ANY?

https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_security_level.html

jmalopoy:
> > Hi,
> 
> I got the error about modification of more than one library, but I cannot see any details of the real issue. As far as I know, this PR is only modifying openssl, nothing about 'ceres-solver' (I have no clue about what it is and where it is being modified). Could you provide any help? (pinging @prince-chrismc but please, let me know who can I address this issue)
> 
> Hello @jmalopoy sorry about that error, it's an internal CI error due concurrent CI jobs. I'll restart the CI build. Also, please, use `@conan-io/barbarians` to ping Conan team when something weird occurs or you need extra support.

Thanks for the clarification

recipes/openssl/3.x.x/conanfile.py
jmalopoy: Thanks for the suggestion.

I will improve the validation following the proposal of @SpaceIm (using a range).

recipes/openssl/3.x.x/conanfile.py
jmalopoy: Thanks for this suggestion.
I have pushed a small modification using a range, which also improves the input validation.

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -85,11 +85,13 @@ class OpenSSLConan(ConanFile):
         "no_whirlpool": [True, False],
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
+        "securitylevel":  [None] + list(range(0, 6)),
```
SpaceIm, line 88: why None?

jmalopoy: I thought it would be a good idea to have it to keep the default behavior of OpenSSL.
As far as I know, in releases < v.3.2.0 the default value is 1, but starting from v3.2.0 it has changed to 2 ( [https://www.openssl.org/news/cl32.txt](https://www.openssl.org/news/cl32.txt) )

> * The default SSL/TLS security level has been changed from 1 to 2. RSA,
   DSA and DH keys of 1024 bits and above and less than 2048 bits and ECC keys
   of 160 bits and above and less than 224 bits were previously accepted by
   default but are now no longer allowed. By default TLS compression was
   already disabled in previous OpenSSL versions. At security level 2 it cannot
   be enabled.

SpaceIm: Fair enough

uilianries, line 88:
```suggestion
        "securitylevel": [0, 1, 2, 3, 4, 5],
```

Use config options to change it:

```
def config_options(self):
    self.options.tls_security_level = 1 Version(self.version) < "3.2" else 2
```

Plus, I would ask to rename the option to `tls_security_level` to reflect its origin.

jmalopoy: I like the suggested change about the config_options and the specification of valid values. It is clearer. Thanks.
Regarding the rename, I think securitylevel is fine, taking into account the OpenSSL documentation uses more the short term than the whole definition name.

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -380,6 +385,9 @@ def _configure_args(self):
         args.append("no-fips" if self.options.get_safe("no_fips", True) else "enable-fips")
         args.append("no-md2" if self.options.get_safe("no_md2", True) else "enable-md2")
 
+        if self.options.securitylevel != None:
+            args.append("-DOPENSSL_TLS_SECURITY_LEVEL=%s" % str(self.options.securitylevel))
```
uilianries, line 389:
```suggestion
        args.append("-DOPENSSL_TLS_SECURITY_LEVEL=%s" % str(self.options.securitylevel))
```

It can not be None

recipes/openssl/3.x.x/conanfile.py
uilianries: >  I think securitylevel is fine, taking into account the OpenSSL documentation uses more the short term than the whole definition name.

Sorry, but I'll insist for `tls_security_level`. The cmake definition from the upstream follows the same name. Other options in this recipe (except openssldir) use underscore to separate option names.

jmalopoy: Ok. I updated it

recipes/openssl/3.x.x/conanfile.py
jmalopoy: Sure. I updated it

uilianries (approved):
LGTM.

jwillikers (approved):

PR #21714, opened 2023-12-11
[fix][ceres-solver] Added IOS_DEPLOYMENT_TARGET variable

franramirez688:
Specify library name and version:  **ceres-solver/***
Closes: https://github.com/conan-io/conan-center-index/issues/21715

`ceres-solver` uses the `IOS_DEPLOYMENT_TARGET` cmake variable to verify that the iOS version is higher than 7.
Indeed, the CMakeLists.txt does not include the iOS.cmake toolchain so it's up to the consumer to include that one as the official documentation reads http://ceres-solver.org/installation.html#ios

### DIFF
--- a/recipes/ceres-solver/all/conanfile.py
+++ b/recipes/ceres-solver/all/conanfile.py
@@ -1,3 +1,6 @@
+import os
+import textwrap
+
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import is_apple_os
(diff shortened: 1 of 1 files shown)


SpaceIm (approved):

11janson:
Thanks for fix this bug.
How should I install Conan after this. brew install Conan or pip install Conan?

RubenRBS (approved):
Nice! :)

uilianries (approved):

PR #16285, opened 2023-02-25
[date] Patch recipe to use tz recipe by default to obtain time zone database

samuel-emrys:
Specify library name and version:  **date/3.0.1**

Date currently initiates its own download of the IANA Timezone Database on first execution. This is effectively a dependency that's being managed by date, so this pull request adds the ability to pull this database from the tz package.

This change allows the usage of date in offline environments.

The following modifications were made to the recipe:
* Remove `use_system_tz_db` option in favour of `with_tzdb` option which accepts multiple string types for the location from which to source the timezone database
* Patch v2.4.1, v3.0.0, v3.0.1 source code to include logic to allow the conan packaged timezone database to be used
* Throw error if `use_tz_db_in_dot` is used when `with_tzdb` is anything other than `download`. This resolves a conflict in logic to set the installation directory of the timezone database.

Blocked by #16284 
Closes #16204

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21719, opened 2023-12-11
[date] Add tz recipe as dependency and add windows compatibility with binary zic-compiled tzdb

### LINKED ISSUE #16204: [package] date/3.0.1: Should package the tzdata database with the package
### Description

Currently, the default date library behaviour is to automatically attempt to download the tz database using curl when initialized. This would be trivial to get within the `source()` method and ship along with the library, and mitigate the need to do this if it doesn't exist on the system. This would have the benefit of making the date package entirely portable in offline environments, with the slight negative side effect that this database will be frozen in time at the time the package is created.

I think this is worth the tradeoff, as the user of the date library can always elect to download a newer version manually and direct date to look at the new location when recompiling, or to use the system timezone database.

NA

### Conan profile

NA

### Steps to reproduce

NA

### Logs

NA

samuel-emrys:
* Initial implementation adding tz recipe as a dependency
* Apply patch from PR HowardHinnant/date#611 to add ability to specify tz database by environment variable and also enable the binary tz database to be parsed on windows
* Deprecate `use_system_tz_db` in favour of `with_tzdb` option to handle all mutually exclusive options
* Add `with_db_format` option to provide flexibility in how the tz package is consumed

Specify library name and version:  **date/3.0.1**

This implementation supersedes #16285. It makes the following improvements:
* Allows `date` to locate the tz database using the `TZDATA` environment variable. This is an improvement over #16285, which hardcoded the path to the tz database within the binary, making it non-portable. Usage of environment variables here allows flexible runtime discovery of the tz database.
* Adds windows compatibility to read the zic-compiled binary tz database, which is functionality that previously only existed for unix platforms.
* No changes were required to the build system other than what has been modified in HowardHinnant/date#611. This has pro's and cons:
** Is the most minimal change required to introduce the features we need to link the tz recipe with conan
** Sacrifices some semantic meaning/the build options that need to be enabled to provide desired features are not intuitive.

Closes #16204

---

### DIFF
--- a/recipes/date/all/conandata.yml
+++ b/recipes/date/all/conandata.yml
@@ -13,6 +13,10 @@ patches:
     - patch_file: "patches/cmake-3.0.1.patch"
       patch_description: "Disable string view to workaround clang 5 not having it"
       patch_type: "portability"
+    - patch_file: "patches/load-tzdb-from-envvar-and-windows-binary-db-support-3.0.1.patch"
+      patch_description: "Allow TZDB to be loaded by envvar and add ability for windows to read zic-compiled binary db"
+      patch_type: "portability"
+      patch_source: "https://github.com/HowardHinnant/date/pull/611"
   "3.0.0":
     - patch_file: "patches/cmake-3.0.0.patch"
       patch_description: "Disable string view to workaround clang 5 not having it"
--- a/recipes/date/all/conandata.yml
+++ b/recipes/date/all/conandata.yml
@@ -21,6 +21,10 @@ patches:
     - patch_file: "patches/cmake-3.0.0.patch"
       patch_description: "Disable string view to workaround clang 5 not having it"
       patch_type: "portability"
+    - patch_file: "patches/load-tzdb-from-envvar-and-windows-binary-db-support-3.0.0.patch"
+      patch_description: "Allow TZDB to be loaded by envvar and add ability for windows to read zic-compiled binary db"
+      patch_type: "portability"
+      patch_source: "https://github.com/HowardHinnant/date/pull/611"
   "2.4.1":
     - patch_file: "patches/0001-fix-uwp.patch"
       patch_description: "Fix Universal Windows Platform (UWP) unhandled exception support. See https://github.com/microsoft/vcpkg/pull/8151#issuecomment-531175393."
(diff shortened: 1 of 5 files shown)


samuel-emrys:
For tracking, I've raised https://github.com/HowardHinnant/date/issues/788 upstream for discussion of changes to the `date` library that might make integration with a package manager easier. Specifically, the inclusion of environment variables to specify the location of the tz database at runtime. The authors position was that they did not have time to provide an implementation or review any implementation that might be contributed by someone else.

The author was also of the opinion that this may inhibit his attempts to encourage migration to C++20 chrono. Unfortunately, migration to C++20 chrono is not a perfect solution as:
* Some projects may be constrained to C++ standards older than C++20
* There is still not good coverage of C++20 chrono implementations across compilers and operating systems

In these cases, it's still necessary for projects to be able to provide `date` features and use an appropriate tz database.

On further research into what my implementation might look like, I discovered https://github.com/HowardHinnant/date/pull/611 which provides close to the fewest possible changes required to introduce the core features that I conan-center-index needs to couple the `date` package with the `tz` package whilst also improving windows compatibility.

For similar reasons above, my belief is that it's unlikely that these changes will be merged upstream (given that it was raised more than 3 years ago). My suggestion is that we incorporate these changes into the `date` package on conan center to improve its compatibility with the timezone database, and flexibility as to where this timezone database is obtained.

samuel-emrys:
@uilianries I appreciate that there have been a number of CI related issues over the past week or two (and I suspect you're tracking the failure here as well) - I only have a macOS environment on which to test these changes for the moment (and docker for Linux), so I'm relying on the CI to provide feedback on Windows compatibility. Should I re-trigger the pipeline or is there something that you need to do on the backend?

uilianries:
Hello @samuel-emrys!! Sorry about the CI situation, indeed we are babysitting the situation.

Usually, we restart manually in CI (Jenkins), but also when marking a PR with `Unexpected error` label, the CI should re-start the PR automatically (it re-runs a failed PR every 10 min).

On the user side, it's possible to enforce it, but we ask to avoid doing it frequently to avoid overloading the CI. To do it, you just need to "rebase" you branch by clicking on the update branch button:

Screenshot 2023-12-13 at 10-32-44 date Add tz recipe as dependency and add windows compatibility with binary zic-compiled tzdb by samuel-emrys · Pull Request #21719 · conan-io_conan-center-index

samuel-emrys:
**This is not ready to be merged**. The tests are failing to read from the binary database, but not erroring appropriately.

samuel-emrys:
I'm closing this in favour of https://github.com/conan-io/conan-center-index/pull/22294. The Windows binary database parser implemented in HowardHinnant/date#611 turns out not to be mature enough to be functional at this stage. I've pared this patch back, and submitted my own PR with just the modifications necessary to use an environment variable to inject the path to a tz database in HowardHinnant/date#807, which I've added a conan implementation for in #22294.

The ability to read binary databases on Windows is an important consideration, but can be implemented in a future PR.

PR #21744, opened 2023-12-13
cyrus-sasl: add Windows support

valgur:
Shared library output only as the MSBuild project hard-codes the build type and DLL defines.


RubenRBS:
Can you address the hook errors before merging @valgur? Thanks!

RubenRBS (approved):

jwillikers (approved):

recipes/cyrus-sasl/all/conanfile.py
```diff
@@ -77,36 +84,31 @@ def layout(self):
...
             self.requires("libmysqlclient/8.1.0")
-        if self.options.with_sqlite3:
+        if self.options.get_safe("with_sqlite3"):
             self.requires("sqlite3/3.44.2")
 
     def validate(self):
-        if self.settings.os == "Windows":
-            raise ConanInvalidConfiguration(
-                "Cyrus SASL package is not compatible with Windows yet."
-            )
```
SpaceIm, line 93: It was raising for Windows and this PR adds support for msvc. What about other compilers on Windows? It still doesn't work with MinGW AFAIK, it would be nice to not lost this information.

valgur: I appreciate the concern, but I don't currently have a setup to test MinGW compatibility and I would not want to disable it just in case either. Feel free to open a PR.

PR #21782, opened 2023-12-15
sdl_mixer: add v2.8.0 as a new subdir

valgur:
The existing recipe is quite outdated. The project has switched over to using CMake.

This also effectively migrates the package to Conan v2.


RubenRBS:
Rebuilding the missing dependency now :)

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+        tc.variables["SDL2MIXER_MIDI"] = self.options.get_safe("nativemidi", False) or self.options.get_safe("tinymidi", False) or self.options.fluidsynth
+        tc.variables["SDL2MIXER_MIDI_FLUIDSYNTH"] = self.options.fluidsynth
+        tc.variables["SDL2MIXER_MIDI_TIMIDITY"] = self.options.get_safe("tinymidi", False)
+        tc.variables["SDL2MIXER_MIDI_NATIVE"] = self.options.get_safe("nativemidi", False)
+        # OPUS
+        tc.variables["SDL2MIXER_OPUS"] = self.options.opus
+        # VORBIS
+        if self.options.vorbis == "stb":
+            tc.variables["SDL2MIXER_VORBIS"] = "STB"
+        elif self.options.vorbis == "voribisfile":
```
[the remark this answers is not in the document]
valgur: Indeed. Thank you!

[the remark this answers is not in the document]
petitlapin: About the remaining issues, do you have an idea of what goes wrong? I don't know much conan but I can try to take a look (with no guarantee) when I have some spare time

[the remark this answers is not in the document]
valgur: @petitlapin I looked into it and I think it's caused by this issue: #22170

recipes/sdl_mixer/cmake/test_package/test_package.c
```diff
@@ -0,0 +1,42 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <SDL.h>
+#include <SDL_mixer.h>
```
petitlapin, line 4: I think it would be better to include SDL2 in the path to ensure we look for SDL2 (and not getting SDL by mistake)
```suggestion
#include <SDL2/SDL.h>
#include <SDL2/SDL_mixer.h>
```

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+    def package(self):
+        copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "SDL2_mixer")
```
petitlapin, line 192: I don't know if it is a good track to follow to try to fix the error but in sdl_ttf: https://github.com/conan-io/conan-center-index/blob/master/recipes/sdl_ttf/all/conanfile.py#L115
```
suffix = "-static" if Version(self.version) >= "2.20.0" and not self.options.shared else ""
self.cpp_info.components["_sdl2_ttf"].set_property("cmake_target_name",f"SDL2_ttf::SDL2_ttf{suffix}")
```
and in the test [CMakeLists.txt](https://github.com/conan-io/conan-center-index/blob/master/recipes/sdl_ttf/all/test_package/CMakeLists.txt#L7C1-L11C8) they use:
```
if(TARGET SDL2_ttf::SDL2_ttf)
    target_link_libraries(${PROJECT_NAME} PRIVATE SDL2_ttf::SDL2_ttf)
else()
    target_link_libraries(${PROJECT_NAME} PRIVATE SDL2_ttf::SDL2_ttf-static)
endif()
```

valgur: Yes, it should. That's what the targets exported by the `sdl_mixer` project look like as well: https://github.com/libsdl-org/SDL_mixer/blob/release-2.6.3/CMakeLists.txt#L164-L172

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+
+        deps = CMakeDeps(self)
+        deps.set_property("flac", "cmake_file_name", "FLAC")
+        deps.set_property("flac", "cmake_target_name", "FLAC")
+        deps.set_property("fluidsynth", "cmake_file_name", "FluidSynth")
+        deps.set_property("fluidsynth", "cmake_target_name", "FluidSynth::FluidSynth")
+        deps.set_property("libxmp", "cmake_file_name", "libxmp")
+        deps.set_property("libxmp", "cmake_target_name", "libxmp::libxmp")
+        deps.set_property("modplug", "cmake_file_name", "modplug")
+        deps.set_property("modplug", "cmake_target_name", "modplug::modplug")
```
SpaceIm, line 166:
```suggestion
        deps.set_property("libmodplug", "cmake_file_name", "modplug")
        deps.set_property("libmodplug", "cmake_target_name", "modplug::modplug")
```

Here is your issue, https://github.com/conan-io/conan-center-index/pull/22170 is irrelevant.

valgur: Thanks!

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,212 @@
...
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
```
petitlapin, line 188: On Windows, it seems to be `self, os.path.join(self.package_folder, "cmake")`. The suggestion is not tested:
```suggestion
        if self.settings.os == "Windows":
            rmdir(self, os.path.join(self.package_folder, "cmake"))
        else:
            rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
```

From the logs:
Linux:
> -- Install configuration: "Release"
> -- Installing: /home/conan/w/prod-v2/bsr/70187/fbaed/p/b/sdl_md2106973f57d3/p/lib/libSDL2_mixer.a
> -- Installing: /home/conan/w/prod-v2/bsr/70187/fbaed/p/b/sdl_md2106973f57d3/p/include/SDL2/SDL_mixer.h
> -- Installing: /home/conan/w/prod-v2/bsr/70187/fbaed/p/b/sdl_md2106973f57d3/p/lib/cmake/SDL2_mixer/SDL2_mixerConfig.cmake


Windows:
> -- Installing: C:/J2/w/prod-v1/bsr@2/84698/ceffe/.conan/data/sdl_mixer/2.6.3/_/_/package/51188b1528b7de4f0f4ac2754c459259f18a23cd/lib/SDL2_mixer-static.lib
> -- Installing: C:/J2/w/prod-v1/bsr@2/84698/ceffe/.conan/data/sdl_mixer/2.6.3/_/_/package/51188b1528b7de4f0f4ac2754c459259f18a23cd/include/SDL2/SDL_mixer.h
> -- Installing: C:/J2/w/prod-v1/bsr@2/84698/ceffe/.conan/data/sdl_mixer/2.6.3/_/_/package/51188b1528b7de4f0f4ac2754c459259f18a23cd/cmake/SDL2_mixerConfig.cmake

After checking, the code in SDL_mixer that set this variable for Windows is: https://github.com/libsdl-org/SDL_mixer/blob/SDL2/CMakeLists.txt#L910. I'm not sure of the reason but if there is a if in the code, there should be a reason so the workaround in conan may make sense...

petitlapin:
For the Linux error, I found https://github.com/conan-io/conan-center-index/pull/22081 which removed the dependency a few days ago for xorg. I don't know if pulseaudio should be rebuilt with this version to remove the dependency?

Ahajha:
> For the Linux error, I found #22081 which removed the dependency a few days ago for xorg. I don't know if pulseaudio should be rebuilt with this version to remove the dependency?

I'm running into the same issue in CPython. It's possible a bunch of dependencies will need to be rebuilt.

Croydon:
Pulseaudio rebuild is happing in  #22241, awaiting approval

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,212 @@
+import os
```
Croydon, line 1: How much of the recipe itself did you change? Is it worth the file split?

valgur: The project changed its CMake config and dependencies substantially since the current v2.0.4.
Cramming it all together seemed like a worse option with no real upside.
Here's a PR to migrate the existing recipe: #21782

petitlapin:
According to the [changelog](https://github.com/libsdl-org/SDL_mixer/releases/tag/release-2.8.0):
```
Use libxmp instead of modplug as the default backend for MOD music
To use libmodplug instead, configure using --enable-music-mod-modplug --disable-music-mod-xmp
```

I guess the recipe should either use libxmp or force the use of modplug.

petitlapin:
hi, is there a reason to not merge it?

zaslavskyv (approved):

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,255 @@
...
+        self.tool_requires("cmake/[>=3.16 <4]")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        rmdir(self, os.path.join(self.source_folder, "external"))
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["CMAKE_POSITION_INDEPENDENT_CODE"] = self.options.get_safe("fPIC", True)
```
Croydon, line 129:
```suggestion
```

Seems unnecessary as Conan does this automatically

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,255 @@
...
+        if Version(self.version) <= "2.6.3":
+            tc.cache_variables["SDL2MIXER_DEBUG_POSTFIX"] = ""
+        tc.variables["SDL2MIXER_VENDORED"] = False
+        tc.variables["SDL2MIXER_SAMPLES"] = False
+        tc.variables["SDL2MIXER_CMD"] = self.options.cmd
+        if self.version == "2.8.0":
+            # The debug postfix will be removed in an upcoming release
```
Croydon, line 136: Context/Source?

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,255 @@
...
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        rmdir(self, os.path.join(self.source_folder, "external"))
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["CMAKE_POSITION_INDEPENDENT_CODE"] = self.options.get_safe("fPIC", True)
+        if Version(self.version) <= "2.6.3":
+            tc.cache_variables["SDL2MIXER_DEBUG_POSTFIX"] = ""
```
Croydon, line 131: This also manipulates `SDL2MIXER_DEBUG_POSTFIX` for some reason, like the lines a few lines down

Based on the version check this seems deleteable as this recipe only manages newer versions. But please clarify in general why this postfix is deleted by us conditionally 🤔

recipes/sdl_mixer/cmake/conanfile.py
```diff
@@ -0,0 +1,255 @@
...
+        elif self.options.vorbis == "vorbisfile":
+            tc.variables["SDL2MIXER_VORBIS"] = "VORBISFILE"
+        elif self.options.vorbis == "tremor":
+            tc.variables["SDL2MIXER_VORBIS"] = "TREMOR"
+        else:
+            tc.variables["SDL2MIXER_VORBIS"] = False
+        # WavPack
+        tc.variables["SDL2MIXER_WAVPACK"] = self.options.wavpack
+
+        # TODO: add support for dynamic loading of dependencies
```
Croydon, line 172: Was this broken in previous recipe revisions? Seemingly it got supported 🤔

Croydon:
Please edit the pull request title regarding the version, so that the merge commit message will be correct

recipes/sdl_mixer/cmake/conanfile.py
valgur: Thanks! The project uses a new enough CMake policy version (3.16) and indeed does not require the use of `cache_variables`.

recipes/sdl_mixer/cmake/conanfile.py
valgur: No idea. I'm not going to add it in this PR either way.

Croydon (approved):

Croydon:
@RubenRBS Can we please get a review? 🙂

recipes/sdl_mixer/cmake/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
```
memsharded, line 1: test_v1_package are no longer necessary, it is better to not add them, specially for new recipes (this under ``cmake`` is a new recipe)

memsharded: No need to change it now, but just FYI. Less is more, specially when the backlog in ConanCenter is so big.

memsharded (approved):

Issue #19008, opened 2023-08-01
[package] catch2/3.4.0: C++14 not compatible with C++20


hobbeshunter:
### Description

Conan thinks that a C++14 build of catch2 is compatible with a C++20 one. This is afaik wrong. E.g. the string_view support in StringMaker is missing.

### Package and Environment Details

* Package Name/Version: **catch2/3.4.0** and probably others
* Operating System+version: **Windows Server 2022**
* Compiler+version: **msvc 17**
* Conan version: **conan 2.0.9**
* Python version: **Python 3.9.13**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows

### Steps to reproduce

Use catch2 in a C++20 project where it finds an already built C++14 package in the conan center. E.g. above profile.

Forcing a build with `--build=*` fixes the problem.

### Logs

When the following happens, you successfully reproduced the bug:

```
======== Computing necessary packages ========
catch2/3.3.2: Checking 3 compatible configurations:
catch2/3.3.2: '256d8e24411ffc8da15f42452ef405d7117905dc': compiler.cppstd=14
catch2/3.3.2: Main binary package 'b85eb13063d5472984e2a54461fae73d82585560' missing. Using compatible package '256d8e24411ffc8da15f42452ef405d7117905dc'
```

```
9/24] Linking CXX executable test\Release\tests.exe
FAILED: test/Release/tests.exe test/tests_tests-f9b55b5.cmake D:/a/json2cpp/json2cpp/build/msvc/test/tests_tests-f9b55b5.cmake 
[...]

PR #21792, opened 2023-12-16
Catch2 cpp17 onwards not compatible with cpp14

Twon:
Specify library name and version:  **catch2/3.5.0**

I am submitting this patch to address https://github.com/conan-io/conan-center-index/issues/19008.  Specifically, there is a bug caused by the fact that Catch2 packages always compile to support only C++ version 14.  However, the Catch2 libraries use [C++ feature testing](https://en.cppreference.com/w/cpp/feature_test) to ensure features specific to C++ versions used are supported.  

However, I did not find examples of other libraries supporting this paradigm. I am happy to make changes to support existing practices if there are any for such library paradigms. Note, [`check_max_cppstd` was mentioned as a potential solution](https://github.com/conan-io/conan-center-index/issues/19008), but unless I am mistaken that is for enforcing a maximum cpp standard, not a minimum that depends on the currently supported cpp standard.

---

Fixes: https://github.com/conan-io/conan-center-index/issues/19008


uilianries:
> Note, https://github.com/conan-io/conan-center-index/issues/19008, but unless I am mistaken that is for enforcing a maximum cpp standard, not a minimum that depends on the currently supported cpp standard.

Yes, `check_max_cppstd` enforces a max version. But here you have to consider that Catch is a header-only project, so it's package ID is always the same. So when installing the Catch2 package, does not matter what compiler or cppstd was used, the package content is always the same. That's why `check_max_cppstd` should be used in order to check if the `settings.compiler.cppstd` is <20, otherwise, Conan will inject via CMakeToolchain, or any generator to use `--std=c++20` (or gnu20), which will not work.

So, please, revert your change and use `check_max_cppstd` as mentioned before.

recipes/catch2/3.x.x/conanfile.py
```diff
@@ -7,7 +7,7 @@
 import os
 import textwrap
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=2.1.0"
```
valgur, line 10:
```suggestion
required_conan_version = ">=1.53.0 <2 || >=2.1.0"
```
Perhaps it's possible to still allow Conan v1 support, but with the responsibility of the user to make sure that `--build` is used for incompatible cppstd versions.

Twon: Thanks for the suggestion, that sounds sensible.

Twon:
Hi @uilianriesm, is anything else required for this fix to the Catch2 package to be accepted? Happy to make any changes required.

Twon:
Hi @uilianries, is anything else required for this fix to the Catch2 package to be accepted? Happy to make any changes required.

jcar87:
Hi @Twon, thanks for your contribution. This looks sensible to me, however I'm trying to verify that this works since the specific code paths are not covered by CI (CI will build and test the package with C++14).

I'm trying the recipe as it is on the remote currently and I'm unable to reproduce the originally reported issue - is there a minimal Catch2 consumer test that we can use to verify this? thanks! (already commented on https://github.com/conan-io/conan-center-index/issues/19008#issuecomment-2090094799)

recipes/catch2/3.x.x/conanfile.py
```diff
@@ -35,6 +35,7 @@ class Catch2Conan(ConanFile):
         "console_width": "80",
         "no_posix_signals": False,
     }
+    extension_properties = {"compatibility_cppstd": False}
```
jcar87, line 38:
```suggestion
    # disallow cppstd compatibility, as it affects the ABI in this library
    # see https://github.com/conan-io/conan-center-index/issues/19008
    extension_properties = {"compatibility_cppstd": False}
```

jcar87 (approved):

Twon:
> Hi @Twon, thanks for your contribution. This looks sensible to me, however I'm trying to verify that this works since the specific code paths are not covered by CI (CI will build and test the package with C++14).
> 
> I'm trying the recipe as it is on the remote currently and I'm unable to reproduce the originally reported issue - is there a minimal Catch2 consumer test that we can use to verify this? thanks! (already commented on [#19008 (comment)](https://github.com/conan-io/conan-center-index/issues/19008#issuecomment-2090094799))
> 
> update: the following reproduces the issue:
> 
> ```
> #include <catch2/catch_test_macros.hpp>
> #include <string_view>
> 
> 
> TEST_CASE( "Test" ) {
> 
>     constexpr std::string_view foobar("foobar");
>     REQUIRE(foobar == "foobar");
>     
> }
> ```
> 
> Before this change:
> 
> ```
> 
> [clipped]
> ======== Computing necessary packages ========
> catch2/3.5.4: Checking 3 compatible configurations
> catch2/3.5.4: Compatible configurations not found in cache, checking servers
> catch2/3.5.4: '256d8e24411ffc8da15f42452ef405d7117905dc': compiler.cppstd=14
> catch2/3.5.4: Main binary package 'b85eb13063d5472984e2a54461fae73d82585560' missing. Using compatible package '256d8e24411ffc8da15f42452ef405d7117905dc': compiler.cppstd=14
> 
> [clipped]
>  error LNK2019: unresolved external symbol "public: static class std::basic_string<char,struct std::char_traits<char>,class std::allocator<char> > __cdecl Catch::StringMaker<class std::basic_string_view<char,struct std::char_traits<char> >,void>:
> :convert(class std::basic_string_view<char,struct std::char_traits<char> >)" (?convert@?$StringMaker@V?$basic_string_view@DU?$char_traits@D@std@@@std@@X@Catch@@SA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@V?$basic_string_view@DU?$char_traits@D@ 
> std@@@4@@Z) referenced in function "private: virtual void __cdecl Catch::BinaryExpr<class std::basic_string_view<char,struct std::char_traits<char> > const &,char const (&)[7]>::streamReconstructedExpression(class std::basic_ostream<char,struct std::char_traits<cha 
> r> > &)const " (?streamReconstructedExpression@?$BinaryExpr@AEBV?$basic_string_view@DU?$char_traits@D@std@@@std@@AEAY06$$CBD@Catch@@EEBAXAEAV?$basic_ostream@DU?$char_traits@D@std@@@std@@@Z) 
> ```
> 
> After this change it will just say missing binaries for `compiler.cppstd=20` if they are not available, and _will not_ fallback on using a different package.

Hi @jcar87,

Thanks, it looks like you have reproduced this now but I set up the following example to recreate this: https://github.com/Twon/ConanCatch2IssueWithCxx17

Note, you will not see this error on Linux because it does not find pre-build packaged for Catch2, so builds them locally and hides the error. It can be reproduced on Windows as there are pre-build versions of the package. See this CI failure for an example: https://github.com/Twon/ConanCatch2IssueWithCxx17/actions/runs/8924838406/job/24512078018

Many thanks!

jcar87:
Hi @Twon - thanks so much for taking the time to set it up! Incredibly helpful!  This PR will be merged shortly, thanks!

Twon:
> Hi @Twon - thanks so much for taking the time to set it up! Incredibly helpful! This PR will be merged shortly, thanks!

Many thanks for looking at this, it will be great to get it resolved!

memsharded (approved):

Issue #13418, opened 2022-10-08
Linking error in qlite3/3.39.4 because it is without libm


yushb0602:
In  sqlite3/3.39.4 the lastest version, please add libm to shell.c. in  /home/ogic/.conan/data/sqlite3/3.39.4/_/_/source/CMakeLists.txt line 170

 target_link_libraries(sqlite3-bin PRIVATE ${PROJECT_NAME} **m**)

otherewise there are  linking errors on linux

sqlite3/3.37.2: CMake command: cmake --build "/home/ogic/.conan/data/sqlite3/3.37.2/_/_/build/e32c1f48bfda04d1f3ad0456e7afb1864c5ecabb/build/Release" '--' '-j2'
[ 25%] Building C object CMakeFiles/sqlite3.dir/src/sqlite3.c.o
[ 50%] Linking C static library libsqlite3.a
[ 50%] Built target sqlite3
[ 75%] Building C object CMakeFiles/sqlite3-bin.dir/src/shell.c.o
[100%] Linking C executable sqlite3
/usr/lib/gcc-cross/aarch64-linux-gnu/11/../../../../aarch64-linux-gnu/bin/ld: libsqlite3.a(sqlite3.c.o): in function `logFunc':
sqlite3.c:(.text+0x356c8): undefined reference to `log'
/usr/lib/gcc-cross/aarch64-linux-gnu/11/../../../../aarch64-linux-gnu/bin/ld: sqlite3.c:(.text+0x35714): undefined reference to `log'
/usr/lib/gcc-cross/aarch64-linux-gnu/11/../../../../aarch64-linux-gnu/bin/ld: sqlite3.c:(.text+0x35764): undefined reference to `log'

PR #21806, opened 2023-12-18
netcdf: fix libm not found in cross build

johningve:
netcdf is unable to find the math library when cross building, similar to #13418. Ran into this issue when cross building from Linux:x86_64 to Linux:armv8. This PR adds a patch that fixes the issue.

---

### DIFF
--- a/recipes/netcdf/all/conandata.yml
+++ b/recipes/netcdf/all/conandata.yml
@@ -10,7 +10,13 @@ patches:
     - patch_file: "patches/4.7.4-0001-fix-cmake.patch"
       patch_description: "fixes for cmake target_link_libraries and using deps"
       patch_type: "conan"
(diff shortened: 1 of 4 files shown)


uilianries:
@johningve Thank you for your PR! Please, ask for access request in #4. Otherwise, the CI bot will not build your files. Regards!

jcar87:
Thanks @johningve for your contribution and thank you for linking the relevant discussion. Indeed, `libm` is a special case where `find_library` should not be used - I've pushed to your branch to fix an issue with your proposed patch (the `HAVE_LIBM` variable is also used to store the name of the library, rather than a true/false, so it was failing because it was trying to pass `-l1` to the linker, instead of `-lm`). Should be fixed now

jcar87 (approved):

RubenRBS (approved):

PR #21557, opened 2023-12-03
json-schema-validator: add version 2.3.0, downgrade nolhmann_json in older versions

toge:
Specify library name and version:  **json-schema-validator/***

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21831, opened 2023-12-20
json-schema-validator: be less choosy regarding nlohmann/json

cbaecker:
Specify library name and version:  **json-schema-validator/all**

Lib depends on nlohmann_json/3.11.2 - be less specific and allow versions above, as i faced an error after updating nlohmann_json to 3.11.3.

---

### DIFF
--- a/recipes/json-schema-validator/all/conanfile.py
+++ b/recipes/json-schema-validator/all/conanfile.py
@@ -69,7 +69,7 @@ class JsonSchemaValidatorConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("nlohmann_json/3.11.2", transitive_headers=True)
+        self.requires("nlohmann_json/[>=3.11.2 <4]", transitive_headers=True)
 
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):


uilianries:
@cbaecker Thank you for your contribution. Could please explain what error are you facing and show us a build log? Maybe it should be reported to the upstream as well.

Using version ranges for that packages is not allowed for now, please, read https://github.com/conan-io/conan-center-index/blob/d7b80d60329bd8ee6185c15d77ebd9748bc4fafa/docs/adding_packages/dependencies.md#version-ranges

uilianries:
@cbaecker Taking a look, your PR could be related to https://github.com/conan-io/conan-center-index/pull/21557. Could you please confirm that combination of versions work for you?

cbaecker:
@uilianries Thanks for your feedback.
My problem is that I wanted to update nlohmann/json to 3.11.3 because one of the improvements is relevant to my project.

Excerpt from my conanfile:
```python
def requirements(self):
        self.requires("json-schema-validator/2.2.0")
```
This led to the following error.
`ERROR: Version conflict: json-schema-validator/2.2.0->nlohmann_json/3.11.2, ->nlohmann_json/3.11.3,`

While updating json-schema-validator to use 3.11.3 would fix the problem, i (and probably others) will be faced with the same error once nlohmann_json/3.11.4 is released.
Since nlohmann/json is header-only I see no need to force the same version as in dependent libraries.

uilianries:
@cbaecker I believe the PR https://github.com/conan-io/conan-center-index/pull/21557 fixes your problem. Could you please test it? You need to install with update:

    conan install --update json-schema-validator/2.2.0
    
Or, just install the newest version: `json-schema-validator/2.3.0`

cbaecker:
Fixed with json-schema-validator/2.3.0, but now stumbling over pybind11_json still forcing nlohmann/json:3.11.2. 🙈😂

cbaecker:
As this is quite annoying i switched to `self.requires("nlohmann_json/3.11.3", force=True)`.

This comment recommends versions/version-ranges...
```
The force and override traits to solve conflicts are not recommended as a general versioning solution, just as a temporary workaround to solve a version conflict. Its usage should be avoided whenever possible, and updating versions or version ranges in the graph to avoid the conflicts without overrides and forces is the recommended approach.
```

Issue #21698, opened 2023-12-09
[package] qt/6.4.2: Invalid: Qt needs recent libstdc++, with charconv. please switch to gcc, or to libc++


n-morales:
### Description

If you try and install qt/6.4.2 with clang but with (very recent) libstdc++, conan gives the invalid configuration error

```
======== Installing packages ========
ERROR: There are invalid packages:
qt/6.4.2: Invalid: Qt needs recent libstdc++, with charconv. please switch to gcc, or to libc++
```

As far as I know, `<charconv>` is in libstdc++, at least in recent versions, so this message seems overly conservative and outdated.

### Package and Environment Details

* Package Name/Version: **qt/6.4.2**
* Operating System+version: **Manjaro Linux 23.1.0**
* Compiler+version: **Clang 16 (libstdc++ 13.2)*
* Conan version: **conan 2.0.13**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.cppstd=gnu23
compiler.libcxx=libstdc++11
compiler.version=16
os=Linux

[conf]
tools.build:cxxflags=["-stdlib=libstdc++"]
tools.build:compiler_executables={"c": "/usr/bin/clang", "cpp": "/usr/bin/clang++"}

### Steps to reproduce

conan install -u -r conancenter --requires qt/6.4.2 -pr:h=clang16 -pr:b=clang16

### Logs

<details><summary>Click to expand log</summary>

```

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.cppstd=gnu23
compiler.libcxx=libstdc++11
compiler.version=16
os=Linux
[conf]
tools.build:cxxflags=['-stdlib=libstdc++']
tools.build:compiler_executables={'c': '/usr/bin/clang', 'cpp': '/usr/bin/clang++'}

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.cppstd=gnu23
compiler.libcxx=libstdc++11
compiler.version=16
os=Linux
[conf]
tools.build:cxxflags=['-stdlib=libstdc++']
tools.build:compiler_executables={'c': '/usr/bin/clang', 'cpp': '/usr/bin/clang++'}


======== Computing dependency graph ========
qt/6.4.2: Checking remote: conancenter
qt/6.4.2: Retrieving from remote 'conancenter'...
[...]

PR #21843, opened 2023-12-21
qt6: enable clang builds

ericLemanissier:
fixes conan-io/conan-center-index#21698

The original error is:
```
/home/conan/w/prod/BuildSingleReference/.conan/data/qt/6.3.0/_/_/source/qt6/qtbase/src/corelib/text/qlocale_tools.cpp:57:10: fatal error: 'charconv' file not found
#include <charconv>
         ^~~~~~~~~~
1 error generated.
```
It can be seen in https://c3i.jfrog.io/c3i/misc/logs/pr/10261/10-linux-clang/qt/6.3.0//8378bea5856cf76fce4bce4f3e960750c8a5402f-build.txt
see also my message https://github.com/conan-io/conan-center-index/pull/10261#issuecomment-1096869182

The original problem is that the docker images used by C3I for the build of clang11 profiles had a libstdc++ which is too old.
Now that clang 11 is not built by C3I any more, it may pass.

Specify library name and version:  **qt/6.***

---


ericLemanissier:
@stevenwdv @n-morales
I expect this to fail, because [C3I uses Xenial for its clang images](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L155) and the default version of libstdc++ on xenial is 6

stevenwdv:
@ericLemanissier I had to look up what those terms meant but now I finally understand (although not why this version is used while a newer version is used for gcc). However, why should this impact my local build? I think the error should probably be removed because the conanfile has no idea what version of libstdc++ I have installed.

recipes/qt/6.x.x/conandata.yml
```diff
@@ -67,14 +67,23 @@ patches:
     - "base_path": "qtwebengine"
       "patch_description": "Workaround for too long .rps file name"
       "patch_file": "patches/c72097e_6.6.0.diff"
```
MartinDelille, line 69: shouldn't we fix the linter issue ? it is polluting the review.

MartinDelille (approved):

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/7329076005/attempts/2)

jwillikers (approved):

RubenRBS (approved):

jwillikers (approved):

PR #21869, opened 2023-12-22
Zserio 2.13.0

Mi-La:
**zserio/2.13.0**

Hello!
I'm one of the main developers of [zserio](https://zserio.org) serialization framework and I'm submitting this PR to make it available in conan center.

Zserio is already available for Python developers via [pypi](https://pypi.org/project/zserio/), for Java developers  on [maven central](https://central.sonatype.com/artifact/io.github.ndsev/zserio-runtime/2.13.0) and we would like to have also better support for our C++ developers.

Currently the package contains a recipe for Zserio C++ runtime library (static-library) and it also provides zserio generator as a jar file, since the generator itself is written in Java. The package also provides a simple cmake function which allows to generate C++ sources from zserio =*.zs= sources. The test_package currently uses only the static library since I don't know how to ensure that Java runtime will be available on CI's bulid machines and it's needed for the cmake function to be able to correctly run the generator in the jar file.

I appreciate any help how to improve the recipes, mainly how to test also the generator in the test_package. So far I see two solutions:
* test_package could try to install java via [tools.system.package_manager](https://docs.conan.io/2/reference/tools/system/package_manager.html)
* test_package could set openjdk as a [build_requirement](https://docs.conan.io/2.0/reference/conanfile/methods/build_requirements.html)

---

   - **conan-center hooks are not available for conan 2.x according to the doc**, and I would like to deal only with conan 2.x, however I've tried to run conan 1.62 with hooks using docker


recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+    def _min_cppstd(self):
+        return 11
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if Version(self.version) < "2.13.0":
+            raise ConanInvalidConfiguration("Minimum available version is 2.13.0")
+
```
uilianries, line 33:
```suggestion
```
Do not need to enforce it, the conandata.yml contains all supported versions.

recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version]["runtime"],
+            pattern="runtime_libs/cpp/*", strip_root=True)
+        shutil.move("cpp/zserio", ".")
+        shutil.move("cpp/CMakeLists.txt", ".")
+        shutil.rmtree("cpp")
```
uilianries, line 42:
```suggestion
```
Please, move files during build step, source should be immutable.

recipes/zserio/all/zserio_compiler.cmake
```diff
@@ -0,0 +1,41 @@
+function(zserio_generate_cpp GENERATED_SOURCES_OUT)
```
uilianries, line 1: I'm curious if the project does not provide this cmake macro already? It could be copied to the package folder directly.

recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        return os.path.join("lib", "cmake", "zserio")
+
+    def package(self):
+        include_dir = os.path.join(self.package_folder, "include")
+        lib_dir = os.path.join(self.package_folder, "lib")
+        copy(self, "*.h", self.source_folder, include_dir)
+        copy(self, "*.lib", self.build_folder, lib_dir)
+        copy(self, "*.a", self.build_folder, lib_dir)
+
+        copy(self, "zserio.jar", self.build_folder, os.path.join(self.package_folder, "bin"))
```
uilianries, line 69: If you need to copy a tool, this package is not a static-library, but `library` instead.

recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        copy(self, "zserio_compiler.cmake", self.export_sources_folder,
+             os.path.join(self.package_folder, self._cmake_module_path))
+
+    def package_info(self):
+        self.cpp_info.components["ZserioCppRuntime"].libs = collect_libs(self)
```
uilianries, line 74:
```suggestion
        self.cpp_info.components["ZserioCppRuntime"].libs = collect_libs(self)
```
Please, do not use `collect_libs` it hides what should be copied and does not respect linkage order in case of multiple libraries.

recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        cmake.configure()
+        cmake.build()
+
+        get(self, **self.conan_data["sources"][self.version]["compiler"], pattern="zserio.jar")
+
+    @property
+    def _cmake_module_path(self):
+        return os.path.join("lib", "cmake", "zserio")
+
+    def package(self):
```
uilianries, line 62: Please, don't forget to copy the license file to `licenses` folder too. The CCI hook will check it.

uilianries (commented):
Thank you for your first contribution! I have some question about your PR.

About the test package, there is no java installed in machines. The only native tools installed that you can expected are Conan, CMake and the compiler, any other tool should be installed as tool requirement. There is the zulu JDK available: https://conan.io/center/recipes/zulu-openjdk?version=21.0.1, but please, try to keep the test package as minimal, it should be used only to check if the expects artifacts are installed and working, not to run an entire an example. Calling a `version();` method is enough, for instance.

Mi-La:
> Thank you for your first contribution! I have some question about your PR.
> 
> About the test package, there is no java installed in machines. The only native tools installed that you can expected are Conan, CMake and the compiler, any other tool should be installed as tool requirement. There is the zulu JDK available: https://conan.io/center/recipes/zulu-openjdk?version=21.0.1, but please, try to keep the test package as minimal, it should be used only to check if the expects artifacts are installed and working, not to run an entire an example. Calling a `version();` method is enough, for instance.

Thank you very much for your help! Currently I did exactly what you propose in the test package - I only check that the .cmake file is present. If it's a big "sin" to install Java runtime as a requirement for the `test_package`, I can probably just use cmake function [`check_function_exists`](https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html) to check that also `zserio_generate_cpp` function is available and thus keep the test simple. In any case we plan to make a deeper test in our [tutorial](https://github.com/ndsev/zserio-tutorial-cpp) which should use the conan package once it becomes available.

While I'll be processing your proposals, should I switch the PR to a draft mode?

recipes/zserio/all/zserio_compiler.cmake
Mi-La: Unfortunately we don't have the cmake function yet in the current zserio release. I would like to add it to the next 2.14 release and maybe then we could just remove the 2.13 from the conan repository.

uilianries:
>  If it's a big "sin" to install Java runtime as a requirement for the test_package

Not a sin at all. We have few recipes following this approach. You would need add in your test_package.py

```
def build_requirements(self):
    tool_requires("openjdk/21.0.1")
    # or tool_requires("zulu-openjdk/21.0.1")
```

> While I'll be processing your proposals, should I switch the PR to a draft mode?

Yes, please.

recipes/zserio/all/conanfile.py
Mi-La: I found it easier than guessing the static library suffix which can be either `.a` or `.lib` and I didn't want to make a complicated platform checks as I found in some other recipes 
Windows MSVC -> lib
Windows MinGW -> a
Linux -> a
...
Is there some clean solution to this? Unfortunately the current zserio `CMakeLists.txt` doesn't have the install step defined.

recipes/zserio/all/conanfile.py
Mi-La: This is just an additional step to get **only** zserio c++ runtime sources. Something like additional functionality to the `pattern` argument of the `get` function. The zserio release contains also sources for Java and Python runtime and I just wanted to keep only C++ sources in a proper source folder. Is it still wrong?

If I will have to move it elsewhere, I would probably do it before `generate` to let cmake find the `CMakeLists.txt`.

recipes/zserio/all/conanfile.py
Mi-La: btw. `collect_libs` seems to be used also in other recipes (found 110x)

recipes/zserio/all/conanfile.py
Mi-La: Added in 53e30672d6cfd980092169f3404e5c084136aa5e.

Mi-La:
> ...
> ERROR: zserio/2.13.0: 'True' is not a valid 'options.shared' value.
> Possible values are ['False']

@uilianries
This is why I originally made the recipe as a `static-library`. Do I have to support both `shared` and `fPIC` in my recipe?

uilianries:
>  This is why I originally made the recipe as a static-library. Do I have to support both shared and fPIC in my recipe?

@Mi-La No, you just need to support what is supported by the upstream. In case it only supports static-library, please use https://github.com/conan-io/conan-center-index/blob/c7023c028fa6803d859d2990c5d9dc41fb35edf5/recipes/libenvpp/all/conanfile.py#L18 as reference. Only fPIC would be required.

Mi-La:
@uilianries 
Hello, thank you again very much for your help! I have yet several questions:
1) Currently the `package_type` is set to `static-library` even though I copy also the tooling (`zserio.jar`). Is it ok?
    * Would it be ok if I will change the `package_type` to `library` instead and would raise ConanInvalidConfiguration for `shared` option? We don't support shared library yet in the upstream, but we probably should to support it sooner or later.
2) I disabled Macos because the test_package couldn't see the environment variable `ZSERIO_JAR_FILE` which is defined in `package_info` step and I don't know how to test it otherwise than using this conan CI. We also don't support Macos in the upstream but we are supposed to support it and it would probably work if I fixed the problem with the env. variable.
   * Can you see some problem in my test_package? I do the check in  [`CMakeLists.txt`](https://github.com/ndsev/zserio-conan-center-index/blob/521e004170fe86cc3b434c048f79d4306d4f048c/recipes/zserio/all/test_package/CMakeLists.txt#L12C1-L14C9).

recipes/zserio/all/conanfile.py
uilianries: We don't filter neither modify the download package, the source folder should be immutable, just like downloading it without changes. I see the package is only 14MB, so it's not big enough. Please, discard those lines, including the `pattern` from `get`. If you need to move something, please, do it in the build, when the source code is copied in the build folder yet. You can use `self.source_folder` as your root.

recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,96 @@
...
+required_conan_version = ">=1.54.0"
+
+class ZserioConanFile(ConanFile):
+    name = "zserio"
+    description = "Zserio C++ Runtime Library"
+    license = "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index/"
+    homepage = "https://zserio.org"
+    topics = ("zserio", "cpp", "c++", "serialization")
+    package_type = "static-library"
```
uilianries, line 19:
```suggestion
    package_type = "library"
```

This recipe packages more than a static library, so it should be classified as `library`. It's the default, but can include more than only libraries. To understand more about traits, please, read https://docs.conan.io/2/reference/conanfile/methods/requirements.html#package-type-trait-inferring

recipes/zserio/all/conanfile.py
uilianries: @Mi-La Please, take a look in the cmake project template as example: https://github.com/conan-io/conan-center-index/blob/7fdbb2410f2b8e682bfc26aee7b6a0a53ad93bed/docs/package_templates/cmake_package/all/conanfile.py You do not need to specify the library extension neither its prefix. For instance: if your library name is `foobar`, it may produce `libfoobar.so` on Linux and `foobar.lib` on Windows, resulting in `self.cpp_info.libs = ["foobar"]` Conan is smart enough to understand it.

recipes/zserio/all/test_package/conanfile.py
```diff
@@ -0,0 +1,26 @@
+import os
+
+from conan import ConanFile
+from conan.tools.build import can_run
+from conan.tools.cmake import cmake_layout, CMake
+
+class ZserioTestPackageConanFile(ConanFile):
+    settings = "os", "arch", "compiler", "build_type"
+    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv"
```
uilianries, line 9:
```suggestion
    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv", "VirtualBuildEnv"
```

You exported `ZSERIO_JAR_FILE` to build_env, but did not say that you need it here. It should fix your error with Macos.

recipes/zserio/all/conanfile.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version]["runtime"],
+            pattern="runtime_libs/cpp/*", strip_root=True)
+        shutil.move("cpp/zserio", ".")
+        shutil.move("cpp/CMakeLists.txt", ".")
+        shutil.rmtree("cpp")
+
+        license_link = f"https://raw.githubusercontent.com/ndsev/zserio/v{self.version}/LICENSE"
+        download(self, license_link, "LICENSE")
```
uilianries, line 51:
```suggestion
        download(self, filename="LICENSE", **source['license'])
```

Please, move it to the conandata.yml as well, so in case of any change in the future, we will know about just by the checksum. Use https://github.com/conan-io/conan-center-index/blob/master/recipes/npcap/all/conandata.yml#L12 as example.

recipes/zserio/all/conanfile.py
uilianries: > btw. collect_libs seems to be used also in other recipes (found 110x)

Yes, it's part of the legacy, unfortunately. But well discouraged nowadays. We had problems using `collect_libs` in the past, related to linker order. Plus, is harder to debug when we don't know which library should be installed.

Mi-La: Oh, this is exactly the feature I needed :-). I was looking to many other recipes and I've always seen that they are guessing the "lib" prefix. Also in the [doc](https://docs.conan.io/2/reference/conanfile/methods/package_info.html#components) there is "libcrypto" used instead of just "crypto".
However if conan can deal with just the lib-name, it's great!

recipes/zserio/all/conanfile.py
Mi-La: Done. But I'm raising ConanInvalidConfiguration for `shared` option.

recipes/zserio/all/conanfile.py
Mi-La: Done.

recipes/zserio/all/conanfile.py
Mi-La: ok, done, using `build_script_folder` now:
https://github.com/ndsev/zserio-conan-center-index/blob/569d8ce57111559a1bdc82bf98c2f80116c93bec/recipes/zserio/all/conanfile.py#L73

recipes/zserio/all/test_package/conanfile.py
Mi-La: trying now

RubenRBS (approved):

RubenRBS (commented):
Are the docs that the hook complained about needed for the actual working of the library? We prefer not to package those in most cases :)

Mi-La:
> Are the docs that the hook complained about needed for the actual working of the library? We prefer not to package those in most cases :)

@RubenRBS: The `zserio_doc` folder is part of the release zip, and it's not needed at all for the library itself. But I was told by @uilianries during review that I shouldn't modify downloaded package in the `source` method :-). And if I understand correctly, the bot does the check right after the `source` method.
However it's not used for building and neither it's used in the `package` method, so I hope it doesn't remain anywhere at the end.

RubenRBS:
Oops, missed that while checking old reviews, sorry for the confussion! Thanks a lot for the explanation :)

RubenRBS (approved):

RubenRBS (changes requested):
Blocking while I check something with @uilianries - no need to do anything else on your side, everything looks good :)

Mi-La:
Thanks @RubenRBS, I've another question. I allowed MacOS build, while we don't officially support it by the upstream. Is it possible to somehow mark that the MacOS support is experimental? Currently I only report a warning from the `validate` step.
https://github.com/conan-io/conan-center-index/blob/4b2f8fd9a72c6e68db4e70d5cdd930e16eee4b3f/recipes/zserio/all/conanfile.py#L54-L55

uilianries:
@Mi-La I created the PR with changes needed to this PR: https://github.com/ndsev/zserio-conan-center-index/pull/1

After talking to Conan team, `static-library` fits better your case.

Mi-La:
> After talking to Conan team, `static-library` fits better your case.

Ok, I think that Conan is so super flexible that it's difficult to say what is the best solution :-). And it's yet more difficult to make the schemas follow some rules. Thanks for the support @uilianries & @RubenRBS!

uilianries (dismissed):
LGTM. Please @RubenRBS review again when possible 😸

uilianries (approved):
LGTM

Mi-La:
@uilianries I've no idea what's wrong now :-(. Do you see where is the problem?

uilianries:
@Mi-La Thank you for reporting the incident. We are facing an outage with Github: https://status.conan.io/

Your PR should be restarted automatically by the CI bot after some hour (after having Github working again). Regards.

uilianries:
Nice, working! Please @RubenRBS take a look when possible 😄

RubenRBS (approved):

Mi-La:
@uilianries I didn't want to wait to automatic restart so I pressed the "Update branch" button.
Will it now merge automatically or do I have to press the button again?

uilianries:
@Mi-La The CI bot will merge it automatically, it takes its time, so don't worry. We documented the process here: https://github.com/conan-io/conan-center-index/blob/master/docs/review_process.md#automatic-merges

In summary the CI bot scans PRs ready to be merged (this one), then merge it to master.

Mi-La:
@uilianries
I just tried to use the zserio package from conancenter and I got an error, probably related to `-fPIC` setting :-/.
```
p/lib/libZserioCppRuntime.a(BitSizeOfCalculator.cpp.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE
```
It was working when I tried the package created on my local machine so I was thinking that it is ok. With the current package in the conancenter I need to add something like `default_options = {"zserio/*:fPIC": True}` to my sample conan recipe and then zserio must be compiled on my local (only -fPIC=False is precompiled in the conancenter). Or I will have to somehow force my cmake to pass `-no-pie` option to `gcc`.

Shouldn't we switch the `fPIC` default to `True`?

(To be honest, I don't fully understand why static library needs -fPIC when linked to an executable).

uilianries:
> Shouldn't we switch the fPIC default to True?

@Mi-La  Yes!!! My mistake I didn't see that! It ALWAYS should True by default. Could you please open another PR with a hotfix?

Mi-La:
> @Mi-La Yes!!! My mistake I didn't see that! It ALWAYS should True by default. Could you please open another PR with a hotfix?

@uilianries Sure, I'll do it ASAP. Maybe there should be a mention somewhere in conan documentation - with an explanation why it's needed to be True by default - and maybe with an explanation why it's needed at all. Is there already something? Is the reason really the address space randomization?

uilianries:
@Mi-La We have it documented already 😄 https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conanfile_attributes.md#predefined-options-and-known-defaults

Also, we have hooks to detect some pre-defined cases, for instance, it could be detected the the hook (ci service) and push a warning. 

Anyway, good catch with that error! Thank you!

Mi-La:
> We have it documented already 😄 https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conanfile_attributes.md#predefined-options-and-known-defaults

@uilianries  It would be nice to have also some link with a deeper explanation why -fPIC is needed - especially for static libraries. It may not be clear to everybody (like to me :-)).

Mi-La:
@uilianries New pull request here https://github.com/conan-io/conan-center-index/pull/22234.

PR #21874, opened 2023-12-23
Glib 2.79.0

ericLemanissier:
Specify library name and version:  **glib/2.79.0**

---


ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/7307530967/attempts/1)

ericLemanissier:
So glib now requires python's `packaging` module https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3740
I guess the only way forward is to install it on all C3I runners ?
I guess glib is not the only package which will migrate from distutils to packaging

ericLemanissier:
superseded by https://github.com/conan-io/conan-center-index/pull/23093

PR #21915, opened 2023-12-26
drwav: add v0.13.14

valgur:


RubenRBS (approved):
Remember that changing the quotes in old builds makes new revisions and thus will need recompilation/redownloading from the users that rely on those packages :)

ericLemanissier (approved):

PR #21917, opened 2023-12-26
ensmallen: add v2.21.0

valgur:
.

### DIFF
--- a/recipes/ensmallen/all/conanfile.py
+++ b/recipes/ensmallen/all/conanfile.py
@@ -25,7 +25,7 @@ class ensmallenRecipe(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("armadillo/12.2.0")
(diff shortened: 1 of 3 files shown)


RubenRBS:

recipes/ensmallen/all/conanfile.py
```diff
@@ -25,7 +25,7 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("armadillo/12.2.0")
+        self.requires("armadillo/12.6.4")
```
czoido, line 28: this is ok as the only other using armadillo in CCI is gdal and already has the `12.6.4` [as dependency](https://github.com/conan-io/conan-center-index/blob/1634381974b91b14804ce996c00fdd1374a29b72/recipes/gdal/post_3.5.0/conanfile.py#L190).

czoido (approved):

PR #21939, opened 2023-12-26
bzip2: fix project URL

valgur:
The current one (http://www.bzip.org) does not look legitimate. Replaced it with https://sourceware.org/bzip2


uilianries (approved):
LGTM

jwillikers (approved):

PR #21952, opened 2023-12-26
libressl: add v3.8.2

valgur:

PR #22003, opened 2023-12-27
serf: add v1.3.10, use SConsDeps

valgur:


RubenRBS:

puetzk:
This "unable to initialize decompress status for section .debug_info" error seems unrelated to the changes being made - it looks like https://sourceware.org/bugzilla/show_bug.cgi?id=23919 (which is exposed if you mix old binutils with newer elfutils) https://sourceware.org/git/?p=elfutils.git;a=commit;h=cf10453f8252df81225796d98548ba6eac113df3

And libressl #21952 is showing the same failure... I wonder if elfutils got upgraded in the CCI build machines in December...

PR #22012, opened 2023-12-27
zookeeper-client-c: add v3.9.1, bump deps, do not write to ~/.m2

valgur:


ericLemanissier (approved):

AbrilRBS (approved):
Thanks! Approving (version bump removes conflictrs with every other CCI package that uses this dependency)

uilianries (approved):
LGTM. Good catch spotting the bug.

PR #22016, opened 2023-12-28
reflect-cpp: new recipe

RazielXYZ:
Specify library name and version:  **reflect-cpp/0.4.0** (https://github.com/getml/reflect-cpp)

Cool static reflection library that focuses on serialization and validation. Supports JSON (through YYJSON), XML (through pugixml) and flexbuffers (through flatbuffers). As far as I know it is not a dependency for anything else.

---


recipes/reflect-cpp/all/conanfile.py
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.options.get_safe("with_json", True):
```
RubenRBS, line 54: They can stay like this for now and maybe be changed in the future to avoid recompilations now, but options can not be missing with how the recipe is currently setup, so just doing `self.options.with_json` would work :)

RubenRBS (approved):
Thanks a lot for the PR and the detailed PR body message :)

recipes/reflect-cpp/all/conanfile.py
RazielXYZ: Might as well change them now!

RubenRBS (approved):

toge (approved):

PR #22022, opened 2023-12-28
gettext: add v0.22.4, better source URLs

valgur:


jcar87:
Superseded by https://github.com/conan-io/conan-center-index/pull/23033 - which fixes issue crossbuilding on macOS.

Re: the URLs - we can use it for newer versions, but we will not accept changes of existing checksums unless there is full traceability of why they're different.

PR #22023, opened 2023-12-28
glibmm: add v2.78.0, drop static build support

valgur:
The support for static builds is quite broken out of the box for this library and the amount of patching required to add support for it seems excessive to me and are a maintenance headache. Especially the code-level patches, in light of the discussion at #17338.

I would have kept the excisting patches for older versions, but I don't think I can delete the `shared` and `fPIC` options selectively without conflicting with `package_type`.


RubenRBS:

jwillikers (approved):

AbrilRBS (approved):
Thanks a lot, I think this is the way going forward :)

PR #22024, opened 2023-12-28
gsl: add v2.7.1, better URLs

valgur:
.


RubenRBS (approved):
Good idea on preferring to use `ftpmirror`.
There are a few recipes that have this issue. I've told the team to allow PRs that update the links gradually over time :) (So not every one at once for anyone reading please, thanks!)

ericLemanissier (approved):

PR #22029, opened 2023-12-28
libpsl: add v0.21.5, bump deps, simplify patching

valgur:
.


eli-schwartz:
How does this "simplify patching"? Patch files can be cleanly verified to still apply, doing the equivalent of a sed will appear to work but potentially do nothing or potentially break things without providing any feedback that it didn't do what's expected.

valgur:
@eli-schwartz It does not, from the perspective of the package consumer, at least. But it achieves the exact same result with a much lower maintenance burden. The changes here are relatively trivial and easy to re-apply, but re-applying such patches really adds up across hundreds of packages, when trying to keep most of them up to date.

> doing the equivalent of a sed will appear to work but potentially do nothing or potentially break things without providing any feedback that it didn't do what's expected

It's a valid concern, but the `replace_in_file` method accounts for this and defaults to `strict=True`, which results in an error being raised in case of the replacement not being applied.

eli-schwartz:
That's the "or potentially break things" case I mentioned. Seds have a distressing tendency to break due to upstream changes that move code around so the sed still does *something*, just not, well, the thing you originally intended it to do.

eli-schwartz:
Anyways, upstream has already changed this to support `-Dtests=false` which will disable this without any form of replacements or patching. Which is proof that you should NOT use a sed. The modifications are only needed for one version, and the next release tag, a patch will stop applying (but a sed will still apply) and you ***want*** to be notified that the patch stops applying, because it is your notification that you should stop modifying the file and start passing a build option instead.

This is a small part of why I hate seds, whether they are implemented in python with strict parameters or not.

valgur:
> Anyways, upstream has already changed this to support -Dtests=false which will disable this without any form of replacements or patching. Which is proof that you should NOT use a sed.

It's still working as expected.

danimtb (approved):

recipes/libpsl/all/conanfile.py
```diff
@@ -48,18 +46,18 @@ def layout(self):
...
         if self.options.with_idna == "icu":
-            self.requires("icu/73.2")
+            self.requires("icu/74.1")
         elif self.options.with_idna == "libidn":
             self.requires("libidn/1.36")
         elif self.options.with_idna == "libidn2":
             self.requires("libidn2/2.3.0")
         if self.options.with_idna in ("libidn", "libidn2"):
-            self.requires("libunistring/0.9.10")
+            self.requires("libunistring/1.1")
```
AbrilRBS, line 55: Podofo is the only library that uses this too, we should bring it in-line with this in a follow-up PR @danimtb @valgur

AbrilRBS (approved):

franramirez688 (approved):

franramirez688 (changes requested):
Sorry, I need to check the new variable `tests` upstream.

recipes/libpsl/all/conanfile.py
```diff
@@ -48,21 +49,22 @@ def layout(self):
...
     def build_requirements(self):
-        self.tool_requires("meson/1.2.1")
+        self.tool_requires("meson/[>=1.2.3 <2]")
         if not self.conf.get("tools.gnu:pkg_config", check_type=str):
-            self.tool_requires("pkgconf/2.0.3")
+            self.tool_requires("pkgconf/[>=2.2.0 <3]")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        self._patch_sources()
```
franramirez688, line 67:
```suggestion
```

franramirez688 (approved):
I just added the new meson option upstream `tests` so we avoid introducing the replace-in-file pattern.

Looking good to me!

AbrilRBS (approved):

PR #22030, opened 2023-12-28
libsecret: add v0.21.2

valgur:
..


RubenRBS:

recipes/libsecret/all/conanfile.py
```diff
@@ -24,24 +25,27 @@ class LibsecretConan(ConanFile):
...
         "shared": [True, False],
         "fPIC": [True, False],
+        "crypto": [False, "libgcrypt", "gnutls"],
         "with_libgcrypt": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "crypto": "libgcrypt",
         "with_libgcrypt": True,
```
uilianries, line 35: Please, deprecate `with_libgcrypt` properly. It's duplicated with `crypto`. For older versions, you can raise an Invalid configuration for crypto=gnutls

valgur: Done.

RubenRBS (dismissed):

recipes/libsecret/all/conanfile.py
```diff
@@ -24,23 +25,24 @@ class LibsecretConan(ConanFile):
...
-
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
         if self.settings.os != "Linux":
             # libgcrypt recipe is currently only available on Linux
+            self.options.crypto = False
+            self.options.with_libgcrypt = False
+        if Version(self.version) >= "0.21.2":
             del self.options.with_libgcrypt
```
uilianries, line 46:
```suggestion
```
with_libcrypt is deprecated.

recipes/libsecret/all/conanfile.py
```diff
@@ -53,23 +55,25 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("glib/2.78.1", transitive_headers=True, transitive_libs=True)
-        if self._use_gcrypt:
+        self.requires("glib/2.78.3", transitive_headers=True, transitive_libs=True)
+        if self.options.crypto == "libgcrypt" or self.options.get_safe("with_libgcrypt") == True:
```
uilianries, line 59:
```suggestion
        if self.options.crypto == "libgcrypt":
```

A deprecated option should not be part of the recipe flow.

recipes/libsecret/all/conanfile.py
```diff
@@ -86,7 +90,10 @@ def generate(self):
         tc.project_options["introspection"] = "false"
         tc.project_options["manpage"] = "false"
         tc.project_options["gtk_doc"] = "false"
-        tc.project_options["gcrypt"] = "true" if self._use_gcrypt else "false"
+        if Version(self.version) >= "0.21.2":
+            tc.project_options["crypto"] = str(self.options.crypto) if self.options.crypto else "disabled"
+        else:
+            tc.project_options["gcrypt"] = "true" if (self.options.with_libgcrypt or self.options.crypto == "libgcrypt") else "false"
```
uilianries, line 96:
```suggestion
            tc.project_options["gcrypt"] = "true" if self.options.crypto == "libgcrypt" else "false"
```

recipes/libsecret/all/conanfile.py
```diff
@@ -106,8 +113,10 @@ def package(self):
 
     def package_info(self):
         self.cpp_info.set_property("pkg_config_name", "libsecret-1")
-        self.cpp_info.requires = ["glib::glib-2.0", "glib::gobject-2.0", "glib::gio-2.0"]
-        if self._use_gcrypt:
-            self.cpp_info.requires.append("libgcrypt::libgcrypt")
         self.cpp_info.includedirs = [os.path.join("include", "libsecret-1")]
         self.cpp_info.libs = ["secret-1"]
+        self.cpp_info.requires = ["glib::glib-2.0", "glib::gobject-2.0", "glib::gio-2.0"]
+        if self.options.crypto == "libgcrypt" or self.options.get_safe("with_libgcrypt") == True:
```
uilianries, line 119:
```suggestion
        if self.options.crypto == "libgcrypt":
```

uilianries (changes requested):
Please, do not forget to delete the option from package id and add a warning message in validate()

valgur:
@uilianries That is not what's typically understood by "deprecated", imo. I would expect the feature to display a warning when used and then be removed entirely in a future update. Ignoring its value would alter the behavior of the recipe silently and unexpectedly.

uilianries:
@valgur by the contrary, having two options doing the same thing is redundant and will cause more confusion. Plus, it's documented to you revisions to avoid having broken packages.

uilianries:
> @uilianries I'm sorry, but that's breaking existing functionality with extra steps, not deprecation.
> 
> I reverted it back to the original logic. There's really no need to expose the `crypto` option to the older versions and anyone upgrading to a newer version will get a clear error message about a missing `with_libgcrypt` option.

@valgur I'll insist, right now the recipe has a logic to use an option or not, plus two options having the same behavior. Conan offers revisions to lock changes, like I said before.

uilianries (approved):
LGTM

ErniGH (approved):

perseoGI (approved):

PR #22037, opened 2023-12-28
miniupnpc: add v2.2.5, do not run upnpDiscover() in test_package

valgur:


recipes/miniupnpc/all/test_package/test_package.c
```diff
@@ -7,6 +7,8 @@ int main(void) {
     int ipv6 = 0;
     unsigned char ttl = 2;
     int error = 0;
-    struct UPNPDev *devlist = upnpDiscover(2000, multicastif, minissdpdpath, localport, ipv6, ttl, &error);
+    if (0) {
```
RubenRBS, line 10: Any reasoning for disabling this, and doing it this way?

RubenRBS (commented):
Thanks!

recipes/miniupnpc/all/test_package/test_package.c
valgur: It probes the local network for UPnP support. It takes a couple of seconds to complete and can look a bit suspicious in terms of network security. For test_package it's sufficient to see that it links correctly.

RubenRBS: Yup! Good idea. Why not remove the whole line though?

valgur: Just to confirm that linking works correctly for the library.

RubenRBS: Ahhh, I see I see, sorry I didnt catch what you meant in your first comment :) thanks as always!

RubenRBS (approved):

ericLemanissier (approved):

PR #22040, opened 2023-12-28
tk: add v8.6.13

valgur:


Ahajha:
I believe Quartz needs to be added to the Apple frameworks in package_info.

Ahajha:
Looks like the last missing symbol is from the "Uniform Type Identifiers" framework, but I'm not sure if that's the exact name (spaces are weird), I can test in a minute. Also, probably should have these extra frameworks conditional on the version.

Ahajha:
Looks like it's just "UniformTypeIdentifiers". While that does fix the link issue for me, I do get a strange error when the test package runs: "Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[TKApplication tkProcessEvent:]: unrecognized selector sent to instance 0x137e12780'". 8.6.10 works fine on my machine, so no immediate theories.

Ahajha:
I found a few prior works, and it seems that the `-ObjC` flag is necessary when linking:
https://core.tcl-lang.org/tk/tktview/85f316beb15108ac43b03fa6c8608e31f3ae5f92
https://github.com/indygreg/PyOxidizer/issues/613

I also have the new frameworks as well as the `-ObjC` flag conditional to the new version. It's possible those are added somewhere between 8.6.10 and 8.6.13, but for our purposes the condition should be fine.

Ahajha:
Not sure about the new error. It works on my machine(tm) but I have a newer OS and compiler, so it might be related to that. I've seen some suggestions but they mostly suggested modifying system settings, which doesn't seem like the right solution to me.

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/7556401529/attempts/1)

Ahajha:
Really struggling to find what the cause of this, the only thread that seems to hint at anything remotely close to this problem is [this one](https://developer.apple.com/forums/thread/710205) which suggests a bug in MacOS Ventura... But that's confusing as it works on my M2 with that OS (if I recall correctly). Not sure what Mac version the CCI runners are using.

Ahajha:
Last test is a _little_ dangerous, but I want to narrow things down a little. I have 8.6.13 pointing to 8.6.12's source (I'll try .11 if this also fails). All of these test commits will be reverted.

Ahajha:
So the regression was somewhere between the .10 and .11 patches, which at least narrows down the scope.

Ahajha:
My only theory at this point is that there's some issue with the test being run on a headless server. The error seems to be in some menu bar initialization code, and it would make sense that the GUI doesn't work on a headless server. Perhaps we could just skip running the test in 8.6.11 and later of Mac?

valgur:
Oh, wow. Nice job, @Ahajha. Thanks!

Ahajha:
@valgur Just some context, essentially I found a function that does nothing if you pass in NULL. This mirrors a lot of what @jcar87 was saying in the recent protobuf PR about the test package being simple.

The only thing I don't like is that the weird runtime dependency on MacOS won't get caught by this package, maybe we can have it so that it uses the simple test only in CCI, so that at least it's caught locally?

Ahajha:
Not sure how to fix that... I must have been using another machine, but that's definitely me.

recipes/tk/all/conanfile.py
```diff
@@ -146,7 +142,7 @@ def generate(self):
 
     def _get_default_build_system(self):
         if is_apple_os(self):
-            return "macosx"
```
Ahajha, line 149: TODO: revert

valgur:
> Not sure how to fix that... I must have been using another machine, but that's definitely me.

Force-pushing with the correct name / email should fix this,

Ahajha:
@valgur I think there was a part of that reverted commit that's still necessary. I'll try to fix the CLA problem.

valgur:
@Ahajha I already overwrote the author and force-pushed that commit, so the CLA issue should be fixed. The CLA check is having issues in all PRs today, it looks like, so it has not updated the status yet.

Ahajha:
@valgur When you get a chance could you try to fix the author for the three earlier commits? I tried just now and got some unrelated merge conflicts that I'm not sure how to fix. I want to try to get this merged since 8.6.15 and 9.0 are out now.

PR #22047, opened 2023-12-28
assimp: add v5.3.1, v5.2.5, simplify patching, drop old versions

valgur:
Also, always add `zlib` dependency since it's a transitive dependency of the non-optional `minizip` anyway.


recipes/assimp/5.x/conanfile.py
```diff
@@ -240,20 +258,52 @@ def _patch_sources(self):
...
+            (os.path.join("zip", "src", "zip.h"), "zip/zip.h"),
+        ]:
+            save(self, os.path.join(self.source_folder, "contrib", contrib_header),
+                 f"#include <{include}>\n")
+
+        # misc
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
+                        "use_pkgconfig(UNZIP minizip)", "set(UNZIP_FOUND TRUE)")
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
+                        "INSTALL( TARGETS zlib", "message(TRACE #", strict=False)
```
SpaceIm, line 306: Never use `strict=False` please. It can hide patch issues.

recipes/assimp/5.x/conanfile.py
```diff
@@ -146,66 +163,56 @@ def _depends_on_clipper(self):
...
         # - Open3DGC
         self.requires("minizip/1.2.13")
-        self.requires("utfcpp/3.2.3")
-        if Version(self.version) < "5.1.0":
-            self.requires("irrxml/1.2")
-        else:
-            self.requires("pugixml/1.13")
+        self.requires("pugixml/1.14")
+        self.requires("utfcpp/4.0.1")
+        self.requires("zlib/[>=1.2.11 <2]")
```
SpaceIm, line 179: Why don't you keep the same condition for zlib as before? I don't understand the rational explained in your first message.

SpaceIm:
I'm not sure that moving patch files to replace_in_file inside conanfile is a simplification. You lose all the CMake file context of these patches, so it can be harder for contributors to understand the all logic.

recipes/assimp/5.x/conanfile.py
valgur: I agree. Fixed.

recipes/assimp/5.x/conanfile.py
valgur: While the existing logic was valid, I found it to mostly add unnecessary maintenance overhead. There is not much point in making `zlib` an optional dependency when it's the most common dependency around and also already added to the build via the non-optional `minizip` dependency. I don't think reviewing the project to keep track of which components exactly need or don't need `zlib` is worth the effort when bumping the package version.

valgur:
> I'm not sure that moving patch files to replace_in_file inside conanfile is a simplification. You lose all the CMake file context of these patches, so it can be harder for contributors to understand the all logic.

@SpaceIm Sure, I agree. I'm not too keen on that choice of wording to summarize these changes either, but I did not manage to think of a better one at the time. "Improve maintainability" would have been a better description.

I think the tradeoff between readability and the reduced maintenance burden is absolutely worth it in this case. I did my best to keep the patching in `patch_sources()` clear and easy to follow.

Generally speaking, I do prefer the standard patch files to `_patch_sources()` replacements when the changes are relatively stable and/or complex. It was not the case here, though.

recipes/assimp/5.x/conanfile.py
SpaceIm: zlib may not be a direct dependency, maintenance overhead is orthogonal.

Distinction between transitive dependencies and direct dependencies is important, specially in conan v2.

valgur: Yes, but that's just one of the reasons. I'm still not convinced that keeping `zlib` optional is a worthwhile endeavor.

EstebanDugueperoux2 (approved):

danimtb (commented):
Agree that having patches gives more context but also in this case having the replaces in file is not too bad in this case.

Regarding the zlib in/direct dependency regarding options, why removing the previous behavior of the recipe? Did that create any issue?

Thanks!

valgur:
@danimtb Each of the optional components has a maintenance overhead due to the need to check whether the component dependencies have changed between versions. The optional zlib one has one of the most complex dependencies in this recipe while simultaneously being the least valuable ones. It's just a waste of time and effort to keep it around, imo.

danimtb (approved):

recipes/assimp/5.x/conan_deps.cmake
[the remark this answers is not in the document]
valgur: It would be, but I don't see myself finding the time to take on proper unvendoring of these dependencies upstream any time soon. I would leave it to someone more invested in the project/recipe.

jwillikers (approved):

PR #22048, opened 2023-12-28
autoconf: add v2.72

valgur:


jcar87 (changes requested):
Thanks for your contribution!

I can see that the implementation of the relocatability is different than how it's currently done for the existing versions: https://github.com/conan-io/conan-center-index/blame/master/recipes/autoconf/all/patches/2.71-0001-relocatable-autoconf.patch

Please ensure that these changes are motivated and explained  in the pull request descriptions. The existing implementation was tested thoroughly on several systems before it was merged, and to my knowledge no issues have been reported in the past several months.  
I do remember testing different alternatives, but don't remember the specifics of why findbin was discarded.

I would err on the side of caution and use the same implementation as before, unless problems are reported.

valgur:
@jcar87 I only touched these patches due to them needing rebasing and extending to one additional file for the new version. The standard function seemed like a slightly better tool for the job in a vacuum, but I agree that it's not worth the potential breakage. I reverted them back to the old logic.

jwillikers (approved):

jcar87 (approved):
Thanks @valgur !

PR #22049, opened 2023-12-28
bear: add v3.1.4, bump deps, build the correct subdir

valgur:
The [root `CMakeLists.txt`](https://github.com/rizsotto/Bear/blob/master/CMakeLists.txt) only builds vendored dependencies and runs integration tests. Neither is useful for CCI. Switched to [`source/CMakeLists.txt`](https://github.com/rizsotto/Bear/blob/master/source/CMakeLists.txt) instead.


RubenRBS:

PR #22053, opened 2023-12-29
libmysqlclient: add v8.2.0, drop some old versions

valgur:
..


RubenRBS:

PR #22076, opened 2023-12-30
openjdk fix JAVA_HOME environment variable

AlexRamallo:
Specify library name and version:  **openjdk/***

The current recipe sets JAVA_HOME using `*env_info.append`, but this results in a leading space, which breaks things that are really picky about it. For example, Gradle fails with the following error:

```
ERROR: JAVA_HOME is set to an invalid directory:  /home/aramallo/.conan2/p/b/openje4780c153b838/p /home/aramallo/.conan2/p/b/openje4780c153b838/p
```

It also gets added twice for some reason, but this fixes it regardless.

---

### DIFF
--- a/recipes/openjdk/all/conanfile.py
+++ b/recipes/openjdk/all/conanfile.py
@@ -67,8 +67,8 @@ class OpenJDK(ConanFile):
     def package_info(self):
         self.output.info(f"Creating JAVA_HOME environment variable with : {self.package_folder}")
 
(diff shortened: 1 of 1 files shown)


uilianries (approved):
LGTM. Thank you for detailing your fix.

AlexRamallo:
@uilianries Any idea what happened here?

RubenRBS:

valgur (approved):

PR #22079, opened 2023-12-31
prometheus-cpp: add new version 1.2.4, require C++11 only

toge:
Specify library name and version:  **prometheus-cpp/***

---


RubenRBS:

uilianries (changes requested):
There is a change regarding the minimal C++ standard required to build this project, it's C++11 again.

https://github.com/jupp0r/prometheus-cpp/compare/v1.1.0...v1.2.3

I would suggest trying to build everything with C++11 directly, excluding the C++14 check.

toge:
@uilianries 
Thank you for your review!
I remove C++14 check and works fine for my environtments.
Let's try cci's CI.

recipes/prometheus-cpp/all/conanfile.py
```diff
@@ -36,15 +35,19 @@ class PrometheusCppConan(ConanFile):
 
     @property
     def _min_cppstd(self):
-        return 11 if Version(self.version) < "1.1.0" else 14
+        return "14" if Version(self.version) != "1.1.0" else "11"
```
uilianries, line 38:
```suggestion
        return "14" if Version(self.version) >= "1.1.0" else "11"
```

Includes 1.1.0 and higher to C++14

recipes/prometheus-cpp/all/test_package/CMakeLists.txt
```diff
@@ -5,8 +5,8 @@ find_package(prometheus-cpp CONFIG REQUIRED)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE prometheus-cpp::push prometheus-cpp::pull)
-if(${prometheus-cpp_VERSION} VERSION_LESS "1.1.0")
```
uilianries, line 8: You could simplify all these line to `target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_14)` only. C++14 is compatible with C++11, so it should not break the test package, so we don't need this condition.

recipes/prometheus-cpp/all/conanfile.py
toge: As you [mentioned](https://github.com/conan-io/conan-center-index/pull/22079#pullrequestreview-1877170924), prometheus-cpp/1.2.4 requires only C++11 again.
So prometheus-cpp/1.1.0 is the only version to require C++14.

I realized the above code is wrong condition, I already fixed it.

recipes/prometheus-cpp/all/test_package/CMakeLists.txt
toge: I believe `cxx_std_11` is required in all versions except 1.1.0.
To simplify this code, should I use `target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)` in all versions?

valgur (approved):

danimtb (approved):

uilianries (approved):

PR #22081, opened 2023-12-31
xorg: remove xvmc as dependency

tsondergaard:
xvmc is the X-Video Motion Compensation API client library. The xvmc library is not referenced in any other package in
conan-center-index. The libXvMC-devel has unceremoniously been removed from RHEL 9. A github search for xvmc finds 15 repositories and not one of them seem to be an important consuming application of this API. libxcb-devel contains libxcb-xvmc.so which provides an XCB based API.

As the xvmc library is unused in conan-center-index, appears to be nearly unreferenced in github and has been removed from a major distribution it should not be included in xorg/system either so xorg/system can be built on RHEL 9.

Documentation for the removal of libXvMC in RHEL 9:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/considerations_in_adopting_rhel_9/index

Specify library name and version:  **xorg/system**

---

### DIFF
--- a/recipes/xorg/all/conanfile.py
+++ b/recipes/xorg/all/conanfile.py
@@ -29,7 +29,7 @@ class XorgConan(ConanFile):
                      "libxcomposite-dev", "libxcursor-dev", "libxdamage-dev", "libxdmcp-dev", "libxext-dev", "libxfixes-dev",
                      "libxi-dev", "libxinerama-dev", "libxkbfile-dev", "libxmu-dev", "libxmuu-dev",
                      "libxpm-dev", "libxrandr-dev", "libxrender-dev", "libxres-dev", "libxss-dev", "libxt-dev", "libxtst-dev",
-                     "libxv-dev", "libxvmc-dev", "libxxf86vm-dev", "libxcb-render0-dev",
+                     "libxv-dev", "libxxf86vm-dev", "libxcb-render0-dev",
                      "libxcb-render-util0-dev", "libxcb-xkb-dev", "libxcb-icccm4-dev", "libxcb-image0-dev",
                      "libxcb-keysyms1-dev", "libxcb-randr0-dev", "libxcb-shape0-dev", "libxcb-sync-dev", "libxcb-xfixes0-dev",
                      "libxcb-xinerama0-dev", "libxcb-dri3-dev", "uuid-dev", "libxcb-cursor-dev"], update=True, check=True)
@@ -39,7 +39,7 @@ class XorgConan(ConanFile):
         yum = package_manager.Yum(self)
         yum.install(["libxcb-devel", "libfontenc-devel", "libXaw-devel", "libXcomposite-devel",
                            "libXcursor-devel", "libXdmcp-devel", "libXtst-devel", "libXinerama-devel",
-                           "libxkbfile-devel", "libXrandr-devel", "libXres-devel", "libXScrnSaver-devel", "libXvMC-devel",
+                           "libxkbfile-devel", "libXrandr-devel", "libXres-devel", "libXScrnSaver-devel",
                            "xcb-util-wm-devel", "xcb-util-image-devel", "xcb-util-keysyms-devel",
                            "xcb-util-renderutil-devel", "libXdamage-devel", "libXxf86vm-devel", "libXv-devel",
                            "xcb-util-devel", "libuuid-devel", "xcb-util-cursor-devel"], update=True, check=True)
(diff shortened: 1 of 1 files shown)


tsondergaard:
@ericLemanissier, would you consider reviewing this?

ericLemanissier (approved):
System xvmc package also have no dependants

RubenRBS:
Thanks for your patience @tsondergaard, I had to investigate this to properly be able to assess the PR and that took me a bit. This looks good on my end after all, so thank you so much! :)

tsondergaard:
@RubenRBS, thanks for the kind words. It sounds from your comment like you meant to approve the PR, but it appears you haven't.

uilianries:
Hello @tsondergaard Thank for your suggestion. Is there any error or blocker related to the xorg package when using xvmc for you? If yes, please, share some log output with us.

The RHEL is more a standard for enterprise than open source community, so packages used by one distro does not reflect others. Also, Github is not a reference to check about distro packages, as most of distros keep their own package repo.

tsondergaard:
Hello @uilianries,

> Is there any error or blocker related to the xorg package when using xvmc for you? If yes, please, share some log output with us.

Ah, yes, I should have included this in the PR. Here it is:

```
xorg/system: Forced build from source
Requirements
    xorg/system#de8fc1917bb3d11cf7a1d15eccf79c16:da39a3ee5e6b4b0d3255bfef95601890afd80709 - Build
libxcb-devel-1.13.1-9.el9.x86_64
libfontenc-devel-1.1.3-17.el9.x86_64
libXaw-devel-1.0.13-19.el9.x86_64
libXcomposite-devel-0.4.5-7.el9.x86_64
libXcursor-devel-1.2.0-7.el9.x86_64
libXdmcp-devel-1.1.3-8.el9.x86_64
libXtst-devel-1.2.3-16.el9.x86_64
libXinerama-devel-1.1.4-10.el9.x86_64
libxkbfile-devel-1.1.0-8.el9.x86_64
libXrandr-devel-1.5.2-8.el9.x86_64
libXres-devel-1.2.0-14.el9.x86_64
libXScrnSaver-devel-1.2.3-10.el9.x86_64
package libXvMC-devel is not installed
xcb-util-wm-devel-0.4.1-22.el9.x86_64
xcb-util-image-devel-0.4.0-19.el9.x86_64
xcb-util-keysyms-devel-0.4.0-17.el9.x86_64
xcb-util-renderutil-devel-0.3.9-20.el9.x86_64
libXdamage-devel-1.1.5-7.el9.x86_64
libXxf86vm-devel-1.1.4-18.el9.x86_64
libXv-devel-1.0.11-16.el9.x86_64
xcb-util-devel-0.4.0-19.el9.x86_64
libuuid-devel-2.37.4-15.el9.x86_64
xcb-util-cursor-devel-0.1.3-13.el9.x86_64
ERROR: xorg/system: Error in system_requirements() method, line 40
        yum.install(["libxcb-devel", "libfontenc-devel", "libXaw-devel", "libXcomposite-devel",
        ConanException: System requirements: 'libXvMC-devel' are missing but can't install because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'
```

As I explained libXvMC-devel is not available in RHEL 9. The removal of this package from the distribution is documented [here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/considerations_in_adopting_rhel_9/index).

> The RHEL is more a standard for enterprise than open source community, so packages used by one distro does not reflect others.

The PR could have been made to only remove xvmc on RHEL 9, but since xvmc is not used by any other recipes in conan-center-index and is also unused by projects that exist in github I thought it would be sensible to remove it altogether.

> Also, Github is not a reference to check about distro packages, as most of distros keep their own package repo.

I understand. I searched github for uses of this library to look wider than conan-center-index to see how many consumers of xvmc that I could find. I found nothing substantial - just a few example programs.

uilianries:
@tsondergaard Thank you for your explanation. Your log shows a misconfiguration, not a missing package error. In order to install a system package, you need to change the package manager mode and the sudo as explained in [docs](https://docs.conan.io/2/reference/tools/system/package_manager.html#configuration-properties-that-affect-how-system-package-managers-are-invoked).

I agree in removing the package xvmc for YUM, as it affects your installation and is well documented by Red Hat. However, please, keep others with xvmc installed as users my use it (we can not track it) and may break someone. In case someone reports the same in the future, for other popular distro, we can drop that package as well.

tsondergaard:
@uilianries,

> @tsondergaard Thank you for your explanation. Your log shows a misconfiguration, not a missing package error. In order to install a system package, you need to change the package manager mode and the sudo as explained in [docs](https://docs.conan.io/2/reference/tools/system/package_manager.html#configuration-properties-that-affect-how-system-package-managers-are-invoked).

Yes, I see what you mean. That log output alone is not very good. I don't like conan to be able to install system packages, so I should have included the following:

```
$ sudo dnf install libXvMC-devel
[...]
No match for argument: libXvMC-devel
Error: Unable to find a match: libXvMC-devel
```

> I agree in removing the package xvmc for YUM, as it affects your installation and is well documented by Red Hat. However, please, keep others with xvmc installed as users my use it (we can not track it) and may break someone. In case someone reports the same in the future, for other popular distro, we can drop that package as well.

I think it is extremely unlikely that other applications will use xvmc, so I thought it better to be consistent. I will modify so xvmc is only dropped for yum/dnf to make the change more narrow.

uilianries (approved):
@tsondergaard thank you for your investigation. I'm okay removing xvmc after your information.

PR #22132, opened 2024-01-03
qt/5.x.x: Fix opengl/system dependency to not be required when using es2

### LINKED ISSUE #22116: [package] qt/5.15.11: Dependency on opengl/system always requires legacy X11 GL interface on Linux
### Description

When building qt/5.15.11 for Linux, there is a dependency on opengl/system, which requires `gl.pc`, the X11 GL interface.
Using this package for an embedded system which uses the es2 interface is problematic because the system doesn't provide or need the X11 GL interface, `libGL` / `gl.pc`.
I think systems only using Wayland will run into the same problem.

* Package Name/Version: **qt/5.15.11**
* Operating System+version: **Yocto SDK**
* Conan version: **conan 1.62.0**

### Conan profile

[settings]
os=Linux
arch=armv7hf
compiler=gcc
compiler.version=7
compiler.libcxx=libstdc++11
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install qt/5.15.11 --options:host 'qt/*:opengl=es2'

### Logs

```
Put your log output here
```

</details>

jwillikers:
The opengl/system dependency is very misleading.
For Linux, it is a requirement for the legacy gl interface used for X11.
This is not necessarily required when using the es2 OpenGL backend and likely Wayland.

This PR only requires `opengl/system` on Linux when `with_x11` is enabled.

Fixes #22116.

Requires:

- [x] #23636

---


ericLemanissier (changes requested):
Thanks, but it looks to me that the package_info is incomplete: egl is never added to the requirements of a module.
I wish I had the time to refactor the requirements(self) and package_info(self) methods to use a single method defining both components and requirements, instead of duplicating everything :(

jcar87:
would it not make sense to make depending on `egl/system` conditional on the value of `self.options.opengl`?

right now it is:
``
        if self.options.get_safe("opengl", "no") != "no":
```

but it could possibly be:
```
        if self.options.get_safe("opengl", "no") == "desktop":
        elif self.options.get_safe("opengl", "no")  == "es2":
           self.requires("egl/system") 
```

seems a bit odd to just sort of ignore the value given to the option and do it depending on a particular combination of other options and settings - may as well be a True/False or not an option at all in that case

assuming that `with_x11=False` and `opengl=es2` might work just with the above

**Comment by jwillikers** [comment]
> would it not make sense to make depending on `egl/system` conditional on the value of `self.options.opengl`?
> 
> right now it is:
> 
> ```
>         if self.options.get_safe("opengl", "no") != "no":
>             self.requires("opengl/system")
> ```
> 
> but it could possibly be:
> 
> ```
>         if self.options.get_safe("opengl", "no") == "desktop":
>             self.requires("opengl/system")
>         elif self.options.get_safe("opengl", "no")  == "es2":
>            self.requires("egl/system") 
> ```
> 
> seems a bit odd to just sort of ignore the value given to the option and do it depending on a particular combination of other options and settings - may as well be a True/False or not an option at all in that case
> 
> assuming that `with_x11=False` and `opengl=es2` might work just with the above

That seems like it should work to me.

File: recipes/qt/5.x.x/conanfile.py [hunk 1]
```diff
@@ -397,9 +397,13 @@ def requirements(self):
             self.requires("libalsa/1.2.10")
         if self.options.get_safe("with_x11"):
             self.requires("xorg/system")
+            if self.options.get_safe("opengl", "no") != "no":
+                self.requires("opengl/system")
```
uilianries, line 401:
```suggestion
            if self.options.get_safe("opengl", "no") == "desktop":
                self.requires("opengl/system")
            elif self.options.get_safe("opengl", "no")  == "es2":
                self.requires("egl/system") 
```

uilianries:
Suggestion https://github.com/conan-io/conan-center-index/pull/22132#issuecomment-2250429768 applied in the commit [d69889f](https://github.com/conan-io/conan-center-index/pull/22132/commits/d69889f97689a0c1e46bf79215468a8190fa0a99)

franramirez688:
Not stale.

PR #22134, opened 2024-01-03
dbus: with_x11=True does not mean libdbus depends on xorg/system

### LINKED ISSUE #22133: [bug] dbus: generated dbus-1.pc incorrectly has Requires: xorg
### Environment details

When building qt with with_dbus=True and dbus option with_x11=True I see a dbus-1.pc like this being generated:

```
$ cat ./p/b/qt1f8e4c91c1b89/b/dbus-1.pc
prefix=/home/ts/.conan2/p/b/dbuse119f1e6f4a56/p
libdir=${prefix}/lib
includedir=${prefix}/include
includedir1=${prefix}/include/dbus-1.0
includedir2=${prefix}/lib/dbus-1.0/include
bindir=${prefix}/bin

Name: dbus-1
Description: Conan package: dbus-1
Version: 1.15.8
Libs: -L"${libdir}" -ldbus-1 -lrt -lpthread
Cflags: -I"${includedir}" -I"${includedir1}" -I"${includedir2}"
Requires: expat xorg
```

It incorrectly lists xorg as dependency, which is incorrect. When dbus is built with with_x11 the libdbus-1 library is not affected - the only impact is that a binary `dbus-launch` is built that depends on x11.

* Operating System+version: AlmaLinux 9.3
* Compiler+version: gcc (GCC) 11.4.1 20230605 (Red Hat 11.4.1-2)
* Conan version: 2.0.16
* Python version: 3.9.18

### Steps to reproduce

_No response_

### Logs

_No response_

tsondergaard:
When dbus is built with with_x11 an extra executable `dbus-launch` will be built. Only this executable depends on x11. By adding the explicit requirement trait visible=False we avoid passing all the X11 libraries in xorg/system as link libraries to applications that link with dbus.

Fixes #22133

---


tsondergaard:
The Conan v1 build failure is caused by this:

```
FAILED: dbus/libdbus-internal.a.p/dbus-sysdeps-util-win.c.obj 
"cl" "-Idbus\libdbus-internal.a.p" "-Idbus" "-I..\src\dbus" "-I." "-I..\src" "/MDd" "/nologo" "/showIncludes" "/utf-8" "/W2" "/Od" "/Zi" "-D_GNU_SOURCE" "/wo4018" "/wd4090" "/wd4101" "/wd4127" "/wd4244" "/we4002" "/we4003" "/we4013" "/we4028" "/we4031" "/we4047" "/we4114" "/we4133" "/Fddbus\libdbus-internal.a.p\dbus-sysdeps-util-win.c.pdb" /Fodbus/libdbus-internal.a.p/dbus-sysdeps-util-win.c.obj "/c" ../src/dbus/dbus-sysdeps-util-win.c
../src/dbus/dbus-sysdeps-util-win.c(1444): error C2022: '312': too big for character
../src/dbus/dbus-sysdeps-util-win.c(1450): error C2022: '312': too big for character
```

I looked at the affected lines [../src/dbus/dbus-sysdeps-util-win.c:1444](https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.15.6/dbus/dbus-sysdeps-util-win.c?ref_type=tags#L1444) and [../src/dbus/dbus-sysdeps-util-win.c:1450](https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.15.6/dbus/dbus-sysdeps-util-win.c?ref_type=tags#L1450) - they both contain DBUS_PREFIX which is a value configured by conan. Looking higher in the log I can see that the prefix path that conan uses:

```
dbus/1.15.6: Meson configure cmd: meson setup --native-file "C:/J2/w/prod-v1/bsr/82476/cdfcb/s\a21702\1\build-debug\conan\conan_meson_native.ini" "C:/J2/w/prod-v1/bsr/82476/cdfcb/s\a21702\1\build-debug" "C:/J2/w/prod-v1/bsr/82476/cdfcb/s\a21702\1\src" -Dprefix="C:/J2/w/prod-v1/bsr/82476/cdfcb/s\470296\1"
```

`\470` (octal) is 312 decimal, so the problem is definitely caused by Conan v1 putting backslashes in the path. Looking at the Conan v2 logs this problem appears to have been fixed there. I believe this is due to commit 89f01e14860fd850 which was introduced in Conan 2.0.9 via https://github.com/conan-io/conan/pull/14295. @memsharded , @RubenRBS would it be possible to cherry-pick this fix to Conan v1? Or is it better to disable the Windows build of dbus on CI with Conan v1?

jwillikers (approved):
LGTM!

valgur:
Just to be clear, does `dbus-launch` always link against X11 statically? Otherwise it will still need the shared libraries at runtime.

tsondergaard:
@valgur,

> Just to be clear, does `dbus-launch` always link against X11 statically? Otherwise it will still need the shared libraries at runtime.

It is not the intention with this PR that it should indicate that you don't need the shared libraries at runtime. I only want to communicate that if you link with libdbus (shared or static) then you do not need to linx with xorg/system.

valgur:
@tsondergaard Conan does not yet have a way to model runtime-only dependencies well, unfortunately.
The best solution I can suggest would be to add a separate component in `package_info()` for the library and the executable and add the `.requires.append("xorg::x11")` only for the latter.

It would not hurt to set the exported requires info for glib and other dependencies in a more exact manner either, since the recipe is probably overlinking many irrelevant libraries from the deps. This can be left for another PR, though.

Anyway, I checked `dbus-launch` with `objdump -p` and it only lists `pthread` and `libc` as runtime dependencies, so marking `xorg` as `visible=False` should be ok.

tsondergaard:
Weird, that you cannot see any X11 dependencies on dbus-launch. This is what I see on the dbus binaries from my AlmaLinux distribution:

```
$ for i in $(rpm -ql dbus-libs dbus-tools dbus dbus-common dbus-x11 | grep -E 'lib64/|bin/'); do echo $i; ldd $i |&grep -i libx; d
one
/usr/lib64/libdbus-1.so.3
/usr/lib64/libdbus-1.so.3.19.13
/usr/bin/dbus-monitor
/usr/bin/dbus-send
/usr/bin/dbus-update-activation-environment
/usr/bin/dbus-uuidgen
/usr/bin/dbus-launch
        libX11.so.6 => /lib64/libX11.so.6 (0x00007fcec950e000)
        libxcb.so.1 => /lib64/libxcb.so.1 (0x00007fcec91fd000)
        libXau.so.6 => /lib64/libXau.so.6 (0x00007fcec8f70000)
```

valgur:
@tsondergaard My comment was about the executable found in the package binary directory. It's ok for the system executable to have different behavior.

valgur (approved):
LGTM. DBus does not use the X11 libraries at all, even for the executables. It only needs X11 to be present during the build for the "X11 autolaunch" feature to be enabled.

I would add the `x11_autolaunch` Meson project option as an option to the recipe as well.

https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.15.8/meson.build?ref_type=tags#L613-629

PR #22138, opened 2024-01-04
rapidjson/* Specify that rapidjson is a header-only library.

bemehiser:
rapidjson/*

Specify that the library is a header-only library to get rid of a warning.
```
ld64.lld: warning: directory not found for option -L/Users/me/.conan2/p/rapidf6bdd1718c30a/p/lib
```

Feature documentation.
https://docs.conan.io/2/tutorial/creating_packages/other_types_of_packages/header_only_packages.html?highlight=libdirs

---

### DIFF
--- a/recipes/rapidjson/all/conanfile.py
+++ b/recipes/rapidjson/all/conanfile.py
@@ -13,6 +13,8 @@ class RapidjsonConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "http://rapidjson.org"
     license = "MIT"
+    package_type = "header-library"
+    package_id_embed_mode = "minor"
     settings = "os", "arch", "compiler", "build_type"
     no_copy_source = True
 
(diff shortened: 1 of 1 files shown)


recipes/rapidjson/all/conanfile.py
RubenRBS, line 1: If this is a header-only library, we might want to add 
```
```

but we're investigating adding extra attributes too

bemehiser: Do you want me to add that? I assumed that having a minimal change set would give be the best odds of actually getting this merged.

RubenRBS: Yes! Let's also add `package_id_embed_mode = "minor"` - let's see if we can avoid having so many missing dependencies this way for header-only packages :)

bemehiser: Done.

RubenRBS (approved):

jwillikers (approved):

sizeak:
After thing change I'm getting a `ERROR: 'minor' is not a known package_id_mode` error when I try and update my package versions. Any idea what the fix is?

sizeak:
Fix: https://github.com/conan-io/conan-center-index/pull/22281

PR #22152, opened 2024-01-04
pcl: add v1.14.0, use llvm-openmp for OpenMP

valgur:
I scanned the diff against previous version and there were no changes to the list of components or their dependencies, either internal or external ones.


recipes/pcl/all/conanfile.py
```diff
@@ -441,52 +454,53 @@ def source(self):
...
+            tc.variables[f"BUILD_{comp}"] = False
 
         tc.cache_variables["PCL_ENABLE_SSE"] = self.options.get_safe("use_sse", False)
 
+        # Do not overwrite CMakeToolchain variables with cache variables
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
         tc.generate()
 
         deps = CMakeDeps(self)
-        deps.set_property("eigen", "cmake_file_name", "EIGEN")
```
EstebanDugueperoux2, line 489: This removal seems the origin of the pipelines failures.

perseoGI:
Hi @valgur, thank you for your contributions!
As mentioned in #24562 `v1.14.1` has fixed some errors which were being patched. 
Instead of adding `v1.14.0` could you better add the latest release?

Thank you in advance!

weypro:
I want PCL 1.14.1, and I'm really looking forward to it being merged.

perseoGI:
Hi @valgur is this PR still in draft?

valgur:
It should be ready to be merged. Needs a build to verify that, though.

Of course, I would love to use `openmp/system` instead of copying the flag export logic for the hundredth time (#22360), but it should work regardless of that.

perseoGI:
> It should be ready to be merged. Needs a build to verify that, though.
We are under maintaining right now, I'll launch them manually

> Of course, I would love to use `openmp/system` instead of copying the flag export logic for the hundredth time (#22360), but it should work regardless of that.
We have been discussing with the team and the potential risk of mixing different openMP backends and other problems we sow do not compensate the extra copy-paste code.
I know how you feel, I also do not like to repeat myself.
We will study more in deep in a near future to find a robust solution. Until that, this is a good to go way :)

valgur:
> We have been discussing with the team and the potential risk of mixing different openMP backends and other problems we sow do not compensate the extra copy-paste code.

Thanks, I appreciate the consideration you have put into the topic with the team. The mixing of OpenMP runtimes was a valid concern for the initial compromise suggestion of defaulting to `llvm-openmp` as the OpenMP provider on CCI. The `openmp/system` recipe avoids that altogether and only uses `llvm-openmp` for CLang and AppleClang, so it is worth another look, if you haven't already. Unless you are concerned about the minor risk of mixing system libomp with one from Conan as well?

valgur:
@AbrilRBS Ok, I reverted most of the OpenMP-related changes. I only kept the use of more correct and robust OpenMP linker and cxx flags instead of adding it as a system lib (the name of which is not fixed on MSVC, for example).

shkwon98:
@valgur Hi, as a PCL user, thank you for your efforts in supporting conan.
Currently, 1.13.1, the only PCL version supported by conan, has a problem with Boost dependencies when building the pcl_gpu_containers target. This issue appears to have been resolved in 1.14.0-rc1, so support in conan is needed.

Looking at the above history, it seems that the check failed because of using the pre-release version of qhull in the Pre-Checks step. Is there a particular reason why you did not use version 8.0.1?

valgur:
The libqhull PR #24532 should be merged first. PCL links against the re-entrant libqhull_r library, but the libqhull recipe does not correctly distinguish between the two in package_info().

PR #22170, opened 2024-01-05
libmodplug: add libcxx dependency for both shared and static builds

valgur:
The macOS build in #21782 appears to be failing due to a missing libcxx dependency in this repo: https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/21782/5-macos-clang/sdl_mixer/2.6.3//summary.json

It only happens for shared builds, which correlates with the libcxx being only exported as a dependency for static builds in this recipe.

@SSE4 or @SpaceIm Can you maybe share your reasoning for only adding libcxx on static builds? This is frequently applied that way in other recipes as well.

### DIFF
--- a/recipes/libmodplug/all/conanfile.py
+++ b/recipes/libmodplug/all/conanfile.py
@@ -70,6 +70,6 @@ class LibmodplugConan(ConanFile):
             self.cpp_info.system_libs.extend(["m"])
         if not self.options.shared:
             self.cpp_info.defines.append("MODPLUG_STATIC")
-            libcxx = stdcpp_library(self)
-            if libcxx:
-                self.cpp_info.system_libs.append(libcxx)
+        libcxx = stdcpp_library(self)
+        if libcxx:
+            self.cpp_info.system_libs.append(libcxx)
--- a/recipes/libmodplug/all/conanfile.py
+++ b/recipes/libmodplug/all/conanfile.py
@@ -70,6 +70,6 @@ class LibmodplugConan(ConanFile):
             self.cpp_info.system_libs.extend(["m"])
         if not self.options.shared:
             self.cpp_info.defines.append("MODPLUG_STATIC")
-        libcxx = stdcpp_library(self)
-        if libcxx:
-            self.cpp_info.system_libs.append(libcxx)
+            libcxx = stdcpp_library(self)
+            if libcxx:
+                self.cpp_info.system_libs.append(libcxx)
--- a/recipes/libmodplug/all/test_package/conanfile.py
+++ b/recipes/libmodplug/all/test_package/conanfile.py
@@ -9,6 +9,10 @@ class TestPackageConan(ConanFile):
     generators = "CMakeToolchain", "CMakeDeps", "VirtualRunEnv"
     test_type = "explicit"
 
+    def configure(self):
+        self.settings.rm_safe("compiler.cppstd")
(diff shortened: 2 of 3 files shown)


SpaceIm:
Libcxx is already linked in shared libraries, so it doesn't make sense to add libcxx in this case.

SpaceIm:
Just a reminder: libcxx trick in package_info() is a workaround for a design flaw in conan (there is no way to express in package_info() that a library/component can link to C++ standard library or not (so that it could be translated properly in CMakeDeps), so it breaks for pure C consumers relying on static C++ libraries with C API.

This workaround is not even sufficient for C++ libraries with C API with C++ dependencies.

valgur:
> Libcxx is already linked in shared libraries, so it doesn't make sense to add libcxx in this case.

Are you sure that this is done for pure-C consuming libraries as well?

RubenRBS (dismissed):

SpaceIm:
> > Libcxx is already linked in shared libraries, so it doesn't make sense to add libcxx in this case.
> 
> Are you sure that this is done for pure-C consuming libraries as well?

What do you mean?

Regarding this PR, I don't understand this modification. test package works fine without this modification, while being pure C and consuming this C++ library, so why this PR?

SpaceIm:
By the way, locally I can build sdl_mixer shared, with libmodplug shared on macOS (I'm using revision of sdl_mixer from https://github.com/conan-io/conan-center-index/pull/17150).

SpaceIm:
See https://github.com/conan-io/conan-center-index/pull/21782#discussion_r1443869579

valgur:
> If there was an issue in test package, I would understand, but there is none.

Indeed. This PR can be closed.

> Shared libmodplug already links to libcxx, it's done during link phase of libmodplug build. Consumers don't have to link private dependencies of a shared library.

Are you sure about that? This would require libcxx to be linked statically into the shared libmodplug library, but this is not the default behavior and requires an additional `-static-libstdc++` flag.

After thinking about it for a moment, the linking of shared libmodplug works fine without an explicit libcxx dependency due to libcxx already being included as a dynamic dependency in the `.dynamic` section of the libmodplug .so file.

Whether the libcxx dependency should be explicitly listed in `system_libs` is a somewhat broader question, still. It's not required when using Conan dependencies in a typical fashion, but I think it would not hurt to be included for more robust deployment of Conan libriaries. If you are deploying shared-only pure-C libraries from Conan, then you would generally assume that you don't need to dynamically link against and redistribute the libcxx shared lib. This is even more of the case for LLVM's libc++, which is usually not available on Linux distros out of the box.

With that in mind, I would generally not bother with making the libcxx `system_lib` conditional on the build being static.

SpaceIm:
> > Shared libmodplug already links to libcxx, it's done during link phase of libmodplug build. Consumers don't have to link private dependencies of a shared library.
> 
> Are you sure about that? This would require libcxx to be linked statically into the shared libmodplug library, but this is not the default behavior and requires an additional -static-libstdc++ flag.

Why do you think that it should be linked to static C++ lib? There is no difference between static linking and dynamic linking here, once libA linked to libB, if libB is linked to libC, linker doesn't need to know libA if it's is a private dependency of libB, libA being shared or static doesn't matter (some gnu linker on Linux may want to know location of libA, but it's another problem).

PR #22171, opened 2024-01-05
RocksDB v8.8.1

Ignition:
Specify library name and version:  **rocksdb/8.8.1**

Moving dependency management over to conan, this is a needed dependency

---


recipes/rocksdb/all/conanfile.py
```diff
@@ -87,7 +91,7 @@ def requirements(self):
 
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):
-            check_min_cppstd(self, 11)
+            check_min_cppstd(self, self._min_cppstd)
```
uilianries, line 94:
```suggestion
            check_min_cppstd(self, self._min_cppstd)
         minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
            raise ConanInvalidConfiguration(
                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
            )
```

recipes/rocksdb/all/conanfile.py
```diff
@@ -51,6 +51,10 @@ class RocksDBConan(ConanFile):
         "use_rtti": False,
     }
 
+    @property
+    def _min_cppstd(self):
+        return "11" if Version(self.version) < "8.8.1" else "17"
+
```
uilianries, line 57:
```suggestion

    @property
    def _compilers_minimum_version(self):
        return {} if self._min_cppstd == "11" else {
            "apple-clang": "10",
            "clang": "7",
            "gcc": "7",
            "msvc": "191",
            "Visual Studio": "15",
        }

```

recipes/rocksdb/all/conanfile.py
```diff
@@ -174,6 +178,9 @@ def package(self):
             self._remove_static_libraries()
             self._remove_cpp_headers() # Force stable ABI for shared libraries
         rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        pc_file = os.path.join(self.package_folder, "lib", "pkgconfig", "rocksdb.pc")
+        if os.path.exists(pc_file):
+            os.remove(pc_file)
```
uilianries, line 183:
```suggestion
        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
```

Just obliterate the pkgconfig in case of existing. It does not raise error when no having it installed.

uilianries (commented):
@Ignition Thank you for your pull request!

uilianries:
The test package is configured for C++11, not C++17. You need to update the CMake file in test package as well:

https://github.com/Ignition/conan-center-index/blob/rocksdb_8.8.1/recipes/rocksdb/all/test_package/CMakeLists.txt#L17

`    target_compile_features(${PROJECT_NAME}_cpp PRIVATE cxx_std_17)`

uilianries (approved):
LGTM

ericLemanissier (approved):

PR #22197, opened 2024-01-07
libxpm: add v3.5.17

valgur:
Also:
* Drop the custom `.def` file for Windows DLLs. It's not clear how it was generated and `WINDOWS_EXPORT_ALL_SYMBOLS` should be ok for such a small library.
* Added three source files that were missing for some reason and caused the build to fail for me.
* Add a comment to clarify why the `*P.c` cannot be used on Windows.
* Explicitly disable the optional `--enable-stat-zfile` feature that requires `gzip` to be present to automatically decompress `.xpm.gz` files. Could be added later as an option.


jwillikers (approved):

recipes/libxpm/all/CMakeLists.txt
```diff
@@ -41,22 +40,27 @@ add_library(
...
+      src/CrPFrI.c
       src/RdFToP.c
       src/WrFFrP.c
   )
   target_include_directories(Xpm PRIVATE "${X11_INCLUDE_DIR}")
   target_link_libraries(Xpm PRIVATE "${X11_LIBRARY}")
 endif()
 
+if(WIN32)
+  set_target_properties(Xpm PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
```
danimtb, line 60: I know the usage of `WINDOWS_EXPORT_ALL_SYMBOLS` is extended across several recipes, however, in recent dicussions we agreed to not exposing symbols if the library does not support it natively. So, in this case, I think we can just drop it and set shared as not supported for this recipe. Thanks 🙏

valgur: This is equivalent to what the Autotools build system in the project would do under Windows. I understand the concern, but I don't think it is justified for a simple C library.

valgur: The non-CMake build system in the project uses libtool to handle shared libs. From libtool docs:
> With Microsoft tools, Libtool digs through the object files that make up the library, looking for non-static symbols to automatically export. I.e., Libtool with Microsoft tools tries to mimic the auto-export feature of contemporary GNU tools. It should be noted that the GNU auto-export feature is turned off when an explicit __declspec(dllexport) is seen. The GNU tools do this to not make more symbols visible for projects that have already taken the trouble to decorate symbols. There is no similar way to limit what symbols are visible in the code when Libtool is using Microsoft tools. In order to limit symbol visibility in that case you need to use one of the options -export-symbols or -export-symbols-regex.

In other words, it runs `dumpbin` on the .o files and creates a .def file from the listed symbols. This is pretty much exactly what `WINDOWS_EXPORT_ALL_SYMBOLS` accomplishes.

valgur: @danimtb

AbrilRBS (approved):
Approving this one from the discussion in the review and it being a C library

PR #22231, opened 2024-01-09
thrust: add v2.2.0

valgur:


valgur:
TODO: the recipe should be updated to use https://github.com/NVIDIA/cccl/blob/main/thrust/thrust/cmake/thrust-config.cmake, which enables the backend to be selected in the consuming CMakeLists.txt.

Issue #19418, opened 2023-08-25
[package] itk/5.1.2: configuration fails with C++20 or higher


SpaceIm:
### Description

With C++20 explicitly set in profile, CMake configuration fails because a third party lib vendored in itk hardcodes a range of supported C++ standards. It can't work without https://github.com/InsightSoftwareConsortium/ITK/pull/2034

### Package and Environment Details

* Package Name/Version: **itk/5.1.2**
* Operating System+version: **Linux Ubuntu 22.04**
* Compiler+version: **GCC 12**
* Conan version: **conan 2.0.9**
* Python version: **Python 3.10.12**

### Conan profile

```
[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=12
compiler.libcxx=libc++
compiler.cppstd=20
build_type=Release
```

### Steps to reproduce

```
conan install --requires "itk/5.1.2" -b missing
``` 

### Logs

<details><summary>Click to expand log</summary>

```
itk/5.1.2: Building from source
itk/5.1.2: Package itk/5.1.2:0b0d8ac25e5c47ec537192c9de02e0c01ffc323e
itk/5.1.2: Copying sources to build folder
itk/5.1.2: Building your package in /home/spaceim/.conan2/p/b/itk7a02baa2df074/b
itk/5.1.2: Calling generate()
itk/5.1.2: Generators folder: /home/spaceim/.conan2/p/b/itk7a02baa2df074/b/build/Release/generators
itk/5.1.2: CMakeToolchain generated: conan_toolchain.cmake
itk/5.1.2: CMakeToolchain generated: CMakePresets.json
itk/5.1.2: CMakeToolchain generated: ../../../src/CMakeUserPresets.json
itk/5.1.2: Generating aggregated env files
itk/5.1.2: Generated aggregated env files: ['conanbuild.sh', 'conanrun.sh']
itk/5.1.2: Calling build()
itk/5.1.2: Apply patch (conan): inject thirdparties from conan
itk/5.1.2: Running CMake.configure()
[...]

PR #22235, opened 2024-01-10
Fix itk/5.1.2 with cpp20

fdgStilla:
Fix https://github.com/conan-io/conan-center-index/issues/19418

---

### DIFF
--- a/recipes/itk/all/conandata.yml
+++ b/recipes/itk/all/conandata.yml
@@ -10,6 +10,18 @@ patches:
     - patch_file: "patches/0001-find-conan-packages.patch"
       patch_description: "inject thirdparties from conan"
       patch_type: "conan"
+    - patch_file: "patches/0002-fix-gcc11-std20.patch"
+      patch_description: "Fix compilation with cppstd for gcc"
+      patch_type: "portability"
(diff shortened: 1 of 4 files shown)


uilianries (changes requested):
@fdgStilla Thank you for your backport, but I see ITK offers new releases (RC) with these patches incorporated already.

I would ask to not try backporting it, but add a new version instead. So, who needs to use >=C++17 should include an official version already patched to support C++20. 

https://github.com/InsightSoftwareConsortium/ITK/tags

https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4rc02/Modules/Core/Common/include/itkThreadPool.h

uilianries:
> Can we add release candidate to conan?

Yes. It's not encouraged because we will have a temporary release which will be available for few time and occupy not only Conan Ceneter, but require CI resource to be built. Same for alpha and beta versions.

However, some projects delay many months between each release, so providing RC is fine. We just don't recommend bumping version for each RC in case they are not really used by users.

AbrilRBS:
Closing in favour of adding the recently released 5.4 version, will do so in the following days. Thanks a lot for your patience in this one :)

PR #22281, opened 2024-01-12
rapidjson/*: Fix incorrect `package_id_embed_mode` value.

sizeak:
Specify library name and version:  **rapidjson/***

The last update to the rapidjson recipe included a `package_id_embed_mode` value of `minor` instead of `minor_mode` which caused build errors such as `AttributeError: 'RequirementInfo' object has no attribute 'minor'` and `'minor' is not a known package_id_mode`.

Problem found in conjunction with @matthewkandiah-sonocent

---

### DIFF
--- a/recipes/rapidjson/all/conanfile.py
+++ b/recipes/rapidjson/all/conanfile.py
@@ -22,26 +22,12 @@ class RapidjsonConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def source(self):
-        get(
-            self,
-            **self.conan_data["sources"][self.version],
-            strip_root=True,
-            destination=self.source_folder
-        )
+        get(self, **self.conan_data["sources"][self.version], strip_root=True,
+                    destination=self.source_folder)
 
     def package(self):
-        copy(
-            self,
-            pattern="license.txt",
-            src=self.source_folder,
-            dst=os.path.join(self.package_folder, "licenses"),
-        )
-        copy(
-            self,
-            pattern="*",
-            src=os.path.join(self.source_folder, "include"),
-            dst=os.path.join(self.package_folder, "include"),
-        )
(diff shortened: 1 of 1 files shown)


klausholstjacobsen:
This new recipe fails with conan 2.0.16 causing: 
```'minor' is not a known package_id_mode```

Docs suggest that this should be `minor_mode` rather than `minor`.

Regards
Klaus

**Review by SpaceIm — APPROVED** [review]

File: recipes/rapidjson/all/conanfile.py [hunk 1]
```diff
@@ -14,7 +14,7 @@ class RapidjsonConan(ConanFile):
     homepage = "http://rapidjson.org"
     license = "MIT"
     package_type = "header-library"
-    package_id_embed_mode = "minor"
+    package_id_embed_mode = "minor_mode"
```
SpaceIm, line 17: Actually, what does it mean `minor_mode` for conan when versions are "cci" versions like rapidjson?

klausholstjacobsen (approved):

ilya-lavrenov (approved):

recipes/rapidjson/all/conanfile.py
joakimono: For "cci" versions, major is "cci" and minor is date, e.g. "20220822"; patch is `None`. My understanding is that setting `package_id_embed_mode = "minor_mode"` (as opposed to the default `full_mode`), leads to a situation where a consuming library only will need to rebuild their binary if minor (or major) version has changed. With `full_mode`, this happens for recipe revisions as well.

Relevant conan docs:
[Package id modes](https://docs.conan.io/2.0/reference/conanfile/attributes.html#package-id-embed-non-embed-unknown-mode)
[The effect of dependencies on `package_id`](https://docs.conan.io/2.0/reference/binary_model/dependencies.html#reference-binary-model-dependencies)

PR #22257, opened 2024-01-11
matplotplusplus: new recipe

valgur:
Adds `matplotplusplus`: https://github.com/alandefreitas/matplotplusplus

Matplot++: A C++ Graphics Library for Data Visualization 📊🗾

- Resolves #7086
- Resolves #11058

[Packaging status](https://repology.org/project/matplotplusplus/versions)

PR #22290, opened 2024-01-13
cimg: add v3.3.3, expose all available options, fix issues with fftw and opencv, avoid overlinking

valgur:
Should also hopefully fix the build error in #22257 caused by the default value for `cimg_display` being incorrect on Macos.

### DIFF
--- a/recipes/cimg/all/test_package/conanfile.py
+++ b/recipes/cimg/all/test_package/conanfile.py
@@ -15,7 +15,7 @@ class TestPackageConan(ConanFile):
         self.options["cimg"].enable_ffmpeg = True
         self.options["cimg"].enable_fftw = True
         self.options["cimg"].enable_heif = True
-        self.options["cimg"].enable_jpeg = True
+        self.options["cimg"].enable_jpeg = "libjpeg"
         self.options["cimg"].enable_magick = False # Not yet Conan 2.0 compatible
         self.options["cimg"].enable_opencv = False # OpenCV v3 Requires OpenEXR v2
         self.options["cimg"].enable_openexr = True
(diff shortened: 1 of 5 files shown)


recipes/cimg/all/test_package/conanfile.py
```diff
@@ -9,6 +9,26 @@ class TestPackageConan(ConanFile):
...
+        self.options["cimg"].enable_openmp = True
+        self.options["cimg"].enable_png = True
+        self.options["cimg"].enable_tiff = True
+        self.options["cimg"].enable_tinyexr = False # Conflicts with OpenEXR and ZLib
+        self.options["cimg"].enable_zlib = True
+        if self.settings.os in ["Linux", "FreeBSD", "Windows"]:
+            self.options["cimg"].enable_display = True
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.options["cimg"].enable_xrandr = True
+            self.options["cimg"].enable_xshm = True
```
RubenRBS, line 30: Why is this needed? Usually test packages should be much simpler than this

RubenRBS (commented):
Thhanks!

RubenRBS (approved):

jcar87:
For future reference - and this is something that we need to update in the guidance for all of Conan Center - exposing _all_ options a library might have is not a goal in Conan Center - rather, we'd like to expose the most likely to be used, and most likely to be useful.

On the other hand, I appreciate this particular recipe has a history of doing this already - but Conan options where designed for things that affect (in most cases) the package contents, and none of these options do - so this causes a situation where not only does it deviate from Conan guidelines, it also deviates from the usage patterns of this very same library when consumed from other repositories. I can see that the equivalent recipe for this library in other repositories don't expose any options and also don't register any dependencies. From a maintainability point of view - it is quite telling that a header only library that has no more than ~30 lines of recipe in other repos, has 10 times as much in Conan Center :(

PR #22299, opened 2024-01-13
iir1: add v1.9.4

valgur:


recipes/iir1/all/conanfile.py
```diff
@@ -66,8 +66,14 @@ def generate(self):
             tc.preprocessor_definitions["IIR1_NO_EXCEPTIONS"] = "1"
         tc.generate()
 
-    def build(self):
+    def _patch_sources(self):
         apply_conandata_patches(self)
+        # Disable test and demo subdirs
```
RubenRBS, line 71: Not asking to change it back, but what I like about using patches is that if upstream changes how it handles tests/demos, the previous method will fail to apply and so we'd know they need to be disabled in a different way, but this way the file will always be written, even if they stop disbaling actual demos etc

RubenRBS (approved):

ericLemanissier (approved):

PR #22300, opened 2024-01-13
glbinding: add v3.3.0

valgur:
.


recipes/glbinding/all/conanfile.py
```diff
@@ -74,6 +74,8 @@ def _patch_sources(self):
         replace_in_file(self, compile_options, "/GL", "")
         # Don't populate rpath
         replace_in_file(self, cmakelists, "if(NOT SYSTEM_DIR_INSTALL)", "if(0)")
+        # Do not copy AUTHORS, VERSION, etc
+        replace_in_file(self, cmakelists, "install(FILES ", "# install(FILES ")
```
RubenRBS, line 78: Doesn't this seem a bit too general? It might end up removing unexpected things at some point

valgur: Dropped it as it's not even required. I initially thought that the existing `cmake-install.patch` is no longer required, but adding it fixes this issue.

RubenRBS (approved):

toge (approved):

PR #22323, opened 2024-01-14
gtsam: unvendor metis, correct option defaults, drop deprecated option

valgur:
The default values of many of the options did not match the project defaults at https://github.com/borglab/gtsam/blob/4.2.0/cmake/HandleGeneralOptions.cmake#L17-L31

That's not good, as these can cause unexpected subtle differences in behavior.

I also think the application of CMakeToolchain variables were probably been broken altogether due to CMake version being set to < 3.13 and CMP0077 not being applied.


valgur:
@RubenRBS Can you maybe somehow check why the Conan v2 build appears to be failing in C3I?

valgur:
@uilianries @RubenRBS @danimtb Can you please help out with the CI issues in the v2 pipeline? It's not clear what exactly is breaking there.

danimtb:
@valgur sorry for the late reply.

It seems so be an issue in the test_package at link time:
```
  test_package.cpp
     Creating library C:/J2/w/prod-v2/bsr/cci-c2efb684/recipes/gtsam/all/test_package/build/msvc-192-x86_64-14-release/Release/test_package.lib and object C:/J2/w/prod-v2/bsr/cci-c2efb684/recipes/gtsam/all/test_package/build/msvc-192-x86_64-14-release/Release/test_package.exp
libgtsam.lib(Ordering.obj) : error LNK2019: unresolved external symbol METIS_NodeND referenced in function "public: static class gtsam::Ordering __cdecl gtsam::Ordering::Metis(class gtsam::MetisIndex const &)" (?Metis@Ordering@gtsam@@SA?AV12@AEBVMetisIndex@2@@Z) [C:\J2\w\prod-v2\bsr\cci-c2efb684\recipes\gtsam\all\test_package\build\msvc-192-x86_64-14-release\test_package.vcxproj]
C:\J2\w\prod-v2\bsr\cci-c2efb684\recipes\gtsam\all\test_package\build\msvc-192-x86_64-14-release\Release\test_package.exe : fatal error LNK1120: 1 unresolved externals [C:\J2\w\prod-v2\bsr\cci-c2efb684\recipes\gtsam\all\test_package\build\msvc-192-x86_64-14-release\test_package.vcxproj]
```

https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/22323/13-windows-msvc/gtsam/4.0.3//summary.json

recipes/gtsam/all/conanfile.py
```diff
@@ -64,16 +64,17 @@ class GtsamConan(ConanFile):
...
-        "tangent_preintegration": False,
+        "support_nested_dissection": True,
+        "tangent_preintegration": True,
         "throw_cheirality_exception": True,
         "use_quaternions": False,
-        "with_TBB": False,
+        "with_TBB": True,
         "with_eigen_MKL": False,
         "with_eigen_MKL_OPENMP": False,
+        "with_vendored_metis": True,  # Cannot be disabled in C3I due to a non-default option value
```
danimtb, line 77: This comment is not clear to me. I think the common practice in CCI is to try to use dependencies from conan as much as possible. In this case, if possible, we should try to avoid adding an option for this and make the metis requirement conditional to `support_nested_dissection`. That way we would simplify the recipe and avoid leaving the decision of using vendored metis to the user (better user experience in the end).

Let me know what you think or point out anything I might be missing. Thanks!

valgur: It references this overridden option value for METIS:
```python
        # GTSAM expects 32-bit integers for Metis
        self.options["metis"].with_64bit_types = False
```
I kind of agree, but the METIS and nested dissection support is pretty important for this library, unfortunately. Disabling it effectively removes parallelization support. So I would leave it as is for now and instead look into improving either GTSAM or METIS to get rid of this deficiency.

valgur: Looking at the METIS recipe again, it looks like it should really be defaulting to 32-bit types in the first place to match the default behavior of the library's build system: https://github.com/KarypisLab/METIS/commit/81d70ba3d799c0143e38085b4cdd4e891e39b49a
I'll open a PR.

danimtb: @valgur thanks a lot for the explanation and quick search. Indeed, we could change the default value in the metis recipe and make the requirement of metis compulsory (unvendored) in this gtsam recipe 👏

valgur: https://github.com/conan-io/conan-center-index/pull/23194

jwillikers (approved):

danimtb (approved):

PR #22325, opened 2024-01-14
[docs] Tweak package_templates

valgur:
Adds notes for a few important but missing details, which I have encountered when creating or updating packages.
Removes a few outdated comments and improves the wording a bit of several others.

I'll add comments for specific changes as review comments.


docs/package_templates/autotools_package/all/conandata.yml
```diff
@@ -6,9 +6,7 @@ sources:
     - "https://mirror2.net/package-1.2.0.tar.gz"
     sha256: "________________________________________________________________"
   "1.1.0":
-    url:
-    - "https://mirror1.net/package-1.1.0.tar.gz"
-    - "https://mirror2.net/package-1.1.0.tar.gz"
+    url: "https://mirror2.net/package-1.1.0.tar.gz"
```
valgur, line 9: To provide an example for newcomers of the more common non-list format for the URLs.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -22,13 +22,13 @@ class PackageConan(ConanFile):
     name = "package"
     description = "short description"
     # Use short name only, conform to SPDX License List: https://spdx.org/licenses/
-    # In case not listed there, use "LicenseRef-<license-file-name>"
+    # In case not listed there, use "DocumentRef-<license-file-name>:LicenseRef-<package-name>"
```
valgur, line 25: The `DocumentRef` format has been recommended in all recent PRs where a non-standard license has been used.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -62,7 +62,6 @@ def _settings_build(self):
         return getattr(self, "settings_build", self.settings)
 
     # no exports_sources attribute, but export_sources(self) method instead
-    # this allows finer grain exportation of patches per version
```
valgur, line 65: A bit confusing and outdated, imo.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -78,11 +77,12 @@ def configure(self):
...
 
     def layout(self):
-        # src_folder must use the same source folder name the project
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        # prefer self.requires method instead of requires attribute
+        # Prefer self.requires method instead of requires attribute
+        # Set transitive_headers=True (which usually also requires transitive_libs=True) if
+        # the dependency is used in any of the packaged header files.
```
valgur, line 85: I considered adding something like "Including a comment with a link to the the public header file using the dependency is encouraged." but it seemed a bit too wordy for me.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -99,13 +99,13 @@ def validate(self):
         if self.settings.os not in ["Linux", "FreeBSD", "Macos"]:
             raise ConanInvalidConfiguration(f"{self.ref} is not supported on {self.settings.os}.")
 
-    # if another tool than the compiler or autotools is required to build the project (pkgconf, bison, flex etc)
+    # if a tool other than the compiler or autotools is required to build the project (pkgconf, bison, flex etc)
     def build_requirements(self):
         # only if we have to call autoreconf
-        self.tool_requires("libtool/x.y.z")
+        self.tool_requires("libtool/2.4.7")
```
valgur, line 105: While the versions should be always checked and updated, the `libtool` and `automake` versions change very rarely, so it does not hurt to include them for convenience.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -143,8 +141,10 @@ def generate(self):
         tc = PkgConfigDeps(self)
         tc.generate()
         # generate dependencies for autotools
-        tc = AutotoolsDeps(self)
-        tc.generate()
+        # some recipes might require a workaround for MSVC:
+        # https://github.com/conan-io/conan-center-index/blob/00ce907b910d0d772f1c73bb699971c141c423c1/recipes/xapian-core/all/conanfile.py#L106-L135
```
valgur, line 145: This is an unfortunately rather frequently required workaround. I think it's beneficial to include it as (a) the ./configure error is just a cryptic "cannot create C executables" and (b) the fix is not at all obvious.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -158,7 +158,7 @@ def generate(self):
             env.define("CC", f"{compile_wrapper} cl -nologo")
             env.define("CXX", f"{compile_wrapper} cl -nologo")
             env.define("LD", "link -nologo")
-            env.define("AR", f"{ar_wrapper} \"lib -nologo\"")
```
valgur, line 161: The `-nologo` option saves a few spammy lines from the build logs, but it has caused the build to fail with a "could not determine ar interface"  in several recipes for me, so it's safer to not include it.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -86,46 +86,51 @@ def validate(self):
...
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
-        # BUILD_SHARED_LIBS and POSITION_INDEPENDENT_CODE are automatically parsed when self.options.shared or self.options.fPIC exist
+        # BUILD_SHARED_LIBS and POSITION_INDEPENDENT_CODE are set automatically as tc.variables when self.options.shared or self.options.fPIC exist
+        # Note that tc.variables require either cmake_minimum_required() >= 3.13 or the CMP0077 policy set to NEW to work correctly.
         tc = CMakeToolchain(self)
         # Boolean values are preferred instead of "ON"/"OFF"
-        tc.variables["PACKAGE_CUSTOM_DEFINITION"] = True
+        tc.variables["PACKAGE_BUILD_TESTS"] = False
```
valgur, line 105: A lightly more relevant example.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -86,46 +86,51 @@ def validate(self):
...
-        # in case it does not work in another configuration, it should validated here too
+        # in case it does not work in another configuration, it should be validated here too
         if is_msvc(self) and self.options.shared:
             raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
 
-    # if another tool than the compiler or CMake is required to build the project (pkgconf, bison, flex etc)
+    # if a tool other than the compiler or CMake newer than 3.15 is required to build the project (pkgconf, bison, flex etc)
     def build_requirements(self):
-        self.tool_requires("tool/x.y.z")
+        self.tool_requires("cmake/[>=3.16 <4]")
```
valgur, line 95: A slightly more typical example for the CMake recipe.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -86,46 +86,51 @@ def validate(self):
...
+        # BUILD_SHARED_LIBS and POSITION_INDEPENDENT_CODE are set automatically as tc.variables when self.options.shared or self.options.fPIC exist
+        # Note that tc.variables require either cmake_minimum_required() >= 3.13 or the CMP0077 policy set to NEW to work correctly.
         tc = CMakeToolchain(self)
         # Boolean values are preferred instead of "ON"/"OFF"
-        tc.variables["PACKAGE_CUSTOM_DEFINITION"] = True
+        tc.variables["PACKAGE_BUILD_TESTS"] = False
         if is_msvc(self):
-            # don't use self.settings.compiler.runtime
             tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
-        # deps_cpp_info, deps_env_info and deps_user_info are no longer used
```
valgur, line 108: No longer relevant with Conan 2.0 being commonplace.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -86,46 +86,51 @@ def validate(self):
...
+        VirtualBuildEnv(self).generate()
 
     def _patch_sources(self):
         apply_conandata_patches(self)
         # remove bundled xxhash
-        rm(self, "whateer.*", os.path.join(self.source_folder, "lib"))
+        rm(self, "whatever.*", os.path.join(self.source_folder, "lib"))
         replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "...", "")
+        # disable subdirectories by truncating their CMakeLists.txt
+        save(self, os.path.join(self.source_folder, "tests", "CMakeLists.txt"), "")
```
valgur, line 130: `test` and `examples` directories need to be disabled quite often and this approach requires less maintenance overhead compared to patching `CMakeLists.txt`.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -139,9 +144,7 @@ def package(self):
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
         rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
         rmdir(self, os.path.join(self.package_folder, "share"))
-        rm(self, "*.la", os.path.join(self.package_folder, "lib"))
```
valgur, line 142: I have never encountered a `*.la` file in a CMake project.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -162,6 +165,18 @@ def package_info(self):
...
+        # you can copy or save a .cmake file under <prefix>/lib/cmake/ with content like
+        #     set(XYZ_VERSION ${${CMAKE_FIND_PACKAGE_NAME}_VERSION})
+        #     set(XYZ_INCLUDE_DIRS ${${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIRS})
+        #     ...
+        # and set the following fields:
+        self.cpp_info.builddirs.append(os.path.join("lib", "cmake"))
+        cmake_module = os.path.join("lib", "cmake", "conan-official-variables.cmake")
+        self.cpp_info.set_property("cmake_build_modules", [cmake_module])
+        self.cpp_info.build_modules["cmake_find_package"] = [cmake_module]
+        self.cpp_info.build_modules["cmake_find_package_multi"] = [cmake_module]
```
valgur, line 178: Matching the `*-config.cmake` output created by the project is often overlooked, imo, and it's difficult to find a good example on how to handle this as a newbie.
`${CMAKE_FIND_PACKAGE_NAME}` is preferable over a hard-coded package name as it allows the name to be overriden in CMakeDeps without breaking things.

docs/package_templates/header_only/all/conanfile.py
```diff
@@ -43,18 +42,16 @@ def _compilers_minimum_version(self):
...
         export_conandata_patches(self)
 
     def layout(self):
-        # src_folder must use the same source folder name than the project
         basic_layout(self, src_folder="src")
 
     def requirements(self):
         # Prefer self.requires method instead of requires attribute
         # Direct dependencies of header only libs are always transitive since they are included in public headers
-        self.requires("dependency/0.8.1", transitive_headers=True)
```
valgur, line 57: `transitive_headers=True` is already the default for `package_type = "header-library"`.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -78,11 +77,12 @@ def configure(self):
         self.settings.rm_safe("compiler.libcxx")
 
     def layout(self):
-        # src_folder must use the same source folder name the project
```
uilianries, line 81: please, keep this comment, it explains why do we use src_folder="src"

docs/package_templates/autotools_package/all/conanfile.py
uilianries: Keeping all the explanation in the same line is more confusing, do it in separated ways. Keep the previous comment, then, add a new dummy requirement that requires transitive flags and add a dummy comment like: "used by foo/baz.hpp:34"

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -99,13 +99,13 @@ def validate(self):
...
-    # if another tool than the compiler or autotools is required to build the project (pkgconf, bison, flex etc)
+    # if a tool other than the compiler or autotools is required to build the project (pkgconf, bison, flex etc)
     def build_requirements(self):
         # only if we have to call autoreconf
-        self.tool_requires("libtool/x.y.z")
+        self.tool_requires("libtool/2.4.7")
         # only if upstream configure.ac relies on PKG_CHECK_MODULES macro
         if not self.conf.get("tools.gnu:pkg_config", check_type=str):
-            self.tool_requires("pkgconf/x.y.z")
+            self.tool_requires("pkgconf/2.1.0")
```
uilianries, line 108: please, revert to x.y.z. Otherwise, people will not check the latest version available when using the template.

docs/package_templates/autotools_package/all/conanfile.py
```diff
@@ -114,20 +114,18 @@ def build_requirements(self):
         # for msvc support to get compile & ar-lib scripts (may be avoided if shipped in source code of the library)
         # not needed if libtool already in build requirements
         if is_msvc(self):
-            self.tool_requires("automake/x.y.z")
```
uilianries, line 117: ditto.

docs/package_templates/autotools_package/all/conanfile.py
uilianries: Please, add a related issue too to be more visible about that idiom and will be done in the future.

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -86,46 +86,51 @@ def validate(self):
...
-        # In case there are dependencies listed on requirements, CMakeDeps should be used
-        tc = CMakeDeps(self)
-        tc.generate()
-        # In case there are dependencies listed on build_requirements, VirtualBuildEnv should be used
-        tc = VirtualBuildEnv(self)
-        tc.generate(scope="build")
+
+        # In case there are dependencies listed under requirements, CMakeDeps should be used
+        deps = CMakeDeps(self)
+        # You can override the CMake package and target names if they don't match the names used in the project
```
uilianries, line 116: nice addition.

docs/package_templates/cmake_package/all/conanfile.py
uilianries: I would not include it. It encourages people doing more patches instead of trying to check it with the upstream an alternative. Remember, patches are the last resource always.

docs/package_templates/cmake_package/all/conanfile.py
uilianries: good catch

uilianries (commented):
looking good, some observations.

docs/package_templates/cmake_package/all/conanfile.py
RubenRBS: My worry with this as explained elsewhere is that in the case where a new version changes how their tests etc are set up, we won't fail as it would have happened with patches, thus making it easier to sneak tests and the like without noticing. wdyt?

docs/package_templates/cmake_package/all/conanfile.py
```diff
@@ -70,11 +69,12 @@ def configure(self):
...
     def layout(self):
-        # src_folder must use the same source folder name the project
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        # prefer self.requires method instead of requires attribute
+        # Prefer self.requires method instead of requires attribute
+        # Set transitive_headers=True (which usually also requires transitive_libs=True) if
+        # the dependency is used in any of the packaged header files.
         self.requires("dependency/0.8.1")
```
RubenRBS, line 78: Maybe add an example of a version range fake dependency, which a comment linking to the FAQ for the allowed ones?

docs/package_templates/msbuild_package/all/conanfile.py
```diff
@@ -91,19 +92,18 @@ def generate(self):
...
         # TODO: to remove once https://github.com/conan-io/conan/pull/12817 available in conan client
-        vcxproj_files = ["path/to/vcxproj_file1", "path/to/vcxproj_file2", "..."]
         platform_toolset = MSBuildToolchain(self).toolset
         import_conan_generators = ""
         for props_file in ["conantoolchain.props", "conandeps.props"]:
             props_path = os.path.join(self.generators_folder, props_file)
             if os.path.exists(props_path):
                 import_conan_generators += f"<Import Project=\"{props_path}\" />"
-        for vcxproj_file in vcxproj_files:
+        for vcxproj_file in self.source_path.rglob("*.vcxproj"):
```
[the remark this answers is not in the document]
RubenRBS: Any example where it does not work?

RubenRBS (commented):
Thanks _a lot_ for this @valgur, we were due taking a look at this. I've some comments on some of the changes, otherwise it's great, thanks!!

docs/package_templates/msbuild_package/all/conanfile.py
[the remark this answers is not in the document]
valgur: Don't remember any, really.

docs/package_templates/autotools_package/all/conanfile.py
valgur: A comment would not hurt, but the current wording is kind of broken, imo. Could you provide a better one, maybe?

docs/package_templates/autotools_package/all/conanfile.py
valgur: Ok. The Meson recipe template already uses real versions, though. Should I blank these out as well?

docs/package_templates/autotools_package/all/conanfile.py
valgur: Done.

docs/package_templates/autotools_package/all/conanfile.py
valgur: Done.

docs/package_templates/cmake_package/all/conanfile.py
valgur: Why not both - use this as a patch and submit a fix upstream as well? :grin: 

Using plain patches is better in principle, but the `add_subdirectory()` statements are typically one of the very last ones in a `CMakeLists.txt` file, which for larger projects means that this needs to be updated for almost every single release. I can remove this example here, but in general, would a `replace_in_file()` with `strict=True` be acceptable instead?

PR #22345, opened 2024-01-15
frugally-deep: add version 0.15.30

Dobiasd:


recipes/frugally-deep/all/conanfile.py
```diff
@@ -39,8 +39,8 @@ def layout(self):
 
     def requirements(self):
         self.requires("eigen/3.4.0")
-        self.requires("functionalplus/0.2.20-p0")
-        self.requires("nlohmann_json/3.11.2")
+        self.requires("functionalplus/0.2.22")
+        self.requires("nlohmann_json/3.10.5")
```
RubenRBS, line 43: Is this a downgrade? Why is that?

RubenRBS (commented):
Question about the version bump, thanks for your contribution!

recipes/frugally-deep/all/conanfile.py
RubenRBS: We usually prefer to have up to date dependencis so that the dependency graph between most libraries is consistent, yeah

Dobiasd: Ah, ok, got it. :+1:

I just changed the `nlohmann_json` version to `3.11.3` (latest) here in this MR.

Also, in the frugally-deep repo, I updated all references to this new version, so that with the next tag, it will be consistent again.

Thanks for the help!

Dobiasd: I just learned, Conan supports [version ranges](https://docs.conan.io/2/tutorial/versioning/version_ranges.html). :bulb:

So I guess something like `nlohmann_json/[>=3.10.0 <4.0]` would be even more preferred than just (superfluously) always requiring the absolutely latest version because it can avoid conflicts when installing multiple libraries, right?

RubenRBS: Yup, but we need to be careful about those in CCI. If for the library itself it  makes sense, I encourage you to use version ranges :)

But for CCI we face specific concerns that do not let us freely add ranges to every dependency. I will check if this one would make sense, thanks for bringing it up :)

RubenRBS (approved):

ericLemanissier (approved):

PR #22349, opened 2024-01-15
tqdm-cpp: new recipe

valgur:
Adds `tqdm-cpp`: https://github.com/mraggi/tqdm-cpp

Easily display progress in C++17. Inspired by python's awesome tqdm library.


valgur:
I also considered https://github.com/tqdm/tqdm.cpp as a more prominent alternative, but it fails with compilation errors when used and seems less capable. Not actively maintained either.

RubenRBS (approved):

toge (approved):

PR #22360, opened 2024-01-15
openmp/system: new recipe

valgur:
Adds a new `openmp/system` package for a much more convenient and less bug-prone OpenMP support in recipes.

This avoids the need for listing the appropriate flags in every recipe using OpenMP, e.g. https://github.com/conan-io/conan-center-index/blob/master/recipes/lightgbm/all/conanfile.py#L123-L133

It relies on the default OpenMP provided by the compiler for compilers that ship with one (GCC and MSVC) and falls back to `llvm-openmp` from CCI for others (Clang and AppleClang).

`transitive_headers=True, transitive_libs=True` can be used to set the correct visibility for libraries that use `#pragma omp` in their public headers.

See #24577 for more context.

Requires
- ~~#22353~~
- #24584


memsharded:
> Edit: Adding yet another openmp to conan is not ideal. I recommend recipes add the argument, 

Recall that it is also possible to inject compiler flags with ``tools.build:cxxflags``, so adding arbitrary flags can be done from the user side without really modifying libraries.

valgur:
> > Edit: Adding yet another openmp to conan is not ideal. I recommend recipes add the argument,
> 
> Recall that it is also possible to inject compiler flags with `tools.build:cxxflags`, so adding arbitrary flags can be done from the user side without really modifying libraries.

@memsharded I do recall that, but the propoesed `openmp/system` is mostly concerned with correct handling OpenMP as a part of the dependency graph rather than a mere optimization flag. Also, I hope you're not seriously suggesting force-enabling OpenMP for all recipes that might support it (https://github.com/conan-io/conan-center-index/issues/24577#issuecomment-2241199222)?

jacobfriedman:
I would explicitly define requirements in packages, rather than wrapping them, which is a workaround. I wouldn't expect force-enabling openmp, rather, it would be an option- for all recipes that might support it.

PR #22364, opened 2024-01-15
moltenvk: add v1.2.7, drop old versions

valgur:
.


uilianries:
I just started an internal build to generate the missing glslang/1.3.239.0 package.

PR #22367, opened 2024-01-15
vk-bootstrap: add v1.3.268

valgur:


recipes/vk-bootstrap/all/conanfile.py
```diff
@@ -57,21 +68,24 @@ def configure(self):
     def layout(self):
         cmake_layout(self, src_folder="src")
 
+    @property
+    def _headers_version(self):
+        return {}.get(self.version, f"{self.version}.0")
```
RubenRBS, line 73: Isn't the mapping missing here?

RubenRBS (commented):
Thanks!

recipes/vk-bootstrap/all/conanfile.py
danimtb: I think so!

perseoGI: Ping @valgur

PR #22371, opened 2024-01-15
oboe: add v1.8.0, allow Conan to set cppstd

valgur:


RubenRBS (approved):

RubenRBS (approved):

ericLemanissier (approved):

recipes/oboe/all/patches/1.7.0-0001-fix-cmake.patch
```diff
@@ -12,7 +12,7 @@
          -Ofast
 -        "$<$<CONFIG:DEBUG>:-Werror>")
 +)
-+target_compile_features(oboe PUBLIC cxx_std_17)
```
SpaceIm, line 15: Why is it removed?

This workaround in conanfile is worse:

```
        if not valid_min_cppstd(self, self._min_cppstd):
```

recipes/oboe/all/conanfile.py
```diff
@@ -40,13 +44,16 @@ def validate(self):
...
+            check_min_cppstd(self, self._min_cppstd)
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
         tc = CMakeToolchain(self)
+        if not valid_min_cppstd(self, self._min_cppstd):
+            tc.variables["CMAKE_CXX_STANDARD"] = self._min_cppstd
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
SpaceIm, line 56: Why do you need this? I don't see any reason looking at upstream CMakeLists

valgur: Just to be safe, due to `cmake_minimum_required(VERSION 3.4.1)`. I don't like checking whether the values are actually overwritten as "CACHE" or not in CMakeLists.txt, as it's prone to change at any time.

recipes/oboe/all/patches/1.7.0-0001-fix-cmake.patch
SpaceIm: This workaround:
- hardcodes C++17 when compiler.cppstd is not set in profile while target_compile_features() is smarter and can fallback to something closer to default compiler version as soon as it honors min C++ version.
- makes conanfile more complex than it needs to be

Moreover the way to go in CCI recipes is to patch and submit fixes upstream to avoid workarounds in conanfile (I have to admit I didn't have the time to submit this one).

PR #22386, opened 2024-01-16
gegles-spdlog_setup: bump to `spdlog/1.13.0`

gegles:
Specify library name and version:  **gegles-spdlog_setup/1.1.0**

---

### DIFF
--- a/recipes/gegles-spdlog_setup/all/conanfile.py
+++ b/recipes/gegles-spdlog_setup/all/conanfile.py
@@ -39,7 +39,7 @@ class SpdlogSetupConan(ConanFile):
 
     def requirements(self):
         self.requires("cpptoml/0.1.1")
-        self.requires("spdlog/1.12.0")
+        self.requires("spdlog/1.13.0")
         self.requires("fmt/10.2.0")
 
     def package_id(self):


josefgraus:
This broke the package as we were using, as now the spdlog version conflicts with another package. There didn't appear to be a version bump of the conan package to avoid backwards-breaking changes?

gegles:
> This broke the package as we were using it, as now the spdlog version conflicts with another package. There didn't appear to be a version bump of the conan package to avoid backwards-breaking changes?

Hello and sorry if this broke your setup. I am still trying to figure out how best to deal with these various version bumps and conflicts... but in general, I think trying to keep everything as up to date as possible is a better way to go.

Are you not able to simply use `override=True` (or explicitly depend on `spdlog/1.12.0` in your package to solve the problem? 
That's what I do in such circumstances..

josefgraus:
Yup, we can fix it on our end. I suppose what's alarming for us is that a fixed version is getting changes. One thing we somewhat depend on is that if we version-lock a requirement, we should be able to rebuild our software without issue in the future. 

I'm not sure what the policy or process is for publishing changes to conan packages. I think I assumed that they would be version bumped if the recipe changed in a way that would break projects that consumed it. Is this something I just made up and hoped was true? haha

gegles:
I hear you, but, I actually think it's better to see the package version change to be more tied to code changes as opposed recipe changes, that can happen independently...
That's why indeed, the package locking mechanism can be key to control that aspect... but I, myself am still figuring how best to use that. If you want more control, you can also just include the revision number in the `self.requires(..)` call... to lock to a specific recipe/package version as well as the normal version...

josefgraus:
I understand, yeah, it's kind of a weird in-between versioning problem. I take it the gegles-spdlog_setup didn't need any changes to compile against the newer version of spdlog? 

Thanks for the tip. We'll go ahead and start adding revisions to our dependencies.

PR #22414, opened 2024-01-17
cairo: Fix cross-compilation when using Meson

jwillikers:
Set `ipc_rmid_deferred_release` property when cross building. Setting this according to runtime behavior is impossible when cross-compiling. Refer to https://gitlab.freedesktop.org/cairo/cairo/-/issues/408. To fix this, set the property `ipc_rmid_deferred_release`. According to the discussion in the issue, this should at least be enabled on Linux. Since other systems may behave differently, this is disabled otherwise.

---

### DIFF
--- a/recipes/cairo/meson/conanfile.py
+++ b/recipes/cairo/meson/conanfile.py
@@ -4,6 +4,7 @@ import os
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import fix_apple_shared_install_name, is_apple_os
+from conan.tools.build import cross_building
 from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rename, replace_in_file, rm, rmdir
 from conan.tools.gnu import PkgConfigDeps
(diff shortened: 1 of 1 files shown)


franramirez688 (approved):

RubenRBS (changes requested):
@jwillikers can I get compilation logs from before/after the change? They can help in the future as maintenance aid to someone, and making sure that the error/fix was proper :)

jwillikers:
> @jwillikers can I get compilation logs from before/after the change before approving? They can help in the future as maintenance aid to someone, and making sure that the error/fix was proper :)

@RubenRBS I've attached those logs below.
The logs are generated using the following command, which dodges cross-building errors in libselinux and newer versions of libmount.

```
conan create . "1.18.0@" --build cascade --build outdated --options:host '*/*:with_selinux=False' --profile:host product/debug --require-override libmount/2.36.2 --update
```


[cairo-cross-failure-log.txt](https://github.com/conan-io/conan-center-index/files/13989911/cairo-cross-failure-log.txt)
[cairo-cross-success-log.txt](https://github.com/conan-io/conan-center-index/files/13989910/cairo-cross-success-log.txt)

RubenRBS (approved):

PR #22432, opened 2024-01-19
Add with_asm and with_opencl options in the x264 recipe

anthonyliot:
Specify library name and version:  **libx264/cci.20220602**

Add in the two options with_asm and with_opencl. When building on Android this two flags can be required.

---


recipes/libx264/all/conanfile.py
```diff
@@ -25,11 +25,15 @@ class LibX264Conan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "bit_depth": [8, 10, "all"],
+        "with_opencl": [True, False],
+        "with_asm": [True, False]
```
valgur, line 29: I would use `enable_opencl` and `enable_asm`, personally, since `with_*` is typically used for options controlling external dependencies on CCI. Matches the project option names better as well.

recipes/libx264/all/conanfile.py
valgur, line 1: Because OpenCL is enabled for bit_depth == 8, you should add a validation check for `self.options.with_opencl and self.options.bit_depth != 10`: https://github.com/mirror/x264/blob/eaa68fad9e5d201d42fde51665f2d137ae96baf0/configure#L1409

valgur (approved):
LGTM, just two minor suggestions.

It's worth noting that the OpenCL feature does not require OpenCL libraries to be present. These are only loaded at runtime, if available.

recipes/libx264/all/conanfile.py
memsharded: Option ``"with_opencl": [True, False],`` already exists in other 6 recipes in ConanCenter, and ``enable_opencl`` only in 1, so better use ``with_opencl``

memsharded:
PR is red due to a bug in the pipeline, but I have already sent a fix for it. I'll push this again when it is merged to get this PR to be merged. Thanks very much for your contribution, and sorry for the long delays to process this!

jcar87 (approved):

memsharded (approved):

PR #22516, opened 2024-01-23
Bump timsort to 3.0.0

Morwenn:
Specify library name and version:  **timsort/3.0.0**

Version 3.0.0 requires C++20, so the automatic update was not enough to automatically bump the library version.

---


toge:
@Morwenn 
To solve CI errors, this changes can help.
https://github.com/conan-io/conan-center-index/pull/22517/files

these changes might help to solve CI errors.

- validate compiler version
- specify c++ version on CMakeLists.txt of test_package

Morwenn:
Thanks, I'm going to try those later tonight.

Morwenn:
I think it doesn't work because `<ranges>` is only truly available in libc++ 15, it's experimental before  - so it could work with Clang < 15 but only if it uses libstdc++. I supposed that it failed there because AppleClang defaults to libc++. Is there anyway I can detect that to fix the recipe?

toge:
@Morwenn 
How about incrementing apple-clang version:
```
                "apple-clang": "14",
```

In [cpp reference](https://en.cppreference.com/w/cpp/compiler_support/20), apple-clang seems to support ranges since 14.

I confirmed timsort/3.0.0 works well on apple-clang 15.

Morwenn:
Hum, I forgot that libstdc++ was considered deprecated with AppleClang, so you're right, there's probably no need to try to be smart about it.

recipes/timsort/all/test_package/CMakeLists.txt
```diff
@@ -5,4 +5,11 @@ find_package(gfx-timsort REQUIRED CONFIG)
...
 add_executable(${PROJECT_NAME} test_package.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE gfx::timsort)
-target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
+
+if(gfx-timsort_VERSION VERSION_LESS "2.0.0")
+  target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_98)
+elseif(gfx-timsort_VERSION VERSION_LESS "3.0.0")
+  target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
+else()
+  target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_20)
```
toge, line 14: Since cxx_std_20 requires 3.12, would you update `cmake_minimum_required`?

https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html

Morwenn: Thinking about it again maybe these lines aren't be useful: shouldn't the conan-generated targets assign the correct compile features for the standard version?

I introduced them on your suggestion to fix the previous problem, but maybe constraining the compiler versions was enough.

uilianries: For Conan 2.x is all solved, the CMakeDeps, etc. will require 3.15. Plus, cppstd is defined by default there, so passing target_compile_features should no longer needed. I guess it's okay keeping the current, otherwise it would fail to build with Conan 2.x in case running an older version of cmake

Morwenn: By "the current", do you mean as it was before this MR, or the `target_compile_features` currently in the MR?

Morwenn: I get errors when removing `target_compile_features` apparently, guess I incorrectly interpreted your message.

Morwenn: Bump because it's been half a year and I'm still unsure what you meant.

recipes/timsort/all/test_package/CMakeLists.txt
```diff
@@ -5,4 +5,3 @@ find_package(gfx-timsort REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE gfx::timsort)
```
uilianries, line 7:
```suggestion
target_link_libraries(${PROJECT_NAME} PRIVATE gfx::timsort)
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_20)
```

The C++ standard is not passed when building with Conan 1.x, because compiler.cppstd is None by default. In Conan 2.x, Conan will pass the standard to CMake via CMAKE_CXX_STANDARD definition.

uilianries (commented):
Sorry the huge delay! I would suggest only adding c++20 in test package. For Conan 2.x it will be managed by cppstd, so it should be safe. 

Please, update the `test_package/CMakeLists.txt` to require version 3.15 (minimum require by Conan 2.x) and will cover cxx_std_20

recipes/timsort/all/test_package/CMakeLists.txt
Morwenn: No problem for the delay, I know that you're quite busy handling a bunch of different recipes 🙂 
Is it sufficient to ensure that all versions work correctly? Only the 3.0.0 requires C++20, older versions are supposed to be compatible with older standards. Or is it just best effort?

uilianries: Yes. Like I commented before, Conan 2.x will the C++ standard based on compiler.cppstd.

perseoGI (approved):

perseoGI (approved):

AbrilRBS (approved):
Hi @Morwenn thanks a lot for your PR, and thanks for your patience while we got back to this PR, it got lost among other PRs - We've taken steps for things like this not to happen again, sorry that this was left unreviewed for so long! :)

Morwenn:
Thanks a lot for taking care of the changes, I really wasn't sure how to get it to work 🙏

PR #22525, opened 2024-01-24
Updated Opene57 to 1.6.5

madduci:
Specify library name and version:  **opene57/1.6.5**

A new release of opene57 is available (1.6.5), therefore the recipe has been updated.

---

### DIFF
--- a/recipes/opene57/all/conandata.yml
+++ b/recipes/opene57/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.6.5":
+    url: "https://github.com/openE57/openE57/archive/1.6.5.tar.gz"
+    sha256: "23f3c159550a70b6a1c2330f0f03ed816cbc27ff5656410bfee527bae31eb186"
   "1.6.4":
     url: "https://github.com/openE57/openE57/archive/1.6.4.tar.gz"
     sha256: "97accc32ae294113a97b8d3f0ecf0d608057a6f0fcdfee17db0c5db4ab28ce69"
--- a/recipes/opene57/all/conandata.yml
+++ b/recipes/opene57/all/conandata.yml
@@ -8,6 +8,3 @@ sources:
   "1.6.3":
     url: "https://github.com/openE57/openE57/archive/refs/tags/1.6.3.tar.gz"
     sha256: "e335afdda98a2707d05ec4bce99f362a6f7f0eb2e8bbf2d7346eae292046f827"
-  "1.6.2":
-    url: "https://github.com/openE57/openE57/archive/refs/tags/1.6.2.tar.gz"
-    sha256: "2d487e663cf43105b19653d34250fd45f947af839e327336b6878464a99d5423"
--- a/recipes/opene57/all/conanfile.py
+++ b/recipes/opene57/all/conanfile.py
@@ -66,7 +66,7 @@ class Opene57Conan(ConanFile):
 
     def requirements(self):
         if self.options.with_tools:
-            self.requires("boost/1.83.0")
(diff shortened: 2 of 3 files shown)


recipes/opene57/all/conandata.yml
uilianries, line 1: I would suggest by removing 1.6.2 from config.yml and all/conandata.yml, because doxygen support is available since 1.6.3 only.

recipes/opene57/all/conanfile.py
```diff
@@ -8,7 +8,7 @@
 from conan.tools.microsoft import is_msvc, is_msvc_static_runtime
 from conan.tools.scm import Version
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=2.0.5"
```
uilianries, line 11:
```suggestion
required_conan_version = ">=1.54.0"
```

Please, keep Conan 1.x compatibility. All recipes in CCI are still compatible.

recipes/opene57/all/conanfile.py
uilianries, line 1: In `generate()` method, is missing `tc.variables["BUILD_DOCS"] = self.options.with_docs`. I can't suggest directly in the code because GH limits which lines accept suggestions.

recipes/opene57/all/conanfile.py
```diff
@@ -39,10 +41,9 @@ def _min_cppstd(self):
     @property
     def _minimum_compilers_version(self):
         return {
-            "Visual Studio": "15",
```
uilianries, line 42: Please, keep the previous configuration. The `Visual Studio` is still used in Conan 1.x. 

I don't get about gcc 7 to 8. Does 1.6.5 require GCC 8?

uilianries (commented):
@madduci Thank you for updating this package. I have some suggestions in the code.

Plus, please, sign the CLA pointed in this PR. It was update some time ago and every user need to sign it again.

recipes/opene57/all/conanfile.py
madduci: It's not an hard requirement, but using gcc 8 as minimum version is suggested. It can be still kept as gcc-7.

I will submit the change with Visual Studio again. I thought that in conan2 it is now "msvc", that's why I removed it

recipes/opene57/all/conandata.yml
madduci: Totally fine for me, the 1.6.2 is pretty old too.

recipes/opene57/all/conanfile.py
madduci: I will fix this, thanks for reviewing

madduci:
@uilianries  i've updated the PR with your suggestions, plus I've signed the CLA again. Could you please have a look at it again? I hope it is fine now

recipes/opene57/all/conanfile.py
```diff
@@ -65,8 +67,13 @@ def layout(self):
     def requirements(self):
         if self.options.with_tools:
             self.requires("boost/1.83.0")
```
uilianries, line 69:
```suggestion
            self.requires("boost/1.84.0")
```

We have Boost 1.84.0 available too https://conan.io/center/recipes/boost?version=1.84.0

uilianries (commented):
Looking good. Just a comment about updating Boost

uilianries (approved):
LGTM. Just  wait the CI now and an extra review 😄

ericLemanissier (approved):

uilianries (approved):
LGTM

ericLemanissier (approved):

PR #22527, opened 2024-01-24
Add protobuf 3.15.8

apogza:
Specify library name and version:  **protobuf/3.15.8**

It is not always easy to migrate to a newer protobuf library. Therefore, we would like to have it in conan center so that it is easily accessible as a dependency.

---


jcar87:
My understanding is that protobuf (at least the 3.xx series) has remained backwards compatible at the source level. Unfortunately we are not considering maintaining older versions, especially considering that new versions tend to work correctly for old code as well. From what I can see, every other recipe in Conan Center that depends on protobuf is able to use newer versions.

If this is not the case, I would advise to open an issue expressing concrete issues so that we can evaluate how many users this could affected.

PR #22539, opened 2024-01-25
abseil: fix build in environments without CMake

Artalus:
Specify library name and version:  **abseil/X**

`abseil` requires CMake to be built, but does it without respect 🙂 

---

### DIFF
--- a/recipes/abseil/all/conanfile.py
+++ b/recipes/abseil/all/conanfile.py
@@ -50,7 +50,7 @@ class AbseilConan(ConanFile):
         }.get(self._min_cppstd, {})
 
     def build_requirements(self):
-        self.tool_requires("cmake/3.27.9")
+        self.tool_requires("cmake/[>=3.16 <4]")
 
     def export_sources(self):
         copy(self, "abi_trick/*", self.recipe_folder, self.export_sources_folder)
--- a/recipes/abseil/all/conanfile.py
+++ b/recipes/abseil/all/conanfile.py
@@ -49,6 +49,9 @@ class AbseilConan(ConanFile):
             },
         }.get(self._min_cppstd, {})
 
+    def build_requirements(self):
+        self.tool_requires("cmake/3.27.9")
+
     def export_sources(self):
         copy(self, "abi_trick/*", self.recipe_folder, self.export_sources_folder)
         export_conandata_patches(self)


recipes/abseil/all/conanfile.py
```diff
@@ -49,6 +49,9 @@ def _compilers_minimum_version(self):
             },
         }.get(self._min_cppstd, {})
 
+    def build_requirements(self):
+        self.tool_requires("cmake/3.27.9")
```
franramirez688, line 53: Let's use version ranges for CMake tool require:

```suggestion
        self.tool_requires("cmake/[>=3.16 <4]")
```

valgur:
Abseil requires CMake v3.10 (https://github.com/abseil/abseil-cpp/blob/20240116.1/CMakeLists.txt#L19). We generally don't add CMake as a tool_requires unless v3.16 or newer is needed.

uilianries (changes requested):
Hello @Artalus! Thank you for submitting your PR, but having CMake installed is something expected already by CCI: https://github.com/conan-io/conan-center-index/blob/master/docs/developing_recipes_locally.md#setup-your-environment

In case we follow your idea, then any recipe in CCI using CMake should have it listed as tool requirement too ... 

The CI service in CCI has CMake 3.15 installed due Conan 2.x design: https://github.com/conan-io/tribe/blob/a36de8dd33852f43091e9f5eb5a7ad83618089f2/design/004-tools-cmake.md?plain=1#L2

For those projects that need a new version, then we list CMake as tool requirement to install a newer version. This is not the case for Abseil, it uses CMake 3.10: https://github.com/abseil/abseil-cpp/blob/master/CMakeLists.txt#L19

If don't have CMake installed in your environment and want to use Conan to install it, I would recommend configuring it directly in your profile instead: https://docs.conan.io/2/reference/config_files/profiles.html#tool-requires

Artalus:
Cheers for such a detailed explanation! Being able to declare `tool-require` in the profile might be of service later.

I might argue that if a package depends on buildtool X that is easily available from Conan, it should declare one as such - esp. since CMake now climbed upon a track of slowly shedding off some of its legacy components. But this would be a large change in CCI policies outside the scope of  this PR :) I guess we can safely close it then 👍

PR #22544, opened 2024-01-25
highfive: add v2.9.0 and bump boost

robomics:
Specify library name and version:  **highfive/all**

---

### DIFF
--- a/recipes/highfive/all/conandata.yml
+++ b/recipes/highfive/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "2.9.0":
+    url: "https://github.com/BlueBrain/HighFive/archive/refs/tags/v2.9.0.tar.gz"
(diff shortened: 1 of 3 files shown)


RubenRBS (dismissed):
Thanks! We found some issues with the recipe itself that we wanted to address, approving this one before we push a PR for the fixes, thanks a lot :)

uilianries:
Hello @robomics !! We did some important changes in your PR. When checking the changes between 2.8 and 2.9, we found bugs in this recipes which are now fixed.

- Deprecated the `with_static_hdf5` as the HDF5 dependency is managed by Conan. The H5 project has no usage over HDF5, but it's something to be consumed with the project.
- As a header-only project, we need to export compiler definitions to include or not external dependencies. It was a bug, now fixed.
- The dependencies options in upstream's CMakeLists.txt are cache, so we need to use `cached_variables` as well.

RubenRBS (approved):

robomics:
Awesome, thanks to both!

toge (approved):

Issue #22485, opened 2024-01-22
[package] xorg/system: Missing xvmc dependency when using prebuilt binaries


Ahajha:
### Description

Since https://github.com/conan-io/conan-center-index/pull/22081 (which removed xvmc) was merged, there have been periodic build failures in a few PRs, for example https://github.com/conan-io/conan-center-index/pull/21387. Even building a direct dependency of `xorg` can fail, for example here: https://github.com/conan-io/conan-center-index/pull/19897

I believe the issue that the package pulls the latest xorg package revision, but because the package ID for xorg is cleared, the latest package (which does contain the `xvmc` requirement) is pulled. When the main package is built, the dependency thinks it needs xvmc, but the main package pulls the latest recipe, which does not pull xvmc. So I think one or both of the following need to happen:

    Encode which packages are part of xorg as part of the package ID.
    Adjust the embed mode of xorg to always rebuild if the package ID changes.

A fix for this potentially be applied to other `system` packages.

### Package and Environment Details

* Package Name/Version: **xorg/system**
* Operating System+version: **Linux Ubuntu 22.04**
* Compiler+version: **GCC 11**
* Conan version: **conan 2.0.17** (haven't reproduced with Conan 1.x yet, unsure if it's a problem)
* Python version: **Python 3.10.12**

### Conan profile

[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=11
compiler.libcxx=libstdc++11
build_type=Release

[conf]
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True
tools.build:compiler_executables={"c": "gcc-11", "cpp": "g++-11"}

### Steps to reproduce

Important: Make sure your system does not contain `xvmc` from the system package manager. It should be safe to remove (see discussion on the PR that removed it from `xorg/system`, but I don't take responsibility if it breaks things.

I've removed all `xorg` and `tk` packages from my system:
```
conan remove tk -c
conan remove xorg -c
```
I can then reproduce the failure by navigating to `recipes/tk/all` and running
```
conan test test_package tk/8.6.10 -o */*:shared=True
```

I was not able to reproduce the problem with different approaches
- A `conan create` instead of a `conan test` (yet, it might be possible)
- Non-shared packages (this I suspect to be part of the problem, as shared libraries are not 'embedded' in Conan 2.0)
- Conan 1.x (yet, it might be possible)

(This is not a `tk` specific problem, just one that has been affected by this)

### Logs

[...]

PR #22549, opened 2024-01-26
[tk] Use xorg components, fix MSVC shared build

Ahajha:
Specify library name and version:  **tk/all**

See https://github.com/conan-io/conan-center-index/issues/22485

C3I runners use a GCC built from source, which does not enable `--as-needed` by default. This adds that to this recipe, and for good measure while we're here add the specific xorg components that are used.

I've fixed the Windows/all shared build as well, which has affected a few previous PRs. What was happening is that there was a line like
```
type << > some_other_file
```
The `<<` expands to a temp file, which for some reason had a `\` which caused the `type` command to fail. So instead, I have it write directly to the final file, now it prints out the contents of the file but this is otherwise harmless.

---


Ahajha:
@jcar87 FYI

RubenRBS (approved):

ericLemanissier (approved):

PR #22554, opened 2024-01-26
[boost] Add default support to C++11

### LINKED ISSUE #22522: Boost 1.84.0 does not generate binaries for apple-clang with Conan 1.x
For now, no packages with Boost 1.84.0 with MacOS and GCC5 are being generated with Conan 1.x. The missing cppstd in the profile is causing it. GCC5 supports C++11 and uses it by default (https://gcc.gnu.org/gcc-5/changes.html), second, Apple Clang supports C++11 and later, but it uses 98 by default. We will need to adjust it, to use C++11 when not defined cppstd.

_Originally posted by @uilianries in https://github.com/conan-io/conan-center-index/issues/21747#issuecomment-1907847429_

Check the build log https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/21747/15/boost/1.84.0//summary.json

uilianries:
Since Boost 1.84.0, basic boost libraries like atomic, filesystem, log, random, thread and test require C++11 (see https://www.boost.org/users/history/version_1_84_0.html), making impossible to build without a `settings.compiler.cppstd` defined when using apple-clang < 14.

The apple-clang compiler uses C++98 as its standard until version 14, so without  `settings.compiler.cppstd`, this recipe would use the default c++ standard supported by the compiler instead.

However, apple-clang supports C++11 since many years ago. Plus, in Conan 1.x is not defined a cppstd by default, even in ConanCenterIndex, it would make impossible to generate new package into Conan Center.

This PR injects `-std=c++11` to cxxflags in case there is NO `settings.compiler.cppstd` defined AND the compiler supports it (see [supported_cppstd](https://docs.conan.io/1/reference/conanfile/tools/build.html#conan-tools-build-supported-cppstd))

fixes #22522

/cc @patmantru 

--- 


uilianries:
Hola @grafikrobot ! Please, take a look in this change if possible. It changes how we pass the `-std=` flag to `b2` in case there is no settings.compiler.cppstd (Conan 1.x). Saludos!

recipes/boost/all/conanfile.py
```diff
@@ -1098,6 +1105,9 @@ def add_defines(library):
         if self.settings.get_safe("compiler.cppstd"):
             cppstd_flag = AutotoolsToolchain(self).cppstd
             flags.append(f"cxxflags={cppstd_flag}")
+        elif self._has_cppstd_11_supported:
+            prefix = "/" if is_msvc(self) else "-"
+            flags.append(f"cxxflags={prefix}std=c++11")
```
grafikrobot, line 1110: You should be using the `cxxstd` feature instead of adding compiler specific flags. Add a `cxxstd=11` argument to the invocation.

grafikrobot (changes requested):

recipes/boost/all/conanfile.py
grafikrobot: PS. If it's not already the case.. The recipe should pass in `cxxstd=XY` for all the other `compiler.cppstd` values.

uilianries: Thank you! I'll update it.

jwillikers (approved):

grafikrobot (approved):

PR #22560, opened 2024-01-26
mgclient: add recipe

toge:
Specify library name and version:  **mgclient/1.4.2**

---


recipes/mgclient/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_cpp": True,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 17
```
RubenRBS, line 34: https://github.com/memgraph/mgclient/blob/master/CMakeLists.txt#L81
Seems like this would not be needed?

Not too sure how this works with the cpp bindings though

RubenRBS (commented):
Thanks! Question about the min cppstd for the bindings

recipes/mgclient/all/conanfile.py
toge: @RubenRBS 
mgclient requires `optional` which is supported since C++17.
https://github.com/memgraph/mgclient/blob/master/mgclient_cpp/include/mgclient.hpp#L108

RubenRBS (approved):
Thanks!

jwillikers (approved):

PR #22563, opened 2024-01-26
onnxruntime: More compatibility patches for 1.16

Artalus:
Specify library name and version:  **onnxruntime/1.16.X**

We have a custom `onnxruntime` recipe with CUDA support for internal use, and while adapting upstream changes to support 1.16 I noticed some weirdness in the patches.
- Comparing `1.16.0-0001-cmake-dependencies.patch` to the predecessor `1.15.1-0001-cmake-dependencies.patch`, it lost the hunk about Abseil. Because of this our CMake configuration logs were lacking a bunch of `Conan: Component target declared 'absl::something'` and we were getting weird `target_link_libraries: target absl::something not found`. Not sure why it doesn't happen in C3I, maybe because we use older version of Abseil.
- In https://github.com/microsoft/onnxruntime/pull/17173 MS started to use `<stacktrace>` header from C++23; then in https://github.com/microsoft/onnxruntime/pull/17209 they made it optional - but left a bug for VS2019 and below. The bug got a fix in https://github.com/microsoft/onnxruntime/pull/17716 but 1.16.0-3 were already released. I backported the fix to enable VS2019 builds in 1.16.

---

### DIFF
--- a/recipes/onnxruntime/all/conandata.yml
+++ b/recipes/onnxruntime/all/conandata.yml
@@ -25,6 +25,10 @@ patches:
     - patch_file: "patches/1.14.1-0004-abseil-no-string-view.patch"
       patch_description: "allow to build with abseil built without c++17 support"
       patch_type: "portability"
+    - patch_file: "patches/1.16.0-0002-vs2019.patch"
+      patch_description: "Do not include C++23 header if it is not present"
+      patch_type: "portability"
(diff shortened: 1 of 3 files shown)


czoido:
I think this should be closed by https://github.com/conan-io/conan-center-index/pull/22557

PR #22574, opened 2024-01-28
reflect-cpp: add version 0.10.0

toge:
Specify library name and version:  **reflect-cpp/0.10.0**

reflect-cpp/0.7.0 supports bson and cbor, but this PR does not mention the two features.
Since reflect-cpp is a header-library and these features do not require extra definitions or configuration files, I do not see the need to add two options for them.

> For BSON support, you must also include the header <rfl/bson.hpp> and link to the libbson library (https://github.com/mongodb/libbson).
https://github.com/getml/reflect-cpp/blob/main/docs/bson.md

---


RubenRBS:
>  I do not see the need to add two options for them.

@toge wouldn't we need to at the very least add the requirements to the recipe?

RazielXYZ:
Anyone who would want to use them could just require them to their own project that uses reflect-cpp and wants those backends. Core reflect-cpp won't have issues compiling without the serialization libraries as backends unless you try to include its specific header for that format/backend.

That said, I'm not sure what the best practice is for something like this. When I made the initial recipe I thought having them in would be helpful, but in that case we should probably keep being consistent with having them.

RazielXYZ:
@RubenRBS any thoughts on this?

RubenRBS:
Hi @RazielXYZ didn't realize you answered back, sorry for the delay, will have something for you tomorrow once I get back to reviewing PRs :)

uilianries:
We have a similar recipe CCI with optional backend, so users need to include them as extra dependency. I would follow the same approach here, not installing these dependencies. Reading the headers and the cmake file, there is no mechanism like a ifdef or linkage that could break the project.

RazielXYZ:
> We have a similar recipe CCI with optional backend, so users need to include them as extra dependency. I would follow the same approach here, not installing these dependencies. Reading the headers and the cmake file, there is no mechanism like a ifdef or linkage that could break the project.

So the suggestion is to remove the with_* options and requires from the recipe? Sounds fine to me

uilianries:
> So the suggestion is to remove the with_* options and requires from the recipe? Sounds fine to me

Not exactly, let's keep with the current options to avoid breaking users.

RazielXYZ (approved):

lkotsonis:
Should there be a separate PR for v0.8.0 that got out today?

valgur (approved):

uilianries (approved):
LGTM. We can take an action over the options in another PR.

uilianries:
> Should there be a separate PR for v0.8.0 that got out today?

@lkotsonis Yes, please. Could you open a new PR? Otherwise we would need to re-build this PR, but is actually good to be merged.

lkotsonis:
> oh. In my local environment(17.7.2+d6990bcfa), it works fine. I try to find workaround.

Any update on this?

toge:
@lkotsonis 
Sorry it took so long.

The reason why 0.8.0 does not pass CI is because 0.8.0 has a compile error in certain versions of msvc.
The latest msvc does not have this problem, so I am going to prepare an older msvc environment used by CI to investigate the cause.
I have not been able to take the time to do this. Please wait.

lkotsonis:
> @lkotsonis 
> Sorry it took so long.
> 
> The reason why 0.8.0 does not pass CI is because 0.8.0 has a compile error in certain versions of msvc.
> The latest msvc does not have this problem, so I am going to prepare an older msvc environment used by CI to investigate the cause.
> I have not been able to take the time to do this. Please wait.

Just to be clear, I wasn't complaining. Thank you for taking care of this and for the useful information!

toge:
@RubenRBS 
Colud you review it again?

If you think the options for dependand packages are mandatory(bson, toml, messagepack etc), I will add these options.

franramirez688 (changes requested):
@toge Thanks a lot for the PR, it looks pretty well! Only one thing before moving this forward. We have to report that patch upstream and wait for the official solution to this bug. Otherwise, we could report it and add an extra step at validation time for that temporary invalid configuration while waiting for the official patch.

toge:
@franramirez688 
Thank you for your review.
I create a PR for this.
Let's wait for reply.

toge:
@franramirez688 
My patch has been merged.

franramirez688 (dismissed):
@toge Great job! Thanks a lot for taking the time to report the patch upstream 👏 That's really appreciated.

RubenRBS (dismissed):
Thanks for creating the upstream PR, it is supper appreciated! :)

toge:
@RubenRBS 
Colud you review it again?
I forgot to remove dangling patch.

RubenRBS (approved):

uilianries (approved):
LGTM https://github.com/getml/reflect-cpp/compare/v0.6.0...v0.10.0.patch

czoido (approved):

RubenRBS (approved):
Sorry for the delay! Looks great now

franramirez688 (approved):

Issue #22532, opened 2024-01-25
[package] sdl_image/2.6.3: failed to build with sdl/*:shared=True


StellaSmith:
### Description

As per the title, SDL_Image fails to build when using a SDL compiled as a shared library.
The logs seems to suggest that SDL_Image is trying to open SDL's `SDL_version.h` but wrongly uses the include directory list?

### Package and Environment Details

* Package Name/Version: **sdl_image/2.6.3**
* Operating System+version: **Fedora Linux 39**
* Compiler+version: **gcc (GCC) 13.2.1 20231205 (Red Hat 13.2.1-6)**
* Docker image: **N/A**
* Conan version: **conan 2.0.8**
* Python version: **Python 3.12.1**

### Conan profile

Profile host:
```ini
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=13.2
os=Linux
[buildenv]
CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
CXXFLAGS=-flto
CCFLAGS=-flto
LDFLAGS=-Wl,--as-needed -Wl,-O3
```

Profile build:
```ini
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=13.2
os=Linux
[buildenv]
CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
CXXFLAGS=-flto
CCFLAGS=-flto
LDFLAGS=-Wl,--as-needed -Wl,-O3
```

### Steps to reproduce

```sh
conan install --requires "sdl_image/2.6.3" --requires "sdl/[~2]" -o "sdl/*:shared=True" -b missing
```

### Logs

<details><summary>Click to expand log</summary>

```

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=13.2
os=Linux
[options]
sdl/*:shared=True
[buildenv]
CMAKE_INTERPROCEDURAL_OPTIMIZATION=ON
CXXFLAGS=-flto
CCFLAGS=-flto
LDFLAGS=-Wl,--as-needed -Wl,-O3

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
[...]

PR #22575, opened 2024-01-28
[sdl_image] Fix static sdl_image with shared sdl

Ahajha:
Specify library name and version:  **sdl_image/2.6.3**

Resolves https://github.com/conan-io/conan-center-index/issues/22532

This only affects 2.6.3, neither 2.0.5 nor the upcoming 2.8.x is affected.

The issue was that sdl_image's CMake would try to read from `${SDL2_INCLUDE_DIR}/SDL_version.h`, where the variable was a list, so it failed. The only thing this ultimately was used for was to get the version as a string, which Conan already provides as a variable directly.

This seems to also be bitten by https://github.com/conan-io/conan-center-index/issues/22485. 
Xorg is used by pulseaudio and xcbcommon, which both specify xorg components, but don't have `-Wl,--as-needed` in their build process. Should be two simple PRs to get those fixed.

---


recipes/sdl_image/all/conanfile.py
```diff
@@ -165,11 +166,14 @@ def generate(self):
             tc.variables["SDL2IMAGE_XV"] = self.options.xv
             tc.variables["SDL2IMAGE_BACKEND_WIC"] = self.options.get_safe("wic")
             tc.variables["SDL2IMAGE_BACKEND_IMAGEIO"] = self.options.get_safe("imageio")
+        if not is_apple_os(self):
+            tc.variables["CMAKE_SHARED_LINKER_FLAGS_INIT"] = '-Wl,--as-needed'
```
valgur, line 170: A code comment on why this is needed would be nice.

valgur (approved):

danimtb (approved):

jcar87 (dismissed):
(see comments)

Ahajha:
@jcar87 What comments? Unless you're referring to @valgur's comment?

jcar87:
> @jcar87 What comments? Unless you're referring to @valgur's comment?

oops sorry! my comments regarding how `--as-needed` has been the default in most Linux distros for years (or should be!) - so it's surprising that it's needed (pun intended) at all - would like to investigate in case that's an actual issue with our gcc docker images

jcar87:
Continuing the discussion here: https://github.com/conan-io/conan-center-index/issues/22485 - really well diagnosed :D

As for this PR, does adding `--as-needed` fix a CI issue that prevents this PR from being merged? If that's the case let me know, otherwise I would suggest removing that line, as by all accounts the issue seems to be somewhere ese

Ahajha:
I will check - I don't think it's entirely necessary at this point (there's been some changes to the xorg recipe since the issue arose). I will try removing it and see what happens.

Ahajha:
@jcar87 Seems like it works, perhaps it would be worth checking if the resulting binaries are still linking to all the extra `xorg` libraries. Maybe the issue still exists, just it doesn't actually cause a problem since all the requisite libraries are there now.

jcar87:
See my comment in https://github.com/conan-io/conan-center-index/issues/22532#issuecomment-2120090210 - 
static `sdl_image` and `shared` SDL fails because when SDL sdl_image is being built as static, it expects the `SDL2::SDL2-static` target to be satisfied by `find_package(SDL2)` - which is not the case when `sdl` is shared. we can fix this in sdl image by creating an alias (without a patch) in a single line, such that `SDL2::SDL2-static` is an alias to `SDL2::SDL2`, however this seems off - I'd be more inclined to mark this as an unsupported/invalid configuration in the validate() method - or to report this upstream first.

Ahajha:
I think I'm just going to close this one. There are newer versions where this isn't an issue, so it's not worth the effort pushing this.

PR #22577, opened 2024-01-29
primesieve: new package

### LINKED ISSUE #14529: [request] primesieve/8.0
### Package Details
  * Package Name/Version: **primesieve/8.0**
  * Website: **https://github.com/kimwalisch/primesieve**
  * Source code: **https://github.com/kimwalisch/primesieve**

primesieve is a command-line program and C/C++ library for quickly generating prime numbers.
It is very cache efficient, it detects your CPU's L1 & L2 cache sizes and allocates its main
data structures accordingly. It is also multi-threaded by default, it uses all available CPU
cores whenever possible i.e. if sequential ordering is not required. primesieve
can generate primes and [prime k-tuplets](https://en.wikipedia.org/wiki/Prime_k-tuple)
up to 2<sup>64</sup>.

primesieve generates primes using the segmented
[sieve of Eratosthenes](https://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) with
[wheel factorization](https://en.wikipedia.org/wiki/Wheel_factorization).
This algorithm has a run time complexity of $O(n\log{\log{n}})$ operations and uses
$O(\sqrt{n})$ memory. Furthermore primesieve uses the
[bucket sieve](http://sweet.ua.pt/tos/software/prime_sieve.html)
algorithm which improves the cache efficiency when generating primes > 2<sup>32</sup>.
primesieve uses 8 bytes per sieving prime, in practice its memory usage is about
$\pi(\sqrt{n})\times 8$ bytes per thread.

Ahajha:
Specify library name and version:  **primesieve/11.2**

This is a continuation of https://github.com/conan-io/conan-center-index/pull/14530, which was essentially already ready. I've just updated the package to 11.2.

Resolves #14529

---


recipes/primesieve/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+required_conan_version = ">=1.53.0"
+
+class PackageConan(ConanFile):
+    name = "primesieve"
+    description = "Fast prime number generator"
+    license = "BSD-2-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/kimwalisch/primesieve"
+    topics = ("math", "prime-numbers", "sieve-of-eratosthenes")
+    settings = "os", "arch", "compiler", "build_type"
```
RubenRBS, line 18: Missing package_type
```suggestion
    package_type = "library"
    settings = "os", "arch", "compiler", "build_type"
```

recipes/primesieve/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  std::cout << "Detect 16-bit overflow: " << errorMsg;
+  check(!errorMsg.empty());
+
+  std::cout << std::endl;
+  std::cout << "All tests passed successfully!" << std::endl;
+
+  return 0;
+}
```
RubenRBS, line 106: Can we simplify this by a lot? `test_package`s should limit themselves to small tests checking that the library was correctly packaged - no need to be generating primes here :)

RubenRBS (commented):
Quick review

recipes/primesieve/all/test_package/test_package.cpp
Ahajha: I agree, I can probably simplify to check if just one number is prime, and that should be a sufficient test.

recipes/primesieve/all/conanfile.py
Ahajha: Done

recipes/primesieve/all/test_package/test_package.cpp
Ahajha: Done, program just generates and prints 5 primes.

Ahajha:
@kimwalisch FYI, the PR from a few years ago was never merged, I'm continuing it now in hopes that we can actually add it.

Side question for you, you require C++11, which GCC 5 does support, but the package seems to fail when using it. Any idea why? Do you want to support this version?

@Alex-PLACET I'm continuing your work, which requires you to have the CLA signed (I think it has 4 month expiration or so), would you mind doing that when you get the chance? No rush, still working on getting things fixed.

kimwalisch:
> Side question for you, you require C++11, which GCC 5 does support, but the package seems to fail when using it. Any idea why? Do you want to support this version?

Yes, primesieve requires full C++11 support. Some old GCC compilers only partially support C++11 (e.g. ```std::align``` from C++11 is not supported by these compilers) and therefore they fail to build primesieve. E.g. I know that GCC 4.9 cannot compile primesieve.

But if the C++ compiler is able to build primesieve, then it must work correctly at runtime (otherwise this is a bug). If you are experiencing primesieve runtime errors then I am interested in seeing those logs.

Ahajha:
> But if the C++ compiler is able to build primesieve, then it must work correctly at runtime (otherwise this is a bug). If you are experiencing primesieve runtime errors then I am interested in seeing those logs.

Conan test packages are only intended to test if the package itself works. Usually this is verifying a simple example compiles, links, and executes without errors. The test package here just generates a handful of prime numbers, which is a simple enough test that it probably won't show any errors unless the packaging is wrong on our end.

Occasionally there is a bug report which involves how it is packaged as the cause of the bug, but those tend to be uncommon.

Alex-PLACET:
@Ahajha, a big thank you for picking up where I left off on my PR. I completely forgot to keep it going. Appreciate your help!

recipes/primesieve/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+        copy(self, pattern="COPYING", dst=os.path.join(self.package_folder, "licenses"), src=self.source_folder)
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.la", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
+        if is_msvc(self) and self.options.shared:
```
RubenRBS, line 89: Any insight into this rename? It looks funny that it gets generated with this name in the first place :)

Ahajha: It seems this is set to avoid issues when packaging both static and shared libraries in the same package. This doesn't apply to us, so I think it's fine to do this? Unless we have a way of telling Conan that this is how it's done.

https://github.com/kimwalisch/primesieve/blob/master/CMakeLists.txt#L138

recipes/primesieve/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        
+        if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) <= "5":
+            raise ConanInvalidConfiguration("GCC<=5 is currently not supported. Contributions with fixes are welcome.")
```
uilianries, line 55:
```suggestion
```

I used to work with GCC 4.7 with C++11 minimal support. GCC 4.8 has a really good coverage of C++11. Usually, we don't add restriction to compiler version when the project only requires C++11

recipes/primesieve/all/conanfile.py
uilianries: Please, keep the original name, it should work as usual, right? I remember having other projects having this same prefix when exporting dll libs

recipes/primesieve/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+        self.cpp_info.set_property("cmake_file_name", "primesieve")
+        self.cpp_info.set_property("cmake_target_name", "primesieve::primesieve")
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs = ["pthread", "m"]
+
+        # Conan 1.x only
+        self.cpp_info.filenames["cmake_find_package"] = "primesieve"
+        self.cpp_info.filenames["cmake_find_package_multi"] = "primesieve"
+        self.cpp_info.names["cmake_find_package"] = "primesieve"
+        self.cpp_info.names["cmake_find_package_multi"] = "primesieve"
```
uilianries, line 103:
```suggestion
```
Not needed: https://github.com/conan-io/conan-center-index/discussions/17406

recipes/primesieve/all/test_v1_package/CMakeLists.txt
uilianries, line 1: The folder test_v1_package can entirely removed. https://github.com/conan-io/conan-center-index/discussions/17406

uilianries (commented):
Thank you for your PR, it's looking good, just few comments. This PR can be reduced a bit.

recipes/primesieve/all/conanfile.py
Ahajha: GCC 5 fails to build, not sure why though. @kimwalisch mentioned above that 4.9 is known not to work, for example.

recipes/primesieve/all/conanfile.py
Ahajha: It would fail without this change, I presume it tries to look for "primesieve".lib but it's "primesieve.dll".lib. I'm fine with changing it back, I just don't know how I would get it to work.

recipes/primesieve/all/conanfile.py
Ahajha: Done

recipes/primesieve/all/test_v1_package/CMakeLists.txt
Ahajha: Done

recipes/primesieve/all/conanfile.py
franramirez688: @uilianries @RubenRBS Many more recipes are doing this rename. You can run a grep and see this ad-hoc helper function `fix_msvc_libname`. The problem is coming from the hook error:
```
primesieve/11.2: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
[HOOK - conan-center.py] post_package_info(): ERROR: [LIBRARY DOES NOT EXIST (KB-H054)] Component primesieve::primesieve library 'primesieve' is listed in the recipe, but not found installed at self.cpp_info.libdirs. Make sure you compiled the library correctly. If so, then the library name should probably be fixed. Otherwise, then the component should be removed. (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H054-LIBRARY-DOES-NOT-EXIST) 
ERROR: 
	ConanException: [HOOK - conan-center.py] post_package_info(): Some checks failed running the hook, check the output
```
@Ahajha We just created an issue to review this process https://github.com/conan-io/hooks/issues/528 Let's figure out if there is a bug in the hooks so we can use the original library name instead of the renamed one.

RubenRBS (approved):

Issue #20768, opened 2023-10-25
Cpython 


jellespijker:
### Package Name/Version

cpython/3.12.0

### Changelog

https://docs.python.org/3/whatsnew/changelog.html

### Context about the new update

The above-mentioned version is newly released by the upstream project and not yet available as a recipe.
Please add this version.

one of our main dependencies for https://github.com/Ultimaker/Cura is CPython, this recipe is also the only recipe which is blocking us for moving forward to using Conan 2.0 instead of Conan 1.x

We're currently investigating how much effort it will be to switch from using CPython 3.10.4 to 3.12 and will probably make the actual switch in a month or so. Having 3.12.x in the CCI would be a huge help for us.

3.12 has ~25% speed improvements along the board compared to 3.10.x and I'm sure other project would also benefit from having 3.12 in the CCI

jellespijker:
@memsharded can I lobby with you to give this recipe a bit of priority? We at UltiMaker are ofcourse more than willing to also put in the effort to get CPython 3.12 in the CCI so we can start upgrading to Conan 2.0

samuel-emrys:
The blocker for cpython at the moment is ncurses in conan 2, for which there have been a few attempts:
* https://github.com/conan-io/conan-center-index/pull/20355
* https://github.com/conan-io/conan-center-index/pull/18568
* https://github.com/conan-io/conan-center-index/pull/17805

Some of these are close, but haven't been able to pass the full complement of CI tests. I haven't had the time to invest in continuing my draft PR - if you have that time/effort available then that would be useful work to do.

jellespijker:
That is good to known, thank you @samuel-emrys I might do some personal digging in that recipe this weekend. What is the policy of the CCI with regards to updating recipes without full Conan v2 support?

Or stack is currently still stuck on using Conan v1, but we need to update our CPython versions because of the deprecation of OpenSSL1.1.1 If we can update the CPython recipe to 3.12 without fully supporting Conan v2 would that be acceptable?

valgur:
I have a branch with ongoing work for the CPython recipe, but it still has some linking issues against third-party libraries that I have not figured out: https://github.com/valgur/conan-center-index/tree/migrate%2Fcpython/recipes%2Fcpython
I would love to see the recipe both migrated and updated to 3.12. I'm personally looking forward to making use of the new stable ABI for forwards-compatible binary modules.
[...]

PR #22599, opened 2024-01-30
CPython: Condense patching, update patch versions, and add 3.11/3.12

Ahajha:

Depends on https://github.com/conan-io/conan-center-index/pull/21387

Resolves https://github.com/conan-io/conan-center-index/issues/20768  

Context for some of the changes:
- The MSVC project patches are extremely fragile, usually breaking on any patch release. I've switched these to an extensive `replace_in_file` system, though admittedly I'm not a massive fan. Ideally I'd like to use fuzzy patching, but as far as I can tell they don't work.
- CPython switched to using PkgConfig for most of its dependencies in 3.12, and removed setup.py for building extensions. 3.11 is in a transition state, where maybe half the deps used PkgConfig, and the other half was "discovered" manually through setup.py.

Once this is merged, I will make a draft PR for 3.13 (yet unreleased), as there are quite a few removals from Python's stdlib that will change some patching and remove at least one dependency: https://github.com/python/cpython/issues/104773

---


Ahajha:
Seems to be bitten by this issue: https://github.com/python/cpython/issues/99042

Very concerning to me that it's fetching another version of Python automatically, definitely don't want that.

Ahajha:
@uilianries @RubenRBS And now this PR is ready for your reviewing pleasure :sweat_smile:

recipes/cpython/all/conanfile.py
```diff
@@ -173,6 +183,9 @@ def validate(self):
                 if self.dependencies["mpdecimal"].ref.version < Version("2.5.0"):
                     raise ConanInvalidConfiguration("cpython 3.9.0 (and newer) requires (at least) mpdecimal 2.5.0")
 
+        if self.settings.compiler == "gcc" and Version(self.settings.compiler.version).major == 9 and Version(self.version) >= "3.12":
```
uilianries, line 186: Could you please point the error?

recipes/cpython/all/conanfile.py
```diff
@@ -236,11 +253,145 @@ def generate(self):
...
+                                "openssl_libdirs = ",
+                                f"openssl_libdirs = {openssl.libdirs + zlib.libdirs} #")
+                replace_in_file(self, setup_py,
+                                "openssl_libs = ",
+                                f"openssl_libs = {openssl.libs + zlib.libs} #")
+
+            if Version(self.version) < "3.11":
+                replace_in_file(self, setup_py, "if (MACOS and self.detect_tkinter_darwin())", "if (False)")
+
+    def _patch_msvc_projects(self):
```
uilianries, line 301: It's really hard to read all this, would be possible to covert to a patch instead?

recipes/cpython/all/conanfile.py
```diff
@@ -102,6 +102,10 @@ def configure(self):
     def layout(self):
         basic_layout(self, src_folder="src")
 
+    def build_requirements(self):
+        if Version(self.version) >= "3.11" and not is_msvc(self) and not self.conf.get("tools.gnu:pkg_config", check_type=str):
```
uilianries, line 106: Why is it needed now?

uilianries:
@Ahajha I see many changes in this PR, please, always try to keep the PR description updated with your changes. It will help us by a lot not only to review, but to track this changes in the future.

recipes/cpython/all/conanfile.py
Ahajha: It would be hell indeed, they basically broke with every new patch version since they were so large. That was the motivation for switching to this method. There's an in-between option that I could explore which is to split the patches by file, in hopes that some of them could be re-used between versions. That might be more readable, but I don't know how much effort that would save us in the long run.

The option I _ideally_ wanted was fuzzy patching, but that seems to just not work, any insight would be appreciated there.

The injection of the dependency's `.prop`s files have to be done this way I think, since we can't configure the build type in the patch.

recipes/cpython/all/conanfile.py
Ahajha: CPython started moving some of its dependency management to `pkg-config` as of 3.11, and (more or less) fully in 3.12.

recipes/cpython/all/conanfile.py
```diff
@@ -116,8 +120,12 @@ def requirements(self):
         if self.settings.os != "Windows":
             if not is_apple_os(self):
                 self.requires("util-linux-libuuid/2.39.2")
-            # If crypt.h is detected, it is included in the public headers.
-            self.requires("libxcrypt/4.4.36", transitive_headers=True, transitive_libs=True)
+            # In <3.9 and lower patch versions of 3.9/10/11, crypt.h was exposed in Python.h
+            # This was removed in 3.11 and backported: https://github.com/python/cpython/issues/88914
+            # For the sake of this recipe, we only have later patch versions, so this version check
+            # may be slightly inaccurate if a lower patch version is desired.
+            transitive_crypt = Version(self.version) < "3.9"
```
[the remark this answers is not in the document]
Ahajha: I can, it would have just been a little complex. I also figured it wasn't the end of the world if they ended up being transitive if they didn't have to be.

recipes/cpython/all/conanfile.py
Ahajha: https://c3i.jfrog.io/c3i/misc/logs/pr/22599/17-linux-gcc-legacy/cpython/3.12.2//2a6be3fe8381eeefd48805e3669056270fb8f4a1-build.txt - search for "posixshmem". Locally I got an internal compiler error, not entirely sure what the error in CCI is, but I do know that it's GCC 9 exclusive (or maybe it affects 8 and 10, but not sure since those aren't tested in CCI)

uilianries: Thank you, I'll check if is something exclusive with docker image, due old ubuntu version, or is gcc9. Depending the case, we can skip the build, but allow users to build with gcc9

Ahajha:
> @Ahajha I see many changes in this PR, please, always try to keep the PR description updated with your changes. It will help us by a lot not only to review, but to track this changes in the future.

All these changes are just those necessary for 3.11 and 3.12. The only slightly irrelevant change is the patching system which I also admittedly am not a fan of.

recipes/cpython/all/conanfile.py
Ahajha: I also get an ICE locally with GCC 9, so my guess is that it's just that version.

Croydon:
Some in-between commits also had Python 2.7 which would be a nice-to-have to build/run legacy stuff. Was it too hard to get it working? What were the challenges?

Ahajha:
> Some in-between commits also had Python 2.7 which would be a nice-to-have to build/run legacy stuff. Was it too hard to get it working? What were the challenges?

We ended up dropping it in the other PR since it's majorly EOL, but I had it more or less working (I think I had to disable one small test, but nothing major). I figured there was a case for having the recipe floating around, but maybe not in CCI. @RubenRBS and @uilianries had some things to say as well, but if you think there's a case for it I can definitely re-add it, probably as a different subdir at this point since it's so old (and while not _that_ different buildwise, enough to not muddy up the main folder).

Ahajha:
BTW the above error is the cryptic Mac install error that occasionally shows up. Haven't found a good solution to it yet.

Ahajha:
I'm fairly sure that if https://github.com/conan-io/python-patch-ng/issues/19 were fixed, we could have a more ideal system where we could use fuzzy patching, which would give us the ease-of-reading of patches and the resilience of replace_in_file/a custom regex_replace_in_file.

@uilianries It looks like that project hasn't been updated in a few years, has it been integrated directly in Conan or has it just not received activity? I could potentially look into fixing it, assuming the update could be added into 1.x as well.

recipes/cpython/all/conanfile.py
```diff
@@ -236,11 +253,145 @@ def generate(self):
...
         setup_py = os.path.join(self.source_folder, "setup.py")
         if Version(self.version) < "3.10":
             replace_in_file(self, setup_py, ":libmpdec.so.2", "mpdec")
+
+        if self.options.get_safe("with_curses", False):
+            # FIXME: this will link to ALL libraries of ncurses. Only need to link to ncurses(w) (+ eventually tinfo)
+            ncurses_info = self.dependencies["ncurses"].cpp_info.aggregated_components()
+            replace_in_file(self, setup_py,
+                "curses_libs = ",
+                "curses_libs = {} #".format(repr(ncurses_info.libs + ncurses_info.system_libs)))
```
valgur, line 282:
```suggestion
            libcurses = self.dependencies["ncurses"].cpp_info.components["libcurses"]
            tinfo = self.dependencies["ncurses"].cpp_info.components["tinfo"]
            libs = libcurses.libs + libcurses.system_libs + info.libs + tinfo.system_libs
            replace_in_file(self, setup_py,
                "curses_libs = ",
                "curses_libs = {} #".format(repr(libs)))
```
Might as well include it since it's a simple fix.

recipes/cpython/all/conanfile.py
valgur: It's not great but looks ok to me. Could use a few more comments for the reason for the patch for each block, if possible.

valgur (approved):

recipes/cpython/all/conanfile.py
Ahajha: Yea I think there's certainly a tradeoff here between readability and resilience to change. Can certainly revisit if fuzzy patching becomes an option again.

I can add some comments, oddly enough one of the issues with the original patches is that they're missing a lot of context as well.

recipes/cpython/all/conanfile.py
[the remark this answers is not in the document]
Ahajha: Personally, I don't see much value in adding it now that I think about it. If these were recent patches, I would do so, but xcrypt has been non-transitive for a few years at this point (late 2021), so it would be a very specific requirement on a specific old patch version that would warrant this necessary.

Ahajha:
@Croydon Regarding Python 2, perhaps make an issue to discuss? I think we're all aware that it will be a while before Python 2 finally kicks the bucket, the concern was general usefulness. If you have a use case I think it would be worth re-adding (shouldn't be difficult to do so).

recipes/cpython/all/conanfile.py
Ahajha: I added some comments, most things are "unvendor XYZ", there are a few things that I may need to take another look at.

Ahajha:
So for some reason the apple install failure is being extremely persistent. I'll have another go at fixing it soon-ish.

Ahajha:
Looks like https://github.com/python/cpython/issues/109796 is the issue. I'll add the workaround mentioned to the package step, which should hopefully avoid the issue.

valgur:
Not directly related to this PR, but these pkg-config aliases need to be turned into lists:
https://github.com/conan-io/conan-center-index/blob/66ad79172b7d1759b89893ae6dde391e0682a020/recipes/cpython/all/conanfile.py#L773
https://github.com/conan-io/conan-center-index/blob/66ad79172b7d1759b89893ae6dde391e0682a020/recipes/cpython/all/conanfile.py#L785

I'm currently seeing a .pc file for each of the letters in `python3.10-embed`. :)

Ahajha:
@valgur That's the second time I've been bitten by that exact problem, yes a check for that would be wonderful. I've fixed the issue, but of course now an unrelated issue popped up :/

valgur (approved):

RubenRBS (approved):
Thanks _a lot_. Can't thank you enough :)

recipes/cpython/all/conandata.yml
```diff
@@ -1,65 +1,79 @@
...
+    - patch_file: "patches/3.12/3.12.1-0001-_ctypes-ffi.patch"
+      patch_description: "Support shared libffi"
+      patch_type: "portability"
+    - patch_file: "patches/3.x-0001-relocatable-python-config.patch"
+      patch_description: "Allow package to be relocatable"
+      patch_type: "conan"
+    - patch_file: "patches/3.12/3.12.1-0002-remove-module-deps.patch"
+      patch_description: "Remove section of solution file forcing projects to be built that might not be used for this recipe"
+      patch_type: "conan"
+  "3.11.8":
```
ericLemanissier, line 31: @Ahajha is this a typo?

Ahajha: Ah, yea it is. I have another PR out, I'll fix it in that one.

ericLemanissier: Thanks, great!

PR #22614, opened 2024-02-01
async_simple: add 1.3, reinstate msvc, add compiled version

RazielXYZ:
Specify library name and version:  **async_simple/1.3**

This quite heavily modifies the recipe to support non-header-only-mode (only on not windows, since the only reason to compile it is for uthread, which only works on linux and osx)
It also removes the invalid throw on msvc - not sure why it was there in the first place, even 1.0 appears to work fine in msvc.
And finally, it adds the latest just released version, 1.3

---


valgur (approved):

danimtb (approved):

danimtb (approved):

recipes/async_simple/all/conandata.yml
```diff
@@ -1,7 +1,14 @@
...
+    sha256: "0BA0DC3397882611B538D04B8EE6668B1A04CE046128599205184C598B718743"
   "1.2":
     url: "https://github.com/alibaba/async_simple/archive/refs/tags/1.2.tar.gz"
     sha256: "a59a2674ac2b0a3997b90873b2bf0fbe4d96fdedbe6a2628c16c92a65a3fa39a"
   "1.0.0":
     url: "https://github.com/alibaba/async_simple/archive/refs/tags/1.0.tar.gz"
     sha256: "b243fb7af5d61b534fe18b662d16498392bbce1deffdc68e58829aa31db7badf"
+patches:
+  "1.0.0":
+    - patch_file: "patches/disable-asan-1.0.patch"
```
uilianries, line 13: Could you please add more information about these patches? What's the origin, the motivation, type ... 

https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conandata_yml_format.md#patches-fields

danimtb (approved):

recipes/async_simple/all/conandata.yml
uilianries: @RazielXYZ Thank for adding extra information.

The patch https://github.com/alibaba/async_simple/commit/ccb760356b9d7481ca73f0c58d15928bf8b0aa0b does quite opposite, it add the option ASYNC_SIMPLE_ENABLE_ASAN and enables it by default. Using this patch, it will disable actually. I don't see it following the upstream default behavior, but diverging from it. Am I missing something?

danimtb (changes requested):
Waiting for answer and requested changes at https://github.com/conan-io/conan-center-index/pull/22614#discussion_r1579653019

recipes/async_simple/all/conandata.yml
RazielXYZ: @uilianries The upstream default behavior is to enable ASAN, yes, but prior to that commit they had no option for it, and therefore no way to disable it. I assume we don't want packages built with ASAN on conancenter, as they do not work when linked against things not built with ASAN (unless I'm wrong!). As such, before they added the option, the only way to not have it build with ASAN would be to remove -fsanitize=address from the cmakefiles which these patches do. If the extra info makes that unclear I am open to suggestions!

uilianries: Agreed. Could you please use the official patch, then pass the cmake definition to disable asan (no conan option). Just to keep officially following the upstream.

recipes/async_simple/all/conanfile.py
```diff
@@ -37,59 +48,109 @@ def _compilers_minimum_version(self):
...
-        copy(
-            self,
-            pattern="*.h",
+        
+        if self.options.header_only:
+            hCopyExcludes = ("test", "uthread", "executors")
+        if not self.options.header_only:
+            hCopyExcludes = ("test", "executors")
+            if self.options.shared:
+                copy(self, pattern="*.so*",
```
uilianries, line 112: Mac generates .dylib, not .so. The current validate supports mac as well.

danimtb (approved):

recipes/async_simple/all/conanfile.py
```diff
@@ -37,59 +48,113 @@ def _compilers_minimum_version(self):
...
-        else:
-            raise ConanInvalidConfiguration("msvc not supported now")
-
-    def build_requirements(self):
-        pass
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support.")
+        if not self.options.header_only and self.settings.os == "Windows":
+            raise ConanInvalidConfiguration(f"{self.ref} cannot use uthread on Windows, and therefore should only be header_only")
```
RubenRBS, line 79:
```suggestion
            raise ConanInvalidConfiguration(f'{self.ref} cannot use uthread on Windows, and therefore it's only supported as a header-only library in this configuration with "{self.name}/*:header_only=True"')
```

RubenRBS (commented):
Thanks! A few minor changes, otherwise looks good :)

danimtb (approved):

jwillikers (approved):

RubenRBS (approved):
Thanks!

PR #22629, opened 2024-02-02
libdwarf: Add 0.9.0 and 0.9.1

jeremy-rifkin:
Specify library name and version:  **libdwarf/0.9.0** **libdwarf/0.9.1**

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #22655, opened 2024-02-04
cpptrace: Add v0.4.0

jeremy-rifkin:
Specify library name and version:  **cpptrace/0.4.0**

I'm the author, updating for 0.4.0. Depends on https://github.com/conan-io/conan-center-index/pull/22629.

---


recipes/cpptrace/all/conanfile.py
```diff
@@ -43,7 +43,10 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("libdwarf/0.8.0")
+        if Version(self.version) >= Version("0.4.0"):
+            self.requires("libdwarf/[>=0.9.1 <20191104]")
+        else:
+            self.requires("libdwarf/[>=0.8.0 <0.9.0 <20191104]")
```
uilianries, line 49:
```suggestion
        if Version(self.version) >= Version("0.4.0"):
            self.requires("libdwarf/[>=0.9.1 <20191104]")
        else:
            self.requires("libdwarf/[>=0.8.0 <0.9.0 <20191104]")
```

Please, do not use version range for libdwarf. Only few versions are allowed to use it in CCI:

There is the risk changing the recipe revision, resulting in missing packages. In small steps we are allowing more packages to use it.

jeremy-rifkin: Thanks!

recipes/cpptrace/all/conanfile.py
```diff
@@ -60,6 +63,8 @@ def generate(self):
         tc = CMakeToolchain(self)
         if is_msvc(self):
             tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+        if Version(self.version) >= Version("0.4.0"):
+            tc.variables["CPPTRACE_USE_EXTERNAL_ZSTD"] = True
```
RubenRBS, line 67: No `self.requires` for zstd?

RubenRBS (commented):
Thanks! One question about `CPPTRACE_USE_EXTERNAL_ZSTD`, else all looks good :)

recipes/cpptrace/all/conanfile.py
jeremy-rifkin: Thank you for noticing that, I'm realizing I should rework something in cpptrace's cmake. Cpptrace doesn't actually need zstd, that's needed by libdwarf.

jeremy-rifkin: I made a change upstream and applied the patch here

jeremy-rifkin:
Thank you so much!

Issue #20548, opened 2023-10-12
[package] libiconv/1.17: recipe is not Conan 2.0 ready


Sage-Cat:
### Description

Wanted to install qt/[6.5.3]

### Package and Environment Details

* Package Name/Version: **libiconv/1.17**
* Operating System+version: **Windows 11 (10.022621 Build 22621)**
* Compiler+version: **gcc.exe (x86_64-posix-seh-rev3, Built by MinGW-W64 project) 11.2.0**
* Docker image: **not used**
* Conan version: **conan 2.0.13**
* Python version: **Python 3.11.6**

### Conan profile

Host profile:
[settings]
arch=x86_64
build_type=Debug
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=11
os=Windows
[options]
*:shared=True

Build profile:
[settings]
arch=x86_64
build_type=Debug
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=11
os=Windows
[options]
*:shared=True

### Steps to reproduce

conanfile.py
```
from conan import ConanFile
from conan.tools.cmake import cmake_layout, CMakeDeps, CMakeToolchain
from conan.tools.files import copy


class SageStorePkg(ConanFile):
    name = "sage_store"
    version = "1.0.0"
    generators = "CMakeDeps", "CMakeToolchain"

    settings = "os", "compiler", "build_type", "arch"
    implements = ["auto_shared_fpic"]

    def requirements(self):
        self.requires("cmake/[3.27.7]")
        self.requires("spdlog/[1.12.0]")
        self.requires("qt/[6.4.2]")

    def layout(self):
        cmake_layout(self, src_folder=".", build_folder="build")

    def generate(self):
        cmake = CMakeDeps(self)
        cmake.generate()

        tc = CMakeToolchain(self)
        tc.generate()

        for dep in self.dependencies.values():
            copy(self, "*.dll", dep.cpp_info.libdirs[0], self.build_folder)
            copy(self, "*.so", dep.cpp_info.libdirs[0], self.build_folder)

```

conan install . --output-folder=build --build=missing -s build_type=Debug

### Logs

<details><summary>Click to expand log</summary>

```
libiconv/1.17: ERROR:
Package '77226ff69664c3e7f944d49a9e66c1e7c00ec32b' build failed
[...]

PR #22665, opened 2024-02-05
libiconv: Fix building on mingw

Ahajha:
Specify library name and version:  **libiconv/all**

Resolves https://github.com/conan-io/conan-center-index/issues/20548  
Resolves https://github.com/conan-io/conan-center-index/issues/506  

The old [bincrafters recipe](https://github.com/bincrafters/conan-libiconv/blob/testing/1.16/conanfile.py) worked with mingw (in Conan 1.x), and I saw the main lines that made it work:
```python
        if self._is_mingw_windows or self._is_msvc:
            prefix = prefix.replace('\\', '/')
            build = False
            if self.settings.arch == "x86":
                host = "i686-w64-mingw32"
                rc = "windres --target=pe-i386"
            elif self.settings.arch == "x86_64":
                host = "x86_64-w64-mingw32"
                rc = "windres --target=pe-x86-64"
```
(plus some added context, but the idea is there)

This PR just migrates these lines to the new build helpers. Admittedly, I have no idea what the `RC` and `WINDRES` flags are for.

---


uilianries:
Hello @Ahajha Did you try to build using mingw with this change? If yes, do you have a full build log to shared your results?

Locally, I can't build:

https://gist.github.com/uilianries/00567ffda78d3b96a63c5dfaaf314cb4

Ahajha:
> Hello @Ahajha Did you try to build using mingw with this change? If yes, do you have a full build log to share your results?
> 
> Locally, I can't build:
> 
> https://gist.github.com/uilianries/00567ffda78d3b96a63c5dfaaf314cb4

I did and it worked, I will get the log shortly.

Ahajha:
@uilianries Here is the log from my local build. I installed mingw separately using this tutorial: https://code.visualstudio.com/docs/cpp/config-mingw

[log.txt](https://github.com/conan-io/conan-center-index/files/14170734/log.txt)

Ahajha:
@uilianries That also works for me.
[log2.txt](https://github.com/conan-io/conan-center-index/files/14170846/log2.txt)
Adding -o */*:shared also works.
[log3.txt](https://github.com/conan-io/conan-center-index/files/14170864/log3.txt)

uilianries (approved):
With the build log using mingw I'm comfortable with this change.

ericLemanissier (approved):

amachado-pie:
This PR fails for cross-compiling wih gcc, when the profile:host is `Windows` but the  profile:build is `Linux` because mingw toolchain binaries in Linux has a name prefix `i686-w64-mingw32-`, ie., `i686-w64-mingw32-windres` , etc. 
So, then we try to do a cross-compilation for Windows from Linux it fails:
```
libtool: compile:  windres --target=pe-i386 --target=pe-i386 -DPACKAGE_VERSION_STRING=\\\"1.17\\\" -DPACKAGE_VERSION_MAJOR=1 -DPACKAGE_VERSION_MINOR=17 -DPACKAGE_VERSION_SUBMINOR=0 -i /home/amachado/.conan2/p/b/libic362ec7a18a879/b/src/lib/../windows/libiconv.rc --output-format=coff -o libiconv.res.o
../libtool: line 1932: windres: command not found
make[1]: *** [Makefile:99: libiconv.res.lo] Error 1
```

jcar87:
based on the report in https://github.com/conan-io/conan-center-index/issues/26445, I think this needs to be reverted.

Unfortunately, while the PR mentions  https://github.com/conan-io/conan-center-index/issues/20548 and  https://github.com/conan-io/conan-center-index/issues/506 - I can't find any evidence in the reported logs that those build failures had anything to do with windres? 

`windres` is autodetected by autotools - it tries `windres` and also `i686-w64-mingw32-windres` which it derives from the compiler target triple)  - this PR breaks the autodetection when the user is using a toolchain that only has the latter. 
* then it is invoked during the libtool phases - but the errors reported in t

The errors reported in those two issues, are compiler errors, reported for older versions of gcc and older versions of libiconv than the ones reported in the PR comments - so I'm not entirely sure we had enough information to approve this in the first place.

hi @Ahajha , @uilianries  - do we have build logs where building libiconv with mingw _fails_ without the changes in this PR? if I remove these lines, I can build with mingw on Windows (both with msys2 itself, and with the mingw-builds conan center package), as well as on Linux (with the toolchain provided by Ubuntu) - I'd hate to introduce a regression by removing these lines - but on the other hand I cant find any logs where the error is related to windres

PR #22667, opened 2024-02-05
gperftools: add a test for plain `malloc()` support

valgur:
Adds `MALLOCSTATS=1` env var, which enable malloc statistics on program termination:
```
gperftools/2.15 (test package): RUN: ./test_package_indirect
------------------------------------------------
MALLOC:          90112 (    0.1 MiB) Bytes in use by application
MALLOC: +       942080 (    0.9 MiB) Bytes in page heap freelist
MALLOC: +        16264 (    0.0 MiB) Bytes in central cache freelist
MALLOC: +            0 (    0.0 MiB) Bytes in transfer cache freelist
MALLOC: +          120 (    0.0 MiB) Bytes in thread cache freelists
MALLOC: +      2490368 (    2.4 MiB) Bytes in malloc metadata
MALLOC:   ------------
MALLOC: =      3538944 (    3.4 MiB) Actual memory used (physical + swap)
MALLOC: +            0 (    0.0 MiB) Bytes released to OS (aka unmapped)
MALLOC:   ------------
MALLOC: =      3538944 (    3.4 MiB) Virtual address space used
MALLOC:
MALLOC:              7              Spans in use
MALLOC:              2              Thread heaps in use
MALLOC:           8192              Tcmalloc page size
------------------------------------------------
Call ReleaseFreeMemory() to release freelist memory to the OS (via madvise()).
Bytes released to the OS take up virtual address space but no physical memory.
```

This allows the use of plain `malloc()` to be tested in a verifiable manner in addition to the existing direct `tc_malloc()` functions from an included header.

The stats output uses stderr, which cannot be captured in Conan v1 easily, so I did not add any asserts for the stats output actually being present in `self.run()`.


recipes/gperftools/all/test_package/conanfile.py
```diff
@@ -19,7 +21,20 @@ def build(self):
         cmake.configure()
         cmake.build()
 
+    def _test(self, executable):
+        bin_path = os.path.join(self.cpp.build.bindir, executable)
+        if conan_version.major >= "2.0.15":
+            stderr = io.StringIO()
+            kwargs = {"stderr".strip(): stderr}
+            self.run(bin_path, env="conanrun",  **kwargs)
```
ericLemanissier, line 29:
```suggestion
            self.run(bin_path, env="conanrun",  stderr=stderr)
```

recipes/gperftools/all/test_package/conanfile.py
```diff
@@ -19,7 +21,20 @@ def build(self):
         cmake.configure()
         cmake.build()
 
+    def _test(self, executable):
+        bin_path = os.path.join(self.cpp.build.bindir, executable)
+        if conan_version.major >= "2.0.15":
```
ericLemanissier, line 26:
```suggestion
        if conan_version>= "2.0.15":
```

ericLemanissier (changes requested):

recipes/gperftools/all/test_package/conanfile.py
valgur: Would love to do it that way, but even with the current obfuscation, the linter still complains about `stderr` being an unsupported keyword argument. :/

recipes/gperftools/all/test_package/conanfile.py
valgur: Thanks!

recipes/gperftools/all/test_package/conanfile.py
ericLemanissier: It's just an opinion, so do as other reviewers will say.
As far as I'm concerned, If there's a linter message anyway I prefer the one I can understand. At least the code is more readable with the suggestion.

recipes/gperftools/all/test_package/test_package_indirect.c
```diff
@@ -0,0 +1,9 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <stdio.h>
+
+int main() {
+    void *p = malloc(100);
+    free(p);
```
ericLemanissier, line 7:
```suggestion
    printf("%p\n", p);    
    free(p);
```
Without this, there is a risk that the compiler choses to optimize this into `return EXIT_SUCCESS` : https://godbolt.org/z/vP3vq9o37 vs https://godbolt.org/z/GvTYEGPqv

jwillikers (approved):

PR #22668, opened 2024-02-05
[config] Add waitaing list for users in authorized users

uilianries:
- This configuration is not running yet.
- Some users change name in middle of asking for access request and when asking about the old username, there is no answer about. These outdated and not responded users will be moved to a waiting list where we can ping from to time, but will not have access to PRs until be move to authorized users list.

---

### DIFF
--- a/.c3i/waitlist_users.yml
+++ b/.c3i/waitlist_users.yml
@@ -1,3 +1,4 @@
 waitlist_users:
 - Mike-Solar
 - retroandchill
+- refactorTractor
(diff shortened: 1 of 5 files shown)


.c3i/waitlist_users.yml
```diff
@@ -0,0 +1,3 @@
+waitlist_users:
+- Mike-Solar
+- retroandchill
```
uilianries, line 3:
```suggestion
- retroandchill
- refactorTractor
```

uilianries:
The action error is wrong, probably Intel is blocking scrapping. We can access https://www.intel.com/content/www/us/en/developer/tools/oneapi/ipp.html normally.

RubenRBS (approved):

danimtb (approved):

PR #22670, opened 2024-02-05
boost: First gcc version to support c++11 was gcc 5

klausholstjacobsen:
boost

First gcc version to support c++11 was gcc 5:
https://gcc.gnu.org/projects/cxx-status.html

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -166,7 +166,7 @@ class BoostConan(ConanFile):
         """ Minimum compiler version having c++ standard >= 11
         """
         return {
-            "gcc": 6,
+            "gcc": 5,
             "clang": 6,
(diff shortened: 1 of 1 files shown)


uilianries:
@klausholstjacobsen It's a different case in Boost recipe, it uses what GCC version uses that C++ standard by default. The GCC 5 does not use C++11, by default, but 6.

https://gcc.gnu.org/gcc-5/changes.html

Indeed GCC 5 supports C++11 and should be able to build Boost (at least most of libraries). So the PR #22554 should fixes it. That PR should be able build GCC 5 as you can see here: https://c3i.jfrog.io/c3i/misc/logs/pr/22554/2-linux-gcc-legacy/boost/1.74.0//5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9-build.txt

Issue #22619, opened 2024-02-01
[package] boost/1.83.0: Pass `tools.build:defines` to `b2` (Boost.Build)


stevenwdv:
### Description

While e.g. `tools.build:cxxflags` is passed to `b2`, `tools.build:defines` is not. This means something like `tools.build:defines=["_WIN32_WINNT=0x0601"]` does not get passed to the compiler when compiling Boost, possibly causing linking issues when the value *is* used when compiling the consumer package, see #14097.

A workaround is to include `'extra_b2_flags': ' '.join(f'define={d}' for d in self.conf.get('tools.build:defines', [])) or None` in the consumer conanfile when requiring Boost.

### Package and Environment Details

* Package Name/Version: **boost/1.83.0**
* Operating System+version: N/A
* Compiler+version: N/A
* Docker image: N/A
* Conan version: **conan 2.0.17**
* Python version: N/A

### Conan profile

```
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.version=192
compiler.runtime=dynamic
os=Windows

[conf]
tools.build:defines=["_WIN32_WINNT=0x0601"]
```

### Steps to reproduce

Run `conan install ./ --build="boost/*" -vvv` with some `tools.build:defines` and observe how no `define=...` is passed to `b2`.

### Logs

Important part:

```
[...]

PR #22679, opened 2024-02-05
boost: Honoring tools.build:defines

klausholstjacobsen:
boost/*

According to the docs [build_and_package](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/build_and_package.md) recipes should honor the configuration `tools.build:defines`

I have thus added defines to cflags and cxxflags in the recipe!

closes https://github.com/conan-io/conan-center-index/issues/22619

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -1143,6 +1143,9 @@ class BoostConan(ConanFile):
         if self._with_zstd:
             add_defines("zstd")
 
+        for define in self.conf.get("tools.build:defines", default=[], check_type=list):
+            flags.append(f"define={define}")
+
         if is_msvc(self):
             flags.append(f"runtime-link={'static' if is_msvc_static_runtime(self) else 'shared'}")
             flags.append(f"runtime-debugging={'on' if 'd' in msvc_runtime_flag(self) else 'off'}")


recipes/boost/all/conanfile.py
```diff
@@ -1336,6 +1336,11 @@ def create_library_config(deps_name, name):
         ldflags = " ".join(self.conf.get("tools.build:sharedlinkflags", default=[], check_type=list)) + " "
         asflags = buildenv_vars.get("ASFLAGS", "") + " "
 
+        defines = " ".join(f"-D{d}" for d in self.conf.get("tools.build:defines", default=[], check_type=list)) + " "
+        if defines.strip():
+            cflags += defines
+            cxxflags += defines
```
uilianries, line 1342: I'm not so sure, because b2 offers `define` as well: https://www.boost.org/doc/libs/1_84_0/tools/build/doc/html/index.html#bbv2.builtin.features.define

At same time, Conan config offers `tools.build:cxxflags` and `tools.build:cflags`

klausholstjacobsen: I have tried adding the defines with:
```
        for d in self.conf.get("tools.build:defines", default=[], check_type=list):
            contents += f'<define>{d} '
```
But it does not work. I have no clue about b2 and its config files, so I have no chance of debugging it.

The user.config.jam produced with the above looks like this:
```
using zlib : 1.2.13 : <include>"[hidden]/zlib5c9e3f01b88a0/p/include" <search>"[hidden]/.conan2/p/b/zlib5c9e3f01b88a0/p/lib" <name>z ;
using "gcc" :  :  "[hidden]/.conan2/p/sdk-x4fa5888332018/p/gcc" :
<define>MYDEFINE1 <define>MYDEFINE2   ;
```
and the subsequent b2 call:
```
b2 -q architecture=x86 address-model=64 abi=sysv --layout=system --user-config=aaa/s/src/tools/build/user-config.jam -sNO_ZLIB=0 -sNO_BZIP2=1 -sNO_LZMA=1 -sNO_ZSTD=1 boost.locale.icu=off --disable-icu boost.locale.iconv=off --disable-iconv threading=multi visibility=hidden link=static variant=release --with-atomic --with-chrono --with-container --with-context --with-contract --with-coroutine --with-date_time --with-exception --with-filesystem --with-graph --with-iostreams --with-math --with-program_options --with-random --with-regex --with-serialization --with-system --with-test --with-thread --with-timer --with-type_erasure --with-wave toolset=gcc cxxstd=14 define=_GLIBCXX_USE_CXX11_ABI=0 pch=on linkflags="" cxxflags="-fPIC" install --prefix=yyy -j16 --abbreviate-paths -d0 --debug-configuration --build-dir="xxx"
```

klausholstjacobsen: And by the way, the `tools.build:cxxflags` and `tools.build:cflags` are already propagated to b2, I just added defines in my original PR.

uilianries: Are cflags and cxxflags not enough to your case? I'm missing what you are really trying to achieve with this change. 🤔

uilianries: @klausholstjacobsen Thank you for explaining! So only to be sure, you need to pass something like `gcc -DQNX` right? If correct, then `tools.build.defines` is the correct way in fact. I just need to check why it didn't work with b2 defines .

klausholstjacobsen: That is absolutely correct!

stevenwdv: Why not reuse this?
https://github.com/conan-io/conan-center-index/blob/dec149b32f3a0670bf2395c0ccade27e7b760bca/recipes/boost/all/conanfile.py#L1074

uilianries: @stevenwdv Please, read the comment above: https://github.com/conan-io/conan-center-index/pull/22679#discussion_r1485671757

uilianries: @klausholstjacobsen Could please try the PR https://github.com/klausholstjacobsen/conan-center-index/pull/2 

I tested locally and worked for me. I tested in Mac and passed `-c "tools.build:defines=['_MSC_VER=191']"`, so it started activate all macros for Visual Studio and push new warning messages. I guess this is what you are looking for.

kudos to @stevenwdv

stevenwdv: Yeah I meant using `define=...` on the `b2` CLI instead of cflags/cxxflags or b2 config files

klausholstjacobsen: @stevenwdv  Implemented your suggestion! Thanks!

uilianries (approved):
LGTM!

franramirez688 (approved):

PR #22690, opened 2024-02-06
date: Add option to use manual tz db

Kaaml:

---

### DIFF
--- a/recipes/date/all/conanfile.py
+++ b/recipes/date/all/conanfile.py
@@ -24,6 +24,7 @@ class DateConan(ConanFile):
         "header_only": [True, False],
         "use_system_tz_db": [True, False],
         "use_tz_db_in_dot": [True, False],
+        "manual_tz_db": [True, False],
     }
     default_options = {
         "shared": False,
@@ -31,6 +32,7 @@ class DateConan(ConanFile):
         "header_only": False,
         "use_system_tz_db": False,
         "use_tz_db_in_dot": False,
+        "manual_tz_db": False,
     }
 
     def export_sources(self):
@@ -53,7 +55,7 @@ class DateConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        if not self.options.header_only and not self.options.use_system_tz_db:
+        if not self.options.header_only and not self.options.use_system_tz_db and not self.options.manual_tz_db:
             self.requires("libcurl/[>=7.78 <9]")
 
     def package_id(self):
@@ -72,6 +74,7 @@ class DateConan(ConanFile):
         tc.variables["ENABLE_DATE_TESTING"] = False
         tc.variables["USE_SYSTEM_TZ_DB"] = self.options.use_system_tz_db
         tc.variables["USE_TZ_DB_IN_DOT"] = self.options.use_tz_db_in_dot
+        tc.variables["MANUAL_TZ_DB "] = self.options.manual_tz_db
         tc.variables["BUILD_TZ_LIB"] = not self.options.header_only
         # workaround for clang 5 not having string_view
         if Version(self.version) >= "3.0.0" and self.settings.compiler == "clang" \
--- a/recipes/date/all/conanfile.py
+++ b/recipes/date/all/conanfile.py
@@ -14,6 +14,7 @@ class DateConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/HowardHinnant/date"
     description = "A date and time library based on the C++11/14/17 <chrono> header"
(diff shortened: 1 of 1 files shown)


danimtb:
Hi @Kaaml, thanks for your PR.

Could you explain a bit how does this work? I mean, setting this option to `True` in the recipe will lead to a different package, but, is this really true? Could you show how would a consumer of this recipe indicate the path to the manual DB? Thank you!

ericLemanissier (approved):

Kaaml:
> I detected other pull requests that are modifying date/all recipe:
> 
> * [[date] Add tz recipe as a dependency #22294](https://github.com/conan-io/conan-center-index/pull/22294)
> 
> This message is automatically generated by https://github.com/ericLemanissier/conan-center-conflicting-prs so don't hesitate to report issues/improvements there.

Hi @danimtb,

First of all, sorry for the long reply. Many users may have problem especially on windows when the date library checks for new version of IANA database and the end user doesn't have or want have 7zip in PATH variable. In my solution I put the flag to the library to block automatic download library. When this flag is set, users of the package have to remember to manually set location of IANA database.
`
date::set_install("/path/to/iana/database");
`
 I saw few issues/PR for solving problem like this but all have all were more complicated and has dependency to other packages. This change can be replaced with more complex solution when will be ready.

danimtb:
Hi @Kaaml and thanks for the reply.

I am not sure I totally understand the usage of the option. Do you mean that without this new option to set a manual path to the database, users have to do `date::set_install("/path/to/iana/database");` in their project?

Also, as pointed out in my previous comment, setting `manual_tz_db` to True/False, will lead to different packages of date, but I am not sure if the binary packages are actually different... If they aren't, then this option will need to be excluded from the pacakge id

Kaaml:
@danimtb 
> Hi @Kaaml and thanks for the reply.
> 
> I am not sure I totally understand the usage of the option. Do you mean that without this new option to set a manual path to the database, users have to do `date::set_install("/path/to/iana/database");` in their project?
> 

Currently when we use `date::set_install("/path/to/iana/database");` the **date** library just changes the path to the IANA database. If there is no IANA database under this path, or if the version is older than on the IANA website, the library will download the new version and try to extract it and replace the original version. 

If we use option `manual_tz_db `:
- autodownloading of IANA database will be disabled
- the `date::set_install("/path/to/iana/database");`  is required; Or you need to put IANA database in deafault path whic is declared in library ( usually something like this in Windows "c:\Users\username\Downloads" )

> Also, as pointed out in my previous comment, setting manual_tz_db to True/False, will lead to different packages of date, but I am not sure if the binary packages are actually different... If they aren't, then this option will need to be excluded from the pacakge id

Yes, they are diffrent. Some functions has excluded some parts of code responsible for automatic download and checking for new version of IANA.

More details
In [Date CMakeLists](https://github.com/HowardHinnant/date/blob/88a3b1512661decb78c60841f819a14eb200423c/CMakeLists.txt#L125)  we have defined `HAS_REMOTE_API=0` and `AUTO_DOWNLOAD=0` variables when we define `manual_tz_db`. Next in
(tz.cpp)[https://github.com/HowardHinnant/date/blob/88a3b1512661decb78c60841f819a14eb200423c/src/tz.cpp#L3569C5-L3569C18] file we have some #IF based on previously defined in CMakeLists.txt. The line 3569 is most important for me because it's doing auto download when file doesn't exist and also updating to new version when file exist. It's worth to notice that if we set `manual_tz_db` the curl library is also redundatn and would not be linked.

recipes/date/all/conanfile.py
```diff
@@ -24,13 +24,15 @@ class DateConan(ConanFile):
...
+        "manual_tz_db": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "header_only": False,
         "use_system_tz_db": False,
         "use_tz_db_in_dot": False,
+        "manual_tz_db": False,
     }
```
valgur, line 36: Please add an `options_description = { "use_system_tz_db": "...", ...}` attribute to the recipe. Their meaning of these options and how they interact is a bit vague otherwise.

I would also maybe call the new option `allow_tz_db_download` instead.

Kaaml: Done

uilianries:
Closing this PR as the author did not sign the CLA and the option is bit too much, as ConanCenterIndex has no registered issues related to a different path for IANA database.

I would suggest using the Conan configuration `tools.cmake.cmaketoolchain:extra_variables` instead, so you could add extra CMake defines without changing the recipe. For instance:

`-c 'tools.cmake.cmaketoolchain:extra_variables={"MANUAL_TZ_DB": "True"}'`

In case still having the same problem, please, re-open the PR so we can discuss about.

PR #22692, opened 2024-02-06
Set CMP0142 policy in package templates

Nekto89:
New behavior fixes warnings for Xcode generator in case of using CMake >= 3.25 and conan 2.

Without this policy CMake passes inexisting subdirectories to linker.
"ld: warning: directory not found for option '-L/Users/user/.conan2/p/b/expatbf806472793ba/p/lib/Release'"
https://cmake.org/cmake/help/latest/policy/CMP0142.html

---

### DIFF
--- a/docs/package_templates/autotools_package/all/test_package/CMakeLists.txt
+++ b/docs/package_templates/autotools_package/all/test_package/CMakeLists.txt
@@ -1,4 +1,9 @@
 cmake_minimum_required(VERSION 3.15)
+
+if(POLICY CMP0142)
+  cmake_policy(SET CMP0142 NEW)
+endif()
+
 project(test_package LANGUAGES C) # if the project is pure C
 # project(test_package LANGUAGES CXX) # if the project uses c++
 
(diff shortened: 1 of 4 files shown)


uilianries:
@Nekto89 Hello! Thank you for your suggestion. Have you talked to https://github.com/conan-io/conan/issues first? It sounds more a feature to be added in `CMakeToolchain` than a workaround to be added in CCI directly.

Nekto89:
> @Nekto89 Hello! Thank you for your suggestion. Have you talked to https://github.com/conan-io/conan/issues first? It sounds more a feature to be added in `CMakeToolchain` than a workaround to be added in CCI directly.

I did not. I've discovered that conan2 always adds lib directories to interface of cmake targets and there is no way to disable it. And Xcode adds one more subdirectory without this policy. I don't think that CMakeToolchain should control such CMake policies globally for user, because it may break some code.

uilianries:
@Nekto89 Let me check with Conan team, maybe are can find a way.

valgur:
Yeah, this would make more sense as a policy set by Conan automatically, since the layout of the package directory is managed by Conan anyway.

PR #22698, opened 2024-02-07
battery-embed: add recipe

toge:
Specify library name and version:  **battery-embed/***

## Is cmake module only package acceptable?

Let me first check if it is OK to create a recipe with only a cmake module.

Since ignition-cmake also provides a cmake module, I would like to achieve the same thing with battery-embed.
https://github.com/conan-io/conan-center-index/tree/master/recipes/ignition-cmake

Is this recipe adequate in cci policy?

## patch description

battery-embed currently has the following two problems:

1. example program is always built
2. compile error with gcc13 due to lack of cstdint

Patch files try to solve these.

## reason two versions are provided

1.2.x requires std::jthread and std::stop_token.
Those are not provided on apple-clang (yet).

---


uilianries:
> compile error with gcc13 due to lack of cstdint

@toge Do you have a log or an issue/PR in the upstream to confirm about this behavior? I don't have GCC13 installed here right now to validate it 😞 

> example program is always built

It could be suggested to the upstream to use your patch instead, so users are free to decide about it.

toge:
> It could be suggested to the upstream to use your patch instead, so users are free to decide about it.

The issue about it has been created on upstream.
https://github.com/batterycenter/embed/issues/2

I'm watching it.

> Do you have a log or an issue/PR in the upstream to confirm about this behavior? I don't have GCC13 installed here right now to validate it 😞

Off course!
Here are:

<details>
======== Exporting recipe to the cache ========
WARN: Name containing special chars is discouraged 'battery-embed'
battery-embed/1.2.8: Exporting package recipe: /home/toge/src/conan-center-index/recipes/battery-embed/all/conanfile.py
battery-embed/1.2.8: exports: File 'conandata.yml' found. Exporting it...
battery-embed/1.2.8: Calling export_sources()
battery-embed/1.2.8: Copied 1 '.yml' file: conandata.yml
battery-embed/1.2.8: Copied 1 '.py' file: conanfile.py
battery-embed/1.2.8: Copied 2 '.patch' files: 0001-disable-examples.patch, 0002-include-cstdint.patch
battery-embed/1.2.8: Exported to cache folder: /home/toge/.conan2/p/battebcc91ef1dfa80/e
battery-embed/1.2.8: Exported: battery-embed/1.2.8#dc406be40a1fbe9305318a815f8cd372 (2024-02-09 09:48:05 UTC)

Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=13
os=Linux

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=13
os=Linux


Graph root
    cli
Build requirements
    battery-embed/1.2.8#dc406be40a1fbe9305318a815f8cd372 - Cache
    cmake/3.27.4#a7e78418b024dccacccc887f049f47ed - Cache
Resolved version ranges
    cmake/[>=3.21 <4]: cmake/3.27.4

battery-embed/1.2.8: Compatible package ID 9a4eb3c8701508aa9458b1a73d0633783ecc2270 equal to the default package ID: Skipping it.
Build requirements
    battery-embed/1.2.8#dc406be40a1fbe9305318a815f8cd372:9a4eb3c8701508aa9458b1a73d0633783ecc2270 - Build
    cmake/3.27.4#a7e78418b024dccacccc887f049f47ed:63fead0844576fc02943e16909f08fcdddd6f44b#fac7109ddc535a744445b6e1c095129c - Cache

cmake/3.27.4: Already installed! (1 of 2)
cmake/3.27.4: Appending PATH environment variable: /home/toge/.conan2/p/cmake9d32db761075c/p/bin
battery-embed/1.2.8: Calling source() in /home/toge/.conan2/p/battebcc91ef1dfa80/s

-------- Installing package battery-embed/1.2.8 (2 of 2) --------
battery-embed/1.2.8: Building from source
battery-embed/1.2.8: Package battery-embed/1.2.8:9a4eb3c8701508aa9458b1a73d0633783ecc2270
battery-embed/1.2.8: Copying sources to build folder
battery-embed/1.2.8: Building your package in /home/toge/.conan2/p/b/battefe07bce641e42/b
battery-embed/1.2.8: Generating aggregated env files
battery-embed/1.2.8: Generated aggregated env files: ['conanbuild.sh', 'conanrun.sh']
battery-embed/1.2.8: Calling build()
battery-embed/1.2.8: Apply patch (conan): disable building examples
battery-embed/1.2.8: Apply patch (portability): include cstdint for gcc 13 later
battery-embed/1.2.8: Package '9a4eb3c8701508aa9458b1a73d0633783ecc2270' built
battery-embed/1.2.8: Build folder /home/toge/.conan2/p/b/battefe07bce641e42/b
battery-embed/1.2.8: Generating the package
battery-embed/1.2.8: Packaging in folder /home/toge/.conan2/p/b/battefe07bce641e42/p
battery-embed/1.2.8: Calling package()
battery-embed/1.2.8: package(): Packaged 1 file: LICENSE
battery-embed/1.2.8: package(): Packaged 1 '.txt' file: CMakeLists.txt
battery-embed/1.2.8: Created package revision 383d669dd11332c6b6339d753c0fb9eb
battery-embed/1.2.8: Package '9a4eb3c8701508aa9458b1a73d0633783ecc2270' created
battery-embed/1.2.8: Full package reference: battery-embed/1.2.8#dc406be40a1fbe9305318a815f8cd372:9a4eb3c8701508aa9458b1a73d0633783ecc2270#383d669dd11332c6b6339d753c0fb9eb
battery-embed/1.2.8: Package folder /home/toge/.conan2/p/b/battefe07bce641e42/p
WARN: deprecated: Usage of deprecated Conan 1.X features that will be removed in Conan 2.X:
WARN: deprecated:     'env_info' used in: cmake/3.27.4

======== Launching test_package ========

Graph root
    battery-embed/1.2.8 (test package): /home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/conanfile.py
Requirements
    battery-embed/1.2.8#dc406be40a1fbe9305318a815f8cd372 - Cache
Build requirements
    cmake/3.27.4#a7e78418b024dccacccc887f049f47ed - Cache

Requirements
    battery-embed/1.2.8#dc406be40a1fbe9305318a815f8cd372:9a4eb3c8701508aa9458b1a73d0633783ecc2270#383d669dd11332c6b6339d753c0fb9eb - Cache
Build requirements
Skipped binaries
    cmake/3.27.4

battery-embed/1.2.8: Already installed! (1 of 1)

======== Testing the package ========
Removing previously existing 'test_package' build folder: /home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release
battery-embed/1.2.8 (test package): Test package build: build/gcc-13-x86_64-20-release
battery-embed/1.2.8 (test package): Test package build folder: /home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release
battery-embed/1.2.8 (test package): Writing generators to /home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release/generators
battery-embed/1.2.8 (test package): Generator 'VirtualRunEnv' calling 'generate()'
battery-embed/1.2.8 (test package): Generator 'CMakeToolchain' calling 'generate()'
battery-embed/1.2.8 (test package): CMakeToolchain generated: conan_toolchain.cmake
battery-embed/1.2.8 (test package): CMakeToolchain generated: CMakePresets.json
battery-embed/1.2.8 (test package): CMakeToolchain generated: ../../../CMakeUserPresets.json
battery-embed/1.2.8 (test package): Generator 'CMakeDeps' calling 'generate()'
battery-embed/1.2.8 (test package): CMakeDeps necessary find_package() and targets for your CMakeLists.txt
    find_package(battery-embed)
    target_link_libraries(... battery-embed::battery-embed)
battery-embed/1.2.8 (test package): Generating aggregated env files
battery-embed/1.2.8 (test package): Generated aggregated env files: ['conanrun.sh', 'conanbuild.sh']

======== Testing the package: Building ========
battery-embed/1.2.8 (test package): Calling build()
battery-embed/1.2.8 (test package): Running CMake.configure()
battery-embed/1.2.8 (test package): RUN: cmake -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE="/home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="/home/toge/src/conan-center-index/recipes/battery-embed/all/test_package" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "/home/toge/src/conan-center-index/recipes/battery-embed/all/test_package"
-- Using Conan toolchain: /home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release/generators/conan_toolchain.cmake
-- Conan toolchain: C++ Standard 20 with extensions OFF
-- The CXX compiler identification is GNU 13.2.1
-- Conan: Target declared 'battery-embed::battery-embed'
-- Conan: Including build module from '/home/toge/.conan2/p/b/battefe07bce641e42/p/lib/cmake/battery-embed/CMakeLists.txt'
-- Configuring done (0.2s)
-- Generating done (0.0s)
-- Build files have been written to: /home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release

battery-embed/1.2.8 (test package): Running CMake.build()
battery-embed/1.2.8 (test package): RUN: cmake --build "/home/toge/src/conan-center-index/recipes/battery-embed/all/test_package/build/gcc-13-x86_64-20-release" -- -j12
[ 20%] Generating embed/autogen/test_package_cpp.cpp
[ 40%] Building CXX object CMakeFiles/test_package.dir/test_package.cpp.o
[ 60%] Building CXX object CMakeFiles/test_package.dir/embed/autogen/test_package_cpp.cpp.o
[ 80%] Building CXX object CMakeFiles/test_package.dir/embed/embed_impl.cpp.o
[100%] Linking CXX executable test_package
[100%] Built target test_package


battery-embed/1.2.8 (test package): Running test()
battery-embed/1.2.8 (test package): RUN: ./test_package
#include <cstdlib>
#include <iostream>
#include <cstdint>

#include "battery/embed.hpp"

int main(void) {
    std::cout << b::embed<"test_package.cpp">() << std::endl;
    return EXIT_SUCCESS;
}
</details>

valgur (approved):
Looks like a perfectly reasonable solution to me. Thanks!

HerrNamenlos123:
Hi! I am the author of battery-embed.

Interesting that people tried to get my library into Conan, without me even knowing!
I apprechiate the effort, however. I would have been more than willing to make upstream changes if they are for the greater good.

Are there still any issues remaining? I added `<cstdint>` upstream, no issue with that. Also, examples are only built if battery-embed is the top level CMake module AND `BUILD_EXAMPLES` is set (which it is, if it's the top-level module).

Is there still an issue with `std::jthread`? It's only used for convenience. If necessary, it can easily be changed to a normal `std::thread` and `std::atomic<bool>` instead of `std::stop_token`...

RubenRBS:
Hi @HerrNamenlos123 I'll get back to you with the status of the PR later today, just wanted to drop by to thank you for offering to help, really appreciated :)

HerrNamenlos123:
@toge I really apprechiate your efforts. Could you explain to me why you want to add 1.2.12 AND 1.1.0? I don't see any reason to use 1.1.0? 1.0.0 is for C++14, but does 1.1.0 give you any advantage over the latest version? If so, can we re-implement it in the latest version?

Also, I don't quite understand why you need to patch the CMakeLists to disable the examples. Can't you define -DBUILD_EXAMPLES=OFF in the conanfile? Isn't that the whole point?

(In case you didn't know, the option() you patched is just defaulted with PROJECT_IS_TOP_LEVEL, meaning the default is based on how you build it. The point of the option is that you can still define it manually and override the default)

I think there shouldn't be any reason to patch anything and if so, i can fix it upstream to keep everything clean...

toge:
@HerrNamenlos123 
Thank you for your comment!
I am glad to discuss the way to package your product.

> Could you explain to me why you want to add 1.2.12 AND 1.1.1? I don't see any reason to use 1.1.1?

The reason to provide 1.1.1 is supporting apple-clang.
As you know, 1.2.x uses std::jthread.
It seems that std::jthread is not supported by apple-clang, and I get the following error message.

```
embed/embed_impl.cpp:51:10: error: no type named 'jthread' in namespace 'std'; did you mean 'thread'?
    std::jthread embeddedFileWatcherThread = std::jthread([](std::stop_token stopToken) {
    ~~~~~^~~~~~~
         thread
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX14.4.sdk/usr/include/c++/v1/__thread/thread.h:164:33: note: 'thread' declared here
class _LIBCPP_EXPORTED_FROM_ABI thread
                                ^
embed/embed_impl.cpp:51:67: error: no type named 'stop_token' in namespace 'std'
    std::jthread embeddedFileWatcherThread = std::jthread([](std::stop_token stopToken) {
                                                             ~~~~~^
embed/embed_impl.cpp:51:51: error: no member named 'jthread' in namespace 'std'
    std::jthread embeddedFileWatcherThread = std::jthread([](std::stop_token stopToken) {
                                             ~~~~~^
embed/embed_impl.cpp:55:40: error: variable '__begin2' cannot be implicitly captured in a lambda with no capture-default specified
            for (auto& [key, filedata] : embeddedFiles) {
                                       ^
```

> Also, I don't quite understand why you need to patch the CMakeLists to disable the examples. Can't you define -DBUILD_EXAMPLES=OFF in the conanfile? Isn't that the whole point?

The reason for patching instead of defining BUILD_EXAMPLES is portability.
I think `BUILD_EXAMPLES` is a popular name and I am afraid of conflicts.

> I think there shouldn't be any reason to patch anything and if so, i can fix it upstream to keep everything clean...

I agree with you.
To solve these problems, I am planning to split CMakeLists.txt into two files.

1. CMakeLists.txt : project, BUILD_EXAMPLES, add_subdirectory.
2. battery_embed.cmake : functions and dependent variables.

Does this fit your policy?

HerrNamenlos123:
Ah, i see. I am not a fan of uploading an old version. give me a bit of time and I will replace std::jthread with a normal thread, then apple-clang can also use the latest version. jthread is not a strict requirement but a convenience.

I see, you package the entire CMake file in a Conan package, Of course. Sorry about that.

I don't exactly know what your proposed solution is. I want to keep the upstream in a single file, for ease-of-use. Do you want to split the file based on line numbers into two separate files for Conan? Not really a fan of that.

However: What if I add an additional flag to the upstream, called BATTERY_DISABLE_EXAMPLES or something like that, that disables the examples definitely? In Conan, since it doesn't need to be a single file, you can add a second file (FindBatteryEmbed or however you need to call it), that defines BATTERY_DISABLE_EXAMPLES and calls add_subdirectory for a subdirectory that contains the main CMake file?

That way not a single patch is needed because BatteryEmbed is in a subdirectory like its supposed to, meaning the project() call does not need to be patched, and the wrapper cmake file from Conan can disable the examples, it would be a hand-written 3 line file.

HerrNamenlos123:
Another, much simpler idea: What if I rename `BUILD_EXAMPLES` to `BATTERY_EMBED_BUILD_EXAMPLES`? I don't see any downsides and all our problems would vanish immediately i think, right?

I don't think anyone needs to set this explicitly because it is cleverly defaulted and if they do, it is documented.

Keeping it in a subdirectory in Conan means we don't even need to disable the examples because it is by default disabled due to being in a subdirectory and not the top level project.

So my suggestion is: I rename the flag and remove std::jthread and you keep the CMakeLists in a subdirectory in the conan archive and add it using a proxy cmake file.

Is that fine for you?

HerrNamenlos123:
I released version 1.2.13. I renamed the flag to `BATTERY_EMBED_BUILD_EXAMPLES` and removed the `std::jthread` dependency. 

It should work with `apple-clang` and all issues with disabling examples should also be gone. No need to upload two different versions.

In the Conan package, create a proxy CMake file and call `add_subdirectory` for battery-embed, which is in a subdirectory. This will make everything magically work. All patches can be removed.

HerrNamenlos123:
@toge and another note: You named the CMake target `battery-embed::battery-embed`. This somehow misses the point since battery-embed is actually supposed to be part of the battery ecosystem (which hasn't grown significantly, but still), so the idea would have been that the main module is `battery::battery`, the ui module is `battery::ui`, and the embed part would be `battery::embed`. It's only called BATTERY-embed and not just embed because it's part of the battery ecosystem. I don't think there should be an issue with naming it `battery::embed` in CMake. That's how it was originally planned and if i continue to develop the battery ecosystem, this will be the way to go.

Now, if you want to keep it `battery-embed::battery-embed` so it's easier to remember because it's packagename-packagename, what about providing both aliases?

toge:
@HerrNamenlos123 
Thank you for your great help!
I agree with your solutions.

I am going to update this PR followings:
- update 1.2.12 to 1.2.13 and remove 1.1.1
- remove patch files and prepare proxy CMakeFiles.txt
- rename target name battery-embed::battery-embed to battery::embed (battery-embed::battery-embed is not needed for me too.)

HerrNamenlos123:
@toge I am terribly sorry but there still seem to be a few errors, that occur when the b_embed function is called in a subdirectory. I fixed the other errors already and I am still investigating why this happens. When i fixed all of them, I will release another version.

toge:
@HerrNamenlos123 
Sorry, it's my fault. 
Forgetting to link `battery::embed` to link pthead on Linux.
Now it works fine.

There is another compilation errors on apple-clang due to `unordered_map` being used without includes.
After applying a temporary patch, CI now passes.

I have created a PR about this patch, could you please check it?
https://github.com/batterycenter/embed/pull/6

HerrNamenlos123:
@toge I finally managed to find a good solution for working across subdirectories (consult Readme). In the process I also fixed other issues as well as unordered_map.

I think it should work fine now. Your PR also looks fine (minus the unordered_map patch), just a minor thing: In `embed.cmake` proxy file, you specified a build directory called `build`. I think it's not bad to specify it since the Conan package might actually be outside of the binary dir, but would you consider renaming it to something like `embed-build` or `embed` or `battery-embed-build`? At the end of the day it would create a folder called `build` within the build directory which might not be uncommon and might create an unnecessary conflict.

Regarding pthread, did you have to manually link it or create a patch? If so, should battery::embed by default link to pthread on Linux? The thing is, I tried it on Linux multiple times, but I never had to do anything with pthread...

toge:
@HerrNamenlos123 
Thank you for your quick response!
I'm going to follow your points.

> Regarding pthread, did you have to manually link it or create a patch? 

Oh, I missed new battery::embed don't require thread features.
I will fix also.

HerrNamenlos123:
Well, i mean i still use `std::thread`, but not `std::jthread` anymore. 

When packaged, the consuming package would still be able to set `B_PRODUCTION_MODE`, right? I was thinking if `B_PRODUCTION_MODE` should also be an option in the package. But I think it shouldn't, as the consuming package should set it before calling (find_package), right?

toge:
@HerrNamenlos123 
I misunderstood your comments!
Linking thread library(if needed) is mandatory for using `<thread>`.
So your suggestion ("should battery::embed by default link to pthread on Linux?") is more elegant than me.e
I created another patch for 1.2.14 to link thread library. (it's temporary patch also.)

Let's see the result of CI.

HerrNamenlos123:
I am going to add the Threads package in upstream. I think it shouldn't create any issues even if it's not strictly needed.

However I think you misunderstood me regarding `B_PRODUCTION_MODE`. If i understand correctly, you added the define to the Conan package, which means it automatically defines it and all consuming packages would inherit production mode unconditionally. Right? This is not the desired behavior as users need to be able to switch it on and off in their projects, ideally without overriding and rebuilding Conan packages.

I only meant that someone needs to test if the final Conan package, if it allows users to set the flag manually and that all Dev/Prod features work as expected in the consuming package.

HerrNamenlos123:
Alright, i did a few changes again (Mainly a bit of internal refactoring/cleanup), and i finally added a proper CI test suite.
From my side everything compiles fine.

Also, i implemented some additional CMake options for settings Production mode, and the entire Thread support is now optional via `B_PRODUCTION_MODE`. That means in theory, battery::embed could also be used for embedded bare-metal programming (only in production mode).

As said, the prod mode flags are now proper CMake options, but they shouldn't be in Conan. They should be controlled entirely by the consuming package.

toge:
@HerrNamenlos123 
Thanks a lot for comments!
I fixed my misunderstood definition and updated to 1.2.15.

HerrNamenlos123:
Alright, i think from my side it's all good now. It can be merged IMO

recipes/battery-embed/all/conanfile.py
```diff
@@ -0,0 +1,75 @@
...
+            "gcc": "11",
+            "clang": "12",
+            "apple-clang": "13",
+            "Visual Studio": "16",
+            "msvc": "192",
+        }
+
+    def export_sources(self):
+        copy(self, "embed.cmake", src=self.recipe_folder, dst=self.export_sources_folder)
+        export_conandata_patches(self)
```
valgur, line 37:
```suggestion
```

recipes/battery-embed/all/conanfile.py
```diff
@@ -0,0 +1,75 @@
...
+    def package_id(self):
+        del self.info.settings.arch
+        del self.info.settings.compiler
+        del self.info.settings.build_type
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def build(self):
+        apply_conandata_patches(self)
```
valgur, line 60:
```suggestion
```

recipes/battery-embed/all/conanfile.py
```diff
@@ -0,0 +1,75 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, get, export_conandata_patches, apply_conandata_patches
```
valgur, line 4:
```suggestion
from conan.tools.files import copy, get
```

valgur (approved):

toge:
@valgur 
Thanks!

HerrNamenlos123:
@toge I took a look at the current package and I propose 2 more changes:

1. Update to the latest Version of 1.2.19, as some important bug fixes were made. 

2. In the Test package, do not include cstdint and cstdlib, because both should be included in embed.hpp itself. Their absence is a bug and the test package should point out the bug and not hide it.

If any other things pop up in the latest version, feel free to immediately create an issue or a PR.

danimtb (approved):

recipes/battery-embed/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def package_id(self):
+        del self.info.settings.arch
+        del self.info.settings.compiler
+        del self.info.settings.build_type
```
AbrilRBS, line 53: I'm confused. Why is this deleting every setting _but_ the os? As far as I can see, if the build() method only has 1 `copy()` call, where's the variance there depending on the os?

recipes/battery-embed/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def package(self):
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        copy(self, "embed.cmake", self.recipe_folder, os.path.join(self.package_folder, "lib", "cmake"))
+        copy(self, "CMakeLists.txt", self.source_folder, os.path.join(self.package_folder, "lib", "cmake", "battery-embed"))
+
+    def package_info(self):
+        self.cpp_info.libdirs = []
+        self.cpp_info.bindirs = []
+        self.cpp_info.includedirs = []
```
AbrilRBS, line 69: This confuses me as to why the test package is even working. If we're setting `includedirs = []`, then how come the test package can do `#include "battery/embed.hpp"` just fine? At what point is then Conan filling that?

AbrilRBS (commented):
Thanks! I have some questions that are confusiong me a bit, otherwise looks good

AbrilRBS (commented):
Thanks! I have some questions that are confusiong me a bit, otherwise looks good

recipes/battery-embed/all/conanfile.py
HerrNamenlos123: I can answer that, that is because embed.hpp is dynamically generated by CMake and does not actually exist until you compile the project. Therefore, by calling b_embed() in CMake, all necessary files are auto-gemerated and their folder is included by CMake. embed.hpp resides in an 'autogen' folder inside the build directory and only exists after the first build.

HerrNamenlos123: It is constructed in a way that, when going from a non-battery-embed project to one that uses it, the only thing you have to change is include the CMake file and call b_embed(). Everything else is done for you, so no modification to your build system is necessary.

recipes/battery-embed/all/conanfile.py
danimtb: You are right, there is no reason to have different packages per OS apparently
```suggestion
        del self.info.settings.build_type
        del self.info.settings.os
```
(tried removing the `settings` attribute but it's required for the `cmake_layout()`, so no luck)

valgur (approved):

danimtb (approved):

recipes/battery-embed/all/conanfile.py
AbrilRBS: Thanks for the insight, makes quite a bit of sense!

AbrilRBS (approved):
Thanks!

HerrNamenlos123:
I am terribly sorry for only noticing now, but unfortunately I found another mistake. The Test package links to battery::embed, but this is not supposed to happen. Like with the include paths, this is meant to happen automatically via `b_embed()`.

The package itself works fine, so it's absolutely not critical, but if we want to be pedantic and ensure correct testing, it should be fixed at some point. Sorry for not noticing earlier, I must have skipped over it.

PR #22701, opened 2024-02-07
stale: enable cache

ericLemanissier:
only 30 PRs are processed per day, so cache is needed to keep state and continue from the last PRs of the last days.

fixes error https://github.com/conan-io/conan-center-index/actions/runs/7810903828/job/21305055330#step:2:1032

```
state: persisting info about 250 issue(s)
Warning: Error delete _state: [403] Resource not accessible by integration
/usr/bin/tar --posix -cf cache.tzst --exclude cache.tzst -P -C /home/runner/work/conan-center-index/conan-center-index --files-from manifest.txt --use-compress-program zstdmt
Failed to save: Unable to reserve cache with key _state, another job may be creating this cache. More details: Cache already exists. Scope: refs/heads/master, Key: _state, Version: fa41d75081481069cfb6b92a5f83a94c6e06ef3ab2e6b762649ac5f86f46153f
```

cf https://github.com/actions/stale/issues/1090#issuecomment-1710940734

---

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -18,6 +18,7 @@ jobs:
     permissions:
       pull-requests: write
       issues: read
+      actions: write
 
 
     steps:


ericLemanissier:
AFAIU, the error happens on this [line](https://github.com/actions/stale/blob/3f3b0175e8c66fb49b9a6d5a0cd1f8436d4c3ab6/src/classes/state/state-cache-storage.ts#L53) because the `DELETE` verb on an action cache needs `actions:write` to use this endpoint : https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28#delete-github-actions-caches-for-a-repository-using-a-cache-key

uilianries (commented):
I'm not happy with this change, because it will give a whole write access. A second option would be running all PRs, but we have +600 now and it would reach out Github API rate limit.

On the other hand, this is an official Action maintained by Github, so I only fell safe because of this. 

Still, I prefer waiting for https://github.com/actions/stale/issues/1133, because more projects are affected because of this problem.

uilianries:
@ericLemanissier Yes, I read the documentation. Still, the author is aware about and more repositories are being affected. I would prefer waiting for a real fix than trying a workaround which would give more permissions now. Thank you for giving more context about the situation.

Issue #22702, opened 2024-02-07
[package] liburing/2.4: Compilation error when cross-compiling


guillaume-michel:
### Description

When trying to cross compile `liburing` for `aarch64` from an `x86_64` machine, I get the following error:
```
-------- Installing package liburing/2.4 (27 of 43) --------
liburing/2.4: Building from source
liburing/2.4: Package liburing/2.4:bb79466d5dce565202f7e1d7ab965454dc5b55b6
liburing/2.4: Copying sources to build folder
liburing/2.4: Building your package in /github/home/.conan2/p/b/libura85eaabeb1392/b
liburing/2.4: Calling generate()
liburing/2.4: Generators folder: /github/home/.conan2/p/b/libura85eaabeb1392/b/build-relwithdebinfo/conan
liburing/2.4: Generating aggregated env files
liburing/2.4: Generated aggregated env files: ['conanbuild.sh', 'conanrun.sh']
liburing/2.4: Calling build()
liburing/2.4: apply_conandata_patches(): No patches defined in conandata
liburing/2.4: Calling:
 > "/github/home/.conan2/p/b/libura85eaabeb1392/b/src/configure" --prefix=/ '--libdir=${prefix}/lib' '--includedir=${prefix}/include' --host=aarch64-linux-gnu --build=x86_64-linux-gnu '--libdevdir=${prefix}/lib' 
liburing/2.4: RUN: "/github/home/.conan2/p/b/libura85eaabeb1392/b/src/configure" --prefix=/ '--libdir=${prefix}/lib' '--includedir=${prefix}/include' --host=aarch64-linux-gnu --build=x86_64-linux-gnu '--libdevdir=${prefix}/lib' 
ERROR: unknown option --host=aarch64-linux-gnu
Try '/github/home/.conan2/p/b/libura85eaabeb1392/b/src/configure --help' for more information

liburing/2.4: ERROR: 
Package 'bb79466d5dce565202f7e1d7ab965454dc5b55b6' build failed
liburing/2.4: WARN: Build folder /github/home/.conan2/p/b/libura85eaabeb1392/b/build-relwithdebinfo
ERROR: liburing/2.4: Error in build() method, line 84
	at.configure()
	ConanException: Error 1 while executing
```

[...]

PR #22704, opened 2024-02-07
liburing: Fix cross-compilation and added version 2.5

guillaume-michel:
Specify library name and version:  **liburing/2.5**

This PR fixes the cross-compilation issue encountered in #22702.

It also adds support for liburing 2.5

---

- [X ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ X] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ X] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.


recipes/liburing/all/patches/disable_samples_and_tests.patch
```diff
@@ -0,0 +1,17 @@
...
+ default: all
+ 
+ all:
+-	@$(MAKE) -C src
+-	@$(MAKE) -C test
+-	@$(MAKE) -C examples
++	@$(MAKE) -C src liburing.a
++	#@$(MAKE) -C test
++	#@$(MAKE) -C examples
+
```
Ahajha, line 15: Minor: For a cleaner diff, could just remove the two bottom lines

recipes/liburing/all/conanfile.py
```diff
@@ -23,11 +23,13 @@ class Liburing(ConanFile):
         "fPIC": [True, False],
         "shared": [True, False],
         "with_libc": [True, False],
+        "use_libc": [True, False],
     }
     default_options = {
         "fPIC": True,
         "shared": False,
         "with_libc": True,
+        "use_libc": True,
```
Ahajha, line 32: It seems strange to have two options that (fundamentally) do the same thing but for different versions. I think instead of adding another option, just keep using `with_libc`, even if the name is slightly off.

recipes/liburing/all/conanfile.py
```diff
@@ -40,8 +42,11 @@ def requirements(self):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
-        if Version(self.version) < "2.2":
+        if Version(self.version) < "2.2" or Version(self.version) >= "2.5":
             del self.options.with_libc
+        if Version(self.version) < "2.5":
+            del self.options.use_libc
+
```
Ahajha, line 49: Assuming the above comment is applied, this can be reverted

recipes/liburing/all/conanfile.py
```diff
@@ -66,6 +71,9 @@ def generate(self):
         tc = AutotoolsToolchain(self)
         tc.update_configure_args({
             "--nolibc": None if self.options.get_safe("with_libc", default=True) else "",
+            "--use-libc": "" if self.options.get_safe("use_libc", default=False) else None,
```
Ahajha, line 74: And I think this would have to be extracted to (untested):
```py
if Version(self.version) >= "2.5":
    if self.options.with_libc:
        tc.configure_args.append("--use-libc")
elif Version(self.version >= "2.2":
    if not self.options.with_libc:
        tc.configure_args.append("--nolibc")
```

recipes/liburing/all/patches/disable_samples_and_tests.patch
```diff
@@ -0,0 +1,17 @@
...
+--- a/Makefile
++++ b/Makefile
+@@ -7,9 +7,9 @@ INSTALL=install
+ default: all
+ 
+ all:
+-	@$(MAKE) -C src
+-	@$(MAKE) -C test
+-	@$(MAKE) -C examples
++	@$(MAKE) -C src liburing.a
```
Ahajha, line 12: Why does this unconditionally build a static liburing, is it used to make the shared version if applicable? I would expect `make src` to do the correct thing.

recipes/liburing/all/patches/disable_samples_and_tests.patch
guillaume-michel: fixed

recipes/liburing/all/conanfile.py
guillaume-michel: kept `with_libc` and removed `use_libc`

recipes/liburing/all/conanfile.py
guillaume-michel: done

recipes/liburing/all/conanfile.py
guillaume-michel: done

recipes/liburing/all/patches/disable_samples_and_tests.patch
guillaume-michel: This was a mistake from my part.
 liburing Makefile builds unconditionnally both static and shared library. The recipe then cleanup in the package phase

recipes/liburing/all/patches/liburing-2.4-disable_samples_and_tests.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/Makefile b/Makefile
+index 73d021c..74aa8d7 100644
+--- a/Makefile
++++ b/Makefile
+@@ -8,8 +8,6 @@ default: all
+ 
+ all:
+ 	@$(MAKE) -C src
+-	@$(MAKE) -C test
+-	@$(MAKE) -C examples
```
Ahajha, line 10: What's the difference between the 2.4 patch and the 2.2 one? Not seeing a difference.

Also, this might be an okay use for `replace_in_file` instead of patching, since the changes are fairly straightforward and only differ in context.

Ahajha (approved):
Looks good, just one minor thing that can be ignored.

recipes/liburing/all/patches/liburing-2.4-disable_samples_and_tests.patch
guillaume-michel: At first, I wanted to have a single patch to remove tests and examples build. The patch applies correctly on all versions on my side but for some reason, it fails in the CI.
I then decided to produce a patch for each version and remove duplicates (md5sum).
I don't know about `replace_in_file`. I will have a look.

Ahajha: Would be essentially something like `replace_in_file(self, os.path.join(self.source_folder, "Makefile"), "@$(MAKE) -C test", "")`

RubenRBS (approved):

toge (approved):

PR #22711, opened 2024-02-08
tcl: Un-vendor sqlite, properly fix apple shared install names

Ahajha:
Specify library name and version:  **tcl/all**

Tcl vendors sqlite3 for an extension, and it seems this had gone unnoticed until now. This also adds an option to remove sqlite entirely in case users don't want this.

Tcl has a lot of sub-directories with libraries in the package folder, and the `fix_apple_shared_install_name` tool wasn't fixing them. Adding the correct folders to `self.cpp.package.libdirs` allows it to fix everything.

Blocks https://github.com/conan-io/conan-center-index/pull/21387

---

### DIFF
--- a/recipes/tcl/all/conanfile.py
+++ b/recipes/tcl/all/conanfile.py
@@ -77,6 +77,7 @@ class TclConan(ConanFile):
     def build_requirements(self):
         if self._settings_build.os == "Windows" and not is_msvc(self):
             self.win_bash = True
+            self.build_requires("autoconf/2.71")
             if not self.conf.get("tools.microsoft.bash:path", check_type=str):
                 self.tool_requires("msys2/cci.latest")
 
--- a/recipes/tcl/all/conanfile.py
+++ b/recipes/tcl/all/conanfile.py
@@ -55,6 +55,16 @@ class TclConan(ConanFile):
         # source folder must be a sub-directory of the build folder
         self.folders.build = "."
 
+        # In order for fix_apple_shared_install_name to work, it needs to know the package layout precisely.
+        # Unfortunately this has to be defined here, before we ever fetch the sources, so we have to do this maually.
+        extra_folders = {
+            "8.6.13": ["itcl4.2.3", "sqlite3.40.0",   "tdbc1.1.5", "tdbcpostgres1.1.5", "thread2.8.8", "tdbcmysql1.1.5", "tdbcodbc1.1.5"],
+            "8.6.11": ["itcl4.2.1", "sqlite3.34.0",   "tdbc1.1.2", "tdbcpostgres1.1.2", "thread2.8.6", "tdbcmysql1.1.2", "tdbcodbc1.1.2"],
+            "8.6.10": ["itcl4.2.0", "sqlite3.30.1.2", "tdbc1.1.1", "tdbcpostgres1.1.1", "thread2.8.5", "tdbcmysql1.1.1", "tdbcodbc1.1.1"],
+        }
+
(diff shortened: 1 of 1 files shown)


RubenRBS (commented):
I think this works, but will delay approving until tomorrow, @uilianries and I want to use this PR for some internal training, sorry about the delay :)

recipes/tcl/all/conanfile.py
```diff
@@ -98,6 +120,26 @@ def yes_no(v): return "yes" if v else "no"
...
+            if self.options.with_sqlite:
+                # tdbcsqlite is the wrapper package for tcl, don't modify it
+                if "sqlite" in folder and "tdbc" not in folder:
+                    rm(self, "*.c", os.path.join(self.source_folder, "pkgs", folder, "compat"), recursive=True)
+                    rm(self, "*.h", os.path.join(self.source_folder, "pkgs", folder, "compat"), recursive=True)
+
+                    # Don't override SQLITE_* variables, causes link errors
+                    if Version(self.version) < "8.6.13":
+                        replace_in_file(self, os.path.join(self.source_folder, "pkgs", folder, "Makefile.in"), "-DSQLITE_API=MODULE_SCOPE", "")
+                    replace_in_file(self, os.path.join(self.source_folder, "pkgs", folder, "Makefile.in"), "-DSQLITE_EXTERN=", "")
```
RubenRBS, line 135: Can you give me some insight into these changes? I haven't been able to find the compilation logs for the failed case. Thanks :)

RubenRBS (commented):
Thanks a lot! I have a question regarding the compilation issues.

Also, is there a reason we can't accept the vendorization? Most (if not all) package managers seem to also vendorize sqlite for this library. What is exactly the blocker here? (Did you get a chance to report this to upstream so they can unvendorize it by themselves?)
Thanks :)

Ahajha:
> Thanks a lot! I have a question regarding the compilation issues.
> 
> Also, is there a reason we can't accept the vendorization? Most (if not all) package managers seem to also vendorize sqlite for this library. What is exactly the blocker here? (Did you get a chance to report this to upstream so they can unvendorize it by themselves?) Thanks :)

It wasn't specifically a problem, but I figured it shouldn't be vendored on principle. Wasn't aware that this was "allowed" to some extent.

I doubt they'd accept such a patch, they actually have a note that while they have an option to use an external sqlite, they don't recommend it. So I imagine this is very intentional.

recipes/tcl/all/conanfile.py
Ahajha: As far as I can tell, it messes with SQLite's linkage. These are overridden in TCL's build, so these are just un-overriding it to the defaults.

Ahajha:
The main issue was that when CPython needed to link against tcl(shared), the rpaths were not correctly fixed on Mac. However, based on your questions, I'm now debating if maybe this could be fixed in a different way:

Perhaps we could keep the vendoring, but I don't think we need to list the extra libraries that are only used for the TCL extensions. So packages that link against it should only use the main TCL library. That would basically just simplify the package info. I'll make another PR in a bit to consider.

Ahajha:
@RubenRBS Sorry for the delay, I've opened an alternate PR that I think fixes the problem in a better way: https://github.com/conan-io/conan-center-index/pull/22939. This can probably be closed, but I'll leave it open in case you think otherwise. The changes here will be pointless if the other PR is merged.

Ahajha:
#22939 was merged, this is no longer necessary.

PR #22713, opened 2024-02-08
[docs] Add changelog 07-February-2024

uilianries:
The new version of C3I Jenkins is here:

- There are few cases when an user can change the username between asking for access in the #4 and the time when the CI bot opens the PR to update the authorized users. In case it happens, we will have a missing username, and sometimes we can't contact the outdated user name. So, this feature pushes those usernames to a second list, where we can ask again from time to time.

- The Bump dependencies PR is used to have all recipe's dependencies aligned to the latest version, or to solve conflicts with other recipes that are consuming a same package with different version. Pushing this kind of PR can result in missing binaries, as an entire dependencies graph will be update and may affect other packages. Second, bump deps usually is not something with high priority, but the current CI service does not know about, resulting in delays when needing to build other priorities, like broken recipes or releasing a very popular package. From now, any bump dependencies PR will be started manually by a Conan team member (please, only ping `@conan-io/barbarians`) to be built in the CI. 

---

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,10 @@
 # Changelog
 
+### 07-February-2024 - 15:43 CET
+
+- [feature] Add waiting list for new collaborators that are not found in access request issue.
+- [feature] Skip building bump dependencies PRs by default. It will require manual CI trigger.
+
 ### 24-January-2024 - 12:13 CET
 
 - [feature] Update Conan 1.x branch to version 1.62.0


RubenRBS (approved):

danimtb (approved):

ericLemanissier:
So PRs only bumping versions will not start until a team member triggers it, but the build will start by simply adding a new-line to the recipe ?
Or do you mean that any PR which modifies dependencies of any package will have to be triggered manually by a team member ?

PR #22715, opened 2024-02-08
plf_list: add v2.72, fix a C++98 support regression in v2.70

valgur:
Adds a fix for https://github.com/mattreecebentley/plf_list/issues/30 to v2.70.


recipes/plf_list/all/conanfile.py
```diff
@@ -43,20 +25,19 @@ def layout(self):
...
+            check_min_cppstd(self, 98)
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def build(self):
         apply_conandata_patches(self)
+        # Do not set PLF_TYPE_TRAITS_SUPPORT fo GCC v5.
+        # Tries and fails to use std::allocator_traits<allocator_type>::is_always_equal otherwise.
+        replace_in_file(self, os.path.join(self.source_folder, "plf_list.h"),
```
AbrilRBS, line 38: This should be reported upstream - is_trivially_copyable is >=5 https://godbolt.org/z/chjMvGnb1, but as you pointed out is_always_equal was not added from the [LWG 2108](https://cplusplus.github.io/LWG/issue2108) until gcc 6 - I wonder if these checks should thus be split! Maybe we can report this upstream

cc @perseoGI

valgur: Already did and a fix has been applied upstream: https://github.com/mattreecebentley/plf_list/pull/32 :slightly_smiling_face:

Thanks for reminding me, though. I bumped the version to include it.

PR #22718, opened 2024-02-08
libsvtav1: Unvendor use of cpuinfo

sjlamerton:
Specify library name and version:  **libsvtav1/***

`libsvtav1` includes a vendored copy of `cpuinfo` in it's `thirdparty` folder. This recipe update instead uses the conan-center package to avoid ODR issues. 

---


uilianries (commented):
I like the idea of using the external cpuinfo, but first let's ask the upstream, so we can move it as official. https://gitlab.com/AOMediaCodec/SVT-AV1/-/issues/2148

sjlamerton:
I agree upstreaming would be ideal, would it still be possible to do this (or an alternative approach) for people using older versions of the library?

uilianries:
I sent a PR to the upstream, asking to support cpuinfo as external dependency: https://gitlab.com/AOMediaCodec/SVT-AV1/-/merge_requests/2178

Let's see what they think. In case accepting, let's use the same patch here.

uilianries:
> would it still be possible to do this (or an alternative approach) for people using older versions of the library?

If it's compatible, then yes.

recipes/libsvtav1/all/conanfile.py
```diff
@@ -43,6 +43,10 @@ def configure(self):
     def layout(self):
         cmake_layout(self, src_folder="src")
 
+    def requirements(self):
+        if self.settings.arch in ("x86", "x86_64"):
+            self.requires("cpuinfo/cci.20231129")
```
uilianries, line 48:
```suggestion
        self.requires("cpuinfo/cci.20231129")
```

I would not keep that restriction, because the cpuinfo package has no limitation. Plus, when using external cpuinfo in libsvtav1, there is no restriction about the arch too.

sjlamerton: Good point, thanks

uilianries (approved):

jwillikers (approved):

PR #22728, opened 2024-02-09
stale: process older PRs first

ericLemanissier:
---

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -62,4 +62,4 @@ jobs:
             recent activity. Thank you for your contributions.
 
         # Max number of operations per run
-        operations-per-run: 100
+        operations-per-run: 30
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -60,4 +62,4 @@ jobs:
             recent activity. Thank you for your contributions.
 
         # Max number of operations per run
-        operations-per-run: 30
+        operations-per-run: 100
```
uilianries, line 65: Let's keep 30. We are not in a hurry with stale PRs.

ericLemanissier: As you which. 
For the record, with the current state of the stale action, 2 operations are consumed to check the last modification date of all PRs having the stale label. This means conan-center-index will never have more than 15 stale PRs

uilianries: No problem, we still can re-visit a bunch of them. Thank you about this information. 

About the rate-limit, we have been doing some experiments using Github API (internally), so as less as we use in actions better will be.

uilianries (approved):
LGTM

jcar87 (approved):

PR #22732, opened 2024-02-09
b2: Don't apply CXXFLAGS env var if it doesn't exist

PeteAudinate:
This is a fix for a regression introduced in #21093

Without this cxxflags are set to "None" which causes a compile failure.

Specify library name and version:  **b2/5.0.0**

---

- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated. **[I couldn't get this working with Conan v2]**

### DIFF
--- a/recipes/b2/portable/conanfile.py
+++ b/recipes/b2/portable/conanfile.py
@@ -159,13 +159,14 @@ class B2Conan(ConanFile):
         if self.options.use_cxx_env:
             envvars = VirtualBuildEnv(self).vars()
 
-            cxx = envvars.get("CXX")
-            if cxx:
-                command += f" --cxx={cxx}"
-                self._write_project_config(cxx)
+            cxx_env = envvars.get("CXX")
+            if cxx_env:
+                command += f" --cxx={cxx_env}"
+                self._write_project_config(cxx_env)
 
             cxxflags_env = envvars.get("CXXFLAGS")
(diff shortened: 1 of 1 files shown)


recipes/b2/portable/conanfile.py
```diff
@@ -159,13 +159,14 @@ def build(self):
...
             envvars = VirtualBuildEnv(self).vars()
 
-            cxx = envvars.get("CXX")
-            if cxx:
-                command += f" --cxx={cxx}"
-                self._write_project_config(cxx)
+            cxx_env = envvars.get("CXX")
+            if cxx_env:
+                command += f" --cxx={cxx_env}"
+                self._write_project_config(cxx_env)
```
grafikrobot, line 165: This looks like a superfluous change.

grafikrobot (changes requested):

recipes/b2/portable/conanfile.py
PeteAudinate: It's a naming tweak to more closely match the section below it (`*_env` for values taken from envvars). 
It means both sections are:
- get env var into `*_env`
- check for None
- take action

I find it clearer this way, but I'm happy to change it back if you prefer.

grafikrobot (approved):

jcar87 (approved):
Thanks @PeteAudinate !

RubenRBS (approved):

PR #22738, opened 2024-02-10
[fastgltf] Update to 0.7.0

spnda:
Specify library name and version:  **fastgltf/0.7.0**

---


RubenRBS (commented):
Hi! Thanks a lot! Quick question :)

recipes/fastgltf/all/conanfile.py
```diff
@@ -83,13 +85,16 @@ def source(self):
...
+            tc.variables["FASTGLTF_DOWNLOAD_SIMDJSON"] = False
         if self.options.enable_small_vector:
-            tc.variables["FASTGLTF_USE_SMALL_VECTOR"] = True
+            tc.variables["FASTGLTF_USE_CUSTOM_SMALLVECTOR"] = True
         if self.options.get_safe("disable_custom_memory_pool"):
             tc.variables["FASTGLTF_DISABLE_CUSTOM_MEMORY_POOL"] = True
         if self.options.get_safe("use_64bit_float"):
             tc.variables["FASTGLTF_USE_64BIT_FLOAT"] = True
+        if self.options.get_safe("compile_as_cpp20"):
+            tc.variables["FASTGLTF_COMPILE_AS_CPP20"] = True
```
uilianries, line 97: @spnda I see you added an option to switch the C++ standard, but actually is a native feature from cmake already. It could be decided by the consumer directly, for instance:

`cmake . -DCMAKE_CXX_STANDARD=20` 

https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html

I could send you a PR with a suggestion making use of the native variable, plus, giving flexibility of using or not from user command line.

Passing it as an option it seems too much, it's actually part of the `settings.compiler.cppstd`

spnda: For the purpose of conan I guess that would work but I disagree with its usage in a library itself that could be included through add_subdirectory or similar, as CMAKE_CXX_STANDARD sets the standard globally. That would override every targets compile standard which could possibly cause other unwanted negative side-effects. That's why I made use of the setting to use `target_compile_features` to guarantee that it's only changed for that target specifically.

If there is some way to set this variable or the cxx standard from the cppstd setting or through some other standard way I'm happy to change that.

uilianries: Thank you for your feedback.

What's the advantage using C++20 in this new release? https://github.com/spnda/fastgltf/compare/v0.6.1...v0.7.0 Looking the diff, I only found `cxx_std_20`, but no feature restriction involved like, `#if __cplusplus >= 202002L`. Am I right? 

If it's correct, you still can pass `target_compile_features(... cxx_std_17)` and build with `cmake -DCMAKE_CXX_STANDARD=20`. The CMake will override, to use C++20. I know you dislike the idea of using `CMAKE_CXX_STANDARD`, but users are free to override it directly, even ignoring your new option. The `cxx_std_17` is more to say "my project requires C++17 at least to be built". Plus, C++17 and C++20 are most compatible (we found errors between C++14 and 20).

spnda: >  Looking the diff, I only found cxx_std_20, but no feature restriction involved like, #if __cplusplus >= 202002L. Am I right?

No. There plenty of small things using C++20 features. Most notably the <bit> header and concepts. Theyre hidden behind a FASTGLTF_CPP_20 macro.

uilianries: Thanks for your answer, so it will be better managing it via cppstd directly in this recipe, and not using the option. The C++ standard adopted is part of settings, not options, so users and configure it via profile. The option should be configured, but based on compiler.cppstd. Plus, the recipe should export the define `FASTGLTF_CPP_20` in case using cppstd 20

uilianries: I'll send a PR to your fork, with these changes.

uilianries: @spnda Please, consider the PR https://github.com/spnda/conan-center-index/pull/2

toge (approved):

PR #22764, opened 2024-02-13
libcoro: add version 0.11.1

toge:
Specify library name and version:  **libcoro/0.11**

---


recipes/libcoro/all/conanfile.py
uilianries, line 1: @toge thank you for updating this project. I helped this project, so I can sure it works on Windows, Mac and Linux. Could you please lift the validation that only works with GCC?

uilianries:
@toge That error is really weird, I'll investigate further. I now more options changed in 0.11 and probably some source is missing when building.

uilianries:
Some weird behavior, when using `cmake_dependent_option` CMake feature with Conan, for some reason, it does not work.

https://c3i.jfrog.io/c3i/misc/logs/pr/22764/2-linux-gcc11/libcoro/0.11//1e83dbe0105b2ee2583fa6fe1489fdb74a950fe7-build.txt

Some options should enabled, they are not:

```
libcoro LIBCORO_EXTERNAL_DEPENDENCIES = ON
libcoro LIBCORO_BUILD_TESTS           = OFF
libcoro LIBCORO_CODE_COVERAGE         = OFF
libcoro LIBCORO_BUILD_EXAMPLES        = OFF
libcoro LIBCORO_FEATURE_NETWORKING    = ON
libcoro LIBCORO_FEATURE_TLS           = ON
libcoro LIBCORO_RUN_GITCONFIG         = OFF
libcoro LIBCORO_BUILD_SHARED_LIBS     = ON
```

Locally, running the same command like, but using only cmake in the docker image, with conanfile.txt, it works.

recipes/libcoro/all/conanfile.py
```diff
@@ -40,8 +40,12 @@ def _min_cppstd(self):
     @property
     def _minimum_compilers_version(self):
         return {
-            "gcc": "10.2.1",
-        }
+                "gcc": "10.2.1",
+                "clang": "16.0.0",
+                "apple-clang": "13",
```
uilianries, line 45:
```suggestion
                "apple-clang": "16",
```

Need to re-check it with the upstream, but probably the project needs an update.

uilianries:
I'll check the sys header missing.

uilianries:
After checking the upstream's code, some options are not available on Windows, but the recipe did not have the logic, because until now we have supported only Linux gcc. I removed dynamically the affected options.

uilianries:
The current error is because shared library was not supported before 0.11 and the custom patch available in CCI is bugged (need extra work due static variables). I'll remove those patches and keep it as static library. Since 0.11 anyone can use shared library with safety (tested in the upstream).

toge:
@uilianries 
Sorry conflicted!
I am goting to revert it.

uilianries:
@toge No, I'm sorry of abusing your PR 😅 The fact is `WINDOWS_EXPORT_ALL_SYMBOLS` will not work here. CMake is capable to export all methods, but not static variables. The libcoro has 2 public static variables that need the dlspec directly in the code. You can see the real fix here: https://github.com/jbaldwin/libcoro/pull/236

toge:
@uilianries 
Thank you for your information!
Now I understand.

uilianries:
The CI is not smart enough to understand dynamic shared option. We will need to keep it, but manage as invalid configuration.

uilianries:
@toge It's finally working! 🎉

uilianries (approved):
LGTM

toge:
@uilianries 
Thanks a lot!

jwillikers (approved):

PR #22772, opened 2024-02-14
duckdb: add version 0.10.1

toge:
Specify library name and version:  **duckdb/0.10.1**

duckdb/0.10.0 uses duckdb_platform to detect platform. https://github.com/duckdb/duckdb/blob/v0.10.0/CMakeLists.txt#L1137-L1157
It causes several errors on cross building.
To avoid them, we have to set `DUCKDB_EXPLICIT_PLATFORM`.

duckdb/0.10.0 began to use private skiplist library.

https://github.com/duckdb/duckdb/compare/v0.9.2...v0.10.1

---


recipes/duckdb/all/conanfile.py
```diff
@@ -159,6 +166,8 @@ def generate(self):
         tc.variables["ENABLE_UBSAN"] = False
         if is_msvc(self) and not self.options.shared:
             tc.preprocessor_definitions["DUCKDB_API"] = ""
+        if Version(self.version) >= "0.10.0" and self.options.platform != "":
+            tc.variables["DDUCKDB_EXPLICIT_PLATFORM"] = self.options.platform
```
uilianries, line 170:
```suggestion
        if Version(self.version) >= "0.10.0" and cross_building(self):
            tc.variables["DDUCKDB_EXPLICIT_PLATFORM"] = f"{self.settings.os}_{self.settings.arch}"
```

I guess we could avoid that bew option and fill the definition when cross-building only.

In the nightly test we can have a hint about what is used: https://github.com/duckdb/duckdb/blob/16938041baa047e77378aff013acde6a6d863c52/.github/workflows/NightlyTests.yml#L660

uilianries (approved):
LGTM

uilianries:
@toge I did a change in your PR: Removed platform option and kept it as automatically defined when cross building.

jwillikers (approved):

PR #22785, opened 2024-02-16
glaze: add version 2.1.4

toge:
Specify library name and version:  **glaze/***

- glaze/2.1.4 uses `decltype(auto)` which is not supported by gcc < 11.3 due to gcc bug. [ref](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102229)

---

### DIFF
--- a/recipes/glaze/all/conandata.yml
+++ b/recipes/glaze/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "2.1.3":
-    url: "https://github.com/stephenberry/glaze/archive/v2.1.3.tar.gz"
-    sha256: "10345cd0fab29a8386815fd94562f014caf90df1640e2e4409466da6a7a10ee2"
+  "2.1.4":
+    url: "https://github.com/stephenberry/glaze/archive/v2.1.4.tar.gz"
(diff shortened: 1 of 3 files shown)


RubenRBS (commented):
Hi! Thanks a lot as always :)

It's weird that gcc does not seem to work here, as their actions seem to be testing for gcc 11 too, and they would have caught a failing build there https://github.com/stephenberry/glaze/compare/v2.1.3...v2.1.4#diff-2ceee9940e7b1f3b7b4cea77c164a4417544c1495fdda80678bcac0313fb1589L23, I won'der whats going on in there?

toge:
@RubenRBS 
Thank you for your comment!
It seems to be gcc bug. [ref](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=102229)
It has been fixed since gcc 11.3.

While cci's gcc11 is 11.1, glaze/2.1.4(uses `decltype(auto)`) met a compilation error.

I will try to relax compilation version check.

RubenRBS:
> While cci's gcc11 is 11.1

Maybe I can do something about it, let me double check, but yeah, checking for 11.3 would be great

RubenRBS (approved):
Approving, will circle bak to this once we update to 11.3 if possible :)

jwillikers (approved):

PR #22727, opened 2024-02-09
fontconfig: update expat

Ahajha:
Fontconfig is causing version conflict in the PR #21387 due expat.

I've simplified the test package for the older versions, for some reason on Mac it is segfaulting if we try to do anything remotely complicated (Can't even call FcFini() without it segfaulting). Seems to be fine on newer versions, so I think this is just some bug in the older versions.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21892, opened 2023-12-25
[mocknetworkaccessmanager] add recipe for 0.12.0

MartinDelille:
Specify library name and version:  **mocknetworkaccessmanager/0.12.0**

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #22788, opened 2024-02-16
[wayland] Bump required dependencies

MartinDelille:

---

### DIFF
--- a/recipes/wayland/all/conanfile.py
+++ b/recipes/wayland/all/conanfile.py
@@ -50,15 +50,15 @@ class WaylandConan(ConanFile):
         if self.options.enable_libraries:
             self.requires("libffi/3.4.4")
         if self.options.enable_dtd_validation:
-            self.requires("libxml2/2.12.3")
-        self.requires("expat/2.5.0")
+            self.requires("libxml2/2.12.4")
+        self.requires("expat/2.6.0")
 
     def validate(self):
         if self.settings.os != "Linux":
             raise ConanInvalidConfiguration(f"{self.ref} only supports Linux")
 
     def build_requirements(self):
-        self.tool_requires("meson/1.3.0")
+        self.tool_requires("meson/1.3.1")
         if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
             self.tool_requires("pkgconf/2.1.0")
         if not can_run(self):


jwillikers (approved):

ericLemanissier:
@conan-io/barbarians We are in a worse situation on all accounts:

- conflicts still happen as they used to
- the fix of these conflict is much slower than before, because we have to wait for a team member twice per PR

also pinging @uilianries @RubenRBS @franramirez688 because it looks like regular users cannot ping barbarians

jcar87:
Please be aware that we are very unlikely to pay attention to a PR without a description, no matter how trivial it may seem.

This other PR: https://github.com/conan-io/conan-center-index/pull/22838 - is a good example - this provides us with enough context to trace back why the version conflict happened in the first place. 
We should have been more diligent before approving https://github.com/conan-io/conan-center-index/pull/22727 - as we have enough tools to see that this would've caused conflicts.


*This* PR, by updating  `libxml2` to version `2.12.4` - actually creates a different conflict - since xkb-common still depends on `2.12.3. So if this PR was automatically merged  it would just create another conflict (I have been testing this from the qt recipe). This is why we have disabled automerging of all dependency bumps  - it requires a coordination to follow up with additional PRs to get consistency, both in terms of opening them, running them and merging them. 

I'm reverting the change for that PR to avoid another conflict and help PRs held up by the `expat` conflict. We will ensure that we minimise the introduction of conflicts while reviewing any future PR. We are simultaneously working on tooling to automate the detection of potential conflicts (to prevent them in the first place), and also the ability to touch multiple recipes in a single PR (to be able to consistently version bump the entire repo, if necessary).

jwillikers (approved):

jcar87 (approved):

ericLemanissier:
> We should have been more diligent before approving #22727 - as we have enough tools to see that this would've caused conflicts.

The fact that the proper tooling to detect these conflicts ready to process all pull requests (even those which do not bump deps can create conflicts by changing an option value for example) makes the PR embargo counter-productive. In a perfect world, this tool would be open-source so that every conan user can benefit, and running as a github action on CCI so that it is transparent to contributors (any contributor would be able to test PRs on any repo, and fix all conflicts before submitting the PR to CCI)


> This PR, by updating libxml2 to version 2.12.4 - actually creates a different conflict - since xkb-common still depends on `2.12.3. So if this PR was automatically merged it would just create another conflict (I have been testing this from the qt recipe), and you would still have an issue with Qt. This is why we have disabled automerging of all dependency bumps - it requires a coordination to follow up with additional PRs to get consistency, both in terms of opening them, running them and merging them.

I can see two solutions here, until you allow several recipes to be modified by a single PR:
1- either you somehow put a definitive hold on all the PRs bumping libxml2 on all recipes, and this effectively prevents all conflicts from happening (but this should be documented)
2- or you somehow allow libxml2 to be bumped sometimes, and when it happens you get the same conflict as the one you just avoided.

MartinDelille:
Hi @jcar87 ! I should have indeed have given a proper description (which I just did). My goal is to get https://github.com/conan-io/conan-center-index/pull/21892 merged, a PR that I am working on since a long time.

Thank you for all your explanations by the way!

ericLemanissier:
@uilianries @RubenRBS @franramirez688 @jcar87 can one of you please trigger the CI so that this PR is merged, and the conflict finally resolved ?

ericLemanissier:
>  @uilianries @RubenRBS @franramirez688 @jcar87 can one of you please trigger the CI so that this PR is merged, and the conflict finally resolved ?

franramirez688 (approved):

PR #22794, opened 2024-02-16
opendis6: Set Minimum C++ Standard to 14 for RHEL 8 Support [CRITICAL]

crhowell3:
Specify library name and version:  **opendis6/0.1.0**

The current minimum cppstd is set to 17, which excludes RHEL 8 GCC 8.x support. It is critical that opendis6 be available for RHEL 8. GCC 8.x natively uses C++14 as the default standard, so I have set the cppstd accordingly for this package.

---

### DIFF
--- a/recipes/opendis6/all/conanfile.py
+++ b/recipes/opendis6/all/conanfile.py
@@ -29,7 +29,7 @@ class OpenDis6Conan(ConanFile):
 
     @property
     def _min_cppstd(self):
-        return "17"
+        return "14"
 
     @property
     def _compilers_minimum_version(self):
--- a/recipes/opendis6/all/conanfile.py
+++ b/recipes/opendis6/all/conanfile.py
@@ -36,9 +36,9 @@ class OpenDis6Conan(ConanFile):
         return {
             "Visual Studio": "15",
             "msvc": "191",
-            "gcc": "8.5",
-            "clang": "6",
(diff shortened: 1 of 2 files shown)


uilianries:
@crhowell3 Thank you for this update! I see you are the author of the project as well. Please, just be sure, the project points C++17 when building: https://github.com/crhowell3/opendis6/blob/main/CMakeLists.txt#L10

So is it correct to use C++14 anyway? Will the upstream be update to C++14 as well? 

Plus, enabling C++14, we could use GCC 6 (because it supports C++14). Do know any blocker by using GCC 6 or older than GCC8? 

Regards!

crhowell3:
So technically, GCC 8.x (the compiler for RHEL 8) supports C++17, but the default C++ standard for 8.x is C++14, and so Conan complains that GCC 8.5 is not compatible for _min_cppstd "17". I don't think the actual OpenDIS6 project will need to downgrade to 14.

uilianries:
@crhowell3 could you please update the compiler list as well, so more compilers could build the project:

```
            "apple-clang": "10",
            "clang": "7",
            "gcc": "7",
            "msvc": "191",
            "Visual Studio": "15",
```

Plus, the test package has a CMakeLists.txt with c++17 listed too. Could you update it to c++14?

ericLemanissier (approved):

PR #22807, opened 2024-02-17
poco: add POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX definition

### LINKED ISSUE #22805: [package] poco/all: Poco::Foundation's SharedLibrary::suffix doesn't work cleanly with build_type Debug
### Description

With build_type=Debug specified in profile and building Poco and a shared library through conan, Poco::SharedLibrary::suffix() will return the incorrect platform-specific extension.

This is due to conan creating shared libraries without a ".d" to specify a debug version of a library as builds are separated into different folders.

For context see comment in the Poco header for this function: [SharedLibrary.h:L115](https://github.com/pocoproject/poco/blob/88be66972a07968f027b68860add1d43aa5f135c/Foundation/include/Poco/SharedLibrary.h#L115)

My suggestion is to add `tc.preprocessor_definitions["POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX"] = "1"` into the recipe for poco. Should a new option for this behaviour be added as well, or should it be added for all builds?

* Package Name/Version: **poco/1.12.4**
* Operating System+version: Linux Ubuntu 16.04
* Compiler+version: GCC 13
* Docker image: conanio/gcc13-ubuntu16.04
* Conan version: conan 2.0.17
* Python version: Python 3.7.13

### Conan profile

[settings]
arch=x86_64
build_type=Debug
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=13
os=Linux

### Steps to reproduce

conan install -r poco/1.12.4

### Logs

No logs, error is during runtime not build time

obnyis:
Specify library name and version:  **poco/all**

Add the POCO_NO_SHARED_LIBRARY_DEBUG_SUFFIX

Fixes #22805

---


uilianries (dismissed):
LGTM

jcar87 (dismissed):
reviewing this for a sec - no issues with the PR

jcar87:
Had a chance to review this further, along with the repro provided by @obnyis  in https://github.com/conan-io/conan-center-index/issues/22805#issuecomment-1953536235.

This was super confusing to us because the name of the option would indicate that this affects the poco libraries themselves, when in fact it affects the poco class to _load_ other shared libraries at runtime: https://docs.pocoproject.org/current/Poco.SharedLibrary.html .

@obnyis  - please note that there is *not* a universal convention as to whether debug libraries end with `d` suffix or not. Most likely to happen on Windows, but not guaranteed. You'll find that some debug libraries will end with `d` and some won't - Conan *does not* change the defaults of the libraries, and recipes in Conan Center *should not* make any changes either - precisely to ensure that the filenames are exactly as they are as if built without Conan, to cover for cases like this where the filenames themselves may be publicly known and not abstracted away.

@uilianries, @RubenRBS  -  it should be fine to expose this as an option, but the option name is still very confusing, if we can come up with something better, that would be great!

uilianries:
What about `comp_shared_lib_debug_suffix` ? 

- It follows positive logic
- Related to Poco component shared library

obnyis:
@uilianries, just to confirm before I update this PR, is your suggestion to have the option name `comp_shared_lib_debug_suffix` with a default value of True, and only add the preprocessor_definitions when it is set to False?

Given that this is in the Foundation component targeting the SharedLibrary class, do you think the name of `comp_foundation_sharedlibrary_debug_suffix` would be clear? Happy to keep what you've suggested, but just want to confirm before I push changes

uilianries:
> @uilianries, just to confirm before I update this PR, is your suggestion to have the option name `comp_shared_lib_debug_suffix` with a default value of True, and only add the preprocessor_definitions when it is set to False?
> 
> Given that this is in the Foundation component targeting the SharedLibrary class, do you think the name of `comp_foundation_sharedlibrary_debug_suffix` would be clear? Happy to keep what you've suggested, but just want to confirm before I push changes

@obnyis the `comp_foundation_sharedlibrary_debug_suffix` looks better. Please, let's use this. Your logic is also correct.

uilianries (approved):
LGTM

RubenRBS (approved):
Thanks for bringing this to our attention! :)

PR #22833, opened 2024-02-20
opencv: Fix lack of config verification

connorsmacd:
Specify library name and version:  **opencv/4.x**

This PR modifies the OpenCV 4.x recipe to enable the configuration option `ENABLE_CONFIG_VERIFICATION`.

According the [OpenCV documentation](https://docs.opencv.org/4.8.0/db/d05/tutorial_config_reference.html), `ENABLE_CONFIG_VERIFICATION` does the following:

> For each enabled dependency (`WITH_` option) verify that it has been found and enabled (`HAVE_` variable). By default feature will be silently turned off if dependency was not found, but with this option enabled cmake configuration will fail. Convenient for packaging systems which require stable library configuration not depending on environment fluctuations.

The recipe did not enable this option, so it was possible for the build to succeed even though an explicitly enabled dependency was not found. I noticed this issue while I was playing around with my own fork of the FFmpeg recipe -- despite `WITH_FFMPEG` being passed, I would see this in the build output:

```
--   Video I/O:
--     FFMPEG:                      NO (find_package)
```

While `find_package` failed due to my own error in the FFmpeg recipe, I would have expected the build to fail as well, but it succeeded due to the error passing silently. With the change in this PR, the build now fails as expected.

---


RubenRBS:
I've restarted the build now that quirc is properly handled :)

uilianries:
@connorsmacd Thank you for reporting your case and suggesting this recipe change. However, Conan does not enforce `ENABLE_CONFIG_VERIFICATION`, instead, it follows its default value/behavior:

- https://github.com/opencv/opencv/blob/4.10.0/CMakeLists.txt#L567
- https://docs.opencv.org/4.x/db/d05/tutorial_config_reference.html

Its default value is `OFF`, and Conan tries to keep it as close as possible to the upstream behavior. Still, your request is valid, but adding a new option, that will affect the Conan package ID, but only causes influence over the setup build step is too much.

What I would suggest is using a Conan configuration to inject that option without touching the recipe. It's a cheaper operation and works well for your case. First, take a look at the supported configuration for Conan 2.x, you can find several configs: https://docs.conan.io/2/reference/config_files/global_conf.html

What you should use here is: 

```
tools.cmake.cmaketoolchain:extra_variables: Dictionary with variables to be injected in CMakeToolchain (potential override of CMakeToolchain defined variables)
```

This will reflect directly in your Conna create/install command line, or, in your Conan profile as follows:

```
conan create 4.x --version=4.10 ... -c tools.cmake.cmaketoolchain:extra_variables='{"ENABLE_CONFIG_VERIFICATION": "ON"}'
```

Or, in profile:

```
[settings]
...
[conf]
opencv/*:tools.cmake.cmaketoolchain:extra_variables={"ENABLE_CONFIG_VERIFICATION": "ON"}
```

As you can see, in your Conan profile you can filter what package should be applied, for this case, it's listed for any version of OpenCV. 

I'll be closing this PR as I'm not inclined to accept it. Please, feel free to comment about the suggested solution and give your feedback. Regards!

PR #22842, opened 2024-02-21
simfil: add version 0.1.1

johannes-wolf:
Specify library name and version:  **simfil/0.1.1**

Simfil - a library that implements a query language for structured (map) data - is a new addition to ConanCenter because we want to use Conan for all our packages and
dependencies. Simfil is a dependency of packages we want to submit in the future.

---


recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+
+required_conan_version = ">=1.62.0"
+
+
+class SimfilRecipe(ConanFile):
+    name = "simfil"
+    description = "simfil is a C++ 17 library and a language for querying structured map feature data. The library provides an efficient in-memory storage pool for map data, optimized for the simfil query language, along with a query interpreter to query the actual data."
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/Klebert-Engineering/simfil"
+    license = "BSD 3-Clause"
```
uilianries, line 17:
```suggestion
    license = "BSD-3-Clause"
```

Always use SDPX short format: https://spdx.org/licenses/BSD-3-Clause.html

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+    topics = ["query language"]
```
uilianries, line 18:
```suggestion
    topics = ["query-language", "json", "data-model"]
```

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+        "fPIC": [True, False],
+        "with_json": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_json": True,
+    }
+
+    exports_sources = "include/*", "src/*"
```
uilianries, line 33:
```suggestion
```

This step is only useful in case not downloading the sources from an external URL, but using with the conanfile.py at same folder.

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+        min_version = self._minimum_compilers_version.get(str(self.settings.compiler))
+        if not min_version:
+            self.output.warning(f"{self.name} recipe lacks information about the {self.settings.compiler} compiler support.")
+        else:
+            if Version(self.settings.compiler.version) < min_version:
+                raise ConanInvalidConfiguration(
+                    f"{self.name} requires Concepts support. The current compiler {self.settings.compiler} {self.settings.compiler.version} does not support it.")
+
+    def build_requirements(self):
+        self.build_requires("cmake/3.28.1")
```
uilianries, line 57:
```suggestion
        self.tools_requires("cmake/[>3.19 <4]")
```
Cmake is allowed to use range versions: 

Plus, build_requires is deprecated, tool_requires is its replacement: https://docs.conan.io/1/migrating_to_2.0/recipes.html#requirements

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+    @property
+    def _minimum_compilers_version(self):
+        return {
+            "Visual Studio": "16",
+            "msvc": "192",
+            "gcc": "10",
+            "clang": "10",
+            "apple-clang": "14",
+        }
+
```
uilianries, line 44:
```suggestion

    def configure(self):
        if self.options.shared:
            self.options.rm_safe("fPIC")

```

The fPIC is always removed when building shared, because shared without fPIC is not used at all.

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["SIMFIL_CONAN"] = True
+        tc.cache_variables["SIMFIL_SHARED"] = bool(self.options.get_safe("shared"))
+        tc.cache_variables["SIMFIL_WITH_REPL"] = False
+        tc.cache_variables["SIMFIL_WITH_COVERAGE"] = False
+        tc.cache_variables["SIMFIL_WITH_TESTS"] = False
+        tc.cache_variables["SIMFIL_WITH_EXAMPLES"] = False
+        tc.cache_variables["SIMFIL_WITH_MODEL_JSON"] = bool(self.options.with_json)
```
uilianries, line 84:
```suggestion
        tc.cache_variables["SIMFIL_WITH_MODEL_JSON"] = self.options.with_json
```

The CMakeToolchain converts to boolean automatically. CMake executable will receive "True", but it's totally capable to understand that string.

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+        tc.generate()
+        deps = CMakeDeps(self)
+        deps.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
```
uilianries, line 94:
```suggestion
    def package(self):
        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
```

Need to copy the license always with the package.

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["simfil"]
+        self.cpp_info.set_property("cmake_target_name", "simfil::simfil")
```
uilianries, line 100:
```suggestion
```

By default, Conan will use the package name as cmake target. The result will be the same as yours.

recipes/simfil/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+find_package(simfil CONFIG REQUIRED)
+
+add_executable(test_package src/example.cpp)
+target_link_libraries(test_package simfil::simfil)
```
uilianries, line 7:
```suggestion
target_link_libraries(test_package simfil::simfil)
target_compile_features(test_package PRIVATE cxx_std_20)
```

Because Conan 1.x does not specify settings.compiler.cppstd by default

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,100 @@
...
+                raise ConanInvalidConfiguration(
+                    f"{self.name} requires Concepts support. The current compiler {self.settings.compiler} {self.settings.compiler.version} does not support it.")
+
+    def build_requirements(self):
+        self.build_requires("cmake/3.28.1")
+
+    def requirements(self):
+        self.requires("sfl/1.2.4", transitive_headers=True)
+        self.requires("fmt/10.2.1", transitive_headers=True)
+        self.requires("bitsery/5.2.3", transitive_headers=True)
```
uilianries, line 62: Transitive headers should be used only when dependencies are part of public headers. It's case here.

uilianries (commented):
@johannes-wolf Thank you for your PR! Please, take a look in my review.

recipes/simfil/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeToolchain, CMakeDeps, cmake_layout
+from conan.tools.files import copy, get
+from conan.tools.scm import Version
+from conan.errors import ConanInvalidConfiguration
+import os
+
+
+required_conan_version = ">=1.62.0"
```
uilianries, line 10:
```suggestion
required_conan_version = ">=1.53.0"
```

This recipe does not require 1.6x.x specific features.

johannes-wolf: How do I know?

uilianries (approved):
LGTM

recipes/simfil/all/conanfile.py
uilianries: There is no easy way, need to read the changelog and compare if you need some feature from there: https://docs.conan.io/1/changelog.html

also, most of tools has which version they are available, e.g.

https://docs.conan.io/1/reference/conanfile/tools/files/basic.html#conan-tools-files-save

RubenRBS (approved):

PR #22849, opened 2024-02-22
rmm: add v24.04.00, bump deps

valgur:
.


recipes/rmm/all/conanfile.py
```diff
@@ -42,8 +42,8 @@ def layout(self):
 
     def requirements(self):
         self.requires("thrust/1.17.2")
-        self.requires("spdlog/1.12.0")
-        self.requires("fmt/10.1.1")
+        self.requires("spdlog/1.13.0")
+        self.requires("fmt/10.2.1")
```
czoido, line 46: ok, I think we can accept this bump deps because there are very few recipes that depend on the old versions

czoido (approved):

czoido (approved):

ericLemanissier (approved):

RubenRBS (approved):

PR #22854, opened 2024-02-22
New package: orc

wgtmac:
Specify library name and version:  **orc/1.9.2**

Add the recipe for Apache ORC: https://orc.apache.org/. I am a PMC member of the Apache ORC community.

---


valgur:
@wgtmac Can we merge this PR and #21007, maybe? Or is this one already a superset of my PR and I can close it once this one is green?

RubenRBS:
@wgtmac feel free to ping me once this is ready :)

wgtmac:
@valgur @RubenRBS This is ready for review now. Thanks!

wgtmac:
Now CIs are all green! This is ready for review now. Thanks! @valgur @RubenRBS

recipes/orc/all/conanfile.py
```diff
@@ -0,0 +1,149 @@
...
+
+        tc = CMakeToolchain(self)
+        tc.variables["ORC_PACKAGE_KIND"] = "conan"
+        tc.variables["BUILD_JAVA"] = False
+        tc.variables["BUILD_CPP_TESTS"] = False
+        tc.variables["BUILD_TOOLS"] = False
+        tc.variables["BUILD_LIBHDFSPP"] = False
+        tc.variables["BUILD_POSITION_INDEPENDENT_LIB"] = bool(self.options.get_safe("fPIC", True))
+        tc.variables["INSTALL_VENDORED_LIBS"] = False
+        tc.variables["BUILD_ENABLE_AVX512"] = False
```
valgur, line 106: This conflicts with the `tc.cache_variables["BUILD_ENABLE_AVX512"] = self.options.get_safe("build_avx512", False)` below.

Also, can all of the `cache_variables` be turned into plain `variable`, maybe? Either way, you will have to add `tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"` for ORC versions below 1.9 because they use a CMake policy version less than 3.12 (https://github.com/apache/orc/blob/v1.8.6/CMakeLists.txt#L18).

recipes/orc/all/patches/0001-thirdparty-toolchain.patch
valgur, line 1: Since it's more of a standalone file than a patch, I would add the file to the recipe root as `ConanThirdpartyToolchain.cmake` and add `copy(self, "ConanThirdpartyToolchain.cmake", self.recipe_folder, os.path.join(self.export_sources_folder, "src", "cmake_modules"))` to `export_sources()`. Makes it easier to read and edit, if necessary.

recipes/orc/all/conanfile.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def build_requirements(self):
+        if not self._is_legacy_one_profile:
+            self.tool_requires("protobuf/<host_version>")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def export_sources(self):
+        export_conandata_patches(self)
```
valgur, line 92: Totally a nitpick, but I recommend following the order used in the [CMake package template](https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py) for the recipe methods to keep the order of their appliation clearer for people new to Conan.

recipes/orc/all/patches/0001-thirdparty-toolchain.patch
```diff
@@ -0,0 +1,55 @@
...
++
++# ----------------------------------------------------------------------
++# ZLIB
++
++find_package (ZLIB REQUIRED)
++
++add_library (orc_zlib INTERFACE)
++add_library (orc::zlib ALIAS orc_zlib)
++target_link_libraries (orc_zlib INTERFACE ${ZLIB_LIBRARIES})
++target_include_directories (orc_zlib SYSTEM INTERFACE ${ZLIB_INCLUDE_DIR})
```
valgur, line 25: You can probably avoid most of these steps in this .cmake file by renaming the generated CMake targets in `generate()`:
```python
        deps = CMakeDeps(self)
        deps.set_property("zlib", "cmake_target_name", "orc::zlib")
        ...
        deps.generate()
```

Which is basically what I did in my PR: https://github.com/conan-io/conan-center-index/pull/21007/files#diff-3cccf7963994e6634e694bd42cc4cf63970514d07a4cdf290d2f90ab0aefcb0bR118-R121

recipes/orc/all/test_v1_package/CMakeLists.txt
valgur, line 1: `test_v1_package` is not required for new recipes and should not be included.

valgur (changes requested):
Looks good overall.
The `cache_variables` CMake policy issue will cause problems and needs to be fixed.

recipes/orc/all/test_v1_package/CMakeLists.txt
wgtmac: Did you mean I do not need to fix any failure in the `Conan v1 pipeline` CI?

recipes/orc/all/patches/0001-thirdparty-toolchain.patch
wgtmac: Thanks for the suggestion! I'd go with https://github.com/conan-io/conan-center-index/pull/22854#discussion_r1501853311 as we have to customize some logic anyway in the cmake file and this is not required any more after ORC v2.0.0.

wgtmac:
Thanks for the review! I've addressed your comments, please take a look again. @valgur

recipes/orc/all/test_v1_package/CMakeLists.txt
valgur: Not quite. Dropping `test_v1_package` only means that legacy v1 generator output (`cmake_find_package_multi`) is not tested. The v1 pipelines are still run and need to succeed, but they only test that the backported v2 generators (`CMakeToolchain`, `CMakeDeps`) work.

Please remove the `test_v1_package` folder.

valgur (approved):
Aside from the minor `test_v1_package` comment, LGTM. Thanks!

wgtmac:
Removed test_v1_package. Thanks @valgur!

recipes/orc/all/conanfile.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def generate(self):
+        VirtualBuildEnv(self).generate()
+        VirtualRunEnv(self).generate(scope="build")
+
+        tc = CMakeToolchain(self)
+        tc.variables["ORC_PACKAGE_KIND"] = "conan"
+        tc.variables["BUILD_JAVA"] = False
+        tc.variables["BUILD_CPP_TESTS"] = False
+        tc.variables["BUILD_TOOLS"] = False
```
RubenRBS, line 108:
```suggestion
        tc.variables["BUILD_TOOLS"] = self.options.build_tools
```

recipes/orc/all/conanfile.py
```diff
@@ -0,0 +1,157 @@
...
+        tc.variables["BUILD_ENABLE_AVX512"] = self.options.get_safe("build_avx512", False)
+        tc.variables["STOP_BUILD_ON_WARNING"] = False
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
+
+        # CMake versions less than 3.12 are supported by ORC pre-1.9.0 versions.
+        # Setting policy CMP0077 to NEW to remove unnecessary cache_variables settings.
+        if self.version < "1.9.0":
+            tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
+
+        protoc_path = os.path.join(self.dependencies["protobuf"].package_folder, "bin", "protoc")
```
RubenRBS, line 122:
```suggestion
        protoc_path = os.path.join(self.dependencies["protobuf"].cpp_info.bindir, "protoc")
```

RubenRBS (commented):
Thanks! Two minor comments, otherwise it looks good :)

wgtmac:
> Thanks! Two minor comments, otherwise it looks good :)

Thanks for the review! Fixed all the comments. Please take a look again. @RubenRBS

recipes/orc/all/ConanThirdpartyToolchain.cmake
```diff
@@ -0,0 +1,49 @@
+# ----------------------------------------------------------------------
+# Snappy
+
+find_package (Snappy REQUIRED)
+
+add_library (orc_snappy INTERFACE)
+add_library (orc::snappy ALIAS orc_snappy)
+target_link_libraries(orc_snappy INTERFACE ${Snappy_LIBRARIES})
+target_include_directories (orc_snappy SYSTEM INTERFACE ${Snappy_INCLUDE_DIR})
```
uilianries, line 9:
```suggestion
target_link_libraries(orc_snappy INTERFACE Snappy::snappy)
```

Is not possible to use CMake targets instead? Passing only Snappy_LIBRARIES and Include dir will not cover all aspects, like compiler definitions (in case used). Same for others in this cmake file.

uilianries: No problem, CMake gives a smart way to do that:

```
target_link_libraries(orc_snappy INTERFACE  
    $<$<BOOL:${BUILD_SHARED_LIBS}>:zstd::libzstd_shared>
    $<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:zstd::libzstd_static>
)
```

wgtmac: Thanks for providing the generator expression! Unfortunately, it seems that `${BUILD_SHARED_LIBS}` is not enough since 3rd-party dependencies may be specified to different static/shared linkings. For example, following command does not compile:
```
conan create . --version=1.8.6 -o orc/1.8.6:shared=True -o lz4/1.9.4:shared=False
```

uilianries: @wgtmac, good catch! In that case, you could link to the target that's available. As you are using Conan, it will load CMakeDeps, so it should be safe, because only one or another target will be available, no mess, no mixes.

```cmake
target_link_libraries(orc_snappy INTERFACE
    $<TARGET_NAME_IF_EXISTS:zstd::libzstd_static>
    $<TARGET_NAME_IF_EXISTS:zstd::libzstd_shared>
)
```

wgtmac: Thanks for the detail guide! Fixed and please take a look again. @uilianries

recipes/orc/all/conanfile.py
```diff
@@ -0,0 +1,157 @@
...
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        if self.settings.os == "Windows" and self.options.shared:
+            mkdir(self, os.path.join(self.package_folder, "bin"))
+            os.rename(os.path.join(self.package_folder, "lib", "orc.dll"),
+                      os.path.join(self.package_folder, "bin", "orc.dll"))
+
+    def package_info(self):
+        self.cpp_info.libs = ["orc"]
```
uilianries, line 157:
```suggestion
        self.cpp_info.libs = ["orc"]
        if self.settings.os in ["Linux", "FreeBSD"]:
            self.cpp_info.system_libs = ["pthread", "m"]
```

This project uses pthread and libm. It's working in test package because is getting a ride from dependencies, but each recipe should be explicit about what is needed.

- https://github.com/apache/orc/blob/main/c%2B%2B/src/CpuInfoUtil.cc#L47
- https://github.com/apache/orc/blob/main/c%2B%2B/src/BloomFilter.hh#L25

uilianries (approved):
LGTM

valgur (approved):

PR #22870, opened 2024-02-23
[config] Fix validate infrastructure entry name

uilianries:
The internal CI build Validate Infrastructure is responsible for updating the document https://github.com/conan-io/conan-center-index/blob/master/docs/supported_platforms_and_configurations.md.

It reads the config yaml file to decide if it should or not open a PR to update the document. But reading the C3I code, it actually wants to read `create_docs_pull-request`, but not `open_docs_pull-request`. By default the C3I code does not open the PR, so we need to fix the entry in config anyway.

Kudos @Croydon that spotted this misconfiguration! 💚 

---

### DIFF
--- a/.c3i/config_v1.yml
+++ b/.c3i/config_v1.yml
@@ -74,7 +74,7 @@ tasks:
   validate_infrastructure:
     macos_executors: 2
     windows_executors: 4
-    open_docs_pull-request: true
+    create_docs_pull-request: true
 
 # Profile configurations to build packages
 configurations:
--- a/.c3i/config_v2.yml
+++ b/.c3i/config_v2.yml
@@ -66,7 +66,7 @@ tasks:
   validate_infrastructure:
     macos_executors: 2
     windows_executors: 4
-    open_docs_pull-request: false
+    create_docs_pull-request: false
 
 configurations:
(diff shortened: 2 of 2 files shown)


RubenRBS (approved):

danimtb (approved):

Croydon:
I guess the documentation is still not entirely up-to-date. For example, Conanv2 builds for VS 2022

uilianries:
Good catch! Indeed we have it disabled:

https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v2.yml#L69

I'll check with @danimtb how to update it.

uilianries:
@Croydon after speaking to @danimtb, we can not do much about right now. That document is updated directly by an internal CI build in Jenkins that checks all configurations, like installed compilers, apps and expected machines present in the CI, then, it opens a PR with all collected information and overrides directly what is listed in the file. Both Conan 1.x and Conan 2.x are using different configurations, including compiler versions, which means, Conan 2.x would override with a total different value, and we running those builds daily.

The best option would be fixing the C3I Jenkins code to fix the situation, but right now we are working in other things, including improving the Docker CI. So it should delay by weeks.

PR #22876, opened 2024-02-24
dispenso: Add new recipe

RubenRBS:
Specify library name and version:  **dispenso/1.3.0**

- Initially tried to add 1.2, but after testing it, it did not build for Linux under Docker, reported as https://github.com/facebookincubator/dispenso/issues/27 upstream - Decided to move to 1.3 after it got fixed there


recipes/dispenso/all/conanfile.py
```diff
@@ -0,0 +1,124 @@
...
+
+        if self.settings.os == "Windows":
+            self.cpp_info.system_libs.extend(["synchronization", "winmm"])
+            self.cpp_info.defines.append("NOMINMAX")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.filenames["cmake_find_package"] = "dispenso"
+        self.cpp_info.filenames["cmake_find_package_multi"] = "dispenso"
+        self.cpp_info.names["cmake_find_package"] = "dispenso"
+        self.cpp_info.names["cmake_find_package_multi"] = "dispenso"
```
uilianries, line 124:
```suggestion
```

Two weeks ago I asked about it, and we can avoid for new recipes :)

recipes/dispenso/all/patches/1.3.0-001-unvendorize-concurrentqueue.patch
uilianries, line 1: I would add a personal note to send a PR to the upstream, adding an option to consume as external library :)

uilianries (dismissed):
Looking good, I would suggest only dropping cmake_find_package and we are good to go.

uilianries (approved):
LGTM

jwillikers (approved):

PR #22907, opened 2024-02-26
libboxes: add recipe for v0.1.1

### LINKED ISSUE #22906: [request] libboxes/0.1.1
### Package Name/Version

libboxes/0.1.1

### Webpage

https://gitlab.com/twdev_projects/boxes

### Source code

https://gitlab.com/twdev_projects/boxes

Boxes is a set of frequently used containers built on top of STL.

dagon666:
**libboxes/0.1.1**

I'm the author of the library.  The library is in its infancy ATM but I plan to actively contribute towards it.

fixes #22906
---

### DIFF
--- a/recipes/libboxes/all/conanfile.py
+++ b/recipes/libboxes/all/conanfile.py
@@ -1,0 +1,95 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, export_conandata_patches, get, rmdir
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.scm import Version
+import os
+
+class libboxesRecipe(ConanFile):
+    name = "libboxes"
+    description = "Boxes is a set of frequently used containers built on top of STL"
+
+    license = "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://gitlab.com/twdev_projects/boxes"
+
+    topics = ("container", "utility")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {"shared": [True, False], "fPIC": [True, False]}
+    default_options = {"shared": False, "fPIC": True}
+
+    @property
+    def _min_cppstd(self):
+        return 20
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "13",
+            "clang": "14",
+            "gcc": "11",
+        }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("xxhash/0.8.2", transitive_headers=True)
+
+    def validate(self):
+        if self.settings.os == "Windows":
+            raise ConanInvalidConfiguration("libboxes does not support Windows yet")
+
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["BUILD_EXAMPLES"] = "OFF"
+        tc.generate()
+
+        tc = CMakeDeps(self)
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE.md", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "libboxes")
+        self.cpp_info.set_property("cmake_target_name", "libboxes::libboxes")
+        self.cpp_info.set_property("pkg_config_name", "libboxes")
+
+        if self.settings.os in ["Linux", "Android", "FreeBSD", "SunOS", "AIX"]:
+            self.cpp_info.system_libs.append("m")
--- a/recipes/libboxes/all/conandata.yml
+++ b/recipes/libboxes/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "0.1.1":
+    url: "https://gitlab.com/twdev_projects/boxes/-/archive/v0.1.1/boxes-v0.1.1.tar.gz"
+    sha256: "a9943a47abd820b7847e1f7abd15fc733c028379542815f8038192a0e4902cd0"
--- a/recipes/libboxes/all/test_package/CMakeLists.txt
+++ b/recipes/libboxes/all/test_package/CMakeLists.txt
@@ -1,0 +1,10 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+find_package(libboxes CONFIG REQUIRED)
+
+
+
+add_executable(example src/example.cpp)
+target_link_libraries(example libboxes::libboxes)
+target_compile_features(example PRIVATE cxx_std_20)
--- a/recipes/libboxes/all/test_package/conanfile.py
+++ b/recipes/libboxes/all/test_package/conanfile.py
@@ -1,0 +1,26 @@
+import os
+
+from conan import ConanFile
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.build import can_run
+
+
+class libboxesTestConan(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps", "CMakeToolchain"
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def layout(self):
+        cmake_layout(self)
+
+    def test(self):
+        if can_run(self):
+            cmd = os.path.join(self.cpp.build.bindir, "example")
+            self.run(cmd, env="conanrun")
--- a/recipes/libboxes/all/test_package/src/example.cpp
+++ b/recipes/libboxes/all/test_package/src/example.cpp
@@ -1,0 +1,16 @@
+#include <boxes/boxes.hpp>
+
+#include <iostream>
+
+int main() {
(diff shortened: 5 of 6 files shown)


RubenRBS (commented):

As for the errors you're seeing, have you taken a look into the [cmake package template](https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py)? It shows a prototypical cmake recipe which avoid some of the common pitfalls you're seeing :)

Let me know if you need further help with this, happy to help!

dagon666:
I'm a bit unsure why I'm unable to reproduce this error locally?  I'm trying in docker using conanio/gcc11 image.  I checked and xxhash is definitely not installed globally.  The only location it can be found is conan cache:

`/home/conan/.conan/data/xxhash/0.8.2/_/_/package/0f64e1d06b9d83ce351cf714fe17bfcfbc6f7722/include/xxhash.h`

The include path seems to be incorrect for some reason.

uilianries:
>  I'm trying in docker using conanio/gcc11 image. 

@dagon666 please, use the docker image `conanio/gcc11-ubuntu16.04` instead. We are using this image in the CI.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("xxhash/0.8.2", transitive_headers=True)
+        self.test_requires("gtest/1.14.0")
```
uilianries: We really avoid running unit test from the recipes in CCI, the idea packaging the project only, not using Conan as a build tool.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+        # in case it does not work in another configuration, it should validated here too
+        if is_msvc(self) and self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
+
+    def build_requirements(self):
+        self.tool_requires("cmake/3.27.9")
+
```
uilianries:
```suggestion
```

The project says it needs Cmake 3.8 installed. The CI will provide CMake 3.15.

https://gitlab.com/twdev_projects/boxes/-/blob/master/CMakeLists.txt?ref_type=heads#L1

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+        cmake = CMake(self)
+        cmake.install()
+
+        # some files extensions and folders are not allowed. Please, read the FAQs to get informed.
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.la", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
```
uilianries, line 112: Please, only keep what is used. The template covers a big number of cases, but most of project does not use all.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_find_mode", "both")
```
uilianries, line 115:
```suggestion
```
Unless there is a real case related, we avoid providing both.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+        self.cpp_info.set_property("cmake_file_name", "libboxes")
+        self.cpp_info.set_property("cmake_target_name", "libboxes::libboxes")
+        self.cpp_info.set_property("pkg_config_name", "libboxes")
+
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs.append("pthread")
+
```
uilianries, line 122:
```suggestion
```

Sorry, but I can not find any usage for threads in the source, only in examples, which should not be part of the package:

https://gitlab.com/twdev_projects/boxes/-/blob/master/src/hashes.cpp?ref_type=heads

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rm, rmdir
```
uilianries, line 5:
```suggestion
from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rm, rmdir
```

Please, keep only imports that are used.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.microsoft import check_min_vs, is_msvc, is_msvc_static_runtime
```
uilianries, line 7:
```suggestion
from conan.tools.microsoft import is_msvc
```

Please, keep only used imports.

uilianries (commented):
Thank you for your PR! Please, take a look in my review.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+        # in case it does not work in another configuration, it should validated here too
+        if is_msvc(self) and self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
```
uilianries, line 66:
```suggestion
        if is_msvc(self) and self.options.shared:
            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
```

Maybe passing `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`? 

recipes/libboxes/all/conanfile.py
twisniewski2222: Hi @uilianries.  Thanks for the review.  This is something unclear to me so, would appreciate some more details.  Should I disable building and packaging all tests/benchmarks in conanfile.py?  I'm not sure if I should have two versions of conanfile.py - the one submitted to conan-center-index - with no test dependencies and a separate one for local use (whatever that might be) - which explicitly defines all dependencies?  Thanks in advance.

recipes/libboxes/all/conanfile.py
twisniewski2222: Will remove.  Thank you!

recipes/libboxes/all/conanfile.py
twisniewski2222: Should I remove tests/benchmarks/examples as well here?  I'm really not sure about it.

recipes/libboxes/all/conanfile.py
twisniewski2222: Will remove.  Thanks!

recipes/libboxes/all/conanfile.py
twisniewski2222: Will remove.  Thanks!

recipes/libboxes/all/conanfile.py
twisniewski2222: Will remove.  Thanks!

recipes/libboxes/all/conanfile.py
twisniewski2222: TBH I don't want to support Windows at all - I wasn't really sure how to express that in conanfile.py so, introduced these rules from the templates.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # In case there are dependencies listed on build_requirements, VirtualBuildEnv should be used
+        tc = VirtualBuildEnv(self)
+        tc.generate(scope="build")
+
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+        # remove bundled xxhash
+        # rm(self, "whateer.*", os.path.join(self.source_folder, "lib"))
+        # replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "...", "")
```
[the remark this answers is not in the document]
twisniewski2222: Will remove.  Thanks!

recipes/libboxes/all/conanfile.py
uilianries: Totally fine in case not support Windows! In that you could update the rule to:

```
def validate(self):
        raise ConanInvalidConfiguration(f"{self.ref} is not supported on Windows.")
```

Of course, it will restrict to any compiler, including who is using GCC (msys2) or Clang.

uilianries:
Hello @dagon666 I created a merge request to your project: https://gitlab.com/twdev_projects/boxes/-/merge_requests/1/

It should improve your CMakeLists.txt a bit and would make easier to use here in Conan side.

recipes/libboxes/all/conanfile.py
uilianries: Yes, please.

dagon666:
Is there anything outstanding on my side that has to be done to progress this MR?

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+from conan.tools.scm import Version
+import os
+
+class libboxesRecipe(ConanFile):
+    name = "libboxes"
+    description = "Boxes is a set of frequently used containers built on top of STL"
+
+    license = "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://boxes-twdev-projects-ba06c75b02a9b24f40d3a6b9b7130a3266c6a3c532.gitlab.io/"
```
valgur, line 15:
```suggestion
    homepage = "https://gitlab.com/twdev_projects/boxes"
```
I think the GitLab repo serves as a slightly better homepage. The Doxygen docs are easy to find from there.

recipes/libboxes/all/conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+        return 20
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "13",
+            "clang": "14",
+            "gcc": "11",
+            "msvc": "192",
+            "Visual Studio": "16.2",
```
valgur, line 34:
```suggestion
            "Visual Studio": "16",
```
You will have to use
```python
def loose_lt_semver(v1, v2):
    return all(int(p1) < int(p2) for p1, p2 in zip(str(v1).split("."), str(v2).split(".")))
```
for the version check to not exclude compler.version=16 otherwise.

valgur (approved):
LGTM, except for the minor Visual Studio version check issue.

recipes/libboxes/all/conanfile.py
valgur: Oh, you can drop VS and MSVC from this list entirely since you don't support Windows anyway.

dagon666: Removed. Thanks

recipes/libboxes/all/conanfile.py
dagon666: Changed as suggested.

valgur (approved):
Thanks!

uilianries (approved):
LGTM according to https://gitlab.com/twdev_projects/boxes/-/blob/v0.1.1/CMakeLists.txt

Thank you!

PR #22917, opened 2024-02-27
soplex: add version 7.0.0

hedtke:
Specify library name and version:  **soplex/7.0.0**

---

### DIFF
--- a/recipes/soplex/all/conanfile.py
+++ b/recipes/soplex/all/conanfile.py
@@ -94,6 +94,7 @@ class SoPlexConan(ConanFile):
         tc.variables["GMP"] = self.options.with_gmp
         tc.variables["BOOST"] = self.options.with_boost
         tc.variables["Boost_VERSION_MACRO"] = "108400"
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
         tc.generate()
         deps = CMakeDeps(self)
         if self.options.with_gmp:
--- a/recipes/soplex/all/conanfile.py
+++ b/recipes/soplex/all/conanfile.py
@@ -73,7 +73,7 @@ class SoPlexConan(ConanFile):
             # see https://github.com/conan-io/conan-center-index/pull/16017#issuecomment-1495688452
             self.requires("gmp/6.3.0", transitive_headers=True, transitive_libs=True)
         if self.options.with_boost:
-            self.requires("boost/1.83.0", transitive_headers=True)  # also update Boost_VERSION_MACRO below!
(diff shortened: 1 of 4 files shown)


uilianries:
We have few cases where need to pass `bigobj`, you should take a look in `harfbuzz` and `cgal` recipe, but basically it's something to fixed in the recipe, not in the test package, otherwise, the package will be broken when consuming.

recipes/soplex/all/conanfile.py
```diff
@@ -90,9 +90,10 @@ def source(self):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.variables["MPFR"] = False
         tc.variables["GMP"] = self.options.with_gmp
         tc.variables["BOOST"] = self.options.with_boost
-        tc.variables["Boost_VERSION_MACRO"] = "108300"
+        tc.variables["Boost_VERSION_MACRO"] = "108400"
```
valgur, line 96:
```suggestion
        tc.variables["Boost_VERSION_MACRO"] = "108400"
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```

The CMake policy version in the project is 3.9 and needs CMP0077 to be enabled for `tc.variables` to work correctly.

hedtke: added

hedtke:
@uilianries I gave it a go to add `/bigobj` to the recipe, but that failed. Any idea what I am doing wrong? I am not very experienced with the Windows stack.

hedtke:
I checked the CMakeLists of soplex. It already enables bigobj: https://github.com/scipopt/soplex/blob/b23232dd27969e585be201487771bc102f0a223f/CMakeLists.txt#L216
That does not seem to be the problem. Do I need to forward this setting to all consumers? How do I do that?

uilianries:
@hedtke Thank you for spotting it! In that case you are right, better pass it directly to the test package then, it should fix the situation and follow the upstream behavior. Sorry about the confusion.

hedtke:
Changed back: `/bigobj` is used only in tests

valgur (approved):

hedtke:
Can I ping here? SCIP 9.0.0 was released in February and depends on this lib. I cannot update SCIP until this update is done. Thank you for your help

AbrilRBS (approved):
Thanks! note that closing and opening the PR caused it to be tagged as "Done" so we lost track of it internally, sorry aboutthe delay! (If you need to force a recompilation, the "Update branch" button usually works best, if not, changing the title also does the trick for now!)

PR #22920, opened 2024-02-27
dbus: Bump expat and glib dependencies to minimize conflicts

jwillikers:
The expat dependency must be bumped to not conflict with wayland. Additionally, glib/2.78.1 and glib/2.78.3 are used much more than glib/2.77.0. I went with glib/2.78.3 since it is newer and also required in the Qt packages. Bump build dependencies as well.

Blocks #22387.

---

### DIFF
--- a/recipes/dbus/1.x.x/conanfile.py
+++ b/recipes/dbus/1.x.x/conanfile.py
@@ -67,9 +67,9 @@ class DbusConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("expat/2.5.0")
+        self.requires("expat/2.6.0")
         if self.options.with_glib:
-            self.requires("glib/2.77.0")
+            self.requires("glib/2.78.3")
         if self.options.get_safe("with_systemd"):
             self.requires("libsystemd/253.6")
         if self.options.with_selinux:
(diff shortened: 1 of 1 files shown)


recipes/dbus/1.x.x/conanfile.py
```diff
@@ -67,9 +67,9 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("expat/2.5.0")
+        self.requires("expat/2.6.0")
```
uilianries, line 70: It will break Qt5 and Qt6. Both are using expat 2.5.0 and require dbus.

jwillikers: @uilianries  Wayland also currently breaks them... just not be default with Wayland disabled. Would you rather I downgrade expat in Wayland to 2.5.0? Or open PRs to bump dependencies in Qt 5 and 6?

ericLemanissier: xkbcommon requires expat 2.6.0, so qt is already using expat/2.6.0 (except if qtwebengine is activated, then there is already a conflict with expat : https://github.com/conan-io/conan-center-index/blob/41876ae1b94e61e249088e43ffcf551be3c9643f/recipes/qt/6.x.x/conanfile.py#L370-L371)

jwillikers: I went ahead and opened #22929 to bump expat to 2.6.0 in Qt.

jwillikers: #22929 has been merged, so this should be good to go now.

RubenRBS (approved):
I'll generate the missing binaries this'd create after getting merged :)

PR #22926, opened 2024-02-28
liblzf: avoid WINDOWS_EXPORT_ALL_SYMBOLS

valgur:
Can use a `.def` file instead since the library has an extremely simple API.


RubenRBS (approved):
Looks good, thanks! :)

jwillikers (approved):

PR #22927, opened 2024-02-28
urdfdom_headers: new recipe

valgur:
Unbundles `urdfdom_headers` from the `urdfdom` recipe, so they can be used directly and versioned separately.


uilianries (approved):
LGTM

jwillikers (approved):

PR #22939, opened 2024-03-01
tcl: Don't include extensions in package info libs

Ahajha:
Specify library name and version:  **tcl/all**

Alternative to https://github.com/conan-io/conan-center-index/pull/22711.

TCL has some extra libraries that it uses just for extensions. I believe these should not be included in the package info, as there is no point in linking against them. They also were not properly fixed by `fix_apple_shared_install_name`, which was causing issues in #21387. This PR fixes this by just removing those extra libraries from the package info altogether.

---


RubenRBS (approved):
After looking into this (Was on holiday last week sorry for the delay!), this seems like a better approach! Thanks a lot :)

jwillikers (approved):

PR #22945, opened 2024-03-01
[openssl] Set tls security level to automatic for OpenSSL 3.x

uilianries:
Specify library name and version:  **openssl/3.x**

The PR #21708 added the new option `tls_security_level` which is capable to configure a OpenSSL property. It's configured according to current versions available for 3.x, but OpenSSL moves fast it can be changed in the future, resulting in a misconfiguration in a solitary bump version PR.

Thanks to @jcar87 for pointing it, this PR keeps the current option, but instead, uses the default value from OpenSSL when not configured to an explicit value.

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -88,13 +88,13 @@ class OpenSSLConan(ConanFile):
         "no_whirlpool": [True, False],
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
-        "tls_security_level": [0, 1, 2, 3, 4, 5],
+        "tls_security_level": [None, 0, 1, 2, 3, 4, 5],
     }
     default_options = {key: False for key in options.keys()}
     default_options["fPIC"] = True
     default_options["no_md2"] = True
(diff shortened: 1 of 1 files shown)


uilianries:
Ops, Conan 1.x is not so smart. Gonna check it now.

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -386,7 +384,8 @@ def _configure_args(self):
 
         args.append("no-fips" if self.options.get_safe("no_fips", True) else "enable-fips")
         args.append("no-md2" if self.options.get_safe("no_md2", True) else "enable-md2")
-        args.append("-DOPENSSL_TLS_SECURITY_LEVEL=%s" % str(self.options.tls_security_level))
+        if str(self.options.tls_security_level) != "None":
```
uilianries, line 387: Both None and 0 are negative boolean values. Plus, `self.options.tls_security_level` is a Conan object, it's converted according to the operation needed.

danimtb (approved):

jcar87 (approved):

PR #22950, opened 2024-03-01
add basu/0.2.1

bobrofon:
Specify library name and version:  **basu/0.2.1**

basu is the sd-bus library, extracted from systemd.

Some projects rely on the sd-bus library for DBus support. However not all systems have systemd or elogind installed. This library provides just sd-bus (and the busctl utility).

---

### DIFF
--- a/recipes/basu/all/conandata.yml
+++ b/recipes/basu/all/conandata.yml
@@ -1,0 +1,11 @@
+sources:
+  "0.2.1":
+    url: "https://git.sr.ht/~emersion/basu/refs/download/v0.2.1/basu-0.2.1.tar.gz"
+    sha256: "d9b373a9fcb5d5eb5f6c1c56355f76edb7f2f52bc744570e80604e83455a19bd"
+
+patches:
+  "0.2.1":
+    - patch_file: "patches/0001-memfd-util.c-Use-F_ADD_SEALS-definition-from-missing.patch"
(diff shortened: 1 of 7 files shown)


RubenRBS (commented):
Thanks a lot for your contribution! My first question here is, given that this comes from systemd, is the current recipe providing this already? Would it make sense to have a conflict if both are present as requirements?

The rest of the recipe looks good :D

bobrofon:
> given that this comes from systemd, is the current recipe providing this already?

Well, yes. Kind of. basu is a subset of libsystemd. basu exports the same symbols, and it's probably binary compatible with libsystemd (but there are no guarantees). The major difference is that basu supports FreeBSD, and libsystemd doesn't. I think that the main usage of basu recipe will be something like this:
```python
if self.settings.os == 'Linux':
    self.requires('libsystemd')
elif self.settings.os == 'FreeBSD':
    self.requires('basu')
```

> Would it make sense to have a conflict if both are present as requirements?

Good question. Imagine the following situation. You have a package with two binaries (let's call them 'client' and 'server'). The server links with libsystemd (by historical reasons). And the client links with basu (maybe because basu is already it's indirect dependency). And the client communicates with the server using dbus. In this situation I don't see why it should be prohibited on the recipe level.

valgur (approved):

danimtb (approved):

PR #22952, opened 2024-03-01
libselinux: Fix cross-compilation

jwillikers:
Apply patches for 3.6 to fix detection of available functions when cross-compiling. Use the `PKG_CONFIG_LIBDIR` to point pkg-config to the correct directory containing Conan's generated pkg-config files. Without this, the `pcre2` from the system will be picked up before the one provided by Conan. When cross-compiling, this causes system include directories to poison the build.

[libselinux-3.6-cross-fixed-log.txt](https://github.com/conan-io/conan-center-index/files/14743730/libselinux-3.6-cross-fixed-log.txt)
[libselinux-3.6-cross-broken-log.txt](https://github.com/conan-io/conan-center-index/files/14743733/libselinux-3.6-cross-broken-log.txt)


valgur (approved):

recipes/libselinux/all/conanfile.py
```diff
@@ -87,7 +87,9 @@ def generate(self):
         sepol_lib_folder = os.path.join(self._sepol_source_folder, "src")
         tc.extra_ldflags.append(f"-L{sepol_lib_folder}")
         tc.make_args.append("USE_PCRE2=y")
-        tc.generate()
+        env = tc.environment()
+        env.append_path("PKG_CONFIG_LIBDIR", self.generators_folder)
+        tc.generate(env=env)
```
franramirez688, line 92: Should we apply this only for cross-compilations?

franramirez688 (commented):
Hi @jwillikers Thanks a lot for this PR (and the associated issue 😁 )

Sorry for the huge delay! I'm just revisiting all my old assigned issues! 🙏 

LGTM! I only have a question regarding to the environment.

recipes/libselinux/all/conanfile.py
jwillikers: I don't see why not. I've updated this.

recipes/libselinux/all/conanfile.py
```diff
@@ -87,7 +88,10 @@ def generate(self):
         sepol_lib_folder = os.path.join(self._sepol_source_folder, "src")
         tc.extra_ldflags.append(f"-L{sepol_lib_folder}")
         tc.make_args.append("USE_PCRE2=y")
-        tc.generate()
+        if cross_building(self):
+            env = tc.environment()
+            env.append_path("PKG_CONFIG_LIBDIR", self.generators_folder)
+            tc.generate(env=env)
```
franramirez688, line 94:
```suggestion
        env = tc.environment()
        if cross_building(self):
            env.append_path("PKG_CONFIG_LIBDIR", self.generators_folder)
        tc.generate(env=env)
```

franramirez688 (approved):

PR #22960, opened 2024-03-02
[fastgltf] Add versions 0.7.2, 0.8.0

spnda:
Specify library name and version:  **fastgltf/0.7.1**

Just a bugfix patch with minor additions, and I've bumped the simdjson version to 3.7.0 in this recipe because I think there's nothing wrong with telling it to use the latest version directly?

### Maintainer edits
* Keep only 0.8.0 and the most recent 0.7.x
* Remove patches and logic in the recipe for versions older than the two mentioned
* Refactor logic to use `check_min_cppstd()`
* Remove test_v1_package
* Refactor test package to be as minimal as possible


recipes/fastgltf/all/conanfile.py
```diff
@@ -67,7 +67,11 @@ def layout(self):
         cmake_layout(self, src_folder='src')
 
     def requirements(self):
-        self.requires("simdjson/3.2.0")
+        # See https://github.com/spnda/fastgltf/commit/aa90dfac290584f35f47dd7c1634f23d376a7207
+        if Version(self.version) >= "0.6.0":
+            self.requires("simdjson/3.7.0")
+        else:
+            self.requires("simdjson/3.2.1")
```
valgur, line 74:
```suggestion
        self.requires("simdjson/3.7.0")
```
Are the older versions incompatible with the latest simdjson?

spnda: Yes. I originally had what you are suggesting, but removed that since it caused build failures on older versions.

Version 3.2.2 moved around a lot of headers and includes, which broke my base64 decoders as they didn't include the headers properly, and the code only worked because `simdjson.h` leaked the SIMD intrinsics. I fixed that in 0.6.0, which is why all versions since then have supported any of the more recent versions of simdjson. See https://github.com/simdjson/simdjson/pull/2039.

spnda:
I have updated this PR to now also include version data for 0.7.2 and 0.8.0. I didn't want to open a separate PR, so I've just included this here.

@toge I saw #24168 and I would be totally fine with removing old versions of fastgltf, which would allow you to remove all of the old versions of simdjson. Only 0.5.0 and earlier required those versions of simdjson, and 0.5.0 came out in May 2023 so it can definitely be viewed as legacy in my opinion.

recipes/fastgltf/all/conanfile.py
spnda: As noted in https://github.com/conan-io/conan-center-index/pull/22960#issuecomment-2294509192, I would be happy to remove these legacy versions of fastgltf since they're pretty old and include a large number of issues. And I don't think anybody actually uses them anymore.

jcar87:
Thanks @spnda  - I've amended the PR to only add `0.7.2` and `0.8.0`, while stopping recipe support for any older version - hopefully this gets simplifies the case to be able to use the most recent version of simdjson.

I've rebased the branch on the most recent master - so if you work with this locally please bear that in mind (you have to resync with the branch in the remote before attempting any push/pull).

Thanks!

spnda:
@jcar87 Thank you. I had suggested this but never got an answer if this was the right thing to do.

jcar87 (approved):

PR #22970, opened 2024-03-04
pango: Fix some component requires

jwillikers:
Several of the requirements are incomplete. This adds at least a few of those missing requirements.

Bump fontconfig to fix dependency conflict with Qt and fltk.

Requires:
- [x] #22696
- [x] #23511


AbrilRBS (commented):
Thanks @jwillikers for the PR and your patience. I have one minor request before merging :)

(Also, as these are non-default options, which is the reason these bugs snuck in the first place, it would be great to have compilation logs, both sucessful post patch, and failing ones before hand would be great!)

PR #22971, opened 2024-03-04
Qt6 check if we are on C3I CI

jcar87:
Specify library name and version:  **qt/6.x**

Check if global environment variable `CONAN_CENTER_BUILD_SERVICE` is defined. This negates the need for consumers to have to define `NOT_ON_C3I`.


RubenRBS (approved):

ericLemanissier (approved):
great !

uilianries (approved):
LGTM

PR #22974, opened 2024-03-04
freetype/2.13.*: Switch to Meson build system

jwillikers:
This requires way less patching. Way less.
Fix the version in the pkg-config file to correctly map to the API version.
Update the font used in the test package and embed the OFL license in the font as suggested in the OFL FAQ.
Add this license to the license attribute of the test packages to make this licensing clear.

#23010 should be merged immediately after this PR to accommodate the correct API version being used now.

No longer blocked by conan-io/hooks#516.

In order to work for CMake versions before 3.29, the Meson generated static library on Windows is renamed from `libfreetype.a` to `freetype.lib`. CMake 3.29 and later properly detect static libraries generated by Meson on Windows for MSVC, i.e. static libraries suffixed with `.a` instead of `.lib`. This change is documented in the [CMake 3.29 release notes](https://cmake.org/cmake/help/latest/release/3.29.html):

> On Windows, when targeting the MSVC ABI, the [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) command now accepts .a file names after first considering .lib. This is symmetric with existing behavior when targeting the GNU ABI, in which the command accepts .lib file names after first considering .a.


jwillikers:
This looks suspiciously like a long paths issue. I can build successfully on Windows locally.

```
Installing libfreetype.a to C:/J2/w/prod-v2/bsr@3/101217/eacfc/p/b/freet613ea5253c3ff/p\lib
```

jwillikers:
Blocked by conan-io/hooks#516

recipes/freetype/meson/test_package/conanfile.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def test(self):
+        if can_run(self):
+            bin_path = os.path.join(self.cpp.build.bindirs[0], "test_package")
+            font_path = os.path.join(self.source_folder, "OpenSans-Bold.ttf")
```
Croydon, line 26: OpenSans License should be mentioned somewhere (both in `all/` and `meson/`)

jwillikers: It might be acceptable to just make sure the license is embedded in the font: https://openfontlicense.org/how-to-use-ofl-fonts/

jwillikers:
I can't recreate this locally on Windows. It builds fine for me.

jwillikers:
Aha, the issue is in the test_module_package instead of the test_package.

jwillikers:
Only happens for release builds. It's caused by CMake assuming that the Freetype library will always have a `d` debug postfix. I hate debug postfix letters.

jwillikers:
I'm starting to get errors with the test package when using CMake 3.28 from the virtualbuildenv. My local testing with CMake 3.29-rc2 works. Doesn't fix the error with the test_package_module...

jwillikers:
Yup, I found the pertinent item in the [CMake 3.29 release notes](https://cmake.org/cmake/help/latest/release/3.29.html):

```
On Windows, when targeting the MSVC ABI, the [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) command now accepts .a file names after first considering .lib. This is symmetric with existing behavior when targeting the GNU ABI, in which the command accepts .lib file names after first considering .a.
```

recipes/freetype/meson/conanfile.py
```diff
@@ -0,0 +1,245 @@
...
+            conan_exec_prefix=${{conan_prefix}}/bin
+            conan_includedir=${{conan_prefix}}/include
+            conan_libdir=${{conan_prefix}}/lib
+            conan_ftversion={version}
+            conan_staticlibs="{staticlibs}"
+        """).format(version=version, staticlibs=staticlibs))
+
+    def _extract_libtool_version(self):
+        conf_raw = load(self, os.path.join(self.source_folder, "builds", "unix", "configure.raw"))
+        return next(re.finditer(r"^version_info='([0-9:]+)'", conf_raw, flags=re.M)).group(1).replace(":", ".")
```
valgur, line 140:
```suggestion
        return re.search(r"^version_info='([0-9:]+)'", conf_raw, flags=re.M).group(1).replace(":", ".")
```

valgur (approved):

valgur (approved):

danimtb (approved):

franramirez688 (changes requested):
Thanks a lot for this, @jwillikers! Blocking only to let me review it calmly 😁 Just a few minutes 🙏

franramirez688 (approved):
LGTM! 👏

jellespijker:
This PR broke our builds, due to the version obtained from the raw file, see: https://github.com/Ultimaker/cura-workflows/actions/runs/8798472364/job/24145560363#step:16:27993
When building fontconfig it searches for freetype2 version 2.13.2 but the generated freetype2.pc uses the version extracted from the textfile which is 26.1.20

Is there a quick fix for this? It's blocking our release of Cura 5.7.1, current approach is setting the `component_version` to `self.version in the `package_info` and then using that in our builds, but I don't know if this is the "proper" way of doing that

ericLemanissier:
yes: https://github.com/conan-io/conan-center-index/pull/23010

jwillikers:
> This PR broke our builds, due to the version obtained from the raw file, see: https://github.com/Ultimaker/cura-workflows/actions/runs/8798472364/job/24145560363#step:16:27993 When building fontconfig it searches for freetype2 version 2.13.2 but the generated freetype2.pc uses the version extracted from the textfile which is 26.1.20
> 
> Is there a quick fix for this? It's blocking our release of Cura 5.7.1, current approach is setting the `component_version` to `self.version` in the `package_info` and then using that in our builds, but I don't know if this is the "proper" way of doing that

@jellespijker Yes, sorry about breaking that. I tried to make it as clear as possible to the team that fixing the pkg-config version would break the `fontconfig` package and require merging my PR #23010 ASAP.

PR #22982, opened 2024-03-05
abseil: Add 20240116.1, remove old versions

Ahajha:
Specify library name and version:  **abseil/20240116.1**

I've removed some of the older versions, mainly out of date "patch" versions (for example, keeping 20230125.3 but removing .2/.1/.0).

---


Ahajha:
Note to self: I don't like the way the edge case is handled, I think it should either be an invalid configuration, or just patch the problem.

RubenRBS:
Just a heads up that `tensorflow-lite` uses one of the removed options

RubenRBS:
I've asked the current tensorflow PR to do that for you :)

Ahajha:
I've reported the GCC7/C++17 issue upstream here: https://github.com/abseil/abseil-cpp/issues/1635. I'm hoping it's a simple fix that could be backported, as that would be the simplest solution rather than the hack I have in place.

Ahajha:
There was a std::result_of workaround (it was removed in C++20) added in #6383, it was fixed as of https://github.com/abseil/abseil-cpp/commit/e19260fd7dbef881492fd73891e0be5bd4a09b95, which was late 2020. None of the versions at this point need these extra defines to work. I also tested locally with an MSVC/C++20 build.

Ahajha:
This should be good to go now, we now have the proper upstream fix backported so no need for any hacky workarounds.

RubenRBS (approved):
Approving to ghet the new version out asap, will bump tensorflow in a bit :) Thanks a lot

franramirez688 (approved):

PR #22990, opened 2024-03-05
ffmpeg: Add missing dependency on OpenSSL for avutil

jwillikers:
This fixes missing symbols errors at link time when using the component dependency `ffmpeg::avutil`.

Also, use `libva` and `libvdpau` instead of the system packages since they are now available as proper Conan package. This prevents conflicts in consumers.

I guess I had to get one more thing in. I've added a `with_libdrm` option as well to include `libdrm`.


jwillikers:
Well, I'm really not sure how I would have cause this error that is showing up on Windows:

```
/c/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.29.30133/bin/HostX64/x64/link: Argument list too long
```

RubenRBS:
Building missing binaries now

MartinDelille (approved):

jcar87:
> This fixes missing symbols errors at link time when using the component dependency ffmpeg::avutil.

Do we have an output of the actual missing symbol errors?

uilianries:
@jwillikers What do you mean by `pipewire`? What exactly scenario could I use to reproduce that error? Could you please share the recipe and commands used?

uilianries:
BTW, I see this PR is more or less following the changes applied in https://github.com/conan-io/conan-center-index/pull/23089/files#diff-13ca8667b24368a779b16b2c336319a64d00888f52929213802d3572e58828bc

jwillikers:
@uilianries See PR #22991

uilianries:
I rebased this PR over the master

recipes/ffmpeg/all/conanfile.py
```diff
@@ -902,11 +908,20 @@ def _add_component(name, dependencies):
...
+            if self.options.get_safe("with_libdrm"):
+                avutil.requires.append("libva::libva-drm")
+            if self.options.get_safe("with_xcb"):
+                avutil.requires.extend(["libva::libva-x11", "xorg::x11"])
 
         if self.options.get_safe("with_vdpau"):
-            avutil.requires.append("vdpau::vdpau")
+            avutil.requires.append("libvdpau::libvdpau")
+
+        if self.options.with_ssl == "openssl":
```
uilianries, line 923: I can confirm this fix working, I tested locally on Linux. I would ask to split this PR. Those extra dependencies will require extra steps to be validated, and replacing packages is even more fragile as ffmpeg is a critical package.

recipes/ffmpeg/all/conanfile.py
```diff
@@ -311,14 +314,14 @@ def requirements(self):
...
         if self.options.get_safe("with_libalsa"):
             self.requires("libalsa/1.2.10")
-        if self.options.get_safe("with_xcb") or self.options.get_safe("with_vaapi"):
+        if self.options.get_safe("with_xcb"):
             self.requires("xorg/system")
         if self.options.get_safe("with_pulse"):
             self.requires("pulseaudio/14.2")
         if self.options.get_safe("with_vaapi"):
-            self.requires("vaapi/system")
+            self.requires("libva/2.20.0")
```
uilianries, line 322: libva looks a good replacement. 👍 

I just compared both packages and they look equivalent. The only possible problem that may cause is breakage for few users, as it is closer to hardware and the package may be incompatible. Of course, rebuilding from source should be enough.

About libva distributed from AUR (Arch Linux):

- https://archlinux.org/packages/extra/x86_64/libva/
- https://gitlab.archlinux.org/archlinux/packaging/packages/libva/-/blob/main/PKGBUILD?ref_type=heads
- https://archlinux.org/packages/extra/x86_64/libva/files/

About libva distributed from APT (Ubuntu):

- https://packages.ubuntu.com/focal/amd64/libva-dev/filelist
- https://git.launchpad.net/ubuntu/+source/libva/tree/debian/control?h=applied/ubuntu/devel

Both packages look be very similar with the Conan libva package, in terms of headers and libraries.

recipes/ffmpeg/all/conanfile.py
```diff
@@ -311,14 +314,14 @@ def requirements(self):
...
         if self.options.get_safe("with_vdpau"):
-            self.requires("vdpau/system")
+            self.requires("libvdpau/1.5")
```
uilianries, line 324: Same for libvdpau, close to hardware, may cause incompatibility, but still looks following the same implementation as from distros:

AUR:
- https://archlinux.org/packages/extra/x86_64/libvdpau/
- https://archlinux.org/packages/extra/x86_64/libvdpau/files/
- https://gitlab.archlinux.org/archlinux/packaging/packages/libvdpau/-/blob/main/PKGBUILD?ref_type=heads

APT:
- https://git.launchpad.net/ubuntu/+source/libvdpau/tree/debian/rules?h=ubuntu/jammy

recipes/ffmpeg/all/conanfile.py
jwillikers: I've reverted those changes to use the `libva` and `libvdpau`.

recipes/ffmpeg/all/conanfile.py
```diff
@@ -902,11 +908,18 @@ def _add_component(name, dependencies):
             if Version(self.version) >= "5.0" and is_apple_os(self):
                 avfilter.frameworks.append("Metal")
 
+        if self.options.get_safe("with_libdrm"):
+            avutil.requires.append("libdrm::libdrm_libdrm")
         if self.options.get_safe("with_vaapi"):
-            avutil.requires.extend(["vaapi::vaapi", "xorg::x11"])
+            avutil.requires.append("vaapi::vaapi")
+            if self.options.get_safe("with_xcb"):
+                avutil.requires.append("xorg::x11")
```
uilianries, line 916: Are you sure about this change? What evidence could be used? At line 317 there is no dependency between vaapi and xcb

jwillikers: It's been a while now, so I don't quite remember the exact details, but I remember this being an important dependency when enabling `vaapi`. It should be able to track down the dependency in the code for `avutil`.

uilianries: I mean, they were independent before. In case not having evidence, please, revert it. The FFMpeg is really critical, so you can open a new PR in the future, in case finding the reason with proper details.

uilianries:
The new CI result shows an error when libdrm is enabled: https://c3i.jfrog.io/c3i/misc-v2/logs/pr/22990/14-linux-gcc/ffmpeg/5.0.3//9e4a072e6a84bebd211fbe2e2d40a0d437ccab5d-build.txt

```
src/libavutil/hwcontext_vaapi.c: In function 'vaapi_map_from_drm':
src/libavutil/hwcontext_vaapi.c:1078:9: error: unknown type name 'VADRMPRIMESurfaceDescriptor'
 1078 |         VADRMPRIMESurfaceDescriptor prime_desc;
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
src/libavutil/hwcontext_vaapi.c:1084:34: error: 'VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2' undeclared (first use in this function); did you mean 'VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME'?
 1084 |                 .value.value.i = VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2,
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                  VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME
src/libavutil/hwcontext_vaapi.c:1084:34: note: each undeclared identifier is reported only once for each function it appears in
src/libavutil/hwcontext_vaapi.c:1093:19: error: request for member 'fourcc' in something not a structure or union
 1093 |         prime_desc.fourcc = va_fourcc;
      |                   ^
src/libavutil/hwcontext_vaapi.c:1094:19: error: request for member 'width' in something not a structure or union
 1094 |         prime_desc.width = src_fc->width;
```

This occurs because the Conan package `vaapi/system` is installing the version 1.7.0 in Ubuntu 16.04 (C3I Jenkins), which is too old for FFMpeg:

```
apt policy libva-dev
libva-dev:
  Installed: 1.7.0-1ubuntu0.1
  Candidate: 1.7.0-1ubuntu0.1
  Version table:
 *** 1.7.0-1ubuntu0.1 500
        500 http://archive.ubuntu.com/ubuntu xenial-updates/universe amd64 Packages
        100 /var/lib/dpkg/status
     1.7.0-1 500
        500 http://archive.ubuntu.com/ubuntu xenial/universe amd64 Packages
```

The required define `VA_SURFACE_ATTRIB_MEM_TYPE_DRM_PRIME_2` is only available for libva >=2.1: https://github.com/intel/libva/commit/51e98b1224794a44ba097baa7a1b4e35c3596d0c

I would recommend disabling `with_libdrm` by default, following the upstream behavior: https://git.ffmpeg.org/gitweb/ffmpeg.git/blob/refs/heads/release/6.1:/configure#l339

As alternative, we could use `libva/2.21.0`, as tried before, I but think is too risk as it may break users due its affinity to the system/hardware. Still, people can use [replace_requires](https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires) to use libva instead of vaapi Conan package. 

BTW, I can build the current commit cf8cc9a without error in Ubuntu 22.04 + Conan 2.5.0 + ffmpeg/5.0.3:with_libdrm=True:

[ffmpeg-5.0.3-ubuntu22.04-libdrm.log](https://github.com/user-attachments/files/16599142/ffmpeg-5.0.3-ubuntu22.04-libdrm.log)

uilianries (approved):
LGTM.

- with_libdrm requires libva >=2.1 due new defines: https://github.com/conan-io/conan-center-index/pull/22990#issuecomment-2286086219
- Conan has libva packaged already, but it depends on the OS and may break consumers: https://github.com/conan-io/conan-center-index/pull/22990#discussion_r1709228082
- with_vaapi is not directly connected to with_xcb, it was libva_x11 dependendy actually. Addressed by the PR #24879
- The OpenSSL is indeed missing for avutil, tested locally. The PR #22991 express the bug.

@jwillikers Thank you for your PR. Please, for future PRs related to ffmpeg, try not mixing things at same PR, try only one change per PR instead, because ffmpeg is not only complex but largely used by the community and any bug may affect several users.

AbrilRBS (approved):

PR #22991, opened 2024-03-05
pipewire: Add recipe

### LINKED ISSUE #18533: [request] pipewire/1.2.7
### Package Name/Version

pipewire/0.3.50

### Webpage

https://pipewire.org/

### Source code

https://gitlab.freedesktop.org/pipewire/pipewire

PipeWire is a server and user space API to deal with multimedia pipelines. It is becoming the de facto standard to handle audio and video in Linux systems.

jwillikers:
[PipeWire](https://pipewire.org/) is a project that aims to greatly improve handling of audio and video under Linux.

Fixes #18533.

Requires the following for optional ffmpeg support to be enabled.
This support won't be enabled by default, so it shouldn't be required to merge this PR.
Fyi, the support for `ffmpeg` should never be enabled by default because it will cause circular dependencies since `ffmpeg` requires `sdl` which now requires `pipewire`.

- [x] #22990


ericLemanissier (approved):

ericLemanissier (approved):

ericLemanissier (approved):

recipes/pipewire/all/conanfile.py
```diff
@@ -0,0 +1,439 @@
...
+        libspa_pc_file = glob.glob(
+            os.path.join(self.package_folder, "lib", "pkgconfig", "libspa-*.pc")
+        )[0]
+
+        libpipewire_api_version = os.path.splitext(
+            os.path.basename(libpipewire_pc_file)
+        )[0].split("-")[1]
+        libspa_api_version = os.path.splitext(os.path.basename(libspa_pc_file))[
+            0
+        ].split("-")[1]
```
valgur, line 275:
```suggestion
        pkconfig_dir = self.package_path.joinpath("lib", "pkgconfig")
        libpipewire_pc_file = next(pkconfig_dir.glob("libpipewire-*.pc"))
        libspa_pc_file = next(pkconfig_dir.glob("libspa-*.pc"))

        libpipewire_api_version = libpipewire_pc_file.stem.split("-")[1]
        libspa_api_version = libspa_pc_file.stem.split("-")[1]
```
Just a suggestion. The pathlib API is much nicer for such tasks.

recipes/pipewire/all/conanfile.py
```diff
@@ -0,0 +1,439 @@
...
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
+
+        fix_apple_shared_install_name(self)
+
+    def package_info(self):
+        libpipewire_api_version = load(self, self._libpipewire_api_version_txt).strip()
+        libspa_api_version = load(self, self._libspa_api_version_txt).strip()
+
+        self.runenv_info.prepend_path(
```
valgur, line 291: Are you sure that prepending/appending is correct here? I.e. would the library be able to use more than one path in the envvars?

jwillikers: Looking at what's documented, I bet you're on the right track there. I'll change this to define it.

jwillikers: Done.

danimtb (dismissed):

recipes/pipewire/all/conanfile.py
```diff
@@ -0,0 +1,431 @@
...
+        return {
+            "gcc": "8",
+        }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if os.getenv("CONAN_CENTER_BUILD_SERVICE") is not None:
+            self.options.with_xfixes = False
```
valgur, line 86: A source-code comment explaining the purpose of this would be nice.

recipes/pipewire/all/conanfile.py
```diff
@@ -0,0 +1,431 @@
...
+        tc.c_args.append("-Wno-error=strict-prototypes")
+        # This appears to be an issue that crops up when using Avahi and libpulse involving the malloc_info and malloc_trim functions.
+        tc.c_args.append("-Wno-error=implicit-function-declaration")
+        for dependency, _value in self.dependencies.direct_host.items():
+            if dependency.ref.name == "linux-headers-generic":
+                for includedir in self.dependencies.direct_host[
+                    "linux-headers-generic"
+                ].cpp_info.includedirs:
+                    tc.c_args.append(f"-I{includedir}")
+                break
```
valgur, line 189: Are the extra for loops really necessary?
```suggestion
        for includedir in self.dependencies["linux-headers-generic"].cpp_info.includedirs:
            tc.c_args.append(f"-I{includedir}")
```

recipes/pipewire/all/conanfile.py
jwillikers: Done.

recipes/pipewire/all/conanfile.py
```diff
@@ -0,0 +1,431 @@
...
+
+        self.runenv_info.define(
+            "PIPEWIRE_CONFIG_DIR", os.path.join(self.package_folder, "res", "pipewire")
+        )
+        self.runenv_info.define(
+            "PIPEWIRE_MODULE_DIR",
+            os.path.join(
+                self.package_folder, "lib", f"pipewire-{libpipewire_api_version}"
+            ),
+        )
```
valgur, line 287: Absolute nitpicking, but could you set the column limit to 100 or 120 for `black`, please?
The line wrapping is a bit excessive and only hurts readability, imo.

recipes/pipewire/all/conanfile.py
jwillikers: For some reason, I think that I had trouble with this a while ago. Oh, it came back to me. I did it this way because otherwise, using a `platform_requires` for `linux-headers-generic` will cause this to fail.

recipes/pipewire/all/conanfile.py
valgur: Thanks!

recipes/pipewire/all/conanfile.py
jwillikers: I reformatted with a longer line length with Ruff. It would be great to have formatting configured for the repository so no one has to think about this kind of thing.

Croydon: If your IDE supports .editorconfig then it is configured https://github.com/conan-io/conan-center-index/blob/2c3904ff13e23e1358e36197e97c90c2188138da/.editorconfig#L12

jacobfriedman:
Looks like this can be unblocked with bumping SELinux to 3.6. Who is maintaining this if not @jwillikers  - and why did the development stop?

danimtb:
@jacobfriedman unfortunatelly the recipe did not get to a good status of support for this library when the PR was opened. Some things have changed since then, as Conan 2.0 has added support for more features, and the CI of Conan Center was completely remodeled as well. I'd suggest opening a new PR for this library or at least leave a comment in the request issue linked in the description of the PR so we can track how important the inclusion of this library in Conan Center is and help us prioritize it. Thank you for your interest

PR #22983, opened 2024-03-05
re2: Add 20240301

Ahajha:
Specify library name and version:  **re2/20240301**

Depends on #22982

---

- [X] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [X] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [X] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #22995, opened 2024-03-05
fuzztest: new recipe

### LINKED ISSUE #22933: [request] fuzztest/cci.20240229
### Package Name/Version

cci.20240229

### Webpage

https://github.com/google/fuzztest

### Source code

https://github.com/google/fuzztest

FuzzTest is a C++ testing framework for writing and executing fuzz tests, which are property-based tests executed using coverage-guided fuzzing under the hood.  
It is our [Google's] latest fuzz testing technology and the successor of previously used fuzzing tools, such as [libFuzzer](https://llvm.org/docs/LibFuzzer.html). It allows you to write powerful fuzz tests more easily than with previously used [fuzz targets](https://llvm.org/docs/LibFuzzer.html#fuzz-target). You can use it together with [GoogleTest](https://google.github.io/googletest/), or other unit testing frameworks, allowing you to write fuzz test side by side with regular unit tests, and just as easily.

Ahajha:
Specify library name and version:  **fuzztest/cci.20240305**

Resolves #22933 

Depends on #22982, #22983

One issue that will come up is that fuzztest uses internal re2 headers ([see here](https://github.com/google/fuzztest/issues/91)). A possible fix is to just expose these headers in re2, but I don't think there's a good solution to this aside from a proper fix from upstream.

---

### DIFF
--- a/recipes/fuzztest/all/conandata.yml
+++ b/recipes/fuzztest/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "cci.20240305":
+    url: "https://github.com/google/fuzztest/archive/e317d5277e34948ae7048cb5e48309e0288e8df3.tar.gz"
+    sha256: "cd13071a90258e9017122903ef1cd61228d8beb3b99d15bd5c60f74379d2681e"
--- a/recipes/fuzztest/all/conanfile.py
+++ b/recipes/fuzztest/all/conanfile.py
@@ -1,0 +1,152 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, load, rm, rmdir, save
+from conan.tools.scm import Version
+import os
+
+import yaml
+
+required_conan_version = ">=1.53.0"
+
+#
+# INFO: Please, remove all comments before pushing your PR!
+#
+
+class _Library:
+    name: str
+    deps: list[str]
+
+class FuzztestConan(ConanFile):
+    name = "fuzztest"
+    description = "A C++ testing framework for writing and executing fuzz tests"
+    license = "Apache-2.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/google/fuzztest"
+    topics = ("fuzzing", "testing")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 17
+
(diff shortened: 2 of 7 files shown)


Ahajha:
@RubenRBS I'd like some feedback, the dependent abseil version bump is solvable, but the big problem is that this package uses internal re2 headers that aren't in that package. There's an issue for it (), but I don't expect it to be fixed anytime soon. As a potential hacky fix, we can add the following to re2's package:
```patch
diff --git a/recipes/re2/all/conanfile.py b/recipes/re2/all/conanfile.py
index 8d2e83b16..bb5455c37 100644
--- a/recipes/re2/all/conanfile.py
+++ b/recipes/re2/all/conanfile.py
@@ -2,7 +2,7 @@ from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeToolchain, CMakeDeps, cmake_layout
-from conan.tools.files import copy, get, rmdir
+from conan.tools.files import copy, get, replace_in_file, rmdir
 import os
 
@@ -86,6 +86,17 @@ class Re2Conan(ConanFile):
         deps.generate()
 
     def build(self):
+        # HACK: fuzztest uses these internal headers
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "set(RE2_HEADERS", """
+set(RE2_HEADERS
+    re2/prog.h
+    util/logging.h
+    re2/pod_array.h
+    re2/sparse_array.h
+    re2/sparse_set.h
+    re2/regexp.h
+    util/utf.h
+""")
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
@@ -94,6 +105,9 @@ class Re2Conan(ConanFile):
         cmake = CMake(self)
         cmake.install()
+        replace_in_file(self, os.path.join(self.package_folder, "include", "re2", "prog.h"), "util/logging.h", "re2/logging.h")
+        replace_in_file(self, os.path.join(self.package_folder, "include", "re2", "regexp.h"), "util/logging.h", "re2/logging.h")
+        replace_in_file(self, os.path.join(self.package_folder, "include", "re2", "regexp.h"), "util/utf.h", "re2/utf.h")
```

If we want a bit of safety, we could also add to the add of the top of each of those files something like
```c++
#ifndef CONAN_RE2_USE_INTERNAL_HEADERS
#error "These headers are internal! Define CONAN_RE2_USE_INTERNAL_HEADERS if you know what you're doing!"
#endif
```

Then define `CONAN_RE2_USE_INTERNAL_HEADERS` within the fuzztest package (and I think it would also need to be exposed to consumers).

Thoughts?

albuquerque:
@Ahajha Hi,

I have been trying to use your recipe as an inspiration to build my own, but personal use only, especially since it depends on re2 internal headers, and there doesn't seem to be any traction on that. Had to build my custom recipe for re2 that bundles the internal headers that fuzztest depends on.

However, despite all this, the test_pacakge for fuzztest fails as it depends on some missing symbols from centipede. Have you encountered this issue? Did you manage to resolve it?

Ahajha:
I don't recall running into that, no. Glad this was useful as a starting point. Kinda gave up on this since the internal headers issue is kinda tough, and review times in this repo have been abysmal as of late.

PR #22996, opened 2024-03-05
avahi: Fix conflicting definition of strlcpy from glibc

jwillikers:
#20706 causes an issue where newer versions of glibc include a definition for `strlcpy` in `/usr/include/string.h`. This results in the following error:

```
/home/jordan/.conan2/p/b/avahi98352b3d5ad69/b/src/avahi-common/domain.c:482:15: error: static declaration of ‘strlcpy’ follows non-static declaration
  482 | static size_t strlcpy(char *dest, const char *src, size_t n) {
      |               ^~~~~~~
In file included from /home/jordan/.conan2/p/b/avahi98352b3d5ad69/b/src/avahi-common/domain.c:24:
/usr/include/string.h:506:15: note: previous declaration of ‘strlcpy’ with type ‘size_t(char * restrict,  const char * restrict,  size_t)’ {aka ‘long unsigned int(char * restrict,  const char * restrict,  long unsigned int)’}
  506 | extern size_t strlcpy (char *__restrict __dest,
      |               ^~~~~~~
```

I was unable to coerce glibc to not include the function declaration for strlcpy. It's included in `/usr/include/string.h` in recent versions of glibc. At least version 2.38 of glibc.
Since I couldn't avoid the definition from glibc, I removed the static keyword to make the definitions the same. Avahi should still use its internal strlcpy in this situation.


uilianries:
@jwillikers Thank for reporting this case and sorry for that bug. Could you please open an issue to the upstream instead? It seems like this bug is more generic than a recipe fault and may affect not only users.

jwillikers:
> @jwillikers Thank for reporting this case and sorry for that bug. Could you please open an issue to the upstream instead? It seems like this bug is more generic than a recipe fault and may affect not only users.

@uilianries Sure, but I wouldn't expect most users to hit this issue since Avahi would use `strlcpy` from glibc by default, not even attempting to compile it's own `strlcpy` function which causes the error.

jwillikers:
I've opened avahi/avahi#579

jcar87:
is it really an issue with `avahi` upstream? 
If the change from this PR https://github.com/conan-io/conan-center-index/pull/20706 is forcing what would otherwise be autodetected - then upstream avahi probably has no problems whatsoever?

Edit: just saw https://github.com/conan-io/conan-center-index/pull/22996#issuecomment-1981053591 - would agree with this.

jwillikers:
> is it really an issue with `avahi` upstream? If the change from this PR #20706 is forcing what would otherwise be autodetected - then upstream avahi probably has no problems whatsoever?
> 
> Edit: just saw [#22996 (comment)](https://github.com/conan-io/conan-center-index/pull/22996#issuecomment-1981053591) - would agree with this.
> 
> Edit2: they were added to glibc recently https://news.ycombinator.com/item?id=36765747
> 
> I have a feeling that even if there was a way of avoiding the compiler from using the glibc variant, there may still be linker issues.

I have a feeling compiling it as a shared library might expose that problem.
Ideally we'd be have the glibc version available as a setting, then we could force it on for new enough versions of `glibc`. I think you're supposed to be able to disable the function using feature definitions, but I tried a bunch of things and couldn't disable it. That to say that there could be more than just the version of `glibc` at play here.

jcar87:
> > is it really an issue with `avahi` upstream? If the change from this PR #20706 is forcing what would otherwise be autodetected - then upstream avahi probably has no problems whatsoever?
> > Edit: just saw [#22996 (comment)](https://github.com/conan-io/conan-center-index/pull/22996#issuecomment-1981053591) - would agree with this.
> > Edit2: they were added to glibc recently https://news.ycombinator.com/item?id=36765747
> > I have a feeling that even if there was a way of avoiding the compiler from using the glibc variant, there may still be linker issues.
> 
> I have a feeling compiling it as a shared library might expose that problem. Ideally we'd be have the glibc version available as a setting, then we could force it on for new enough versions of `glibc`. I think you're supposed to be able to disable the function using feature definitions, but I tried a bunch of things and couldn't disable it. That to say that there could be more than just the version of `glibc` at play here.

It's probably a bit trickier than that - the library assumes that if it "is" available at compile time, then it doesn't need to build it's variant - and rely on an externally provided one (but undetermined who provides this one). Then the same artifact may or may not work, and it wasn't only glibc but also `libbsd`?

Given that this function is _only_ used internally in an implementation file, may as well keep `ac_cv_func_strlcpy=no` and rename the function altogether - then it will never conflict and still uses a valid function.

jwillikers:
> Given that this function is _only_ used internally in an implementation file, may as well keep `ac_cv_func_strlcpy=no` and rename the function altogether - then it will never conflict and still uses a valid function.

I think I'd been working for too long yesterday, because for some reason I didn't see where `strlcpy` was actually called. Now that I see that, yeah, renaming those would certainly do the trick.

jwillikers:
I've renamed the `strlcpy` symbol to `avahi_strlcpy`.

RubenRBS (approved):
Looks great now :)

uilianries:
> I have a feeling compiling it as a shared library might expose that problem. Ideally we'd be have the glibc version available as a setting, then we could force it on for new enough versions of `glibc`. I think you're supposed to be able to disable the function using feature definitions, but I tried a bunch of things and couldn't disable it. That to say that there could be more than just the version of `glibc` at play here.

Adding `glibc` to settings is a very, very old topic, since Conan 0.x. You can find several issues in Conan-io related to it: https://github.com/search?q=repo%3Aconan-io%2Fconan+glibc&type=issues

In the future we should be running new compilers with new version of glibc installed, so it should not be a problem (I hope). 

Let me update the issue there in the upstream. Indeed is very fragile case.

recipes/avahi/all/conandata.yml
```diff
@@ -2,3 +2,8 @@ sources:
   "0.8":
     url: "https://github.com/lathiat/avahi/releases/download/v0.8/avahi-0.8.tar.gz"
     sha256: "060309d7a333d38d951bc27598c677af1796934dbd98e1024e7ad8de798fedda"
+patches:
+  "0.8":
+    - patch_file: "patches/0.8-0001-Avoid-conflicting-definition-with-strlcpy-in-glibc.patch"
+      patch_description: "Avoid conflicting definition with glibc strlcpy"
+      patch_type: "portability"
```
uilianries, line 9:
```suggestion
      patch_type: "conan"
```

It's only affecting Conan.

jcar87 (approved):

uilianries (approved):
LGTM

PR #22997, opened 2024-03-05
[llvm-core] bring package up to date. Support for Conan 2.

### LINKED ISSUE #20339: llvm-core/13.0.0: recipe is broken (conan 2)
### Description

I get "Recipe 'llvm-core/13.0.0' seems broken." when trying to build `llvm-core/13.0.0` with conan 2.

* Package Name/Version: **llvm-core/13.0.0**
* Operating System+version: **Windows 11**
* Compiler+version: **msvc_2022**
* Conan version: **conan 2.0..13**
* Python version: **Python 3.11.14**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows

### Steps to reproduce

use this conanfile.py:

```
from conan import ConanFile


class CompressorRecipe(ConanFile):
    generators = "CMakeToolchain", "CMakeDeps"

    def requirements(self):
        self.requires("llvm-core/13.0.0")
```

run `conan install . --build=missing`

### Logs

```
```
llvm-core/13.0.0: Not found in local cache, looking in remotes...
llvm-core/13.0.0: Checking remote: conancenter
llvm-core/13.0.0: Downloaded recipe revision 95ce239bdd688987d2281e44fe8422e8
Graph root
    conanfile.py: C:\git\conan_playground\conanfile.py
Requirements
    abseil/20230125.3#5431a4c609f5fb48bb8d8567e953243f - Cache
    c-ares/1.19.1#420a0b77e370f4b96bee88ef91837ccc - Cache
    grpc/1.54.3#3e0d68cd1578502c9c3a0e8366f0ab77 - Cache
    openssl/3.1.3#35e30a39a89ee3ae98929377ebe78913 - Cache
    protobuf/3.21.12#8210c0b1bb46b08ff2814614ec091a9c - Cache
    re2/20230301#56bcddd1eaca2b093fd34525ae40ee9b - Cache
    zlib/1.2.13#97d5730b529b4224045fe7090592d4c1 - Cache
Build requirements
    nasm/2.15.05#ca1c14c2060af0ec151048b54ced9262 - Cache
    strawberryperl/5.32.1.1#8f83d05a60363a422f9033e52d106b47 - Cache
Resolved version ranges
    openssl/[>=1.1 <4]: openssl/3.1.3
Recipe 'llvm-core/13.0.0' seems broken.
It is possible that this recipe is not Conan 2.0 ready
If the recipe comes from ConanCenter, report it at https://github.com/conan-io/conan-center-index/issues
If it is your recipe, check if it is updated to 2.0

ERROR: Package 'llvm-core/13.0.0' not resolved: llvm-core/13.0.0: Cannot load recipe.
Error loading conanfile at 'C:\Users\pierr\.conan2\p\llvm-b0efe2cdcc15c\e\conanfile.py': Unable to load conanfile in C:\Users\pierr\.conan2\p\llvm-b0efe2cdcc15c\e\conanfile.py
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "C:\Users\pierr\.conan2\p\llvm-b0efe2cdcc15c\e\conanfile.py", line 6, in <module>
    from conans import CMake
ImportError: cannot import name 'CMake' from 'conans' (C:\Users\pierr\AppData\Local\Programs\Python\Python311\Lib\site-packages\conans\__init__.py)
.
```
```

planetmarshall:
Specify library name and version:  **llvm-core/***

Attempt to bring the llvm-core recipe up to date. Includes contributions from @jusito and #17509. Fixes #20339

Note that I have opted **not** to add the latest version (18.1.3) as part of this PR, as the existing recipe needed updating and is already quite complex. I (or someone else) will add the latest version as a followup.

---


recipes/llvm-core/all/conanfile.py
```diff
@@ -69,361 +76,265 @@ class LLVMCoreConan(ConanFile):
...
+            'LLVM_ENABLE_ZLIB': "FORCE_ON" if self.options.with_zlib else False,
+            'LLVM_ENABLE_LIBXML2': "FORCE_ON" if self.options.with_xml2 else False,
+            'LLVM_ENABLE_TERMINFO': self.options.with_terminfo,
+            'LLVM_RAM_PER_COMPILE_JOB': self.options.ram_per_compile_job,
+            'LLVM_RAM_PER_LINK_JOB': self.options.ram_per_link_job
+        }
+        if is_msvc(self):
+            build_type = str(self.settings.build_type).upper()
+            cmake_definitions['LLVM_USE_CRT_{}'.format(build_type)] = self.settings.compiler.runtime
+
```
jusito, line 184: Just saw your CI issues, beside the patch there is a second opt needed LLVM_USE_SPLIT_DWARF [its available in llvmorg-11.0.0](https://github.com/llvm/llvm-project/blob/llvmorg-11.0.0/llvm/CMakeLists.txt#L497) therefore I guess this snippet should help. [Since llvm 14 it is also in the docs as recommendation for elf platforms.](https://github.com/llvm/llvm-project/blob/llvmorg-14.0.0/llvm/docs/CMake.rst?plain=1#L757-L761)
```suggestion
        # Conan Center Index CI optimization, in build_type=Debug the CI is killing it because of high memory usage:
        # quote: This option reduces link-time memory usage by reducing the debug information the linker needs to resolve.
        # It is recommended for platforms using the ELF object format, like Linux systems, when linker memory usage is too high.
        is_platform_ELF_based = self.settings.os in [
            'Linux', 'Android', 'FreeBSD', 'SunOS', 'AIX', 'Neutrino', 'VxWorks'
        ]
        if is_platform_ELF_based:
            self.output.info(
                f"your platform \"{self.settings.os}\" is using the ELF format, optimizing memory usage during debug build linking.")
            cmake_definitions.update(
                {
                    'LLVM_USE_SPLIT_DWARF': True
                }
            )
            
```

jusito: If you don't like setting this value for every build we could use it just for CI with this trick: https://github.com/conan-io/conan-center-index/pull/17509#discussion_r1365797505

planetmarshall: > If you don't like setting this value for every build we could use it just for CI with this trick: [#17509 (comment)](https://github.com/conan-io/conan-center-index/pull/17509#discussion_r1365797505)

It's a good idea - but if I understand it correctly that suggestion is the reverse of what I want - it requires the user to define the config variable to get non-default behaviour. I've made a feature request to get the reverse, that is, have a config variable that can be checked by recipes to determine if they are running on CCI so that they can take action accordingly. https://github.com/conan-io/conan-center-index/issues/23050

jwillikers (approved):

recipes/llvm-core/all/conanfile.py
```diff
@@ -29,6 +29,8 @@
 
 required_conan_version = ">=1.62.0"
 
+LLVM_TARGETS = "AArch64;AMDGPU;ARM;AVR;BPF;Hexagon;Lanai;LoongArch;Mips;MSP430;NVPTX;PowerPC;RISCV;Sparc;SystemZ;VE;WebAssembly;X86;XCore"
```
jwillikers, line 32: I recommend using a list for this variable instead of a special delimited string.

planetmarshall: Being a bit lazy here as just copied and pasted from the LLVM docs.

planetmarshall:
> ### Conan v1 pipeline ❌
> 
> Failure in build 8 (`81bf61c336b6491d34db1cc30dab668ae40a7d68`):
> 
>     * **llvm-core/12.0.0:**
>       CI failed to create some packages ([All logs](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/22997/8-linux-gcc11/llvm-core/12.0.0//summary.json))
>       Logs for packageID `9e2762e3c828f6e056512746def7c7bed8d4112a`:
>       ```
>       [settings]
>       arch=x86_64
>       build_type=Debug
>       compiler=gcc
>       compiler.libcxx=libstdc++11
>       compiler.version=11
>       os=Linux
>       [options]
>       llvm-core:shared=True
>       ...
>       collect2: fatal error: ld terminated with signal 9 [Killed]
>       ```

Have disabled Shared Debug builds on CCI due to resource constraints. This is at least an improvement over the original recipe which disabled shared builds entirely for all users.

jwillikers:
> > ### Conan v1 pipeline ❌
> > Failure in build 8 (`81bf61c336b6491d34db1cc30dab668ae40a7d68`):
> > ```
> > * **llvm-core/12.0.0:**
> >   CI failed to create some packages ([All logs](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/22997/8-linux-gcc11/llvm-core/12.0.0//summary.json))
> >   Logs for packageID `9e2762e3c828f6e056512746def7c7bed8d4112a`:
> >   ```
> >   [settings]
> >   arch=x86_64
> >   build_type=Debug
> >   compiler=gcc
> >   compiler.libcxx=libstdc++11
> >   compiler.version=11
> >   os=Linux
> >   [options]
> >   llvm-core:shared=True
> >   ...
> >   collect2: fatal error: ld terminated with signal 9 [Killed]
> >   ```
> > ```
> 
> Have disabled Shared Debug builds on CCI due to resource constraints. This is at least an improvement over the original recipe which disabled shared builds entirely for all users.

If only you could use `lld`...

jusito:
Actually it should work if split dwarf is activated. For this issue I wrote some scripts which builds it in docker and collect cpu/memory consumption. I used your current recipe and run if for llvm 11.1, gcc 10, debug build, shared, 2048/16384 it was: peak CPU: 2369% peak Memory: 13312MiB. Looks like it should work in CI and split dwarf/job pools are working.

jusito:
It would be helpful to have exact memory constraints for the CI. Guessing into the blue, maybe double the ram per compile job (its just for ci) because compile and link jobs are not exclusive. Maybe n compile jobs scratching at the limit and if one finished and n-1 compile jobs running + one link job this is to much. Couldn't find a way to do this exclusiveness via ninja so this could be the issue here. Still guessing if we don't know the limits.

jwillikers (approved):

planetmarshall:
Could I get s second review on this? I have v18.1.1 and a Clang recipe waiting in the wings :) Pretty please!

jwillikers:
> Could I get a second review on this? I have v18.1.1 and a Clang recipe waiting in the wings :) Pretty please!

@planetmarshall You probably want to ping one of the team members, although  I think they are pretty heads-down right now fixing some of the major problems in CCI.

planetmarshall:
> > Could I get a second review on this? I have v18.1.1 and a Clang recipe waiting in the wings :) Pretty please!
> 
> @planetmarshall You probably want to ping one of the team members, although I think they are pretty heads-down right now fixing some of the major problems in CCI.

https://github.com/prince-chrismc/conan-center-index-pending-review/issues/1 mentions 12 days before raising an alarm so I'll give it another week before pestering anyone.

jwillikers:
@planetmarshall Thanks for updating this recipe btw. I'm more in need of  a general `llvm` package as the Mesa package requires some very specific components, but this really helps move things forward. Thank you!

planetmarshall:
Pinging @danimtb and @prince-chrismc for review if available.

prince-chrismc:
I haven't been involved in the project for a while so that table might not reflect the current trends. Nor will I be able to help with reviewing since i haven't used conan in a long while. 🙊

planetmarshall:
Pinging also @ericLemanissier and @RubenRBS 

Apologies for the spam but keen to move forward with this.

Cheers.

jwillikers (approved):

planetmarshall:
@RubenRBS  The build seems to have been stuck in limbo for a while. Are there still CI issues? Thanks.

jwillikers (approved):
Woot!

RubenRBS:
@planetmarshall sorry about that - llvm takes quite a long time to compile - I see that the compilation is now done and the results posted and they are passing, yay!! I'll get a review going for monday, thanks a lot for your patience and for the contribution in the first place, we really aprpeciate it :)

planetmarshall:
> @planetmarshall sorry about that - llvm takes quite a long time to compile - I see that the compilation is now done and the results posted and they are passing, yay!! I'll get a review going for monday, thanks a lot for your patience and for the contribution in the first place, we really aprpeciate it :)

Thanks, I also appreciate the efforts of the CCI team.

jacobfriedman:
@jusito possible to review this week? I opted not to use llvm-core and your PR instead (llvm-16) but am eagerly waiting on this PR and can help update/test/review the later PRs.

@jcar87 It would be a big plus to be able to simply pull down transparent/conan-compiled compilers in Dockerfiles e.g. https://github.com/conan-io/conan-docker-tools/blob/master/modern/clang/Dockerfile

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+        "A toolkit for the construction of highly optimized compilers,"
+        "optimizers, and runtime environments."
     )
-    license = 'Apache-2.0 WITH LLVM-exception'
-    topics = ('llvm', 'compiler')
-    homepage = 'https://llvm.org'
-    url = 'https://github.com/conan-io/conan-center-index'
-
-    settings = ('os', 'arch', 'compiler', 'build_type')
+    license = "LLVM-exception"
```
Croydon, line 63: This license change is wrong. Please undo

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+        "with_z3": True,
+        "with_zlib": True,
+        "ram_per_compile_job": "auto",
+        "ram_per_link_job": "auto"
     }
-
-    # Older cmake versions may have issues generating the graphviz output used
-    # to model the components
-    build_requires = [
-        'cmake/3.20.5'
```
Croydon, line 80: Why is it not required anymore?

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
-        if self.options.get_safe('with_xml2', False):
-            self.requires('libxml2/2.9.10')
-
-    def package_id(self):
-        del self.info.options.use_llvm_cmake_files
+            self.requires("libffi/3.4.4")
+        if self.options.get_safe("with_libedit"):
+            self.requires("editline/3.1")
+        if self.options.with_zlib:
+            self.requires("zlib/1.3.1")
```
Croydon, line 166:
```suggestion
            self.requires("zlib/[>=1.2.11 <2]")
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+        if self.options.with_xml2:
+            self.requires("libxml2/2.12.4")
```
Croydon, line 168:
```suggestion
            self.requires("libxml2/[>=2.12.5 <3]")
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+        if self.options.with_z3:
+            self.requires("z3/4.12.4")
```
Croydon, line 170:
```suggestion
            self.requires("z3/4.13.0")
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+
+    def build_requirements(self):
+        self.tool_requires("ninja/1.11.1")
```
Croydon, line 173:
```suggestion
        self.tool_requires("ninja/1.12.1")
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
-        'use_sanitizer': [
-            'Address',
-            'Memory',
-            'MemoryWithOrigins',
-            'Undefined',
-            'Thread',
-            'DataFlow',
-            'Address;Undefined',
-            'None'
+        "shared": [True, False],
```
Croydon, line 69: This pull request is a night mare to review due to size and patches, you really shouldn't have made formatting changes in such a pull request 🙈

recipes/llvm-core/all/conanfile.py
planetmarshall: Are you sure? This is the SPDX identifier for the LLVM license, according to https://spdx.org/licenses/LLVM-exception.html, and according to the [conan docs](https://docs.conan.io/2/reference/conanfile/attributes.html#license), "..it is strongly recommended that recipes of Open Source projects use [SPDX](https://spdx.dev/) identifiers from the [SPDX license list](https://spdx.org/licenses/)"

recipes/llvm-core/all/conanfile.py
planetmarshall: Because graphviz is not used to model component dependencies, as noted in the comment.

recipes/llvm-core/all/conanfile.py
planetmarshall: Sorry that it's been a "nightmare" for you, but the recipe was unmaintained and did not build with conan2. Now it does.

recipes/llvm-core/all/conanfile.py
valgur: The existing license was correct. See https://spdx.org/licenses/exceptions-index.html
> The SPDX License List includes a list of exceptions. These exceptions grant an exception to a license condition or additional permissions beyond those granted in a license; they are not stand-alone licenses. Exceptions are added to a license using the [License Expression](https://spdx.org/specifications) operator, "WITH".

Croydon: What you linked is just an exception identifier not a license identifier. @planetmarshall

recipes/llvm-core/all/conanfile.py
Croydon: Thank you for your work! Didn't mean to sound ungrateful or something 🙂

recipes/llvm-core/all/conanfile.py
planetmarshall: Reverted - thanks for the clarification.

recipes/llvm-core/all/conanfile.py
planetmarshall: To clarify further - the original recipe used graphviz output which was then parsed and sanitized to generate component dependencies. Instead of doing this I opted to just use the information that LLVM already outputs in its CMake files, I thought this was more robust since the best source of cmake dependencies is cmake itself rather than going via graphviz and back again.

jwillikers (approved):

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
 import os
+from pathlib import Path
+import re
 import textwrap
 
-required_conan_version = ">=1.50.2" # Due to conan.tools.scm.Version
+
+required_conan_version = ">=1.62.0"
+
+# LLVM's default config is enable all targets, but end users can significantly reduce
```
valgur, line 31:
```suggestion
# LLVM's default config is to enable all targets, but end users can significantly reduce
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+        if self.options.shared:
+            if self._is_windows:
+                raise ConanInvalidConfiguration("Shared builds not currently supported on Windows")
```
valgur, line 186:
```suggestion
                raise ConanInvalidConfiguration("Shared builds are currently not supported on Windows")
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+            raise ConanInvalidConfiguration("Cannot enable exceptions without rtti support")
 
-    def source(self):
-        get(self, **self.conan_data["sources"][self.version], strip_root=True,
-            destination=self._source_subfolder)
-        self._patch_sources()
+        if cross_building(self):
+            # FIXME support cross compilation, at least for common cases like Apple Silicon -> X86
+            #  requires a host-compiled version of llvm-tablegen
+            raise ConanInvalidConfiguration("Cross compilation is not supported")
```
valgur, line 203:
```suggestion
            raise ConanInvalidConfiguration("Cross compilation is not supported. Contributions are welcome!")
```
Just to let the user know that this is not a limitation of LLVM and could be added in the future.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,510 @@
...
+        # handle conan legacy setting
+        if msvcrt in ["MDd", "MTd", "MD", "MT"]:
+            return msvcrt
+
+        if self.settings.build_type in ["Debug", "RelWithDebInfo"]:
+            crt = {"static": "MTd", "dynamic": "MDd"}
+        else:
+            crt = {"static": "MT", "dynamic": "MD"}
+
+        return crt[msvcrt]
```
valgur, line 243: Perhaps you can use `conan.tools.microsoft.visual.msvc_runtime_flag()` to simplify this logic?

jwillikers (approved):

recipes/llvm-core/all/conanfile.py
planetmarshall: Good spot - I just didn't know it existed.

jcar87:
Hi @planetmarshall - please let me extend my apologies on behalf of the entire team for the delay in reviewing updates to the llvm-core recipe. We appreciate your contribution and the time you have dedicated to this task, especially considering the intricacies of the LLVM build system.

Coinciding with our sprint planning this week, we are prioritizing the review of this PR in this coming sprint. 

At Conan Center, we have reached a point where the user base and the volume of contributions are exceeding our capacity to review and merge contributions as quickly as both us and our contributors would like. There is also a less visible part of maintaining the infrastructure for the CI service. This situation is forcing us to prioritize - however we don’t always get it right, so please accept our unreserved apologies. Sometimes it is difficult for us to prioritise - as there can be conflicting priorities between users and contributors, and the fact that for historical reasons, PRs in Conan Center tend to concentrate too many changes, making reviews more difficult.

I am aware that certain pull requests tend to fall off the stack due to their complexity: this iis one such case, not only from the scope of changes proposed (~1500 lines) and the intricacies of the LLVM build system, but also the CI times tend to delay the review process. This PR specifically is taking close to 17 hours to go through CI validation - so any changes the team or the contributors make, would effectively take an entire business day to validate - any “back and forth” in terms of review comments and feedback would end up delaying the review process even further.

When reviewing pull requests - we take into consideration several factors that go beyond CI validation:
Do we understand the changes, and are we comfortable maintaining the recipe moving forward in terms of bug fixes and features that may arise in the future? 
Are there code paths or functionality that is not tested by CI (for whatever reason) that merit additional testing, either by us or by the contributor?
Are there any risks to our users? Typically any sort of disruption: broken builds, version conflicts, etc. But also from a security perspective.

The growing user base of Conan Center means that any disruption has the potential to affect a great number of users, and our role in the software supply chain and the increasing focus in security and the Open Source community demand that we are thorough in the review process. 

However - this is not an excuse. While this PR is complex and thus is likely to require a longer review than others - we should have allocated proper time to look into this and provided an update. Once again, please accept our unreserved apologies. The team will review this between this week and the next.

planetmarshall:
> Hi @planetmarshall - please let me extend my apologies on behalf of the entire team for the delay in reviewing updates to the llvm-core recipe. We appreciate your contribution and the time you have dedicated to this task, especially considering the intricacies of the LLVM build system.
> 

Many thanks for the response. I appreciate that the team have to prioritize PR reviews and other tasks and I have absolutely no problem with this - I'm sure the number of open issues and PRs for such a large project must get overwhelming. However the key word in your response from my perspective is "visibility" - from a contributor's perspective there's no visibility into what that prioritization looks like. I don't have a solution to this but even some automated indication that PRs that have had no activity for extended periods of time despite passing CI checks and review requests are in some way being flagged to the team for attention.

Thanks again, and your time is appreciated.

recipes/llvm-core/all/conandata.yml
```diff
@@ -10,18 +10,57 @@ sources:
...
+    - patch_file: patches/13x/0000-cmake-dependencies.patch
+      patch_description: fix references to third party libs to match conan variables and targets
+      patch_type: conan
+    - patch_file: patches/13x/0001-calculate-job-pools.patch
+      patch_description: calculate resource limits for compiling LLVM
+      patch_type: portability
+      patch_source: https://github.com/llvm/llvm-project/pull/65274
+    - patch_file: patches/13x/0002-missing-includes.patch
+      patch_description: fix missing header files
+      patch_type: portability
```
franramirez688, line 23: Just to double-check, does this patch come from https://github.com/llvm/llvm-project-release-prs/pull/3?

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
+        "use_perf": False,
+        "use_sanitizer": "None",
+        "with_libedit": True,
+        "with_ffi": False,
+        "with_terminfo": False,  # differs from LLVM default
+        "with_xml2": True,
+        "with_z3": True,
+        "with_zlib": True,
+        "ram_per_compile_job": "auto",
+        "ram_per_link_job": "auto"
```
franramirez688, line 120:
```suggestion
```
I saw how they are used in the recipe. IMO, I think it could be better to use user configurations instead:

```python
    def _apply_resource_limits(self, cmake_definitions):
        # ....
        else:
            ram_per_compile_job = self.conf.get("user.llvm:ram_per_compile_job")
            ram_per_link_job = self.conf.get("user.llvm:ram_per_link_job")
            if ram_per_compile_job:
                cmake_definitions["LLVM_RAM_PER_COMPILE_JOB"] = ram_per_compile_job
            if ram_per_link_job:
                cmake_definitions["LLVM_RAM_PER_LINK_JOB"] = ram_per_link_job
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
-        'with_xml2': [True, False],
-        'use_llvm_cmake_files': [True, False],
+        "with_ffi": [True, False],
+        "with_libedit": [True, False],
+        "with_terminfo": [True, False],
+        "with_zlib": [True, False],
+        "with_xml2": [True, False],
+        "with_z3": [True, False],
+        "ram_per_compile_job": ["ANY"],
+        "ram_per_link_job": ["ANY"],
```
franramirez688, line 97:
```suggestion
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
-        self._supports_compiler()
-        if cross_building(self, skip_x64_x86=True):
-            raise ConanInvalidConfiguration('Cross-building not implemented')
+            raise ConanInvalidConfiguration("Cannot enable exceptions without rtti support")
 
-    def source(self):
-        get(self, **self.conan_data["sources"][self.version], strip_root=True,
-            destination=self._source_subfolder)
-        self._patch_sources()
+        if cross_building(self):
```
franramirez688, line 200: Should it not be `can_run` instead?

```suggestion
        if can_run(self):
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
+            self.output.info("Applying CCI Resource Limits")
+            cmake_definitions.update({
+                "LLVM_RAM_PER_LINK_JOB": "16384",
+                "LLVM_RAM_PER_COMPILE_JOB": "2048"
+            })
+        else:
+            if self.options.ram_per_compile_job != "auto":
+                cmake_definitions["LLVM_RAM_PER_COMPILE_JOB"] = self.options.ram_per_compile_job
+            if self.options.ram_per_link_job != "auto":
+                cmake_definitions["LLVM_RAM_PER_LINK_JOB"] = self.options.ram_per_link_job
```
franramirez688, line 219: As commented:

```suggestion
        else:
            ram_per_compile_job = self.conf.get("user.llvm:ram_per_compile_job")
            ram_per_link_job = self.conf.get("user.llvm:ram_per_link_job")
            if ram_per_compile_job:
                cmake_definitions["LLVM_RAM_PER_COMPILE_JOB"] = ram_per_compile_job
            if ram_per_link_job:
                cmake_definitions["LLVM_RAM_PER_LINK_JOB"] = ram_per_link_job
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
+
+    @property
+    def _is_windows(self):
+        return self.settings.os == "Windows"
+
+    @staticmethod
+    def load(filename):
+        # regex fails on Windows when using conan's built in 'load' method
+        with open(filename, "r", encoding="utf-8")as fp:
+            return fp.read()
```
franramirez688, line 321: Conan's method was not working because of the `Path` management I guess:

```suggestion
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
-                alias_targets
-            )
+                    data["requires"].append(component)
+            return data
 
-            self._create_cmake_module_alias_targets(
-                os.path.join(self.package_folder, self._old_alias_module_file_rel_path),
-                old_alias_targets
-            )
+        cmake_exports = self.load(Path(self.package_folder) / "lib" / "cmake" / "llvm" / "LLVMExports.cmake")
```
franramirez688, line 367: Let's try the Conan's one, but using the `as_posix()` method:

```suggestion
        cmake_exports = load(self, (Path(self.package_folder) / "lib" / "cmake" / "llvm" / "LLVMExports.cmake").as_posix())
```
For all the `Path` uses in this recipe, `PurePath` should be more than enough.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
+            if llvm_lib in components:
+                components[llvm_lib].update(_parse_deps(dependencies))
 
-        rm(self, "LLVMExports*.cmake", self.package_folder, recursive=True)
-        rename(self, os.path.join(self.package_folder, self._module_subfolder, 'LLVM-Config.cmake'),
-               os.path.join(self.package_folder, self._module_subfolder, 'LLVM-ConfigInternal.cmake'))
-        rename(self, os.path.join(self.package_folder, self._module_subfolder, 'LLVMConfig.cmake'),
-               os.path.join(self.package_folder, self._module_subfolder, 'LLVMConfigInternal.cmake'))
+    def _llvm_build_info(self):
+        cmake_config = self.load(Path(self.package_folder) / "lib" / "cmake" / "llvm" / "LLVMConfig.cmake")
```
franramirez688, line 376:
```suggestion
        cmake_config = load(self, (Path(self.package_folder) / "lib" / "cmake" / "llvm" / "LLVMConfig.cmake").as_posix())
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
+    def package(self):
+        copy(self, "LICENSE.TXT", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
 
+        build_info = self._write_build_info()
+
+        package_folder = Path(self.package_folder)
+        cmake_folder = package_folder / "lib" / "cmake" / "llvm"
+        rm(self, "LLVMConfig.cmake", cmake_folder)
```
franramirez688, line 451: I tend to say that we should pass everything to the Conan methods as `Path().as_posix()` instead of `Paths` objects to avoid any unexpected errors.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,496 @@
...
+            build_info,
+            package_folder / self._build_module_file_rel_path
+        )
 
-            for lib, target in targets.items():
-                if lib in deps:
-                    self.cpp_info.components[component].requires.append(target)
+    def package_id(self):
+        del self.info.options.ram_per_compile_job
+        del self.info.options.ram_per_link_job
```
franramirez688, line 471: Using the confs, we would not need these lines:

```suggestion
```

franramirez688 (commented):
Hi @planetmarshall 

As @jcar87 commented, sorry for the huge delay reviewing this 😓 

Thanks a lot for the tremendous effort of bringing this PR 👏  I just left some comments to understand better some patches and to suggest other things that could help in the future.

recipes/llvm-core/all/conandata.yml
```diff
@@ -10,18 +10,57 @@ sources:
...
+    - patch_file: patches/13x/0003-split-dwarf.patch
+      patch_description: enable split dwarf option from 14.x
+      patch_source: https://reviews.llvm.org/D108776
```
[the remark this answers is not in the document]
planetmarshall: Yes - additional changes were necessary to backport the patch from LLVM 14

recipes/llvm-core/all/conanfile.py
planetmarshall: No - I specifically need to check for cross building (even if the host can run foreign architectures like Rosetta on Macos) as this needs special setup to build LLVM. This is left as an exercise for the next contributor!

recipes/llvm-core/all/conandata.yml
planetmarshall: Yes - updated `patch_source`

franramirez688 (dismissed):
Hi @planetmarshall 

As I said, thanks for all the efforts to get this PR, and I am sorry for the delay in reviewing.
IMO, it looks great! Approving so far 👏

jwillikers (approved):

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
-        'with_xml2': [True, False],
-        'use_llvm_cmake_files': [True, False],
+        "with_ffi": [True, False],
+        "with_libedit": [True, False],
+        "with_terminfo": [True, False],
+        "with_zlib": [True, False],
+        "with_xml2": [True, False],
+        "with_z3": [True, False],
     }
+    options.update({f"with_target_{target.lower()}": [True, False] for target in LLVM_TARGETS})
```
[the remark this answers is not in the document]
planetmarshall: TBH I don't actually remember why I did this. I was probably modeling it on another recipe that has a similar set of options like `boost`. LLVM's default behaviour is to always build with all targets enabled.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
-        'with_ffi': [True, False],
-        'with_zlib': [True, False],
-        'with_xml2': [True, False],
-        'use_llvm_cmake_files': [True, False],
+        "with_ffi": [True, False],
+        "with_libedit": [True, False],
+        "with_terminfo": [True, False],
+        "with_zlib": [True, False],
+        "with_xml2": [True, False],
+        "with_z3": [True, False],
```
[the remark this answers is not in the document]
planetmarshall: I assume it was an oversight from the previous recipe. as Z3 is an optional dependency just like zlib or xml2. Perhaps the z3 recipe was not available at the time.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
+from pathlib import PurePosixPath
+import re
 import textwrap
 
-required_conan_version = ">=1.50.2" # Due to conan.tools.scm.Version
+
+required_conan_version = ">=1.62.0"
+
+# LLVM's default config is to enable all targets, but end users can significantly reduce
+# build times for the package by disabling the ones they don't need with the corresponding option
```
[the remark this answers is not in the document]
planetmarshall: TBH I don't actually remember why I did this. I was probably modeling it on another recipe that has a similar set of options like `boost`. LLVM's default behaviour is to always build with all targets enabled.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
+                raise ConanInvalidConfiguration("Shared builds are currently not supported on Windows")
+            if os.getenv("CONAN_CENTER_BUILD_SERVICE") and self.settings.build_type == "Debug":
+                raise ConanInvalidConfiguration("Shared Debug build is not supported on CCI due to resource limitations")
+            if is_apple_os(self):
+                # FIXME iconv contains duplicate symbols in the libiconv and libcharset libraries (both of which are
+                #  provided by libiconv). This may be an issue with how conan packages libiconv
+                iconv_dep = self.dependencies.get("libiconv")
+                if iconv_dep and not iconv_dep.options.shared:
+                    raise ConanInvalidConfiguration("Static iconv cannot be linked into a shared library on macos "
+                                                    "due to duplicate symbols. Use libxml2/*:iconv=False.")
```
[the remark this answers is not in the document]
planetmarshall: No, this is caused by the macos build of libiconv containing both libiconv and libcharset which contain duplicate symbols. It's easily reproduced by creating an executable that depends on that library. If I have time I'll raise a separate bug report.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
-        self._supports_compiler()
-        if cross_building(self, skip_x64_x86=True):
-            raise ConanInvalidConfiguration('Cross-building not implemented')
+        # this capability is back-ported from LLVM 14.x
+        is_platform_ELF_based = self.settings.os in [
+            "Linux", "Android", "FreeBSD", "SunOS", "AIX", "Neutrino", "VxWorks"
+        ]
+        if is_platform_ELF_based:
+            self.output.info("ELF Platform Detected, optimizing memory usage during debug build linking.")
+            cmake_variables["LLVM_USE_SPLIT_DWARF"] = True
```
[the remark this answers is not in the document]
planetmarshall: This was necessary to get a debug build to succeed on the CCI build system. Without it I would have to deactivate all Debug builds. I believe the patch originated with one of the other LLVM recipes on which I based this one.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
-            destination=self._source_subfolder)
-        self._patch_sources()
+        if is_msvc(self):
+            build_type = str(self.settings.build_type).upper()
+            cmake_variables[f"LLVM_USE_CRT_{build_type}"] = msvc_runtime_flag(self)
+
+        if not self.options.shared:
+            cmake_variables.update({
+                "DISABLE_LLVM_LINK_LLVM_DYLIB": True,
+                "LLVM_ENABLE_PIC": self.options.get_safe("fPIC", default=False)
```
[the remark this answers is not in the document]
planetmarshall: Sure, but that is a change from the original recipe behaviour.

recipes/llvm-core/all/patches/11x/0000-cmake-dependencies.patch
```diff
@@ -0,0 +1,118 @@
...
+-    endforeach()
++    cmake_push_check_state()
++    list(APPEND CMAKE_REQUIRED_INCLUDES ${ZLIB_INCLUDE_DIRS})
++    list(APPEND CMAKE_REQUIRED_LIBRARIES ${ZLIB_LIBRARIES})
++    check_symbol_exists(compress2 zlib.h HAVE_LIBZ)
++    cmake_pop_check_state()
+   endif()
+ 
+-  # Don't look for these libraries on Windows.
+-  if (NOT PURE_WINDOWS)
```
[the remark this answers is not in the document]
planetmarshall: Because conan makes it easy to do so. I assume that the reason LLVM does not check for these libraries on Windows is dues to the difficulty in providing them. As conan allows us to provide them, I see no reason not to do so.

recipes/llvm-core/all/conanfile.py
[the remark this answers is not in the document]
planetmarshall: Having said that - it's been a while since I tested a lot of these changes and there have been various changes to the recipe since then. It might take me some time to go back through the various configurations but I will check.

recipes/llvm-core/all/conanfile.py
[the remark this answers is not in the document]
planetmarshall: One advantage to the current approach is that the targets are available in the conanfile for inspection, and it doesn't require the user to go digging into the LLVM build files to find out what the available targets are. Having said that, if users are missing with this option then they're probably comfortable doing that anyway.

recipes/llvm-core/all/conanfile.py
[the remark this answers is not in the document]
jusito: I added it in my recipe: https://github.com/conan-io/conan-center-index/pull/17509/commits/e4bd6fa50f584775f6950e193cc2e08554e3e9dc

yes its not default but as far as I know was recommended in the llvm docs to reduce memory consumption during linking, at least for llvm 14+ https://releases.llvm.org/14.0.0/docs/CMake.html
> If enabled CMake will pass -gsplit-dwarf to the compiler. This option reduces link-time memory usage by reducing the amount of debug information that the linker needs to resolve. It is recommended for platforms using the ELF object format, like Linux systems when linker memory usage is too high.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
-            content += textwrap.dedent("""\
-                if(TARGET {aliased} AND NOT TARGET {alias})
-                    add_library({alias} INTERFACE IMPORTED)
-                    set_property(TARGET {alias} PROPERTY INTERFACE_LINK_LIBRARIES {aliased})
-                endif()
-            """.format(alias=alias, aliased=aliased))
-        save(self, module_file, content)
+        cmake_exports = load(self, (self._package_folder_path / "lib" / "cmake" / "llvm" / "LLVMExports.cmake").as_posix())
+        match_dependencies = re.compile(
+            r'''^set_target_properties\((\w+).*\n?\s*INTERFACE_LINK_LIBRARIES\s+"(\S+)"''', re.MULTILINE)
```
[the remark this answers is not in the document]
planetmarshall: I have an LLVM-18 recipe in the works that will follow this one - I'll try the Graphviz approach again but I seem to recall it had issues of its own, such as introducing circular dependencies that then need to be sanitized. Maybe the cmake file API could be used to do this?

planetmarshall:
@jcar87 Most recent CI build failed but i'm unable to view any logs.

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,491 @@
...
+                self.cpp_info.components[component_name].libs = [component_name]
+                requires = data.get("requires")
+                if requires is not None:
+                    self.cpp_info.components[component_name].requires += requires
+                system_libs = data.get("system_libs")
+                if system_libs is not None:
+                    self.cpp_info.components[component_name].system_libs += system_libs
+        else:
+            self.cpp_info.set_property("cmake_target_name", "LLVM")
+            self.cpp_info.libs = collect_libs(self)
```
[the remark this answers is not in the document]
planetmarshall: Explicitly listing the shared libraries fails the Conan v1 pipeline with `ERROR: [LIBRARY DOES NOT EXIST (KB-H054)] Component llvm-core::llvm-core library 'LLVM-11' is listed in the recipe, but not found installed at self.cpp_info.libdirs.`

Conan v2 works. I have reverted to `collect_libs`

recipes/llvm-core/all/conanfile.py
planetmarshall: In the latter case, LLVM will attempt to build a native toolchain in a subdirector of the build directory. For this to work, that subdirectory would need to have the cmake configuration files generated by conan for the **build** platform - as well as the files generated for the **host** platform for the cross compilation build. The error is as follows - 

```
Configuring NATIVE LLVM...
FAILED: NATIVE/CMakeCache.txt 
cd /Users/jenkins/workspace/prod-v2/bsr/82122/cabdf/p/b/llvm-2a22810816590/b/build/Release/NATIVE && /Users/jenkins/cmake/cmake-3.20.1/bin/cmake -G Ninja -DCMAKE_MAKE_PROGRAM="/Users/jenkins/workspace/prod-v2/bsr/82122/cabdf/p/ninja7130a04140d30/p/bin/ninja" -DCMAKE_C_COMPILER=/Applications/conan/xcode/13.0/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -DCMAKE_CXX_COMPILER=/Applications/conan/xcode/13.0/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ /Users/jenkins/workspace/prod-v2/bsr/82122/cabdf/p/b/llvm-2a22810816590/b/src -DLLVM_TARGET_IS_CROSSCOMPILE_HOST=TRUE -DLLVM_TARGETS_TO_BUILD="AArch64;AMDGPU;ARM;AVR;BPF;Hexagon;Lanai;Mips;MSP430;NVPTX;PowerPC;RISCV;Sparc;SystemZ;WebAssembly;X86;XCore" -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD="" -DLLVM_DEFAULT_TARGET_TRIPLE="arm64-apple-darwin23.4.0" -DLLVM_TARGET_ARCH="host" -DLLVM_ENABLE_PROJECTS="" -DLLVM_EXTERNAL_PROJECTS="" -DLLVM_TEMPORARILY_ALLOW_OLD_TOOLCHAIN="OFF" -DCMAKE_BUILD_TYPE=Release
```

recipes/llvm-core/all/conanfile.py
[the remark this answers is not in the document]
jcar87: Thanks! yeah, if we could have an issue with the errors that would be great, so that we can look into this

recipes/llvm-core/all/conanfile.py
jcar87: Thanks @planetmarshall  - this makes perfect sense. 

please feel free to add a link to https://github.com/conan-io/conan-center-index/pull/22997/files#r1740847096 in the recipe in the comments, so that in the future we know what the reason  in case it comes up

recipes/llvm-core/all/conanfile.py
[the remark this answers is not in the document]
jcar87: Thanks @planetmarshall  and @jusito for the added context.

If the big motivator is the Conan Center CI, and given that this affects Linux _debug_ builds - I personally think that we could probably avoid the complexity of backporting a feature from a newer version, and simply skipping that configuration in our CI (checking against the value of the environment variable, so that it is only disabled for this CI). 

Bear in mind that Linux debug binaries are generally less requested than release anyway.

recipes/llvm-core/all/conanfile.py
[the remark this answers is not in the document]
jcar87: perfect, this makes sense, especially if the library/file names have a lot of variation

recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,478 @@
...
-                'LibXml2::LibXml2': 'xml2'
-            }
+            cmake_variables.update({
+                "DISABLE_LLVM_LINK_LLVM_DYLIB": True,
+                "LLVM_ENABLE_PIC": self.options.get_safe("fPIC", default=True)
+            })
 
-            components = defaultdict(list)
-            for lib, dep in deps:
-                if not lib.startswith('LLVM'):
```
cynix, line 296: I think this needs to be kept. Without something like this, I’m seeing shared libraries (`libRemarks.so` and `libLTO.so`) in the produced package even when `shared=False`.

cynix: Might also want to delete `*.so` and `*.so.*` from the package.

planetmarshall: [libRemarks](https://llvm.org/docs/Remarks.html) and [libLTO](https://llvm.org/docs/LinkTimeOptimization.html#liblto) are part of the LLVM infrastructure and not available as static libraries.

libLTO specifically is used by some out of tree linkers but not by LLVM itself - See https://discourse.llvm.org/t/how-to-use-dynamic-library-liblto-so/74541 for example.

cynix: I understand these are not available statically. However I don’t think it’s correct to include shared libs in a static package. For example, if my intention of using a static package is to produce an executable that doesn’t depend on any shared libs for ease of deployment, now I’m forced to link to these 2 shared libs when I link to `llvm-core::llvm::core`, and the executable will not run in the target environment as intended.

I think we should delete them, and the users of the static package should simply be aware that these 2 components are not available.

cynix: > `llvm-core::llvm-core` is a conan-generated target and would not ordinarily be available to consumers of LLVM

By “not available” do you just mean “not recommended”? Because I certainly have code using it, and it works fine (except for linking to the 2 shared libs).

planetmarshall: I do think that `LTO` and `Remarks` should probably be made separate components for the shared build, but IMHO this is something that can be added as a future improvement and shouldn't block this PR.

jwillikers (approved):

franramirez688 (approved):

AbrilRBS (approved):

jcar87 (approved):
Thanks @planetmarshall  and everyone involved - this was not an easy recipe to port, and hopefully this can unblock some of the usages that are required by the community.

There's still a few things to look into further down the line - the issue with libiconv in macOS, and the parsing of CMake files (it is obvious from the recipe code that this can be fragile) - however, I dont think these are an impediment to merge the recipe as proposed in this PR. 

Thanks so much!

PR #20172, opened 2023-09-27
tensorflow lite: fix include 'tensorflow/lite/version.h'

Wawha:
Library name and version:  **tensorflow-lite/2.12.0**

I was not possible to include `tensorflow/lite/version.h` because file `tensorflow/core/public/version.h` was not included inside library include path.

---

- [X] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [X] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [X] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #21762, opened 2023-12-14
xnnpack: Add cci.20231026

talyz:
Specify library name and version:  **xnnpack/cci.20231026**

Depends on #21761 and  #21759 which should be merged before this.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #23006, opened 2024-03-06
tensorflow-lite: resolve conflicts

jcar87:
Work in progress pending tests.

Solve dependency conflicts caused by https://github.com/conan-io/conan-center-index/pull/21762/files.

Need to ensure all tests pass - as it's hard to reason whether tensorflow-lite is still compatible with these newer versions.

Before this PR:
* `conan graph info --requires=tensorflow-lite/2.12.0 --requires=opencv/4.5.5 --profile:host Linux-gcc11-x86_64-Release --profile:build Linux-gcc11-x86_64-Release` fails

After this PR: the command above should pass without conflicts.

### DIFF
--- a/recipes/tensorflow-lite/all/conanfile.py
+++ b/recipes/tensorflow-lite/all/conanfile.py
@@ -83,9 +83,9 @@ class TensorflowLiteConan(ConanFile):
         if self.settings.arch in ("x86", "x86_64"):
             self.requires("intel-neon2sse/cci.20210225")
         if self.options.with_xnnpack:
-            self.requires("xnnpack/cci.20220801")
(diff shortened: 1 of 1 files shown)


jcar87:
Note: will include this fix https://github.com/conan-io/conan-center-index/pull/20172 before merging.

RubenRBS:
Bumping abseil to the latest available patch for its current version would solve the blocker in https://github.com/conan-io/conan-center-index/pull/22982, while also bringing it up-to-date with the rest of the CCI libraries that require this version, all of which point to the latest patch - thus they currently conflict with this one

/cc @Ahajha

czoido (approved):

PR #23007, opened 2024-03-06
avahi/0.8: Support cross-compilation

jwillikers:
To cross-compile Avahi, the `--with-distro` flag must be passed.
Disable introspection.
This must be disabled to cross-compile.
Strangely enough, it doesn't work when enabled when doing a native build.
It only seems to work when set to auto.
When gobject-introspection is updated to support Conan V2 an option could be added to control enabling or disabling introspection.


ericLemanissier (approved):

RubenRBS:
Thanks! Could you please share compilation logs of your cross-compile test? Would help a lot on this one :)

(Also, just noticed in line 81 that you can safely remove the check!)

RubenRBS (approved):

ericLemanissier (approved):

Issue #5782, opened 2021-06-06
[package] fontconfig/2.13.93: How to use with sysconfdir set on build machine


hobbeshunter:
### Package and Environment Details (include every applicable attribute)
  * Package Name/Version: **fontconfig/2.13.93**
  * Operating System+version: **Linux Ubuntu 20.04**
  * Compiler+version: **GCC 9.3.0**
  * Conan version: **conan 1.36.0**
  * Python version: **Python 3.8.5**

### Conan profile
```
[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=9
compiler.libcxx=libstdc++11
build_type=Release
[options]
[build_requires]
[env]
```

### Steps to reproduce (Include if Applicable)
Currently I'm trying to build an app with Qt6. When I run the app I get the error 
```
Fontconfig error: Cannot load default config file: No such file: (null)
```
When investigating this with strace I found that fontconfig looks at 
```
/home/conan/w/BuildSingleReference/.conan/data/fontconfig/2.13.93/_/_/package/211285f7c5eb7915075b241bd6a053e345cec58c/bin/etc/fonts/fonts.conf
```

which seems to come from the way sysconfdir is defined in the recipe: https://github.com/conan-io/conan-center-index/blob/38d85e32c8352c7faabf191901ac126ec1214f95/recipes/fontconfig/all/conanfile.py#L61

Of course this behavior can be avoided by setting the environment variable `FONTCONFIG_PATH`.

As I want my app easily distributable through conan, what is the recommended way to hint fontconfig to the right path (e.g. `/etc/fonts` on Ubuntu)?

Thanks in advance!

rockdreamer:
This issue affects me too, our applications look.. less that stellar due to use of base X11 fonts on linux

paulharris:
I'm thinking a package option might be an approach.
Then you build for a target platform, but now it is up to every conan user to build fontconfig with the right paths compiled in ...
Still better than asking the user to set an env variable, or have to drop scripts to run binaries to set the env.

uilianries:
Usually we install those extra files in `package_folder/res`, then people can redistribute as they want by [imports](https://docs.conan.io/en/latest/reference/conanfile_txt.html#imports) or [deploy](https://docs.conan.io/en/latest/reference/generators/deploy.html). Conan is not focused in deploying artifacts ...

[...]

PR #23010, opened 2024-03-06
fontconfig: Remove patch for freetype libtool version number

jwillikers:
This patch is a bad workaround for improper versioning in the freetype package. This is fixed in the freetype package itself in #22974.

In addition, this fixes the following error, which can be seen when running the test package for the Meson variant of the package:

```
Fontconfig error: Cannot load default config file: No such file: (null)
```

This is fixed by installing the fontconfig configuration file.
It was being installed to the wrong location in the Meson package previously.
Additionally, don't set `FONTCONFIG_FILE`, but instead append to `FONTCONFIG_PATH` only.
This works and makes it easier for consumers to opt for the system font directory.
See #5782.

Requires:

- [x] #22974


ericLemanissier (approved):
this PR seems to be urgent. LGTM

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8799846282/attempts/1)

jellespijker:
thanks for helping out @ericLemanissier

jellespijker (approved):
lgtm

jellespijker:
> I told the team yesterday to merge this immediately after the freetype one. I'll update it to at least get it building.

Out of curiosity, did you guys ever experimented with GitHub merge queue's? It's on my bucket list to investigate the usefulness for our project. It seems to be helpful for these kind of stacked PR's

recipes/fontconfig/meson/conanfile.py
```diff
@@ -75,15 +74,14 @@ def generate(self):
             "doc": "disabled",
             "nls": "disabled",
             "tests": "disabled",
-            "tools": "disabled"
+            "tools": "disabled",
+            "sysconfdir": os.path.join(self.package_folder, "res", "etc"),
+            "datadir": os.path.join(self.package_folder, "res", "share"),
```
ericLemanissier, line 79:
```suggestion
            "sysconfdir": os.path.join("res", "etc"),
            "datadir": os.path.join("res", "share"),
```
according to https://c3i.jfrog.io/c3i/misc/logs/pr/23010/8-linux-gcc-legacy/fontconfig/2.14.2//1dd5f947341f0c35b104c3cbc22073e69527248e-build.txt
these two paths seem to be relative
see also https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/2.14.2/meson.build?ref_type=tags#L291 and 
https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/2.14.2/meson.build?ref_type=tags#L301

jwillikers: Oh yeah, I'll need to update the version requirement too.

ericLemanissier (approved):

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8799846282/attempts/2)

jellespijker (approved):
LGTM

franramirez688 (approved):

RubenRBS (approved):

jellespijker:
Thanks everyone on behalf of team Cura for giving this PR priority.

It's appreciated :hearth: !

recipes/fontconfig/meson/conanfile.py
```diff
@@ -3,16 +3,15 @@
...
 )
 from conan.tools.gnu import PkgConfigDeps
 from conan.tools.layout import basic_layout
 from conan.tools.meson import Meson, MesonToolchain
-from conan.tools.scm import Version
 
 import os
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=1.64.0 <2 || >=2.2.0"
```
valgur, line 14: @jwillikers Why was this massive bump to the required Conan version necessary?

Related to https://github.com/conan-io/conan/issues/16140

PR #23011, opened 2024-03-06
harfbuzz: Disable auto-features which pull in system dependencies

jwillikers:
Harfbuzz autodetects cairo and pulls it in, which breaks reproducibility. This also breaks cross compilation by pulling in dependencies on the build system. Disabling auto features in Meson stops this unexpected behavior from happening. See [here](https://github.com/harfbuzz/harfbuzz/blob/06749fa481c02ced563ea20685ac7697ea933dfe/meson_options.txt#L6)


ericLemanissier (approved):

RubenRBS (approved):

RubenRBS:
Looks good to me :)

uilianries (approved):

PR #23013, opened 2024-03-06
libudev: Set the pkg-config version to the real version number

jwillikers:
The pkg-config file contains the version of the Conan package. This is problematic because it is `system` which causes version checks for libudev to fail. This happens in Weston I'm making for instance:

```
Dependency libudev found: NO found system but need: '>= 136'
Found CMake: /usr/local/bin/cmake (3.28.20240202)
Run-time dependency libudev found: NO (tried cmake)

../src/libweston/meson.build:196:2: ERROR: Dependency lookup for libudev with method 'pkgconfig' failed: Invalid version, need 'libudev' ['>= 136'] found 'system'.
```

I've set the version in the pkg-config file to that of the version on the system. This fixes version checks in consumers using pkg-config.

### DIFF
--- a/recipes/libudev/all/conanfile.py
+++ b/recipes/libudev/all/conanfile.py
@@ -48,3 +48,7 @@ class LibUDEVConan(ConanFile):
         self.cpp_info.libdirs = []
         pkg_config = PkgConfig(self, "libudev")
         pkg_config.fill_cpp_info(self.cpp_info)
+        self.cpp_info.set_property("system_package_version", str(pkg_config.version))
+
+        # todo Remove this workaround for Conan v1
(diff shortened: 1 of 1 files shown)


uilianries:
Hello @jwillikers Could you please shared your entire build log? Please, try to follow the bug package template when sending hotfix without an issue associate, so we can reproduce your error quickly, without asking for extra information.

ericLemanissier (approved):

uilianries:
@jwillikers Thank your for the details, now I see your scenario!

uilianries (approved):
LGTM. The follow-up clarifies this fix.

PR #23014, opened 2024-03-07
libxslt: use version range for libxml2

temap:
Specify library name and version:  **libxslt/all**

---

### DIFF
--- a/recipes/libxslt/all/conanfile.py
+++ b/recipes/libxslt/all/conanfile.py
@@ -62,7 +62,7 @@ class LibxsltConan(ConanFile):
     def requirements(self):
         if Version(self.version) >= "1.1.39":
             # see https://github.com/conan-io/conan-center-index/pull/16205#discussion_r1149570846
-            self.requires("libxml2/2.12.3", transitive_headers=True, transitive_libs=True)
+            self.requires("libxml2/[>=2.12 <2.13]", transitive_headers=True, transitive_libs=True)
         else:
             self.requires("libxml2/2.11.6", transitive_headers=True, transitive_libs=True)
 


uilianries (dismissed):
@temap Thank you for your intention of using ranged versions.

Only few packages are allowed to use ranged versions: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges

Could you please point what issue you are trying to achieve?

temap:
@uilianries I intend to minimize security and operational risks when using third-party libraries, while also ensuring all changes are upstreamed to avoid unnecessary version bumps and minimizing the need for overrides in my consumer packages.

temap:
@uilianries I just wanted to mention that I've updated the patch version, but it hasn't been merged for a month, and a new patch version has been released. Improving the review process and setting up Dependabot properly could really help address these poorly maintained recipes and prevent attempts to create such pull requests in the future.

uilianries:
@temap thank you for your quick feedback! Indeed ranged versions make life easier than bumping recipes manually. The problem when building packages and disturbing them, they should be aligned, using the same version, otherwise, it will result in missing packages when having a dependency in common.
Right now there is no such mechanism to keep all aligned and fix automatically missing dependencies, so adding a range version would break packages in Conan Center. Instead, Conan team is adding manually that support, so they can check and build everything before having errors with CI. This should be fixed in the future, like building more than a package at the same PR, then aligning the dependencies. 

Thank you for your intention, but this PR cannot be merged.

temap:
@jcar87 @uilianries thank you so much for the feedback! 

Just wanted to let you know that for our product, we have a two-week requirement to fix CVEs.

Sometimes, tackling such tasks can be quite exhausting. I've found a solution by using Renovate, which automatically bumps our Conan dependencies. Right now, all I need to do is accept the merge request. However, in cases where libraries don't have ranges, I have to manually add an override in the commit.

I was wondering if there are any plans to incorporate Renovate for `conandata.yml`, or if you have your own bot that can handle dependency updates based on CVEs? Let me know your thoughts.

Thanks

RubenRBS:
Closing in favour of https://github.com/conan-io/conan-center-index/pull/23633 - We're moving ahead with version ranges for libxml2, and I didn't see this PR until I approved the newer one, sorry about that!

> I was wondering if there are any plans to incorporate Renovate for conandata.yml, or if you have your own bot that can handle dependency updates based on CVEs? Let me know your thoughts.

We're working on security tools to elp in this regard, yes :)

Issue #18912, opened 2023-07-26
[request] OpenSUSE <opengl>/<system>


chriztheanvill:
### Package Name/Version

opengl/system

### Changelog

https://github.com/nigels-com/glew/releases/tag/glew-2.2.0

### Context about the new update

Hi.
In opensuse trying to install glew, install opengl/system, and shows this message:
```
opengl/system: RUN: rpm -q glproto-devel
package glproto-devel is not installed

ERROR: opengl/system: Error in system_requirements() method, line 39
        zypper.install(["Mesa-libGL-devel", "glproto-devel"], update=True, check=True)
        ConanException: System requirements: 'glproto-devel' are missing but can't install because tools.system.package_manager:mode is 'check'.Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'

```

In opensuse the package glproto-devel does not exist. Solving: edit: .conan2/p/opengc3fc46f257b85/e/conanfile.py

```
38        zypper = package_manager.Zypper(self)
39        zypper.install(["Mesa-libGL-devel", "glproto-devel"], update=True, check=True)

to this
38        zypper = package_manager.Zypper(self)
39        zypper.install(["Mesa-libGL-devel"], update=True, check=True)

```

ericLemanissier:
Did you try what the error message says: `Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'`
The systems test of opengl on opensuse does pass on https://github.com/bincrafters/system-packages-checks/actions/runs/5655446415/job/15320677068#step:6:420 and https://github.com/bincrafters/system-packages-checks/actions/runs/5655446415/job/15320677266#step:6:347

chriztheanvill:
Yes.

> Did you try what the error message says: `Please update packages manually or set 'tools.system.package_manager:mode' to 'install' in the [conf] section of the profile, or in the command line using '-c tools.system.package_manager:mode=install'` The systems test of opengl on opensuse does pass on https://github.com/bincrafters/system-packages-checks/actions/runs/5655446415/job/15320677068#step:6:420 and https://github.com/bincrafters/system-packages-checks/actions/runs/5655446415/job/15320677266#step:6:347
> 
> see also https://pkgs.org/search/?q=glproto-devel&on=provides

Yes, tried and no success.

[...]

PR #23015, opened 2024-03-07
Add install_substitutes for zypper with "xorgproto-devel"

czoido:
Using install_substitutes in zypper because in opensuse tumbleweed `glproto-devel` is not available but `xorgproto-devel` provides the same.

I think this should close: https://github.com/conan-io/conan-center-index/issues/18912
And also close: https://github.com/conan-io/conan/issues/15335


jcar87 (approved):

uilianries (approved):
https://software.opensuse.org/package/xorgproto?locale=gl

https://repology.org/project/glproto/versions

glproto is Available for OpenSuse Leap.

PR #23016, opened 2024-03-07
Update range list to add libpng (And mention libxml)

RubenRBS:
After consideration, we'll allow libpng to the range list

As for the libxml dependency, we're also open to it being added, but still need to do some more research

### DIFF
--- a/docs/adding_packages/dependencies.md
+++ b/docs/adding_packages/dependencies.md
@@ -182,10 +182,12 @@ Currently, these are:
 * OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
 * CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts
 * Libcurl: `[>=X.YY <9]`, where `X.YY` is the minimum version of Libcurl required, starting from `7.78`
+* Zlib: `[>=1.2.11 <2]`
+* Libpng: `[>=1.X <2]`, where `1.X` is the minimum version of libpng required by the recipe
 
-> **Note**: You might also see Zlib ranges in some PR by CCI maintainers.
+> **Note**: You might also see Libpng & Libxml ranges in some PR by CCI maintainers.
 > We're adding them little by little to avoid missing binaries and conflict errors.
-> Please do not open PRs moving Zlib to ranges for now, we'll update this page when PRs are free to add new ranges.
+> Please do not open PRs moving Libpng to ranges for now, we'll update this page when PRs are free to add new ranges.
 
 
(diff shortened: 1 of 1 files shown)


docs/adding_packages/dependencies.md
```diff
@@ -182,10 +182,12 @@ Currently, these are:
 * OpenSSL: `[>=1.1 <4]` for libraries known to be compatible with OpenSSL 1.x and 3.x
 * CMake: `[>3.XX <4]`, where `3.XX` is the minimum version of CMake required by the relevant build scripts
 * Libcurl: `[>=X.YY <9]`, where `X.YY` is the minimum version of Libcurl required, starting from `7.78`
+* Zlib: `[>=1.2.11 <2]`
+* Libpng: `[>=1.X <2]`, where `1.X` is the minimum version of libpng required by the recipe
```
jcar87, line 186: I'd recommend `[>=1.6 < 6]`, unless a higher one is needed - I know it's unlikely we'll ever solve to 1.5, but those are too old.

uilianries: I agree, keeping a real version range is much safer for users when reading the documentation.

docs/adding_packages/dependencies.md
```diff
@@ -182,10 +182,12 @@ Currently, these are:
...
 
-> **Note**: You might also see Zlib ranges in some PR by CCI maintainers.
+> **Note**: You might also see Libpng & Libxml ranges in some PR by CCI maintainers.
```
jcar87, line 188: "You might also see version ranges in some PRs by CCI maintainers. These are being done on a case-by-case basis, and are being rolled out in phases to ensure that they do not cause problems to users."

(i.e. just because it's in a single recipe by us doesn't mean it should/can be in all recipes _unless it's on this list_))

docs/adding_packages/dependencies.md
```diff
@@ -182,10 +182,12 @@ Currently, these are:
...
 > We're adding them little by little to avoid missing binaries and conflict errors.
-> Please do not open PRs moving Zlib to ranges for now, we'll update this page when PRs are free to add new ranges.
+> Please do not open PRs moving Libpng to ranges for now, we'll update this page when PRs are free to add new ranges.
```
jcar87, line 190: I'd mention to please not open PRs where the only change is adding a version range to a requirement, unless the PR is solving a conflict (in which case the PR is very much welcome and will be prioritised).

uilianries (dismissed):
LGTM

docs/adding_packages/dependencies.md
uilianries: Good catch! Let's do it explicit, otherwise we may have new PRs only changing it.

uilianries (approved):
https://github.com/RubenRBS/conan-center-index/blob/rr/docs-ranegs-libpng/docs/adding_packages/dependencies.md#adding-version-ranges

I like it.

PR #23030, opened 2024-03-08
glib: Bump libmount to 2.39.2 to avoid conflicts with libsystemd

jwillikers:
Also avoids conflicts with libwebsockets and eudev.


jwillikers:
I'm a bit confused as to why this one broke.

jwillikers:
I think something breaks when `pkgconf` is bumped in the test package. It appears that `PkgConfig` is not actually using the `pkgconf` executable from the `tool_requires` but looking for an executable named `pkg-config`, which it doesn't always find.

valgur (approved):

ericLemanissier (approved):

danimtb (approved):

uilianries (approved):

uilianries (approved):
LGTM. All other recipes are using the newest version available in Conan Center:

```
../glib/all/conanfile.py:            self.requires("libmount/2.39")
../libsystemd/all/conanfile.py:        self.requires("libmount/2.39.2")
../libwebsockets/all/conanfile.py:            self.requires("libmount/2.39.2")
../eudev/all/conanfile.py:            self.requires("libmount/2.39.2")
```

PR #23033, opened 2024-03-08
add gettext 0.22.5

jcar87:
* Add `gettext/0.22.5` (latest version)
* Add patch (for 0.22.5 only) to force `use_macos_tools=yes` - this is needed when cross building, because Conan passes both the `host` and `build` triplets, and the build scripts otherwise assume older macOS
* Remove patches to remove `memmove` and `memset` from source files when building on Windows. The functions are built because the `ac_cv_func_memmove`  and `ac_cv_func_memset` checks are erroneously coming back with a "no" - simply setting them to `yes` during configure achieves the same effect.
* Simplify the test package - there's only 4 executables and the only useful thing the autotools project was doing with regards to this recipe, is checking that they are found in path.


jcar87:
There appears to be an issue where this won't "crossbuild" properly. Our CI service is now doing cross-builds from `armv8 ` to `x86_64` on macOS. We've verified that this is a cross-build issue, that is, it also doesn't work the other way around (`x86_64` to `armv8`) - currently investigating why...

RubenRBS (approved):
Yayyy

franramirez688 (approved):
👏

PR #23044, opened 2024-03-10
genie: fix compiler paths

Ahajha:
Specify library name and version:  **genie/all**

Previously, the build system would not respect `tools.build:compiler_executables`. On my system, I don't have a "g++" command, just a handful of profiles that specifically call "g(cc/++)-\<version\>" using that.

This also simplifies a decent amount of logic.

Now it should behave similar to other make-based packages:

1. Use `tools.build:compiler_executables` if defined
2. Use CC if defined
3. Fallback to cc/cxx

---


RubenRBS (approved):
Great catch!

toge (approved):

Issue #23058, opened 2024-03-11
`zlib` does not use a correct `pkg-config` name for the library with `clang`


mmomtchev:
https://github.com/conan-io/conan-center-index/blob/5b9ff5f781fee9e9eb8eaa9f78796e3277ec8855/recipes/zlib/all/conanfile.py#L103

Reproduction:
```
[requires]
zlib/[>=1.2.0]

[generators]
PkgConfigDeps
```
```
conan install .
```

I wonder what should be the correct test, since `clang` is not `mingw`? Or maybe `clang` should be `mingw`?

mmomtchev:
In `conanfile.py`:

```python
    def layout(self):
      if self.dependencies['zlib']:
        # https://github.com/conan-io/conan-center-index/issues/23058
        print('Monkey patching zlib')
        zlib = self.dependencies.host['zlib']._conanfile
        zlib._package_info = zlib.package_info
        def zlib_package_info():
          zlib._package_info()
          zlib.cpp_info.libs = ['z']
        zlib.package_info = zlib_package_info
```

PR #23060, opened 2024-03-11
[zlib] use the MinGW pkg-config name with clang on Windows

mmomtchev:
Specify library name and version:  **libz/1.3.1**

Refs: #23058 

---

### DIFF
--- a/recipes/zlib/all/conanfile.py
+++ b/recipes/zlib/all/conanfile.py
@@ -29,7 +29,7 @@ class ZlibConan(ConanFile):
 
     @property
     def _is_mingw(self):
-        return self.settings.os == "Windows" and (self.settings.compiler in ["gcc", "clang"])
+        return self.settings.os == "Windows" and self.settings.compiler in ["gcc", "clang"]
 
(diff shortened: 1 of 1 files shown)


recipes/zlib/all/conanfile.py
```diff
@@ -29,7 +29,7 @@ class ZlibConan(ConanFile):
 
     @property
     def _is_mingw(self):
-        return self.settings.os == "Windows" and self.settings.compiler == "gcc"
+        return self.settings.os == "Windows" and self.settings.compiler in ["gcc", "clang"]
```
SpaceIm, line 32: It would break clang-cl.

I've proposed a helper in conan client https://github.com/conan-io/conan/pull/12678, but it has been rejected.

mmomtchev: It got rejected because it was a conan 1.x issue. If this is common throughout conan, I think there must be a generic and centralized solution to it. Especially since these days, if someone is using a gcc-derivative on Windows, the most probable choice would be `clang` because of `libc++`.

PR #23086, opened 2024-03-13
metal-cpp: Add header only library metal-cpp from Apple

irieger:
Specify library name and version:  **metal-cpp/x**

Add metal-cpp header only library provided by Apple under the Apache 2.0 license: https://developer.apple.com/metal/cpp/

This library wraps some libraries provided by Apple as Objective-C interfaces into C++ interfaces. Headers only provide the wrapping and it requires one file doing the implementation (as described on the linked project page).

Important point: No clear versioning, packages only reference the version of the macOS/iOS SDKs they wrap into C++. Also the library requires relatively modern macOS/iOS versions to be compiled.

---


recipes/metal-cpp/all/src/implementation.cpp
uilianries, line 1: I would not include implementation in this package, instead, keep only the upstream content, the headers. Who is consuming this package will use those headers as needed.

irieger: Ok, sure. Can do that.
I'd like to have both options but if it is the preferred way, I can do the library thing locally for my project. But having multiple things consume it I'd prefer it if the package offers both. Maybe alternative with the header only as the default but still offering the implementation?

Also, is it valid to have the header only package build - as it doesn't need any tools - but the test package fail on all CI hosts. There it would not work as far as I looked at it as all build hosts seem to have an older OS on it.

uilianries: I understand you need to use this package, but it's your particular case using that specific implementation. I would suggest creating a local package, where you include metal-cpp. 

Packages in ConanCenterIndex should follow the upstream as close as possible. Some of them have patches, but nothing related to breaking behavior. 

Please, keep the upstream's header only for this PR.

irieger: Ok, cleaned it to be a pure header only library

irieger:
Some open questions to discuss:

1. Version naming: Current suggestion uses the version number of the target macOS version of the headers as the version number. Would it be better to use cci.xyz or something else?
2. Version checking and settings are a bit unclear to me here. "Building" the header only library obviously just requires unpacking the ZIP and packing it into a conan package. Applying it however depends mainly on the target SDK, so version "13" builds on everything targeting macOS 13 or newer, while version "14.2" requires a config with "os.version=14.2". How to best reflect this?

jcar87:
Hi @irieger - thank you so much for this PR and for investigating and troubleshooting the SDK version compatibility.

I have pushed a couple refactorings in this PR - while your PR is correct, I think _requiring_ the users to have `os.version` or `os.sdk_version` defined may be a tad too far, as doing so would cause all package IDs to be different due to the new settings, and it may be incompatible with what they had before.

As a fallback, I'm proposing to ask `xcrun` what the version of the SDK is - it is the SDK headers after all that define the compatible versions of macOS (and others). This should work in most cases when the other two are not defined, and should return the baseline SDK version (which is what ultimately will determine whether the required link libraries are available). One is to assume that if `xcrun` returns this SDK version, this is what will be used de-facto and that the target OS version will typically be just this (or lower), and if the user wants a target version - they would use `os.version` anyway.

Additionally - added checks for tvOS as well, since it falls under the "iOS" umbrella (if I'm not mistaken, iOS in this context is both iOS and ipadOS, and tvOS is versioned alongside those two). I've moved the dict of versions to Conandata.yml - as that seems to be a better place for "version-dependent" information.

irieger:
Hey @jcar87 thank you for the proposed changes. I think they make sense.

I just fixed one version requirement and did a small cleanup but from my point of view we could move forward.

Build all 4 versions locally with success.

AbrilRBS (approved):

jcar87 (approved):

jcar87:
> Hey @jcar87 thank you for the proposed changes. I think they make sense.
> 
> 
> 
> I just fixed one version requirement and did a small cleanup but from my point of view we could move forward.
> 
> 
> 
> Build all 4 versions locally with success.

Thanks so much!

irieger:
Nice. Thank you very much for the review and the general work. Working on my side projects is so much more fun now since I switched everything to conan.

PR #23087, opened 2024-03-13
Add libsndio recipe

spiderkeys:
libsndio/1.9.0

The current `libsndfile` recipe has a [line](https://github.com/conan-io/conan-center-index/blob/master/recipes/libsndfile/all/conanfile.py#L76C9-L76C137):
```py
tc.variables["CMAKE_DISABLE_FIND_PACKAGE_Sndio"] = True  # FIXME: missing sndio cci recipe (check whether it is really required)
```
This was ultimately causing issues for me within my machine's environment, for partially unknown reasons. I described my findings here (meant to file the issue here, though):
https://github.com/conan-io/conan/issues/15240

By adding this sndio recipe, the complete chain of deps when building `ffmpeg` in a configuration that pulls in `pulseaudio` can be resolved (ffmpeg->pulseaudio->libsndfile->libsndio), and I was able to get the test to pass using the `libsndio` package.

I know that PRs should only change one recipe at a time, but I've currently put in the commit that adjusts `libsndfile` as well, for visibility and in case I'm missing something. I will remove this if this PR moves forward/gets buy-in.

I also haven't written a test-package specifically for this recipe yet, but will also do that once any issues are addressed. I have tested it indirectly by doing a `conan create --version 6.1` for `ffmpeg` which had previously been exhibiting my linked issues.

`libsndio` is a little weird compared to other projects I've written recipes for in that it has a handwritten `configure` file and no `configure.ac`, so the standard approach for using the AutoToolsToolchain didn't really work. I have very little experience writing autotools recipes, so please let me know if there are ways to improve this.

---

^- will do this at some point


RubenRBS:
Thanks! The bad thing about modifying recipes/libsndfile/all/conanfile.py the current CI will complain and it won't try to compile the new recipe - I'd recommend removing that change from this PR and focusing on getting this merged knowing that we need to additionally check ffmpeg :)

spiderkeys:
Thanks @RubenRBS. Just requested access and removed the commit against libsndfile..

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8277348867/attempts/1)

recipes/libsndio/all/conanfile.py
```diff
@@ -0,0 +1,120 @@
...
+        tc.configure_args.append("--datadir=${prefix}/res")
+        tc.configure_args.append("LDFLAGS=-lm")
+        
+        if self.options.get_safe("with_alsa"):
+            tc.configure_args.append("--enable-alsa")
+
+        tc.generate()
+
+        tc = PkgConfigDeps(self)
+        tc.generate()
```
jwillikers, line 81: If `pkg-config` is used, you should include it in the tool requires like in the template package [here](https://github.com/conan-io/conan-center-index/blob/2c12fe30fd472a22e5f885b36e09b7551e7f627b/docs/package_templates/autotools_package/all/conanfile.py#L107).

jwillikers (changes requested):

recipes/libsndio/all/conanfile.py
spiderkeys: Just some leftover cruft from copying another recipe. Not needed, so removed.

spiderkeys:
Added a test package for conan v2. It passes, but it seems there is a missing dependency still:
```
libsndio/1.9.0@spiderkeys/test (test package): Running test()
libsndio/1.9.0@spiderkeys/test (test package): RUN: ./test_package
ALSA lib /home/conan/w/prod-v2/bsr/9663/cafde/p/b/libal9a348007750da/b/src/src/conf.c:4028:(snd_config_hooks_call) Cannot open shared library libasound_module_conf_pulse.so (//lib/alsa-lib/libasound_module_conf_pulse.so: cannot open shared object file: No such file or directory)
ALSA lib /home/conan/w/prod-v2/bsr/9663/cafde/p/b/libal9a348007750da/b/src/src/pcm/pcm.c:2675:(snd_pcm_open_noupdate) Unknown PCM hw:0
couldn't open play stream: No such file or directory
sio_open() failed
```
This `libasound_module_conf_pulse.so` appears to be an alsa plugin which is provided by:
https://github.com/alsa-project/alsa-plugins

recipes/libsndio/all/conanfile.py
jwillikers: Thanks!

spiderkeys:
No idea why the v1 pipeline is failing. I've run `conan create` for the package in conan docker containers with both conan v1.60 and conan2 (in conanio/gcc11-ubuntu16.04:latest) and the build+test succeeds...

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8293463420/attempts/1)

recipes/libsndio/all/conanfile.py
```diff
@@ -0,0 +1,135 @@
...
+            deps_cpp_info = dependency.cpp_info.aggregated_components()
+
+            dep_cflags.extend( deps_cpp_info.cflags + deps_cpp_info.defines )
+            for path in deps_cpp_info.includedirs:
+                dep_cflags.append( f"-I{path}" )
+
+            dep_ldflags.extend( deps_cpp_info.sharedlinkflags + deps_cpp_info.exelinkflags )
+            for path in deps_cpp_info.libdirs:
+                dep_ldflags.append( f"-L{path}" )
+            for lib in reversed(deps_cpp_info.libs + deps_cpp_info.system_libs):
```
ericLemanissier, line 86:
```suggestion
            for lib in deps_cpp_info.libs + deps_cpp_info.system_libs:
```
linker arguments should have the dependents before the dependencies, otherwise the dependencies are automatically dropped

recipes/libsndio/all/conandata.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+  "1.9.0":
+    "source":
+      url: "https://github.com/ratchov/sndio/archive/refs/tags/v1.9.0.tar.gz"
+      sha256: "4e8d77069b651d7c4bb2a50ba49004bd7fcfdfec1516f88a76457f831c17a546"
+    "license":
+      url: "https://raw.githubusercontent.com/ratchov/sndio/e10e4f836b88f8d5ea2864845e3712f560e6bed6/LICENSE"
+      sha256: "4ae392ab9cfbf2d6ac8522f5f4b025685d60c824979fad783aaf72d55165730f"
```
jwillikers, line 8: You'll want to copy the license out of t he downloaded source code instead. That's the typical idiom on CCI.

jwillikers (changes requested):

recipes/libsndio/all/conandata.yml
spiderkeys: It doesn't exist in any distribution of the source by the authors. They added the license to the git repo 9 months ago but the latest release is from 2022 so it doesn't contain it. The tarball on the authors website also does not contain the license.

The license is partially present in the actual source code files as well, but the authors are spread out across several different files.

I found the approach I took here in a number of CCI recipes, such as in bazel here:

https://github.com/conan-io/conan-center-index/blob/master/recipes/bazel/all/conandata.yml

jwillikers: Ah, I didn't realize it wasn't yet in a release. Thanks!

jwillikers (approved):

jwillikers (approved):

recipes/libsndio/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class LibsndioConan(ConanFile):
+    name = "libsndio"
+    license = "ISC"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://sndio.org/"
+    topics = ("sndio", "sound", "audio", "midi")
+    description = "A small audio and MIDI framework that provides a lightweight audio & MIDI server \
+        and a user-space API to access either the server or the hardware directly in a uniform way."
```
valgur, line 20:
```suggestion
    description = ("A small audio and MIDI framework that provides a lightweight audio & MIDI server"
                   "and a user-space API to access either the server or the hardware directly in a uniform way.")
```
To avoid unnecessary whitespace in the description.

recipes/libsndio/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+    default_options = {
+        "with_alsa": True,
+    }
+
+    def configure(self):
+        self.settings.rm_safe("compiler.libcxx")
+        self.settings.rm_safe("compiler.cppstd")
+
+    def validate(self):
+        if self.settings.os != "Linux":
```
valgur, line 35:
```suggestion
        if self.settings.os not in ["Linux", "FreeBSD"]:
```
I see it being available on FreeBSD (https://repology.org/project/sndio/versions) as well, so it's probably supported.

recipes/libsndio/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+        exclusions = ["--enable-shared", "--disable-shared", "--disable-static", "--enable-static", "--sbindir", "--oldincludedir"]
+        tc.configure_args = [arg for arg in tc.configure_args if not any(exclusion in arg for exclusion in exclusions)]
+
+        # Add alsa support
+        if self.options.get_safe("with_alsa"):
+            tc.configure_args.append("--enable-alsa")
+        else:
+            tc.configure_args.append("--disable-alsa")
+
+        # Inject conan dependency information into sndio configuration
```
valgur, line 91: A comment on why `AutotoolsDeps` is not sufficient here would not hurt.

valgur (approved):

RubenRBS:
Feel free to ping once you've addressed @valgur review :)

recipes/libsndio/all/conandata.yml
```diff
@@ -0,0 +1,8 @@
+sources:
+  "1.9.0":
+    "source":
+      url: "https://github.com/ratchov/sndio/archive/refs/tags/v1.9.0.tar.gz"
```
memsharded, line 4: Why the sources from https://sndio.org/install.html (https://sndio.org/sndio-1.9.0.tar.gz) aren't used? It seems they say their sources work in Linux, it is necessary to use this Github fork instead of the original project sources?

memsharded (commented):
It looks good.

recipes/libsndio/all/conandata.yml
spiderkeys: I had originally used those sources, but found they didn't include the license file. I thought the GitHub release would because I saw the license file on the main branch, so I switched to GitHub sources. It turned out that the license file had been added in a commit later than the latest release, though, so it ended up not mattering.

I can switch this back to the sndio.org sources, if it is preferable. They should be the same.

spiderkeys: It's worth noting that Alexandre Ratchov is one of the authors of sndio though, which is why I thought it would be acceptable. Their fork/repo is the only place where I could find a complete license file that includes all authors.

spiderkeys:
I've added the suggested changes from @valgur 's review. @memsharded let me know if you want the source URL changed, based on the above conversation, otherwise I think this should be ready to land.

danimtb (approved):

memsharded (approved):

recipes/libsndio/all/conandata.yml
memsharded: I have approved it, sorry for the delays, it is very difficult to keep up with everything.
I think in the future it would make sense to use the official URLs, but that is not critical, I preferred to approve to not further delay it, and it can be changed in the future.

Many thanks again for this contribution!

AbrilRBS (approved):

PR #23089, opened 2024-03-14
[Experiment] FFmpeg backend for Qt6 QtMultimedia and support for VAAPI drivers

spiderkeys:
Attempts to add support for building Qt6.4+'s FFMpeg backend for QtMultimedia.

This PR contains several changes to try and work towards functional ffmpeg-based hardware decoding (and other multimedia functions) within Qt6, which shifted focus from their gstreamer-based backend to ffmpeg in Qt6.4, so as to target all platforms with one library.

This is not meant to ever be merged, but to serve as grounds for discussion to figure out the right way to achieve these goals.

The impetus behind these changes come from difficulties we have had when trying to use conan-packaged ffmpeg within our Qt application. We were originally using Qt's official release binaries, but once Qt shifted to ffmpeg and added support for ffmpeg's hardware decoders, we discovered that we had a dependency conflict between the ffmpeg version Qt built and statically linked against (used by their MediaPlayer QML element, among others) and our conan-supplied shared version of ffmpeg+libva. (context, if desired: https://bugreports.qt.io/browse/QTBUG-122757)

We are now trying to shift to using the conan Qt recipe, but there is currently only support for the gstreamer qtmultimedia backend. Upon trying to follow the same process within the recipe as the gstreamer plugin, I ran into some hurdles:

1. Qt has a handful of its own FindX.cmake files which all differ in behavior, making it hard to get Qt to find conan dependencies. Some get overridden, while others explicitly do not. Some of their cmake modules do not follow the same patterns or implementations as the others, so each needs to be handled on a case-by-case basis.
2. `with_pulseaudio` was forbidden by the recipe due to some outstanding bug/build limitation
3. pulseaudio has a cmake find file called WrapPulseAudio, which I don't seem to properly have accounted for yet
4. libva also has a Find file, which I think I have properly added support for, but may be wrong

I've included a non-trivial Qt application, which I am trying to use to test the various aspects of the Qt recipe that I ultimately depend on. It isn't complete yet, because I'm stuck on the Qt6 package build with the current conan option set.

My current blocker is a long list of linking errors to pulseaudio symbols:
```
...
/usr/bin/ld: /home/spiderkeys/.conan2/p/b/pulse864d9de525735/b/src/src/pulse/stream.c:1694: undefined reference to `pa_log_level_meta'
/usr/bin/ld: /home/spiderkeys/.conan2/p/b/pulse864d9de525735/b/src/src/pulse/stream.c:1696: undefined reference to `pa_detect_fork'
/usr/bin/ld: /home/spiderkeys/.conan2/p/b/pulse864d9de525735/b/src/src/pulse/stream.c:1700: undefined reference to `pa_memblockq_get_length'
...
collect2: error: ld returned 1 exit status
[4954/5077] Building CXX object qtmulti...ediaPlugin.dir/qv4l2cameradevices.cpp.o

qt/6.6.2: ERROR: 
Package 'dd971a2f5642e628f9fa9524b36fa90ea936552b' build failed
qt/6.6.2: WARN: Build folder /home/spiderkeys/.conan2/p/b/qt098fadc0ba2b7/b/build/Debug
ERROR: qt/6.6.2: Error in build() method, line 813
        cmake.build()
```
So my next step is to figure out what is wrong with the attempted inclusion of pulseaudio into the recipe, most likely relating to the cmake find file.

Any advice/feedback on this effort is much appreciated.

---
Some random notes:
 - There was some issue with either my `nvidia-vaapi-driver` recipe or the `libva` recipe which prevented `nvidia-vaapi-driver`'s meson configure process from finding libdrm. I had a custom libva recipe of my own (from before there was one in conan center) which worked, so I replaced the libva recipe content with my own, just to get things working. This required a change in ffmpeg as well, to make sure it found the correct target components.
 - I added what I think should be necessary to enable `cuvid` decoding support in the ffmpeg recipe, such that it builds, but I haven't been able to test it yet.


spiderkeys:
I was able to get the pulseaudio errors resolved by changing it to `shared=True`. Cleaned up a bit more, though the test project is still quite a mess. That said, it now fully builds with all of my Qt features enabled, and a qml app window launches, but the media player cannot load/play a video yet.
image

Not sure why it is trying to load vdpau when I've asked for vaapi.

recipes/intel-media-driver/all/conanfile.py
```diff
@@ -0,0 +1,95 @@
...
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "intel-media-driver")
+        self.cpp_info.set_property("cmake_target_name", "intel-media-driver")
+        self.cpp_info.set_property("pkg_config_name", "intel-media-driver")
+
+        self.cpp_info.components["intel-media-driver"].libs = collect_libs(self)
```
jwillikers, line 91: You should avoid using `collect_libs` if possible. Since this is a runtime-loadable plugin, I'm not sure that you even need to provide a library in this way.

You'll want to make the VA-API libraries accessible at runtime:

```
self.runenv_info.prepend_path("LIBVA_DRIVERS_PATH", os.path.join(self.package_folder, "lib", "dri"))
```

jwillikers (changes requested):

recipes/intel-media-driver/all/conanfile.py
jwillikers: Yeah, I think it might be worth setting here and letting consumer's figure out whether or not to include it as a dependency as appropriate for the specific system.

spiderkeys: One thought just now, how would you expect to handle the case where there are two VAAPI driver packages that both do this?

As an example, in our app we pull in both `intel-media-driver` and `nvidia-vaapi-driver` and the correct one can be used by the application based on the hardware it is running on. There can only be one `LIBVA_DRIVERS_PATH`, so it seems like the value couldn't be the individual package folder itself in that case?

jwillikers: Seriously? That seems like a ridiculous limitation. Do you know if there's a good reason for that?

jwillikers: Hmm, I found this: https://github.com/canonical/snapcraft/pull/3591/files which seems to indicate someone at least tried to append a path to that variable.

jwillikers: Aha! It should work with multiple paths: https://github.com/intel/libva/issues/639#issuecomment-1385276574

spiderkeys: https://github.com/intel/libva/blob/907b2b5405ca1091b4360bf35060e143bd704b62/va/va.c#L393
Great thanks, just validated it myself in the source. Good find!

spiderkeys:
As of my last push, I now believe I have everything working together. Qt's MediaPlayer loads a remote MP4 containing h264, loads the FFMpeg plugin, and FFmpeg reports that it has found our conan-provided `libva` and `intel-media-driver` and loaded them as the HWaccel decoder:
```
"FFmpeg log: Trying to use DRM render node for device 0."
"FFmpeg log: libva: VA-API version 1.20.0"
"FFmpeg log: libva: User environment variable requested driver 'iHD'"
"FFmpeg log: libva: Trying to open /home/spiderkeys/z/workspace/conan-center-fork/qt_test_project/build/x86_64/Debug/thirdparty/dri/iHD_drv_video.so"
"FFmpeg log: libva: Found init function __vaDriverInit_1_20"
"FFmpeg log: libva: va_openDriver() returns 0"
"FFmpeg log: Initialised VAAPI connection: version 1.20"
"FFmpeg log: Format 0x41524742 -> bgra."
"FFmpeg log: Format 0x42475241 -> argb."
"FFmpeg log: Format 0x41424752 -> rgba."
"FFmpeg log: Format 0x52474241 -> abgr."
"FFmpeg log: Format 0x58524742 -> bgr0."
"FFmpeg log: Format 0x42475258 -> 0rgb."
"FFmpeg log: Format 0x58424752 -> rgb0."
"FFmpeg log: Format 0x52474258 -> 0bgr."
"FFmpeg log: Format 0x30335241 -> unknown."
"FFmpeg log: Format 0x30334241 -> unknown."
"FFmpeg log: Format 0x30335258 -> x2rgb10le."
"FFmpeg log: Format 0x30334258 -> unknown."
"FFmpeg log: Format 0x36314752 -> unknown."
"FFmpeg log: Format 0x50424752 -> unknown."
"FFmpeg log: Format 0x50524742 -> unknown."
"FFmpeg log: Format 0x56555941 -> unknown."
"FFmpeg log: Format 0x30303859 -> gray."
"FFmpeg log: Format 0x3231564e -> nv12."
"FFmpeg log: Format 0x3132564e -> unknown."
"FFmpeg log: Format 0x32595559 -> yuyv422."
"FFmpeg log: Format 0x59565955 -> uyvy422."
"FFmpeg log: Format 0x32315659 -> yuv420p."
"FFmpeg log: Format 0x30323449 -> yuv420p."
"FFmpeg log: Format 0x50313134 -> yuv411p."
"FFmpeg log: Format 0x48323234 -> yuv422p."
"FFmpeg log: Format 0x56323234 -> yuv440p."
"FFmpeg log: Format 0x50343434 -> yuv444p."
"FFmpeg log: Format 0x33434d49 -> unknown."
"FFmpeg log: Format 0x30313050 -> p010le."
"FFmpeg log: VAAPI driver: Intel iHD driver for Intel(R) Gen Graphics - 23.4.3 ()."
"FFmpeg log: Driver not found in known nonstandard list, using standard behaviour."
"FFmpeg log: detected 8 logical cores"
"FFmpeg log: nal_unit_type: 7(SPS), nal_ref_idc: 3"
"FFmpeg log: Decoding VUI"
"FFmpeg log: nal_unit_type: 8(PPS), nal_ref_idc: 3"
```
image

I also fixed some issues with qtimageformat while I was at it, and brought in my fixes to the whole libsndio->pulseaudio chain, and tested that audio decoding/playback through ffmpeg was also working correctly.

Now that everything works as a proof of concept, if there is any advice/feedback on any element of the approach at a high level, I'd be glad to hear and iterate on it here for a bit longer. 

From there, I can split things out into individual PRs per recipe and close this PR.

uilianries:
Closing this PR to avoid new CI actions, as it has multiple recipes. 

Please, feel free to open new PRs splitting it. Thank you!

PR #23091, opened 2024-03-14
python-packaging 24.0

ericLemanissier:
Specify library name and version:  **python-packaging/24.0**

---

### DIFF
--- a/recipes/python_packaging/all/conanfile.py
+++ b/recipes/python_packaging/all/conanfile.py
@@ -1,5 +1,6 @@
 from conan import ConanFile
 from conan.tools.files import copy, get
+from conan.tools.layout import basic_layout
 import os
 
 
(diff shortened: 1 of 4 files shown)


ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8278029715/attempts/1)

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8278029715/attempts/2)

uilianries:
I have a branch with this package here: https://github.com/uilianries/conan-center-index/tree/package/python-packaging/recipes/python-packaging

Still under study with the Conan team. Please, be patient, there are many points to be considered.

jwillikers (approved):

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8278029715/attempts/3)

jwillikers (approved):

recipes/python_packaging/all/conanfile.py
```diff
@@ -0,0 +1,45 @@
...
+from conan.tools.files import copy, get
+from conan.tools.layout import basic_layout
+import os
+
+
+required_conan_version = ">=1.47.0"
+
+
+class PackageConan(ConanFile):
+    name = "python_packaging"
```
Croydon, line 11: We should have consistence naming. #23098 used a hypen instead of underscore.
Personally, I would prefer a hypen. It is also closer to how we handle double package names.

Croydon, line 11:
```suggestion
    name = "python-packaging"
```

uilianries (changes requested):
Python packages are not allowed in CCI for now. Still under discussion with Conan team to have a better insight. Thank you for investigating.

recipes/python_packaging/all/conanfile.py
jwillikers: I used a hyphen to stay as close as possible to how Linux distributions manage Python packages, fyi.

PR #23096, opened 2024-03-14
emsdk: Update emsdk versions + add Node.js version option

johannes-wolf:
Specify library name and version:  **emskd/1.3.X**

- Add versions up to 1.3.55
- Add an option to set the node.js Version used (defaulting to the previous constant 16.3.0)

There is another PR bumping emsdk to 1.3.53. I can rebase my PR if that one gets merged.

---

### DIFF
--- a/recipes/emsdk/all/conandata.yml
+++ b/recipes/emsdk/all/conandata.yml
@@ -1,4 +1,19 @@
 sources:
+  "3.1.55":
+    url: "https://github.com/emscripten-core/emsdk/archive/3.1.55.tar.gz"
+    sha256: "86a6af30e43d7b501772e5d2993457c924b73f1d1c0a3484a8c6c48452af549f"
+  "3.1.54":
+    url: "https://github.com/emscripten-core/emsdk/archive/3.1.54.tar.gz"
+    sha256: "e19eb3cfd2f8c6c2e3404467396f07733837c1e6f66d964bd29095b083186990"
+  "3.1.53":
+    url: "https://github.com/emscripten-core/emsdk/archive/3.1.53.tar.gz"
+    sha256: "1ae9753740dddd271a18bd6ce874bf0c96fa59cc64783fbf23dec263047375c0"
+  "3.1.52":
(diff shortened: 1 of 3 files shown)


johannes-wolf:
How can I see the build errors from the CI?

recipes/emsdk/all/conandata.yml
```diff
@@ -1,4 +1,19 @@
 sources:
+  "3.1.55":
```
uilianries, line 2: @johannes-wolf do you really need all these version? Otherwise, could please keep only the latest version available. Building all these versions will only consume CI resource basically. For who needs other versions, we can add by issue request.

recipes/emsdk/all/conanfile.py
```diff
@@ -19,6 +19,18 @@ class EmSDKConan(ConanFile):
...
+            "16.3.0",
+            "16.20.2",
+            "18.15.0"
+        ]
+    }
+
+    default_options = {
+        "nodejs_version": "16.3.0"
+    }
+
```
uilianries, line 33:
```suggestion
```
Please, don't do that. The requirement should be package of the package ID already, adding it as an option is a good approach, you will can use use a recipe and override to use the version that you want.

recipes/emsdk/all/conanfile.py
```diff
@@ -29,7 +41,7 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("nodejs/16.3.0")
+        self.requires("nodejs/" + str(self.options.nodejs_version))
```
uilianries, line 44:
```suggestion
        self.requires("nodejs/16.3.0")
```

Please, revert it.

uilianries (changes requested):
@johannes-wolf Please, do not add dependencies versions as options, they should be part of the recipe directly. Please, take a look in the documentation, about conflict https://docs.conan.io/2/tutorial/versioning/conflicts.html, in case you need to use another version.

johannes-wolf:
Closed it favor of `replace_requires`.

PR #23108, opened 2024-03-15
xz_utils: skip building tests

Ahajha:

---

### DIFF
--- a/recipes/xz_utils/cmake/conanfile.py
+++ b/recipes/xz_utils/cmake/conanfile.py
@@ -57,6 +57,7 @@ class XZUtilsConan(ConanFile):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.variables["BUILD_TESTING"] = False
         tc.generate()
 
     def build(self):


recipes/xz_utils/cmake/conanfile.py
```diff
@@ -57,6 +57,7 @@ def source(self):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.variables["BUILD_TESTING"] = False
```
uilianries, line 60: @Ahajha Why do you need it? The upstream is False by default and we always re-check CMakeList.txt when bumping a version. Did you find any bug? 

https://github.com/tukaani-project/xz/blob/v5.6.0/CMakeLists.txt#L1975

jwillikers: @uilianries This is one of those weird built-in CMake variables that's automatically enabled when including the `CTest` module: https://cmake.org/cmake/help/latest/module/CTest.html

jwillikers (approved):

recipes/xz_utils/cmake/conanfile.py
Ahajha: Not strictly necessary, but I noticed that the tests were being built when making the package, I figured it would be a good idea to skip those.

RubenRBS (approved):

PR #23111, opened 2024-03-15
pulseaudio/17.0: Add missing tool requirement for m4

jwillikers:
PulseAudio 17.0 requires `m4`, even though it's using Meson as the build system. The AutoTools build system for previous versions most likely brought in `m4` transitively, which is why it wasn't explicitly required.

Attached are logs for both before and after this change.
[pulseaudio-17-missing-m4-log.txt](https://github.com/conan-io/conan-center-index/files/14620932/pulseaudio-17-missing-m4-log.txt)
[pulseaudio-17-with-m4-log.txt](https://github.com/conan-io/conan-center-index/files/14620947/pulseaudio-17-with-m4-log.txt)

### DIFF
--- a/recipes/pulseaudio/meson/conanfile.py
+++ b/recipes/pulseaudio/meson/conanfile.py
@@ -76,9 +76,10 @@ class PulseAudioConan(ConanFile):
                 )
 
     def build_requirements(self):
-        self.tool_requires("meson/1.3.1")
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):
Thanks for attaching the logs, they are super helpful in these ocasions, I really appreciate it :)

franramirez688 (approved):

PR #23112, opened 2024-03-16
OpenColorIO: Fix dynamic linking for OpenColorIO >= 2.3.1 & fix package_info

irieger:
Specify library name and version:  **opencolorio/>=2.3.1**

I'm running into problems building my project with dynamic linking and one of the two culprits is OpenColorIO which currently prevents real dynamic linking and doesn't advertise its dependencies.

---


recipes/opencolorio/all/patches/2.3.1-0001-fix-cmake-source-dir-and-targets.patch
```diff
@@ -59,9 +59,18 @@ index 175d89c2..ac93b87a 100644
...
-index 26b4bb4c..d54cff3d 100755
+index 26b4bb4c..f366bdd1 100755
 --- src/OpenColorIO/CMakeLists.txt
 +++ src/OpenColorIO/CMakeLists.txt
+@@ -301,7 +301,7 @@ target_include_directories(OpenColorIO
+ )
+ 
+ target_link_libraries(OpenColorIO
+-    PRIVATE
++    PUBLIC   # Change to public so that the tools properly link
```
RubenRBS, line 70: Has this been submitted upstream? This looks like something that would better live as an official change upstream, where it can be properly maintained

RubenRBS (commented):
Hi! Thanks a lot! I have a small question regarding the CMakeLists change :)

recipes/opencolorio/all/patches/2.3.1-0001-fix-cmake-source-dir-and-targets.patch
irieger: I'm going to put up a pull request, but I'm currently working on getting into the CLA for the Linux Foundation which the upstream is part of. But I can also start the pull request later without the CLA standing and retrigger the check later. (And then reference the PR in the conandata)

uilianries: @irieger If you have a blocker with CLA because your job (for instance), don't worry, we can send the patch too. Thank you for checking it!

irieger: No, the job is not a problem. It is more the other way round. Trying to get the CLA through my company (we have a company CLA there) as I don't like the Linux foundation requireing a physical address, which is why I refuse to sign it with my personal details. (An open source focused organisation like that having such a bad privacy statement is a red flag for me when asked for personal details.)

Me doing this is no problem for my company and although this contributions here are for my personal interest, I guess it is good to be active in open source projects from a company perspective anyway and I have some colleagues contributing to the Academy Software Foundation projects (ASFW is hosted - or whatever the right english word is here - by the Linux Foundation).

irieger: Created a PR: https://github.com/AcademySoftwareFoundation/OpenColorIO/pull/1958

irieger: As this is anyway only a small part of the overall compatibility conan patching, I guess it doesn't make sense to add the upstream URL maybe?

jcar87 (changes requested):
Thanks for the contribution @irieger -

I'm not 100% changing the private to public is the right solution. By this I mean that it _may_ solve the problem, but it may not be the right solution :)

Any chance we can get a reproducible example of the problems you are running into? including the actual build errors that you are getting?

So that we can verify this and ensure it is the right solution (it's also entirely possible that this is a Conan limitation, which I'm currently investigating).

irieger:
> Thanks for the contribution @irieger -
> 
> I'm not 100% changing the private to public is the right solution. By this I mean that it _may_ solve the problem, but it may not be the right solution :)
> 
> Any chance we can get a reproducible example of the problems you are running into? including the actual build errors that you are getting?
> 
> So that we can verify this and ensure it is the right solution (it's also entirely possible that this is a Conan limitation, which I'm currently investigating).

https://github.com/conan-io/conan-center-index/issues/23421 is basically what I have as analysis and understanding. Logically to me the public is wrong too, but for some reason the Path of dependencies of Minizip isn't passed in non-public linking to the downstream executables. (The other alternative that works is manually adding minizip to every one of the executables that fails linking.)
But as the same happens in OpenImageIO (as well in my project consuming this dependnecies), I think there is a larger problem in shared linking regarding the runpath handling.

irieger:
Trying to extract the relevant parts with #24395

PR #23113, opened 2024-03-16
lielab: new package and v0.1.1

msparapa:
Adding new package to ConanCenter: Lielab.

lielab/0.1.1

I am the author of this library.

---


recipes/lielab/all/conanfile.py
```diff
@@ -0,0 +1,60 @@
...
+
+    exports_sources = ("CMakeLists.txt",
+                       "LICENSE",
+                       "lielab/*")
+    
+    no_copy_source = True
+
+    def requirements(self):
+        self.requires("eigen/3.4.0")
+
```
valgur, line 29:
```suggestion

    def build_requirements(self):
        self.tool_requires("cmake/[>=3.23 <4]")
```
Need to specify CMake tool_requires when requiring CMake newer than v3.15.

recipes/lielab/all/conanfile.py
```diff
@@ -0,0 +1,60 @@
...
+                  " on these spaces."
+    topics = ("Lie-theory", "Lie-group", "Lie-algebra", "numerical", "header-only")
+    package_type = "header-library"
+    license = "MIT"
+
+    settings = "os", "arch", "compiler", "build_type"
+
+    exports_sources = ("CMakeLists.txt",
+                       "LICENSE",
+                       "lielab/*")
```
valgur, line 23:
```suggestion
```
These are only necessary for in-source recipes and are not needed on CCI.

recipes/lielab/all/conanfile.py
valgur, line 1: Does the library require a minimal C++ version? If it's C++11 or greater, it should be checked by the recipe. See https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py for an example.

valgur (changes requested):
LGTM, just a few minor tweaks required.

recipes/lielab/all/conanfile.py
uilianries, line 29:
```suggestion

    @property
    def _min_cppstd(self):
        return 20

    @property
    def _compilers_minimum_version(self):
        return {
            "gcc": "11",
            "clang": "12",
            "apple-clang": "13.1",
            "Visual Studio": "17",
            "msvc": "193",
        }

    def validate(self):
        if self.settings.compiler.cppstd:
            check_min_cppstd(self, self._min_cppstd)
        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
            raise ConanInvalidConfiguration(f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support.")

```

Requires C++20: 

https://github.com/sandialabs/Lielab/blob/v0.1.1/CMakeLists.txt#L131

msparapa:
> @msparapa Thank you for your PR! Could you please sign the CLA commented in this PR? Also, your username is listed already in the authorized users.

On it. Sorry, my institution has a fairly slow turnaround with these things.

msparapa:
Hi @uilianries . I did not try locally with any of the conan-center hooks but I think otherwise it's good to go now.

uilianries:
@msparapa Thank you for checking locally. I see this is a new recipe and missing the test package part. Would mind adding the test package folder and adding a very minimum test to validate the package? We have a template for header-only projects, so you don't start it from the scratch: https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates/header_only/all

msparapa:
I added the test package. Question though, as of v0.3.0, the way Lielab gets imported changes so the `#include <lielab/lielab>` becomes `#include <Lielab.hpp>`. In the future, how would I set up the recipe so that there's a different test_package for versions < 0.3.0 and versions >= 0.3.0?

msparapa:
Hi @uilianries . I think this is good to go now. I fixed some of the errors on the main branch and changed the version to 0.4.0 since it was the easiest thing to do. I think the remaining failures are from Conan v1 not using the `validate()` method to make sure it's using C++ 20 or higher which causes it to not compile, but let me know if there's other things I need to fix.

uilianries (approved):
LGTM. Sorry the delay to review. I pushed a small fix to the test package.

perseoGI (approved):
LGTM

PR #23117, opened 2024-03-16
libassert: add 2.0.0

jeremy-rifkin:
Specify library name and version:  **libassert/2.0**

I'm the author of the library

---

### DIFF
--- a/recipes/libassert/all/conandata.yml
+++ b/recipes/libassert/all/conandata.yml
@@ -1,5 +1,9 @@
 sources:
   # Newer versions at the top
+  "2.0.0":
+    url:
+      - "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v2.0.0-beta.tar.gz"
+    sha256: "8615f10979bf3a940f6b19920050f4aa1ee12c9a3a3d5e74aa431136c7b6fa73"
   "1.2.2":
     url:
       - "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v1.2.2.tar.gz"
--- a/recipes/libassert/all/conandata.yml
+++ b/recipes/libassert/all/conandata.yml
@@ -2,8 +2,8 @@ sources:
   # Newer versions at the top
   "2.0.0":
     url:
-      - "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v2.0.0-beta.tar.gz"
-    sha256: "8615f10979bf3a940f6b19920050f4aa1ee12c9a3a3d5e74aa431136c7b6fa73"
+      - "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v2.0.0.tar.gz"
+    sha256: "d4b2da2179a94637b34d18813a814531a1eceb0ddc6dd6db6098050dd638f4a1"
   "1.2.2":
     url:
       - "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v1.2.2.tar.gz"
@@ -25,6 +25,7 @@ patches:
   "2.0.0":
     - patch_file: "patches/2.0.0/0001-fix-export-and-include.patch"
       patch_type: "bugfix"
+      patch_source: https://github.com/jeremy-rifkin/libassert/commit/25c1f3e43737ab18490a0d9430cb1c70f976a662
       patch_description: "Fix export header and an incorrect #include"
   "1.2.2":
     - patch_file: "patches/1.2.2/0001-cpptrace-dll-copy.patch"
--- a/recipes/libassert/all/conandata.yml
+++ b/recipes/libassert/all/conandata.yml
@@ -22,6 +22,10 @@ sources:
     sha256: "e1bb3b50767994ca4d0f60b7977b279cf32b8569ff92c5830e7a1de567b82fd5"
 
 patches:
+  "2.0.0":
+    - patch_file: "patches/2.0.0/0001-fix-export-and-include.patch"
+      patch_type: "bugfix"
+      patch_description: "Fix export header and an incorrect #include"
   "1.2.2":
     - patch_file: "patches/1.2.2/0001-cpptrace-dll-copy.patch"
       patch_type: "conan"
--- a/recipes/libassert/all/conanfile.py
+++ b/recipes/libassert/all/conanfile.py
@@ -158,14 +158,13 @@ class LibassertConan(ConanFile):
             self.cpp_info.components["assert"].names["cmake_find_package_multi"] = "assert"
             self.cpp_info.components["assert"].requires = ["cpptrace::cpptrace"]
             self.cpp_info.components["assert"].libs = ["assert"]
+            if not self.options.shared:
+                self.cpp_info.components["assert"].defines.append("LIBASSERT_STATIC_DEFINE")
         else:
             self.cpp_info.filenames["cmake_find_package"] = "assert"
             self.cpp_info.filenames["cmake_find_package_multi"] = "assert"
             self.cpp_info.names["cmake_find_package"] = "assert"
-            self.cpp_info.names["cmake_find_package_multi"] = "assert"
-
-        if not self.options.shared:
-            self.cpp_info.defines.append("LIBASSERT_STATIC_DEFINE")
+            self.cpp_info.names["cmake_find_package_multi"] = "assert"        
 
         if Version(self.version) < "1.2.1":
             # pre-cpptrace
--- a/recipes/libassert/all/conanfile.py
+++ b/recipes/libassert/all/conanfile.py
@@ -158,6 +158,9 @@ class LibassertConan(ConanFile):
             self.cpp_info.filenames["cmake_find_package_multi"] = "assert"
             self.cpp_info.names["cmake_find_package"] = "assert"
             self.cpp_info.names["cmake_find_package_multi"] = "assert"
+            
+        if not self.options.shared:
+            self.cpp_info.defines.append("LIBASSERT_STATIC_DEFINE")
 
         if Version(self.version) < Version("1.2.1"):
             # pre-cpptrace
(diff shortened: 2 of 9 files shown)


recipes/libassert/all/conanfile.py
```diff
@@ -82,7 +84,14 @@ def generate(self):
         if is_msvc(self):
             tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
 
-        if Version(self.version) >= Version("1.2.1"):
+        if Version(self.version) >= Version("2.0.0"):
+            # if not self.options.shared:
+                # tc.variables["LIBASSERT_STATIC_DEFINE"] = True
+            tc.variables["LIBASSERT_USE_EXTERNAL_CPPTRACE"] = True
+            tc.variables["LIBASSERT2"] = True
```
jeremy-rifkin, line 91: I found I needed to do this to check in the test package whether to do `find_package(libassert)` or `find_package(assert)`. Not sure if there is a better solution.

recipes/libassert/all/conanfile.py
```diff
@@ -52,13 +52,15 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        if Version(self.version) >= Version("1.2.2"):
+        if Version(self.version) >= Version("2.0.0"):
+            self.requires("cpptrace/0.4.1", transitive_headers=True)
```
jeremy-rifkin, line 56: Can be 0.5.0 after #23119

recipes/libassert/all/conanfile.py
```diff
@@ -52,13 +52,16 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        if Version(self.version) >= Version("1.2.2"):
+        if Version(self.version) >= "2.0.0":
+            # libassert::detail::process_assert_fail
+            self.requires("cpptrace/0.4.1", transitive_headers=True, transitive_libs=True)
```
RubenRBS, line 57:
```suggestion
            self.requires("cpptrace/0.5.0", transitive_headers=True, transitive_libs=True)
```

jeremy-rifkin:
I’ll check the remaining errors later tonight, what what I briefly skimmed it looks like there might be a missing export annotation on the libassert side of things

jeremy-rifkin:
Ok - my attempted fix was not correct. There is something up with the export macro definition but `libassert_lib_EXPORTS` is the correct define from cmake. Somehow on static the library is seeing LIBASSERT_STATIC_DEFINE, which is correct, but the test package is not so it's doing dllimport/export stuff.

jeremy-rifkin:
The libassert-config.cmake is also properly configured as
```
...
if(TRUE)
    target_compile_definitions(libassert::assert INTERFACE LIBASSERT_STATIC_DEFINE)
endif()
```so the test package should be getting the define... I'll have to look at this more tomorrow

**Comment by jeremy-rifkin** [comment]
Oh but testing on windows I see a different libassert-config.cmake placed in `conan-center-index/recipes/libassert/all/test_package/build/msvc-193-x86_64-17-release/generators`... Still not sure why there was no issue with cpptrace with this setup.

**Comment by jeremy-rifkin** [comment]
To update:

The conanfile for libassert sets `self.cpp_info.defines.append("LIBASSERT_STATIC_DEFINE")` properly for static, which disables the `__declspec(dllimport/dllexport)` in the library's header.

This appears to be working for building of the library and testing locally on windows I see the built library not containing export symbols.

However, the linker error indicates the test package is looking for a dllexport symbol so the `LIBASSERT_STATIC_DEFINE` define must not be propagating to the test package, and testing locally I confirmed this. Is there anything that could this?


What I’m most puzzled by is the setup here is essentially identical to the setup for my other library, cpptrace, but there were no issues there.

**Comment by uilianries** [comment]
I'm on Windows™ right now. I'll take a look.

File: recipes/libassert/all/conanfile.py [hunk 4]
```diff
@@ -124,22 +133,46 @@ def package(self):
...
+            self.cpp_info.components["assert"].requires = ["cpptrace::cpptrace"]
+            self.cpp_info.components["assert"].libs = ["assert"]
+        else:
+            self.cpp_info.filenames["cmake_find_package"] = "assert"
+            self.cpp_info.filenames["cmake_find_package_multi"] = "assert"
+            self.cpp_info.names["cmake_find_package"] = "assert"
+            self.cpp_info.names["cmake_find_package_multi"] = "assert"
+
+        if not self.options.shared:
+            self.cpp_info.defines.append("LIBASSERT_STATIC_DEFINE")
```
uilianries, line 168:
```suggestion
            self.cpp_info.components["assert"].libs = ["assert"]
            if not self.options.shared:
                self.cpp_info.components["assert"].defines.append("LIBASSERT_STATIC_DEFINE")
        else:
            self.cpp_info.filenames["cmake_find_package"] = "assert"
            self.cpp_info.filenames["cmake_find_package_multi"] = "assert"
            self.cpp_info.names["cmake_find_package"] = "assert"
            self.cpp_info.names["cmake_find_package_multi"] = "assert"        
```

uilianries:
@jeremy-rifkin I found that the definition should be exported via assert component too. Exporting it as global does not work. It worked here in my machine™ for both static and shared.

recipes/libassert/all/conandata.yml
```diff
@@ -1,5 +1,9 @@
 sources:
   # Newer versions at the top
+  "2.0.0":
+    url:
+      - "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v2.0.0-beta.tar.gz"
```
uilianries, line 5:
```suggestion
    url: "https://github.com/jeremy-rifkin/libassert/archive/refs/tags/v2.0.0.tar.gz"
```

Please, use the correct version. We try to avoid betas in case there is a stable release close to be finished.

recipes/libassert/all/conandata.yml
```diff
@@ -18,6 +22,10 @@ sources:
     sha256: "e1bb3b50767994ca4d0f60b7977b279cf32b8569ff92c5830e7a1de567b82fd5"
 
 patches:
+  "2.0.0":
+    - patch_file: "patches/2.0.0/0001-fix-export-and-include.patch"
```
uilianries, line 26: @jeremy-rifkin Would it be possible to integrate this patch in the upstream?

recipes/libassert/all/conanfile.py
```diff
@@ -52,13 +52,16 @@ def layout(self):
...
+        elif Version(self.version) >= "1.2.2":
             self.requires("cpptrace/0.3.1")
-        elif Version(self.version) >= Version("1.2.1"):
+        elif Version(self.version) >= "1.2.1":
             self.requires("cpptrace/0.2.1")
 
     def validate(self):
-        if self.settings.compiler == "apple-clang":
+        if Version(self.version) <= "2.0.0" and self.settings.compiler == "apple-clang":
             raise ConanInvalidConfiguration("apple-clang not supported")
```
uilianries, line 65:
```suggestion
        if Version(self.version) <= "2.0.0" and is_apple_os(self):
            raise ConanInvalidConfiguration(f"{self.ref} is not supported on Mac. Please, update to version >=2.0.0")
```

recipes/libassert/all/conanfile.py
```diff
@@ -82,7 +85,13 @@ def generate(self):
         if is_msvc(self):
             tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
 
-        if Version(self.version) >= Version("1.2.1"):
+        if Version(self.version) >= "2.0.0":
+            # if not self.options.shared:
+                # tc.variables["LIBASSERT_STATIC_DEFINE"] = True
```
uilianries, line 90:
```suggestion
```

Not needed here, the upstream CMakeLists.txt manages it automatically.

recipes/libassert/all/conanfile.py
```diff
@@ -109,7 +118,7 @@ def package(self):
             copy(
                 self,
                 "*.dll",
-                src=self.build_folder,
+                src=os.path.join(self.package_folder, "lib"),
```
uilianries, line 121: Why needed to change it now? I see the version 2.0.0 install in the bin folder already.

uilianries (commented):
I would say splitting this recipe in different folders for each major version (1.x and 2.x) would not be a bad idea. The recipe looks polluted, but it could wait to a next PR, let's make it work first.

recipes/libassert/all/conandata.yml
jeremy-rifkin: Thank you for catching this, I had this here as I was trying to iron out the release upstream and I forgot to bump :)

recipes/libassert/all/conandata.yml
jeremy-rifkin: Yes indeed, already done at https://github.com/jeremy-rifkin/libassert/commit/25c1f3e43737ab18490a0d9430cb1c70f976a662.

recipes/libassert/all/conanfile.py
jeremy-rifkin: I don't quite recall, I'll try reverting it and see if anything breaks

jeremy-rifkin:
I think splitting into another folder would be helpful, happy to do it now or for a later PR. Will wait for CI to finish at the very least.

valgur (approved):

RubenRBS (approved):

RubenRBS:
> I think splitting into another folder would be helpful, happy to do it now or for a later PR. Will wait for CI to finish at the very least.

Good idea, thanks :)

jeremy-rifkin:
Thank you everyone so much for the help getting this sorted out!

PR #23136, opened 2024-03-17
cpp-channel: add recipe

toge:
Specify library name and version:  **cpp-channel/***

---


recipes/cpp-channel/all/conanfile.py
```diff
@@ -0,0 +1,59 @@
...
+            src=os.path.join(self.source_folder, "include"),
+        )
+        copy(
+            self,
+            pattern="*.inl",
+            dst=os.path.join(self.package_folder, "include"),
+            src=os.path.join(self.source_folder, "include"),
+        )
+
+    def package_info(self):
```
RubenRBS, line 57: seems like the library in upstream is called `msd_channel` https://github.com/andreiavrammsd/cpp-channel/blob/master/CMakeLists.txt#L14

I'll reach out to upstream to ask if they have plans to use targets and if so, what name they would expect to use

RubenRBS (commented):
hi! Thanks! I've a concern about the package name for cmake, but other than that it looks great :)

valgur (approved):

franramirez688 (approved):

PR #22993, opened 2024-03-05
[google-cloud-cpp] add v2.22.0 version

dbolduc:
Specify library name and version:  **google-cloud-cpp/2.22.0**

Update `google-cloud-cpp` to the latest version (v2.22.0). Remove the recipe for v2.12.0, and clean up recipe logic only relevant to v2.12.0.

Tested with the following on linux:

```cc
env CC=gcc-11 CXX=g++-11 conan create --build missing  --build-require  --version 2.22.0 recipes/google-cloud-cpp/2.x

env CC=gcc-11 CXX=g++-11 conan create --build missing  --build-require  --version 2.19.0 recipes/google-cloud-cpp/2.x

env CC=gcc-11 CXX=g++-11 conan create --build missing  --build-require  --version 2.15.1 recipes/google-cloud-cpp/2.x
```

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #23149, opened 2024-03-17
[google-cloud-cpp] skip testing with `os=macOS`, `shared=True` on conan v1

dbolduc:
Following the lead of `grpc`, skip testing with `os=macOS`, `shared=True` on conan v1.

---

In #22993 we attempted to upgrade `google-cloud-cpp` to `2.22.0`.

We see failures[^1] to build on `os=macOS` with `shared=True` for older versions of the library. This makes us suspect that the recipe will not build at HEAD.

This PR tests the recipe at HEAD by adding a no-op comment. If the builds fail, I will use this PR to address the build failures. We failed to reproduce this on our local machines, hence the PR.

[^1]: The failures have to do with not finding `libprotoc` while running the grpc protoc plugin: `dyld[48161]: Library not loaded: @rpath/libprotoc.32.dylib`

### DIFF
--- a/recipes/google-cloud-cpp/2.x/conanfile.py
+++ b/recipes/google-cloud-cpp/2.x/conanfile.py
@@ -9,6 +9,9 @@ from conan.tools.microsoft import check_min_vs, is_msvc
 from conan.tools.scm import Version
 from conan.errors import ConanInvalidConfiguration
 
+# This PR tests whether google-cloud-cpp builds at HEAD. This test is motivated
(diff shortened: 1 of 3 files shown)


dbolduc:
Is there a better way to handle `DYLD_LIBRARY_PATH` in the `os=macOS` + `shared=True` configuration?

A workaround is discussed in #10668 and implemented for some recipes. Is that solution preferred to disabling the configuration on conan v1?

dbolduc:
Closing, as this fixed nothing, and #23267 looks promising.

PR #23150, opened 2024-03-18
Updates libsndfile to use conan-provided libsndio

spiderkeys:
This PR depends on and follows https://github.com/conan-io/conan-center-index/pull/23087, solving issue https://github.com/conan-io/conan/issues/15240.


recipes/libsndfile/all/conanfile.py
```diff
@@ -59,7 +59,9 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
+        self.requires("libsndio/1.9.0")
         if self.options.get_safe("with_alsa"):
+            self.options["libsndio"].with_alsa = True
```
danimtb, line 64: It seems that this option is defaulted to True in libsndio recipe, to it should be true by default in this recipe as well. Otherwise, there will be no prebuilt packages to libsndio that have with_alsa=False option and the CI will fail (it builds default options only and does not build missing binaries from dependencies).

I am taking a look and will commit some changes to make the logic work in this recipe by setting the with_alsa=True by default. Hope that makes sense. Thank you!

danimtb:
The latest changes work fine as well when the option is set to false: `with_alsa=False`
So this should be fine 😄

AbrilRBS (approved):

danimtb (approved):

AbrilRBS (approved):

Issue #23094, opened 2024-03-14
[package] grpc/all: configuration is designed to possibly fail, but succeed on a retry


js-nano:
### Description

It appears that the configure step of gRPC is designed such that it can fail once but succeed on a retry.  This should be accounted for in the recipe.

Looking at [the section of gRPC's `CMakeFiles.txt` that downloads 3rd party dependencies](https://github.com/grpc/grpc/blob/868412b573a0663c8db41558498caf44098f4390/CMakeLists.txt#L376), we see it tries to download from one URL and, if that fails, it then tries to download from another URL:

```CMake
# Setup external proto library at third_party/opencensus-proto/src with 2 download URLs
if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src)
  # Download the archive via HTTP, validate the checksum, and extract to third_party/opencensus-proto/src.
  download_archive(
    ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src
    https://storage.googleapis.com/grpc-bazel-mirror/github.com/census-instrumentation/opencensus-proto/archive/v0.3.0.tar.gz
    b7e13f0b4259e80c3070b583c2f39e53153085a6918718b1c710caf7037572b0
    opencensus-proto-0.3.0/src
  )
endif()
if (NOT EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src)
  # Download the archive via HTTP, validate the checksum, and extract to third_party/opencensus-proto/src.
  download_archive(
    ${CMAKE_CURRENT_SOURCE_DIR}/third_party/opencensus-proto/src
    https://github.com/census-instrumentation/opencensus-proto/archive/v0.3.0.tar.gz
    b7e13f0b4259e80c3070b583c2f39e53153085a6918718b1c710caf7037572b0
    opencensus-proto-0.3.0/src
  )
endif()
```

The idea seems to be:
- Run configure
- Try to download from first URL
- If that fails, fall back to downloading from second URL
- Configure step fails (due to failed download from first URL)
- Re-run configure
- CMake detects the file with the correct hash (that was successfully downloaded from the second URL), and so no further download attempts are made
- Configure step succeeds

[...]

PR #23153, opened 2024-03-18
grpc: set gRPC_DOWNLOAD_ARCHIVES to false

js-nano:
The gRPC CMake configuration step can fail to due to a failure when downloading third party archives.  These failures are expected, and the user is instructed to retry.  So, if configuration fails, retry it

Also set the `gRPC_DOWNLOAD_ARCHIVES` CMake flag to `False`, to avoid downloading these unnecessary archives in the first place, in versions of gRPC that understand the flag

Specify library name and version:  **grpc/all**

This fixes https://github.com/conan-io/conan-center-index/issues/23094

---

### DIFF
--- a/recipes/grpc/all/conanfile.py
+++ b/recipes/grpc/all/conanfile.py
@@ -2,7 +2,7 @@ import os
 import yaml
 
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
+from conan.errors import ConanInvalidConfiguration, ConanException
 from conan.tools.apple import is_apple_os
 from conan.tools.build import cross_building, valid_min_cppstd, check_min_cppstd
 from conan.tools.cmake import cmake_layout, CMake, CMakeToolchain, CMakeDeps
@@ -190,6 +190,11 @@ class GrpcConan(ConanFile):
         tc.cache_variables["gRPC_BUILD_GRPC_RUBY_PLUGIN"] = self.options.ruby_plugin
         tc.cache_variables["gRPC_BUILD_GRPCPP_OTEL_PLUGIN"] = self.options.get_safe("otel_plugin", False)
 
+        # Never download unnecessary archives
+        # (supported in gRPC >= 1.62.0)
+        tc.cache_variables["gRPC_DOWNLOAD_ARCHIVES"] = False
+
+
         # Consumed targets (abseil) via interface target_compiler_feature can propagate newer standards
         if not valid_min_cppstd(self, self._cxxstd_required):
             tc.cache_variables["CMAKE_CXX_STANDARD"] = self._cxxstd_required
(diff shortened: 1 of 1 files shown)


valgur (approved):

recipes/grpc/all/conanfile.py
```diff
@@ -174,6 +174,11 @@ def generate(self):
         tc.cache_variables["gRPC_BUILD_GRPC_PYTHON_PLUGIN"] = self.options.python_plugin
         tc.cache_variables["gRPC_BUILD_GRPC_RUBY_PLUGIN"] = self.options.ruby_plugin
 
+        # Never download unnecessary archives
+        # (supported in gRPC >= 1.62.0)
+        tc.cache_variables["gRPC_DOWNLOAD_ARCHIVES"] = False
```
samuel-emrys, line 179: I'm not particularly familiar with `gRPC` - these archives _aren't_ required? I agree that a conan recipe should never attempt to download external files, but if they are required then they should be added to `conandata.yml` and downloaded/extracted in the `source()` method.

js-nano: Based on the issue where the flag was introduced, it seems that they're not required for anything directly related to building or testing gRPC
https://github.com/grpc/grpc/pull/34587#issuecomment-1779337256

In addition, for all but one of the archives, no attempt is made to download the archives -- the CMake script checks for the presence of directories that are (in most cases) already present in the Conan source checkout.

samuel-emrys:
@RubenRBS looks like the v2 pipeline is failing on a weird error - it can't find itself as a dependency?

samuel-emrys:
@js-nano from @RubenRBS [on slack](https://cpplang.slack.com/archives/C41CWV9HA/p1718800211571939?thread_ts=1718713845.604409&cid=C41CWV9HA), to get the conan 2 pipeline passing you'll need to simplify the `test_package` so that `self.tested_reference_str` is only requested as a `requires()`, and not a `tool_requires()`, i.e.

```diff
    def build_requirements(self):
        if not self._is_legacy_one_profile:
            # For the grpc-cpp-plugin executable at build time
-           self.tool_requires(self.tested_reference_str)
            self.tool_requires("protobuf/<host_version>")
```
This is due to a limitation in the existing CI infrastructure

jcar87 (approved):
LGTM - we've only kept the passing of `gRPC_DOWNLOAD_ARCHIVES` to False

A try/except pattern is not really acceptable  (and it really does complicate build logs) - I would strongly advise reporting this upstream if the behaviour is flaky (otherwise it never gets fixed!) - on the other hand, for older versions of grpc where this is still a problem, we would accept a PR that fully inhibits the behaviour and does not attempt to download anything that we know is not needed. But this appears to be addressed for grpc >=1.62.0

PR #22508, opened 2024-01-23
(#22506) arrow: fix config_options

wgtmac:
Specify library name and version:  **arrow/15.0.0**

Fix https://github.com/conan-io/conan-center-index/issues/22506

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #23163, opened 2024-03-18
Arrow: Refactor auto options

danimtb:
The aim of this PR is to solve recurrent issues with the options in the Arrow recipe. We have received many PRs trying to fix this situation but instead of making the recipe simpler, it has become complex this the management of "auto" options.

Making the recipe too smart is a way to make the recipe more likely to fail and in the end users may face bad experiences at cosume time.

Hope this helps to improve the situation!

Related to https://github.com/conan-io/conan-center-index/pull/22508#issuecomment-1992220451

---

### DIFF
--- a/recipes/arrow/all/conanfile.py
+++ b/recipes/arrow/all/conanfile.py
@@ -1,5 +1,5 @@
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
+from conan.errors import ConanInvalidConfiguration, ConanException
 from conan.tools.build import check_min_cppstd, cross_building
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
@@ -71,13 +71,13 @@ class ArrowConan(ConanFile):
         "shared": False,
         "fPIC": True,
         "gandiva": False,
-        "parquet": "auto",
+        "parquet": False,
         "skyhook": False,
         "substrait": False,
         "acero": False,
         "cli": False,
-        "compute": "auto",
-        "dataset_modules": "auto",
+        "compute": False,
+        "dataset_modules": False,
         "deprecated": True,
         "encryption": False,
         "filesystem_layer": False,
@@ -86,29 +86,29 @@ class ArrowConan(ConanFile):
         "simd_level": "default",
         "runtime_simd_level": "max",
         "with_backtrace": False,
-        "with_boost": "auto",
+        "with_boost": False,
         "with_brotli": False,
         "with_bz2": False,
         "with_csv": False,
         "with_cuda": False,
-        "with_flight_rpc": "auto",
+        "with_flight_rpc": False,
         "with_flight_sql": False,
         "with_gcs": False,
-        "with_gflags": "auto",
-        "with_jemalloc": "auto",
+        "with_gflags": False,
+        "with_jemalloc": False,
         "with_mimalloc": False,
-        "with_glog": "auto",
-        "with_grpc": "auto",
+        "with_glog": False,
+        "with_grpc": False,
         "with_json": False,
-        "with_thrift": "auto",
-        "with_llvm": "auto",
-        "with_openssl": "auto",
+        "with_thrift": False,
+        "with_llvm": False,
+        "with_openssl": False,
         "with_opentelemetry": False,
         "with_orc": False,
-        "with_protobuf": "auto",
-        "with_re2": "auto",
+        "with_protobuf": False,
+        "with_re2": False,
         "with_s3": False,
-        "with_utf8proc": "auto",
+        "with_utf8proc": False,
         "with_lz4": False,
         "with_snappy": False,
         "with_zlib": False,
@@ -158,22 +158,6 @@ class ArrowConan(ConanFile):
         if Version(self.version) < "8.0.0":
             del self.options.substrait
 
-        self.options.parquet = self._parquet()
-        self.options.compute = self._compute()
-        self.options.dataset_modules = self._dataset_modules()
-        self.options.with_boost = self._with_boost()
-        self.options.with_flight_rpc = self._with_flight_rpc()
-        self.options.with_gflags = self._with_gflags()
-        self.options.with_glog = self._with_glog()
-        self.options.with_grpc = self._with_grpc()
-        self.options.with_jemalloc = self._with_jemalloc()
-        self.options.with_thrift = self._with_thrift()
-        self.options.with_llvm = self._with_llvm()
-        self.options.with_openssl = self._with_openssl()
-        self.options.with_protobuf = self._with_protobuf()
-        self.options.with_re2 = self._with_re2()
-        self.options.with_utf8proc = self._with_utf8proc()
-
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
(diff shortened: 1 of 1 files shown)


valgur (changes requested):
This is definitely a much saner approach. Thanks!

recipes/arrow/all/conanfile.py
```diff
@@ -71,13 +71,13 @@ class ArrowConan(ConanFile):
         "shared": False,
         "fPIC": True,
         "gandiva": False,
-        "parquet": "auto",
+        "parquet": False,
```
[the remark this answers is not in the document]
danimtb: I see parquet is enabled in most "recipes" for other _package manages_ in their repo, so it make sense: https://github.com/search?q=repo%3Aapache%2Farrow%20ARROW_PARQUET&type=code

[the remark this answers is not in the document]
danimtb: Also, the `with_thrift` options seems to be hihgly dependant on the `parquet` one according to its cmake https://github.com/apache/arrow/blob/6c06b37434f865e52c69951ad9de40b36895be0d/cpp/cmake_modules/ThirdpartyToolchain.cmake#L390 so I will make it true by default

danimtb:
Waiting for changes at https://github.com/conan-io/conan-center-index/pull/23185

danimtb:
@valgur did not have much luck activating the parquet option although I think it makes sense. However, I just want to push the refactor of options management in this recipe, so let's leave the option value tweaks for a future PR

recipes/arrow/all/conanfile.py
```diff
@@ -181,115 +168,8 @@ def configure(self):
...
-        else:
-            return bool(self.options.with_openssl)
-
     def _requires_rapidjson(self):
-        if self.options.with_json:
-            return True
-        if Version(self.version) >= "7.0.0" and self.options.encryption:
-            return True
-        return False
+       return self.options.with_json or (Version(self.version) >= "7.0.0" and self.options.encryption)
```
valgur, line 172:
```suggestion
        return self.options.with_json or (Version(self.version) >= "7.0.0" and self.options.encryption)
```
Indentation

recipes/arrow/all/conanfile.py
```diff
@@ -348,11 +228,31 @@ def requirements(self):
...
 
     def validate(self):
-        # validate options with 'auto' as default value
-        auto_options = ["parquet", "compute", "dataset_modules", "with_boost", "with_flight_rpc", "with_gflags", "with_glog",
-                        "with_grpc", "with_jemalloc", "with_thrift", "with_llvm", "with_openssl", "with_protobuf", "with_re2", "with_utf8proc"]
-        for option in auto_options:
-            assert "auto" not in str(self.options.get_safe(option)), f"Option '{option}' contains 'auto' value, wich is not allowed. Generally the final value should be True/False"
+        # FIXME: Temporaly discard these configurations
+        if self.version == "1.0.0" and self.settings.compiler == "apple-clang":
+            raise ConanInvalidConfiguration("Not supported")
```
valgur, line 233:
```suggestion
```

valgur (approved):
LGTM. The default parquet value can be addressed later, I agree.

ericLemanissier (changes requested):
please consider https://github.com/danimtb/conan-center-index/pull/6

recipes/arrow/all/conanfile.py
```diff
@@ -348,11 +215,28 @@ def requirements(self):
...
 
     def validate(self):
-        # validate options with 'auto' as default value
-        auto_options = ["parquet", "compute", "dataset_modules", "with_boost", "with_flight_rpc", "with_gflags", "with_glog",
-                        "with_grpc", "with_jemalloc", "with_thrift", "with_llvm", "with_openssl", "with_protobuf", "with_re2", "with_utf8proc"]
-        for option in auto_options:
-            assert "auto" not in str(self.options.get_safe(option)), f"Option '{option}' contains 'auto' value, wich is not allowed. Generally the final value should be True/False"
+        # FIXME: Temporaly discard these configurations
+        if self.settings.compiler == "apple-clang":
+            raise ConanInvalidConfiguration("Not supported")
```
uilianries, line 220:
```suggestion
        # FIXME: Temporaly discard these configurations
        if self.settings.compiler == "apple-clang":
            raise ConanInvalidConfiguration("Not supported")
```

It's not clear why this limitation now. Please, update it in the PR description if possible, and update the recipe with a better description.

danimtb: you are right, I added this to test some changes. I will remove this. Thanks! 🙏

danimtb:
```suggestion
```

ericLemanissier (approved):

danimtb:
I have opened an issue in the arrow project to ask for some collaboration https://github.com/apache/arrow/issues/40979

kou:
I'm an Apache Arrow developer. What should I answer to complete this PR?

danimtb:
Hi @kou and thanks again for your response.

As you can see in the changes of the PR, we are trying to simplify the management of "auto" options (CMake variables) in the Arrow recipe. As some option values affected other options the UX was quite ricky for the user consuming this recipe and the maintenance of the recipe was harder at different points.

We reviewed the logic and finally reduced most of the option default values to `False`, however, there are some of them that were quite dependant on their "auto" behavior to other ones and we added some warnings in the recipe:

```
        # Warn/error on some option values
        if not self.options.with_boost:
            if self.options.gandiva:
                self.output.warning("Option 'with_boost=True' could be required when option 'gandiva=True'")
        if not self.options.with_re2:
            if self.options.gandiva or self.options.parquet:
                self.output.warning("Option 'with_re2=True' could be required when option 'gandiva=True' or 'parquet=True'")
            if self.options.compute or self.options.dataset_modules:
                self.output.warning("Option 'with_re2=True' could be required when 'compute=True' or 'dataset_modules=True'")
        if not self.options.with_jemalloc:
            if "BSD" in self.settings.os:
                self.output.warning("Option 'with_jemalloc=True' could be required for BSD in settings.os")
```

We would like to simplify this more but we are not familiar with the intrinsics of the Arrow build, so basically we would need to answer these questions:
- Should `with_jemalloc` (CMake's `ARROW_JEMALLOC`) option be True in BSD platforms? Are there any other platforms where jemalloc should be used mandatory?
- Should Boost be required whenever the `gandiva` (`ARROW_GANDIVA`) option is activated?
- Should re2 dependency be required whenever `gandiva`, `parquet` (`ARROW_PARQUET`), `compute` (`ARROW_COMPUTE`) or `dataset_modules` (`ARROW_DATASET`) are used?

Also, we noticed that this recipe was somehow used in the https://github.com/apache/arrow repository, so we want to know if there would be any issue with these changes.

Thanks a lot! 😄

kou:
>  * Should `with_jemalloc` (CMake's `ARROW_JEMALLOC`) option be True in BSD platforms? Are there any other platforms where jemalloc should be used mandatory?

We don't need to use `True` for `with_jemalloc` on BSD platforms. There are no platforms that require jemalloc.

(In general, jemalloc is better than system allocator. So we recommend jemalloc but it's not required.)
 
>  * Should Boost be required whenever the `gandiva` (`ARROW_GANDIVA`) option is activated?

Yes.

> * Should re2 dependency be required whenever `gandiva`, `parquet` (`ARROW_PARQUET`), `compute` (`ARROW_COMPUTE`) or `dataset_modules` (`ARROW_DATASET`) are used?

No. re2 is required only for `gandiva`.

> Also, we noticed that this recipe was somehow used in the https://github.com/apache/arrow repository, so we want to know if there would be any issue with these changes.

We use this recipe only for CI. We're testing this recipe with unreleased Apache Arrow. It's for smooth release process. We want to avoid that downstreams including this recipe have patches that fixes problems in released Apache Arrow. If we can detect problems before release a new version, downstreams don't need to have patches. :-)

But maintaining this recipe (keeping this recipe buildable) in apache/arrow difficult... It's broken recently...

ericLemanissier (approved):

RubenRBS (approved):

recipes/arrow/all/conanfile.py
```diff
@@ -332,8 +200,7 @@ def requirements(self):
             self.requires("lz4/1.9.4")
         if self.options.with_snappy:
             self.requires("snappy/1.1.9")
-        if Version(self.version) >= "6.0.0" and \
-            self.options.get_safe("simd_level") != None or \
+        if self.options.get_safe("simd_level") != None or \
```
valgur, line 203:
```suggestion
        if self.options.get_safe("simd_level") is not None or \
```

recipes/arrow/all/conanfile.py
```diff
@@ -332,8 +200,7 @@ def requirements(self):
...
             self.options.get_safe("runtime_simd_level") != None:
```
valgur, line 204:
```suggestion
            self.options.get_safe("runtime_simd_level") is not None:
```

recipes/arrow/all/conanfile.py
```diff
@@ -613,9 +485,8 @@ def package_info(self):
                 self.cpp_info.components["libgandiva"].requires.append("boost::boost")
             if self.options.parquet and self.settings.compiler == "gcc" and self.settings.compiler.version < Version("4.9"):
                 self.cpp_info.components["libparquet"].requires.append("boost::boost")
-            if Version(self.version) >= "2.0":
-                # FIXME: only headers components is used
-                self.cpp_info.components["libarrow"].requires.append("boost::boost")
+            # FIXME: only headers components is used
+            self.cpp_info.components["libarrow"].requires.append("boost::boost")
```
valgur, line 489:
```suggestion
            self.cpp_info.components["libarrow"].requires.append("boost::headers")
```

wgtmac:
Happy to see this gets resolved. Thanks!

BTW, I port this to the Apache Arrow repo and let's see how it goes: https://github.com/apache/arrow/pull/39729

wgtmac:
I have reported an issue with the new recipe: https://github.com/conan-io/conan-center-index/issues/23475. Could you please take a look? @danimtb

PR #23183, opened 2024-03-20
[changelog] Deployment March 20, 2023

uilianries:
A new deployment arrives with the beginning of the spring (north hemisphere).

- When opening PR related to change regular version to version range, it will be classified as bump dependencies PR too. In past the rule was the opposite and very strict because the CI bot was merging PRs automatically without reviews.

- When a recipe uses package type `static-library`, it will be built with both `*/*:shared=True/False`. This behavior will avoid missing package when using these packages as dependency.

---

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,9 +1,14 @@
 # Changelog
 
-### 13-Mar-2024 - 11:08 CET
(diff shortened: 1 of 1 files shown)


danimtb (approved):
Nice!

PR #23191, opened 2024-03-21
pkgconf: set tools.gnu:pkg_config

ericLemanissier:
It is used by conan's PkgConfig helper
https://github.com/conan-io/conan/blob/59029095e86eacd402083fca7c308b06889f6301/conan/tools/gnu/pkgconfig.py#L25

Specify library name and version:  **pkgconf/***

---

### DIFF
--- a/recipes/pkgconf/all/conanfile.py
+++ b/recipes/pkgconf/all/conanfile.py
@@ -138,8 +138,8 @@ class PkgConfConan(ConanFile):
         self.buildenv_info.prepend_path("ACLOCAL_PATH", pkgconf_aclocal)
         # TODO: evaluate if `ACLOCAL_PATH` is enough and we can stop using `AUTOMAKE_CONAN_INCLUDES`
         self.buildenv_info.prepend_path("AUTOMAKE_CONAN_INCLUDES", pkgconf_aclocal)
-        
-        self.conf_info.define("tools.gnu:pkg_config", "pkgconf")
+
+        self.conf_info.define("tools.gnu:pkg_config", pkg_config)
 
         # TODO: remove in conanv2
         automake_extra_includes = unix_path_package_info_legacy(self, pkgconf_aclocal.replace("\\", "/"))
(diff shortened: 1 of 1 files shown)


ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8372667985/attempts/1)

ericLemanissier:
this fails because macos-m1-clang runner runs after macos-clang

jwillikers:
@ericLemanissier Thanks for this!

jwillikers (approved):

ericLemanissier:
@jcar87 @RubenRBS @uilianries 
Is it possible to schedule macos native compilation before cross compilation ?
In the current situation no package tested as tool_requires can pass on macos.

uilianries:
> @jcar87 @RubenRBS @uilianries 
> Is it possible to schedule macos native compilation before cross compilation ?
> In the current situation no package tested as tool_requires can pass on macos.

@ericLemanissier Thank you for reporting the case. Indeed is a bug that should be fixed. We will be working in a patch to the next maintenance window. Sorry for the delay with this error, it's not the first PR that we have in this situation, usually, restarting the PR works, but is not something sustentable.

ericLemanissier:
No matter how hard I try, the CI always runthe macos cross build before the macos native build, so it systematically fails. It looks to me that the problem is not that the two configurations are not sequenced, but that they are sequenced backwards.

jwillikers:
I believe that this change will also allow the `*/system` packages to have dependency on `pkgconf` as necessary.

@ericLemanissier Won't this introduce a kind of race condition since most packages check this variable before introducing the `pkgconf` package as a `tool_requires`? 

```
if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
    self.tool_requires("pkgconf/2.1.0")
```

Whatever package get's evaluated first would then set the version of `pkgconf` for all other packages then right? I know there have been specific instances of build failures due to bumping the `pkgconf`, and while this may not effect CCI, it could effect anyone building multiple dependencies at once.

ericLemanissier:
I did not consider this order of evaluation issue indeed.
This is probably not the correct solution then. I can see alternatives:
- patch https://github.com/conan-io/conan/blob/59029095e86eacd402083fca7c308b06889f6301/conan/tools/gnu/pkgconfig.py#L25 to honor PKG_CONFIG env var (like meson and autotools do
- patch pkgconf recipe to create a [`pkg-config` symlink to pkgconf](https://github.com/pkgconf/pkgconf?tab=readme-ov-file#pkg-config-symlink). Does it work on windows though ?

regarding `*/system` packages, I do not think that it would actually allow to use pkgconf, because pkg-config/pkgconf still needs to know where the system .pc files are stored, and only the executable provided by the distribution has this information (`pkg-config --variable pc_path pkg-config`)

jwillikers:
> regarding `*/system` packages, I do not think that it would actually allow to use pkgconf, because pkg-config/pkgconf still needs to know where the system .pc files are stored, and only the executable provided by the distribution has this information (`pkg-config --variable pc_path pkg-config`)

Good point!

PR #23199, opened 2024-03-21
openassetio: migrate to Conan v2

valgur:
The recipe is almost up to date but it can be fully migrated now that the cpython package is available.


recipes/openassetio/all/test_v1_package/conanfile.py
```diff
@@ -8,18 +8,16 @@
...
 class TestPackageConan(ConanFile):
     settings = "os", "arch", "compiler", "build_type"
-    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv"
+    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv", "VirtualBuildEnv"
     test_type = "explicit"
 
     def build_requirements(self):
-        self.tool_requires("cmake/3.25.3")
+        self.tool_requires("cmake/[>=3.25 <4]")
+        self.tool_requires("cpython/3.10.0")
```
AbrilRBS, line 16: Why is this necessary even if no python option?

valgur: Probably an oversight on my part. I'll simply drop the test_v1_package.

PR #23206, opened 2024-03-21
ipaddress: add recipe

VladimirShaleev:
Specify library name and version:  **ipaddress/1.0.1**

Hi! This is a new library for working with IP addresses and networks.

It is inspired by the well-thought-out API from the Python module for working with addresses and networks. Therefore, it is familiar and understandable for use by many programmers.

This library fills gaps in the implementation of other libraries: network parsing, working with stringize **scope id**, as well as network manipulation

---

### DIFF
--- a/recipes/ipaddress/all/conanfile.py
+++ b/recipes/ipaddress/all/conanfile.py
@@ -4,7 +4,6 @@ from conan.tools.build import check_min_cppstd
 from conan.tools.files import copy, get
 from conan.tools.layout import basic_layout
 from conan.tools.scm import Version
-from conan.tools.microsoft import is_msvc
 import os
 
 
@@ -22,8 +21,32 @@ class IpAddressConan(ConanFile):
     settings = "os", "arch", "compiler", "build_type"
     no_copy_source = True
 
-    options = {"exceptions": [True, False], "overload_std": [True, False], "ipv6_scope": [True, False], "ipv6_scope_max_length": ["ANY"]}
-    default_options = {"exceptions": True, "overload_std": True, "ipv6_scope": True, "ipv6_scope_max_length": 16}
+    options = {
+        "exceptions": [True, False],
+        "overload_std": [True, False],
+        "ipv6_scope": [True, False],
+        "ipv6_scope_max_length": ["ANY"],
+    }
+    default_options = {
+        "exceptions": True,
+        "overload_std": True,
+        "ipv6_scope": True,
+        "ipv6_scope_max_length": 16,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 11
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "13.0",
+            "clang": "6.0",
+            "gcc": "7.5",
+            "msvc": "192",
+            "Visual Studio": "16",
+        }
 
     def layout(self):
         basic_layout(self, src_folder="src")
@@ -32,19 +55,13 @@ class IpAddressConan(ConanFile):
         self.info.clear()
 
     def validate(self):
-        compiler = self.settings.compiler
-        compiler_version = Version(self.settings.compiler.version)
-
-        if compiler == "gcc" and compiler_version < "7.5":
-            raise ConanInvalidConfiguration(f"{self.ref} doesn't support gcc < 7.5")
-        if compiler == "clang" and compiler_version < "6.0":
-            raise ConanInvalidConfiguration(f"{self.ref} doesn't support clang < 6.0")
-        if compiler == "apple-clang" and compiler_version < "13.0":
-            raise ConanInvalidConfiguration(f"{self.ref} doesn't support Apple clang < 13.0")
-        if is_msvc(self) and compiler_version < "16":
-            raise ConanInvalidConfiguration(f"{self.ref} doesn't support MSVC < 16")
         if self.settings.compiler.get_safe("cppstd"):
-            check_min_cppstd(self, 11)
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} doesn't support {self.settings.compiler} < {minimum_version}"
+            )
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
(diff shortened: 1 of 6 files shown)


RubenRBS:
Feel free to ping me once this is ready for review :) The recipe looks good as-is tho!

VladimirShaleev:
Hi! Thanks for the comment. I think PR is ready for review, I’ll change it to ready

recipes/ipaddress/all/conanfile.py
```diff
@@ -0,0 +1,72 @@
...
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/VladimirShaleev/ipaddress"
+    topics = ("ipv4", "ipv6", "ipaddress", "ip", "network", "header-only")
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
+    no_copy_source = True
+
+    options = {"exceptions": [True, False], "overload_std": [True, False], "ipv6_scope": [True, False], "ipv6_scope_max_length": ["ANY"]}
+    default_options = {"exceptions": True, "overload_std": True, "ipv6_scope": True, "ipv6_scope_max_length": 16}
```
valgur, line 26:
```suggestion
    options = {
        "exceptions": [True, False],
        "overload_std": [True, False],
        "ipv6_scope": [True, False],
        "ipv6_scope_max_length": ["ANY"],
    }
    default_options = {
        "exceptions": True,
        "overload_std": True,
        "ipv6_scope": True,
        "ipv6_scope_max_length": 16,
    }
```
For readability.

recipes/ipaddress/all/conanfile.py
```diff
@@ -0,0 +1,72 @@
...
+        compiler_version = Version(self.settings.compiler.version)
+
+        if compiler == "gcc" and compiler_version < "7.5":
+            raise ConanInvalidConfiguration(f"{self.ref} doesn't support gcc < 7.5")
+        if compiler == "clang" and compiler_version < "6.0":
+            raise ConanInvalidConfiguration(f"{self.ref} doesn't support clang < 6.0")
+        if compiler == "apple-clang" and compiler_version < "13.0":
+            raise ConanInvalidConfiguration(f"{self.ref} doesn't support Apple clang < 13.0")
+        if is_msvc(self) and compiler_version < "16":
+            raise ConanInvalidConfiguration(f"{self.ref} doesn't support MSVC < 16")
```
valgur, line 45: Not wrong, but it would be better to follow the conventional format:
https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/header_only/all/conanfile.py#L30-L75

valgur (approved):
Looks good. Thanks! Just some cosmetic suggestions.

recipes/ipaddress/all/conanfile.py
VladimirShaleev: Yes, it will probably be better. I'll do it as in the template. But let me change the error text to more accurately convey the cause of the error.

If the new text is unacceptable, please respond to me. I will return the text from template.

VladimirShaleev:
@valgur Thank you! Yes, these will be really good changes. Thanks for the feedback!

Thank you everyone for your help and work on my PR! Let me know if any other changes are needed

recipes/ipaddress/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+project(test_package LANGUAGES CXX)
+
+find_package(ipaddress CONFIG REQUIRED)
+
+add_executable(${PROJECT_NAME} test_package.cpp)
+target_link_libraries(${PROJECT_NAME} ipaddress::ipaddress)
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_14)
```
RubenRBS, line 8: This is setting C++ 14, but the recipe indicates min to be C++ 11 (And the readme does too)
```suggestion
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
```

To ensure we work for the min version :)

Unless there's a reason this is set to 14, any insight would be appreciated in that case!

RubenRBS (commented):
Hi! Thanks a lot for taking the time to add the recipe and your patience while we were on Easter break

The recipe looks good overall, I just have a minor question with the test package :)

recipes/ipaddress/all/test_package/CMakeLists.txt
VladimirShaleev: Hi, Rubén! I almost understand what needs to be done, but I want to clarify just in case:

```cmake
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
```

does not install compilation with C++11. It only checks that the compiler supports C++11 and higher. In reality, the copying will be with the default version (usually 14 or 17).

In order to configure compilation with standard C++11, need to do the following:

```cmake
set_target_properties(${PROJECT_NAME}
     PROPERTIES
         CXX_STANDARD 11
)
```

or

```cmake
set_target_properties(${PROJECT_NAME}
     PROPERTIES
         CXX_STANDARD 11
         CXX_STANDARD_REQUIRED ON
         CXX_EXTENSIONS OFF
)
```

In this case, cmake will pass the necessary flag during compilation (`-std=gnu++11` or `-std=c++11` etc).

Do I need to set a minimum standard or force the test to compile with C++11? In other words, should I use `target_compile_features` or `set_target_properties`?

uilianries: @VladimirShaleev Thank you for your feedback. The reason behind is because in Conan 1.x , the settings.compiler.cppstd is not configured by default, so the test package will use what is configured by the compiler as standard version (e.g. GCC5 will fail). In Conan 2.x, the c++ standard always follows what is configured by settings.compiler.cppstd, so it would not be needed to use `target_compile_features` neither `set_target_properties` in that case (future, when dropping Conan 1.x). But answering your question: `target_compile_features` is more than fine because it will check if the compiler supports c++11 and accommodate in case needed.

VladimirShaleev: @uilianries Thanks for your reply! I'll follow your advice and save target_compile_features with cxx_std_11

RubenRBS (approved):
Thanks a lot! This looks super good now :)

VladimirShaleev:
Thank you, Rubén, for your insightful feedbacks during the pull request review) I really appreciate it!

toge (approved):

PR #23216, opened 2024-03-22
libcvd: new recipe

### LINKED ISSUE #23173: [request] libCVD
### Package Name/Version

libcvd/2.3.1

### Webpage

https://github.com/edrosten/libcvd

### Source code

https://github.com/edrosten/libcvd

libcvd - efficient and easy to use C++ computer vision library.

valgur:
Adds libCVD: https://github.com/edrosten/libcvd

libcvd - efficient and easy to use C++ computer vision library.

Resolves #23173

[Packaging status](https://repology.org/project/libcvd/versions)


recipes/libcvd/all/conanfile.py
```diff
@@ -0,0 +1,180 @@
...
+            self.requires("libjpeg/9e")
+        elif self.options.with_libjpeg == "mozjpeg":
+            self.requires("mozjpeg/4.1.5")
+        if self.options.with_libpng:
+            self.requires("libpng/[>=1.6 <2]")
+        if self.options.with_libtiff:
+            self.requires("libtiff/4.6.0")
+        if self.options.get_safe("with_opengl", True):
+            # https://github.com/edrosten/libcvd/blob/RELEASE_2_5_0/cvd/videodisplay.h#L18-L20
+            self.requires("opengl/system", transitive_headers=True, transitive_libs=True)
```
jwillikers, line 96: Would it be possible to target `libglvnd` here on `Linux` and `FreeBSD` instead of `opengl/system`?

jwillikers (changes requested):

recipes/libcvd/all/conanfile.py
valgur: That sounds good to me. My only general doubt is whether it could conflict with other packages still using `opengl/system`?
`libglvnd` does not set `provides = "opengl"`, but doesn't it effectively conflict with `opengl/system` nevertheless?

jwillikers: You're totally correct about that, and I think we need to add that to the `libglvnd` recipe. However, I'd prefer to have less recipes to update in the future to use `libglvnd` on Linux and FreeBSD.

jwillikers (approved):

RubenRBS (approved):
Thanks! (And to @jwillikers too for the opengl review, I'd have not catch that!)

PR #23218, opened 2024-03-22
bump sdl_image dependencies

StellaSmith:


valgur (approved):

recipes/sdl_image/all/conanfile.py
```diff
@@ -96,17 +96,17 @@ def layout(self):
 
     def requirements(self):
         # Headers are exposed https://github.com/conan-io/conan-center-index/pull/16167#issuecomment-1508347351
-        self.requires("sdl/2.28.3", transitive_headers=True)
+        self.requires("sdl/2.28.5", transitive_headers=True)
```
memsharded, line 99: Maybe it would be better if we could use the latest sdl on ConanCenter, which is the 2.30.1? So the latest of ``sdl_image`` doesn't conflict with the latest of ``sdl``?

recipes/sdl_image/all/conanfile.py
```diff
@@ -96,17 +96,17 @@ def layout(self):
...
         if self.options.with_libtiff:
             self.requires("libtiff/4.6.0")
         if self.options.with_libjpeg:
             self.requires("libjpeg/9e")
         if self.options.with_libpng:
-            self.requires("libpng/1.6.40")
+            self.requires("libpng/1.6.43")
```
memsharded, line 105:
```suggestion
            self.requires("libpng/[>=1.6 <2]")
```
libpng is already accepting version ranges, this would be better for the future

recipes/sdl_image/all/conanfile.py
```diff
@@ -96,17 +96,17 @@ def layout(self):
...
         if self.options.with_libwebp:
             self.requires("libwebp/1.3.2")
         if self.options.get_safe("with_avif"):
-            self.requires("libavif/1.0.1")
+            self.requires("libavif/1.0.4")
```
memsharded, line 109: This looks good

recipes/sdl_image/all/conanfile.py
memsharded: Well, responding to myself, and checking other recipes in ConanCenter, the other recipes in ConanCenter depends on 2.28, so this cannot be done and would require some coordination, please be patience, this might take some time. We are trying to improve the CI to be able to do these changes more streamlined, but this is a big effort.

PR #23222, opened 2024-03-23
libva: Require libglvnd instead of opengl/system

jwillikers:
The with_glx option of the libva package pulls in a dependency on OpenGL. This option is only available on Linux and FreeBSD, so instead of using `opengl/system`, use the `libglvnd` package instead which provides the OpenGL libraries. This avoids using a system dependency.

### DIFF
--- a/recipes/libva/all/conanfile.py
+++ b/recipes/libva/all/conanfile.py
@@ -67,11 +67,13 @@ class PackageConan(ConanFile):
         if self.options.get_safe("with_wayland"):
             self.requires("wayland/1.22.0")
         if self.options.get_safe("with_glx"):
-            self.requires("opengl/system")
+            self.requires("libglvnd/1.7.0", options={"glx": True})
 
     def validate(self):
         if self.options.get_safe("with_glx") and not self.options.get_safe("with_x11"):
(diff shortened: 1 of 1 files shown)


valgur:
Very interesting. Should `libglvnd` be preferred to `opengl/system` in general?

jwillikers:
> Very interesting. Should `libglvnd` be preferred to `opengl/system` in general?

Definitely on FreeBSD / Linux. It provides the OpenGL implementations in way that they are automatically dispatched to different vendor implementations. I think it should be the preferred default on FreeBSD and Linux. The Mesa project can also be built without `libglvnd` support, in which case, it would provide the OpenGL libraries. I'm still working on getting the Mesa package merged, bet it would make sense to be able to choose between these configurations.

valgur (approved):

jcar87:

> Very interesting. Should `libglvnd` be preferred to `opengl/system` in general?

No. Given that it interfaces with the system's X server and the drivers, it's best to leave this to the distro.
At build time it's not an issue (C interfaces, binary compatible, etc) - but at runtime it may just not work.
AFAIK: Conda, binary wheels in PyPi, vpckg, homebrew - do NOT get libglvnd from their layer, and expect it from the system.

PR #23223, opened 2024-03-23
premake: add 5.0.0-beta2, unvendor deps, simplify

valgur:
Although a pre-release, v5.0.0 is being actively used by the `yojimbo` recipe, for example, so it's worth an update, imo.

Also unvendored `libcurl`, `mbedtls`, `libzip` and `zlib` to ensure that the most up-to-date versions of these are used.

Applied all patches to the non-bootstrapped build system based on the original premake build files instead of the bootstrap Makefiles and vcxprojects, which keeps the recipe much simpler.


recipes/premake/5.x/conanfile.py
```diff
@@ -49,58 +46,26 @@ def package_id(self):
         del self.info.settings.compiler
 
     def requirements(self):
-        if self.settings.os != "Windows":
-            self.requires("util-linux-libuuid/2.39")
+        self.requires("libcurl/[>=7.78.0 <9]")
+        self.requires("libzip/1.10.1")
+        self.requires("zlib/1.3.1")
```
Croydon, line 51:
```suggestion
        self.requires("zlib/[>=1.2.13 <2]")
```

recipes/premake/5.x/conanfile.py
```diff
@@ -49,58 +46,26 @@ def package_id(self):
         del self.info.settings.compiler
 
     def requirements(self):
-        if self.settings.os != "Windows":
-            self.requires("util-linux-libuuid/2.39")
+        self.requires("libcurl/[>=7.78.0 <9]")
```
Croydon, line 49: Version ranges are currently only allowed for a handful of recipes. Is libcurl one of them?

recipes/premake/5.x/conandata.yml
```diff
@@ -1,7 +1,19 @@
 sources:
+  "5.0.0-beta2":
+    url: "https://github.com/premake/premake-core/archive/refs/tags/v5.0.0-beta2.tar.gz"
+    sha256: "84a7775fe0a90c0f649eb18b10779b84626ad8c58dea4a8f24cca83690dd47d4"
   "5.0.0-alpha15":
-    url: "https://github.com/premake/premake-core/releases/download/v5.0.0-alpha15/premake-5.0.0-alpha15-src.zip"
-    sha256: "880f56e7cb9f4945d1cb879f059189462c1b7bf62ef43ac7d25842dfb177dd53"
+    url: "https://github.com/premake/premake-core/archive/refs/tags/v5.0.0-alpha15.tar.gz"
```
Croydon, line 6: Please don't use the auto generated GitHub source archives when there are official downloads

Croydon (commented):
> Applied all patches to the non-bootstrapped build system based on the original premake build files instead of the bootstrap Makefiles and vcxprojects, which keeps the recipe much simpler.

I don't get it. What does the mean?

Is the way to build Premake that you choose official supported on all configurations?

> Although a pre-release, v5.0.0 is being actively used by the yojimbo recipe, for example, so it's worth an update, imo.

This is really no criteria for a project that gets actively developed but is so afraid of stable releases, that their last official stable release is from 2010... see e.g. https://github.com/premake/premake-core/issues/178

So, in the case of Premake it is completely okay to have alpha and beta releases added.

recipes/premake/5.x/conanfile.py
valgur: Yes.

recipes/premake/5.x/conandata.yml
valgur: They are more or less identical, except the packaged source distributions include the "bootstrapped" build files for a dozen or so build configurations. These are no longer used by the updated PR and can be skipped for a smaller download.

valgur:
@Croydon 
> I don't get it. What does the mean?

See https://github.com/premake/premake-core/blob/master/BUILD.txt#L27

The packaged source archives include build scripts for each different platform (Linux and macOS Makefiles, VS projects, etc) that have been "bootstrapped" from the [premake5.lua](https://github.com/premake/premake-core/blob/master/premake5.lua) Premake build script.

Since this PR rewrites the build scripts to rely on dependencies from Conan instead of the vendored versions, it's much more efficient to modify the input `premake5.lua` script instead of the generated build files. That's the only real reason for the change. The recipe becoming simpler is a nice bonus, too.

> Is the way to build Premake that you choose official supported on all configurations?

Yes. It's identical to what the project itself does for the source archives.

recipes/premake/5.x/conandata.yml
Croydon: GitHub does generate the source archives on the fly, they are not necessary stable. They broke them once recently and broke dozen of projects. And in no way are they official from a project point of view. So please use the official download

valgur: Eh, fine. Reverted.

Issue #16747, opened 2023-03-28
[request] libssh/0.10.3


grabowski-d:
### Package Name/Version

libssh/0.10.3

### Webpage

https://www.libssh.org/

### Source code

https://git.libssh.org/

### Description of the library/tool

Library description by authors:
libssh is a multiplatform C library implementing the SSHv2 protocol on client and server side. With libssh, you can remotely execute programs, transfer files, use a secure and transparent tunnel, manage public keys and much more.

I wish to create a recipe for this package by myself.
Any concerns?

gjasny:
The recipe needs Conan 2 conversion as well. I'll give it a try.

gjasny:
> I wish to create a recipe for this package by myself.

Oh sorry, I missed that. I'm under the impression that `[request]` issues are to request recipes from others. I picked up the stale and closed PR #13449 and updated it according to the cmake packaging template: #16899 .

grabowski-d:
Sure, go ahead. I didn't start yet :)

uilianries:
The libssh is available in Conan Center. See #23230

PR #23230, opened 2024-03-24
libssh: add recipe

obnyis:
Related: #13449

Related: #16899

Fixes: #16747

Specify library name and version:  **libssh/0.10.6**

Changed replace_in_file from previous PR with apply_conandata_patches

Added `tc.cache_variables["CMAKE_TRY_COMPILE_CONFIGURATION"]` to support Visual Studio / Ninja Multi-Config generators

---

### DIFF
--- a/recipes/libssh/all/conanfile.py
+++ b/recipes/libssh/all/conanfile.py
@@ -22,12 +22,14 @@ class LibSSHRecipe(ConanFile):
         "fPIC": [True, False],
         "with_zlib": [True, False],
         "crypto_backend": ["openssl", "gcrypt"],
+        "with_symbol_versioning": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "with_zlib": True,
         "crypto_backend": "openssl",
+        "with_symbol_versioning": True,
     }
 
     def export_sources(self):
@@ -36,6 +38,7 @@ class LibSSHRecipe(ConanFile):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+            del self.options.with_symbol_versioning
 
     def configure(self):
         if self.options.shared:
@@ -66,7 +69,7 @@ class LibSSHRecipe(ConanFile):
         tc.variables["WITH_GSSAPI"] = False
         tc.variables["WITH_MBEDTLS"] = False
         tc.variables["WITH_NACL"] = False
-        tc.variables["WITH_SYMBOL_VERSIONING"] = False
+        tc.variables["WITH_SYMBOL_VERSIONING"] = self.options.get_safe("with_symbol_versioning", True)
         tc.variables["WITH_ZLIB"] = self.options.with_zlib
         if is_msvc(self):
             tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
--- a/recipes/libssh/all/conanfile.py
+++ b/recipes/libssh/all/conanfile.py
@@ -116,8 +116,3 @@ class LibSSHRecipe(ConanFile):
         # pkg-config defined at https://git.libssh.org/projects/libssh.git/tree/CMakeLists.txt?h=libssh-0.10.6#n124
         self.cpp_info.set_property("pkg_config_name", "libssh")
 
-        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
-        self.cpp_info.filenames["cmake_find_package"] = "libssh"
-        self.cpp_info.filenames["cmake_find_package_multi"] = "libssh"
-        self.cpp_info.names["cmake_find_package"] = "libssh"
-        self.cpp_info.names["cmake_find_package_multi"] = "libssh"
--- a/recipes/libssh/all/conanfile.py
+++ b/recipes/libssh/all/conanfile.py
@@ -103,11 +103,17 @@ class LibSSHRecipe(ConanFile):
             self.cpp_info.system_libs.extend(["ws2_32", "iphlpapi"])
 
         self.cpp_info.set_property("cmake_file_name", "libssh")
+        # target and alias names defined at:
+        # ssh         https://git.libssh.org/projects/libssh.git/tree/src/CMakeLists.txt?h=libssh-0.10.6#n351
+        # ssh::ssh    https://git.libssh.org/projects/libssh.git/tree/src/CMakeLists.txt?h=libssh-0.10.6#n371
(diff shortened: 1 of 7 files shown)


recipes/libssh/all/conanfile.py
```diff
@@ -0,0 +1,110 @@
...
+        tc.variables["WITH_GCRYPT"] = self.options.crypto_backend == "gcrypt"
+        tc.variables["WITH_GSSAPI"] = False
+        tc.variables["WITH_MBEDTLS"] = False
+        tc.variables["WITH_NACL"] = False
+        tc.variables["WITH_SYMBOL_VERSIONING"] = False
+        tc.variables["WITH_ZLIB"] = self.options.with_zlib
+        if is_msvc(self):
+            tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
+        tc.cache_variables["CMAKE_TRY_COMPILE_CONFIGURATION"] = str(self.settings.build_type)
```
valgur, line 74: Nice! This indeed seems to be the correct workaround for this CMake/Conan issue:  https://github.com/conan-io/conan/issues/12180

recipes/libssh/all/test_v1_package/CMakeLists.txt
valgur, line 1: New recipes should avoid including `test_v1_package`. You can drop it entirely.

recipes/libssh/all/conanfile.py
```diff
@@ -0,0 +1,110 @@
...
+    def package_info(self):
+        self.cpp_info.libs = ["ssh"]
+        if not self.options.shared:
+            self.cpp_info.defines.append("LIBSSH_STATIC=ON")
+        if self.settings.os == "Windows":
+            self.cpp_info.system_libs.extend(["ws2_32", "iphlpapi"])
+
+        self.cpp_info.set_property("cmake_file_name", "libssh")
+        self.cpp_info.set_property("cmake_target_name", "ssh")
+        self.cpp_info.set_property("pkg_config_name", "libssh")
```
valgur, line 104:
```suggestion
        self.cpp_info.set_property("cmake_file_name", "libssh")
        self.cpp_info.set_property("cmake_target_name", "ssh::ssh")
        self.cpp_info.set_property("cmake_target_aliases", ["ssh"])
        self.cpp_info.set_property("pkg_config_name", "libssh")
```
The library exports both `ssh` and `ssh::ssh` as targets: https://git.libssh.org/projects/libssh.git/tree/src/CMakeLists.txt#n345

The rest are correct:
cmake config file: https://git.libssh.org/projects/libssh.git/tree/src/CMakeLists.txt#n387
.pc file: https://git.libssh.org/projects/libssh.git/tree/CMakeLists.txt#n102

You could also add `ssh-static` and `ssh::static` aliases if not self.options.shared: https://git.libssh.org/projects/libssh.git/tree/src/CMakeLists.txt#n407

valgur (changes requested):
LGTM, should only add a missing exported target alias. Thanks!

recipes/libssh/all/conanfile.py
valgur: Including one or more of the above links as a reference in a comment would not hurt either.

recipes/libssh/all/test_v1_package/CMakeLists.txt
obnyis: Neat, I've removed that now

valgur (approved):
Thanks!

recipes/libssh/all/conanfile.py
obnyis: I added the missing targets, but I'm not sure which lines in particular you want comments on. Could you make a suggested change with what you are after and I'll add it in?

obnyis: Hi @valgur, I added some comments with links to the lines these definitions are in the libssh-0.10.6 tag as comments. Can you review this and let me know if it's what you had in mind?

uilianries (commented):
Looking good, only few details

recipes/libssh/all/conanfile.py
```diff
@@ -0,0 +1,123 @@
...
+            ["ssh"] if self.options.shared else ["ssh", "ssh-static", "ssh::static"],
+        )
+        # pkg-config defined at https://git.libssh.org/projects/libssh.git/tree/CMakeLists.txt?h=libssh-0.10.6#n124
+        self.cpp_info.set_property("pkg_config_name", "libssh")
+
+        # TODO: to remove in conan v2 once cmake_find_package_* generators removed
+        self.cpp_info.filenames["cmake_find_package"] = "libssh"
+        self.cpp_info.filenames["cmake_find_package_multi"] = "libssh"
+        self.cpp_info.names["cmake_find_package"] = "libssh"
+        self.cpp_info.names["cmake_find_package_multi"] = "libssh"
```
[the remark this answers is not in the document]
obnyis: Done. Thanks for the review :)

recipes/libssh/all/conanfile.py
```diff
@@ -0,0 +1,118 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["CLIENT_TESTING"] = False
+        tc.variables["SERVER_TESTING"] = False
+        tc.variables["WITH_EXAMPLES"] = False
+        tc.variables["WITH_GCRYPT"] = self.options.crypto_backend == "gcrypt"
+        tc.variables["WITH_GSSAPI"] = False
+        tc.variables["WITH_MBEDTLS"] = False
```
uilianries, line 70: quick question here, did you try to use the mbedtls package here? https://conan.io/center/recipes/mbedtls?version=3.6.0

It's not a blocker, but I would recommend adding a TODO comment to improve it in the future.

uilianries, line 70:
```suggestion
        # TODO: Add support for mbedtls. Package available in Conan Center already.
        tc.variables["WITH_MBEDTLS"] = False
```

recipes/libssh/all/patches/0.10.6-0001-fix-cmake.patch
uilianries, line 1: I just checked the upstream and they are actually using CMake targets in the master branch. So it should not be needed in the future.

uilianries (approved):
LGTM

PR #23234, opened 2024-03-24
added opus 1.5.1

dmpriso:
Specify library name and version:  opus/1.5.1

Added new opus version with support for DRED and deep PLC

---


recipes/opus/all/conanfile.py
```diff
@@ -56,6 +62,9 @@ def generate(self):
         tc = CMakeToolchain(self)
         tc.variables["OPUS_FIXED_POINT"] = self.options.fixed_point
         tc.variables["OPUS_STACK_PROTECTOR"] = self.options.stack_protector
+        tc.variables["OPUS_OSCE"] = self.options.osce
+        tc.variables["OPUS_DEEP_PLC"] = self.options.deep_plc
+        tc.variables["OPUS_DRED"] = self.options.dred
```
valgur, line 67:
```suggestion
        tc.variables["OPUS_DRED"] = self.options.dred
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
Needed for `tc.variables` to set CMake `options()` correctly.

valgur (changes requested):
LGTM, only need to add CMP0077. Thanks!

recipes/opus/all/conanfile.py
Croydon: Could you please explain the need for this CMake policy @valgur ? 🤔

valgur: Without it any values set using `option()` in CMake do not get overridden by `tc.variables[...]`. `tc.variables` sets its values using `set()` commands loaded within the `project()` command. `tc.cache_variables` get set via `-D...` command line arguments. Only the latter work correctly when this policy is not enabled.

The policy gets set automatically when CMake's policy version is set to 3.12 or above, which is usually set via `cmake_minimum_required()`.

Tbh, this should be set by Conan automatically. I consider it a bug.

More info in CMake docs: https://cmake.org/cmake/help/latest/policy/CMP0077.html

Croydon:
@dmpriso You need to write a short comment in this issue, then you are going to be added and CI runs for your pull requests https://github.com/conan-io/conan-center-index/issues/4

dmpriso:
There seems to be a TLS issue when downloading the builds from xiph.org

dmpriso:
Anything I can do about this? The releases are hosted on xiph.org, but somehow the build servers don't accept the certificate or TLS version, see e.g. https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23234/4-linux-gcc/opus/1.5.1//1d2e365eb8c339034b0dc24d7dd442ac92ae783a-build.txt

Croydon:
Might have been a temporary thing since it only happened in a Conan v2 build. v1 had a regular build failure

recipes/opus/all/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "1.5.1":
+    url: "https://downloads.xiph.org/releases/opus/opus-1.5.1.tar.gz"
+    sha256: "b84610959b8d417b611aa12a22565e0a3732097c6389d19098d844543e340f85"
```
Croydon, line 4:
```suggestion
  "1.5.2":
    url:
      - "https://ftp.osuosl.org/pub/xiph/releases/opus/opus-1.5.2.tar.gz"
      - "https://downloads.xiph.org/releases/opus/opus-1.5.2.tar.gz"
    sha256: "65c1d2f78b9f2fb20082c38cbe47c951ad5839345876e46941612ee87f9a7ce1"
```

The downloads.xiph.org seems to redirect to mirrors. So let's try to use one of them as the preferred download.
For older versions we also have used their GitHub downloads, but they did not upload them there for 1.5x. And we also used a Mozilla mirror, but they don't seem to have these versions yet either.

Checksum is taken from https://opus-codec.org/release/stable/2024/04/12/libopus-1_5_2.html

And let's use 1.5.2 instead of 1.5.1 at this point

recipes/opus/config.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  "1.5.1":
```
Croydon, line 2:
```suggestion
  "1.5.2":
```

Croydon (changes requested):

Croydon:
For protocol, the error is

```
opus/1.5.1: ERROR: Error downloading file https://downloads.xiph.org/releases/opus/opus-1.5.1.tar.gz: 'HTTPSConnectionPool(host='downloads.xiph.org', port=443): Max retries exceeded with url: /releases/opus/opus-1.5.1.tar.gz (Caused by SSLError(SSLError(1, '[SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:1091)')))'
ERROR: opus/1.5.1: Error in source() method, line 59
	destination=self.source_folder, strip_root=True)
	ConanException: Error downloading file https://downloads.xiph.org/releases/opus/opus-1.5.1.tar.gz: 'HTTPSConnectionPool(host='downloads.xiph.org', port=443): Max retries exceeded with url: /releases/opus/opus-1.5.1.tar.gz (Caused by SSLError(SSLError(1, '[SSL: TLSV1_ALERT_PROTOCOL_VERSION] tlsv1 alert protocol version (_ssl.c:1091)')))'
```

downloads.xiph.org seems to support ONLY TLS 1,3, not any older versions.
ftp.osuosl.org seems to only support TLS 1.2, not any other version.

The Python 3.7 version in the docker containers is using the following OpenSSL version: `OpenSSL 1.0.2g  1 Mar 2016`
Printed via
```
import ssl
[...]
    def configure(self):
        self.output.info(ssl.OPENSSL_VERSION)
```

OpenSSL added TLS 1.3 support in 1.1.1+

recipes/opus/all/conandata.yml
danimtb: I think it will be better to use the gitlab official repo source code URL:

https://gitlab.xiph.org/xiph/opus/-/archive/v1.5.2/opus-v1.5.2.tar.gz

```suggestion
  "1.5.2":
    url: "https://gitlab.xiph.org/xiph/opus/-/archive/v1.5.2/opus-v1.5.2.tar.gz"
    sha256: "b84610959b8d417b611aa12a22565e0a3732097c6389d19098d844543e340f85"
```

recipes/opus/all/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "1.5.2":
+    url: "https://gitlab.xiph.org/xiph/opus/-/archive/v1.5.2/opus-v1.5.2.tar.gz"
+    sha256: "b84610959b8d417b611aa12a22565e0a3732097c6389d19098d844543e340f85"
```
danimtb, line 4:
```suggestion
    sha256: "5cf92b5b577d8ed203424f1e0f618f30bc6b6e42a26eae88bdb649ea63961cc9"
```

danimtb:
I think this PR is ready to merge even though the CLA Assistant complains about "Daniel" having not signed it (however, I belive it is the same person as @dmpriso).

danimtb (dismissed):

recipes/opus/all/conanfile.py
```diff
@@ -21,14 +22,24 @@ class OpusConan(ConanFile):
...
         "fixed_point": False,
         "stack_protector": True,
+        "osce": False,
+        "deep_plc": False,
+        "dred": False,
     }
 
+    def build_requirements(self):
+        if self.version == "1.5.2":
+            self.tool_requires("cmake/[>=3.16 <4]")
```
danimtb, line 41: this was required by the original cmakelist of the library

recipes/opus/all/conanfile.py
```diff
@@ -21,14 +22,24 @@ class OpusConan(ConanFile):
         "fPIC": [True, False],
         "fixed_point": [True, False],
         "stack_protector": [True, False],
+        "osce": [True, False],
+        "deep_plc": [True, False],
+        "dred": [True, False],
```
uilianries, line 27:
```suggestion
```

I would suggest removing these new options. 

When any of them is enabled, it will require DNN support as well:

https://gitlab.xiph.org/xiph/opus/-/compare/v1.4...v1.5.2?from_project_id=36&straight=false#9a2aa4db38d3115ed60da621e012c0efc0172aae_368_401

Which results in the follow error:

```
-- Configuring done (14.7s)
CMake Error at cmake/OpusFunctions.cmake:180 (target_sources):
  Cannot find source file:

    dnn/fargan_data.h

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm
  .ccm .cxxm .c++m .h .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90
  .f95 .f03 .hip .ispc
  CMakeLists.txt:400 (add_sources_group)
```

The `dnn/fargan_data.h` a generated file, resulting from the pre-training commands using Python, a pre-steps, before building: https://gitlab.xiph.org/xiph/opus/-/blob/v1.5.2/dnn/torch/fargan/README.md#L37 

As those options are OFF by default, even in the project (https://gitlab.xiph.org/xiph/opus/-/blob/v1.5.2/CMakeLists.txt#L86) we don't see it, but it's the scenario that will result in new issues for CCI in the future.

uilianries (approved):
LGTM

danimtb (approved):

Issue #23205, opened 2024-03-21
[sbepp] <sbepp>/<1.2.0>: Decrease minimum compiler version to gcc 8.5


ujos:
### Description

Because of restrictions on the project I'm working on, I cannot use gcc9. Instead we use gcc 8. When I run `connan install` command, I get the following error:

```
sbepp/1.2.0: Invalid ID: sbepp/1.2.0 requires C++17, which your compiler does not support.
```
I've checked the `conanfile.py` and found following code:

```py
   def _compilers_minimum_version(self):
        if self.options.with_sbeppc:
            return {
                "gcc": "9",
                "clang": "9",
                "apple-clang": "11"
            }
        else:
            return {
                "gcc": "4.8.1",
                "clang": "3.3",
                "apple-clang": "9.4"
            }
```

This code sets the minimum required gcc version to 9 in case if `with_sbeppc=True`.

I've downloaded the source code of the sbepp/1.2.0 library and successfully built it with gcc 8.5. So theoretically we can use the `sbepp` library as a git module, however I'd like to use it from the Conan.

Taking into account the information above, could you please update the minimum GCC version requirements to v8.5?

### Package and Environment Details

* Package Name/Version: **sbepp/1.2.0**
* Operating System+version: **CentOS Linux release 7.9.2009 (Core)**
* Compiler+version: **GCC 8.5**
* Docker image: **centos7.9.2009**
* Conan version: **conan 1.58.0**
* Python version: **Python 3.6.8**

### Conan profile

```
[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=8
compiler.libcxx=libstdc++11

[options]

[build_requires]

[buildenv]
CC=/home/dev/bin/GCC-8.5-abi/bin/gcc
CXX=/home/dev/bin/GCC-8.5-abi/bin/g++

```

### Steps to reproduce

1. Install gcc 8.5
```
yum update
yum install texinfo gcc gcc-c++ wget bzip2 make

wget https://ftp.gnu.org/gnu/gcc/gcc-8.5.0/gcc-8.5.0.tar.gz
tar xvf gcc-8.5.0.tar.gz
cd gcc-8.5.0
./contrib/download_prerequisites
cd ..
mkdir obj
../gcc-8.5.0/configure --prefix=$HOME/bin/GCC-8.5-abi --disable-multilib --enable-languages=c,c++ --disable-libsanitizer
make -j 8
```
[...]

PR #23245, opened 2024-03-25
+ [sbepp] Added support of gcc8.1 to sbeppc

ujos:
Specify library name and version:  **sbepp/1.2.0**

This commit extends the list of supported gcc compilers up to g++ v8.1

Following is a link to the ticket:  https://github.com/conan-io/conan-center-index/issues/23205

---

### DIFF
--- a/recipes/sbepp/all/conanfile.py
+++ b/recipes/sbepp/all/conanfile.py
@@ -37,7 +37,7 @@ class PackageConan(ConanFile):
     def _compilers_minimum_version(self):
         if self.options.with_sbeppc:
             return {
-                "gcc": "8.1",
+                "gcc": "8",
                 "clang": "9",
                 "apple-clang": "11"
             }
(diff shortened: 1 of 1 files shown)


recipes/sbepp/all/conanfile.py
```diff
@@ -37,7 +37,7 @@ def _min_cppstd(self):
     def _compilers_minimum_version(self):
         if self.options.with_sbeppc:
             return {
-                "gcc": "9",
+                "gcc": "8.1",
```
valgur, line 40:
```suggestion
                "gcc": "8",
```
You will have to use additional logic to not exclude compiler.version=8 otherwise:
```python
def loose_lt_semver(v1, v2):
    return all(int(p1) < int(p2) for p1, p2 in zip(str(v1).split("."), str(v2).split(".")))
```

ujos: I did not test gcc8. Most probably it works too, but it is not available on the ftp.

franramirez688 (approved):

ericLemanissier (approved):

PR #23246, opened 2024-03-25
zstd: remove vulnerable versions (< 1.4.9)

valgur:
Currently all versions < 1.4.9 are affected: https://repology.org/project/zstd/cves

Also a good reason to clean up the list of outdated versions a bit.


jwillikers (approved):

RubenRBS (approved):
Note that this does not remove them from the conancenter remote, but new revisions won't be published for those, which I'm ok with, as no recipes in CCI depend on the deleted versions

Hopefuly we get to share some cool security integrations for these kind of situations we're working on soon-ish! :)

Hopobcn:
> Note that this does not remove them from the conancenter remote, but new revisions won't be published for those, which I'm ok with, as no recipes in CCI depend on the deleted versions
> 
> Hopefuly we get to share some cool security integrations for these kind of situations we're working on soon-ish! :)

Looking forward for that!
Conan being able to warn/error if you're building with a library/version that has published CVE would be awesome.

PR #23255, opened 2024-03-25
harfbuzz: add v8.3.1, drop vulnerable versions (< 7.0)

valgur:
Based on https://repology.org/project/harfbuzz/information


jwillikers (approved):

RubenRBS (approved):
Approving because I can't find issues/PR comments mentioning the need to keep these around

RubenRBS (changes requested):
In fact, let me check with the team before :)

PR #23261, opened 2024-03-25
libftp: add recipe

toge:
Specify library name and version:  **libftp/***

---

### DIFF
--- a/recipes/libftp/all/conanfile.py
+++ b/recipes/libftp/all/conanfile.py
@@ -72,6 +72,7 @@ class LibFTPConan(ConanFile):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
         tc.variables["LIBFTP_BUILD_TEST"] = False
         tc.variables["LIBFTP_BUILD_EXAMPLE"] = False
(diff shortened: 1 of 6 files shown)


valgur (approved):

recipes/libftp/all/conanfile.py
```diff
@@ -0,0 +1,104 @@
...
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
```
czoido, line 75: Thanks lot for the contribution @toge! I think that if the library was not originally intended to work with shared It's better to add a `ConanInvalidConfiguration` for `Windows` and `shared=True`. I know we already have recipes that add this but this can be problematic and I think for new recipes we should try to avoid this unless it's really necessary for some reason.

czoido: I have just pushed some changes to add this `ConanInvalidConfiguration`

franramirez688 (approved):

PR #23272, opened 2024-03-26
glfw: limit xorg library dependencies

johningve:
Previously the glfw package would implicitly require linking with xorg::xorg, which includes all the libraries exposed by the xorg package.

This change explicitly declares the requirements of glfw, limiting the xorg libraries to only those needed.

From version 3.4 these dependencies are no longer necessary since glew implemented runtime platform detection.

Also removes vulkan_static option in version 3.4, since it is no longer supported by glfw.

Specify library name and version:  **glfw/all**

---


recipes/glfw/all/conanfile.py
```diff
@@ -217,6 +219,17 @@ def package_info(self):
...
+        # Starting with version 3.4, glfw loads the platform libraries at runtime
+        self.cpp_info.requires = []
+        if Version(self.version) < "3.4":
+            if self.options.get_safe("vulkan_static"):
+                self.cpp_info.requires.extend(["vulkan-loader::vulkan-loader"])
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                if self.options.get_safe("with_x11", True):
+                    self.cpp_info.requires.extend(["xorg::x11"])
+            if self.options.get_safe("with_wayland"):
+                self.cpp_info.requires.extend(["wayland::wayland", "xkbcommon::xkbcommon"])
```
valgur, line 231:
```suggestion
        if self.options.get_safe("vulkan_static"):
            self.cpp_info.requires.extend(["vulkan-loader::vulkan-loader"])
        if self.settings.os in ["Linux", "FreeBSD"]:
            if self.options.get_safe("with_x11", True):
                self.cpp_info.requires.extend(["xorg::x11"])
        if self.options.get_safe("with_wayland"):
            self.cpp_info.requires.extend(["wayland::wayland", "xkbcommon::xkbcommon"])
```
I don't think the version check is necessary as it's already handled by the deleted options in `configure()`, no?

johningve: I think the version check is needed because even when you specify `with_x11=True` or `with_wayland=True`, the glfw library does not require linking with these after version 3.4, which implemented the runtime platform detection.

jwillikers (approved):

perseoGI (dismissed):
LGTM

uilianries (approved):
LGTM: https://github.com/glfw/glfw/compare/3.3.8..3.4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

valgur:
@perseoGI Needs a follow-up PR.

perseoGI:
@valgur Which logic are you refering?

valgur:
The requires being added for vulkan, xorg and wayland on lines 230-251 are being duplicated by the logic added in this PR.

perseoGI:
Okey I see. Lines https://github.com/conan-io/conan-center-index/blob/dc3c064276ea0f59a50a2d63ab0763ed5c88ef60/recipes/glfw/all/conanfile.py?plain=1#L230-L251 seems to be more complete.
But in this PR the version control was added. Should we keep the previous lines with the version control?
Could you @johningve or @valgur open a PR to fix this?

PR #23281, opened 2024-03-26
Use (ftp.gnu.org) as fallback of (ftpmirror.gnu.org), as the second might fail due to unsafe crypto

es20490446e:
With the departure towards OpenSSL 3 some mirrors are still using old unsafe renegotiation, and that's preventing from building this recipe.

I have noticed the mirror here about the issue, but for the time being it's just better to use the non mirrored FTP.

I will edit this back when I'm noticed it's fixed.

autoconf/2.71: ERROR: Error downloading file https://ftpmirror.gnu.org/autoconf/autoconf-2.71.tar.xz: 'HTTPSConnectionPool(host='ftp.rediris.es', port=443): Max retries exceeded with url: /mirror/GNU/autoconf/autoconf-2.71.tar.xz (Caused by SSLError(SSLError(1, '[SSL: UNSAFE_LEGACY_RENEGOTIATION_DISABLED] unsafe legacy renegotiation disabled (_ssl.c:1006)')))' ERROR: autoconf/2.71: Error in source() method, line 51
        ConanException: Error downloading file https://ftpmirror.gnu.org/autoconf/autoconf-2.71.tar.xz: 'HTTPSConnectionPool(host='ftp.rediris.es', port=443): Max retries exceeded with url: /mirror/GNU/autoconf/autoconf-2.71.tar.xz (Caused by SSLError(SSLError(1, '[SSL: UNSAFE_LEGACY_RENEGOTIATION_DISABLED] unsafe legacy renegotiation disabled (_ssl.c:1006)')))'

---


recipes/autoconf/all/conandata.yml
```diff
@@ -1,9 +1,9 @@
 sources:
   "2.72":
-    url: "https://ftpmirror.gnu.org/autoconf/autoconf-2.72.tar.xz"
+    url: "https://ftp.gnu.org/gnu/autoconf/autoconf-2.72.tar.xz"
```
valgur, line 3:
```suggestion
    url:
      -  "https://ftpmirror.gnu.org/autoconf/autoconf-2.72.tar.xz"
      -  "https://ftp.gnu.org/gnu/autoconf/autoconf-2.72.tar.xz"
```
The reliability of ftpmirror is definitely a problem, but it should still be the first option since the download speed from the main site can be painfully slow, taking several minutes to download an archive. It can and should be added as a backup source instead.

recipes/autoconf/all/conandata.yml
```diff
@@ -1,9 +1,9 @@
...
     sha256: "ba885c1319578d6c94d46e9b0dceb4014caafe2490e437a0dbca3f270a223f5a"
   "2.71":
-    url: "https://ftpmirror.gnu.org/autoconf/autoconf-2.71.tar.xz"
+    url: "https://ftp.gnu.org/gnu/autoconf/autoconf-2.71.tar.xz"
```
valgur, line 6:
```suggestion
    url:
      -  "https://ftpmirror.gnu.org/autoconf/autoconf-2.71.tar.xz"
      -  "https://ftp.gnu.org/gnu/autoconf/autoconf-2.71.tar.xz"
```

jcar87:
> Shall I perform any extra action?
> 
> Is it mandatory that I port this recipe for this change to be accepted?

The recipe is already compatible, this was an issue as the CLA was not signed by the time the build started.

I can relaunch it, however, I would advise to apply the changes as suggested by @valgur  and add two lines instead.

As reviewers this is a constant back and forth - some experience issues with `ftp.gnu.org` and others have intermittent issues with specific mirrors in `ftpmirror.gnu.org`. Given that Conan can easily fallback, I'd suggest listing both URLS, perhaps with `ftpmirror.gnu.org` as the first attempt, as advised by GNU themselves:

https://www.gnu.org//server/mirror.html

> We recommend using this generic ftpmirror.gnu.org address wherever possible in links, documentation, etc., to reduce load on the main GNU server.

 Listing just one may work for you today, but not for others, or not in the future, so the fallback is probably sensible. 

On the other hand, if there an issue with the mirrors, especially on systems with new OpenSSL 3, I would advise reporting this issue to GNU directly - the closest contact information I can find related to mirrors is at webmasters@gnu.org. I think it's sensible that it should be pointed out that clients on newer OpenSSL, that some mirrors are not complying with https://www.rfc-editor.org/rfc/rfc5746.html - if this can be confirmed.

es20490446e:
@conan-io/barbarians This pull request seems ready to merge.

valgur (approved):

RubenRBS (approved):
Thanks!

es20490446e:
@RubenRBS 👌😉

PR #23288, opened 2024-03-27
fontconfig: add option to use system directories

### LINKED ISSUE #5782: [package] fontconfig/2.13.93: How to use with sysconfdir set on build machine

johningve:
The fontconfig package currently tries to load fonts from some directory within the conan package directory. This results in the following error when this directory is not present:

```
Fontconfig error: Cannot load default config file: No such file: (null)
```

This commit adds an option `system_dirs` to use system directories instead.

Fixes #5782

Specify library name and version:  **fontconfig/2.13.93+**

---


recipes/fontconfig/meson/conanfile.py
```diff
@@ -79,12 +82,23 @@ def generate(self):
...
+        build_script = os.path.join(self.source_folder, "meson.build")
+        replace_in_file(self, build_script,
                         "freetype_req = '>= 21.0.15'",
                         f"freetype_req = '{Version(self.dependencies['freetype'].ref.version)}'")
 
+        if self.settings.os in ("Linux", "FreeBSD") and self.options.get_safe("system_dirs", False):
+            self._replace_configuration_string(build_script, "CONFIGDIR", "fc_configdir", "'/etc/fonts/conf.d'")
+            self._replace_configuration_string(build_script, "FC_CACHEDIR", "fc_cachedir", "'/var/cache/fontconfig'")
+            self._replace_configuration_string(build_script, "FC_TEMPLATEDIR", "fc_templatedir", "'/usr/share/fontconfig/conf.avail'")
+            self._replace_configuration_string(build_script, "FONTCONFIG_PATH", "fc_baseconfigdir", "'/etc/fonts'")
```
valgur, line 100: You should be able to accomplish the same without patching simply via project_options in generate(), I think:
```
tc.project_options["fc_configdir"] = "/etc/fonts/conf.d"
tc.project_options["fc_cachedir"] = "/var/cache/fontconfig"
tc.project_options["fc_templatedir"] = "/usr/share/fontconfig/conf.avail"
tc.project_options["fc_baseconfigdir"] = "/etc/fonts"
```
Also, you should probably override `xml-dir` as well? https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/meson_options.txt?ref_type=heads#L32-45

valgur (changes requested):
This fixes what seems like an important oversight. Thanks!

Should use `project_options` instead, though.

valgur:
Could setting either of these options get the best of both worlds without introducing a recipe option, maybe?
https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/meson_options.txt?ref_type=heads#L25-29

johningve:
Incidentally I noticed that the paths get different defaults on Conan 2 compared to Conan 1 (that is, when `system_dirs=False`). The Paths on Conan 2 are almost correct for a Linux system. Though the Template and XML directory paths are wrong.

Conan 2:

```
  Defaults
    Hinting                    : slight
    Font directories           : /usr/share/fonts, /usr/local/share/fonts
    Additional font directories:

  Paths
    Cache directory            : /var/cache/fontconfig
    Template directory         : /share/fontconfig/conf.avail
    Base config directory      : /etc/fonts
    Config directory           : /etc/fonts/conf.d
    XML directory              : /share/xml/fontconfig

```

Conan 1:

```
  Defaults
    Hinting                    : slight
    Font directories           : /usr/share/fonts, /usr/local/share/fonts
    Additional font directories:

  Paths
    Cache directory            : /home/johnio/.conan/data/fontconfig/2.14.2/_/_/package/91934fc8bcc809a17bd34af66d8491f41cb49e99/var/cache/fontconfig
    Template directory         : /home/johnio/.conan/data/fontconfig/2.14.2/_/_/package/91934fc8bcc809a17bd34af66d8491f41cb49e99/share/fontconfig/conf.avail
    Base config directory      : /home/johnio/.conan/data/fontconfig/2.14.2/_/_/package/91934fc8bcc809a17bd34af66d8491f41cb49e99/etc/fonts
    Config directory           : /home/johnio/.conan/data/fontconfig/2.14.2/_/_/package/91934fc8bcc809a17bd34af66d8491f41cb49e99/etc/fonts/conf.d
    XML directory              : /home/johnio/.conan/data/fontconfig/2.14.2/_/_/package/91934fc8bcc809a17bd34af66d8491f41cb49e99/share/xml/fontconfig

```

recipes/fontconfig/meson/conanfile.py
johningve: I thought I had tried that previously, but you are right, it works on Conan 2.

Unfortunately, on Conan 1, Meson attempts to install something to /usr and fails. Looks like it is due to missing this change: https://github.com/conan-io/conan/pull/15706

johningve:
> Could setting either of these options get the best of both worlds without introducing a recipe option, maybe? https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/meson_options.txt?ref_type=heads#L25-29

Looks like those options are used to specify the location of fonts, not the config/cache directories.

uilianries (changes requested):
@johningve Thank you for your suggestion of fixing that case. I would not go for adding a new option only change an environment variable, it's too much. Conan offers mechanisms to do it without touching the recipe. Please, read https://github.com/conan-io/conan-center-index/issues/5782#issuecomment-2149686757

johningve:
> @johningve Thank you for your suggestion of fixing that case. I would not go for adding a new option only change an environment variable, it's too much. Conan offers mechanisms to do it without touching the recipe. Please, read [#5782 (comment)](https://github.com/conan-io/conan-center-index/issues/5782#issuecomment-2149686757) and check if fits your case.

It is not about changing an environment variable, that is a workaround to this issue. What this is about is changing the  location where fontconfig will look for fonts by default. I want to be able to include fontconfig in my application and have it work without the user having to worry about environment variables. 

> I see a fragility here as you are hardcoding system paths that we can not guarantee. Conan only knows its own cache, outside of that is risky. Still, in case that option is not enough, we could add a Conan configuration to handle those variables, they fit better because are not part of the package ID.

I would say that at least on Linux, these paths are well known and relied upon. I get that Conan doesn't want to rely upon the outside too much, but the purpose of fontconfig is to enable applications to find and use the fonts available on the local system.

uilianries:
Hello again!

I’m following up on the discussion regarding the default system directories used by fontconfig. I’ve provided a more detailed explanation of how to parameterize the system folders in my comment here: https://github.com/conan-io/conan-center-index/issues/5782#issuecomment-2161050812.

Thank you once again for your pull request and for proposing an alternative to address this limitation. However, the current implementation seems to introduce additional complexity to the Conan recipe. This includes modifying the package ID with a new recipe option, requiring several custom user configurations, and applying patches dynamically.

For these reasons, I’ll be closing this pull request, as I’m not inclined to accept the proposed changes. I encourage you to explore using Conan’s native configuration to extend the Meson configuration for this use case.

Please feel free to share your feedback or ask for help on how to use Conan configuration for your particular case.

Thank you for your understanding!

PR #23291, opened 2024-03-27
open62541pp: new recipe

### LINKED ISSUE #18540: [request] open62541pp/0.4.1
### Package Name/Version

open62541pp/0.4.1

### Webpage

https://open62541pp.github.io/open62541pp/

### Source code

https://github.com/open62541pp/open62541pp

C++ wrapper of the amazing [open62541](https://open62541.org/) OPC UA library

High-level and easy to use classes similar to the [Python opcua/asyncua API](https://python-opcua.readthedocs.io/):
        [opcua::Server](https://open62541pp.github.io/open62541pp/classopcua_1_1Server.html)
        [opcua::Client](https://open62541pp.github.io/open62541pp/classopcua_1_1Client.html)
        [opcua::Node](https://open62541pp.github.io/open62541pp/classopcua_1_1Node.html)
    [Safe wrapper classes](https://open62541pp.github.io/open62541pp/group__TypeWrapper.html) for open62541 UA_* types to prevent memory leaks
    Native open62541 objects can be accessed using the handle() method of the wrapping classes
    [Extensible type conversion system](https://github.com/open62541pp/open62541pp#-type-conversion) to convert arbitrary types to/from native UA_* types
    Cross-platform (tested on Windows, Linux and MacOS)
    Compatible with all stable open62541 versions ≥ v1.0
    [Easy installation and integration with CMake](https://github.com/open62541pp/open62541pp#-getting-started)
    Use modern C++ (C++ 17) and best practices
    Less hurdle to get started with OPC UA

I'd like to use this library and consume it via conan, preferably via conan-center.

Library developer is open to this idea.
https://github.com/open62541pp/open62541pp/issues/20#issuecomment-1512697890
What is the preferred location for the recipe? With the source code, or separate in CCI? Or both?

I already have the basics working in my own [fork](https://github.com/ErikTempelaarVO/open62541pp/tree/feat/conan-recipe)
but I am unsure how to proceed.

valgur:
Based on https://github.com/ErikTempelaarVO/open62541pp/blob/feat/conan-recipe/conanfile.py by @ErikTempelaarVO.

@ErikTempelaarVO, I assume you are ok with me copying parts of your recipe for this PR?

* closes #18540

[Packaging status](https://repology.org/project/open62541pp/versions)


ErikTempelaarVO:
I used the same MPL-2.0 license as the original open62541pp-library and I believe that license allows that.
I'm pleased that this library will be in conan-center, and so does the author of the library (going by the thumbs up)

For the 0.12.0 version of the library you might need some of the changes from here:
https://github.com/oliver/open62541pp/commits/feat/conan-recipe/

lukasberbuer:
The open62541 requirement is currently pinned to version `v1.3.9`. We could relax this since open62541pp supports all open62541 versions sind `v1.0`.

valgur:
@lukasberbuer CCI does not generally allow version ranges for dependencies, unfortunately. However, consumers can still always add `self.requires("open62541/1.x.y", override=True)` to use a different version.

lukasberbuer:
Ok, thanks @valgur 👍

lukasberbuer:
Hi @valgur, can I support you with the IPO issue?

recipes/open62541pp/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,32 @@
+#include <open62541pp/open62541pp.h>
+
+#include <iostream>
+
+
+int main() {
+    opcua::Server server(4840 /* port */);
```
RubenRBS, line 7: Let's simplify the test package not to have to use server classes - reads like this could be hitting the network layer :)

lukasberbuer: Yes it does. The example is similar to the open62541 one: https://github.com/conan-io/conan-center-index/blob/master/recipes/open62541/all/test_package/test_package.c

I could provide a test without any server/client connection if needed.

RubenRBS: That'd be ace yes :) thanks!

lukasberbuer: Ops, I was wrong. The example seems to be fine. The actual server connection is only instantiated if `server.run()` is called.

valgur: I can simply move the test code to a dummy function that does not get called from main().

valgur:
@lukasberbuer Yes, your assistance on the IPO issue would be very welcome.

recipes/open62541pp/all/test_package/test_package.cpp
RubenRBS: > The actual server connection is only instantiated if server.run() is called.

Having this as a comment in the code itself then should be enough :)

lbckmnn:
Is someone still working on this? 
Can I assist somehow?

valgur:
@AbrilRBS The GCC build fails in CI with:

> .conan/data/open62541/1.3.9/_/_/package/75ef360a57ade25004dd5837d5c97cae478a69ec/lib/libopen62541.a' generated with LTO version 11.0 instead of the expected 11.3

My guess is that this might be fixed by regenerating the `open62541` binaries with the same newer GCC version. Can you help with that, please?

AbrilRBS:
There are a few packages, will update this branch when the binaries have been built, thanks for the heads-up @valgur!

AbrilRBS:
Seems like some configurations are having a few issues running the tests https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/prod-v1/TapaholesRepo/135-linux-gcc11/open62541/1.2.6//summary.json but it's progressing nicely otherwise

ErikTempelaarVO:
Looks like the problem @valgur mentioned is still there?

`"lto1: fatal error: bytecode stream in file '/home/conan/workspace/prod-v1/bsr/90552/bbfdc/.conan/data/open62541/1.2.6/_/_/package/176240f9a6169082d460c41d91c4bea5bc22a932/lib/libopen62541.a' generated with LTO version 11.0 instead of the expected 11.3"`

But only in profiles with

```
[options]
open62541:shared=False
```

lukasberbuer:
Hi @valgur, thanks for you efforts!
Could you please update the version to the latest v0.15.0? The header structure has finally been tidied up.
As a test, the [`server_minimal.cpp`](https://github.com/open62541pp/open62541pp/blob/v0.15.0/examples/server_minimal.cpp) should be sufficient.

There is one remaining error due to a version mismatch between open62541 and open62541pp:
```
lto1: fatal error: bytecode stream in file '/home/conan/workspace/prod-v1/bsr/12322/fadda/.conan/data/open62541/1.3.9/_/_/package/75ef360a57ade25004dd5837d5c97cae478a69ec/lib/libopen62541.a' generated with LTO version 11.0 instead of the expected 11.3
```

@AbrilRBS will this issue be resolved with a rebuild of open62541? Could you trigger a rebuild and run the CI workflows for this PR again?

AbrilRBS:
Forcing recompilation of the faulty binaries again, this time I'm forcing a full recompilation to avoid this issue, thanks for your patience. For those in the know, you're looking for tapaholes 184, will retrigger this PR once the builds have finished

AbrilRBS:
@valgur the remaining errors looks like a valid issue in the test package :)

lukasberbuer:
Thanks @AbrilRBS  for triggering the rebuild 👍 Big step forward!
@valgur, I think the update to the latest open62541pp release with the minimal example should fix the remaining issue.

lukasberbuer:
Thanks @valgur for the updates.
The build fails with GCC 7. We need full C++17 support for open62541pp. To my understanding this should require GCC >= 8.
Is there a way to require C++17 support for the compiler without specifying each minimum compiler version?

Following compiler versions are tested in the open62541pp CI: https://github.com/open62541pp/open62541pp/blob/master/.github/workflows/ci.yml
- GCC >= 9
- Clang >= 13
- Apple Clang >= 14
- Visual Studio >= 16

lukasberbuer:
I would recommend a simple test like this:

```cpp
#include <open62541pp/node.hpp>
#include <open62541pp/server.hpp>

int main() {
    opcua::Server server;
    opcua::Node parentNode(server, opcua::ObjectId::ObjectsFolder);
    opcua::Node myIntegerNode = parentNode.addVariable({1, 1000}, "TheAnswer");
    myIntegerNode.writeValueScalar(42);
}
```

This tests important components of open62541pp:
- `opcua::Server` creation
- `opcua::Node` interface, which is a high-level interface for all service sets (will be all included)
- Node management service set (`Node::addVariable`)
- `opcua::Variant` interface (`Node::writeValueScalar` is an alias for `Node::writeValue(Variant::fromScalar(...))`

recipes/open62541pp/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,10 @@
+#include <open62541pp/node.hpp>
+#include <open62541pp/server.hpp>
+
+int main() {
+    int port = 0;
+    opcua::Server server(port);
```
AbrilRBS, line 6: As far as I can see, this actually tries to bind a server right? Or is there some kind of initialization needed for that? The examples upstream seem to imply that it does not, and object instantiation enables it by default?

lukasberbuer: But why set port to `0`?

valgur: Ok, I just wanted to avoid opening any connections, if possible. It was not clear to me that an additional method call is needed for that. I can revert this part.

recipes/open62541pp/all/conanfile.py
```diff
@@ -39,10 +39,10 @@ def _min_cppstd(self):
     def _compilers_minimum_version(self):
         return {
             "gcc": "7",
```
lukasberbuer, line 41: This should be set to `9`.

recipes/open62541pp/all/test_package/test_package.cpp
lukasberbuer: open62541pp's `Server::run` wraps the `UA_Server_run_startup` and `UA_Server_run_iterate` functions.

The `UA_Server_run_startup` will spin-up the event loop and the network layers:

```c
/* Start: Spin up the workers and the network layer and sample the server's
 *        start time.
 * Iterate: Process repeated callbacks and events in the network layer. This
 *          part can be driven from an external main-loop in an event-driven
 *          single-threaded architecture.
 * Stop: Stop workers, finish all callbacks, stop the network layer, clean up */
UA_StatusCode UA_Server_run_startup(UA_Server *server);
UA_UInt16     UA_Server_run_iterate(UA_Server *server, UA_Boolean waitInternal);
UA_StatusCode UA_Server_run_shutdown(UA_Server *server);
```

valgur:
@lukasberbuer test_package failed for MSVC Debug shared build. The return code 3 doesn't suggest any kind of DLL issue or similar. Do you have any ideas what might have failed there?

lukasberbuer:
@valgur I have no idea at the moment. According to the docs, return code 3 is:
> Execution terminated due to manually stopping the process with Ctrl+C key combination.
> https://docs.conan.io/1/reference/commands/return_codes.html#user-ctrl-c

This problem occurs with both the shared library Debug and Release build.
Are there any options to get more debug information @AbrilRBS?

valgur:
This hangs and crashes with STATUS_STACK_BUFFER_OVERRUN error code on MSVC. Are you sure it's correct?
```cpp
    opcua::Node myIntegerNode = parentNode.addVariable({1, 1000}, "TheAnswer");
    myIntegerNode.writeValueScalar(42);
```
I dropped these lines from test_package for now.

lukasberbuer:
@valgur I just tested it locally with Visual Studio 17 and shared libraries and all examples and tests run fine (the test_package is just the stripped down `server_minimal.cpp` example).

AbrilRBS (dismissed):

AbrilRBS (approved):

recipes/open62541pp/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["UAPP_INTERNAL_OPEN62541"] = False
+        tc.variables["UAPP_BUILD_DOCUMENTATION"] = False
+        if is_msvc(self):
+            tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+        tc.variables["open62541_ipo"] = self.options.ipo
```
AbrilRBS, line 52: This option seems a bit iffy, but it's true that the interface library from Conan is not helping matters - any idea @jcar87?

perseoGI (dismissed):
LGTM thank you 🐸

ErikTempelaarVO:
@jcar87 
Any idea when you'll be able to review this?

AbrilRBS (approved):

PR #23301, opened 2024-03-28
libinput: Make wayland-protocols a build dependency

jwillikers:
This aligns with how wayland-protocols is handled in every other package. I missed this in my original PR to add the `libinput` package.


RubenRBS (approved):
Thanks for circling back to it :)

toge (approved):

Issue #22796, opened 2024-02-17
[question] How to use `openssl/3.2.1` with the FIPS module from `openssl/3.0.8`


gegles:
### What is your question?

We are trying to create a FIPS compliant app. For that, we want to use the latest `openssl/3.2.1` package via `self.requires("openssl/3.2.1")` but we also need to package the FIPS module (i.e. `fips.so` or `fips.dll` or `fips.dylib`) from the officially supported version, i.e. `openssl/3.0.8`. [See the official OpenSSL doc](https://www.openssl.org/source/):
```
Please follow the Security Policy instructions to download, build and install a validated OpenSSL FIPS provider. 
Other OpenSSL Releases MAY use the validated FIPS provider, but MUST NOT build and use their own FIPS provider. 
For example you can build OpenSSL 3.2 and use the OpenSSL 3.0.8 FIPS provider with it.
```

My initial solution (that sorta works) was the following:
```python
    def build_requirements(self):
        # FIPS
        # Note: Ideally, this should be a `self.requires` in `requirements`.
        # This openssl will be built with the `build` profile when ideally it should be built with the `host` profile
        self.tool_requires(f"openssl/3.0.8")

    def requirements(self):
        self.requires("openssl/3.2.1")  # Only explicitly needed to extract the openssl executable

    def generate(self):
        # Get fips shared library
        openssl_fips = self.dependencies.build["openssl"]
        openssl_mod_dir = openssl_fips.runenv_info.vars(self)["OPENSSL_MODULES"]
        build_lib_dir = os.path.join(self.build_folder, "lib", str(self.settings.build_type), "ossl-modules")
        copy(self, "fips.*", openssl_mod_dir, build_lib_dir)
        ...
```
This sorta works, except for when cross-compiling (i.e. macOS Silicon -> Intel). Indeed, the fips module will be of the `build` arch type (`armv8`) instead of the `host` arch type (`x86_64`).

Ideally, it would be great if we could somehow `self.requires()` 2 different versions of the same package, but that might be tricky.

Otherwise, is there a way (as part of the conanfile.py and using simple methods) to merely download a separate version of a package and extract what's needed out of it?

gegles:
FYI @uilianries @RubenRBS @grafikrobot

gegles:
@uilianries @RubenRBS @grafikrobot @Hopobcn @Croydon @jcar87, I know your busy, but any thoughts on this? Thx!

grafikrobot:
[...]

PR #23303, opened 2024-03-28
openssl-fips: create "alias" recipe for FIPS validated OpenSSL

gegles:
This is especially necessary for when 2 distinct versions of the FIPS module and the rest of the openssl libs/execs.

Specify library name and version:  **openssl-fips/3.0.9**

This is my first attempt to address https://github.com/conan-io/conan-center-index/issues/22796

With this PR, I was successfully able to do the following in my recipe:

```
    def requirements(self):
        self.requires("openssl/3.2.1")
        self.requires("openssl-fips/3.0.9")

    def generate(self):
        # Get fips shared library
        openssl_fips = self.dependencies["openssl-fips"]
        openssl_mod_dir = openssl_fips.runenv_info.vars(self)["OPENSSL_MODULES"]
        build_lib_dir = os.path.join(self.build_folder, "lib", str(self.settings.build_type), "ossl-modules")
        copy(self, "fips.*", openssl_mod_dir, build_lib_dir)
```

This PR currently duplicates the whole recipe as well as the test_package which is far from ideal...

Ideally, all we really need to do is the ability to optionally change the package name of the recipe, so we can build both `openssl` and `openssl-fips` from the same recipe...

I know `--name ` option can be used in the `conan create`, but I don't think conan center uses that. 

Any thoughts?

---

### DIFF
--- a/recipes/openssl-fips/all/conandata.yml
+++ b/recipes/openssl-fips/all/conandata.yml
@@ -1,0 +1,12 @@
+sources:
+  # Note: Only ever add the versions of OpenSSL that have been FIPS certified
+  3.0.9:
+    url:
+      - "https://www.openssl.org/source/openssl-3.0.9.tar.gz"
+      - "https://github.com/openssl/openssl/releases/download/openssl-3.0.9/openssl-3.0.9.tar.gz"
+    sha256: eb1ab04781474360f77c318ab89d8c5a03abc38e63d65a603cabbf1b00a1dc90
+  3.0.8:
+    url:
+      - "https://www.openssl.org/source/openssl-3.0.8.tar.gz"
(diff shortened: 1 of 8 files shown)


gegles:
FYI @Croydon

Croydon (changes requested):
I don't think this is a reasonable approach. Why would you use two different OpenSSL versions in the same build, instead of just the one that supports FIPS?

gegles:
Closing this in favor of solution presented [here](https://github.com/conan-io/conan-center-index/issues/22796#issuecomment-2080011745) by @jcar87 and worked on [in this Conan PR](https://github.com/conan-io/conan/pull/16132). Thx!

PR #23304, opened 2024-03-29
openssl: make things a bit more generic

gegles:
This removes some hardcoded names and allows for the package to be "renamed" more easily...

For example if one wanted to create a distinct `openssl-fips` package so as to use it in conjunction with the standard `openssl` package without any collision.

See https://github.com/conan-io/conan-center-index/issues/22796 for more context

Specify library name and version:  **openssl/3.x.x**

---


gegles:
FYI @Croydon

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -631,6 +631,7 @@ def _module_file_rel_path(self):
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "OpenSSL")
         self.cpp_info.set_property("cmake_find_mode", "both")
+        self.cpp_info.set_property("cmake_target_name", "openssl::openssl")
```
SpaceIm, line 634:
```suggestion
```

Unecessary and misleading since there is no such openssl::openssl in official FindOpenSSL.cmake

gegles: Removed. thx!

gegles:
@RubenRBS any progress here?

valgur (approved):

gegles:
@SpaceIm @RubenRBS could you approve/merge this, please?

Much appreciated!

Croydon:
The change here is harmless, but since the purpose is hidden and since it differs from other recipes, it might be very well, that this gets broken regularly one way or another 🤔

gegles:
> The change here is harmless, but since the purpose is hidden and since it differs from other recipes, it might be very well, that this gets broken regularly one way or another 🤔

I hear you @Croydon, but I'm willing to take the risk.. 😉 
Of course I would love for Conan Center and the community to come up with a good/better solution to [the very unusual FIPS versioning requirements](https://github.com/conan-io/conan-center-index/issues/22796), but at least this would enable us to have a very trivial `openssl-fips` recipe alias that we can simply host locally...

gegles:
@SpaceIm @RubenRBS, @Croydon  could you approve/merge this, please?

Much appreciated!

Croydon (approved):

gegles:
Thanks @Croydon and @valgur!

@RubenRBS, could we merge this now? Thx!

gegles:
@RubenRBS 🙏🏻. Thanks!

jcar87 (changes requested):
Hi @gegles, thanks for your contribution and for providing insights in https://github.com/conan-io/conan-center-index/pull/23304.

based on the details you have provided in the issue https://github.com/conan-io/conan-center-index/issues/22796 I think we can actually implement something along the lines of your initial proposal (a requires on an earlier version of itself) - which would be a lot simpler (a few lines, if I'm not mistaken), and would require no further changes - in particular, no two different recipes

gegles:
@jcar87 @RubenRBS, I think it would still make the code cleaner and less "hardcoded" to merge this in, despite the great & elegant alternaitve solution currently worked on in parallel, but if not, I am also fine closing this PR. LMK. Thx!

gegles:
@jcar87 @RubenRBS, I think it would still make the code cleaner and less "hardcoded" to merge this in, despite the great & elegant alternaitve solution currently worked on in parallel, but if not, I am also fine closing this PR. LMK. Thx!

gegles:
Closing due to lack of interest/consensus....

PR #23308, opened 2024-03-29
fix: duckdb 0.10.1 jemalloc no longer built for non ARM Linux

Taepper:
Specify library name and version:  **duckdb/0.10.1**

The recipe did not build on ARM Linux.

Here is an analog change in the setup.py provided by duckdb:
https://github.com/duckdb/duckdb/commit/4a89d97db8a5a23a15f3025c8d2d2885337c2637

---

### DIFF
--- a/recipes/duckdb/all/conanfile.py
+++ b/recipes/duckdb/all/conanfile.py
@@ -232,7 +232,8 @@ class DuckdbConan(ConanFile):
                 self.cpp_info.libs.append("visualizer_extension")
             if self.options.with_httpfs:
                 self.cpp_info.libs.append("httpfs_extension")
-            if Version(self.version) >= "0.6.0" and self.settings.os == "Linux":
+            if (Version(self.version) >= "0.6.0" and self.settings.os == "Linux" and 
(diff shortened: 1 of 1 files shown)


RubenRBS (commented):
Hi! Thanks a lot for taking the time to fix this

I also think this warrants some changes to the `validate()` method of the recipe to notify Conan that this combination is not possible? wdty? Pinging @uilianries too for his opiniong too :)

uilianries:
@RubenRBS I see no direct dependency to [jemalloc/5.3.0](https://conan.io/center/recipes/jemalloc?version=5.3.0) package in this recipe 🤔 It would need more investigation to understand, but in the current condition, changing only cpp_info, looks correct to me.

RubenRBS (approved):
After checking with @uilianries this looks good, we got confused for a second there, sorry about the noise :)

Approving now

corneliusroemer:
Unless I'm really misunderstanding something here, the title of the issue is wrong - it should drop the "non". It's not build for ARM anymore, not the opposite as implied by the title

> fix: duckdb 0.10.1 jemalloc no longer built for ~non~ ARM Linux

PR #23315, opened 2024-03-30
Add LibRaw 0.22.0-Devel202403 development snapshot with a ton of newl…

irieger:
Specify library name and version:  **libraw/0.22.0-devel202403**

Add LibRaw 0.22.0-Devel202403 development snapshot with a ton of newly supported image file formats. As there is only irregular releases every ~18 Months for LibRaw, development snapshot versions are offered to allow early access to features. As LibRaw is a library to read raw photo image formats and there is often new cameras introduced with slight changes in the image format, regular updates are helpful.

Is it viable to introduce this to cci? Already added it to my local artifactory so not too urgent for me but I'd assume this can also be helpful for other users.

---

### DIFF
--- a/recipes/libraw/all/conandata.yml
+++ b/recipes/libraw/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.22.0-devel202403":
+    url: "https://github.com/irieger/LibRaw/archive/0.22.0-Devel202403.tar.gz"
+    sha256: "c22c38b4a6b27e2ba08f0bf11c55974bcde9810cf332006d5c3744210ebb9135"
   "0.21.2":
     url: "https://github.com/LibRaw/LibRaw/archive/0.21.2.tar.gz"
     sha256: "7ac056e0d9e814d808f6973a950bbf45e71b53283eed07a7ea87117a6c0ced96"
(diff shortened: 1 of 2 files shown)


SpaceIm:
I don't think it's a good idea to add a version coming from a fork. Nobody reviews your fork, and no offense, it could be malicious.

irieger:
> I don't think it's a good idea to add a version coming from a fork. Nobody reviews your fork, and no offense, it could be malicious.

The only reason was to trigger a release of a commit from the original repository without any additional changes. However I understand the problem. Sadly the snapshot is not officially tagged & released on the original repository. Would there be a consistent way on github to get a fixed archive URL with fixed checksum to a specific commit properly?

uilianries:
@irieger Why not just asking the upstream to create a new tag instead?

uilianries (changes requested):
Blocking for discussion first about the fork.

irieger:
> @irieger Why not just asking the upstream to create a new tag instead?

On it, just requested the tagging there.

irieger:
It seems to be a very fluid thing with this "snapshot" so I'll close this MR and just keep my personal conanfile with the current tag or maybe switch it to use newer commits later.

(https://github.com/LibRaw/LibRaw/issues/637#issuecomment-2030324219)

PR #23334, opened 2024-04-02
gfortran: revive and redesign deprecated recipe

valgur:
This PR revives the `gfortran` recipe by adapting the GCC recipe from #21193 to focus solely on GFortran.

There is currently no usable Fortran compiler available on CCI, which is limiting for quite a few numerics libraries.

I don't think melding GFortran with the broader GCC recipe was a good idea. The additional C and C++ compilers plus the rest of the toolchain provided by the recipe are not required for Fortran compilation and they are only very likely to cause issues by accidentally overriding the C/C++ compiler. Plus, it makes for a slightly smaller package to download as well.

### DIFF
--- a/recipes/gfortran/all/conandata.yml
+++ b/recipes/gfortran/all/conandata.yml
@@ -1,7 +1,11 @@
 sources:
   "13.2.0":
-    url:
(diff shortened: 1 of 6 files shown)


valgur:
@uilianries Can you help with the missing binutils binary, please?

valgur:
Should be a part of the CI image or installed on the consumer system, really.

Issue #23327, opened 2024-04-01
[package] cpr/1.10.5: Unable to build with Android NDK.


sizeak:
### Description

When building for Android with the NDK (external & tool_requires), the OpenSSL headers cannot be found:
```
src/cpr/ssl_ctx.cpp:8:10: fatal error: 'openssl/err.h' file not found
```

This is because while the OpenSSL package is found, the `OPENSSL_INCLUDE_DIR` variable is not set. The usual workaround for this seems to be setting the `OPENSSL_ROOT_DIR` variable in the environment (or by passing it directly to CMake with `-DOPENSSL_ROOT_DIR`, which is a bit impractical when the build is happening in the middle of a package dependency tree. 

An alternative solution is to instead compute the include path from the link paths, which are set correctly, when it is found to be unset. This seems to work well here and doesn't impact currently working builds, since `OPENSSL_INCLUDE_DIR` will already be set in those cases. I'l open a PR with a patch and see about getting support in upstream too, unless someone suggests a cleaner fix.

### Package and Environment Details

* Package Name/Version: **cpr/1.10.5**
* Operating System+version: **Linux**
* Compiler+version: **Android NDK r26c (clang)**
* Conan version: **conan 2.2.1**
* Python version: **Python 3.11.8**

### Conan profile

[settings]
arch=x86_64
os=Android
os.api_level=33
compiler=clang
compiler.version=14
compiler.libcxx=c++_static
compiler.cppstd=20

[tool_requires]
android-ndk/r26c

[conf]
tools.cmake.cmake_layout:build_folder_vars=['settings.arch', 'settings.os']

### Steps to reproduce

conan create . --build=missing --profile android_x86_64 --version=1.10.5

### Logs

<details><summary>Click to expand log</summary>

```
conan create . --build=missing --profile conan_profiles/android_x86_64 --version=1.10.5

======== Exporting recipe to the cache ========
cpr/1.10.5: Exporting package recipe: /home/si/Projects/conan-center-index/recipes/cpr/all/conanfile.py
cpr/1.10.5: exports: File 'conandata.yml' found. Exporting it...
cpr/1.10.5: Calling export_sources()
cpr/1.10.5: Copied 1 '.py' file: conanfile.py
cpr/1.10.5: Copied 1 '.yml' file: conandata.yml
cpr/1.10.5: Copied 1 '.patch' file: 008-1.10.0-remove-warning-flags.patch
[...]

Issue #23251, opened 2024-03-25
[package] cpr/1.10.5: Not possible to build with openssl on Windows


nclindroos:
### Description

When building a project that depends on cpr with openssl on Windows, cpr fails to build with error 
```
error C1083: Cannot open include file: 'openssl/err.h': No such file or directory
```
It appears that the cpr recipe does not set the openssl dependency correctly on Windows. Here a small example project the reproduces the issue: https://github.com/nclindroos/cpr-openssl-windows

Looking in https://github.com/conan-io/conan-center-index/blob/master/recipes/cpr/all/conanfile.py it appears to me that the requires section is incorrect, it is
```python
    def requirements(self):
        self.requires("libcurl/[>=7.78.0 <9]", transitive_headers=True, transitive_libs=True)
        # FIXME: This is a very dirty hack.
        # with_ssl == _AUTO_SSL, cpr needs the openssl header to compile. But Conan recipe does not know which SSL library to use.
        # because cpr's CMakeLists.txt automatically detects SSL libraries with CPR_ENABLE_SSL == ON.
        if self.settings.os in ["Linux", "FreeBSD"] and self.options.with_ssl in ["openssl", CprConan._AUTO_SSL]:
            self.requires("openssl/[>=1.1 <4]")
```
but should be
```python
    def requirements(self):
        self.requires("libcurl/[>=7.78.0 <9]", transitive_headers=True, transitive_libs=True)
        # FIXME: This is a very dirty hack.
        # with_ssl == _AUTO_SSL, cpr needs the openssl header to compile. But Conan recipe does not know which SSL library to use.
        # because cpr's CMakeLists.txt automatically detects SSL libraries with CPR_ENABLE_SSL == ON.
        if (self.settings.os in ["Linux", "FreeBSD"] and self.options.with_ssl == CprConan._AUTO_SSL) or self.options.with_ssl == "openssl":
            self.requires("openssl/[>=1.1 <4]")
```

If this looks like a reasonable solution I can look into adding a pull request.

### Package and Environment Details

* Package Name/Version: **cpr/1.10.5**
* Operating System+version: **Windows 11 Pro (23H2)**
* Compiler+version: **msvc 19.39.33519**
* Docker image: n/a
* Conan version: **conan 2.0.17**
* Python version: **Python 3.12.2**

### Conan profile

Profile host:
[settings]
arch=x86_64
build_type=Debug
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
[...]

PR #23339, opened 2024-04-02
cpr: Fix build with openssl on Windows and Android (#23251, #23327) 

nclindroos:
* Add self.requires("openssl...") call when using openssl on windows

Specify library name and version:  **cpr/1.10.5**

This makes it possible to build cpr on windows with openssl. See issue #23251 for more discussion.

---

### DIFF
--- a/recipes/cpr/all/conanfile.py
+++ b/recipes/cpr/all/conanfile.py
@@ -86,7 +86,8 @@ class CprConan(ConanFile):
         # FIXME: This is a very dirty hack.
         # with_ssl == _AUTO_SSL, cpr needs the openssl header to compile. But Conan recipe does not know which SSL library to use.
         # because cpr's CMakeLists.txt automatically detects SSL libraries with CPR_ENABLE_SSL == ON.
-        if (self.settings.os in ["Linux", "FreeBSD"] and self.options.with_ssl == CprConan._AUTO_SSL) or self.options.with_ssl == "openssl":
+        if ((self.settings.os in ["Linux", "FreeBSD", "Android"] and self.options.with_ssl == CprConan._AUTO_SSL) or
+            self.options.with_ssl == "openssl"):
             self.requires("openssl/[>=1.1 <4]")
 
     def validate(self):
@@ -102,7 +103,7 @@ class CprConan(ConanFile):
             raise ConanInvalidConfiguration(f"Cannot compile {self.ref} with libstdc++ on clang < 9")
 
         ssl_library = str(self.options.with_ssl)
-        
+
         if ssl_library == "openssl" and is_apple_os(self):
             raise ConanInvalidConfiguration("OpenSSL is not supported on macOS")
         if ssl_library == "darwinssl" and not is_apple_os(self):
--- a/recipes/cpr/all/conanfile.py
+++ b/recipes/cpr/all/conanfile.py
@@ -86,7 +86,7 @@ class CprConan(ConanFile):
         # FIXME: This is a very dirty hack.
         # with_ssl == _AUTO_SSL, cpr needs the openssl header to compile. But Conan recipe does not know which SSL library to use.
         # because cpr's CMakeLists.txt automatically detects SSL libraries with CPR_ENABLE_SSL == ON.
-        if self.settings.os in ["Linux", "FreeBSD"] and self.options.with_ssl in ["openssl", CprConan._AUTO_SSL]:
+        if (self.settings.os in ["Linux", "FreeBSD"] and self.options.with_ssl == CprConan._AUTO_SSL) or self.options.with_ssl == "openssl":
             self.requires("openssl/[>=1.1 <4]")
 
     def validate(self):
--- a/recipes/cpr/all/conanfile.py
+++ b/recipes/cpr/all/conanfile.py
@@ -12,9 +12,6 @@ required_conan_version = ">=1.53.0"
 
 
 class CprConan(ConanFile):
-    _AUTO_SSL = "auto"
-    _NO_SSL = "off"
-
     name = "cpr"
     description = "C++ Requests: Curl for People, a spiritual port of Python Requests"
     license = "MIT"
@@ -27,14 +24,14 @@ class CprConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
-        "with_ssl": ["openssl", "darwinssl", "winssl", _AUTO_SSL, _NO_SSL],
+        "with_ssl": [False, "openssl", "darwinssl", "winssl"],
         "signal": [True, False],
         "verbose_logging": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
-        "with_ssl": _AUTO_SSL,
+        "with_ssl": "openssl",
         "signal": True,
         "verbose_logging": False,
     }
@@ -70,6 +67,10 @@ class CprConan(ConanFile):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+            self.options.with_ssl = "winssl"
+
+        if is_apple_os(self):
+            self.options.with_ssl = "darwinssl"
 
         if Version(self.version) < "1.10.0":
             del self.options.verbose_logging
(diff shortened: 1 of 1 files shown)


sizeak:
As mentioned in this issue https://github.com/conan-io/conan-center-index/issues/23251#issuecomment-2031565850, this PR can also fix the Android build with a slight tweak and is a much cleaner fix than https://github.com/conan-io/conan-center-index/pull/23328.

sizeak:
Please close this in favour of https://github.com/conan-io/conan-center-index/pull/23339

SpaceIm:
Does it fix Windows (in the title) or Android (from code)?

recipes/cpr/all/conanfile.py
```diff
@@ -86,7 +86,8 @@ def requirements(self):
         # FIXME: This is a very dirty hack.
         # with_ssl == _AUTO_SSL, cpr needs the openssl header to compile. But Conan recipe does not know which SSL library to use.
         # because cpr's CMakeLists.txt automatically detects SSL libraries with CPR_ENABLE_SSL == ON.
-        if self.settings.os in ["Linux", "FreeBSD"] and self.options.with_ssl in ["openssl", CprConan._AUTO_SSL]:
+        if ((self.settings.os in ["Linux", "FreeBSD", "Android"] and self.options.with_ssl == CprConan._AUTO_SSL) or
```
SpaceIm, line 89: In another PR, I would suggest to remove this ugly `_AUTO_SSL` and `_OFF` (leading to convoluted and fragile logic, this recipe can still fail with other OS not listed here), and keep things under control.

In upstream CMakeLists, when autodetection of SSL is enabled, it leads to:

- Windows => winssl
- Apple OS => darwinssl
- otherwise => OpenSSL (or MbedTLS)

Something under control should look like this:

```python
    options = {
        (...)
        "with_ssl": [False, "openssl", "darwinssl", "winssl"],
        (...)
    }
    default_options = {
        (...)
        "with_ssl":  "openssl",
        (...)
    }

    (...)

    def config_options(self):
        (...)
            self.options.with_ssl = "winssl"
        elif is_apple_os(self):
            self.options.with_ssl = "darwinssl"

    (...)

    def requirements(self):
        (...)
        if self.options.with_ssl == "openssl":

    def validate(self):
        (...)
        if ssl_library in ("openssl", "darwinssl") and ssl_library != self.dependencies["libcurl"].options.with_ssl:
                f"{self.ref}:with_ssl={self.options.with_ssl} requires libcurl:with_ssl={self.options.with_ssl}"
            )
        (...)

    (...)

    def generate(self):
       (...)
        tc.variables["CPR_FORCE_DARWINSSL_BACKEND"] = self.options.with_ssl == "darwinssl"
        tc.variables["CPR_FORCE_OPENSSL_BACKEND"] = self.options.with_ssl == "openssl"
        tc.variables["CPR_FORCE_WINSSL_BACKEND"] = self.options.with_ssl == "winssl"
        tc.variables["CMAKE_USE_OPENSSL"] = (self.options.with_ssl == "openssl")
        tc.variables["CPR_FORCE_MBEDTLS_BACKEND"] = False # TODO: add mbedtls backend
        tc.variables["CPR_ENABLE_SSL"] = bool(self.options.with_ssl)
```

nclindroos:
> Does it fix Windows with OpenSSL (in the title) or Windows with OpenSSL and Android (from code)?

I did the changes and have tested OpenSSL on Windows. I do not have a good setup to test Android, but according to @sizeak  this does fix his issue #23327 as well. I can update the title that it fixes both if you want to. Should I put both issue number in the title?

recipes/cpr/all/conanfile.py
nclindroos: I have updated the code as you suggested.

Something I still find surprising with this recipe on Windows is that libcurl uses OpenSSL as default, and CPR uses winssl. You either have to set libcurl/*:with_ssl or cpr/*:with_ssl for windows to build. Before the refactor the check that libcurl and cpr use the same SSL did not trigger due to AUTOSSL weirdness.

If I change CPR default to OpenSSL to match libcurl that could break build for projects that depended on default being winssl.

recipes/cpr/all/conanfile.py
```diff
@@ -101,24 +99,25 @@ def validate(self):
...
+            if ssl_library == "darwinssl" and not is_apple_os(self):
+                raise ConanInvalidConfiguration("DarwinSSL is only supported on macOS")
+            if ssl_library == "winssl" and self.settings.os != "Windows":
+                raise ConanInvalidConfiguration("WinSSL is only on Windows")
 
-        if ssl_library == "winssl" and self.dependencies["libcurl"].options.with_ssl != "schannel":
-            raise ConanInvalidConfiguration(
-                f"{self.ref}:with_ssl=winssl requires libcurl:with_ssl=schannel"
-            )
+            if ssl_library != "winssl" and ssl_library != self.dependencies["libcurl"].options.with_ssl:
```
SpaceIm, line 112:
```suggestion
            if ssl_library != "winssl" and ssl_library != self.dependencies["libcurl"].options.with_ssl:
```

You don't want to do this check when `self.options.with_ssl=False`

recipes/cpr/all/conanfile.py
```diff
@@ -101,24 +99,25 @@ def validate(self):
...
+                raise ConanInvalidConfiguration(
+                    f"{self.ref}:with_ssl={self.options.with_ssl} requires libcurl:with_ssl={self.options.with_ssl}"
+                )
+
+            if ssl_library == "winssl" and self.dependencies["libcurl"].options.with_ssl != "schannel":
+                raise ConanInvalidConfiguration(
+                    f"{self.ref}:with_ssl=winssl requires libcurl:with_ssl=schannel"
+                )
```
SpaceIm, line 120: The problem of this check is that now by default it raises for Windows, because by default cpr:with_ssl=winssl, and libcurl:with_ssl=openssl

So either:
- this check is not correct and should be removed, and ssl backend of libcurl doesn't matter (before this PR, cpr was built with winssl and libcurl with openssl, but it didn't raise because cpr:with_ssl was equal to "auto", and yet there was no issue)
- Change default value of with_ssl of cpr for windows (to openssl), even if it doesn't honor upstream default value

recipes/cpr/all/conanfile.py
```diff
@@ -144,7 +143,7 @@ def generate(self):
         tc.variables["CPR_FORCE_WINSSL_BACKEND"] = (self.options.with_ssl == "winssl")
         tc.variables["CMAKE_USE_OPENSSL"] = (self.options.with_ssl == "openssl")
         # If we are on a version where disabling SSL requires a cmake option, disable it
-        if self.options.with_ssl == CprConan._NO_SSL:
+        if not self.options.with_ssl:
             tc.variables["CPR_ENABLE_SSL"] = False
```
SpaceIm, line 147: Prefer to always set this option:

```python
tc.variables["CPR_ENABLE_SSL"] = bool(self.options.with_ssl)
```

nclindroos: I have update the line

recipes/cpr/all/conanfile.py
nclindroos: I did some testing on Windows, and building with libcurl/*:with_ssl set to OpenSSL and cpr/*:with_ssl set to winssl seems to produce a working binary that can perform https requests. This is the same as the default behavior before, and I will remove the check.

sizeak:
This also fixes the build with Emscripten if I append "Emscripten" to the `os` list `self.settings.os in ["Linux", "FreeBSD", "Android", "Emscripten"]`. Would you mind adding it to the PR please?

recipes/cpr/all/conanfile.py
```diff
@@ -101,24 +99,15 @@ def validate(self):
...
-            raise ConanInvalidConfiguration(
-                f"{self.ref}:with_ssl={self.options.with_ssl} requires libcurl:with_ssl={self.options.with_ssl}"
-            )
-
-        if ssl_library == "winssl" and self.dependencies["libcurl"].options.with_ssl != "schannel":
-            raise ConanInvalidConfiguration(
-                f"{self.ref}:with_ssl=winssl requires libcurl:with_ssl=schannel"
-            )
+            if ssl_library == "openssl" and is_apple_os(self):
+                raise ConanInvalidConfiguration("OpenSSL is not supported on macOS")
```
uilianries, line 106:
```suggestion
                raise ConanInvalidConfiguration("OpenSSL is not supported on macOS. Use DarwinSSL instead.")
```

uilianries (commented):
It look much clean now! Only a small suggest to guide users.

uilianries:
I see there is a window for MbedTLS support too, but it's a topic to another PR.

uilianries (approved):
Compared the automatic detection using https://github.com/libcpr/cpr/blob/1.10.5/CMakeLists.txt and looks fine. There is forced option for backends, but the current implementation follows the excepted behavior as well.

RubenRBS (approved):
Thanks a lot! :)

PR #23347, opened 2024-04-02
diligent-core: migrate to Conan v2

MikeLankamp:
Specify library name and version:  **diligent-core/\***

Upgrading diligent-core to Conan 2.

The main problems in using diligent-core with Conan 2 as-is are:
* a conflicting requirement of vulkan-headers (1.3.224.0 from `vulkan-validationlayers/1.3.224.1`, but 1.3.224.1 from `diligent-core` itself  and `volk/1.3.224.1`). This has been fixed by setting the conflicting dependencies to 1.3.224.0.
* The `cmake/3.24.2` tool requirement is not Conan2-enabled. The dependency has been generalized to `[>=3.24 <4]`, which allows Conan to pick a Conan2-enabled version.

This PR is similar to PR #21522, but doesn't attempt to bump dependencies, which comes with its own set of problems apparently. This PR just tries to get this recipe Conan2-compatible.

Some additional fixes in the conanfile are required to support Conan2. The largest of which is the change that the source is no longer downloaded into "source_subfolder". This was originally to provide a custom top-level CMakeLists.txt which defines library aliases for conan dependency. Without these, Diligent's own CMake files attempts to build empty folders in "ThirdParty". This was changed into modifying the original top-level CMakeLists.txt to _include_ the file with target aliases instead (similar to PR #21522).

---


valgur (approved):

MikeLankamp:
Hi @valgur, how could this PR get the required approvals? @RubenRBS, can you provide a required review?

jwillikers:
@MikeLankamp Could you review https://github.com/MikeLankamp/conan-center-index/pull/1? That change will use the `libglvnd` on Linux and FreeBSD. I figured it would be best to get that in your branch.

MikeLankamp:
>  I figured it would be best to get that in your branch.

May I ask why? Your change seems unrelated with my Conan 2.0 support and aims to get `libglvnd` used. Adding it to my branch would only pollute my PR to master with unrelated stuff (potentially dragging out the review).

Additionally, I'm not knowledgable enough about `libglvnd` and the Linux/FreeBSD ecosystem to judge whether its inclusion here makes sense, so it seems better for you to create a separate PR to master once my PR is in.

RubenRBS:
Hi! We have moved the discussion on `libglvnd` to https://github.com/conan-io/conan-center-index/issues/23649, so no need to apply those changes for now

As for the review of the PR itself, this is something I plan on doing next monday, sorry for the delay! From what I've seen in a quick glance, it looks mostly good to go! :)

MikeLankamp:
Hi, @danimtb, @RubenRBS, I don't want to sound pushy, but this PR has been awaiting a review for almost two months now. I'd like to remind you that without this change, this package is unusable with Conan 2.

I hope you can find the time soon to have a look at this change.

recipes/diligent-core/all/test_v1_package/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+cmake_minimum_required(VERSION 3.1)
```
danimtb, line 1: test_v1_package is not needed anymore

recipes/diligent-core/all/test_v1_package/conanfile.py
```diff
@@ -0,0 +1,17 @@
+from conans import ConanFile, CMake, tools
```
danimtb, line 1: test_v1_package is not needed anymore, please leave only test_package files and the cpp there

danimtb:
Hi @MikeLankamp,

I'm sorry for not getting back to you sooner. It is somehow hard to keep pace with the reviews. I have noticed you added files for test_v1_package, which are no longer needed. I don't see anything else to prevent this PR from getting merged. But let's wait for @RubenRBS review.

Thanks a lot for your patience 🙏

danimtb (approved):

jwillikers (approved):

recipes/diligent-core/all/conanfile.py
```diff
@@ -48,29 +50,28 @@ def validate(self):
...
         if not min_version:
-            self.output.warn("{} recipe lacks information about the {} compiler support.".format(
+            self.output.warning("{} recipe lacks information about the {} compiler support.".format(
                 self.name, self.settings.compiler))
         else:
             if Version(self.settings.compiler.version) < min_version:
                 raise ConanInvalidConfiguration("{} requires C++{} support. The current compiler {} {} does not support it.".format(
                     self.name, self._minimum_cpp_standard, self.settings.compiler, self.settings.compiler.version))
-        if self.settings.compiler == "Visual Studio" and "MT" in self.settings.compiler.runtime:
+        if visual.is_msvc_static_runtime(self):
```
RubenRBS, line 59:
```suggestion
        if visual.is_msvc(self) and visual.is_msvc_static_runtime(self):
```

MikeLankamp-TomTom: Is this really necessary? `is_msvc_static_runtime` already checks for `is_msvc`.

MikeLankamp: Is this really necessary? `is_msvc_static_runtime` already checks for `is_msvc`.

RubenRBS: Oh great catch, I didn't realize! TIL :)

RubenRBS (approved):

MikeLankamp:
Hi @valgur, @danimtb, @RubenRBS, sorry to bother you after this is merged (thanks for approving this PR!), but when I try to use this package, Conan 2 complains as follows:
```
diligent-core/2.5.2: Compatible package ID 0b6ce91bc61db9be4c90e8138b36d3ae166dc8be equal to the default package ID: Skipping it.
diligent-core/2.5.2: Checking 3 compatible configurations
diligent-core/2.5.2: Compatible configurations not found in cache, checking servers
diligent-core/2.5.2: 'f9de1b468d6419cb25174c1f3ce2c2b9866a247f': compiler.cppstd=14
diligent-core/2.5.2: Main binary package '0b6ce91bc61db9be4c90e8138b36d3ae166dc8be' missing. Using compatible package 'f9de1b468d6419cb25174c1f3ce2c2b9866a247f': compiler.cppstd=14
ERROR: Invalid setting 'MD/MDd' is not a valid 'settings.compiler.runtime' value.
Possible values are ['static', 'dynamic']
Read "http://docs.conan.io/2/knowledge/faq.html#error-invalid-setting"
```

I'm using MSVC and my local profile has `compiler.cppstd = 17`, so it can't find an exact prebuilt match. It uses `cppstd_compat.py` to find the prebuilt `cppstd=14` package, but in processing that one, it seems to re-validate the settings from the package_id. And "MD/MDd" is not valid.

Is this is a bug in Conan, or a bug in this package?

On the one hand, [documentation](https://docs.conan.io/2/reference/conanfile/methods/package_id.html#partial-information-erasure) seems to indicate you can put any string in `self.info` as its just for package_id creation, so in that case, why does Conan re-validate it? On the other hand, documentation also says this is practice discouraged.

That said, I also don't quite get why the runtime property is set to both release/debug. I suppose this reuses the same prebuilt package for both configurations, but I don't know why. It's been this way since this recipe got added in #7804. @AndreyMlashkin, do you perhaps remember why?

PR #23349, opened 2024-04-02
OpenImageIO: Add version 2.5.10.1

irieger:
Specify library name and version:  **openimageio/2.5.10.1**

Add the newest version of OpenImageIO. Contains - as always with OIIO a larger patch, but it has only two lines changed compared to the previous patch for version 2.5.9.0, which are lines 175 and 176 in the patch file.

---


RubenRBS (approved):
> but it has only two lines changed compared to the previous patch for version 2.5.9.0, which are lines 175 and 176 in the patch file.

Thanks for that! :)

toge (approved):

PR #23358, opened 2024-04-03
Libgpiod update source url

ashley-b:
Specify library name and version:  **libgpiod**

- Source URLs are broken. This PR updates URLs to point to authors recommend location 
https://github.com/brgl/libgpiod/issues/71

- Also added new version 2.1.1

---

### DIFF
--- a/recipes/libgpiod/all/conandata.yml
+++ b/recipes/libgpiod/all/conandata.yml
@@ -1,13 +1,13 @@
 sources:
   "2.1":
-    url: "https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/snapshot/libgpiod-2.1.tar.gz"
-    sha256: "fd6ed4b2c674fe6cc3b481880f6cde1eea79e296e95a139b85401eaaea6de3fc"
+    url: "https://mirrors.edge.kernel.org/pub/software/libs/libgpiod/libgpiod-2.1.tar.gz"
+    sha256: "1f7ddd3d8b1331da8735a88ed6ec55c82c1c469b4c38a118ddcd651db4808d98"
   "2.0.1":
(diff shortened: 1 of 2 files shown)


uilianries (commented):
@ashley-b Thank you for checking with the upstream first! 👏 

As a mirror, it should have the very same checksum, so we could use as a list in conandata.yml. I would ask to investigate a bit more to understand why they don't match. 

Plus, I checked locally and the original url is working. It could be a temporary error, could you please check it again.

ashley-b:
>
> As a mirror, it should have the very same checksum, so we could use as a list in conandata.yml. I would ask to investigate a bit more to understand why they don't match. 

From the authors comments 
"
You'll be better off downloading tarballs from https://mirrors.edge.kernel.org/pub/software/libs/libgpiod/ too - they're signed with my GPG key and generated using make dist as opposed to just the repo snapshot.
"
So make dist has created some changes 

I have done a diff and can see there are generated Makefile.in in the new link

> 
> Plus, I checked locally and the original url is working. It could be a temporary error, could you please check it again. 

I've check it today and the URL is still broken for me and check with some one and they have the same issue

uilianries:
@ashley-b You informed to the author the mirror is down for you too, but it's working. Not only the mirror, but also the snapshot:

- https://www.isitdownrightnow.com/mirrors.edge.kernel.org.html
- https://www.isitdownrightnow.com/git.kernel.org.html

I would suggest checking your network first. What exactly command and output do you have when trying to download? I guess you are having some limitation in your organization only.

ashley-b:
I've tested the snapshot today, and it seems to be back. It's a strange fault.
I did find someone else with the same issue https://github.com/microsoft/vcpkg/issues/37948

ashley-b:
The source URL still seems to be working. I will close this PR

PR #23361, opened 2024-04-03
libpcap: expose the with-snf option for recipes

darakelian:
Specify library name and version:  libpcap/*

Some of the build agents we use have myricom support. This causes libpcap to try and link against the myricom snf libraries (since with-snf=yes is the default) which causes problems when we use our internal libpcap pre-built package on a host that does not have myricom snf support. We had an internal fork of this recipe for conan 1 with this functionality supported but as we are migrating to conan 2, it would be easier if this could be mainlined and we can decom the fork of the recipe.

---


recipes/libpcap/all/conanfile.py
```diff
@@ -29,11 +29,13 @@ class LibPcapConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "enable_libusb": [True, False],
+        "with_snf": [None, True, False],
```
uilianries, line 32:
```suggestion
        "with_snf": [True, False],
```

Let's keep binary decision, no auto values. We can sure about its default value: https://github.com/the-tcpdump-group/libpcap/blob/master/configure.ac#L1515

recipes/libpcap/all/conanfile.py
```diff
@@ -29,11 +29,13 @@ class LibPcapConan(ConanFile):
...
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "enable_libusb": False,
+        "with_snf": None,
```
uilianries, line 38:
```suggestion
        "with_snf": True,
```

recipes/libpcap/all/conanfile.py
```diff
@@ -115,6 +117,11 @@ def generate(self):
                 "--disable-dbus",
                 "--disable-rdma",
             ])
+            
+            snf = self.options.get_safe("with_snf")
+            if snf:
+                tc.configure_args.append(f"--with-snf={yes_no(snf)}")
+
```
uilianries, line 124:
```suggestion
                "--disable-rdma",
                "--with-snf={yes_no(self.options.with_snf)}",
            ])
```

uilianries (commented):
@darakelian Thank you for your PR. It's an authentic case for adding this option. Let's simplify a bit, the default value is known, so adding auto would complicated more.

valgur:
You should also export a dependency on libsnf in `package_info()`. Otherwise static builds will not work without additional linker flags on the consumer side.
```py
if self.options.get_safe("with_snf"):
    self.cpp_info.system_libs.append("snf")
```

recipes/libpcap/all/conanfile.py
valgur: Missing an `f` from the f-string.

recipes/libpcap/all/conanfile.py
darakelian: After reading through the configure file, I think this should default to `False` (though maybe `None` still makes sense) since I believe the "yes, if present" means that `--with-snf=yes` and `--with-snf` are equivalent.

recipes/libpcap/all/conanfile.py
darakelian: Why also `--disable-rdma`?

darakelian:
> You should also export a dependency on libsnf in `package_info()`. Otherwise static builds will not work without additional linker flags on the consumer side.
> 
> ```python
> if self.options.get_safe("with_snf"):
>     self.cpp_info.system_libs.append("snf")  # the library name might be incorrect
> ```

How would I go about confirming the correct name? I don't see a recipe for libsnf either.

valgur:
@darakelian `snf` is the correct library name to add to `system_libs`: https://github.com/the-tcpdump-group/libpcap/blob/libpcap-1.10.4/configure.ac#L1612-L1614

You need to add it as a "system library" because it is not provided by this package or other Conan packages, but it's still required for linking.

recipes/libpcap/all/conanfile.py
valgur: `--disable-rdma` is just an existing unchanged line.

recipes/libpcap/all/conanfile.py
darakelian: Hi, I was reading through the libpcap stuff more and I do believe None is a valid option here, when `--with-snf` is not present, the autoconfigure stuff will attempt to see if the snf library is present at `/opt/snf` and if so link against it. This is different behavior than explicitly passing yes/no

darakelian: I am not sure how to best model this with conan2, if it is even possible? Maybe defaulting to false/true is sufficient for this use case?

recipes/libpcap/all/conanfile.py
```diff
@@ -115,6 +117,11 @@ def generate(self):
                 "--disable-dbus",
                 "--disable-rdma",
             ])
+            
+            snf = self.options.get_safe("with_snf")
+            if snf is not None:
+                tc.configure_args.append(f"--with-snf={yes_no(snf)}")
```
uilianries, line 123:
```suggestion
            tc.configure_args.append(f"--with-snf={yes_no(self.options.with_snf)}")
```
Always present don't need to check.

valgur (approved):

recipes/libpcap/all/conanfile.py
```diff
@@ -29,11 +29,13 @@ class LibPcapConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "enable_libusb": [True, False],
+        "with_snf": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "enable_libusb": False,
+        "with_snf": False,
```
uilianries, line 38:
```suggestion
        "with_snf": False,
```

https://github.com/the-tcpdump-group/libpcap/blob/libpcap-1.10.4/CMakeLists.txt#L521
It's enabled by default, but I'm fine accepting the opposite, because it's private product.

uilianries (approved):
LGTM. libpcap tries to find, when enabled, but no adds REQUIRED once enabled.

PR #23362, opened 2024-04-04
libffi: Add 3.4.6, remove 3.4.2 and 3.4.3

Ahajha:
Specify library name and version:  **libffi/3.4.6**

The patch `0007-3.4.3-forward-declare-open_temp_exec_file.patch` (and friends) is no longer needed for the new version, as it has been upstreamed.

The MSVC build changed slightly: https://github.com/libffi/libffi/commit/2b2f61f6e6a79d46fa277965745a8cd8bb5f04bc, essentially there's a new `FFI_STATIC_BUILD` that replaces `FFI_BUILDING`.

I'm not certain that the `FFI_BUILDING`/`FFI_STATIC_BUILD` macros need to be in the package info for non-MSVC builds. I've kept them for now, but maybe we could revisit at some point.
 
Older versions of libffi don't support Apple ARM CPUs, it looks like there was an attempt to backport the fix, but the patch itself wasn't actually merged: https://github.com/conan-io/conan-center-index/pull/5085. There are plenty of newer versions so I don't think it's worth fixing at this point. Unsure why this was still an issue.

---


Ahajha:
@uilianries @RubenRBS Do you think it's about time to remove libffi 3.4.2 and 3.4.3? 3.4.4 is used everywhere (except llvm-core which is still on 3.3). 3.4.4 is already about two years old, I don't think we need anything older.

Ahajha:
...Very confused as to why 3.3 broke. Will investigate.

RubenRBS (dismissed):
Thanks a lot for the detailed PR body, it's super appreciated :)

Ahajha:
@RubenRBS What do you think about removing the older versions? Only 3.3 is used, and it's by llvm-core, which is getting updated in another PR. 3.4.4 itself has been out for nearly 2 years, I don't think we need anything older.

uilianries:
@Ahajha you could remove `3.4.3` and `3.4.2`. Others I would prefer in another PR, after updating other recipes first. Thank you for checking it.

Ahajha:
@uilianries I've removed those two versions. I realize my comment wasn't clear, 3.4.4 I think should stay for now (and would have to anyways), 3.3 was the other version that I think should be removed, but I'll wait until the llvm-core PR is merged to do that.

uilianries:
@Ahajha You need to remove both version from conandata.yml too.

Ahajha:
@uilianries Thanks, was a bit tired last night :)

uilianries:
> @uilianries Thanks, was a bit tired last night :)

Thank you for contributing!!!! 👏

uilianries (approved):
LGTM

recipes/libffi/all/conandata.yml
```diff
@@ -1,17 +1,19 @@
...
-    sha256: "4416dd92b6ae8fcb5b10421e711c4d3cb31203d77521a77d85d0102311e6c3b8"
-  "3.4.2":
-    url: "https://github.com/libffi/libffi/releases/download/v3.4.2/libffi-3.4.2.tar.gz"
-    sha256: "540fb721619a6aba3bdeef7d940d8e9e0e6d2c193595bc243241b77ff9e93620"
   "3.3":
     url: "https://github.com/libffi/libffi/releases/download/v3.3/libffi-3.3.tar.gz"
     sha256: "72fba7922703ddfa7a028d513ac15a85c8d54c8d67f55fa5a4802885dc652056"
 patches:
+  "3.4.6":
+    - patch_file: "patches/0002-3.4.6-fix-libtool-path.patch"
```
Croydon, line 13: Please start adding `patch_description` and `patch_type`. Additionally `patch_source` if applicable

Ahajha: These were from existing patches, but I can do some digging and figure out their purposes.

Issue #23309, opened 2024-03-29
[package] ffmpeg/6.1: Build fails on Windows, missing aom


elvisdukaj:
### Description

Trying to build `ffmpeg` I have the following error on Windows:

```
ERROR: aom >= 1.0.0 not found using pkg-config
```

It seems that autotools is failing to find the packages with `pkg-config` . The possilbe issue is that `PKG_CONFIG_PATH` is not set correctly. 

### Package and Environment Details

* Package Name/Version: **ffmpeg/6.1**
* Operating System+version: **Windows**
* Compiler+version: **MSVC 193**
* Conan version: **conan 2.2.2**
* Python version: **Python 3.11.3**

### Conan profile

```
======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=23
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[tool_requires]
*: ninja/1.11.1, cmake/3.28.1
[conf]
tools.cmake.cmaketoolchain:generator=Ninja Multi-Config
tools.env.virtualenv:powershell=False

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
```

### Steps to reproduce

conan install --requires ffmpeg/6.1 --build=missing -r conancenter

### Logs

<details><summary>Click to expand build log</summary>

```
======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=23
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[tool_requires]
*: ninja/1.11.1, cmake/3.28.1
[conf]
tools.cmake.cmaketoolchain:generator=Ninja Multi-Config
tools.env.virtualenv:powershell=False

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows


======== Computing dependency graph ========
Graph root
    cli
Requirements
    brotli/1.1.0#d56d7bb9ca722942aba17369cb5c0519 - Cache
    bzip2/1.0.8#457c272f7da34cb9c67456dd217d36c4 - Cache
[...]

Issue #21694, opened 2023-12-09
[package] ffmpeg/6.1: x264 not found using pkg-config


0000mz:
### Description

When building `ffmpeg/6.1` using MSVC, ffmpeg fails to find the dependencies during the configure step. This is because the pkg-config path is not being updated to point to the generator's folder where the `*.pc` files for all the dependencies have been copied to.

A simple fix to this would be to set the `PKG_CONFIG_PATH` variable to include the generator's folder path. I have tested that this fixes the issue with this addition: https://github.com/0000mz/conan-center-index/commit/277d22fe5b3062d66371ab8211033f60bfd438c0

### Package and Environment Details

* Package Name/Version: **ffmpeg/ 6.1**
* Operating System+version: **Microsoft Windows 10 Pro 10.0.19045 N/A Build 19045**
* Compiler+version: **MSVC 193**
* Conan version: **conan 2.0.14**
* Python version: **Python 3.10.11**

### Conan profile

[settings]
arch=x86_64
build_type=Debug
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.version=193
os=Windows

### Steps to reproduce

```conanfile.txt
[requires]
ffmpeg/6.0

[generators]
CMakeDeps
CMakeToolchain
```

```profile.txt
[settings]
arch=x86_64
build_type=Debug
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.version=193
os=Windows
```

`conan install . --profile profile.txt --build=missing -c tools.cmake.cmaketoolchain:generator="Visual Studio 17 2022"`

### Logs

<details><summary>Click to expand log</summary>

```
conan install . --profile profile.txt --build=missing -c tools.cmake.cmaketoolchain:generator="Visual Studio 17 2022"

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Debug
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Debug
compiler.version=193
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Visual Studio 17 2022

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=17
compiler.libcxx=libstdc++
compiler.version=12
os=Windows
[conf]
tools.microsoft.bash:path=c:\Program Files\Git\bin\bash.exe
tools.microsoft.bash:subsystem=msys2


[...]

PR #23368, opened 2024-04-04
ffmpeg: PKG_CONFIG_PATH not set in case of build with msvc

elvisdukaj:
Specify library name and version:  **ffmpeg**

This PR tries to solve the issue(s) in:

- https://github.com/conan-io/conan-center-index/issues/23309
- https://github.com/conan-io/conan-center-index/issues/21694

In particular when build with `msvc` the ffmpeg recipe has a branch on using `AutotoolsDeps` as they actually are broken when the compiler is `msvc`. A better solution would probably be fix the conan `AutotoolsDeps` to better support `msvc` instead of the many workarounds we have. Indeed other packages would benefit from this (`coin-utils`, `ffmpeg`, `gf-complete`,  `libgettext`, `ncurses`, `xapian-core`).

Please note that simply `prepend_path` doesn't work if the `PKG_CONFIG_PATH` was empty before, see [issue 21694](https://github.com/conan-io/conan-center-index/issues/21694) for the detailed reason

---


elvisdukaj:
If https://github.com/conan-io/conan/pull/16016 is approved this PR is not needed.

elvisdukaj:
After deeper investigation the issue wasn't in https://github.com/conan-io/conan/pull/16016 but is in `pkgconfig` that is unable to split the environment variable correctly on msys2. As a patch we can override the PKG_CONFIG_PATH environment variable to don't contain multiple paths.

elvisdukaj:
Building ffmpeg with the `pkgconf` package coming from `msys2` works out of the box. I modified `msys2` to don't remove `pkgcong` and removed the `build_requirment` from msys2.

The problem is with `pkgconf` on Windows and how paths are handled.

I think having `pkgconf` built on Windows with Visual C++ (no msys2) and used from `msys2` environment cannot work.

paulharris:
Any progress on this?   I'm still using eric's original patched version of this...

elvisdukaj:
I made a lot of trouble shooting with this https://github.com/conan-io/conan/pull/1601. I had the feeling that the `PKG_CONFIG_PATH` was not properly set. Indeed, trying to print the environment variable from the conanfile, this was empty. But I failed to reproduce this situation with a test.
I stopped there, but I think it's worth continuing investigating.

jcar87:
We're unable to reproduce this issue, more details here: https://github.com/conan-io/conan-center-index/issues/23309#issuecomment-2213803996

We're trying to understand what is failing where, but so far can't get the configure step to fail on windows with msvc.

elvisdukaj:
@jcar87 I added all my observation here: https://github.com/conan-io/conan-center-index/issues/23309

PR #23369, opened 2024-04-04
alembic: Downgrade imath dependency to 3.1.9 to avoid conflitcs

RubenRBS:
Specify library name and version:  **alembic/all**

Pushes back a patch version to get in line with the rest of the imath requirements for now - evetually we'll bump them at once :)


valgur (approved):
Thanks!

czoido (approved):

uilianries (approved):

PR #23371, opened 2024-04-04
commata: include <set> in 0.2.9

toge:
Specify library name and version:  **commata/0.2.9**

There is a compilation error in commata/0.2.9 on gcc 13.

---


RubenRBS (commented):
You even sent a PR upstream, that's great! Thanks a lot!! Let's see what they have to say, but I expect the PR to pass - will approve here when so :)

RubenRBS:
Feel free to add other missing headers that upstream has mentioned now that we're on it :)

uilianries (approved):
LGTM. Thank you for the fix in the upstream! ❤️

PR #23374, opened 2024-04-04
zenoh-c: new recipe

valgur:
Adds the C API for Zenoh.

Zenoh is a pub/sub/query protocol unifying data in motion, data at rest and computations. It elegantly blends traditional pub/sub with geo distributed storage, queries and computations, while retaining a level of time and space efficiency that is well beyond any of the mainstream stacks.

https://zenoh.io/
https://github.com/eclipse-zenoh/zenoh-c

Requires #20991 for `rust/1.76.0`, since the internals of the library are implemented in Rust. All Rust dependencies and the Rust standard library are linked statically into a single static or shared library, so no dependencies on Rust remain for the library consumer.


valgur:
@jcar87 @memsharded @uilianries Can we please add support for Rust purely as a build tool on CCI via #20991, please? It's fundamentally no different from `zulu-openjdk` or other compiler toolchains as long as the projects take care to statically link all of the Rust dependencies for the exported C/C++-ABI libraries, as is done by the project in this PR (and is also the default for Rust in general).

oteffahi:
@valgur Ideally it is better to use Rust 1.22 to build zenoh-c, in order to have it support the pre-built plugins from the 0.10.1-rc releases. I notice that 1.76 was recently added to #20991, so maybe we can add 1.72 too?

oteffahi (commented):
Ideally it is better to use Rust 1.72 to build zenoh-c, in order to have it support the pre-built plugins/backends from [the 0.10.1-rc releases](https://download.eclipse.org/justj/?file=zenoh). I notice that 1.76 was recently added to #20991, so maybe we can add 1.72 too?

PR #23376, opened 2024-04-05
libcurl: Add 8.7.1

### LINKED ISSUE #23295: [request] libcurl 8.7.1
### What is your problem/feature request?

https://curl.se/download/curl-8.7.1.tar.gz

Ahajha:
Specify library name and version:  **libcurl/8.7.1**

Resolves #23295 

---


RubenRBS (commented):
I see that from our latest version, there's the new Cmake `USE_OPENSSL_QUIC`, and that `ENABLE_CURL_MANUAL` has changed default to true.
I think it'd be nice to ensure that `ENABLE_CURL_MANUAL` is set to false if we're remocing the docs (Not that I think they could get built if they seem to require Perl).

Does it make sense to have the new option also available? wdyt? Thanks!

jcar87:
> USE_OPENSSL_QUIC

I'd leave it as default to off outside of this PR - it requires an additional dependency which we don't currently have (`nghttp3`), and requires handling the same logic when invoking both the `./configure` and CMake. Orthogonal to this recipe unless these capabilities were introduced in the newest version.


> ENABLE_CURL_MANUAL

Setting this to OFF unconditionally - matches what we do for makefiles which is disable the manual/docs altogether

RubenRBS (approved):
Sorry for the delay on this one, it fell thru the cracks :)

jcar87 (approved):

perseoGI:
I'm closing this PR in favor of https://github.com/conan-io/conan-center-index/pull/24068 which adds `libcurl 8.8.0`.
Thanks for your contribution @Ahajha!

PR #23384, opened 2024-04-05
dbus: don't propagate X11 libs, drop `with_glib`, update license info

### LINKED ISSUE #22133: [bug] dbus: generated dbus-1.pc incorrectly has Requires: xorg

valgur:
I dropped (deprecated) the `with_glib` option altogether since it's only used within tests and nowhere else. I guess this might have been intended to enable [`dbus-glib`](https://gitlab.freedesktop.org/dbus/dbus-glib), which has been deprecated and moved to a separate repo a long while ago.

Set the X11 dependency to `visible=False`. As noted by the recipe comment, this is not ideal but it's the best that is currently possible with the tools provided by Conan. 

I also made the licensing details a bit more accurate, since the mix of licenses covering the project is actually quite complex:
https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/COPYING
https://gitlab.freedesktop.org/dbus/dbus/-/tree/master/LICENSES

Also bumped `libselinux` to avoid a conflict with `libsystemd` and removed `with_selinux` option on non-Linux OS-s.

- Fixes #22133
- Resolves #22134

### DIFF
--- a/recipes/dbus/1.x.x/conanfile.py
+++ b/recipes/dbus/1.x.x/conanfile.py
@@ -52,8 +52,8 @@ class DbusConan(ConanFile):
     def config_options(self):
         if self.settings.os not in ("Linux", "FreeBSD"):
             del self.options.with_systemd
-        if self.settings.os not in ("Linux", "FreeBSD"):
             del self.options.with_x11
+            del self.options.with_selinux
         if self.settings.os == "Windows":
             del self.options.fPIC
 
@@ -72,8 +72,8 @@ class DbusConan(ConanFile):
             self.requires("glib/2.78.3")
         if self.options.get_safe("with_systemd"):
             self.requires("libsystemd/253.6")
-        if self.options.with_selinux:
-            self.requires("libselinux/3.3")
+        if self.options.get_safe("with_selinux"):
+            self.requires("libselinux/3.5")
         if self.options.get_safe("with_x11"):
             self.requires("xorg/system")
 
@@ -82,7 +82,7 @@ class DbusConan(ConanFile):
             raise ConanInvalidConfiguration(f"{self.ref} requires at least gcc 7.")
 
     def build_requirements(self):
-        self.tool_requires("meson/1.3.2")
+        self.tool_requires("meson/1.4.0")
         if not self.conf.get("tools.gnu:pkg_config",check_type=str):
             self.tool_requires("pkgconf/2.1.0")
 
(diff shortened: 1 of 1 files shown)


recipes/dbus/1.x.x/conanfile.py
```diff
@@ -109,13 +109,18 @@ def generate(self):
             tc.project_options["launchd_agent_dir"] = os.path.join(self.package_folder, "res", "LaunchAgents")
         tc.project_options["x11_autolaunch"] = "enabled" if self.options.get_safe("with_x11", False) else "disabled"
         tc.project_options["xml_docs"] = "disabled"
+        tc.project_options["modular_tests"] = "enabled" if self.options.with_glib else "disabled" # glib is not found otherwise due to a buggy build.meson
```
jwillikers, line 112: Can we get this fixed upstream?

RubenRBS: Agree with @jwillikers, we should at leats let upstream know :)

RubenRBS (approved):
Thanks! Approving, but let's try to open an issue upstream for the buggy meson file!

RubenRBS (changes requested):
nevermind, I think that also addresing https://github.com/conan-io/conan-center-index/issues/22133 would be good to have in the same if you can :)

valgur:
I dropped (deprecated) the `with_glib` option altogether since it's only used within tests and nowhere else.

Set the X11 dependency to `visible=False`. As noted by the recipe comment, this is not ideal but it's the best that is currently possible with the tools provided by Conan. 

I also made the licensing details a bit more accurate, since the mix of licenses covering the project is actually quite complex:
https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/COPYING
https://gitlab.freedesktop.org/dbus/dbus/-/tree/master/LICENSES

jwillikers (approved):
Thanks @valgur!

jwillikers (approved):

PR #23356, opened 2024-04-03
aerospike-client-c: new recipe

ybogo:
Specify library name and version:  **aerospike-client-c/6.6.0**

I am not an author of this package but I moving my pet project to use conan and I am using aerospike client in it. So I decided to create a recipe to manage all my dependencies with only one tool.

Closes https://github.com/conan-io/conan-center-index/issues/23243

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #23388, opened 2024-04-05
aersopike-common: new recipe

ybogo:
Specify library name and version:  **aerospike-common/0009245**

This library is only needed for aerospike-client-c (PR already created https://github.com/conan-io/conan-center-index/pull/23356) but in root project it is used as a submodule thats why it needs it's own recipe.

Also there is no any version of this library officially exist (no tags, no branches, etc). That why I used commit hash a version.

---


recipes/aerospike-common/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "0009245":
```
RubenRBS, line 2: For these cases, we usually use `cci.YYYYMMDD` with the date of the picked commit

RubenRBS (commented):
Thanks! you mention that this was only used internally from another specific library, makes me wonder if vendorizing it in that one would be easier for you, but happy to have this availale in CCI too! Let us know if you need help fixing the issues detected in CI :)

ybogo:
> Thanks! you mention that this was only used internally from another specific library, makes me wonder if vendorizing it in that one would be easier for you, but happy to have this availale in CCI too! Let us know if you need help fixing the issues detected in CI :)

Actually I don't want to create separate packages for each library from aerospike client but as I said in root project it is used as a submodule. As far as I know git is not allowed to pull sources https://github.com/conan-io/conan-center-index/pull/23356#discussion_r1551084529 . So what is the most correct way to create a recipe for repo with submodules?

As for CI errors I have I question how to disable windows support? This package use pure `make` as a build tool which is not supported for windows. I raised an `ConanException` in `validate(self)` method but CI treat this as an error

And last but not least problem with cross-platform build. I implemented my own function which generate arch falgs for different compilers and it works nice. But I am getting an error during linakage of test package. Somehow tests package still building for host arch and not for specified in in build settings. Take a look at failed job [logs](https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23388/9-macos-clang/aerospike-common/cci.20240124//a136f562e464712e44a28cfce3a67b3fa0c6bb34-test.txt). I managed to reproduce it locally with `VERBOSE=1`:
```
/Library/Developer/CommandLineTools/usr/bin/c++  -isystem /Users/yurybogomolov/.conan2/p/b/aeros5cd64c47a6652/p/include -stdlib=libc++ -O3 -DNDEBUG -arch arm64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -std=gnu++17 -MD -MT CMakeFiles/test_package.dir/main.cpp.o -MF CMakeFiles/test_package.dir/main.cpp.o.d -o CMakeFiles/test_package.dir/main.cpp.o -c /Users/yurybogomolov/develop/conan-center-index/recipes/aerospike-common/all/test_package/main.cpp
```

PR #18836, opened 2023-07-22
sdl_mixer/all: migrate to Conan v2

valgur:
Linking in test_package currently fails for GCC due to #12826.

PR #23391, opened 2024-04-05
sdl_mixer: Validate out invalid requirement options

RubenRBS:
Specify library name and version:  sdl_mixer/all

Prettier error messages


czoido (approved):

valgur:
Probably needs to be migrated to v2 before these changes can be merged (#18836).

RubenRBS:
Afaik this should only build the new versions, those are the ones changing recipe revision, but we'll see. I've started reviewing your linked PR too either way, thanks for the ping :)

Issue #23151, opened 2024-03-18
[package] CPython: Misc recipe improvements


Ahajha:
I'm making this issue to serve as a TODO list of things that I believe need to be done in the CPython recipe. Currently waiting on the 2.0 migration and new version PRs to be reviewed and merged, so I don't plan to start on these just yet, but I do plan on working on them soon:

- [ ] Emulate CMake's built-in `find_package(Python)`: Currently this is just `find_package(cpython)`, but we need to manually pass in a bunch of extra variables to make sure we find the Conan one. We should be able to use `find_package(Python)` without this extra setup. We would need to include Python_add_library, which is implemented in CMake here: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPython/Support.cmake?ref_type=heads#L4016. Potentially we could have this as an option.

- [ ] Check which options are actually usable: Many of the options (like with_lzma) imply that there is a way to not include certain dependencies/packages, but I suspect with recent changes that many of these no longer work (or maybe never did). For each of these we should determine if it is actually possible to implement it properly, and remove it if it is either impossible or highly unfeasible.

- [ ] Add 3.13 (once released): I expect this to not have too many breaking changes regarding the build system, but I at least know that some standard library modules have been removed: https://peps.python.org/pep-0594/. The main ones of note are `nis` and `crypt`, which I believe are the only two that would affect the recipe (mainly just disabling things in 3.13+)

- [ ] Enabling the `optimizations` and `lto` options, and ideally turn it on by default, at least in Release mode: This would ensure the Conan Python packages are on par speed-wise with the official binaries.

- [ ] Setting shared by default: Windows static mode has not worked since about 3.10, and seems to only be supported on a best-effort basis (https://github.com/python/cpython/issues/110234). Linux/Mac static builds work and are fully functional, but on all OSs shared seems to be the preferred distribution method. For Windows specifically, it's a bit annoying for the default configuration to be invalid.

Ahajha:
Regarding conf info and environment variables:

Currently, we have the following defined:
Env: PATH, PYTHON, PYTHONHOME, PYTHON_ROOT
Conf: user.cpython:python, user.cpython:pythonhome, user.cpython:module_requires_pythonhome, user.cpython:python_root

Env:
- PATH: fine
- PYTHON: I can't find any docs on this, inclined to remove it if we don't know its purpose. My guess is that it was just used to mirror the `python` conf variable.
[...]

PR #23394, opened 2024-04-06
cpython: Add native support for CMake's builtin FindPython(3)

Ahajha:

With these new changes, consumers should be able to use `find_package(Python)` or `find_package(Python3)` without any additional setup. See notes about CMake and FindPython here: #23151

I've removed the tests for the old `find_package(PythonInterp)` and `find_package(PythonLibs)`. Nothing's preventing them from working, but I figure most people at this point would be interested in the new versions, so this is just to clean up the test package (these are still tested in the test_v1_package, along with the "manual" `find_package(Python3)` integration, since I haven't added support for the v1 CMake genererators).

Going to wait for the new versions to be merged before taking this out of draft.

---


recipes/cpython/all/conanfile.py
```diff
@@ -512,6 +512,70 @@ def _msvc_package_copy(self):
...
+
+    def _write_cmake_findpython_wrapper_file(self):
+        template = textwrap.dedent("""
+        if (DEFINED Python3_VERSION_STRING)
+            set(_CONAN_PYTHON_SUFFIX "3")
+        else()
+            set(_CONAN_PYTHON_SUFFIX "")
+        endif()
+        set(Python${{_CONAN_PYTHON_SUFFIX}}_EXECUTABLE {})
+        set(Python${{_CONAN_PYTHON_SUFFIX}}_LIBRARY {})
```
valgur, line 540: Using the standard Python formatter with this template is a bit of an overkill and bug-prone due to the numerous braces.
I would simplify the template and use
```py
content = template.replace("@PYTHON_EXE@", python_exe).replace("@PYTHON_LIBRARY@", python_library)
```

recipes/cpython/all/conanfile.py
```diff
@@ -512,6 +512,70 @@ def _msvc_package_copy(self):
         lib_dir_path = os.path.join(self.package_folder, self._msvc_install_subprefix, "Lib").replace("\\", "/")
         self.run(f"{interpreter_path} -c \"import compileall; compileall.compile_dir('{lib_dir_path}')\"")
 
+    def _exact_lib_name(self, folder):
+        possible_extensions = ("a", "so", "dylib", "lib")
+        for file in os.listdir(folder):
+            for extension in possible_extensions:
+                if re.match(f".*\\.{extension}", file):
+                    return file
```
valgur, line 520:
```suggestion
        patterns = ["*.a", "*.so*", "*.dylib", "*.lib"]
        for pattern in patterns:
            for file in Path(folder).glob(pattern):
                return file.name
```
Would be slightly nicer, imo. Needs a `from pathlib import Path` as well.

recipes/cpython/all/conanfile.py
```diff
@@ -512,6 +512,70 @@ def _msvc_package_copy(self):
...
+        
+        include(${{CMAKE_ROOT}}/Modules/FindPython${{_CONAN_PYTHON_SUFFIX}}.cmake)
+                                   
+        # Sanity check: The former comes from FindPython(3), the latter comes from the injected find module
+        if(NOT Python${{_CONAN_PYTHON_SUFFIX}}_VERSION STREQUAL Python${{_CONAN_PYTHON_SUFFIX}}_VERSION_STRING)
+            message(FATAL_ERROR "CMake detected wrong cpython version - this is likely a bug with the cpython Conan package")
+        endif()
+
+        if (TARGET Python${{_CONAN_PYTHON_SUFFIX}}::Module)
+            target_link_libraries(Python${{_CONAN_PYTHON_SUFFIX}}::Module INTERFACE cpython::python)
```
valgur, line 554: This fails for some versions of CMake. Use this pattern instead:
```cmake
set_target_properties(<target> PROPERTIES INTERFACE_LINK_LIBRARIES <libs>)
```

recipes/cpython/all/conanfile.py
```diff
@@ -647,6 +712,12 @@ def package_info(self):
         )
         self.cpp_info.components["embed"].requires = ["python"]
 
+        # Transparent integration with CMake's FindPython(3)
+        self.cpp_info.set_property("cmake_file_name", "Python3")
+        self.cpp_info.set_property("cmake_module_file_name", "Python")
```
valgur, line 717: Too bad there's no way to create file name aliases in Conan.
This will break for `find_package(Python3 MODULE)`, but it's rarely used that way, fortunately.

recipes/cpython/all/conanfile.py
valgur: Could maybe avoid the globbing altogether?
```py
def _exact_lib_name(self):
        return self._lib_name if self.options.shared else f"{self._lib_name}.a"
    if is_apple_os(self):
        return f"lib{self._lib_name}.dylib" if self.options.shared else f"lib{self._lib_name}.a"
    return f"lib{self._lib_name}.so" if self.options.shared else f"lib{self._lib_name}.a"
```

recipes/cpython/all/conanfile.py
Ahajha: Yea, I think I like that better. In my initial prototypes I didn't have that many braces, and then I realized I had to handle the difference between FindPython/FindPython3.

recipes/cpython/all/conanfile.py
Ahajha: My original approach was to avoid the pattern matching, but I was concerned if there were any odd edge cases, for example using MinGW. Though I don't know if MinGW would even work for this recipe at all. Also I think on Windows it's a `.lib` regardless, but I'd have to double check.

recipes/cpython/all/conanfile.py
Ahajha: Interesting, can do. What version of CMake would that fail for, out of curiosity?

recipes/cpython/all/conanfile.py
Ahajha: Yea, that would be useful here. I saw Freetype used a similar pattern so that's how I did it.

I'm not exactly sure what `MODULE` mode does differently anyways, or if FindPython(3) even supports it.

valgur: `MODULE` simply excludes `*-config.cmake`/`*Config.cmake` and only allows `Find*.cmake` to be used by `find_package()`.

Ahajha: Ah, yea that's probably rare enough to not worry about. I could also swap them if for whatever reason we think `find_package(Python3 MODULE)` would be more common than `find_package(Python MODULE)`.

recipes/cpython/all/conanfile.py
valgur: Not sure which ones are affected exactly. I've had issues with v3.28.

Ahajha: Odd, I thought I was testing with the latest (3.29). I'll test with 3.28 and see what happens.

Ahajha:
@valgur Since you're reviewing, what would you think about having an option to opt-out of this behavior? We could install the wrapper module regardless, and then the option would just affect the package info, and not change the ID.

valgur:
@Ahajha That's not a bad idea, but I can't think of a reason why someone would include the cpython package in requires, but would not want it to be picked up by `find_package(Python)`. I would keep it simple for now and not include it unless someone complains. But either way would be fine by me.

recipes/cpython/all/conanfile.py
valgur: @Ahajha I did not actually encounter any issues with `target_link_libraries` for this recipe. It's more of a best practices / FYI suggestion.

Ahajha:
@valgur My only concern is that perhaps it would break someone's workflow. But it *likely* won't, so yea I think we can leave it for now.

recipes/cpython/all/conanfile.py
Ahajha: Hadn't heard that, I didn't even know there was a different way to link libraries in this way. I'll have to do some reading.

recipes/cpython/all/conanfile.py
Ahajha: I do kind of like the non-globbed approach here, we can always fix if an issue comes up. I've mashed some of these implementations together, will push shortly.

recipes/cpython/all/conanfile.py
Ahajha: Can't find much documentation on it, but it looks like Conan does it internally, so I'll add your suggestion once I test.

Ahajha:
Want to see if I can get `WITH_SOABI` working, I haven't used it before so I don't know exactly what to expect, but what I'm seeing is almost certainly wrong. On Windows I'm seeing `.pyd.pyd` files, for example.

recipes/cpython/all/conanfile.py
Ahajha: New implementation pushed

recipes/cpython/all/conanfile.py
Ahajha: New implementation pushed

recipes/cpython/all/conanfile.py
Ahajha: Done

valgur (approved):
Thanks!

test_package looks so much saner as well now. :)

Ahajha:
@valgur I'm going to see if I can get WITH_SOABI working - hopefully can figure that out by tomorrow (or at least by when my other PR gets merged).

Ahajha:
@valgur Seems that there's a bug in FindPython regarding Windows debug builds, I've reported it here: https://gitlab.kitware.com/cmake/cmake/-/issues/25874. It's possible the difficulty in auto-detection is related, I might make a separate issue for that.

Just going to fix the lint warning and then this should be go (aside from waiting for the other PR to be merged).

recipes/cpython/all/conandata.yml
```diff
@@ -28,7 +28,7 @@ patches:
     - patch_file: "patches/3.12/3.12.1-0002-remove-module-deps.patch"
       patch_description: "Remove section of solution file forcing projects to be built that might not be used for this recipe"
       patch_type: "conan"
-  "3.11.8":
+  "3.11.9":
```
Ahajha, line 31: @ericLemanissier FYI

recipes/cpython/all/conanfile.py
```diff
@@ -648,6 +648,72 @@ def _msvc_package_copy(self):
...
+        # In order for the package to be relocatable, these variables must be relative to the installed CMake file
+        if is_msvc(self):
+            python_exe = "${CMAKE_CURRENT_LIST_DIR}/../../" + self._cpython_interpreter_name
+            python_library = "${CMAKE_CURRENT_LIST_DIR}/../" + self._exact_lib_name
+        else:
+            python_exe = "${CMAKE_CURRENT_LIST_DIR}/../../bin/" + self._cpython_interpreter_name
+            python_library = "${CMAKE_CURRENT_LIST_DIR}/../" + self._exact_lib_name
+
+        cmake_file = os.path.join(self.package_folder, self._cmake_module_path, "use_conan_python.cmake")
+        content = template.replace("@PYTHON_EXECUTABLE@", python_exe).replace("@PYTHON_LIBRARY@", python_library)
```
jwillikers, line 714: It might be possible to use an f-string here.

jwillikers (approved):

recipes/cpython/all/conanfile.py
Ahajha: I originally had it as an fstring, @valgur suggested this change since  it has a ton of escaped {}s.

jwillikers: Oh yeah, that makes sense.

danimtb (approved):

jwillikers (approved):

PR #23404, opened 2024-04-06
llvm-core: unpin cmake version

pierricgimmig:
In llvm-core, the cmake version was pinned to 3.20.5, making it impossible to build with VS 2022 on Windows since VS 2022 generators were introduced with cmake 3.21. From the original comment, it seems the intention was not to pin cmake, but rather specify a minimum version.

This change simply transforms the cmake pinning to a mere minimum requirement, making it possible to compile llvm-core with toolchains born after cmake 3.20.5.

Specify library name and version:  **llvm-core/all**

---

### DIFF
--- a/recipes/llvm-core/all/conanfile.py
+++ b/recipes/llvm-core/all/conanfile.py
@@ -77,7 +77,7 @@ class LLVMCoreConan(ConanFile):
     # Older cmake versions may have issues generating the graphviz output used
     # to model the components
     build_requires = [
-        'cmake/3.20.5'
+        'cmake/[>=3.20.5]'
     ]
 
     generators = 'cmake', 'cmake_find_package'
--- a/recipes/llvm-core/all/conanfile.py
+++ b/recipes/llvm-core/all/conanfile.py
@@ -77,7 +77,7 @@ class LLVMCoreConan(ConanFile):
     # Older cmake versions may have issues generating the graphviz output used
     # to model the components
     build_requires = [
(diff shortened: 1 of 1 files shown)


recipes/llvm-core/all/conanfile.py
```diff
@@ -77,7 +77,7 @@ class LLVMCoreConan(ConanFile):
     # Older cmake versions may have issues generating the graphviz output used
     # to model the components
     build_requires = [
-        'cmake/3.20.5'
+        'cmake/[>=3.20.5]'
```
RubenRBS, line 80:
```suggestion
        'cmake/[>=3.20.5 <4]'
```

Just in case :)

RubenRBS (dismissed):
Hi! Thanks a lot for taking the time to fix this issue, we really appreciate it

Please remember to request access by writing in https://github.com/conan-io/conan-center-index/issues/4 before we can merge this PR, after that the PR will be picked automatically once we leave maintenance mode in the CI systems, so no need for further actions on your part :)

pierricgimmig:
Hi @RubenRBS , do I need to do somehting more on my end? Thanks!

jcar87:
Thanks @pierricgimmig  and apologies for the delay on this.
I believe this no longer applies, as the graphviz output is no longer used so there is no strict requirement on any cmake version now: https://github.com/conan-io/conan-center-index/blob/master/recipes/llvm-core/all/conanfile.py

Closing this as no longer needed - but please let us know if there are any issues.

PR #23411, opened 2024-04-07
nmos-cpp: component bindirs

garethsb:
Specify library name and version:  **nmos-cpp**

Add components for _nmos-cpp-node_ and _nmos-cpp-registry_ with `bindirs`. This fixes the bug that these executables were not available in the path on Windows when running:
```sh
conan install --tool-requires=nmos-cpp/cci.20240223
conanbuild.bat
```

Also remove direct dependency on **zlib** - I believe originally added only to ensure no version conflicts between the indirect dependencies via **boost** and **cpprestsdk** at the time CCI policy was to use specific versions and aggressively update recipes to latest dependency versions.

---

### DIFF
--- a/recipes/nmos-cpp/all/conanfile.py
+++ b/recipes/nmos-cpp/all/conanfile.py
@@ -57,7 +57,6 @@ class NmosCppConan(ConanFile):
         self.requires("openssl/[>=1.1 <4]")
         self.requires("json-schema-validator/2.3.0")
         self.requires("nlohmann_json/3.11.3")
-        self.requires("zlib/[>=1.2.11 <2]")
         if Version(self.version) >= "cci.20240222":
             self.requires("jwt-cpp/0.7.0")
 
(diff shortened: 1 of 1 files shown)


garethsb:
> Also remove direct dependency on **zlib** - I believe originally added only to ensure no version conflicts between the indirect dependencies via **boost** and **cpprestsdk** at the time CCI policy was to use specific versions and aggressively update recipes to latest dependency versions.

Would it be better to keep it, as an `override=True` requirement?

uilianries:
> Would it be better to keep it, as an override=True requirement?

@garethsb Please, don't. I know it fixes several problems with version conflict, but the side-effect is linking to a different version, which may result in abi incompatibility. You still can say, "but this specific version has no conflicts and works fine", but then we will need to consider any package in ConanCenterIndex. 

The idea is using more range versions in ConanCenterIndex, we need to check the backward compatibility for projects, but in small steps we are updating. As second step, accepting a PR with multiple recipes, so we can fix all together, it's a CI feature, but is under development.

uilianries (approved):
LGTM. Tested locally and looks good.

jwillikers (approved):

PR #23427, opened 2024-04-09
Remove binutils unused ``conf_info`` entries

memsharded:
Specify library name and version:  **binutils**

The binutils recipe had 2 ``conf_info`` entries that are "options", not basic Python types, and this is broken, it crashes for example when using ``--format=json``. Furthermore, those are not used at all in all ConanCenter, so their existence seems not necessary. The values of those options can be read by consumers via ``self.dependencies["dep"].options``.

Close https://github.com/conan-io/conan/issues/16018

---


RubenRBS (approved):

uilianries (approved):
LGTM

PR #23428, opened 2024-04-09
libjpeg: add missing stdlib includes in the public header

valgur:
`libjpeg` is apparently missing `#include <stddef.h>` and `#include <stdio.h>` in its exported `jpeglib.h` main public header.

### DIFF
--- a/recipes/libjpeg/all/test_package/test_package.c
+++ b/recipes/libjpeg/all/test_package/test_package.c
@@ -12,12 +12,12 @@
  * JPEG markers.
  */
 
+#define JPEG_CJPEG_DJPEG /* to get the command-line config symbols */
+#include <jpeglib.h>    /* get auto-config symbols, <stdio.h> */
+
 #include <stdio.h>
 #include <stddef.h>
 #include <stdlib.h>
-
-#define JPEG_CJPEG_DJPEG /* to get the command-line config symbols */
-#include "jpeglib.h"    /* get auto-config symbols, <stdio.h> */
 
 #ifdef HAVE_LOCALE_H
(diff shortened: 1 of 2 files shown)


valgur:
Simply re-ordered the includes in test_package to not include anything else before `#include <jpeglib.h>` to demonstrate the issue.

valgur:
Huh. Apparently it's by design: https://github.com/libjpeg-turbo/libjpeg-turbo/issues/17#issuecomment-139620135
I suppose we should keep it that way then.

jcar87:
Hi @valgur - thanks for opening this PR. 

A couple notes:
* In general for all PRs that claim to fix an issue, please add details about the issue - on which systems does it fail, what is the error?  One of us from the reviewing team will make every effort to replicate the issue and validate the fix. This also leaves traceability for the future
* Thanks for checking that this was an issue reported upstream: https://github.com/libjpeg-turbo/libjpeg-turbo/issues/17#issuecomment-139620135

on the other hand, i can see this is a very complex test package that really doesn't add much. The typical test package, unless there's a strong justification, should just be limited to including a public header and calling a function where the symbol is provided by the library (as opposed to a function that is self-contained in a header). ANY such function would do - even if it the expected results in an error.  So I would strongly encourage to simplify this test package as much as possible.

valgur:
@jcar87 Sorry, I did not see your message before closing.

As I mentioned in the previous comment, it was currently just a draft PR without an actual fix just do demonstrate that using libjpeg without a preceding `#include <stdio.h>` does not work. I could have included more details, though, I agree. The build fails due to undefined `size_t` and `FILE` without the include.

This would normally be considered a bug in an upstream project, but it's apparently intentional and should be kept that way for consistency.

PR #20374, opened 2023-10-05
Package/implot add implot demo.cpp

FranzPoize:
Specify library name and version:  **implot/0.15**

Adds implot_demo.cpp into the recipe so that user can use the main documentation of implot

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #23434, opened 2024-04-09
implot: bump imgui for consistency with other packages

valgur:
Also adds support for v1.89.7+ for v0.14 by renaming a single imgui constant in a .cpp file.

`IMGUI_DEFINE_MATH_OPERATORS` no longer needs to be set in CMake. Dropped it to get rid of a related warning.

The imgui version bump is for compatibility with `imguizmo` (#23435) and `iridescence` (#23436). There are no other packages depending on `implot`.

### DIFF
--- a/recipes/implot/all/CMakeLists.txt
+++ b/recipes/implot/all/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.15)
 project(implot LANGUAGES CXX)
 
 file(GLOB HEADER_FILES ${IMPLOT_SRC_DIR}/*.h)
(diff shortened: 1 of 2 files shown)


recipes/implot/all/CMakeLists.txt
```diff
@@ -1,20 +1,18 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.15)
 project(implot LANGUAGES CXX)
 
 file(GLOB HEADER_FILES ${IMPLOT_SRC_DIR}/*.h)
 
 add_library(${PROJECT_NAME}
     ${IMPLOT_SRC_DIR}/implot.cpp
     ${IMPLOT_SRC_DIR}/implot_items.cpp
-    ${IMPLOT_SRC_DIR}/implot_demo.cpp
```
RubenRBS, line 9: We added this last year because someone requested it in https://github.com/conan-io/conan-center-index/pull/20374

Honestly I don't mind both options, but if it's useful? Don't know. Do you have any extra insight I might be missing? Thanks!

valgur: Huh, that seems like a rather niche use case. None of the symbols in `implot_demo.cpp` are even declared in the public header, I think. I would have pushed back on the previous PR and directed them to convince the upstream project to refactor that functionality into the library itself, really. Nevertheless, it does not actively hurt either and I'm ok with keeping it intact. I'll simply include a comment this time to prevent any future head scratches.

valgur: Reverted.

RubenRBS (approved):

valgur:
@jwillikers @ericLemanissier @toge a community review would be welcome here

ericLemanissier (approved):

PR #23437, opened 2024-04-09
dbus: Use relative folders for directories

jwillikers:
Use relative folders due to this change: https://github.com/conan-io/conan/pull/15706. Remove the useless attempt to modify the separators used for the prefix path. The separators are still inconsistent for the prefix and exec_prefix on Windows. They use both backslashes and forward slashes.


RubenRBS (approved):
Thanks!

franramirez688 (approved):

Issue #641, opened 2020-01-10
[question] Approach for system requirements


danimtb:
There are some recipes that may need to install tools that can be considered as system requirements.

There are already some recipes in the index that check for system tools installed but do not install them. Others like `libusb` do install system requirements as `libudev`.

We feel it is intrusive and risky to install system requirements when using a Conan dependency and we want to avoid this in Conan (specially asking for sudo permissions 😩).

The question here is what approach should be taken when packaging recipes that require any kind os system requirement:

- Should we create recipes for system requirements and package binaries?
- Should we just check the existence of that tools installed and just raise?
- Should we create "proxy recipes" that perform those install or checks to model that dependency in the conan graph?

Feecback please!

SSE4:
for the reference, I am copying here problems/disadvantages/limitations caused by system requirements we previously faced on bincrafters:

---------------------------------------------------------------------------------------------------------

System Requirements in conan known to cause lots of various issues, they are hard to maintain, and don't scale well. some problems with System Requirements:

* they require running system package manager from sudo, which is not a desired in many cases. they pollute build machine, they require user interaction, and they require permissions which developer might not have at all.
* system requirements don't support cross-building at all, especially cases like cross-compiling from Windows/Mac to Linux are impossible, and even Linux -> Linux is problematic.
* due to the fragmentation, there are many major package managers to support, like yum, apt, pacman, etc. support of each new package manager requires additional baby-sitting (and possibly modifications to the conan itself, to add new distributions like Gentoo, Slackware, etc.)
* within single package manager, there are various architectures to support - and this adds an additional level of scalability problem.
* even within single package manager, various distributions may have different names of packages. for instance, Fedora, RHEL and CentOS are known to have different package names for certain packages.
* even within single distro, packages are deprecated and removed from time to time, so it requires continuous maintenance with every new release.
* the code to use system package manager in recipes looks ugly - it has lots of nested conditions.
* due to the various bugs in distributions, packages often conflict with each other, making it impossible to build a conan package on the system. for instance, on Ubuntu, 32-bit python-development package conflicts with 64-bit one.
* installing additional software from system package manager, may confuse conan build. for instance, libprotobuf-dev might be installed as a dependency of some other package, then it may override protoc executable from the conan package, which is not a desired effect.
* builds are no longer truly reproducible, because various distros have different versions of system packages, therefore behaviour may change after recompiling on another distro.
[...]

PR #23439, opened 2024-04-09
glfw: Use libglvnd on FreeBSD and Linux

jwillikers:
The libglvnd package provides the necessary opengl/system dependency. The libglvnd package is a proper, non-system Conan package.


valgur (approved):

jcar87:
Hi @jwillikers  - thank you for opening this PR.

We'd like to understand the motivation better, as well as any potential implications. As far as I'm aware, `opengl/system` already handles bringing `libglvnd` from the system.

jwillikers:
> Hi @jwillikers - thank you for opening this PR.
> 
> We'd like to understand the motivation better, as well as any potential implications. As far as I'm aware, `opengl/system` already handles bringing `libglvnd` from the system.

@jcar87 That's exactly the problem, it's a _system_ Conan package. Why use a system Conan package when a we have a proper Conan package that builds the library from source? My understanding is that system packages are primarily an escape-hatch when we don't have or want a proper Conan package, like in the case of X11. The system Conan packages have multiple problems, including that they don't support cross-compilation for a system like Yocto and that they need to be manually removed and rebuilt after a major distribution upgrade, i.e. Ubuntu 20.04 to Ubuntu 22.04. If we should instead use the `opengl/system` package, then why even have or use the `libglvnd` package in CCI?

I'm not sure that there are any major implications. We're just using a version of `libglvnd` that is built from Conan instead of the one installed on the system. We're currently using the `libglvnd` Conan package with my Mesa and Weston Conan packages that are in progress to run a Wayland compositor. This works well in practice, using the Conan `libglvnd` and `mesa` packages to provide graphics acceleration. The `libglvnd` Conan package should also work with loading system graphics drivers when used without the Mesa Conan package.

jcar87:
> My understanding is that system packages are primarily an escape-hatch when we don't have or want a proper Conan package, like in the case of X11.

This is not correct, although I can see why this can be confusing.

It is perfectly feasible to build all of X11 from source in a Conan recipe. And at "build" time, all consumer packages would build against this version perfectly fine.

The problem comes from runtime: if the version of the X11 shared libraries from Conan is different than the X server running on the system - things may not work well, especially if the Conan package is in a newer version. The applications need to talk to the server running on the host machine - the server is going to be the exact version from the system, as it is a single program running. I believe that at this stage the application (and its libraries) talk to the server via the X protocol, so we have different layers of compatibility:
* ABI compatibility between the apps/libraries we've built, and the loaded x11 libraries - here Im not too concerned, x11 is backwards compatible 
* compatibility at the X level protocol at runtime The reason there is an `xorg/system` is to avoid the case where there may be a version mismatch
* configuration: I suspect that runtimes like x11 may load configuration files from specified locations, handled by the distros - and more often than not, the compiled code is wired to

I suspect the reason we don't have x11 built from source in Conan is precisely to guarantee these two compatibilities without mismatches:
* Conan Center binaries use "older" distro versions, so that the built artifacts against older x11 can still be satisfied by _newer_ versions of the same libraries in newer distros - old school backwards abi compatibility
* Because the system libraries are always used at runtime, the clients are the server are in the same version, and the configuration is loaded where the OS expects it 

these two things make graphical GUIs work when built from Conan, as far as I understand. If these assumptions are not correct, I'd be happy to start maintaining a recipe for x11 from source, and finally solve those issues :)  - my concern is that if the recipes depend on "too new x11 runtime", there may be issues running them on older distros that people still use.

As for `libglvnd`, if I remember correctly it interfaces with both:
* x11 libraries
* system configuration: `libglvnd` queries the system to find the lower-level implementations
* lower-level implementations - that are installed at the system level, presumably in some location that the `libglvnd` implementation needs to know about

So we need to ensure that:
* `libglvnd` is compatible with the system x11 libraries and maybe the runtime too
* that our conan-provided `libglvnd` is capable of finding the vendor-specific implementations that may be installed on the system. This is critical, the search mechanism is described below:
    - https://github.com/NVIDIA/libglvnd/blob/908086d22dc307d17d0eb35c522c35fd190718cc/src/EGL/icd_enumeration.md#icd-discovery
   - https://github.com/NVIDIA/libglvnd/blob/908086d22dc307d17d0eb35c522c35fd190718cc/src/EGL/icd_enumeration.md#icd-installation
* that `libglvnd` is compatible and can interface with those vendor-specific implementations (e.g. nvidia, etc)


As it stands, because `libglvnd` has to interface with system-provided components (the x11 libraries, the system configuration folders, and the vendor-specific implementations), I am unsure of moving forward with this will actually work, and it if works, if it works across most distros.

I'm particularly concerned about the compile-time setting of the configuration directory that then the runtime searches for implementations - is it _always_ at `/usr/share/glvnd/egl_vendor.d`? do some distros or systems have this somewhere else? do some distros build their `libglvnd` set to a different location? 

That's why in some scenarios we prefer system versions rather than conan packages:
* when the libraries/apps need to interface with a system service of which there can't be multiple instances, and the version needs to be guaranteed (in particular when it comes to talking to hardware, like the graphics driver)
* when the libraries/app need have hardcoded assumptions about system configuration locations - it's safer to rely on the system libraries to find the right configurations, rather than create the potential for stuff not working at all

jcar87:
Side notes:

> The system Conan packages have multiple problems, including that they don't support cross-compilation for a system like Yocto

We probably _should_ have a versioned, built-from-source recipe for x11 to cover for cases like this. When there is not an expectation of running anything, and x11 is already _not_ provided by the system - it's perfectly sensible to provide it ourselves. This would also apply to x11 on macOS and Windows, if needed

> and that they need to be manually removed and rebuilt after a major distribution upgrade, i.e. Ubuntu 20.04 to Ubuntu 22.04

what would need to be force-rebuilt? conan packages that were built against a previous version of the x11 headers/libraries? This should not be necessary, as far as I'm aware: x11 is probably one of the most backwards compatible libraries out there, something built against an older version should run on a system that is running a newer version - I'd be happy to look into a reproducible example where this isn't the case.

jwillikers:
@jcar87 Fyi, you probably want to reference links to the true `libglvnd` and not NVIDIA's fork: https://gitlab.freedesktop.org/glvnd/libglvnd. NVIDIA's fork is the first thing that shows up in Google, which is really annoying.

First, I'll point out that I'm very confused why we only use X11 as a system package then. As I've been doing this, I've been following what we do for the Wayland and Vulkan packages. I think that Vulkan and libglvnd use a similar runtime loading mechanism, which would mean that it might have similar concerns, right? And,  I thought that Wayland's client and server libraries _should_ always be using the same version, which would also be cause for concern.

It's very likely that containers and containerized application deployments like snap and Flatpak run into some similar concerns. For instance, the Freedesktop Flatpak packages its own version of `libglvnd`, seen [here](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/components/libglvnd.bst) and mesa, seen [here](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/extensions/mesa/mesa.yml). This leads me to believe it is possible to use multiple, different versions of `libglvnd` and `mesa` on a system.

Looking at the `meson_options.txt`, there is no specific configuration option for the path to `/usr/share/glvnd/egl_vendor.d`, which would require an actual patch or using a wonky data directory other than `/usr/share`.

On the needing to rebuild system packages, that's referring to the need to rebuild the system packages in the Conan cache to update the information retrieved from the pkg-config files. I can open a separate issue if/when I hit this again.

Croydon:
Historical speaking, system packages have been an escape hatch, when proper Conan packages were not working, either because a lack of resources on our hands or because it might not be feasible in general. The goal was to get rid of system packages as soon as possible, if it is possible at all (similar with Conan packages that re-package downloaded binaries).

In Bincrafters we had several x11 related packages, but they never really worked at runtime
https://github.com/bincrafters/conan-x11
https://github.com/bincrafters/conan-xorgproto
https://github.com/bincrafters/conan-xorg-macros

Croydon:
Speaking of system packages, #641 never got closed.

PR #23440, opened 2024-04-09
freeglut: Don't require xkbcommon's with_wayland option

jwillikers:
This option only enables building Wayland utility programs in xkbcommon, not the library. I added this previously, so I'm fixing my mistake.

Also, remove `opengl/system` requirement from the test package.

### DIFF
--- a/recipes/freeglut/all/test_package/conanfile.py
+++ b/recipes/freeglut/all/test_package/conanfile.py
@@ -14,7 +14,6 @@ class TestPackageConan(ConanFile):
 
     def requirements(self):
         self.requires(self.tested_reference_str)
-        self.requires("opengl/system")
 
     def build(self):
         cmake = CMake(self)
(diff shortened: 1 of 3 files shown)


recipes/freeglut/all/test_package/conanfile.py
```diff
@@ -14,7 +14,6 @@ def layout(self):
 
     def requirements(self):
         self.requires(self.tested_reference_str)
-        self.requires("opengl/system")
```
Croydon, line 17: Please explain. How does it not require OpenGL at runtime?

jwillikers: I believe it still requires it, but the OpenGL dependency is pulled in transitively from the actual FreeGLUT package. In the case of `libglvnd`, I would assume that nothing actually needs dispatched to a graphics driver in the test package, which seems possible since it appears to only be registering some FreeGLUT error handling functions and printing the FreeGLUT version.

RubenRBS (approved):
<3 thanks a lot for taking the time to fix that once you identified it, really appreciated! :)

PR #23442, opened 2024-04-09
Upgrade opentelemetry-cpp to 1.14.2

wadimklincov:
Specify library name and version:  **opentelemetry-cpp/1.14.2**

I've also added a check for the logs option, since it's been out of preview since `1.11` and added a missing link to `opentelemetry_otlp_recordable` and fixed the missing `opentelemtry-proto` dependency when running with `-o with_otlp_grpc=False`.

---


RubenRBS (approved):
Thanks _a lot_! It's really appreciated. We try to run as many configurations as possible when reviewing, but sometimes we miss some, so it's great when someone double checks for us and fixes it :)

czoido (approved):

PR #23443, opened 2024-04-09
sdl: Use libglvnd instead of opengl/system and egl/system

jwillikers:
The libglvnd package provides the necessary opengl/system and egl/system dependencies on Linux. The libglvnd package is a proper, non-system Conan package.

Fix check for whether the PulseAudio library is shared or not since that option has been removed.

### DIFF
--- a/recipes/sdl/all/conanfile.py
+++ b/recipes/sdl/all/conanfile.py
@@ -139,6 +139,10 @@ class SDLConan(ConanFile):
             self.options.rm_safe("fPIC")
         self.settings.rm_safe("compiler.libcxx")
         self.settings.rm_safe("compiler.cppstd")
+        if self.options.get_safe("opengl"):
+            self.options["libglvnd"].glx = True
+        if self.options.get_safe("wayland"):
+            self.options["libglvnd"].egl = True
 
     def requirements(self):
         if self.options.get_safe("iconv", False):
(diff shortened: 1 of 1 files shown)


valgur (approved):

Croydon (commented):
I have been wondering if we can make more use of libglvnd for some years at this point, but never had the resources to look into it.

Some questions:
1. Is my assumption correct that libglvnd does not actual need an opengl provider at build time?
2. If this works for SDL, can libglvnd replace the usage of opengl/system for all recipes?

jwillikers:
> I have been wondering if we can make more use of libglvnd for some years at this point, but never had the resources to look into it.
> 
> Some questions:
> 
>     1. Is my assumption correct that libglvnd does not actual need an opengl provider at build time?

Yes, it provides the GL interface including the standard GL libraries like EGL, GL, GLX, GLESv1, and GLESv2, the includes, and the commonly used pkg-config files. Hopefully, this PR building in CI in Conan Center is proof enough that no drivers are required. All dispatch to the GPU drivers is done at runtime. If drivers are available on a system, the `libglvnd` Conan package should be able to dispatch to them at runtime. It can also dispatch to drivers from a Mesa Conan package, like the one I currently have a PR open for.

>     2. If this works for SDL, can libglvnd replace the usage of opengl/system for all recipes?

On FreeBSD and Linux, yes, which is why I'm trying to update all of the modern recipes to do this. It's important to be aware of the fact that Mesa can be used in place of or in tandem with `libglvnd`. In most cases, `libglvnd` should be the preferred method, but I can imagine that certain use cases such as embedded may want to use the Mesa package directly in-place of `libglvnd`, but I figure we'll cross that bridge when we get there.

jcar87:

PR #23448, opened 2024-04-09
sfml: Use libglvnd on FreeBSD and Linux

jwillikers:
The libglvnd package provides the necessary opengl/system dependency. The libglvnd package is a proper, non-system Conan package.

### DIFF
--- a/recipes/sfml/all/conanfile.py
+++ b/recipes/sfml/all/conanfile.py
@@ -54,7 +54,9 @@ class SfmlConan(ConanFile):
     def requirements(self):
         if self.options.window:
             # FIXME: use cci's glad
-            if self.settings.os in ["Windows", "Linux", "FreeBSD", "Macos"]:
+            if self.settings.os in ["FreeBSD", "Linux"]:
+                self.requires("libglvnd/1.7.0")
+            elif self.settings.os in ["Macos", "Windows"]:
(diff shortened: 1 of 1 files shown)


danimtb (approved):

franramirez688 (commented):
@jwillikers thanks a lot for the PR!

I was reading about the libglvnd. I did not know much about it to be honest (apart from the Changelog lines mentioned in this PR).
Now, I was wondering if we should set the CMake variable mentioned here: https://github.com/SFML/SFML/blob/d6e1961112c29eee7f60128c29ed699795505167/src/SFML/Window/CMakeLists.txt#L294-L299

jwillikers:
> @jwillikers thanks a lot for the PR!
> 
> I was reading about the libglvnd. I did not know much about it to be honest (apart from the Changelog lines mentioned in this PR). Now, I was wondering if we should set the CMake variable mentioned here: https://github.com/SFML/SFML/blob/d6e1961112c29eee7f60128c29ed699795505167/src/SFML/Window/CMakeLists.txt#L294-L299

Yes, I think so. That variable is a built-in one for CMake's `FindOpenGL` module. I've already set it in the `generate` method in this PR.

Also, it should be noted there's been further discussion on using `libglvnd` in PR #23439.

jcar87:

PR #23461, opened 2024-04-10
djvulibre: new recipe

valgur:
Adds DjVuLibre: https://djvu.sourceforge.net/

The main, if not the only, open-source library to read and write DjVu files.

[Packaging status](https://repology.org/project/djvulibre/versions)

### DIFF
--- a/recipes/djvulibre/all/conanfile.py
+++ b/recipes/djvulibre/all/conanfile.py
@@ -1,0 +1,131 @@
+from conan import ConanFile
+from conan.tools.apple import fix_apple_shared_install_name
+from conan.tools.build import cross_building, stdcpp_library
+from conan.tools.env import Environment, VirtualBuildEnv, VirtualRunEnv
+from conan.tools.files import copy, get, rm, rmdir, save, rename
+from conan.tools.gnu import Autotools, AutotoolsToolchain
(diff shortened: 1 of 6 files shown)


recipes/djvulibre/all/conanfile.py
```diff
@@ -0,0 +1,131 @@
...
+    def package_info(self):
+        self.cpp_info.set_property("pkg_config_name", "ddjvuapi")
+
+        self.cpp_info.libs = ["djvulibre"]
+        self.cpp_info.resdirs = ["res"]
+
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs.extend(["m", "pthread"])
+
+        # Implemented in C++ but exports a pure C API
```
RubenRBS, line 129: Only issue now seems to be that the min c++ 17 enforcement in CCI is making things difficult. The new CI pipeline should be able to handle things like these, but that's still a bit ways off. If the only issue is the register keyword, maybe we can patch those out?

RubenRBS (commented):
Thanks!

RubenRBS (approved):
Thanks!

czoido (approved):

PR #23477, opened 2024-04-11
zeus_expected: new recipe

X1aomu:
**zeus_expected**

A header-only library to backport `std::expected` to C++17.

---


valgur (approved):

RubenRBS (approved):
Hi! Thanks for taking the time to make your library available in Conan Center, we really appreciate it!

The recipe looks super good, I have nothing to suggest changing :)

(Don't worry about the CI, it will be retriggered automatically once the underlying infra issues have been fixed)

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,62 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, get
+from conan.tools.cmake import cmake_layout
+import os
+
+required_conan_version = ">=2.0"
```
uilianries, line 7:
```suggestion
required_conan_version = ">=1.54.0"
```

Conan 1.x is still mandatory.

recipes/zeus_expected/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
...
+project(test_package LANGUAGES CXX)
+
+find_package(zeus_expected REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} test_package.cpp)
+target_link_libraries(${PROJECT_NAME} PRIVATE zeus::expected)
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
+if (MSVC)
+    target_compile_options(${PROJECT_NAME} PRIVATE "/Zc:__cplusplus")
+endif ()
```
uilianries, line 11: Why is it needed? What error does it prevent when running the test package?

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,62 @@
...
+    license = "MIT"
+
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
+
+    no_copy_source = True
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, 17)
```
uilianries, line 26:
```suggestion
    @property
    def _min_cppstd(self):
        return 17

    @property
    def _minimum_compilers_version(self):
        return {
            "gcc": "7",
            "clang": "5",
            "Visual Studio": "15.7",
            "apple-clang": "10",
        }


    def validate(self):
        if self.settings.compiler.get_safe("cppstd"):
            check_min_cppstd(self, self._min_cppstd)

        min_version = self._minimum_compilers_version.get(str(self.settings.compiler))
        if min_version and Version(self.settings.compiler.version) < min_version:
            raise ConanInvalidConfiguration(f"{self.name} requires C++{self._min_cppstd}, which your compiler does not support.")
```

Need to check compiler version when not having settings.compiler.cppstd defined (Conan 1.x)

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,62 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, get
+from conan.tools.cmake import cmake_layout
```
uilianries, line 4:
```suggestion
from conan.tools.cmake import cmake_layout
from conan.errors import ConanInvalidConfiguration
```

uilianries (commented):
Looking good, only some details.

recipes/zeus_expected/all/conanfile.py
RubenRBS: Nice catch @uilianries I totally glossed over it :)

recipes/zeus_expected/all/test_package/CMakeLists.txt
X1aomu: the library utilizes `__cplusplus` to check what standard is in use, without `/Zc:__cplusplus`, programs will not compile (fail with static_assert). 

A workaround is to make use of `_MSVC_LANG`. To make this library for wider and eaiser use, I am considering that approach and would make a change in the future.

Tell me if the current `/Zc:__cplusplus` way could have any harm to conan. Should this recipe wait until the library abandon the requirement of `/Zc:__cplusplus` ?

recipes/zeus_expected/all/conanfile.py
X1aomu: Thanks for the detailed example. As a C++ library, I haven't consider that before. Is it only need for conan 1.x? If so, I would leave a comment so that we can remove those boilerplate code later.

uilianries: Don't need to comment because is well used over all ConanCenterIndex and part of templates: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py

In the future, after dropping Conan 1.x, will be able to track it anyway and remove. Thank you for asking.

recipes/zeus_expected/all/test_package/CMakeLists.txt
uilianries: But the `__cplusplus` should be evaluated by default in Visual Studio as well: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

We are passing `cxx_std_17`, it should be enough to work the test package.

recipes/zeus_expected/all/conanfile.py
X1aomu: > Do you mean if the CI run both? 

No, I mean personal development env. I managed to do it by create a new user for conan 1.x testing only. Just want to know if there are any eaiser way.

uilianries: Ah, now I see. Well, you have some options, depending what you prefer:

- You could use Docker images to validate linux builds:

```
cd conan-center-index/recipes/
docker pull conanio/gcc11-ubuntu16.04:1.64.0
docker run -ti -v $PWD:/home/conan/project conanio/gcc11-ubuntu16.04:1.64.0
# In the container
cd project/zeus_expected/
conan create all 1.0.2@
```

Note that I passed the docker tag as the Conan version. Avoid the tag `latest` always.
All images are hosted in hub.docker.com/r/conanio, but only of few are maintained/used nowadays: https://github.com/conan-io/conan-docker-tools?tab=readme-ov-file#storage-and-availability

- Use [pyenv](https://github.com/pyenv/pyenv) to create separated environments (I use this one in Mac)

```
pyenv virtualenv conan1
pyenv activate conan1
pip install conan==1.64.0
cd conan-center-index/recipes/zeus_expected/
conan create all 1.0.2@
...
pyenv deactivate conan1
```

The environment `conan1` will be always available, so you can activate again. You can do the same for Conan 2.x in case wanting to have more versions installed.

recipes/zeus_expected/all/test_package/CMakeLists.txt
uilianries: @X1aomu Okay! Now I see, so we need to move it to `package_info` instead. Otherwise, users that consume this package, will need to pass that flag manually. I'll point the change in the code.

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,82 @@
...
+        copy(
+            self,
+            "*",
+            src=os.path.join(self.source_folder, "include"),
+            dst=os.path.join(self.package_folder, "include"),
+        )
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "zeus_expected")
+        self.cpp_info.set_property("cmake_target_name", "zeus::expected")
```
uilianries, line 75:
```suggestion
        self.cpp_info.set_property("cmake_target_name", "zeus::expected")
        if is_msvc(self):
            self.cpp_info.cxxflags.append("/Zc:__cplusplus")
```

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,82 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, get
+from conan.tools.cmake import cmake_layout
```
uilianries, line 4:
```suggestion
from conan.tools.cmake import cmake_layout
from conan.tools.microsoft import is_msvc
```

recipes/zeus_expected/all/test_package/CMakeLists.txt
uilianries, line 11:
```suggestion
```
Moved to package_info

uilianries:
> Hi, @uilianries
> 
> Though I've successfully run `conan create` against conan 1.54 with the hook enabled, there are still a few more questions I have,
> 
>     1. does it need a seperated test_v1_package folder for test package? what content should lie in there?
> 
>     2. I saw some old things about conan 1.x from the recipe of `tl-expected`. Are they still necessary nowadays? (they cause the deprecated warning in conan 2 that I don't like)
> 
> 
> https://github.com/conan-io/conan-center-index/blob/85854290b51c1bb754e672bd2884e08814206750/recipes/tl-expected/all/conanfile.py#L50-L61

@X1aomu Thank you for validating even Conan 1.x locally! 

>  1. does it need a seperated test_v1_package folder for test package? what content should lie in there?

No, for new recipes, the same test_package folder should be used by both Conan 2.x and 1.x, with the exactly same code, no condition related to the version. The legacy test_v1_package folder was kept to prove compatibility with legacy Conan generators like `cmake` and `cmake_find_package`. So you do not need to create test_v1_package, your current test_package is good enough. More details about the deprecation of support for legacy generators here: https://github.com/conan-io/conan-center-index/discussions/17406

> 2. I saw some old things about conan 1.x from the recipe of `tl-expected`. Are they still necessary nowadays? (they cause the deprecated warning in conan 2 that I don't like)

Do you mean this line, right? https://github.com/conan-io/conan-center-index/blob/master/recipes/tl-expected/all/conanfile.py#L50

It was added only for recipes in ConanCenterIndex that was present since Conan 1.x and then migrated to support Conan 2.x. As users would need time to migrate too, we kept both legacy and new generators supported in the recipes. But this a decision only for old recipes, new recipes like yours should ignore that rule and do not add support for those legacy generators. 

> Though I've successfully run `conan create` against conan 1.54 with the hook enabled

About the hooks, don't worry too much about the warnings, it's more a notification than an obligation, some of them are covered already by Conan 2.x as built-in rule (e.g package name should be lower-case). In Conan 2.x we still don't have hooks in ConanCenterIndex, but we are designing other ways to check it.

recipes/zeus_expected/all/test_package/CMakeLists.txt
X1aomu: Well, as the author of zeus_expected, i just released a [new version](https://github.com/zeus-cpp/expected/releases/tag/v1.1.0) to drop the requirement of `/Zc:__cplusplus`.

I'd like to let the conan recipe start from that new version. Thus conan users would forget about `/Zc:__cplusplus` :)

recipes/zeus_expected/all/conanfile.py
X1aomu: Resolved by a [new upstream version](https://github.com/zeus-cpp/expected/releases/tag/v1.1.0) that don't need `"/Zc:__cplusplus` anymore.

recipes/zeus_expected/all/conanfile.py
X1aomu: Resolved by a [new upstream version](https://github.com/zeus-cpp/expected/releases/tag/v1.1.0) that don't need `"/Zc:__cplusplus` anymore.

recipes/zeus_expected/all/test_package/CMakeLists.txt
uilianries: Awesome!!! 👏

recipes/zeus_expected/all/conanfile.py
uilianries: Thank you!!

recipes/zeus_expected/all/conandata.yml
```diff
@@ -0,0 +1,10 @@
+sources:
+  "1.1.0":
+    url: "https://github.com/zeus-cpp/expected/archive/refs/tags/v1.1.0.tar.gz"
+    sha256: "a963eba43f227498da2cbb924265344209696320c75ee63a92073936bb49f7e5"
+
+patches:
+  "1.1.0":
+    - patch_file: "patches/0001-fix-compiler-errors-with-msvc-v142.patch"
```
uilianries, line 8:
```suggestion
    - patch_file: "patches/0001-fix-compiler-errors-with-msvc-v142.patch"
      patch_source: "https://github.com/zeus-cpp/expected/pull/7/commits/82ed513aa65a3c2c7b8f8af7c63e0b8b57637dfb"
```

Try to keep source of the change tracked

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+    homepage = "https://github.com/zeus-cpp/expected"
+    description = "Backporting std::expected to C++17."
+    topics = ("cpp17", "expected")
+    license = "MIT"
+
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
+
+    # patches applied
+    # no_copy_source = True
```
uilianries, line 26:
```suggestion
    no_copy_source = True
```

As hearder only project and you are actually only copying file, does not worth keeping a build folder, it would be a copy of a copy. So we can copy from source directly. The patch will be applied directly to source, but that's a common usage.

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+            check_min_cppstd(self, self._min_cppstd)
+
+        min_version = self._minimum_compilers_version.get(str(self.settings.compiler))
+        if min_version and Version(self.settings.compiler.version) < min_version:
+            raise ConanInvalidConfiguration(
+                f"{self.name} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
```
uilianries, line 55:
```suggestion
        basic_layout(self, src_folder="src")
```

Cmake layout is nice when you are actually using CMake. The layout organizes the folder tree used when building and copying sources. But in this case, you want only to copy files, so basic_layout is lighter and fits better.

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+        copy(
+            self,
+            "*",
+            src=os.path.join(self.source_folder, "include"),
+            dst=os.path.join(self.package_folder, "include"),
+        )
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "zeus_expected")
```
uilianries, line 78: Just to you know, do not need to change the code, it's totally fine: Conan uses the package name as `cmake_file_name` in case not defined. So what you are defining here, Conan would use the very same name.

X1aomu:
glad to see all checks passed for the first time :)

I am going to investigate the bug further and merge the patch to upstream.

recipes/zeus_expected/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import copy, get
+from conan.tools.cmake import cmake_layout
```
uilianries, line 4:
```suggestion
from conan.tools.layout import basic_layout
```

uilianries (commented):
Almost good to go, only small details. Thank you for releasing a new version!

recipes/zeus_expected/all/conanfile.py
X1aomu: Thanks for explanation. I misunderstood it

recipes/zeus_expected/all/conanfile.py
X1aomu: ahh, i see

uilianries (approved):
LGTM. Thank you for your PR!

toge (approved):

PR #23478, opened 2024-04-11
boost: ignore dependency if it doesn't provide a library [v1]

kambala-decapitator:
Specify library name and version:  **boost/any**

For Conan v1 I use the following custom recipes that wrap system libraries on Apple platforms:
- bzip2: https://github.com/kambala-decapitator/conan-system-libs/blob/main/bzip2/conanfile.py
- zlib: https://github.com/kambala-decapitator/conan-system-libs/blob/main/zlib/conanfile.py

But when I try to build Boost against those, it fails due to attempt to access first element of an empty array. This patch prevents that.

Tested by building Boost as shared library and my project and then running my project successfully.

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -1344,6 +1344,9 @@ class BoostConan(ConanFile):
 
         def create_library_config(deps_name, name):
             aggregated_cpp_info = self.dependencies[deps_name].cpp_info.aggregated_components()
+            if len(aggregated_cpp_info.libs) == 0:
+                return ""
+
             includedir = aggregated_cpp_info.includedirs[0].replace("\\", "/")
             includedir = f"\"{includedir}\""
             libdir = aggregated_cpp_info.libdirs[0].replace("\\", "/")


uilianries:
@kambala-decapitator Thank you for your PR! This current Boost recipe is linking to both zlib and bzip2 libraries present in Conan Center and is working. Why do you need these custom recipes instead of using the officials?

Plus, could you please share your full build log with your described error?

kambala-decapitator:
> @kambala-decapitator Thank you for your PR! This current Boost recipe is linking to both zlib and bzip2 libraries present in Conan Center and is working. Why do you need these custom recipes instead of using the officials?
> 
> Plus, could you please share your full build log with your described error?

@uilianries I'd like to use system libraries - the less custom dependencies the better :) This has worked great with a few other libraries from CCI. I know that v2 has a built-in way to declare system libs, but currently our project is still using v1.

sorry, don't have the build log at hand any more, but initially it said something like "index out of bounds" pointing at the line

    includedir = aggregated_cpp_info.includedirs[0].replace("\\", "/")

at first I modified my custom recipes to include library name in `libs` as well as empty string in `includedirs` and `libdirs` (arrays that access the first element) and boost built successfully, but then at cmake configure time it started complaining

```
-- Conan: Target declared 'BZip2::BZip2'
CMake Error at conan-generated/cmakedeps_macros.cmake:39 (message):
  Library 'bz2' not found in package.  If 'bz2' is a system library, declare
  it with 'cpp_info.system_libs' property
  conan-generated/BZip2-Target-release.cmake:24 (conan_package_library_targets)
  conan-generated/BZip2Targets.cmake:26 (include)
  conan-generated/BZip2Config.cmake:16 (include)
  /usr/local/Cellar/cmake/3.29.1/share/cmake/Modules/CMakeFindDependencyMacro.cmake:76 (find_package)
  conan-generated/BoostConfig.cmake:24 (find_dependency)
  CMakeLists.txt:462 (find_package)
```

you can see full log at https://github.com/vcmi/vcmi/actions/runs/8637368273/job/23679378148

And with the current patch it all builds and works perfectly.

uilianries:
> I'd like to use system libraries - the less custom dependencies the better :) This has worked great with a few other libraries from CCI. I know that v2 has a built-in way to declare system libs, but currently our project is still using v1.

@kambala-decapitator Sorry, I still miss the big picture, because you are actually using a Conan package for Boost, but can not use for zlib and bzip2? Those packages are built exactly as recommended by the upstream and most excepted in Linux distros, we usually compare patches from distros linux like Ubuntu and from AUR to make sure. Also, you can build from source all packages and inspect the recipe exported, in case you have security reasons. Is there something wrong with those packages that we could improve?

> at first I modified my custom recipes to include library name in libs as well as empty string in includedirs and libdirs (arrays that access the first element) and boost built successfully, but then at cmake configure time it started complaining

Yes, packaging system libraries is a bit more complicated. You can use some recipes from CCI as reference:

- https://github.com/conan-io/conan-center-index/blob/b229eb291ea027346087e1e2de11aac6657c4187/recipes/glu/all/conanfile.py#L31
- https://github.com/conan-io/conan-center-index/blob/b229eb291ea027346087e1e2de11aac6657c4187/recipes/gtk/system/conanfile.py#L31
- https://github.com/conan-io/conan-center-index/blob/b229eb291ea027346087e1e2de11aac6657c4187/recipes/opengl/all/conanfile.py#L25

> you can see full log at [vcmi/vcmi/actions/runs/8637368273/job/23679378148](https://github.com/vcmi/vcmi/actions/runs/8637368273/job/23679378148)

Thank you for sharing!

jwillikers (approved):

jwillikers:
@uilianries I've run into a similar problem elsewhere when using `replace_requires` with a system package. Trying to use `cpp_info` of system libraries tends not to work because the `libs` or certain other bits are not set.

jwillikers:
@kambala-decapitator You might try setting the `libs` and `libdirs` variables appropriately. Ideally you could use `pkg-config` which handles transitive dependencies, but if I remember correctly, `bzip` doesn't provide one yet. They have one in `master` I think.

kambala-decapitator:
@uilianries @jwillikers thank you for the responses!

> you are actually using a Conan package for Boost, but can not use for zlib and bzip2?

Apple OSs have these libraries in the system (I also have custom recipes to wrap iconv and sqlite3 btw) and I want to use them instead of building them from source. It's common practice in Apple world to use these built-in libraries in programs. I don't think that I _must_ use them from CCI only?

> Yes, packaging system libraries is a bit more complicated. You can use some recipes from CCI as reference:

these examples install something from package managers. I don't need this, all the aforementioned libs are present in a freshly installed OS.

> You might try setting the libs and libdirs variables appropriately

yes, that's what I tried initially. But once I set `libs`, then cmake configuration of my project fails, see https://github.com/conan-io/conan-center-index/pull/23478#issuecomment-2049294082

<hr>

btw I've just thought of this: would it make more sense to adjust the patch to check if `system_libs` is present and return empty string in that case?

P.S. Do I need to take any action because of the unexpected build error?

kambala-decapitator:
@uilianries please check my above comment

uilianries:
> Apple OSs have these libraries in the system (I also have custom recipes to wrap iconv and sqlite3 btw) and I want to use them instead of building them from source. It's common practice in Apple world to use these built-in libraries in programs. I don't think that I must use them from CCI only?

@kambala-decapitator Thank you for more details about your case. Could please refer a source about this common practice in Apple world? I've never heard about, so I'm curious. The problem using system dependencies is tracking the changes and reproducing the same build with total safety. You are totally free to mix Conan packages with system packages at your risk, if this enough to cover your case, so is totally fine :)

uilianries (approved):
LGTM

kambala-decapitator:
@uilianries thanks for the merge!

I don't have a source in the sense of an article from documentation, but if you open any Xcode project and go to the `Link Binary With Libraries` section, you can select a wide range of system libraries (besides frameworks) there, see SS. When one works purely with Xcode project, it's common to choose libraries from there.

Screenshot 2024-04-23 at 10 53 21

PR #23498, opened 2024-04-12
cyrus-sasl: add option to disable saslauthd

adamws:
The `saslauthd` is a daemon that handles authentication requests on behalf of the SASL library.
Package consumers might not be interested in compiling this daemon when integrating SASL protocol in own application.

Introduced `with_saslauthd` option which defaults to `True`. On Linux & FreeBSD that was the default when not explicitly added to `configure`.

For some cross-compilation scenarios, having this option always enabled was an issue because `-lcrypt` might be missing and it would fail to link (for example when using
arm-linux-gnueabihf-gcc toolchain).
From now on this optional daemon can be disabled using `with_saslauthd=False`. In such case, `crypt` is not added to `system_libs`.

Specify library name and version:  **cyrus-sasl/2.1.28**

---


valgur (approved):
LGTM. Thanks!

RubenRBS (approved):
Thanks a lot! This seems to compile nicely for me :)

PR #23519, opened 2024-04-14
make minimum compiler version explicit in error message

BenBudr:
Minor addition to an error message. Was mildly frustrating being told that my compiler (gcc 11.0) didn't support C++20 and having to dig into the conanfile to work out exactly what the issue was.


RubenRBS (approved):
I like how that reads, thanks a lot! Note that these blocks will be removed on most recipes once Conan 1 support is dropped, but for now it's ok :)

Please remember to request access in the issue that the bot has linked so that this can get merged :)

PR #23522, opened 2024-04-14
fix template conan doesn't pass the linter

SSE4:

code from template doesn't pass the linter:
```
[C3001(unnecessary-lambda-assignment), OpenldapConan.generate.<lambda>] Lambda expression assigned to a variable. Define a function using the "def" keyword instead.
```

---

### DIFF
--- a/docs/package_templates/autotools_package/all/conanfile.py
+++ b/docs/package_templates/autotools_package/all/conanfile.py
@@ -132,7 +132,7 @@ class PackageConan(ConanFile):
         # --enable/disable-shared is automatically managed when 'shared' option is declared
         tc = AutotoolsToolchain(self)
         # autotools usually uses 'yes' and 'no' to enable/disable options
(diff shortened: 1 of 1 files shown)


RubenRBS (commented):
I think the proper fix is to disable this linter! I very much prefer to have lambdas as variables than using the single-line def syntax which imo seems uglier :)

SSE4:
> I very much prefer to have lambdas as variables than using the single-line def syntax which imo seems uglier :)

that's just about personal preference. I don't see and use lambdas often in python code, so for me it takes some more time to read and understand the code compare to the regular function.

RubenRBS:
ok, I'll approve to unblock you here - note that in the future we'll have less and less lint rules running as actions, and for "superficial" ones rely on pre-commit hooks for those interested in running them

RubenRBS (approved):

ericLemanissier (approved):

PR #23531, opened 2024-04-15
glpk: add v4.48 and v4.65 for COIN-OR, add GMP support

valgur:
Used by multiple COIN-OR packages, but only v4.48 is compatible. I also included the latest available v4.x version in case anyone needs it.


EstebanDugueperoux2:
The build of this PR works on debian:bullseye based Docker image on Linux.
The "mkdir: cannot create directory '/dev/mqueue': Read-only file system" error message seems due to a bad Windows build environment.
@AbrilRBS @jcar87  Could the CI of this PR triggered again to check if this error occurs again?
Regards.

PR #23530, opened 2024-04-15
coin-buildtools: new recipe

valgur:
Packages Autotools M4 files required for running `autoreconf` on the COIN-OR libraries on CCI. https://github.com/coin-or-tools/BuildTools

The COIN-OR libraries have been struggling with cross-building support on macOS and shared builds on MSVC due to the included configure scripts being generated with a fairly old Autoconf version. This will hopefully resolve these issues.

It also adds newer versions of the Autotools toolchain packages as transitive build-time requirements (but as `self.requires()` due to Conan v1 limitations) to ensure that up-to-date versions are being used.

PR #23532, opened 2024-04-15
coin-utils: add v2.11.11, apply autoreconf, use OpenBLAS, add GLPK support

valgur:
Also sets `--enable-coinutils-threads` on systems where it is supported.

I did not make BLAS and LAPACK support optional as using COIN-OR libraries without them barely makes sense, really. Can be added later, if necessary.

GLPK is used in most of the COIN-OR packages, but is GPL-licensed, so I included an option to disable it.

Requires
- #23530
- #23531


recipes/coin-utils/all/conanfile.py
```diff
@@ -55,22 +58,15 @@ def layout(self):
...
-
-    def validate(self):
-        if self.settings.os == "Windows" and self.options.shared:
-            raise ConanInvalidConfiguration("coin-utils does not provide a shared library on Windows")
-        # FIXME: This issue likely comes from very old autotools versions used to produce configure.
-        #        It might be fixed by calling autoreconf, but https://github.com/coin-or-tools/BuildTools
-        #        should be packaged and added to build requirements.
-        if hasattr(self, "settings_build") and cross_building(self) and self.options.shared:
-            raise ConanInvalidConfiguration("coin-utils shared not supported yet when cross-building")
+        self.requires("openblas/0.3.26")
```
EstebanDugueperoux2, line 61:
```suggestion
        self.requires("openblas/0.3.27")
```

EstebanDugueperoux2 (changes requested):
Just a requirement bump

recipes/coin-utils/all/conanfile.py
valgur: I'll handle this once the two prerequisite PRs have been merged.

PR #23537, opened 2024-04-15
gnu-getopt: new recipe

valgur:
Packages the GNU variant of the `getopt` command-line utility. Not to be confused with the `getopt` library.

Required on macOS for the shell scripts included in the f2c PR #23417.

The current version of the tool was originally hosted at http://frodo.looijaard.name/project/getopt but has now been merged into `linux-utils`. There's a slight question of whether we should simply build and package all of the utils in `linux-utils` instead. However, I don't think they are all compatible with other OS-s so it's safer to package `getopt` in isolation for now.

[Packaging status](https://repology.org/project/getopt-standalone/versions)

### DIFF
--- a/recipes/gnu-getopt/all/conanfile.py
+++ b/recipes/gnu-getopt/all/conanfile.py
@@ -1,0 +1,72 @@
+import os
+
+from conan import ConanFile
+from conan.tools.env import Environment, VirtualBuildEnv
+from conan.tools.files import copy, get
+from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.layout import basic_layout
+from conan.tools.microsoft import is_msvc, unix_path
+
+required_conan_version = ">=1.54.0"
+
+
+class GnuGetoptConan(ConanFile):
+    name = "gnu-getopt"
+    description = "GNU getopt(1) command-line utility"
+    license = "GPL-2.0-or-later"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "http://frodo.looijaard.name/project/getopt"
+    topics = ("gnu", "getopt", "utility", "command-line", "parsing")
+
+    package_type = "application"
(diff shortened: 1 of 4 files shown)


jwillikers:
Ugh, this will mean another separate project to convert to Meson. 😢 I just got changes in the stable branch for 2.40 that should make it possible to build this with Meson on macOS, but it looks like you're also supporting Windows. I'd better see if Meson needs any changes to support that.

jwillikers (approved):
LGTM! I'd still like to lump all of the util-linux recipes together in one recipe eventually.

jwillikers (approved):

recipes/gnu-getopt/all/conanfile.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        env = VirtualBuildEnv(self)
+        env.generate()
+
+        tc = AutotoolsToolchain(self)
+        # liblastlog support requires sqlite3
```
RubenRBS, line 58: Any insight into not adding this as an option? I'm happy to keep as-is, but wanted to hear some of your insight if you had any :)

RubenRBS (commented):
Thanks! Minor question regarding the sqlite3 dependency, other than that, looks good :)

recipes/gnu-getopt/all/conanfile.py
jwillikers: We almost certainly shouldn't add any extra libraries to avoid potential conflicts with `libmount` and `util-linux-libuuid`.

valgur: As far as I can understand the project setup of `util-linux`, the `getopt` executable does not even make use of the liblastlog dependency anyway.

danimtb (approved):

Issue #16821, opened 2023-03-31
[package] osqp/0.6.2: Build fails with Visual Studio and Ninja as CMake generator


SpaceIm:
### Description

Build fails with `ninja: error: build.ninja:320: multiple rules generate out/osqp.lib` with Ninja generator and Visual Studio 2022

### Package and Environment Details

* Package Name/Version: **osqp/0.6.2**
* Operating System+version: **Windows 11**
* Compiler+version: **Visual Studio 2022**
* Conan version: **conan 2.0.2**
* Python version: **Python 3.11.1**

### Conan profile

```jinja
[settings]
os=Windows
arch=x86_64
compiler=msvc
compiler.version=193
compiler.runtime=dynamic
build_type=Release
[options]
[tool_requires]
[conf]
tools.microsoft.msbuild:installation_path=C:/Program Files/Microsoft Visual Studio/2022/Community
[buildenv]
CMAKE_EXPORT_COMPILE_COMMANDS=ON
```

### Steps to reproduce

`conan install --requires="osqp/0.6.2" -o "*:shared=True" -b missing`

### Logs

<details><summary>Click to expand log</summary>

```
osqp/0.6.2: Building from source
osqp/0.6.2: Package osqp/0.6.2:f7dcd1cec4586c154a242a765a1de011300db8e2
osqp/0.6.2: Copying sources to build folder
osqp/0.6.2: Building your package in C:\Users\spaceim\.conan2\p\t\osqp6542c78057d19\b
osqp/0.6.2: Calling generate()
osqp/0.6.2: Generators folder: C:\Users\spaceim\.conan2\p\t\osqp6542c78057d19\b\build\Release\generators
osqp/0.6.2: CMakeToolchain generated: conan_toolchain.cmake
osqp/0.6.2: CMakeToolchain generated: CMakePresets.json
osqp/0.6.2: CMakeToolchain generated: ..\..\..\src\CMakeUserPresets.json
osqp/0.6.2: Generating aggregated env files
osqp/0.6.2: Generated aggregated env files: ['conanbuild.bat', 'conanrun.bat']
osqp/0.6.2: Calling build()
osqp/0.6.2: Running CMake.configure()
[...]

PR #23539, opened 2024-04-15
osqp 0.6.2 and 0.6.3 with Ninja generator + osqp build options

tkhyn:
Specify library name and version:  **osqp/0.6.2** and **osqp/0.6.3**

This PR fixes issue #16821 by patching OSQP's cmake file to generate the shared and static libraries in different folders.
It also exposes the OSQP build options in the recipe.

---


recipes/osqp/all/patches/0.6.3-set-libraries-output-paths.patch
valgur, line 1: Since only either static or shared libraries get packaged, ultimately, I would recommend simply disabling the other one based on `BUILD_SHARED_LIBS` instead.

recipes/osqp/all/conanfile.py
```diff
@@ -18,12 +18,29 @@ class OsqpConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "printing": [True, False],
+        "profiling": [True, False],
+        "ctrlc": [True, False],
+        "dfloat": [True, False],
+        "dlong": [True, False],
+        "coverage": [True, False],
```
valgur, line 26: `coverage` does not really make much sense as a package option. It only has an effect when unit tests are being run.

recipes/osqp/all/conanfile.py
```diff
@@ -18,12 +18,29 @@ class OsqpConan(ConanFile):
...
         "shared": False,
         "fPIC": True,
+        "printing": True,
+        "profiling": True,
+        "ctrlc": True,
+        "dfloat": False,
+        "dlong": True,
+        "coverage": False,
+        "mkl_paradisio": True
     }
```
valgur, line 39: An `options_description = {...}` field would be nice, since the meaning of the options isn't super obvious.

recipes/osqp/all/conanfile.py
```diff
@@ -18,12 +18,29 @@ class OsqpConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "printing": [True, False],
+        "profiling": [True, False],
+        "ctrlc": [True, False],
+        "dfloat": [True, False],
+        "dlong": [True, False],
+        "coverage": [True, False],
+        "mkl_paradisio": [True, False]
```
valgur, line 27: I would suggest skipping the `mkl_paradisio` option as it is being replaced by `OSQP_ALGEBRA_BACKEND` in the upcoming v1 release and would have to be deprecated. https://github.com/osqp/osqp/blob/master/CMakeLists.txt#L77-L79

recipes/osqp/all/conanfile.py
```diff
@@ -18,12 +18,29 @@ class OsqpConan(ConanFile):
...
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "printing": True,
+        "profiling": True,
+        "ctrlc": True,
```
valgur, line 34: Has been renamed to `interrupt` in upcoming 1.0 version. Might be worth renaming here as well. https://github.com/osqp/osqp/blob/master/CMakeLists.txt#L83

recipes/osqp/all/patches/0.6.3-set-libraries-output-paths.patch
RubenRBS: I like this approach better too

recipes/osqp/all/conanfile.py
tkhyn: Yes, as a library user I don't really care about coverage so happy to skip (done in commits I just pushed), but as the `UNITTEST` option was handled by the recipe possibly the `COVERAGE` could be too.

tkhyn:
Thank you @valgur and @RubenRBS for the feedback. I have implemented the requested changes and learnt a few things on the way!

Note: it doesn't look like `shared=True`  works on Windows though due to the library not exporting any symbols. It looks like it's more a library issue than a recipe issue though (I understand it's fixed in version 1.0), although if you think it's a good thing we could force `shared=False` on windows.

RubenRBS:
Let's validate out the shared case for windows until 1.0 yes :)

uilianries:
The release 1.0 is out and has separate options for shared and static: https://github.com/osqp/osqp/compare/v0.6.3...v1.0.0#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR59

It sounds better skipping patching it for 0.x and using the latest version instead.

Issue #23557, opened 2024-04-15
[question] new versions of glib 


ericLemanissier:
### What is your question?

Since glib 2.79.0, glib uses python's `packaging` [module ](https://pypi.org/project/packaging/) (cf https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3740 at build time and run time. this module is not installed in conan-center CI runners, which prevents from building all versions of glib since 2.79.0 (cf https://github.com/conan-io/conan-center-index/pull/21874 https://github.com/conan-io/conan-center-index/pull/23093)
Conan team rejected all the solutions proposed by the community(https://github.com/conan-io/conan-docker-tools/pull/554 https://github.com/conan-io/conan-center-index/pull/23091)
glib does not intend to drop the dependency on this `packaging` module : https://gitlab.gnome.org/GNOME/glib/-/issues/3301

What's the plan ? 

@uilianries @memsharded

valgur:
I'm feeling foolish enough to take this one on... Here's my humble attempt: #23559

ericLemanissier:
https://github.com/conan-io/conan-center-index/pull/24165

PR #23193, opened 2024-03-21
Glib 2.80.0

ericLemanissier:
Specify library name and version:  **glib/2.80.0**

Patch is necessary to remove dependency to python packaging, following https://github.com/conan-io/conan-docker-tools/pull/554#issuecomment-1999860485

don't use `PkgConfig` because it invokes `pkg-config` which is not installed on macos runners.

remove unused versions:
- https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+glib%2F2.75.3&type=code
- https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+glib%2F2.77.0&type=code
- https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+glib%2F2.77.1&type=code
- https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+glib%2F2.77.2&type=code

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #23559, opened 2024-04-15
glib: add v2.80.0

valgur:
Continues from #23193. Related to #23557 and https://github.com/conan-io/conan-docker-tools/pull/554#issuecomment-1999860485

~~Adds `self.tool_requires("cpython/3.12.2")` and installs a pinned version of `packaging` from PyPI locally into the build folder.~~

~~The published package artifacts on PyPI are immutable, so the pinned version will provide reproducible builds and is reflected in the package hash via the conanfile.py contents, if I'm not mistaken.~~

~~I think this approach is a simple and scalable way to deal with build-scope Python dependencies on CCI in general.~~

Edit: @ericLemanissier pointed out to me that the `packaging` dependency is a transitive build dep. Since we decided not to package it separately, the only option left was to vendor the two required files. Which I did.

### DIFF
--- a/recipes/glib/all/test_package/conanfile.py
+++ b/recipes/glib/all/test_package/conanfile.py
@@ -2,7 +2,8 @@ from conan import ConanFile
 from conan.tools.build import can_run
 from conan.tools.cmake import cmake_layout, CMake, CMakeDeps, CMakeToolchain
 from conan.tools.env import Environment, VirtualBuildEnv, VirtualRunEnv
-from conan.tools.gnu import PkgConfig, PkgConfigDeps
+from conan.tools.gnu import PkgConfigDeps
+import io
 import os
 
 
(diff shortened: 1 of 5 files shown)


ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/8691972924/attempts/1)

recipes/glib/all/test_package/conanfile.py
```diff
@@ -56,4 +56,4 @@ def test(self):
             if self.settings.os != "Windows":
                 pkg_config = PkgConfig(self, "gio-2.0", pkg_config_path=self.generators_folder)
                 gdbus_codegen = pkg_config.variables["gdbus_codegen"]
```
ericLemanissier, line 58:
```suggestion
                mybuf = io.StringIO()
                self.run(f"PKG_CONFIG_PATH={self.generators_folder} pkgconf --variable=gdbus_codegen gio-2.0", mybuf, env="conanbuild")
                gdbus_codegen = mybuf.getvalue().strip()
```

ericLemanissier (approved):

ericLemanissier (commented):
@memsharded @uilianries I think this PR is implementing the solution you had in mind in https://github.com/conan-io/conan-docker-tools/pull/554#issuecomment-1999860485
Can we get a review in order to finally make progress ?

valgur:
Closing as redundant to #24591. Most of the changes here were related to the Python 3.12 issues.

PR #23562, opened 2024-04-15
openssl/3.2.1: Patch for MD5 cert names on Android.

sizeak:
Specify library name and version:  **openssl/3.2.1**

Android ships CA Certificates with MD5 hash names which OpenSSL stopped supporting in v1.0.0 in favour of canonically encoded SHA1 names (according to https://github.com/openssl/openssl/issues/13565#issuecomment-736579342). This means that by default you can't use the certificates Android ships with OpenSSL to enable https support with libcurl etc. Instead you would have to include your own CA bundle with your app/library or do some work to copy & rehash the system certificates, neither of which are ideal.

There is a [PR open](https://github.com/openssl/openssl/pull/24002) (not mine) with OpenSSL which addresses this problem in a more generic way by allowing the library to fallback to MD5 when no file is found with SHA1, but this PR appears to be slightly contentious and it's not clear if/when this will be merged. Even if it is merged, it will not enable us to use OpenSSL easily on Android right now, with the currently released versions.

This PR provides a patch allowing the use of the older MD5 based names in favour of the SHA1 names when the `FORCE_MD5_X509_NAME_HASHES` preprocessor definition is detected. Combined with a new option in the recipe, this enables the build to be configured to work with the Android system CA certs without forcing this on anyone already building for Android and using other workarounds.

Hopefully if/when the linked PR is approved, this will no longer be needed, but until then it makes it much easier to get https working on Android with curl etc. (for anyone finding this from a search engine, you also need to point curl at the certs using something like `curl_easy_setopt(handle, CURLOPT_CAPATH, "/system/etc/security/cacerts");` )

---

### DIFF
--- a/recipes/openssl/3.x.x/conandata.yml
+++ b/recipes/openssl/3.x.x/conandata.yml
@@ -36,7 +36,7 @@ sources:
     sha256: f93c9e8edde5e9166119de31755fc87b4aa34863662f67ddfcba14d0b6b69b61
 patches:
   3.2.1:
-    - patch_file: "patches/3.2.1-android-x509-md5-names.patch"
+    - patch_file: "patches/3.2.1-x509-md5-names.patch"
       patch_description: "Android ships MD5 hash named certificates which OpenSSL no longer supports by default."
       patch_type: portability
       patch_source: https://github.com/openssl/openssl/issues/13565
--- a/recipes/openssl/3.x.x/conandata.yml
+++ b/recipes/openssl/3.x.x/conandata.yml
@@ -35,6 +35,11 @@ sources:
       - "https://github.com/openssl/openssl/releases/download/openssl-3.0.12/openssl-3.0.12.tar.gz"
     sha256: f93c9e8edde5e9166119de31755fc87b4aa34863662f67ddfcba14d0b6b69b61
 patches:
+  3.2.1:
+    - patch_file: "patches/3.2.1-android-x509-md5-names.patch"
+      patch_description: "Android ships MD5 hash named certificates which OpenSSL no longer supports by default."
+      patch_type: portability
+      patch_source: https://github.com/openssl/openssl/issues/13565
   3.2.0:
     - patch_file: "patches/3.2.0-fix-winsock2.patch"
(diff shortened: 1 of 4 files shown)


jcar87:
> There is a https://github.com/openssl/openssl/pull/24002 (not mine) with OpenSSL which addresses this problem in a more generic way by allowing the library to fallback to MD5 when no file is found with SHA1, but this PR appears to be slightly contentious and it's not clear if/when this will be merged. Even if it is merged, it will not enable us to use OpenSSL easily on Android right now, with the currently released versions.

Hi @sizeak - thank you for taking the time to research this and propose a solution.

However, and considering the security implications of how OpenSSL is used, if the feature upstream is already contentious and not clear _if_ it would be merged, then it's not good for us to consider accepting it in Conan Center either. The recipes in Conan Center are designed to be built from the source code as it is provided upstream.

Even if it is protected under a non-default option, this will not be checked by our CI. Moving forward we are making every effort that unverified changes, in particular unverified source code changes, make it into the recipes at all. 

For this reason I'm afraid we cannot accept this PR.

Please be aware that with Conan you are able to host your _own_ copy of the recipe, with your own custom changes, to satisfy your team's needs. Please let us know if you would require additional help on this topic.

jcar87:
Added note, from here:  https://stackoverflow.com/questions/26935662/openssl-1-0-2-to-read-md5-ca-certificates
Two solutions are proposed: 
* using your own CA bundle, in a way that it does not use the system ones
* using boringssl - which is actually the library used by Android, IIRC

I'm not very familiar with boringssl, but you may be able to link `curl` against the system boringssl if this is something provided by the Android NDK

sizeak:
Yes thank you, I'm aware of those options. My intention with this PR was to make life easier for anyone else needing to do something similar and avoiding the need to host something custom; using https with curl from native code on Android doesn't seem to be a particularly niche use case.

I understand though that while this change is small and easily verifiable by reading the patch, that a blanket policy is a blanket policy and avoiding source changes makes sense given the recent spate of supply chain attacks.

Thank you for the quick response. We already host our internal Conan packages in an Artifactory instance so it shouldn't be too problematic to host a custom build of OpenSSL.

PR #23577, opened 2024-04-16
leopard: add recipe

toge:
Specify library name and version:  **leopard/2.0.0**

---

### DIFF
--- a/recipes/leopard/all/conanfile.py
+++ b/recipes/leopard/all/conanfile.py
@@ -30,7 +30,7 @@ class LeopardConan(ConanFile):
         return {
             "gcc": "11",
             "clang": "12",
-            "apple-clang": "13",
+            "apple-clang": "14",
             "Visual Studio": "16",
             "msvc": "192",
         }
--- a/recipes/leopard/all/conanfile.py
+++ b/recipes/leopard/all/conanfile.py
@@ -77,4 +77,4 @@ class LeopardConan(ConanFile):
         self.cpp_info.set_property("pkg_config_name", "ThreadPool")
 
         if self.settings.os in ["Linux", "FreeBSD"]:
-            self.cpp_info.system_libs.extend(["pthread"])
+            self.cpp_info.system_libs.extend(["pthread", "dl", "m"])
--- a/recipes/leopard/all/conanfile.py
+++ b/recipes/leopard/all/conanfile.py
@@ -1,0 +1,80 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import get, copy
+from conan.tools.layout import basic_layout
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.52.0"
+
+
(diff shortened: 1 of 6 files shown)


recipes/leopard/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+    def package_info(self):
+        self.cpp_info.bindirs = []
+        self.cpp_info.libdirs = []
+
+        self.cpp_info.set_property("cmake_file_name", "ThreadPool")
+        self.cpp_info.set_property("cmake_target_name", "ThreadPool::ThreadPool")
+        self.cpp_info.set_property("pkg_config_name", "ThreadPool")
+
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs.extend(["pthread"])
```
RubenRBS, line 80: https://github.com/hosseinmoein/Leopard/blob/main/src/Makefile.Linux.GCC64#L16 we might be missing a few others

recipes/leopard/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+            pattern="*.h",
+            dst=os.path.join(self.package_folder, "include"),
+            src=os.path.join(self.source_folder, "include"),
+        )
+        copy(
+            self,
+            pattern="*.tcc",
+            dst=os.path.join(self.package_folder, "include"),
+            src=os.path.join(self.source_folder, "include"),
+        )
```
RubenRBS, line 69: I don't understand what theses files are, they look to be like cpp sources implementing some of the functions in the headers? Do you have any insight?

RubenRBS (commented):
Hi! Thanks a lot. I have a question regarding the .tcc files :)

recipes/leopard/all/conanfile.py
toge: Thanks!
fixed.

recipes/leopard/all/conanfile.py
toge: I think this is an implementation technique.
It looks to be making a distinction between .h files as interface and .tcc files as implementation.

While on development the macro `HMTHRP_DO_NOT_INCLUDE_TCC_FILES` seems to be defined so that tcc can be compiled.

RubenRBS: Oh, that's super interesting, thanks for the extra info, I had never heard of something similar :)

RubenRBS (changes requested):
@uilianries realized a possible name conflict, blocking until we figure this out :) No need for further action on your part

RubenRBS (approved):
Approving, seems like https://github.com/catid/leopard?tab=readme-ov-file#leopard-rs prefers leopard-rs so no issue here :)

uilianries (approved):
There is another Leopard project even more popular: https://github.com/catid/leopard. And it's packaged already in Arch Linux: 
https://aur.archlinux.org/packages/leopard-git

However, I see no problem using this one as Leopard. The another uses the name `Leopard RS` in the readme, so in the future we could add it as `leopard-rs` or something similar.

PR #23578, opened 2024-04-16
libsodium: add emscripten support to the libsodium recipe

laci-aura:
**libsodium/1.0.19**

The `--disable-ssp` option was missing which lead to linking errors due to the missing stack check functions in emscripten.

Also, the libsodium-provided
```
./dist-build/emscripten.sh --standard
```
script results in running:
```
./configure --enable-minimal --disable-shared "--prefix=libsodium-stable/libsodium-js" --without-pthreads --disable-ssp --disable-asm --disable-pie
```
so I've added the missing minimal/pthreads/asm/ssp options to the recipe to mimic the standard emscripten build options

---


jcar87:
Thanks @laci-aura  and apologies for the delay.

We've merged emscripted support as part of https://github.com/conan-io/conan-center-index/pull/26960 - if some options are not valid _at all_ on emscripted, may as well just pass them directly - we prefer this to adding new options. Unless of course there's evidence users would need to toggles, in that case please let us know, thanks!

PR #23580, opened 2024-04-16
capnproto: Simplify testpackage 

RubenRBS:
Same idea as in https://github.com/conan-io/conan-center-index/pull/23573


uilianries:
You could call `capnp` tool directly, so `.capnp` could be preserved. 

https://capnproto.org/capnp-tool.html

recipes/capnproto/all/test_package/CMakeLists.txt
```diff
@@ -3,19 +3,9 @@ project(test_package LANGUAGES CXX)
...
 
-capnp_generate_cpp(addressbookSources addressbookHeaders addressbook.capnp)
-add_executable(addressbook addressbook.c++ ${addressbookSources})
+
+add_executable(addressbook test_package.c++ ${addressbookSources})
 target_include_directories(addressbook PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 target_link_libraries(addressbook PRIVATE CapnProto::capnp)
 target_compile_features(addressbook PRIVATE cxx_std_14)
 
-capnp_generate_cpp(calculatorSources calculatorHeaders calculator.capnp)
```
valgur, line 12: Please try to preserve the `capnp_generate_cpp()` call in CMake, even if its output is not used. It's important to verify that the CMake modules configured in `package()` and `package_info()` are working correctly.
You should be able to use `VirtualRunEnv(self).generate(scope="build")` and a `NOT CMAKE_CROSSCOMPILING` check in CMake for this.

jcar87: Appreciate the feedback but we have decided not to do this, as per the discussion in the protobuf PR.

jcar87 (approved):

recipes/capnproto/all/test_package/CMakeLists.txt
ericLemanissier: @jcar87 didn't you say in https://github.com/conan-io/conan-center-index/pull/23573#issuecomment-2059824846

> If we are concerned that the logic inside protobuf-generate.cmake is not being tested, we can leave the logic to invoke protobuf_generate - perhaps even double check what the resulting command looks like, short of letting cmake invoke protoc during the build. The combination of self.run() and that check would mean that the usage is still covered, and the package is still correct.

uilianries (approved):
Both library and binary are used.

recipes/capnproto/all/test_package/CMakeLists.txt
jcar87: Yes, but that's not what is being suggested here.

Looking at the capnproto recipe, the relevant "custom" functionality is here:
https://github.com/conan-io/conan-center-index/blob/b822a71374d88f3689a32ecde23dbb87b729e939/recipes/capnproto/all/conanfile.py#L178-L190

The rest of the file is already part of the library - and testing the logic inside I think it's beyond the scope of the test_package - that should work as it does in upstream. In this particular case, the logic I've highlighted is already running as part of the resolution of `find_package` - if we are concerned that may not work, we can simply check in the `CMakeLists.txt` of the test_package that `CAPNPC_CXX_EXECUTABLE` and `CAPNP_EXECUTABLE` are found ? that would be as far as I go.

For protobuf is a very similar case, right? the only "Custom" logic is setting up the `protoc::protoc` target - the rest is just _verbatim_ the upstream code of  the library? 

That is, we can test that the bits that the recipe actually touch, result in the correct postconditions on the consumer side, and we do not need to test the upstream CMake code any more than the we need to test the internals of the library.

valgur: Let's not create wholly artificial constructs just to avoid touching upstream code. It does not make any rational sense to me. The cmake code you linked to is extremely janky for starters and definitely needs to be covered by the tests. And even the CMake modules copied verbatim from upstream still need to interface with the output of Conan v1 and v2 generators, which is not without its pitfalls. I struggle to see any measurable gains for the CI load or maintability.

I understand if there are higher priority items in your backlog and you mostly wish to unblock the recipe and move on, but please just say so and I'll gladly follow up on it myself. No need to turn it into a precedent and a best practice on CCI unnecessarily.

jcar87: >  The cmake code you linked to is extremely janky for starters and definitely needs to be covered by the tests. 

It already is called - since it is automatically included when `find_package()` is being called. We can add a check in CMake to make sure those calls succeeded.

> And even the CMake modules copied verbatim from upstream still need to interface with the output of Conan v1 and v2 generators,

I completely disagree, because you can make an argument about testing the entire functionality of the library, and then you can argue to test anything and everything. Our CI service has never been intended to be a "testing" service: https://github.com/conan-io/conan-center-index/blob/b822a71374d88f3689a32ecde23dbb87b729e939/docs/adding_packages/README.md#the-build-service

>  interface with the output of Conan v1 and v2 generators, 

Conan 2 was released over a year ago - the generators are independently tested in the Conan repository. The test_package should highlight issues with the _recipes_ - such that if a test_package fails, we know the fix is in either the test package or the recipe. Otherwise it's a Conan bug - discovering potential Conan bugs, while a nice side effect, is also not the goal of Conan Center.

>  which is not without its pitfalls.
The _only_ thing I can see in both protobuf and capnproto is that the recipe adds a rather thin layer to locate the executable - everything else is from upstream. What other potential pitfalls are we missing that put us at risk of PR breaking the consumers?


> I struggle to see any measurable gains for the CI load or maintability.
`conan-center-index` is forked by a multitude of enterprise users, and we already hace received complaints (in public and in private), that the test packages for these recipes force them to have to hardcode a build order in their CI scripts. So it's not only Conan Center CI maintainability, is that of all of our enterperise users too.

>  No need to turn it into a precedent and a best practice on CCI unnecessarily.

The proposed changes in both this PR and the one in protobuf _are_ the recommended best practice, for the test packages to be as simple as possible.
- For library packages, test the workflow with a consumer conanfile
- For executable packages, run a simple `self.run()` that calls the executable and make sure it comes from the COnan cache and is the right version
- For hybrid packages (like protobuf and captain proto): do both, have a test CMake project consuming the library, and a `self.run()` calling the executable, guarded by a `can_run`. 

Anything beyond this is subject to be corrected in Conan Center, and was _never_ intentional by us.

Some exception could be made for actual issues or as a means of regression testing - but that's subject to us analysing it on a case by case basis.

For reference, this is the "test" stage of the homebrew protobuf package: https://github.com/Homebrew/homebrew-core/blob/648afdc2845af8dce545f65d7c7f60b38bdf5488/Formula/p/protobuf.rb#L52-L63.

jcar87: Bear in mind that until Conan 2.x was released, this wasn't a problem that we had these dual requirements (host and build context). The tests _were_ useful during the time when Conan 2.x was in beta, because they actually highlighted issues/limitations of Conan 2, with the added burden of keeping the recipe backwards compatible with Conan 1.x in both single and dual-profile modes. These complex test packages did play a role... in highlighting things that had to be implemented in the Conan client - and not necessarily in the recipes. 

So what it may appear to you that the patterns in these test packages were best practice, they are actually relatively new constructs, that have only permeated to ~20 recipes, and we are now in a position to correct.

valgur: This getting a bit out of hand. :grin:

I have not once argued against the replacement of `self.tool_requires(self.tested_reference_str)` with a host-context-only one. It makes complete sense to me given the limitations and overhead it produces and I'm grateful for providing a clear fix/workaround for it.

The only point I'm trying to make is that due to Conan ripping out most of the CMake config and the related targets, variables and sometimes macros/functions and replacing them with

PR #23582, opened 2024-04-16
cmake: add version 3.29.2

j-carl:
Specify library name and version:  **cmake/3.29.2**

---

### DIFF
--- a/recipes/cmake/binary/conandata.yml
+++ b/recipes/cmake/binary/conandata.yml
@@ -1,4 +1,23 @@
 sources:
+  "3.29.2":
+    Linux:
+      armv8:
+        url: "https://cmake.org/files/v3.29/cmake-3.29.2-linux-aarch64.tar.gz"
+        sha256: "ca883c6dc3ce9eebd833804f0f940ecbbff603520cfd169ee58916dbbc23c2b8"
+      x86_64:
+        url: "https://cmake.org/files/v3.29/cmake-3.29.2-linux-x86_64.tar.gz"
+        sha256: "0416c70cf88e8f92efcbfe292e181bc09ead7d70e29ab37b697522c01121eab5"
+    Macos:
+      universal:
+        url: "https://cmake.org/files/v3.29/cmake-3.29.2-macos10.10-universal.tar.gz"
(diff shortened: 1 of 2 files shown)


recipes/cmake/binary/conanfile.py
```diff
@@ -16,21 +16,23 @@ class CMakeConan(ConanFile):
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/Kitware/CMake"
     license = "BSD-3-Clause"
-    settings = "os", "arch"
+    settings = "os", "arch", "compiler", "build_type"
```
RubenRBS, line 19: Any insight into why this is added?

jcar87 (changes requested):
The changes in `conanfile.py` are unnecessary and go beyond the PR description. This recipe only has the two settings that are already required (no more, and no less), thus how it is was intentional. 

having `compiler` and `build_type` and removing it in `package_id()` is _only_ needed if a compiler is involved during the build phase, which is not the case here.

recipes/cmake/binary/conanfile.py
j-carl: The build failed on with this warning
> [HOOK - conan-center.py] pre_export(): WARN: [MANDATORY SETTINGS (KB-H070)] The values 'compiler', 'build_type' are missing on 'settings' attribute. Update settings with the missing values and use 'package_id(self)' method to manage the package ID.
> ERROR: [HOOK - conan-center.py] pre_export(): Some checks failed running the hook, check the output

And I addressed it by using the guideline from the error code page https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#kb-h070-mandatory-settings

RubenRBS: Oh, thanks for reading those, we'll make sure to clarify this does not apply for this recipe so that in the future it does not confuse anyone else :)

/cc @jcar87

j-carl: Ok, I will revert the change to `conanfile.py`

j-carl:
@jcar87 I removed the changes to `conanfile.py`.

jcar87:
Thanks a lot! This was also a good exercise to highlight with the issue with the hook giving the wrong advice :D

RubenRBS (approved):

PR #23589, opened 2024-04-17
hictk: add v0.0.12 and drop v0.0.11

robomics:
Specify library name and version:  **hictk/all**

hictk v0.0.11 turned out to be broken due to some targets not being installed properly.
Updating the test_package to include one of hictk's main header files highlights this issue (as in the test package no longer compiles).
I think the best way forward is to entirely drop v0.0.11. Should this not be acceptable, then I can use #ifdef guards to include the old test_package code when version == 0.0.11. 

---


RubenRBS (approved):
Thanks, I think this is ok as-is :)
Note that removal from here does not mean that users can't install the old version, it just won't get updated with new changes!

uilianries (approved):
LGTM

Issue #23558, opened 2024-04-15
[service] macos cross-compilation ordering


ericLemanissier:
### What is your problem/feature request?

comes from https://cpplang.slack.com/archives/C41CWV9HA/p1711022464815949

I think there an ordering problem between conan 2 macos-clang and macos-m1-clang runners (cf https://github.com/conan-io/conan-center-index/pull/23191#issuecomment-2012071902)
In case of a package building an executable, the macos ARM (M1) should be built first, because it is used both when targetting the ARM and X86_64 arch. In other words, native compilation has to be scheduled before cross-compilation, which seems to be sequenced backwards currently.
see also
https://github.com/conan-io/conan-center-index/pull/21622#issuecomment-2011268669
https://github.com/conan-io/conan-center-index/pull/22378#issuecomment-1976439636

@danimtb @RubenRBS @uilianries 

PS: retriggering build never works around this situation, because the order is always wrong.

ericLemanissier:
The team clearly stated that this won't happen. The following recipes now need contributors to fix them and make sure that the test_package does not `tool_requires` the tested reference:
```
libtool/all/test_package/conanfile.py
mbits-lngs/all/test_package/conanfile.py
mfast/all/test_package/conanfile.py
extra-cmake-modules/all/test_package/conanfile.py
grpc/all/test_package/conanfile.py
tree-gen/all/test_package/conanfile.py
flatcc/all/test_package/conanfile.py
btyacc/all/test_package/conanfile.py
protobuf-c/all/test_package/conanfile.py
wayland/all/test_package/conanfile.py
cfgfile/all/test_package/conanfile.py
libtasn1/all/test_package/conanfile.py
ignition-cmake/all/test_package/conanfile.py
gsoap/all/test_package/conanfile.py
flex/all/test_package/conanfile.py
```
(list comes from https://github.com/conan-io/conan-center-index/pull/23594#issuecomment-2060996942)

Qt is not worth changing anything, because the recipe does not support cross-compilation, and won't do in a foreseeable future

RubenRBS:
Thanks for the recipe traceability on what needs fixing @ericLemanissier - appreciated :)

PR #23594, opened 2024-04-17
[config] Re-order Mac M1 configuration

uilianries:
This change puts Mac M1 to be listed first when building packages in ConanCenterIndexCI. It does not guarantee the result as builds are execute in parallel, so a test package could ask for package that's not available yet.

Besides this change, Conan team will check for internal options/fix in C3I Jenkins code.

Related to #23558

---

### DIFF
--- a/.c3i/config_v2.yml
+++ b/.c3i/config_v2.yml
@@ -81,19 +81,6 @@ configurations:
               compiler.libcxx: ["libstdc++11"]
               compiler.version: ["11"]
               build_type: ["Release"]
-  - id: macos-clang
-    hrname: "macOS, Clang"
-    build_profile:
-      os: "Macos"
-      arch: "armv8"
-    content:
-      - os: [ "Macos" ]
-        arch: [ "x86_64" ]
-        compiler:
-          - "apple-clang":
-              compiler.version: [ "13" ]
-              compiler.libcxx: [ "libc++" ]
-              build_type: [ "Release"]
   - id: macos-m1-clang
     hrname: "macOS M1, Clang"
     build_profile:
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):

jcar87 (changes requested):
under review

jcar87:
having taken the decision to simplify the test_packages, I'm not sure this is worth the effort.

The test_packages should have never reached the level of complexity that some have reached. Guaranteeing a build order "would" fix the issue of test_packages failing when they require two package IDs of the same reference (that are built as part of the same PR), but it has a downside that it serialises what can otherwise be parallel. 

Out of 1743 test_packages in the repository, only the following have a dual requirement against the tested reference:
```
libtool/all/test_package/conanfile.py
mbits-lngs/all/test_package/conanfile.py
mfast/all/test_package/conanfile.py
extra-cmake-modules/all/test_package/conanfile.py
grpc/all/test_package/conanfile.py
tree-gen/all/test_package/conanfile.py
flatcc/all/test_package/conanfile.py
btyacc/all/test_package/conanfile.py
protobuf-c/all/test_package/conanfile.py
qt/5.x.x/test_package/conanfile.py
qt/6.x.x/test_package/conanfile.py
wayland/all/test_package/conanfile.py
pkgconf/all/test_package/conanfile.py
cfgfile/all/test_package/conanfile.py
libtasn1/all/test_package/conanfile.py
ignition-cmake/all/test_package/conanfile.py
gsoap/all/test_package/conanfile.py
flex/all/test_package/conanfile.py
```
that's 18, in addition to `protobuf` and `capnproto` which have already been fixed. Guaranteeing a build order by serialising the creation of the binary packages is not worth it only to support these 18 cases vs the 1743 test packages we have currently. Personally I dont think most of the above are justified in having a complex test package, other than Qt.

I think we should refocus our efforts in ensuring that the very valid PRs against protobuf and the likes that are impacted by the issues can be swiftly merged, while in the background we continue our efforts to bring the test packages back in line with the original intention of the test packages.

If this PR doesn't really guarantee that the test_packages will succeed due to race conditions, I wouldn't bother too much.

ericLemanissier:
@jcar87 OTOH this change has no adverse effect, and allows easy progress on all the recipes you listed

jcar87:
> @jcar87 OTOH this change has no adverse effect, and allows easy progress on all the recipes you listed

All the recipes I've listed will have their test package refactored accordingly, should this become a problem.

jwillikers (approved):

Croydon (approved):

uilianries:
Closing as not planned.

jcar87:
After further testing, this did not mitigate the issue.

PR #23602, opened 2024-04-17
cyrus-sasl: patch packaged headers to work with msvc compiler

adamws:
Specify library name and version:  **cyrus-sasl/2.1.28**

When using msvc compiler and `saslplug.h` include, it would fail to compile test_package.
I addressed that problem here: https://github.com/cyrusimap/cyrus-sasl/issues/831 - until solved upstream I think this qualifies for patching here (which this PR does)

---


valgur (approved):

RubenRBS (commented):
Thanks a lot! We really appreciate you opening the issue upstream and linking back there in the code <3

My understanding is that upstream will fix this similarly to how you've implemented here, but let's give them a few days just in case they supprise us, else I'll aprove by monday :)

uilianries (approved):
@adamws Thank you for your PR!

I made a change in your PR, I replaced your patch by the official one from the upstream.

Still, I can confirm your bug, without that patch, my build on Windows fails: [cyrus-sasl-2.1.28-no-patch.log](https://github.com/user-attachments/files/19290607/cyrus-sasl-2.1.28-no-patch.log)

Then, after applying the patch, it works without any error: [cyrus-sasl-2.1.28-patched.log](https://github.com/user-attachments/files/19290612/cyrus-sasl-2.1.28-patched.log)

Regards!

PR #23612, opened 2024-04-17
util-linux-libuuid, make python bindings optional

basiliscos:
**util-linux-libuuid/2.39.2**

I hope this will fix the compilation in debug mode, see https://ci.appveyor.com/project/basiliscos/cpp-rotor/builds/49638104/job/2pg3xasvpoof4h8x

```
checking for ECONF... no
checking whether C:\Python38-x64 version is >= 2... no
configure: error: Python interpreter is too old
util-linux-libuuid/2.39.2: ERROR: 
Package '659f5046bee3c5a8463ca82f0ce7aaba50748d87' build failed
```

i.e. it find some old win32 python, while it is building in linux-on-windows, and than fails.

---


Ahajha:
Minor, I would rename the option to avoid a negative, so `with_python` (or even better, `with_python_bindings`). Then if it's false, then set the flag.

Ahajha (approved):

RubenRBS (commented):
Thanks! My main concern here is that some people might be relying on having these by default - if this is fixing a specific configuration, maybe modifying the setting in the `config_option` method in that case would work best?

basiliscos:
@RubenRBS any news so far? It does not matter what is the default, i.e. on or off, but the option is definitely needed.

valgur (approved):

basiliscos:
ping

basiliscos:
@uilianries ping

uilianries:
Hello @basiliscos! I'm currently on vacation :palm_tree:, please, ping `@conan-io/barbarians` for support. Have a nice day!

basiliscos:
@ErniGH could you approve, please?

recipes/util-linux-libuuid/all/conanfile.py
```diff
@@ -23,10 +23,12 @@ class UtilLinuxLibuuidConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "with_python_bindings": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "with_python_bindings": False,
```
AbrilRBS, line 31:
```suggestion
        "with_python_bindings": True,
```

Unless there's a good reason to, let's try to avoid changing the default value for current users once they get this new revision

AbrilRBS (approved):
Thanks!

ErniGH (approved):
LGTM Thanks for your patience 😄

PR #23642, opened 2024-04-18
libpsl: added version 0.21.5

laci-aura:
Specify library name and version:  **libpsl/0.21.5**

---

### DIFF
--- a/recipes/libpsl/all/conanfile.py
+++ b/recipes/libpsl/all/conanfile.py
@@ -5,6 +5,7 @@ from conan.tools.files import apply_conandata_patches, copy, export_conandata_pa
 from conan.tools.gnu import PkgConfigDeps
 from conan.tools.layout import basic_layout
 from conan.tools.meson import Meson, MesonToolchain
+from conan.tools.scm import Version
 import os
 
 required_conan_version = ">=1.53.0"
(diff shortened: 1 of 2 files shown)


recipes/libpsl/all/conanfile.py
```diff
@@ -76,7 +79,11 @@ def generate(self):
         env.generate()
         tc = MesonToolchain(self)
         tc.project_options["runtime"] = self._idna_option
-        tc.project_options["builtin"] = self._idna_option
+        if Version(self.version) >= "0.21.5":
+            tc.project_options["builtin"] = self.options.enable_builtin
```
RubenRBS, line 83: If this option is only valid from this version onward, it should be removed in the `config_options()` method so as to not influcence the pkgid of older versions :)

RubenRBS: Also, this PR that tried to target an older release https://github.com/conan-io/conan-center-index/pull/22029/files#diff-308e90de1d7646e4db40d401de640dd72698dbb96cb43871ac4fb7bbd8eb8566R77

has it still check for the old ._idna_option. I take that yours is more extensible, but could you confirm what the original upstream intention is? Thanks! :)

laci-aura: No it should not be removed, older versions of the lib used to allow setting the 'builtin' option to the same values as the 'runtime' option (libicu, libidn, etc) but now they changed the meaning/possible values to just True/False.
I preserved compatibility with older versions
https://github.com/conan-io/conan-center-index/blob/41dc055e5210007408ff51c09d02a9e7c1b2ebb7/recipes/libpsl/all/conanfile.py#L86

laci-aura: Sorry took me a while to understand your comment :), fixed in https://github.com/conan-io/conan-center-index/pull/23642/commits/1e108a372847eaad594eab031be84212da3d81ea

franramirez688 (commented):
Hi @laci-aura 

The PR looks great. Could you please rebase your branch from the master one so I can relaunch the CI?

Issue #18564, opened 2023-07-15
[openjdk/19.0.2] : Permission error


MartinDelille:
### Description

When installing openjdk/19.0.2 with another settings (in my case when changing the os.version), I have a permission error during build:

```
...
Downloading openjdk-19.0.2_macos-x64_bin.tar.gz:  99%|#########8| 182Downloading openjdk-19.0.2_macos-x64_bin.tar.gz:  99%|#########9| 183Downloading openjdk-19.0.2_macos-x64_bin.tar.gz: 100%|#########9| 184                                                                     Downloading openjdk-19.0.2_macos-x64_bin.tar.gz completed [188067.37k]                   openjdk/19.0.2:
openjdk/19.0.2:
openjdk/19.0.2:
openjdk/19.0.2: ERROR: Package '4505918cd3c6f93860716a090bf9acb3b037d0ad' build failed
openjdk/19.0.2: WARN: Build folder /Users/martin/.conan/data/openjdk/19.0.2/_/_/build/4505918cd3c6f93860716a090bf9acb3b037d0ad
ERROR: openjdk/19.0.2: Error in build() method, line 35
	get(self, **self.conan_data["sources"][self.version][str(key)],
	PermissionError: [Errno 13] Permission denied: '/Users/martin/.conan/data/openjdk/19.0.2/_/_/source/jdk-19.0.2.jdk/Contents/Home/legal/java.xml.crypto/santuario.md'
make: *** [test] Error 1
```

### Package and Environment Details

* Package Name/Version: **openjdk/19.0.2**
* Operating System+version: **MacOS 13.4.1 (c) (22F770820d)**
* Compiler+version: **N/A**
* Docker image: **N/A**
* Conan version: **conan 1.60.0**
* Python version: **Python 3.10.6**

### Conan profile

[settings]
os=Macos
os.version=10.13
os_build=Macos
arch=x86_64
arch_build=x86_64
compiler=apple-clang
compiler.version=14.0
compiler.libcxx=libc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

```
conan remove "openjdk/*" -f
conan install -r conancenter openjdk/19.0.2@ -s os.version=11.0 -b missing
conan install -r conancenter openjdk/19.0.2@ -s os.version=12.0 -b missing
```

### Logs

<details><summary>Click to expand log</summary>

```
conan remove "openjdk/*" -f
[...]

PR #23644, opened 2024-04-18
Remove `os.version` from package ID and also removing `no_copy_source = True` in openjdk recipe

czoido:
This PR should fix: https://github.com/conan-io/conan-center-index/issues/18564

The issue is that the zip from openjdk contains a file that is decompressed with read only values, and as we have the `no_copy_source = True` the zip is always decompressed in exactly the same location so the second time that you decompress with a subsetting that is different, for example `os.version` the decompressing fails. Removing the subsetting from the package ID and also removing `no_copy_source = True` I think should fix the issue. If you force the build it will be decompressed in a different location than the previous build and when the package ends being the same package ID than the other the older package ID will be removed leaving only one folder.

### DIFF
--- a/recipes/openjdk/all/conanfile.py
+++ b/recipes/openjdk/all/conanfile.py
@@ -21,7 +21,8 @@ class OpenJDK(ConanFile):
     def package_id(self):
         del self.info.settings.compiler
         del self.info.settings.build_type
-
+        self.info.settings.rm_safe("os.version")
+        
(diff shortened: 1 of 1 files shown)


RubenRBS (approved):

toge (approved):

MartinDelille:
Thank you very much for the fix! I 🙏

PR #23646, opened 2024-04-18
Add ICU 75.1

datalogics-staylor:
ICU version 75.1 was just released.

---


jwillikers (commented):
I'm curious whether we need the mingw patch or not for this release.

recipes/icu/all/conandata.yml
```diff
@@ -27,6 +30,8 @@ sources:
     url: "https://github.com/unicode-org/icu/releases/download/release-68-2/icu4c-68_2-src.tgz"
     sha256: "c79193dee3907a2199b8296a93b52c5cb74332c26f3d167269487680d479d625"
 patches:
+  "75go.1":
```
jwillikers, line 33:
```suggestion
  "75.1":
```

jwillikers (changes requested):

recipes/icu/all/conandata.yml
jwillikers: It happens. I usually blame my code editor, whether or not that's what actually caused it ;)

jwillikers (approved):

jwillikers:
According to the [release announcement](https://icu.unicode.org/):

> C++ code now requires C++17 and is being made more robust. 

Looks like that constraint will need to be enforced for version 75 and later. You should be able to pull the logic from the Qt 6 recipe to require this in the `validate` method.

datalogics-staylor:
> According to the [release announcement](https://icu.unicode.org/):
> 
> > C++ code now requires C++17 and is being made more robust.
> 
> Looks like that constraint will need to be enforced for version 75 and later. You should be able to pull the logic from the Qt 6 recipe to require this in the `validate` method.

Makes sense. Thanks and I will do that!

recipes/icu/all/conanfile.py
```diff
@@ -75,6 +75,9 @@ def validate(self):
             if not os.path.exists(str(self.options.dat_package_file)):
                 raise ConanInvalidConfiguration("Non-existent dat_package_file specified")
 
+        if self.settings.compiler.get_safe("cppstd"):
```
jwillikers, line 78: You'll probably also need something like [this](https://github.com/conan-io/conan-center-index/blob/7c80e9e24abaa4963b6cd42eb7dab023d4d256a2/recipes/qt/6.x.x/conanfile.py#L188-L197) and [this](https://github.com/conan-io/conan-center-index/blob/7c80e9e24abaa4963b6cd42eb7dab023d4d256a2/recipes/qt/6.x.x/conanfile.py#L258-L262). That enforces an additional check that the minimum version of the compiler supports C++17.

datalogics-staylor: Thank you for all the help! In case it wasn't obvious, this aspect of conan packages is fairly new to me.

jwillikers: No problem. It's always good to see a new contributor helping out, and I'm looking forward to the new version. I should also point you to the CMake package template, [here](https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py), which you'll probably find really useful.

jwillikers:
Looks like it is just breaking on the `test_package` now.

datalogics-staylor:
> Looks like it is just breaking on the `test_package` now, but only for Conan V1.

Yes, I've been trying to figure out what to do about those failures and I'm afraid I'm coming up short.

recipes/icu/all/conanfile.py
```diff
@@ -58,6 +58,16 @@ def _enable_icu_tools(self):
...
         return not self.conf.get("tools.build:skip_test", default=True, check_type=bool)
 
+    @property
+    def _minimum_compilers_version(self):
+        return {
+            "Visual Studio": "16",
+            "msvc": "192",
+            "gcc": "8",
+            "clang": "9",
+            "apple-clang": "12" if Version(self.version) >= "6.5.0" else "11"
```
jwillikers, line 68: This just needs adapted from the Qt 6 recipe.

```suggestion
            "apple-clang": "11"
```

jwillikers (changes requested):
@datalogics-staylor The issue is due to the `test_package` not having a C++ standard set when being built with Conan V1. You need to add `target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)` to the `CMakeLists.txt` in the `test_package` if the version of ICU is greater 75 or later.

Something like the following will probably work.

```
if (ICU_VERSION VERSION GREATER_EQUAL "75")
else()
    target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
endif()
```

recipes/icu/all/conanfile.py
datalogics-staylor: Thank you so much! This makes perfect sense now that I understand what went wrong. I was having trouble understanding what the difference from Conan 1 to Conan 2 causing this was from just the logs.

jwillikers:
MSVC isn't using the C++17 standard:

> note: 'auto' in non-type template parameters requires at least '/std:c++17'

@RubenRBS What's the best way to go about this?

Issue #9333, opened 2022-02-10
[package] cpython/3.9.7: venvlauncher is not built


ZbigniewRA:
### Package and Environment Details (include every applicable attribute)
  * Package Name/Version: **cpython/3.9.7**
  * Operating System+version: **Windows 10**
  * Compiler+version: **Visual Studio 2019**
  * Docker image: **???**
  * Conan version: **conan 1.45.0**
  * Python version: **Python 3.9.6**

### Conan profile (output of `conan profile show default` or `conan profile show <profile>` if custom profile is in use)
```
[settings]
os=Windows
os_build=Windows
arch=x86_64
arch_build=x86_64
compiler=Visual Studio
compiler.version=16
build_type=Release
[options]
[conf]
[build_requires]
[env]
```

### Steps to reproduce (Include if Applicable)

Python interpreter for Windows is built correctly, and it includes the `venv` package.
Unfortuantely `python -m venv` doesn't work on Windows without the `venvlauncher.exe`/`venvwlauncher.exe`, which should be built as well, and included in the `bin` directory (alongside `python.exe` and `pythonw.exe`).

### Logs (Include/Attach if Applicable)
<details><summary>Click to expand log</summary>

```
Output from "python -m venv MyVenv":

Unable to copy 'C:\\Users\\Zbigniew\\.conan\\data\\cpython\\3.9.7\\_\\_\\package\\d140711d95cc16a85766a8fc3a551dfafe84cf63\\bin\\venvlauncher.exe'
Unable to copy 'C:\\Users\\Zbigniew\\.conan\\data\\cpython\\3.9.7\\_\\_\\package\\d140711d95cc16a85766a8fc3a551dfafe84cf63\\bin\\venvwlauncher.exe'
error : [WinError 2] The system cannot find the file specified
```

</details>

ZbigniewRA:
The problem above is happening when I use default settings for `cpython`, like this:
```
[requires]
cpython/3.9.7

[generators]
cmake_multi
```

However if I add the following option everything works as expected:
```
[options]
cpython:shared=True
```
`venvlauncher.exe` is present, `ensurepip` works, and `venv` works.

If default setting `shared=False` is unsupported on Windows, maybe it can be added to the validation step.

ZbigniewRA:
Conan is downloading in my case:
- for Release: `cpython/3.9.7:6b6a5b8333d2f388bcda263a58a797b73d8e6016`
[...]

PR #23648, opened 2024-04-19
CPython: Use shared by default

Ahajha:

(See https://github.com/python/cpython/issues/110234 for more context, I will summarize below)

CPython static builds (especially in newer versions) are only supported on a best-effort basis. Shared builds are the best supported and recommended way to use it. Notably on Windows, the static build has been broken since 3.10, and extension modules only work in shared mode, so when using CPython as a requirement, it is almost always mandatory that it be built in shared mode (on Windows at least).

Given all of this, I think it makes sense to make shared the default on Windows at least certainly, on other platforms this is less important, but still useful as it is the recommended way.

This will also fix/avoid issues like [this one](https://github.com/conan-io/conan-center-index/issues/9333) which arise from static being the default. I've also seen a few other PRs pop up (now that this recipe is usable) which set shared to true, so that provides some more evidence, plus some places to clean up just a little.

### DIFF
--- a/recipes/cpython/all/conanfile.py
+++ b/recipes/cpython/all/conanfile.py
@@ -44,7 +44,7 @@ class CPythonConan(ConanFile):
         "env_vars": [True, False],  # set environment variables
     }
     default_options = {
-        "shared": False,
+        "shared": True,
         "fPIC": True,
         "optimizations": False,
         "lto": False,


Ahajha:
@RubenRBS @uilianries Can an exception be made to the hook? (please see above for reasoning). If not (but you still agree), perhaps I could trick the hook by using `configure()`?

RubenRBS:
@Ahajha thanks for the PR and the reasoning behind it (Along with the one in Slack!)
We'll bring this up with this the team next Tuesday, we'll let you know how that goes :)

jcar87:
Hi @Ahajha - thanks for this!

I'm actually surprised by this! Not so much on Windows - but I thought that a statically-linked-python-runtime-into-the-interpreter was actually widely used in some distros (Debian based) and Conda?

Historically:
* the statically linked runtime performed better
* loadable modules are expected to *not* record a dependency on `libpython` (that is, do not force the linker to try and load it during `dlopen()`, because the symbols are already provided by the executable)

I think other distros  did not follow the path - apparently there's a flag you can pass at compile time to also get performance benefits for the shared one as well, see this thread: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/NWPVQSKVWDKA75PDEIJNJIFL5C5SJXB2/

Just a curiosity - if the static variant doesn't work on Windows, I would make it an invalid option in `validate()` only on that platform. 
The problem with making the `shared` variant with default for all platforms is that it introduces the risk of the runtime library embedding the contents of all the other library dependencies (static) if no care is taken. Potentially leading to runtime issues loadable modules then need to load the shared variant of those shared libraries.

Ahajha:
@jcar87 Thanks for the context, I was not aware (though also not surprised) that there's more to the story for non-Windows platforms. I'll do some reading and think about that case separately. I think I recall there being a lengthy discussion in nanobind about... something related (see here https://github.com/wjakob/nanobind/pull/360#discussion_r1388675361, may or may not be related).

As for Windows, static builds are already invalid, but I think it provides for a poor user experience for the default state to be invalid. Every single user would have to set it to shared anyways, we might as well set that as the default.

So perhaps maybe it should only be default shared on Windows only?

Ahajha:
@jcar correct my if I'm wrong: So from my rough understanding, the difference between shared and static is that the static package ships the binary with a statically linked libpython. For our purposes, we also need to ship the static library, since it's a "generic" package that can also be used for embedding. The shared package ships an exe linked to libpython dynamically, again with both shipped.

The part that I don't quite understand is where it would be necessary for a C extension to link to libpython, but I'm not sure if that's relevant here.

I think I'm okay with leaving non-Windows platforms as static in this case, after reading about the performance increase that comes with static.

Ahajha:
@valgur I think that's the change I'll end up going with. I'll push that later today.

Ahajha:
@jcar87 Can you take another look at this one?

roalter (approved):
This would completely make sense, over all as cpython is mostly used a tool_requires and the developer will not explicitly enabled shared there.

Voting for including this asap.

PR #23653, opened 2024-04-19
boost: bump deps

robomics:
Specify library name and version:  **boost/all**

---


uilianries:
@robomics Sorry the delay. We are avoiding bump dependencies at first sight as they could result in version conflicts. Boost is not only well used in CCI, but its dependencies are also used in many other recipes. Do you need this bump dependencies to fix a conflict, or you are trying to use the latest version only of those libraries?

We are also working on trying to use version ranges for more projects. I'm taking a look if we can use zstd as well, so it could help to avoid conflicts in the future.

recipes/boost/all/conanfile.py
```diff
@@ -651,14 +651,14 @@ def requirements(self):
         if self._with_bzip2:
             self.requires("bzip2/1.0.8")
         if self._with_lzma:
-            self.requires("xz_utils/5.4.4")
+            self.requires("xz_utils/5.4.5")
```
valgur, line 654: +1 for using `5.4.5`, which is the version used in nearly all other recipes except ones still awaiting migration to Conan 2.0.

robomics:
Thanks for looking into this @uilianries!

When I submitted this PR I was declaring dependencies using `conanfile.txt` in several of my projects, and so this PR was an attempt to address various version conflicts I was facing.
In the mean time I've switched to using `conanfile.py` files, so that I can force/override dependency versions as needed.

Let me know if i should push a commit to use version ranges for zstd and v5.4.5 for xz_utils, as suggested by valgur.

uilianries (approved):
@robomics Sorry for blocking the PR, it was intended to avoid merging it by mistake. 

After talking to Conan team, the number of affected packages is small and we can open separated PRs to fix as well. At time, Boost is largely used in CCI and should continue working with your change.

I updated your PR with few changes:

- Use version range for zstd. Still, it's the first recipe using it.
- Add version range for xz_utils. Not documented yet, but looks very stable and compatible. 

Hope these changes fix your current conflicts.

robomics:
Awesome! Many thanks!

valgur (approved):

PR #23659, opened 2024-04-19
Qt5: fix crash at launch on Android < 23

kambala-decapitator:
Specify library name and version:  **Qt/5.15.13**

Support of Android 21 & 22 is broken in Qt 5.15.13, this patch fixes it.

https://bugreports.qt.io/browse/QTBUG-120627

---

### DIFF
--- a/recipes/qt/5.x.x/conandata.yml
+++ b/recipes/qt/5.x.x/conandata.yml
@@ -117,6 +117,11 @@ patches:
       "patch_file": "patches/5.15.12-fix-macos-cpp-lib-memory-resource.patch"
       "patch_source": "https://codereview.qt-project.org/c/qt/qtbase/+/482392"
       "patch_type": "portability"
+    - "base_path": "qt5/qtbase"
+      "patch_description": "Fix crash at launch on Android < 23"
+      "patch_file": "patches/5.15.13-android-21-22.diff"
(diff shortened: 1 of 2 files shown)


RubenRBS (commented):
Hi! Thanks a lot for providing the patch and the link to upstream's report so that it can be traced back there - Let's wait on upstream's decision on how to fix this, see if they approve the patch, before merging it here too as a backport - feel free to ping me once something new comes up on that side :)

kambala-decapitator:
@RubenRBS upstream bug was simply closed as a duplicate, but apparently the original bug's patch is accessible only to commercial license holders

ericLemanissier (commented):
Thanks for the work @kambala-decapitator, unfortunately I am not able to review this patch, because IIUC it was created by you, and there is no sign of review from qt team.
The actual upstream fix is in https://codereview.qt-project.org/c/qt/tqtc-qtbase/+/473563, which is under a commercial license, so we pretty much have to wait until qt 5.15.14 is released under LGPL, which has to be done one year after the commercial release,so it should be before the end of the may 2024.

kambala-decapitator:
for anyone stumbling on this: this fix didn't make it into .14 fully, only the `getRootWindowInsets()` call was removed. The rest of the patch is still required for Android 21-22.

PR #23662, opened 2024-04-19
[rapidjson] fix for gcc14 compile

hoyhoy:
**rapidjson/1.1.0**

GCC 14 fails to compile rapidjson 1.1.0.  Looks like a syntax error actually.  You can't set a constant member.

---


recipes/rapidjson/all/conandata.yml
```diff
@@ -15,3 +15,7 @@ sources:
   "cci.20200410":
     url: "https://github.com/Tencent/rapidjson/archive/8f4c021fa2f1e001d2376095928fc0532adf2ae6.zip"
     sha256: e6fc99c7df7f29995838a764dd68df87b71db360f7727ace467b21b82c85efda
+patches:
+  "1.1.0":
+    - patch_file: "patches/0001-gcc14.patch"
+      patch_description: "Fix for GCC 14 compiler error in document.h."
```
RubenRBS, line 21: Has this been reported upstream?

RubenRBS (commented):
Hi! Thanks a lot for the contribution, we appreciate it - I don't see an URL in the patch_source field, so it's hard to trace theis back to an upstream issue. Has it been fixed upstream?

recipes/rapidjson/all/conandata.yml
Ahajha: 1.1.0 is actually from 2016, I'm surprised we support it to begin with (though it seems it's used at least). The patch probably isn't relevant by now. Would be worth checking to see if the latest works with GCC 14 though.

hoyhoy: We need this for our CI.  I patched the recipe and the source.  We have a legacy app still using 1.1.0 which was the last official release.

hoyhoy: In fact, this is fixed origin/master.  There's just no released fix for it.

hoyhoy: I just changed the patch to do what was done in rapid json's main.  They just removed the `=` operator altogether.

hoyhoy:
1.1.0 is the latest release.  This now merges a single line (removing the syntax error on an unused line) with rapidjson's upstream to compile on GCC 14.

recipes/rapidjson/all/patches/0001-gcc14.patch
```diff
@@ -0,0 +1,13 @@
...
+index e3e20dfb..b0f1f70b 100644
+--- a/include/rapidjson/document.h
++++ b/include/rapidjson/document.h
+@@ -316,8 +316,6 @@ struct GenericStringRef {
+ 
+     GenericStringRef(const GenericStringRef& rhs) : s(rhs.s), length(rhs.length) {}
+ 
+-    GenericStringRef& operator=(const GenericStringRef& rhs) { s = rhs.s; length = rhs.length; }
+-
+     //! implicit conversion to plain CharType pointer
```
Ahajha, line 11: Looking at the commit history, it seems there was another change immediately after which explicitly disabled the copy assignment operator by making it private, perhaps it would be a good idea to include that change too if we added this? (https://github.com/Tencent/rapidjson/commits/master/?after=ab1842a2dae061284c0a62dca1cc6d5e7e37e346+734, specifically this commit: https://github.com/Tencent/rapidjson/commit/862c39be371278a45a88d4d1d75164be57bb7e2d).

Ahajha (commented):
Would this potentially break builds? I'm thinking no, since if it wouldn't compile anyways then it can't be used, but just checking. If not then LGTM.

jcar87:
What are exactly the relevant commits in the upstream repository? It's typically ok for us to merge a backport when the upstream authors have fixed the issue - provided it doesn't break existing code. On that note and for traceability purposes, a small reproducible example and a copy of the log with the compiler error would also be good. Anything that CI does not validate (gcc14 in this case), we will try and verify ourselves when reviewing

hoyhoy:
It's just the one unused line you need to delete for gcc 14 to work.  If anyone were to have ever used that line, it wouldn't have compiled on any compiler anyway.

Also, I just found a UndefinedBehavior issue with rapidjson as well.    We ended up just forking the fixes to our local conan-center repo since nobody seems to care about it here.  A

nd, honestly, rapidjson probably should just be viewed as deprecated anyway.  They haven't updated a release since 2016.  There are commits still going into the rapidjosn repository, but we're hesitant to bring in a "cci" release to our production build.  We already switched to [nlohmann/json](https://github.com/nlohmann/json) for the most part, but have one legacy app still built on rapidjson.

jcar87:
> There are commits still going into the rapidjosn repository, but we're hesitant to bring in a "cci" release to our production build.

This is the approach that has been already been followed by other maintainers such as [Conda](https://github.com/conda-forge/rapidjson-feedstock/pull/7) and [vcpkg](https://github.com/microsoft/vcpkg/blob/master/ports/rapidjson/vcpkg.json#L3) - precisely to reflect the fact that there is are no new tagged/versioned releases, but development has continued, with bugfixes such as the one requested here. 

I can see that `rapidjson/cci.20230929` already has the fix for this, so I would advise using that instead. We would consider backporting a patch for older versions only when there isn't already a published version by us that contains the fix.

Issue #18780, opened 2023-07-21
[service] PR builds trigger on PR description edits


valgur:
### What is your problem/feature request?

I noticed that all of the PRs that I recently updated the description on for a status update triggered a new build shortly after.
That's not necessarily an issue, but it's probably not something worth spending the limited CI capacity on either. Should this perhaps be disabled?

valgur:
@jcar87

valgur:
@jcar87 The need for more context info in the description is usually proportional to the size of the PR, which in turn usually depends on the complexity of the recipe and the project. And the more complex PRs usually need incremental updates as more bugs and edge cases get uncovered. I guess I'll just re-trigger a multi-hour CI run each time in these cases then?

uilianries:
> ### What is your problem/feature request?
> 
> I noticed that all of the PRs that I recently updated the description on for a status update triggered a new build shortly after. That's not necessarily an issue, but it's probably not something worth spending the limited CI capacity on either. Should this perhaps be disabled?

It's not something we can manage easily, because it's a webhook that triggers the build to Jenkins, but in jenkins side is much harder to filter it.

You still can add a new comment in the PR, reviewers should read anything. Better having a separated comment than an empty description.

valgur:
Should be fixed in the new CI setup: #25772

PR #23678, opened 2024-04-20
libkml: add a patch for C++17 compatibility

valgur:
https://github.com/libkml/libkml/pull/275

### DIFF
--- a/recipes/libkml/all/conandata.yml
+++ b/recipes/libkml/all/conandata.yml
@@ -8,3 +8,7 @@ patches:
     - patch_file: "patches/0002-fix-for-mingw-and-empty-literal-for-vc.patch"
     - patch_file: "patches/0003-export-extern-symbols-for-msvc.patch"
     - patch_file: "patches/0004-minizip-no-crypt-header.patch"
+    - patch_file: "patches/0005-cpp17-compatibility.patch"
+      patch_description: "Rmove unnecessary std::unary_function for C++17 compatibility"
+      patch_type: "portability"
+      patch_source: "https://github.com/libkml/libkml/pull/275"
--- a/recipes/libkml/all/conanfile.py
+++ b/recipes/libkml/all/conanfile.py
@@ -1,8 +1,9 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir, save
-from conan.tools.microsoft import is_msvc, is_msvc_static_runtime
+from conan.tools.microsoft import is_msvc_static_runtime
 import os
 import textwrap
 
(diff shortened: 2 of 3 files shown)


RubenRBS (approved):
Thanks! Feel free to ping me once upstream ahs approved this :)

valgur:
@RubenRBS I'm doubtful the project/fork is actually being maintained. The last release was in 2015.

jcar87:
> @RubenRBS I'm doubtful the project/fork is actually being maintained. The last release was in 2015.

Thanks @valgur  for looking into this.
As per our policy, we do not accept patches to C or C++ where upstream might not. If this is an unmaintained library that does not support C++17, then it should raise an the InvalidConfiguration exception when cppstd=17. We provide libraries built from their sources as-is. By plugging fixes that the upstream maintainers won't, this essentially makes the Conan team responsible for the ongoing maintenance of a library fork.

In general, if the change _truly_ is trivial, please provide more context in the PR description to support this claim. Opening a PR upstream is unfortunately not enough, especially if the authors are known to not respond. We do review upstream PRs/issues and require that the upstream maintainers acknowledge, comment, and confirm whether the proposed solution is what they would include in future versions.

Changing the base class of a C++ class or struct has side effects in the ABI. Conan will not cause a rebuild of consumers_ of `libkml/1.3.0` unless the major or minor components have changed, or other settings have.  So the change proposed in this PR _could_ cause link errors in consumers, provided some conditions were met. I think in this case this would typically be that an public function accepts an object (by value) of the `GetElement` struct. It's unclear to me whether that is the case, but we are not familiar with the codebase to make a call on this.

valgur:
@jcar87 Fair enough.

It would indeed argue that this change is trivial and should not have any effects on the ABI, API or the behavior of the library. The `GetElement` functor is used only locally in an `std::for_each()` and is not declared in any headers, so it won't be linked against even in the same project let alone external libs.

Here's the single usage:
https://github.com/libkml/libkml/blob/916a801ed3143ab82c07ec108bad271aa441da16/src/kml/xsd/xsd_file.cc#L51-L69

jcar87:
On which compiler is this a problem?
The last time this recipe built on CI, it passed with C++17 on both Linux and macOS (with gcc and apple-clang).
https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/20193/1/libkml/1.3.0//summary.json

ericLemanissier:
on msvc : https://github.com/eirikb/proof-of-conan/actions/runs/8830426628/job/24243544036#step:11:521

valgur:
> would have been useful in the description as well

I'll start actually updating descriptions when you fix #18780.

jcar87:
> > would have been useful in the description as well
> 
> I'll start actually updating descriptions when you fix #18780.

Please keep in mind that the team may not review, prioritise and should certainly not merge PRs with incomplete information.  

Even if updating the description results in a new run, the packages are not rebuilt if the commit hasn't changed - it reruns the tests, which should be faster.

jcar87:
> @jcar87 The need for more context info in the description is usually proportional to the size of the PR, which in turn usually depends on the complexity of the recipe and the project.

I would say this PR is a good example of that not being the case. That's why we found it confusing knowing that there _are_ binaries published where C++17 was actually used - we don't know where this fails, why, and if the fixes proposed actually fix what the title is claiming to fix, which turned out not to be the case. Having the information at hand will help the review team get to this quicker.

jcar87:
However this is a distraction. We are asking to provide this information when **opening** the PR.

PR #23680, opened 2024-04-20
flac: ignore DLL_EXPORT define in export.h

valgur:
The Windows export header checks both `FLAC__NO_DLL` and `DLL_EXPORT`, but the latter overly broadly named define is only set by autotools and we are using the CMake to build the project.

Fixes the `unresolved external symbol __imp_FLAC__stream_decoder_new` etc errors due to conflicting `DLL_EXPORT` define being set in 
- #18836


recipes/flac/all/conanfile.py
```diff
@@ -56,6 +56,7 @@ def generate(self):
         tc.variables["BUILD_EXAMPLES"] = False
         tc.variables["BUILD_DOCS"] = False
         tc.variables["BUILD_TESTING"] = False
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
uilianries, line 59: What specific error is it fixing? There is no information in the description.

recipes/flac/all/patches/1.4.2-002-ignore-dll_export-define.patch
uilianries, line 1: It makes sense, found similar fix in other places.

uilianries (approved):
LGTM

RubenRBS (approved):
Approving after going over this with @uilianries, thanks!

PR #23696, opened 2024-04-22
oatpp: add version 1.3.0-latest

toge:
Specify library name and version:  **oatpp/1.3.0-latest**

https://github.com/oatpp/oatpp/compare/1.3.0...1.3.0-latest

---


recipes/oatpp/all/conanfile.py
```diff
@@ -49,6 +50,10 @@ def validate(self):
         if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "5":
             raise ConanInvalidConfiguration("oatpp requires GCC >=5")
 
+    def build_requirements(self):
+        if Version(self.version) >= "1.3.0":
```
RubenRBS, line 54: This does not sadly work as you expect. `Version("1.3.0-latest") < Version("1.3.0")` because of prerelease syntax :(

recipes/oatpp/all/conanfile.py
```diff
@@ -59,6 +64,8 @@ def generate(self):
         if is_msvc(self) and Version(self.version) >= "1.3.0":
             tc.variables["OATPP_MSVC_LINK_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
         tc.generate()
+        venv = VirtualBuildEnv(self)
```
RubenRBS, line 67: We also probably want to set `OATPP_LINK_TEST_LIBRARY` to false in the `CMakeToolchain` here

RubenRBS (commented):
Thanks! The tag is creating some issues, let's think about how we can best tackle this :)

recipes/oatpp/all/conanfile.py
toge: @RubenRBS 
Thanks!
I add `_version` property to fix it.

recipes/oatpp/all/conanfile.py
toge: @RubenRBS 
Thanks a lot!
I add `with_test_library` option and use it's value for OATPP_LINK_TEST_LIBRARY.

perseoGI (approved):
LGTM

valgur (approved):

recipes/oatpp/all/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "1.3.0-latest":
```
perseoGI, line 2:
```suggestion
  "1.3.0_latest":
```
Modify version to make `Version("1.3.0_latest") > Version("1.3.0")`

recipes/oatpp/all/conanfile.py
```diff
@@ -23,12 +24,21 @@ class OatppConan(ConanFile):
...
         "shared": False,
         "fPIC": True,
+        "with_test_library": False,
     }
 
+    @property
+    def _version(self):
+        version = str(self.version)
+        if version.endswith("-latest"):
+            version = version[:-len("-latest")]
```
perseoGI, line 39:
```suggestion
        if version.endswith("_latest"):
            version = version[:-len("_latest")]
```

recipes/oatpp/config.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  "1.3.0-latest":
```
perseoGI, line 2:
```suggestion
  "1.3.0_latest":
```

perseoGI (changes requested):
Modify version to make `Version("1.3.0_latest") > Version("1.3.0")`

DannyLihard:
Hello, can we make the last necessary changes to merge this PR ? I'd like to use this version via Conan, and I intend to create PRs for some of Oat++'s extensions too.

I agree with you about the bad versioning scheme, not helping there. Another solution would be to create a recipe v1.3.1 that points to Oat++'s `1.3.0-latest` tag, now that we know the next official release will be v1.4. While a bit confusing for users (still, the information can easily be found in `conandata.yml`), this would be easier (simply patching the `#define`d version in source code) than modifying the recipe for specific version string suffixes.
Moreover, these modifications would be similar for all extensions' PRs (`oatpp-openssl`, `oatpp-postgresql`, etc.). So the easy `1.3.1` workaround would be the same in those, as opposed as the recipe modification in all of them.

What do you think ?

perseoGI:
The thing is that, given the conan `Version` behavior:

```py
>>> from conan.tools.scm import Version
>>> Version("1.3.0-latest") > Version("1.3.1")
False
>>> Version("1.3.0-latest") > Version("1.3.0")
False
>>> Version("1.3.0-latest") < Version("1.3.0")
True
>>> Version("1.3.0_latest") > Version("1.3.1")
False
>>> Version("1.3.0_latest") > Version("1.3.0")
True
```

If we use `1.3.0-latest` tag, we will have to deal with this problems. Thus, it is better to rename the tag to the `underscore` one @toge

DannyLihard:
> The thing is that, given the conan `Version` behavior:
> 
> ```python
> >>> from conan.tools.scm import Version
> >>> Version("1.3.0-latest") > Version("1.3.1")
> False
> >>> Version("1.3.0-latest") > Version("1.3.0")
> False
> >>> Version("1.3.0-latest") < Version("1.3.0")
> True
> >>> Version("1.3.0_latest") > Version("1.3.1")
> False
> >>> Version("1.3.0_latest") > Version("1.3.0")
> True
> ```
> 
> If we use `1.3.0-latest` tag, we will have to deal with this problems. Thus, it is better to rename the tag to the `underscore` one @toge

That's why I'm proposing to use neither `1.3.0-latest` nor `1.3.0_latest` from the Conan point of view, but `1.3.1` instead.
I think that if we want to merge this PR, we need to either :
- Have the changes you proposed done (use `1.3.0_latest` in Conan).
- Go for the `1.3.1` solution.

Both these solutions would use the `1.3.0-latest` Git tag of the `oatpp` repository behind the scenes anyway.

AbrilRBS (approved):

perseoGI (approved):

perseoGI (approved):

PR #23700, opened 2024-04-22
libftp: recipe improvements

deniskovalchuk:
Specify library name and version:  **libftp/0.5.0**

Hi!

#23687 has added libftp v0.5.0 (thanks, @toge).

As FTPS support has been added in this version, I would like to suggest the
following improvements to the recipe:
1. Update the description.
2. Update the topics.
3. Update the test package:
    - ~~Test that libftp provides OpenSSL routines.~~
    - Avoid network operations.

What do you think?

---

### DIFF
--- a/recipes/libftp/all/conanfile.py
+++ b/recipes/libftp/all/conanfile.py
@@ -12,7 +12,7 @@ required_conan_version = ">=1.53.0"
 
 class LibFTPConan(ConanFile):
     name = "libftp"
-    description = "A cross-platform FTP client library based on Boost.Asio"
+    description = "A cross-platform FTP/FTPS client library based on Boost.Asio"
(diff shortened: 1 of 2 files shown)


toge (approved):
Thank a lot for your contribution!

deniskovalchuk:
As far as I understand, the test package should be compatible with v0.4.1 so
I revert 2a51554.

RubenRBS (approved):
Thanks a lot, we appreciate you taking the time to improve the recipe :)

RubenRBS:
> As far as I understand, the test package should be compatible with v0.4.1 so

You can if-def out code based on the version, but nevertheless I like the current simplicity better! :)

toge (approved):

PR #23712, opened 2024-04-23
refactor: test_package for libjpeg

ErniGH:
* write the minimal test that tests the package (the compiler can find the correct includes, the static linker can resolve symbols, the runtime linker can resolve symbols)
* do not perform heavy logic in CI
* remove binary blob - these will no longer be allowed anywhere in the repository
* Remove the `test_v1_package` and the image - it is not worth backporting the refactor to the legacy test


valgur (approved):
I'm not sure if it's ok to remove test_v1_packages already, but otherwise looks good. Thanks! The test executable was doing way too much, indeed.

recipes/libjpeg/all/test_package/CMakeLists.txt
```diff
@@ -10,11 +10,3 @@ if (MSVC)
     target_compile_definitions(${PROJECT_NAME} PRIVATE _CRT_SECURE_NO_WARNINGS _CRT_NONSTDC_NO_WARNINGS)
 endif()
 
-if(BUILD_TRANSUPP)
-    add_executable(test_transupp test_transupp.c ${LIBJPEG_RES_DIR}/transupp.c)
-    target_link_libraries(test_transupp PRIVATE JPEG::JPEG)
```
jcar87, line 15: we can probably still test this minimally - this appears to be unconditionally included in the recipe.
perhaps just add an `#include "transupp.h"` in `test_package.c` and that's it.

jcar87 (commented):
Just one small comment, but rest is great :D

jcar87:
> I'm not sure if it's ok to remove test_v1_packages already, but otherwise looks good. Thanks! The test executable was doing way too much, indeed.

Thanks @valgur!

One of the goals of this PR is not only to simplify the test package, but to also remove any test that make use of binary files - these will no longer be accepted in any test package or in any PR.  `test_v1_package` are no longer required for newly added recipes (as per [announcement](https://github.com/conan-io/conan-center-index/discussions/17406)) - but they can be removed if they get in the way of maintenance - in this case I see no point in backporting the simplification to the old test package. Perhaps we can word this better in the announcement. What we won't accept and is only reserved to Conan team maintainers, is removal of `test_v1_package` without any reason - the team will deal with the removal at a later date (@ErniGH is a member of the team).

toge (approved):

PR #23714, opened 2024-04-23
Add rotor v0.30

basiliscos:
Specify library name and version:  **rotor/0.30**

I'm library author.

---

### DIFF
--- a/recipes/rotor/all/conanfile.py
+++ b/recipes/rotor/all/conanfile.py
@@ -58,6 +58,8 @@ class RotorConan(ConanFile):
             del self.options.fPIC
         if Version(self.version) < "0.26":
             del self.options.enable_ev
+        if Version(self.version) < "0.30":
+            del self.options.enable_fltk
 
     def configure(self):
         if self.options.shared:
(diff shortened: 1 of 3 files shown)


recipes/rotor/all/conanfile.py
```diff
@@ -32,6 +33,7 @@ class RotorConan(ConanFile):
         "enable_thread": False,
         "multithreading": True,
         "enable_ev": False,
+        "enable_fltk": False,
```
RubenRBS, line 36: This option should be removed in the `config_options` method below as `enable_ev` is, when the version is less than 0.30 so that it does not affect the package if it's not valid

RubenRBS (commented):
Hi! Thanks a lot, this looks good, there's only one minor issue :)

recipes/rotor/all/conanfile.py
```diff
@@ -126,3 +132,7 @@ def package_info(self):
         if self.options.get_safe("enable_ev", False):
             self.cpp_info.components["ev"].libs = ["rotor_ev"]
             self.cpp_info.components["ev"].requires = ["core", "libev::libev"]
+
+        if self.options.get_safe("enable_fltk", False):
+            self.cpp_info.components["fltk"].libs = ["rotor_fltk"]
+            self.cpp_info.components["fltk"].requires = ["core", "fltk::fltk"]
```
RubenRBS, line 138:
```suggestion
            self.cpp_info.components["fltk"].requires = ["core", "fltk::fltk"]
        if self.settings.os in ["Linux", "FreeBSD"]:
            self.cpp_info.components["core"].system_libs.append("m")
```

basiliscos: `rotor` code does not contains any `<cmath>`, it is pulled in by some other dependencies (probably, boost).

toge (approved):

PR #23722, opened 2024-04-23
refactor: test_package for freexl

ErniGH:


recipes/freexl/all/test_package/test_package.c
```diff
@@ -1,317 +1,10 @@
...
 
-  ret = freexl_close(handle);
-  if (ret != FREEXL_OK) {
-    fprintf(stderr, "CLOSE ERROR: %d\n", ret);
-    return -2;
-  }
+int main() {
+    const void *handle;
+    int ret;
+    ret = freexl_open(NULL, &handle);
```
RubenRBS, line 8: This sends `NULL` to `fopen` which is UB afaik, should probably find another function :)

RubenRBS (changes requested):

recipes/freexl/all/test_package/test_package.c
valgur: Passing a nonexistent file path will probably work just fine here. It will simply return an error code, which can be ignored.

franramirez688 (approved):

toge (approved):

RubenRBS (approved):

PR #23730, opened 2024-04-23
Update matio to v1.5.27

tbeu:
Specify library name and version:  **matio/1.5.27**

---


RubenRBS (commented):
Hi! Thanks a lot for taking the time to add the new version, we appreciate it :)

I have a few minor comments about the changes, but the PR is mostly good to go

The new version adds `MATIO_ENABLE_CPPCHECK` but is turned off by default, could you nevertheless add it to the `CMakeToolchain`'s `variables` and set it to `False` in the `generate` method, to ensure new versions in the future dont inadvertedly turn it on without us realizing? Thanks!

RubenRBS:
> I wonder how is this possible at all? Do we need to care?

Good catch, I'd say let's ignore that by now

RubenRBS (approved):
Thanks a lot, we really appreciate you taking the time to open the PR, and thanks for such a quick review response! :)

toge (approved):

PR #23731, opened 2024-04-23
Update laszip to v3.4.4

tbeu:
Specify library name and version:  **laszip/3.4.4**

This PR also reflects the license change of LASzip.

---


RubenRBS (dismissed):

recipes/laszip/all/conanfile.py
```diff
@@ -35,6 +35,8 @@ def config_options(self):
             del self.options.fPIC
 
     def configure(self):
+        if Version(self.version) >= "3.4.4":
+            self.license = "Apache-2.0"
```
tbeu, line 39: Ah, thanks. The license change is hidden in the configure method and thus less prominent than setting the class variable directly. Did not know about this option.

uilianries (changes requested):
The version 3.4.4 adds C++17 as requirement. The current recipe is not prepared to work with that, for instance, running older and incompatible compiler will ask for C++17 and will break.

https://github.com/LASzip/LASzip/compare/3.4.3...3.4.4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR4

As hotfix, I would suggest updating `validate()`, based on the version and checking for C++17. As reference, you can use this recipe: https://github.com/conan-io/conan-center-index/blob/a5c52a107e77f3f7a860d72a8a41d768f8fcedb2/recipes/osmanip/all/conanfile.py#L70

tbeu:
Did you try to remove the C++17 requirement in the CMakeLists.txt via patch? I still could compile it on MSVC with /std:c++14 set.

uilianries:
> Did you try to remove the C++17 requirement in the CMakeLists.txt via patch? I still could compile it on MSVC with /std:c++14 set.

@tbeu In case the project is C++14 compatible, then you should take this information from upstream to be sure. All packages in ConanCenterIndex should follow as close as possible the upstream behavior, to avoid any surprises. So far, I only found a reference about moving to C++17 (https://github.com/LASzip/LASzip/commit/cafdc6bb1ab16d42d1ba8c1e0b10bbe3bfe6364a).

tbeu:
It's at least resolved upstream by https://github.com/LASzip/LASzip/commit/0a209056e5616eae54e8f39c8eaea3830fb92aa6.

Not sure about any new release though. Should it be patched here?

uilianries (approved):
LGtM

perseoGI (approved):
LGTM

PR #23733, opened 2024-04-23
up-cpp: new recipe

PLeVasseur:
Specify library name and version:  **up-cpp/0.1.2-dev**

Hi there :wave: 

I contribute to the [Eclipse uProtocol](https://github.com/eclipse-uprotocol) project for developing a Software Defined Vehicle (SDV) middleware to knit together compute within the vehicle, cloud, mobile. We'd like to be able to host the `up-cpp` package on CCI since it seems like the reasonable first choice.

Forgive me as I've not contributed before I'm not entirely sure I've done everything correctly :sweat_smile: 

I left a few questions and TODOs in the conanfile.py as guidance to where I'm unsure in case the recipe is unable to build in certain spots. Will remove those once I can confirm the recipe passes as-is.

---

### DIFF
--- a/recipes/up-cpp/all/conandata.yml
+++ b/recipes/up-cpp/all/conandata.yml
@@ -1,0 +1,10 @@
+sources:
+  # Newer versions at the top
+  "0.1.2-dev":
+    url: "https://github.com/eclipse-uprotocol/up-cpp/archive/refs/tags/v0.1.2-dev.tar.gz"
+    sha256: "1c1006256d0e5fd385732767e0e3b0b4867342e29e4c74e11a0d8a9a763c1288"
+proto:
+  "0.1.2-dev":
+    ssh: "git@github.com:eclipse-uprotocol/up-core-api.git"
+    https: "https://github.com/eclipse-uprotocol/up-core-api.git"
+    tag: "uprotocol-core-api-1.5.7"
--- a/recipes/up-cpp/all/conanfile.py
+++ b/recipes/up-cpp/all/conanfile.py
@@ -1,6 +1,6 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
-from conan.tools.build import check_min_cppstd
+from conan.tools.build import check_min_cppstd, stdcpp_library
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rm, rmdir, collect_libs
@@ -89,6 +89,10 @@ class UpCppConan(ConanFile):
             )
         if is_msvc(self) and self.options.shared:
             raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
+        self.output.info(f"self.settings.compiler: {self.settings.compiler}")
+        self.output.info(f"stdcpp_library(self): {stdcpp_library(self)}")
+        if self.settings.compiler == "clang" and stdcpp_library(self) == "c++":
+            raise ConanInvalidConfiguration(f"{self.ref} Experiencing some interference between math.h and the protobuf generated udiscovery.pb.h when compiling with libc++ for some reason.")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
--- a/recipes/up-cpp/all/conanfile.py
+++ b/recipes/up-cpp/all/conanfile.py
@@ -87,8 +87,8 @@ class UpCppConan(ConanFile):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
-        if is_msvc(self) and self.options.shared:
-            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
+        if is_msvc(self):
+            raise ConanInvalidConfiguration(f"{self.ref} need to have paths in CMake files updated to handle both \ and /.")
         self.output.info(f"self.settings.compiler: {self.settings.compiler}")
         self.output.info(f"stdcpp_library(self): {stdcpp_library(self)}")
         if self.settings.compiler == "clang" and stdcpp_library(self) == "c++":
--- a/recipes/up-cpp/all/conanfile.py
+++ b/recipes/up-cpp/all/conanfile.py
@@ -127,7 +127,7 @@ class UpCppConan(ConanFile):
         apply_conandata_patches(self)
 
     def build(self):
-        self._patch_sources()  # It can be apply_conandata_patches(self) only in case no more patches are needed
+        self._patch_sources()
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
@@ -137,7 +137,6 @@ class UpCppConan(ConanFile):
         cmake = CMake(self)
         cmake.install()
 
-        # some files extensions and folders are not allowed. Please, read the FAQs to get informed.
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
         rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
         rmdir(self, os.path.join(self.package_folder, "share"))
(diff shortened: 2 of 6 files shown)


PLeVasseur:
I noticed in a previous attempt to package up-cpp when we were using a git submodule that was [seen as unacceptable](https://github.com/conan-io/conan-center-index/pull/22684#issuecomment-1930247543).

I read that it's possible to instead use the Conan Git helper and clone in a certain tag based on our version in `conandata.yml`. I've implemented it that way in the current PR.

Would that suffice?

PLeVasseur:
Hey folks :wave: 

Is it possible to receive a review to know what I'll need to fix / update? Thanks for your support!

-Pete

recipes/up-cpp/all/conandata.yml
uilianries, line 1: @PLeVasseur Could you please explain why do you need 2 different packages here? What are they (e.g library and compiler)?

We don't use Git feature for download, only the tag directly, because the source should be immutable.

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+    def _min_cppstd(self):
+        return 14
+
+    # unclear on what I should choose for msvc and Visual Studio
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "13",
+            "clang": "13",
+            "gcc": "11",
```
uilianries, line 52: GCC 6 supports C++14 already, and other compilers also can be lowered: https://en.cppreference.com/w/cpp/compiler_support/14

Or, is there a real feature required?

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+            "msvc": "191",
+            "Visual Studio": "15",
+        }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
```
uilianries, line 59:
```suggestion
```

No patches to be applied, so not needed :)

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+        self.folders.build = os.path.join("build", str(self.settings.build_type))
+        self.folders.install = "install"
```
uilianries, line 71:
```suggestion
```
Please, keep the default layout.

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def build_requirements(self):
+        self.build_requires("cmake/[>=3.18.0]")
+
+    def requirements(self):
+        self.requires("protobuf/3.21.12")
+        self.requires("spdlog/1.13.0")
+        if self.options.build_testing:
+            self.requires("gtest/1.14.0")
```
uilianries, line 80:
```suggestion
        if self.options.build_testing:
            self.requires("gtest/1.14.0")
```

Testing is not something really supported, because the idea is packaging the project, not acting like a build tool. Still, in case you really want to be able to run unit tests, please, use the follow approach: https://github.com/conan-io/conan-center-index/blob/master/recipes/tree-gen/all/conanfile.py#L33 It will not affect the package ID, which is really important.

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+        if is_msvc(self):
+            raise ConanInvalidConfiguration(
+                f"{self.ref} need to have paths in CMake files updated to handle both \ and /."
```
uilianries, line 92: What do you mean with this comment? Does up-cpp support Windows? In case is a bug related how Conan manages the path, you still can replace "\\" by "/", for instance.

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+            )
+        if self.settings.compiler == "clang" and stdcpp_library(self) == "c++":
+            raise ConanInvalidConfiguration(
```
uilianries, line 95: Do you have a build log with that error? It would be nice pointing it as a comment in the recipe. For the message, it can be simplified like:
f"{self.ref} can not be built with clang and libc++ due math.h errors"

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+        # we clone protos from the appropriate repo at the appropriate tag to our local root to build
+        rmdir(self, self.proto_containing_folder)
+
+        git = Git(self)
+        git_method = os.getenv("GIT_METHOD", "https")
+        git.clone(self.conan_data["proto"][self.version][git_method], target=self.proto_containing_folder)
+        git.folder=self.proto_containing_folder
+        git.checkout(self.conan_data["proto"][self.version]["tag"])
```
uilianries, line 110:
```suggestion
        git = Git(self)
        git_method = os.getenv("GIT_METHOD", "https")
        git.clone(self.conan_data["proto"][self.version][git_method], target=self.proto_containing_folder)
        git.folder=self.proto_containing_folder
        git.checkout(self.conan_data["proto"][self.version]["tag"])
```
Need more detail about it, we don't use Git in ConanCenterIndex, instead, we download the tar.gz to be sure about the immutability of the package.

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["PROTO_PATH"] = os.path.join(self.proto_containing_folder, os.path.join(self.proto_subfolder, self.proto_folder))
+        tc.variables["BUILD_TESTING"] = self.options.build_testing
+        tc.variables["BUILD_UNBUNDLED"] = self.options.build_unbundled
+        tc.variables["BUILD_SHARED_LIBS"] = self.options.shared
```
uilianries, line 117:
```suggestion
```
Automatically handled by CMakeToolchain: https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#extending-and-advanced-customization

recipes/up-cpp/all/conanfile.py
```diff
@@ -0,0 +1,159 @@
...
+        self.cpp_info.set_property("cmake_file_name", "up-cpp")
+        self.cpp_info.set_property("cmake_target_name", "up-cpp::up-cpp")
+        self.cpp_info.set_property("pkg_config_name", "up-cpp")
+        self.cpp_info.libs = collect_libs(self)
+
+        self.cpp_info.set_property("cmake_target_name", "up-cpp::up-cpp")
+        self.cpp_info.requires = ["spdlog::spdlog", "protobuf::protobuf"]
+
+        self.cpp_info.names["cmake_find_package"] = "up-cpp"
+        self.cpp_info.names["cmake_find_package_multi"] = "up-cpp"
```
uilianries, line 159:
```suggestion
```
All these configuration is generated by default ;)

recipes/up-cpp/all/test_package/conanfile.py
```diff
@@ -0,0 +1,29 @@
...
+
+# It will become the standard on Conan 2.x
+class TestPackageConan(ConanFile):
+    settings = "os", "arch", "compiler", "build_type"
+    generators = "CMakeDeps", "CMakeToolchain", "VirtualRunEnv"
+    test_type = "explicit"
+
+    def requirements(self):
+        self.requires("protobuf/3.21.12")
+        self.requires("spdlog/1.13.0")
```
uilianries, line 15:
```suggestion
```
Transitive dependencies are managed by Conan already.

uilianries (changes requested):
@PLeVasseur Thank you for your first contribution, it looks an interesting project. Please, take a look in my review and please, try to clarify the proto part in this project, is not clear why is it needed and if should be added as a separated package.

PLeVasseur:
Hey @uilianries :wave: I wanted to follow up here. My employer has decided to part ways with many staff, myself included. I was working on Eclipse uProtocol as my "day job". My ability to contribute to Eclipse uProtocol is reduced at the moment. I'd propose closing this for now and revisiting when myself or another maintainer has bandwidth.

uilianries:
Hello @PLeVasseur! I'm currently on vacation :palm_tree:, please, ping `@conan-io/barbarians` for support. Have a nice day!

PR #23735, opened 2024-04-23
assimp: add v5.4.1

valgur:
https://github.com/assimp/assimp/releases/tag/v5.4.0

No new formats or exporters were added in this release.

### DIFF
--- a/recipes/assimp/5.x/conanfile.py
+++ b/recipes/assimp/5.x/conanfile.py
@@ -2,6 +2,7 @@ from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import stdcpp_library, check_min_cppstd
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import apply_conandata_patches, collect_libs, copy, export_conandata_patches, get, replace_in_file, rmdir, save
 from conan.tools.microsoft import is_msvc
 from conan.tools.scm import Version
@@ -205,6 +206,10 @@ class AssimpConan(ConanFile):
         if Version(self.version) < "5.3.0" and self._depends_on_clipper and Version(self.dependencies["clipper"].ref.version).major != "4":
             raise ConanInvalidConfiguration("Only 'clipper/4.x' is supported")
 
+    def build_requirements(self):
+        if Version(self.version) >= "5.4.0":
+            self.tool_requires("cmake/[>=3.22 <4]")
+
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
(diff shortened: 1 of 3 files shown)


EstebanDugueperoux2 (approved):

RubenRBS:
Should we ensure we don't get bitten by changes like https://github.com/assimp/assimp/compare/v5.3.1...v5.4.0#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR140 in the future? `ASSIMP_BUILD_ZLIB` changed value but we don't set it in the generate method

valgur:
@RubenRBS Thanks for catching that. I added explicit handling of all relevant CMake options.

sunbubble:
https://github.com/assimp/assimp/releases/tag/v5.4.1

5.4.1 has just been released and addresses some Draco relating issues. Could it be related to the warnings emitted above?

sunbubble:
@valgur thanks for bumping this so fast. Out of interest, do you know whether the warnings in the build is what's holding up the merging of this PR? Or is just the queue taking its time?

valgur:
@sunbubble The queue of open PRs is simply very long (https://github.com/ericLemanissier/conan-center-index-pending-review/issues/3), although simple PRs like this one usually get reviewed faster.
I think the Conan devs are currently focusing on a major overhaul of the CCI CI system and Conan itself, so reviews can be somewhat slow.

recipes/assimp/5.x/conanfile.py
```diff
@@ -298,6 +318,8 @@ def _patch_sources(self):
         ]:
             save(self, os.path.join(self.source_folder, "contrib", contrib_header),
                  f"#include <{include}>\n")
+        if Version(self.version) >= "5.4.0":
+            rmdir(self, self.source_path.joinpath("contrib", "utf8cpp"))
```
RubenRBS, line 322: This seems to only be used in samples and there's a change in the release mentioning this, but why is it only now removed? (I might have checked the wrong diff now that I think about it!)

RubenRBS (commented):
Thanks a lot @valgur :)
Minor question, otherwise looks good

recipes/assimp/5.x/conanfile.py
valgur: Oh, it's probably unnecessary in that case then. Let me see if it builds ok without this...

valgur: It was necessary. :slightly_smiling_face:

RubenRBS: Thanks for double-checking :)

RubenRBS:
I'm seeing some `assimp/5.4.1: WARN: replace_in_file didn't find pattern '-fPIC' in '/Users/jenkins/workspace/prod-v2/bsr/38132/eabea/p/b/assimcb1569f627615/b/src/code/CMakeLists.txt' file.` in https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23735/3-macos-m1-clang/assimp/5.4.1//76e545471fe165ddd092b4cbb0361627c0f28db3-build.txt but it seems it's still there! https://github.com/assimp/assimp/blob/942518a4ff148ec10c62c425825c8186eec7fc65/CMakeLists.txt#L311

Are we thus not patching things properly?

valgur:
@RubenRBS The logic was correct, but I cleaned up the patching to avoid the use of `strict=False` nevertheless.

RubenRBS (approved):
Thanks!

toge (approved):

PR #23745, opened 2024-04-24
refactor: open62541 test_package

ErniGH:


recipes/open62541/all/test_package/CMakeLists.txt
```diff
@@ -4,31 +4,8 @@ project(test_package LANGUAGES C)
...
 find_package(Threads REQUIRED)
 
-find_package(Python3 REQUIRED)
-set(PYTHON_EXECUTABLE ${Python3_EXECUTABLE})
-
-# Output directory for Nodeset Compiler
-set(GENERATE_OUTPUT_DIR "${CMAKE_BINARY_DIR}/src_generated/")
-file(MAKE_DIRECTORY "${GENERATE_OUTPUT_DIR}")
-include_directories("${GENERATE_OUTPUT_DIR}")
-ua_generate_nodeset_and_datatypes(
```
jcar87, line 14: This check is a little more special
What does it use to "generate"? I'm assuming it generates files from the `csv` and `bsd` files? in that case - we need to identify how they are generated - and attempt to replicate this in the `test_package` as well (like we did recently for protobuf)

recipes/open62541/all/test_package/test_package.c
```diff
@@ -1,75 +1,11 @@
...
 
-UA_Boolean running = false;
-
-static void server_stop_callback(UA_Server *server, void *data) {
-    UA_LOG_INFO(UA_Log_Stdout, UA_LOGCATEGORY_SERVER, "stopping server due to timer");
-    running = false;
-}
-
 int main(void) {
     UA_Server *server = UA_Server_new();
```
jcar87, line 10: is there a "delete" function that handles the pointer cleanup? If there is one, I would call it to free the pointer. I think OSs will free the resources upon process termination, but I would still do the minimal thing.

RubenRBS, line 10:
```suggestion
    UA_Server *server = UA_Server_new();
    UA_Server_delete(server);
```

RubenRBS (changes requested):

RubenRBS:
The LTO errors I think go outside the scope of the test package changes, I'd focus on the rest of the PRs and we can take a look at it at some point later :)

AbrilRBS:
Already done in a different PR - the loggin option fix remains to be applied but it can wait until something worthwhile of a new revision arrives

PR #23746, opened 2024-04-24
openblas: add v0.3.27, recipe cleanup, add options

valgur:
https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.27

I also 
* cleaned up some messier parts of the recipe (mostly badly-formatted patching),
* added `build_relapack` and `use_locking` options,
* set CMP0077 to ensure that fPIC is handled correctly.

### DIFF
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -4,7 +4,7 @@ from conan.tools.apple import fix_apple_shared_install_name
 from conan.tools.build import cross_building
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 from conan.tools.files import copy, get, replace_in_file, rmdir
-from conan.tools.microsoft import is_msvc_static_runtime
+from conan.tools.microsoft import is_msvc_static_runtime, is_msvc
 from conan.tools.scm import Version
 import os
 import textwrap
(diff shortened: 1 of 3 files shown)


recipes/openblas/all/conanfile.py
```diff
@@ -64,100 +78,108 @@ def validate(self):
...
 
-    def source(self):
-        get(self,
-            **self.conan_data["sources"][self.version],
-            strip_root=True,
-            destination=self.source_folder
-        )
+        if self.options.build_relapack:
+            if not self.options.build_lapack:
+                raise ConanInvalidConfiguration("build_relapack option requires build_lapack")
```
RubenRBS, line 83:
```suggestion
                raise ConanInvalidConfiguration(f'"{self.name}/*:build_relapack=True" option requires "{self.name}/*:build_lapack=True"')
```

After being bitten a few times by unclear messages in the validate message, from now I'd appreciate these kind of messages that ask the user to set a specific option to be abkle to copy&paste directly form the message

recipes/openblas/all/conanfile.py
```diff
@@ -64,100 +78,108 @@ def validate(self):
...
-            **self.conan_data["sources"][self.version],
-            strip_root=True,
-            destination=self.source_folder
-        )
+        if self.options.build_relapack:
+            if not self.options.build_lapack:
+                raise ConanInvalidConfiguration(f'"{self.name}/*:build_relapack=True" option requires "{self.name}/*:build_lapack=True"')
+            if self.settings.compiler not in ["gcc", "clang"]:
+                # ld: unknown option: --allow-multiple-definition on apple-clang
+                raise ConanInvalidConfiguration("build_relapack option is only supported for GCC and Clang")
```
RubenRBS, line 86:
```suggestion
                raise ConanInvalidConfiguration(f'"{self.name}/*:build_relapack=True" option is only supported for GCC and Clang')
```

RubenRBS: Asking the team if this looks alright for them when the message is not asking the user to set the option by itself

RubenRBS (approved):

toge (approved):

PR #23766, opened 2024-04-26
thorvg: New 0.13.2 version and several improvements

franramirez688:
* Added new patch version: `0.13.2`
* Added option `with_simd` (legacy option `vector` renamed to `simd` upstream)
* Added explicit requirements.

cc @toge


toge (approved):
Thanks!

recipes/thorvg/all/conanfile.py
```diff
@@ -77,6 +80,8 @@ def _compilers_minimum_version(self):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+        if Version(self.version) > "0.13.0":
```
RubenRBS, line 83: But does the `with_simd` still have an effect in old versions? What happens if their values differs in versions where both exist?

RubenRBS (commented):
One question about the handling of the new option, but looks good overall. Thanks for adding option descriptions! :)

recipes/thorvg/all/conanfile.py
franramirez688: The thing is that it does not exist in the old versions. Perhaps, it is not clear in the changelog 😓
In summary:

* Versions < 0.13.0 -> `vector`
* Versions >= 0.13.1 -> ~`vector`~ `simd`

franramirez688: See here https://github.com/thorvg/thorvg/compare/v0.13.0...v0.13.1#diff-76ea9078099f2ee8cc0e760a874dd153f225ba16ad10375c24ab6cd53518f3c4L24

RubenRBS (dismissed):

recipes/thorvg/all/conanfile.py
```diff
@@ -54,7 +56,8 @@ class ThorvgConan(ConanFile):
         "loaders": "Enable File Loaders in thorvg",
```
valgur, line 56: The option names in `options_description` should be adjusted to match their names in `options` and `default_options`.

recipes/thorvg/all/conanfile.py
```diff
@@ -33,7 +33,8 @@ class ThorvgConan(ConanFile):
         "with_bindings": [False, 'capi', 'wasm_beta'],
         "with_tools": [False, 'svg2tvg', 'svg2png', 'lottie2gif', 'all'],
         "with_threads": [True, False],
-        "with_vector": [True, False],
+        "with_vector": [True, False],  # removed in thorvg 0.13.1. Renamed to simd
+        "with_simd": [True, False],  # legacy with_vector
```
valgur, line 37: I would mark the legacy option as deprecated (set the default value to `"deprecated"` and log a warning if it is being set).

franramirez688: But this case is different. This case is renamed so it is not a "deprecated" issue at all. I think that putting "deprecated" is a bit confusing as it does not exist anymore.

RubenRBS (approved):

toge (approved):

PR #23767, opened 2024-04-26
md4qt: bump version

igormironchik:
Specify library name and version:  **md4qt/2.8.2**

I'm the author of this library.

---

### DIFF
--- a/recipes/md4qt/all/conanfile.py
+++ b/recipes/md4qt/all/conanfile.py
@@ -60,7 +60,10 @@ class Md4QtConan(ConanFile):
         pass
 
     def package(self):
-        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        if Version(self.version) <= "2.8.1":
+            copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        else:
+            copy(self, "*.txt", src=os.path.join(self.source_folder, "LICENSES"), dst=os.path.join(self.package_folder, "licenses"))
         copy(self, "*.hpp", src=os.path.join(self.source_folder, "md4qt"), dst=os.path.join(self.package_folder, "include", "md4qt"))
 
     def package_info(self):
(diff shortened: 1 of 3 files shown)


recipes/md4qt/all/conanfile.py
```diff
@@ -60,7 +60,10 @@ def build(self):
         pass
 
     def package(self):
-        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        if Version(self.version) <= "2.8.1":
+            copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        else:
+            copy(self, "*.txt", src=os.path.join(self.source_folder, "LICENSES"), dst=os.path.join(self.package_folder, "licenses"))
```
RubenRBS, line 66: Hi! I'd like to understand this change a bit better upstream.

What license is the library now licensed under?

igormironchik: Hi. The license wasn't changed, it's MIT still. But to follow REUSE SOFTWARE any file in the sources was licensed, so in in the LICENSES directory there are some additional licenses for that files.

igormironchik:
@uilianries, @RubenRBS Pay attention, please.

recipes/md4qt/all/conanfile.py
uilianries: >  The license wasn't changed, it's MIT still.

Would not be better only copying everything then?

```
copy(self, "*.txt", src=os.path.join(self.source_folder, "LICENSES"), dst=os.path.join(self.package_folder, "licenses"))
```

igormironchik: > Would not be better only copying everything then?

There is no more `LICENSE` file since version 2.8.2, there are license files in `LICENSES` folder. There are `BSD-2-Clause.txt`,  `CC-BY-4.0.txt`,  `CC-BY-SA-4.0.txt`,  `MIT.txt` in `LICENSES` folder now.

uilianries: @igormironchik There is no problem try to copying LICENSE even when it does not exist, Conan will not raise an error.

>  In case of Conan maybe they are not necessary, I give to user only headers of the library, and not the benchmark, converter to HTML application.

Okay, it clarifies the change. Thank you!

uilianries (approved):
LGTM, the author not only justified the change but explained us why packaging only MIT.txt

RubenRBS (approved):

PR #23777, opened 2024-04-26
Add think-cell-library

RubenRBS:
Specify library name and version:

### DIFF
--- a/recipes/think-cell-library/all/conanfile.py
+++ b/recipes/think-cell-library/all/conanfile.py
@@ -1,0 +1,93 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get
+from conan.tools.layout import basic_layout
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.52.0"
+
+
+class PackageConan(ConanFile):
+    name = "think-cell-library"
+    description = "This library consists of several core C++ utilities that we at think-cell Software have developed and consider to be useful."
+    # Use short name only, conform to SPDX License List: https://spdx.org/licenses/
+    # In case it's not listed there, use "LicenseRef-<license-file-name>"
+    license = "BSL-1.0 license"
+    url = "https://github.com/conan-io/conan-center-index"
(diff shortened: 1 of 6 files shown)


recipes/think-cell-library/all/conanfile.py
```diff
@@ -0,0 +1,73 @@
...
+            "gcc": "12",
+            "msvc": "191",
+            "Visual Studio": "15",
+        }
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("boost/1.84.0")
```
uilianries, line 42:
```suggestion
        self.requires("boost/1.84.0", transitive_headers=True, transitive_libs=True)
```
Or does package_type=header_library exposes automatically?

recipes/think-cell-library/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,147 @@
...
+	{
+		print(tc::as_dec(*it), ", ");
+	}
+	print("\n");
+
+	print("-- Done ----------\n");
+	std::fflush(stdout);
+
+	return EXIT_SUCCESS;
+}
```
uilianries, line 147:
```suggestion
#include "tc/range/meta.h"
#include "tc/range/filter_adaptor.h"
#include "tc/string/format.h"
#include "tc/string/make_c_str.h"

#include <vector>
#include <cstdio>
#include <cstdlib>

namespace {

template <typename... Args>
void print(Args&&... args) noexcept {
	std::printf("%s", tc::implicit_cast<char const*>(tc::make_c_str<char>(std::forward<Args>(args)...)));
}

int main(void) {
	std::vector<int> v = {1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20};

	tc::for_each(
		tc::filter(v, [](const int& n){ return (n%2==0);}),
		[&](auto const& n) {
			print(tc::as_dec(n), ", ");
		}
	);
	print("\n");
	return EXIT_SUCCESS;
}
```

recipes/think-cell-library/all/conanfile.py
RubenRBS: It does! https://docs.conan.io/2/reference/conanfile/methods/requirements.html#default-traits-for-each-kind-of-requires

PR #23787, opened 2024-04-27
pybind11_json: relax version requirements

mologie:
Specify library name and version:  **pybind11_json/all**

The pybind11_json recipe has outdated and overly strict requirements. This relaxes them. I have also lowered the minimum version required to that set by the original author before Conan 2 migration, since there were no code changes to require them.

---

### DIFF
--- a/recipes/pybind11_json/all/conanfile.py
+++ b/recipes/pybind11_json/all/conanfile.py
@@ -23,8 +23,8 @@ class Pybind11JsonConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("nlohmann_json/3.11.2")
-        self.requires("pybind11/2.10.4")
+        self.requires("nlohmann_json/[>=3.9 <4]")
+        self.requires("pybind11/[>=2.6 <3]")
 
     def package_id(self):
         self.info.clear()


recipes/pybind11_json/all/conanfile.py
```diff
@@ -23,8 +23,8 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("nlohmann_json/3.11.2")
-        self.requires("pybind11/2.10.4")
+        self.requires("nlohmann_json/[>=3.9 <4]")
```
memsharded, line 26: I have checked the occurences for this in ConanCenter.
There are like 38 of them, the vast majority to ``3.11.3``. This is also the latest version on ConanCenter.
So the version range wouldn't solve much, using versions like 3.9 is a bit untested, and allowing newer versions doesn't change much if the latest is 3.11.3.

We also did some research, to propose version ranges for ``nlohmann_json``, but we didn't find any evidence (not favor not against) the binary compatibility of the library, so we didn't decide to move forward with it yet.

If the previous ``3.11.2`` was raising conflicts, maybe updating it to ``3.11.3`` would be enough to solve the conflicts?

recipes/pybind11_json/all/conanfile.py
```diff
@@ -23,8 +23,8 @@ def layout(self):
...
+        self.requires("pybind11/[>=2.6 <3]")
```
memsharded, line 27: This library:
- is only used by other 3 libs in conancenter, depending on ``pybind11/2.10.1``. So this probably wouldn't create new conflicts
- IT is true that there are newer ``pybind11/2.13.6``, and it would make sense to let ``pybind11_json`` to move to them.
- I think it is better to keep the lower limit here
```suggestion
        self.requires("pybind11/[>=2.10.4 <3]")
```

memsharded (commented):

As this change is more related to general alignment with other existing recipes in ConanCenter, I'll try to discuss this with the team, to see what makes sense in terms of version ranges and possible conflicts.

recipes/pybind11_json/all/conanfile.py
mologie: My reasoning for choosing a range here was that `nlohmann_json` is a header-only library, as is pybind11 and this recipe, so there is no ABI to worry about to begin with.

Conan does not appear to have a system in place yet to ensure that recipe dependencies do not become stale, so I'd prefer ranges whenever possible, otherwise recipes like this one become unavailable when the target application requires a more recent dependency of this recipe as well. For pybind11 and nlohmann_json, that is bound to happen very frequently.

recipes/pybind11_json/all/conanfile.py
mologie: Your comment mentions that you think it's better to keep the lower limit, yet the code change request raises the limit, could you clarify please? Either way is fine with me, though I selected this range simply because old pybind11 versions are available and compatible. I'd let it be the user's choice what they want to combine in the end.

PR #23788, opened 2024-04-27
Update CMakeFile for ohPipeline

DoomHammer:
CMAKE_BUILD_TYPE is apparently not set in Windows builds. The new approach should handle this.

---


DoomHammer:
BTW: if there is a better way to handle the ServiceGen path handover than this: https://github.com/conan-io/conan-center-index/blob/master/recipes/ohnet/all/conanfile.py#L113 I'd be happy to learn. Somehow this feels extremely hacky.

recipes/ohpipeline/all/conandata.yml
```diff
@@ -4,8 +4,8 @@ sources:
       url: "https://github.com/openhome/ohPipeline/archive/refs/tags/ohMediaPlayer_1.139.1000.tar.gz"
       sha256: "070eac6d974bbe520d0b292df1ca4a72160a2b607a60dabdaa0971c3ef96e5e4"
     cmake:
-      url: "https://raw.githubusercontent.com/merakiacoustic/ohPipeline/ohMediaPlayer_1.139.1000_meraki2/CMakeLists.txt"
-      sha256: "f9e11d4adb8426a8375dae02e3b641d66efc6ef3abd6f96cb81f6e85352083a3"
+      url: "https://raw.githubusercontent.com/merakiacoustic/ohPipeline/ohMediaPlayer_1.139.1000_meraki3/CMakeLists.txt"
```
Croydon, line 7: Needing to maintain a completely custom build system for a project is already a nightmare, but even more so when externalizing the scripts. I am not sure why this was accepted.

If it is too big to be included in CCI's repository, it is too big to be maintained. Can we please integrate it back here or have a general discussion about this?

DoomHammer: I've also created a separate PR that tries to work with existing build system but it fails in ways I can't reproduce :(

memsharded (approved):

recipes/ohpipeline/all/conandata.yml
memsharded: Agree this should be done, but I know it can be complicated too. 
So approving this fix in the meantime, thanks very much for contributing it @DoomHammer !

perseoGI (approved):

PR #23790, opened 2024-04-27
openni2: new recipe

valgur:
Adds OpenNI2: https://github.com/structureio/OpenNI2

The OpenNI 2.0 API provides access to PrimeSense-compatible depth sensors.

[Packaging status](https://repology.org/project/openni2/versions)

An optional dependency for the following packages on CCI:
- libfreenect
- libfreenect2
- librealsense
- opencv
- pangolin
- pcl


valgur:
That's a fair point, but the old repo is pretty much dead and the fork is used as the main repo by all other package managers, including Debian/Ubuntu: https://repology.org/project/openni2/information

czoido:
> That's a fair point, but the old repo is pretty much dead and the fork is used as the main repo by all other package managers, including Debian/Ubuntu: https://repology.org/project/openni2/information

Hi @valgur, thanks for the info, I was digging a bit and looks like the own owners of the original repo recommend using the fork, so I think it's ok for the recipe to be named openni2 but please let's keep a comment in the recipe stating that this is the recommended fork from the original repo.

PR #23791, opened 2024-04-27
Increase version-range of libwebsockets->openssl to include openssl 3

memsharded:
Specify library name and version:  **libwebsockets**

So far libwebsockets couldn't link with openssl3, so the range couldn't be defined as in other recipes.

https://github.com/warmcat/libwebsockets/issues/2894 reported the issue as a question, and it seems the problem is not defining that the library links with ``zlib`` too, which is brought as a transitive dependency of ``openssl``, but the consumer ``libwebsockets`` has its own definition of a variable when ``zlib`` is to be considered by the library.

NOTE: This changes the default ``with_zlib="zlib"`` option, and this can have other consequences.

### DIFF
--- a/recipes/libwebsockets/all/conanfile.py
+++ b/recipes/libwebsockets/all/conanfile.py
@@ -249,7 +249,7 @@ class LibwebsocketsConan(ConanFile):
         if self.options.with_ssl == "openssl":
             no_zlib = self.dependencies["openssl"].options.get_safe("no_zlib")
             if not no_zlib and self.options.with_zlib != "zlib":
-                raise ConanInvalidConfiguration(f"{self.name} needs option with_zlib=zlib, as its dependency openssl links with it")
+                raise ConanInvalidConfiguration(f'{self.name} needs option "{self.name}/*:with_zlib=zlib", as its dependency openssl links with it')
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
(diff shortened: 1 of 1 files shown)


recipes/libwebsockets/all/conanfile.py
```diff
@@ -246,6 +246,11 @@ def validate(self):
             raise ConanInvalidConfiguration("Library hubbub not implemented (yet) in CCI")
             # TODO - Add hubbub package when available.
 
+        if self.options.with_ssl == "openssl":
+            no_zlib = self.dependencies["openssl"].options.get_safe("no_zlib")
+            if not no_zlib and self.options.with_zlib != "zlib":
+                raise ConanInvalidConfiguration(f"{self.name} needs option with_zlib=zlib, as its dependency openssl links with it")
```
RubenRBS, line 252:
```suggestion
                raise ConanInvalidConfiguration(f'{self.name} needs option "{self.name}/*:with_zlib=zlib", as its dependency openssl links with it')
```

RubenRBS, line 252:
```suggestion
                raise ConanInvalidConfiguration(f'{self.name} needs option "{self.ref}:with_zlib=zlib", as its dependency openssl links with it')
```

RubenRBS, line 252: To discuss with the team what variant we want, either it showing
`libwebsockets/*:with_zlib=zlib` or `libwebsockets/x.y.z:with_zlib`, and what is more UX friendly

PR #23792, opened 2024-04-27
fix abuse of Conan attributes in fftw

memsharded:
Specify library name and version:  **fftw**

This recipe is abusing the ``build_folder``, overwriting it. According to the docs: https://docs.conan.io/2/reference/conanfile.html

> Public attributes and methods, like build(), self.package_folder, are reserved for Conan. Don’t use public members for custom fields or methods in the recipes.

The changes proposed seems to both simplify the recipe and avoid this issue. 

This was originated in https://github.com/conan-io/conan/issues/16166, as recent changes in ``develop2`` makes the ``-DCMAKE_TOOLCHAIN_PATH`` defined in Presets at generation time relative between build-folder and generators-folder, but if ``build_folder`` is changed under the hood, things do not longer match.

Close https://github.com/conan-io/conan/issues/16166

### DIFF
--- a/recipes/fftw/all/conanfile.py
+++ b/recipes/fftw/all/conanfile.py
@@ -49,7 +49,6 @@ class FFTWConan(ConanFile):
         "combinedthreads": False,
         "simd": False,
     }
-    _current_precision = None
 
     def export_sources(self):
         export_conandata_patches(self)
@@ -94,11 +93,6 @@ class FFTWConan(ConanFile):
         tc.generate()
 
     @property
-    def build_folder(self):
-        bf = super().build_folder
-        return os.path.join(bf, self._current_precision) if self._current_precision else bf
-
-    @property
     def _all_precisions(self):
         return [p for p in ALL if self.options.get_safe(f"precision_{p}")]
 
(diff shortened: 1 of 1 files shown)


jcar87 (approved):
LGTM

The resulting contents are the same before and after:

```
├── conaninfo.txt
├── conanmanifest.txt
├── include
│   ├── fftw3.f
│   ├── fftw3.f03
│   ├── fftw3.h
│   ├── fftw3l.f03
│   └── fftw3q.f03
├── lib
│   ├── libfftw3.a
│   ├── libfftw3f.a
│   └── libfftw3l.a
└── licenses
    └── COPYRIGHT
```

recipes/fftw/all/conanfile.py
```diff
@@ -107,29 +101,21 @@ def on_off(value):
...
+                "ENABLE_FLOAT": on_off(current_precision == SINGLE),
+                "ENABLE_LONG_DOUBLE": on_off(current_precision == LONGDOUBLE),
+                "ENABLE_QUAD_PRECISION": on_off(current_precision == QUAD)
             }
             cmake.configure(variables=variables)
             cmake.build()
-
-        # Potentially avoid side effects due to build_folder property tweak.
-        self._current_precision = None
+            cmake.install()
```
jcar87, line 113: I think the proposed approach makes sense: files are added to the install folder (the libraries), and every other file is just overwritten with exactly the same contents. This is equivalent to the way it's done in CMake to support multi-configuration (Release/Debug) in the same install tree: configure twice, build twice, install twice to same location.

Was wondering if it would be worth passing `--fresh` to CMake configure, or clearing the build folder ahead of starting, but from what I can see, we wouldn't benefit much, as the result is the same.

RubenRBS (approved):

PR #23798, opened 2024-04-28
lapack: new recipe

### LINKED ISSUE #4509: [request] lapack/3.7.1
It is archived in [conan-community](https://github.com/conan-community/conan-lapack)
I think we should move it to conan-center to continue recipe polish :)

valgur:
Adds LAPACK and LAPACKE from the reference implementation at https://github.com/Reference-LAPACK/lapack/releases

LAPACK is a library of Fortran subroutines for solving the most commonly occurring problems in numerical linear algebra.

Together wih BLAS, it is required for nearly all packages dealing with linear algebra that don't use Eigen. tT's currently available as a part of the OpenBLAS package on CCI, but it is not quite sufficient in practice. CMake projects that need LAPACK usually rely on [`FindLAPACK`](https://cmake.org/cmake/help/latest/module/FindLAPACK.html) from CMake to find the package, which OpenBLAS and others cannot provide with the current CMakeDeps functionality.

The other motivation for decoupling LAPACK from OpenBLAS is that there are several alternative BLAS implementations available and LAPACK, which builds upon BLAS, can generally be handled separately and use the reference implementation.

[Packaging status](https://repology.org/project/lapack/versions)

- Requires #23334

- Resolves #4509

### DIFF
--- a/recipes/lapack/all/conandata.yml
+++ b/recipes/lapack/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "3.12.0":
+    url: "https://github.com/Reference-LAPACK/lapack/archive/refs/tags/v3.12.0.tar.gz"
+    sha256: "eac9570f8e0ad6f30ce4b963f4f033f0f643e7c3912fc9ee6cd99120675ad48b"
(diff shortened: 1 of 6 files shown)


valgur:
Needs Fortran support.

PR #23801, opened 2024-04-28
cr: new recipe

valgur:
Adds cr: https://github.com/fungos/cr

cr.h: A Simple C Hot Reload Header-only Library

[Packaging status](https://repology.org/project/cr/versions)


RubenRBS (approved):
Thanks!

PR #23805, opened 2024-04-29
libgettext: fix configure so it doesnt block on MSVC debug builds

paulharris:
During ./configure, the "intl" configure step tests this condition:
checking whether printf supports the 'n' directive...

This fails on MSVC, which is fine on Release builds.

However, in Debug builds, a GUI Dialog box pops up telling the user about an assert failure, and would the user like to Abort,Retry,Ignore.

This blocks the configuration until the user manually presses a button.

In gettext-runtime/intl/configure (in the tarball release),
line 29702 has a call to _set_invalid_parameter_handler()
This almost does the job, but not quite.
adding the line: _CrtSetReportMode(_CRT_ASSERT, 0);
changes the behaviour to fail without the dialog box.

There are other places in the tarball where _set_invalid_parameter_handler() is called, and this line could be added in all of those places to be sure, but my tests show it is enough to add it to just this one place,

### DIFF
--- a/recipes/libgettext/all/conandata.yml
+++ b/recipes/libgettext/all/conandata.yml
@@ -9,6 +9,10 @@ sources:
     sha256: "66415634c6e8c3fa8b71362879ec7575e27da43da562c798a8a2f223e6e47f5c"
     url: "https://ftp.gnu.org/pub/gnu/gettext/gettext-0.20.1.tar.gz"
 patches:
+  "0.22":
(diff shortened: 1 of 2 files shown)


jcar87:
Thanks @paulharris for opening this issue and opening a detailed issue. 

We have fixed this by passing an argument to the configure script to preset the result of this check, as we already use that approach in the gettext recipe for and it makes ongoing maintenance easier.

Thank you for reporting this upstream, your proposed fix looks sensible to be, but the gettext maintainers are better placed to advise.

PR #23831, opened 2024-04-30
libwebsockets: use find_package(OpenSSL) and use version range for OpenSSL

jcar87:
Fix issue where `libwebsockets` could not be built with OpenSSL 3 because the build scripts were not correctly forwarding the transitive dependencies of OpenSSL to CMake's TryCompile

In this PR:
* Stop defining `LWS_OPENSSL_LIBRARIES` and `LWS_OPENSSL_INCLUDE_DIRS`, and ensure that `find_package(OpenSSL)` is called early instead, defining the required `OPENSSL_*` variables when needed.
* By locating OpenSSL early on, this prevents fallback logic that try to locate it with PkgConfig in addition to `find_package` (this was redundant)
* Define `CMAKE_TRY_COMPILE_CONFIGURATION` to ensure that it matches the current build type. This causes CMake to correctly propagate the Conan-generated targets.
* Use an OpenSSL version range that includes versions `>=3`

Fixes:
* The build configuration step correctly locates the required OpenSSL functionality: `-- Looking for HMAC_CTX_new - found` 
* Include changes from https://github.com/conan-io/conan-center-index/pull/19157, preserving authorship. 

Close https://github.com/conan-io/conan-center-index/pull/23791
Close https://github.com/conan-io/conan-center-index/pull/19157


memsharded (approved):

uilianries (approved):

oavitzur:
@jcar87 Hi, this merge causes all our builds that use libwebsockets in our environment to fail. 

we get this error:

`
project could not find requested file:
   /home/devbox/.conan/data/libwebsockets/4.3.3/_/_/build/9418abf133f9bf2de819612f91599e720fe872d8/src/project_include.cmake
```
`

And indeed, the libwebsocket repo does not have this `project_include.cmake` file

 What am I missing ? 
Thanks

oavitzur:
@jcar87  I think I found the  issue. I am building libwebsockets withnout ssl : `"libwebsockets:with_ssl": False,` 
This cause this failure. When I remove this option then the build succeeds

PR #23838, opened 2024-05-01
eudev: Bump libxslt to avoid conflicts

jwillikers:
Fixes conflicts with libxml2 in the graph caused by older version of libxslt.


RubenRBS:
Rebuilding the missing binary :)

PR #23494, opened 2024-04-12
libjpeg - Use MSBuild for windows platforms #21628

datalogics-rgailiunas:
- Allow libjpeg to be built using MSBuild on windows platforms adding support for Windows ARM64.
 - Fixes issue [#21628](https://github.com/conan-io/conan-center-index/issues/21628)

Specify library name and version:  **libjpeg/9e**

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

Issue #23453, opened 2024-04-09
libjpeg/9e: Package fails to build correctly for Windows x86


dcoburnMEDITECH:
### Description

When trying to use the libjpeg/9e package in a x86 build environment the code will crash when using the libjpeg functions.  We found that there was an argument with size_t type.  This cause the wrong width of the size_t and breaking the call to cause our code to crash. 

The error is that the compiler was called with the wrong arguments:

cl -c -DLIBJPEG_BUILDING -DLIBJPEG_STATIC -DCRTAPI1=_cdecl -DCRTAPI2=_cdecl -nologo -GS -D_AMD64_=1 -DWIN64 -D_WIN64  -DWIN32 -D_WIN32 -W4 -D_WINNT -D_WIN32_WINNT=0x0500 -DNTDDI_VERSION=0x05000000 -D_WIN32_IE=0x0500 -DWINVER=0x0500 -Zi -Od -DDEBUG -Fdlibjpeg.pdb  -D_MT -MTd -I. jquant2.c

The above line is not for 32 bit builds, but for 64 bit builds.

[Config:x86d] cl -c -DLIBJPEG_BUILDING -DLIBJPEG_STATIC -DCRTAPI1=_cdecl -DCRTAPI2=_cdecl -nologo -GS  -DWIN32 -D_WIN32 -W4 -D_WINNT -D_WIN32_WINNT=0x0500 -DNTDDI_VERSION=0x05000000 -D_WIN32_IE=0x0500 -DWINVER=0x0500 -Zi -Od -DDEBUG -Fdlibjpeg.pdb  -D_MT -MTd -I. jquant2.c

The line above is the correct line for 32 bit builds.

This was traced to win32.mak section of code:

!IF "$(CPU)" == "i386"
cflags = $(ccommon) -D_X86_=1  -DWIN32 -D_WIN32 -W3
scall  = -Gz

!ELSEIF "$(CPU)" == "IA64"
cflags = $(ccommon) -D_IA64_=1 -DWIN64 -D_WIN64  -DWIN32 -D_WIN32
cflags = $(cflags) -W4
scall  =

!ELSEIF "$(CPU)" == "AMD64"
cflags = $(ccommon) -D_AMD64_=1 -DWIN64 -D_WIN64  -DWIN32 -D_WIN32
cflags = $(cflags) -W4
scall  =

!ENDIF

if one goes back to the top of win32.mak
We find:
[...]

Issue #21628, opened 2023-12-06
[libjpeg] libjpeg/9e: won't build on Windows ARM but can build manually


datalogics-robb:
### Description

The jpeg library recipe needs to add support for Windows ARM64 - instructions are present in the 9e sources as follows:

nmake /f makefile.vs setup-v16
"Open the solution file jpeg.sln, build the library project."
"If you want to build ARM/ARM64 code, change the platform setting
in Visual Studio Installer (Workloads tab Optional components or Individual components tab)."
"apps.sln to build the application projects."

### Package and Environment Details

* Package Name/Version: **libjpeg/9e**
* Operating System+version: **Microsoft Windows 10/11**
* Compiler+version: **Visual Studio 2019/2022**
* Conan version: **conan 1.62.0** (any)
* Python version: **Python 3.11** 

### Conan profile

[settings]
os=Windows
os_build=Windows
arch=armv8
arch_build=armv8
compiler=Visual Studio
compiler.version=17
compiler.runtime=MD
build_type=Release
[options]
[conf]
tools.microsoft.bash:sybsystem=msys2
tools.microsoft.bash:path="C:\msys64\clangarm64.exe"
[build_requires]
[env]

### Steps to reproduce

Attempt to build a libjpeg/9e package:
`conan create -pr:b visual-studio-17-arm64 -pr:h visual-studio-17-arm64 --build=missing . libjpeg/9e@_/_`

### Logs

<details><summary>Conan create output</summary>

```
(python-env-desktop-e4e6vqq) c:\Users\rboeh\Development\conan-center-index\recipes\libjpeg\all>conan create -pr:b visual
-studio-17-arm64 -pr:h visual-studio-17-arm64 --build=missing . libjpeg/9e@_/_
Conan 1 is on a deprecation path, please consider migrating to Conan 2
Exporting package recipe
libjpeg/9e exports: File 'conandata.yml' found. Exporting it...
libjpeg/9e exports: Copied 1 '.yml' file: conandata.yml
libjpeg/9e: Calling export_sources()
libjpeg/9e: Copied 1 '.Mak' file: Win32.Mak
libjpeg/9e: A new conanfile.py version was exported
libjpeg/9e: Folder: C:\Users\rboeh\.conan\data\libjpeg\9e\_\_\export
libjpeg/9e: Using the exported files summary hash as the recipe revision: 8e24c99af83e8b5c3002b8f1e7cbb4f2
[...]

Issue #10116, opened 2022-04-05
[package] libjpeg/9c: compiler.toolset is ignored for VS2022


Nekto89:
### Package and Environment Details (include every applicable attribute)
  * Package Name/Version: **libjpeg/9c**
  * Operating System+version: **Windows 10**
  * Compiler+version: **Visual Studio 17 2022 (17.1.2)**
  * Docker image: **-**
  * Conan version: **conan 1.47.0**
  * Python version: **Python 3.10.3**

### Conan profile (output of `conan profile show default` or `conan profile show <profile>` if custom profile is in use)
```
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Debug
compiler=Visual Studio
compiler.runtime=MDd
compiler.toolset=v142
compiler.version=17
os=Windows
os_build=Windows
[options]
[build_requires]
[env]
```

### Steps to reproduce (Include if Applicable)
conan install -s build_type=Debug libjpeg/9c@ --build=libjpeg

I was expecting that binaries will be built with v142 (14.29.*) but from log and resulting static library I see that build was using native toolset for VS2022 (14.31).

### Logs (Include/Attach if Applicable)
<details><summary>Click to expand log</summary>

```
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Debug
compiler=Visual Studio
compiler.runtime=MDd
compiler.toolset=v142
compiler.version=17
os=Windows
os_build=Windows
[options]
[build_requires]
[env]

libjpeg/9c: Forced build from source
Installing package: libjpeg/9c
Requirements
    libjpeg/9c from 'conancenter' - Cache
Packages
    libjpeg/9c:d057732059ea44a47760900cb5e4855d2bea8714 - Build

Installing (downloading, building) binaries...
libjpeg/9c: Copying sources to build folder
libjpeg/9c: Building your package in C:\Users\mklymov\.conan\data\libjpeg\9c\_\_\build\d057732059ea44a47760900cb5e4855d2bea8714
libjpeg/9c: Generator txt created conanbuildinfo.txt
libjpeg/9c: Aggregating env generators
libjpeg/9c: Calling build()
**********************************************************************
** Visual Studio 2022 Developer Command Prompt v17.1.2
** Copyright (c) 2022 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'

Microsoft (R) Program Maintenance Utility Version 14.31.31105.0
[...]

PR #23840, opened 2024-05-01
libjpeg: use msbuild

jcar87:
Use the libjpeg provided `.sln`/`.vcxproj` solution to build libjpeg

* Fix ability to build for `ARM64` and `Win32` (32-bit intel)
* Drop old `9c` version and add new `9f` version
* Use https for all source tarballs

Fix: https://github.com/conan-io/conan-center-index/issues/23453
Fix: https://github.com/conan-io/conan-center-index/issues/21628
Close: https://github.com/conan-io/conan-center-index/pull/23494

Should fix: https://github.com/conan-io/conan-center-index/issues/10116, but needs to be tested


recipes/libjpeg/all/conanfile.py
```diff
@@ -82,46 +82,64 @@ def generate(self):
...
+                # Inject conan-generated .props file
+                conantoolchain_props = os.path.join(self.generators_folder, MSBuildToolchain.filename)
+                replace_in_file(
+                    self, jpeg_vcxproj,
+                    """<Import Project="$(VCTargetsPath)\\Microsoft.Cpp.targets" />""",
+                    f"""<Import Project="{conantoolchain_props}" /><Import Project="$(VCTargetsPath)\\Microsoft.Cpp.targets" />""",
+                )
+
+                # Fix toolset 
+                toolset_before = "v142" if self.version in ["9d", "9e"] else "$(DefaultPlatformToolset)"
```
memsharded, line 113: In theory it is possible to override the ``vcxproj`` defined toolset with the ``conantoolchain.props`` one, as long as the inclusion is done in the right place. But not a problem with this either, should be good too.

memsharded (approved):

recipes/libjpeg/all/conanfile.py
jcar87: Good catch!
the right place is to import it right before `Microsoft.Cpp.props` - then the toolset version set by .conan props is respected, making this redundant :D - with the added benefit that the Visual Studio IDE then shows the Conan .props file correctly as well

memsharded (dismissed):

uilianries (dismissed):

jcar87:
Screenshot 2024-05-03 at 09 15 35

Need to update one last thing. the previous revision of the package calls the DLL `libjpeg-9.dll`, while the new one so far would do `libjpeg.dll` - breaking compatibility. Will fix this before merging

uilianries (approved):

RubenRBS (approved):

Issue #23854, opened 2024-05-03
[package] gtest/1.14.0: GMock C++14 build not compatible with C++17


Twon:
### Description

The GMock component of the GTest package checks the C++ language version and if C++17 make use of C++17 features such as `std::string_view`.  However, when Conan pulls down a pre-built package it detects the C++14 build as compatible with C++17 which is incorrect.  This has been identified as an issue on the GTest repository: https://github.com/google/googletest/issues/4357

Note this issue is similar in nature to https://github.com/conan-io/conan-center-index/issues/19008, and the same fix should be applied here.

### Package and Environment Details

* Package Name/Version: **gtest/1.14.0**
* Operating System+version: **Microsoft Windows Server 2022**
* Compiler+version: **msvc 11**
* Conan version: **conan 2.2.3**
* Python version: **Python 3.12.3**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

I have built a simple Github repository which reproduces this issue within its Github Action workflow, which can be found here: https://github.com/Twon/ConanGTestIssueWithCxx17/pull/1

### Logs

<details><summary>Click to expand log</summary>

```
======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows


======== Computing dependency graph ========
gtest/1.14.0: Not found in local cache, looking in remotes...
gtest/1.14.0: Checking remote: conancenter
gtest/1.14.0: Downloaded recipe revision 0bfd0373[13](https://github.com/Twon/ConanGTestIssueWithCxx17/actions/runs/8929927784/job/24528859496?pr=1#step:11:14)8714a7567dcdabd133a722
Graph root
    conanfile.txt: D:\a\ConanGTestIssueWithCxx17\ConanGTestIssueWithCxx17\conanfile.txt
Requirements
[...]

PR #23856, opened 2024-05-03
GTest - Flag packages built with different cxx langauge support as incompatible

Twon:
Specify library name and version:  **gtest/1.14.0**

GTest packages built with a C++14 compiler result in linkage errors when used in a C++17 program and attempting to use C++17 features such as `std::sting_view`.  I have built a reproduction of this issue here: https://github.com/Twon/ConanGTestIssueWithCxx17/pull/1

Note this mirrors the fix applied to Catch2 to ensure packages were compatible with C++17 caused the same underlying cause.  This change applies the same solution to GTest: https://github.com/conan-io/conan-center-index/pull/21792

Fixes: #23854

---


CJCombrink (approved):

RubenRBS (approved):
Thanks for taking the time to implement the fix, we appreciate it :)

uilianries (approved):

uilianries:
@hhowe29 I just talked to @RubenRBS now and we will revert the minimum Conan client required. 

As result, older Conan version will not break, just will not use the feature, because is an attribute.

uilianries:
@hhowe29 I just opened the PR https://github.com/conan-io/conan-center-index/pull/23909

hhowe29:
> It's specially recommended for companies and users that are using ConanCenterIndex in production to keep a lockfile or your own server, to avoid any unexpected behavior from CCI recipes/packages.

Thanks for the advice. Obviously we are not using a lock file. Will look into getting one added.

jcar87:
> @Twon, [This change](https://github.com/conan-io/conan-center-index/pull/23856/files#diff-30c143c486109e8ef87666fdfc05fe43889e15121afa1addd541aa409c91ee95R10) broke our build because we are still on conan 2.0.x.
> 
> ```
> ERROR: Package 'gtest/1.14.0' not resolved: gtest/1.14.0: Cannot load recipe.
> Error loading conanfile at '/root/.conan2/p/gtestc561d3ff51c0d/e/conanfile.py':
> Current Conan version (2.0.13) does not satisfy the defined one (>=1.53.0 <2 || >=2.1.0)..
> ```
> 
> Is forcing 2.1.x necessary? Seems odd that 1.64 is allowed, but a version 2.0 release from a few months ago is forcibly rejected. I get that 2.1 is the first version the implements `extension_properties`, but it seems like 2.0 users who had green CI builds would prefer to keep those builds green instead of being compelled to upgrade conan without warning.

While it happens to be the case that we could keep the recipe compatible with older versions as unknown attributes are ignored - it has always been the longstanding practice in Conan Center recipes to assume that the client can be updated to the latest. Significant efforts go into making sure Conan is backwards compatible and we don't introduce behaviour changes. We are relaxing the version constraint for this recipe, however - if the mindset is "users who had green CI builds would prefer to keep them green", it does make me wonder why users would pull the latest recipe revision in the first place - lockfiles are recommended here, in the same way that they are for pip (python) and cargo (Rust), for exactly the same reasons.

PR #23864, opened 2024-05-03
restinio: bump 3p to `boost/1.85.0`

gegles:
Specify library name and version:  **restinio/0.7.2**

---


gegles:
@memsharded @uilianries @RubenRBS any chance this could be picked up? It's been about 5,6 weeks... Thx!

recipes/restinio/v0.7/conanfile.py
```diff
@@ -37,12 +37,12 @@ def layout(self):
     def requirements(self):
         self.requires("llhttp/9.1.3")
         self.requires("fmt/10.2.1")
-        self.requires("expected-lite/0.6.3")
+        self.requires("expected-lite/0.8.0")
```
memsharded, line 40: This will create a conflict with jinja2cpp, still using previous version. Bump deps should be better orchestrated with others or wait until/if they can be moved to version-ranges

recipes/restinio/v0.7/conanfile.py
```diff
@@ -37,12 +37,12 @@ def layout(self):
...
 
         if self.options.asio == "standalone":
-            self.requires("asio/1.29.0")
+            self.requires("asio/1.30.2")
```
memsharded, line 43: This will create conflicts with around 14 other libraries depending on previous versions of ``asio``

recipes/restinio/v0.7/conanfile.py
```diff
@@ -37,12 +37,12 @@ def layout(self):
...
         else:
-            self.requires("boost/1.84.0")
+            self.requires("boost/1.85.0")
```
memsharded, line 45: There are like other 110 recipes depending on boost, practically none to ``1.85``, so this will conflict with all of them.

memsharded (commented):
The problem with these version bumps might create more version conflicts that they solve. This is the reason why version bumps are not being merged as quickly as before, they are being prepared and orchestrated (mostly by @RubenRBS), so they are done simultaneously in order to reduce conflicts to users.

gegles:
> The problem with these version bumps might create more version conflicts that they solve. This is the reason why version bumps are not being merged as quickly as before, they are being prepared and orchestrated (mostly by @RubenRBS), so they are done simultaneously in order to reduce conflicts to users.

Thanks for the clarification @memsharded. I get it, maybe previously the per-recipe deps bump would happen too aggressively, but the opposite, i.e. the status quo is not great either, and I am afraid things will get very stale over time...

I hope a clear process will emerge so that deps can be upgraded frequently and globally... 

Would a cross-recipe PR upgrading a single dep at a time throughout the recipes be a better way? Or would it be too costly from a CI builds POV?

memsharded:
> I hope a clear process will emerge so that deps can be upgraded frequently and globally...

Yes, the process is now already started, and moving to version ranges will be more and more spread. It just takes time to do it with the current infrastructure/CI without breaking too much users.

> Would a cross-recipe PR upgrading a single dep at a time throughout the recipes be a better way? Or would it be too costly from a CI builds POV?

This is exactly what we are trying to do with the current work for a new CI pipeline. This new pipeline will support much better this kind of actions. 

At this moment it is not possible and @RubenRBS has been orchestrating these changes with multiple parallel PRs, but we are getting closer to the new pipeline, so it might be better to wait until it is running.

recipes/restinio/all/conanfile.py
```diff
@@ -38,7 +38,7 @@ def layout(self):
 
     def requirements(self):
         self.requires("http_parser/2.9.4")
-        self.requires("fmt/10.1.1")
+        self.requires("fmt/11.0.1")
```
memsharded, line 41:
```suggestion
        self.requires("fmt/10.2.1")
```

This is the most popular version by far in ConanCenter, I'd use this one to avoid conflicts

recipes/restinio/v0.7/conanfile.py
memsharded: This is still conflicting with jinja2cpp

memsharded (changes requested):
I am afraid the commented conflicts are still there, this bumping of versions will create more issues than problems it might solve.

recipes/restinio/v0.7/conanfile.py
gegles: @memsharded, as I am now also adding `0.7.3` and that version requires these, I am adding a conditional check on the version..

Hope this will allow us merging this. thx!

gegles:
Also now added version [0.7.4](https://github.com/Stiffstream/restinio/releases/tag/v.0.7.4). Thanks @eao197!

danimtb (approved):

memsharded (dismissed):
I think this PR can be merged, I have recommended to remove the intermediate patches versions, unless there is evidence they are specifically requested by users.

recipes/restinio/v0.7/conanfile.py
```diff
@@ -36,13 +36,24 @@ def layout(self):
...
         if self.options.asio == "standalone":
-            self.requires("asio/1.29.0")
+            if Version(self.version) >= "0.7.4":
+                self.requires("asio/1.32.0")
+            else:
+                self.requires("asio/1.31.0")
         else:
-            self.requires("boost/1.84.0")
+            if Version(self.version) >= "0.7.4":
+                self.requires("boost/[>=1.87.0]")
```
uilianries, line 54:
```suggestion
                self.requires("boost/1.87.0")
```

I don't recommend using version range for Boost because there is no guarantee about full compatibility between minors. Boost libraries can change their ABI compatibility (e.g. require a newer C++ standard) and the API (e.g. Boost process v2 become the default over v1)

uilianries: Doing a second look, RestinIO requires C++17 already, so it should not break in case Boost changes the required C++ for one of used modules here.

I did a quick search in RestinIO 0.7.7 and only Boost ASIO and Regex are used:

```
find . -type f -name "*.hpp" -exec grep boost/ {} +                                                                                                                                                                
./dev/restinio/impl/tls_socket.hpp:  #include <boost/asio/ssl.hpp>
./dev/restinio/asio_include.hpp:#include <boost/asio.hpp>
./dev/restinio/router/boost_regex_engine.hpp:#include <boost/regex.hpp>
```

Also, checking Boost changelog (https://www.boost.org/releases/1.88.0/ and older), I did not find any breaking change, most fixes.

So, I'm fine accepting version ranges for Boost here. Plus, I would recommend using always instead, replacing the if..else by:

```python
self.requires("boost/[>=1.86.0]")
```

It should work, and will make easier to maintain the recipe later.

memsharded (dismissed):

uilianries (approved):
LGTM.

Just tested locally using Boost as ASIO implementation:

- [restinio-0.7.2-mac-boost.log](https://github.com/user-attachments/files/20775097/restinio-0.7.2-mac-boost.log)
- [restinio-0.7.7-mac-boost.log](https://github.com/user-attachments/files/20775098/restinio-0.7.7-mac-boost.log)

It requires Boost 1.86.0 or early for RestinIO <= 0.7.4. Otherwise, it fails with:

```
In file included from /Users/uilian/.conan2/p/b/resti42603d4f50b58/p/include/restinio/impl/acceptor.hpp:17:
/Users/uilian/.conan2/p/b/resti42603d4f50b58/p/include/restinio/impl/connection.hpp:1134:18: error: no template named 'buffer_cast' in namespace 'restinio::asio_ns'; did you mean 'buffer_copy'?
 1134 |                                                                 asio_ns::buffer_cast< const char * >(
      |                                                                 ~~~~~~~~~^
/Users/uilian/.conan2/p/b/boostaecf3d227aa67/p/include/boost/asio/buffer.hpp:2469:20: note: 'buffer_copy' declared here
 2469 | inline std::size_t buffer_copy(const MutableBufferSequence& target,
      |                    ^
In file included from /Users/uilian/Development/conan/conan-center-index/recipes/restinio/v0.7/test_package/test_package.cpp:2:
In file included from /Users/uilian/.conan2/p/b/resti42603d4f50b58/p/include/restinio/core.hpp:20:
In file included from /Users/uilian/.conan2/p/b/resti42603d4f50b58/p/include/restinio/http_server.hpp:14:
/Users/uilian/.conan2/p/b/resti42603d4f50b58/p/include/restinio/impl/acceptor.hpp:266:46: error: no member named 'max_connections' in 'boost::asio::socket_base'
  266 |                                 m_acceptor.listen( asio_ns::socket_base::max_connections );
      |                                                    ~~~~~~~~~~~~~~~~~~~~~~^
/Users/uilian/.conan2/p/b/resti42603d4f50b58/p/include/restinio/impl/acceptor.hpp:623:36: error: no member named 'from_string' in 'boost::asio::ip::address'
  623 |                                 result = asio_ns::ip::address::from_string( str_addr );
      |                                          ~~~~~~~~~~~~~~~~~~~~~~^
3 errors generated.
```

danimtb (approved):

PR #23873, opened 2024-05-05
Qt5: add patch for 5.15.13 to fix QDir::entryInfoList() with nameFilters

kambala-decapitator:
Specify library name and version:  **qt/5.15.13**

Backport of patch that's available only for 5.15.14

Upstream bug: https://bugreports.qt.io/browse/QTBUG-112738

---


ericLemanissier (commented):
Thanks @kambala-decapitator, but unfortunately the situation is very similar to https://github.com/conan-io/conan-center-index/pull/23659. I think we should just wait for 5.15.14 to be released under lgpl (before the end of may 2024).
Also, this patch is fixing a bug on a feature which is not central to qt (in the sense that it's not a bug preventing compilation, or preventing all uses of qt on a given platform), so I think it's out of scope for conan-center

PR #23880, opened 2024-05-06
ncurses: add v6.5, drop v6.2

valgur:
https://invisible-island.net/ncurses/announce.html#h2-release-notes

Dropped v6.2 solely to simplify the recipe a bit, since it includes a crazy amount of patches. v6.2 was released 4 years ago.


RubenRBS (approved):
What supprises me is that the newer versions do not have patches addressing some of the points, thanks a lot!

ericLemanissier (approved):

PR #23894, opened 2024-05-06
Compile as C

datalogics-staylor:
Specify library name and version:  **qr-code-generator/1.8.0**

* Compile as C when the package is built with the new "compile_as_c" option.

---

### DIFF
--- a/recipes/qr-code-generator/all/conanfile.py
+++ b/recipes/qr-code-generator/all/conanfile.py
@@ -27,10 +27,12 @@ class QrCodeGeneratorConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "compile_as_c": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "compile_as_c": False,
     }
 
     @property
@@ -54,6 +56,10 @@ class QrCodeGeneratorConan(ConanFile):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
+        if not self.options.compile_as_c:
+            self.options.rm_safe("compiler.libcxx")
+            self.options.rm_safe("compiler.cppstd")
+
     def layout(self):
         cmake_layout(self, src_folder="src")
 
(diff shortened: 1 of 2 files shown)


jcar87:
Hi @datalogics-staylor , thanks for your contribution.
Is this a case where it might make sense for the package to just build _both_ libraries, C and C++? if the resulting library files have a different name, they can coexist, and then these can be different components? I suspect that's the case with other libraries in the same situation as well.

datalogics-staylor:
Hi @jcar87. That's a very good point. I looked around a bit for recipes that do both C and C++ and didn't find many, so I chose to go this route so that it doesn't build packages that aren't required. It would be a pretty minor change to just build both, so I'd be happy to go that route if that seems to make more sense. Conan and CCI are relatively new to me, so I'm still learning common usages. Thanks for the feedback!

uilianries:
@datalogics-staylor Your change looks correct in terms of CMake, but still need to update the Conanfile recipe in order to list the new library too. Could please update the pacakge_info as well?

recipes/qr-code-generator/all/conanfile.py
```diff
@@ -102,7 +102,7 @@ def package(self):
 
     def package_info(self):
         self.cpp_info.libs = [
-            "qrcodegen" if Version(self.version) < "1.7.0" else "qrcodegencpp"
+            "qrcodegen" if Version(self.version) < "1.7.0" else "qrcodegencpp", "qrcodegenc"
         ]
```
uilianries, line 106:
```suggestion
        ]
        self.cpp_info.components["qrcodegencpp"].set_property("cmake_target_name", "qr-code-generator::qrcodegencpp")
        self.cpp_info.components["qrcodegencpp"].libs = ["qrcodegen" if Version(self.version) < "1.7.0" else "qrcodegencpp"]
        self.cpp_info.components["qrcodegenc"].set_property("cmake_target_name", "qr-code-generator::qrcodegenc")
        self.cpp_info.components["qrcodegenc"].libs = ["qrcodegenc"]
```

I would suggest adding separated components too, so you can link to C++ or C only, by using cmake targets.

datalogics-staylor: Ah, that makes sense. I'm pretty new to both Conan and CMake, so I appreciate the feedback!

uilianries (approved):
Looks good. Both C++ and C libraries are available as separated targets.

toge (approved):

datalogics-robb:
@franramirez688 We're just waiting for one more review - the last approval was 2 weeks ago - can we get some eyes on this one for us?

Thanks!

jcar87 (approved):

AbrilRBS (approved):

PR #23902, opened 2024-05-07
uriparser: add version 0.9.8

toge:
Specify library name and version:  **uriparser/0.9.8**

https://github.com/uriparser/uriparser/compare/uriparser-0.9.7...uriparser-0.9.8

The new CMake option `URIPARSER_SHARED_LIBS` is added since 0.9.8.
But it's default value refers `BUILD_SHARED_LIBS` values.
So I think it is not needed to set `URIPARSER_SHARED_LIBS`.

```cmake
if(DEFINED BUILD_SHARED_LIBS)
    set(_URIPARSER_SHARED_LIBS_DEFAULT ${BUILD_SHARED_LIBS})
else()
    set(_URIPARSER_SHARED_LIBS_DEFAULT ON)
endif()
option(URIPARSER_SHARED_LIBS "Build shared libraries (rather than static ones)" ${_URIPARSER_SHARED_LIBS_DEFAULT})
```

---

### DIFF
--- a/recipes/uriparser/all/conandata.yml
+++ b/recipes/uriparser/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.9.8":
+    url: "https://github.com/uriparser/uriparser/releases/download/uriparser-0.9.8/uriparser-0.9.8.tar.xz"
+    sha256: "1d71c054837ea32a31e462bce5a1af272379ecf511e33448e88100b87ff73b2e"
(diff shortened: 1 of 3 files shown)


RubenRBS (approved):
Good catch, I agree with you regarding the new option :)

uilianries (approved):
Thank you for checking the diff of versions 😄

Issue #23647, opened 2024-04-18
[package] <lzham>/<cci.20220103>: linking statically doesn't work for msvc


soroosh-sdi:
### Description

In my project I successfully build again lzham's statically, but when I wanted to run the executable it fails and says it needs lzham's dll.

### Package and Environment Details

* Package Name/Version: **zlib/1.2.8**
* Operating System+version: **Linux Ubuntu 18.04**
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.version=193
compiler.runtime=static
os=Windows

### Steps to reproduce

git clone https://github.com/conan-io/conan-center-index.git

cd conan-center-index/recipes/lzham/all

conan create . --build=missing -tf=test_package --profile:b=./conan.profile --profile:h=./conan.profile --version=cci.2022010

### Logs

<details><summary>Click to expand log</summary>

```
======== Exporting recipe to the cache ========
lzham/cci.20220103: Exporting package recipe: D:\conan-center-index\recipes\lzham\all\conanfile.py
lzham/cci.20220103: exports: File 'conandata.yml' found. Exporting it...
lzham/cci.20220103: Calling export_sources()
lzham/cci.20220103: Copied 1 '.yml' file: conandata.yml
lzham/cci.20220103: Copied 1 '.py' file: conanfile.py
lzham/cci.20220103: Copied 6 '.patch' files
lzham/cci.20220103: Exported to cache folder: C:\Users\soroosh\.conan2\p\lzham3152bc4f48d77\e
lzham/cci.20220103: Exported: lzham/cci.20220103#104dc0cfb5e94d5422e4d616e6ba56d6 (2024-04-18 15:52:12 UTC)

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=static
compiler.runtime_type=Release
compiler.version=193
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=static
compiler.runtime_type=Release
compiler.version=193
os=Windows


======== Computing dependency graph ========
Graph root
    cli
Requirements
[...]

PR #23907, opened 2024-05-07
Add all the lzham lib files for msvc builds.

soroosh-sdi:
Specify library name and version:  **lzham/cci.20220103**

Adding all lzham library static libs to the `cpp_info.libs`
I describe the issue I found in the bug #23647.

---

### DIFF
--- a/recipes/lzham/all/conanfile.py
+++ b/recipes/lzham/all/conanfile.py
@@ -1,12 +1,14 @@
 import os
 
 from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 from conan.tools.files import (
     apply_conandata_patches,
     copy,
     export_conandata_patches,
     get,
+    replace_in_file,
     rmdir
 )
 from conan.tools.microsoft import (
@@ -51,6 +53,28 @@ class PackageConan(ConanFile):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
+    def validate(self):
+        if self.settings.os == "Windows" and self.settings.arch not in ["x86", "x86_64"]:
+            raise ConanInvalidConfiguration("On Windows, only x86 and x86_64 are supported")
+        
+    def _msvc_libs(self, targets=False):
+        arch = "x86" if self.settings.arch == "x86" else "x64"
+        suffix = f"{arch}D" if self.settings.build_type == "Debug" else arch
+        if self.options.shared == True:
+            if targets:
+                 # Note: this causes its dependencies to be built too
+                return ['lzhamdll']
+            
+            files = [f"lzham_{suffix}.dll", f"lzham_{suffix}.lib"]
+        else:
+            libs = ['lzhamcomp', 'lzhamdecomp', 'lzhamlib']
+            if targets:
+                return libs
+            
+            files = [f"{lib}_{suffix}.lib" for lib in libs]
+            
+        return files
+
     def layout(self):
         if is_msvc(self):
             vs_layout(self)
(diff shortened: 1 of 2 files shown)


recipes/lzham/all/conanfile.py
```diff
@@ -141,10 +156,13 @@ def package_info(self):
...
         if is_msvc(self):
-            lib_name = "lzham_x64"
+            lib_names = ["lzhamlib_x64", "lzhamcomp_x64", "lzhamdecomp_x64"]
             if self.settings.build_type == "Debug":
-                lib_name += "D"
-            self.cpp_info.libs = [lib_name]
+                # add D to the end of all the names in the lib_names list
+                lib_names = [f"{lib_name}D" for lib_name in lib_names]
+            print(self.settings.build_type)
+            print(lib_names)
```
jcar87, line 164:
```suggestion
```

RubenRBS (approved):

toge (approved):

jwillikers (approved):

jcar87:
Hi @soroosh-sdi , thank you for opening this PR and for the detailed description of the bug in https://github.com/conan-io/conan-center-index/issues/23647.

Upon reading the reported issue, this was telling:

>  In my project I successfully build again lzham's statically, but when I wanted to run the executable it fails and says it needs lzham's dll..

a static build of lzham should not result in missing DLLs from it. Looking further into this, it appears that on windows when `shared=False` this is building shared libraries, which is wrong. This PR doesn't actually fix this issue, it would just hide it further or potentially fail elsewhere -I'm not sure Conan 2 would propagate the runtime directories (PATH env var) if the library is reporting to be static. This is why it's very useful that you have opened an issue describing the problem, thank you for doing that! We will fix the underlying cause to address this.

on the other hand, and by no means a blocker - with the proposed changes there would be 5 consecutive `copy()` invocations that could be handled in a loop for better readability. Will address this too - it's just stylistic and not a major concern.

jcar87:
Hi @soroosh-sdi - I've pushed changes to your branch to fix the handling of static/shared on Windows. The Visual Studio solution was building _both_ the shared and static variants,  all were being copied, and the package_info was always pointing to link to the shared library, which was causing your issues. I've amended the recipe such that when `shared=False`, only the static library is built and packaged, and only the static libraries are advertised in `package_info()`

Additionally:
* added `validate()` method to fail on other windows architectures that are not supported (like arm)
* fixed handling of x86 architecture (x86_64 was hardcoded when it shouldn't have been)
* fixed issue where the vcxproj files were overriding the msvc runtime set by MSBuildToolchain
* fixed potential issue where the consumer of the static libraries may fail to link if using a different compiler toolset than the one used to produce the object files in the static libraries

tested all the following variants:

```
conan create . --version=cci.20220103 -s arch=x86 -s build_type=Debug
conan create . --version=cci.20220103 -s arch=x86 -s build_type=Release
conan create . --version=cci.20220103 -s arch=x86 -o "*:shared=True" -s build_type=Debug
conan create . --version=cci.20220103 -s arch=x86 -o "*:shared=True" -s build_type=Release

conan create . --version=cci.20220103 -s arch=x86_64 -s build_type=Debug
conan create . --version=cci.20220103 -s arch=x86_64 -s build_type=Release
conan create . --version=cci.20220103 -s arch=x86_64 -o "*:shared=True" -s build_type=Debug
conan create . --version=cci.20220103 -s arch=x86_64 -o "*:shared=True" -s build_type=Release
```

please note that you would have to include `lzham.h` instead of the specific shared or static ones. if you want to test this locally you'll have to fetch from the remote and reset your local branch.

soroosh-sdi:
Hi @jcar87 
Thanks for taking a deep look at problem and resolving the root cause.
I took a quick look at your changes and I totally agree with it.
I'm on vacation and will not have access to my build machine for three weeks, wanted to make sure you tested it with conan2 as well.

franramirez688 (approved):

RubenRBS (approved):

PR #23920, opened 2024-05-08
yomm2: new recipe

franramirez688:
Adding:  **yomm2/1.5.1**

Repository: https://github.com/jll63/yomm2

### DIFF
--- a/recipes/yomm2/all/conanfile.py
+++ b/recipes/yomm2/all/conanfile.py
@@ -7,7 +7,7 @@ from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout, CMakeDeps
 from conan.tools.files import get, rm, rmdir, copy
 from conan.tools.scm import Version
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=1.54.0"
 
 
 class yomm2Recipe(ConanFile):
--- a/recipes/yomm2/all/conanfile.py
+++ b/recipes/yomm2/all/conanfile.py
@@ -1,0 +1,98 @@
+import os
+
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout, CMakeDeps
+from conan.tools.files import get, rm, rmdir, copy
+from conan.tools.scm import Version
+from conans.errors import ConanInvalidConfiguration
+
+required_conan_version = ">=1.53.0"
+
+
+class yomm2Recipe(ConanFile):
+    name = "yomm2"
+    package_type = "library"
+    # Optional metadata
+    license = "BSL-1.0"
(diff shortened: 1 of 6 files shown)


recipes/yomm2/all/conanfile.py
```diff
@@ -0,0 +1,110 @@
...
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
+        if self.options.shared:
+            rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        else:  # header-only one
```
uilianries, line 100: Do you mean this project offer a header-only version? In that case, I would suggest adding `hearder_only` option in the recipe. The CI has a logic to process recipes that can be both library and header-only. The fmt is a good example: https://github.com/conan-io/conan-center-index/blob/master/recipes/fmt/all/conanfile.py#L23

RubenRBS: I agree with @uilianries, Cona can perfectly hanble the `header_only` option too and will set the package_type accordingly

franramirez688: Well, I do not think so. `fmt` offers that option and even creates a separate target if header-only, but `yomm2` does not do that. It creates the same target regardless of whether you are building a shared lib, but it behaves like a header-only or shared library. IMO, I think that adding that option could be confusing for consumers.

uilianries (dismissed):
LGTM

uilianries (dismissed):

uilianries (approved):

recipes/yomm2/all/conanfile.py
```diff
@@ -0,0 +1,116 @@
...
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "YOMM2")
+        self.cpp_info.set_property("cmake_target_name", "YOMM2::yomm2")
+        if self.options.header_only:
+            self.cpp_info.bindirs = []
+            self.cpp_info.libdirs = []
+        else:  # shared-library
+            self.cpp_info.libs = ["yomm2"]
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                self.cpp_info.system_libs.append("m")
```
franramirez688, line 116:
```suggestion
        else:  # shared-library
            self.cpp_info.libs = ["yomm2"]
        if self.settings.os in ["Linux", "FreeBSD"]:
            self.cpp_info.system_libs.append("m")
```

RubenRBS (approved):

uilianries (approved):

czoido (approved):

PR #23925, opened 2024-05-08
detools: refactor test package

ErniGH:


RubenRBS (approved):

RubenRBS (commented):
Where were the old files mentioned?

uilianries (approved):

PR #23940, opened 2024-05-09
confu_json version 1.0.1

werto87:
confu_json/1.0.1

version bump of confu_json and dependencies. using cmake install to discribe what needs to be coppied.
Switched from
include(${CMAKE_BINARY_DIR}/conanbuildinfo.cmake) conan_basic_setup(TARGETS)
to
find_package(magic_enum REQUIRED)


recipes/confu_json/all/conanfile.py
```diff
@@ -43,30 +43,24 @@ def _compilers_minimum_version(self):
...
 
     def layout(self):
-        basic_layout(self, src_folder="src")
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.81.0")
-        self.requires("magic_enum/0.8.2")
+        self.requires("boost/1.84.0")
+        self.requires("magic_enum/[>=0.9.5 <10]")
```
RubenRBS, line 50: While we're always on the lookout for more libraries to add, for now we only allow a small subset listed in https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges for version ranges, as those need careful consideration. Could you please pin this to the latest magic_enum release instead? Thanks!

recipes/confu_json/all/conanfile.py
```diff
@@ -43,30 +43,24 @@ def _compilers_minimum_version(self):
         }.get(self._min_cppstd, {})
```
RubenRBS, line 43: You can probably simplify this function now that only C++17 is needed :)

RubenRBS (commented):
Small review, thanks a lot for taking the time to keep the recipe update, we appreciate it :)

jcar87:
Apologies for the delay, this should already have been handled by https://github.com/conan-io/conan-center-index/pull/25374 - thanks!

PR #23942, opened 2024-05-09
sdf: refactor test package

ErniGH:


recipes/sdf/all/test_package/test_package.c
```diff
@@ -8,30 +8,8 @@
...
-        return 1;
-    }
-
-    sdfBuildDistanceField(dest_data, width, 2.0f, img_data, width, height, width);
-
-    free(dest_data);
-    stbi_image_free(img_data);
-
+    unsigned char* img_data = stbi_load("not a valid/path/fake_img.png", &width, &height, &bpp, 0);
+    printf("img_data: %p\n", img_data);
```
RubenRBS, line 13: It's probably UB to print a `NULL` pointer in this case, (https://stackoverflow.com/a/44996472) I'd keep the error message with the check in there, but make it return ok

SpaceIm:
It was already a minimal test package. Your proposed test package doesn't even call any function of sdf :/

ErniGH:
@SpaceIm Oh well noted. The test was very good and was actually quite small. The change is actually to remove the binary files from our test packages. 😄

recipes/sdf/all/test_package/test_package.c
```diff
@@ -8,27 +8,17 @@
...
+    unsigned char* img_data = stbi_load("not a valid/path/fake_img.png", &width, &height, &bpp, 0);
+    if (img_data == NULL)
     {
-        stbi_image_free(img_data);
-        return 1;
+        printf("Test\n"); // This should be printed always
     }
 
-    sdfBuildDistanceField(dest_data, width, 2.0f, img_data, width, height, width);
+    int result = sdfBuildDistanceField(dest_data, width, 2.0f, img_data, width, height, width);
```
SpaceIm, line 20: You can't give a NULL pointer as img_data to this function. If you want to drop binary image from conan-center repo, you'll have to programatically create pixels in this img_data. And no need to keep stb dependency anymore.

uilianries (dismissed):

SpaceIm (approved):

SpaceIm:
`stb` should be removed from requirements in test package

RubenRBS (approved):

SpaceIm:
There is still `stb` in conanfile of test_v1_package.

RubenRBS:
Great catch @SpaceIm :)

RubenRBS (approved):

toge (approved):

PR #23945, opened 2024-05-10
openssh: Allow all openssl 3.1.x releases

klausholstjacobsen:
openssh/*

Changed requirements to allow all openssl/3.1.x releases and not only 3.1.0

---

### DIFF
--- a/recipes/openssh/all/conanfile.py
+++ b/recipes/openssh/all/conanfile.py
@@ -66,7 +66,7 @@ class PackageConan(ConanFile):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
-        if self.version == "9.1p1":
+        if self.version in ["9.1p1", "9.6p1"]:
             # Backport configure script fix to accept OpenSSL versions in the 3.x series
             # See https://github.com/openssh/openssh-portable/commit/2eded551ba96e66bc3afbbcc883812c2eac02bd7
             replace_in_file(self, join(self.source_folder, "configure"), "300*", "30*")
--- a/recipes/openssh/all/conanfile.py
+++ b/recipes/openssh/all/conanfile.py
@@ -52,7 +52,7 @@ class PackageConan(ConanFile):
     def requirements(self):
         self.requires("zlib/[>=1.2.11 <2]")
         if self.options.with_libcrypto == "openssl":
-            self.requires("openssl/[>=1.1 <=3.1]")
+            self.requires("openssl/[>=1.1 <3.2]")
         elif self.options.with_libcrypto == "libressl":
             self.requires("libressl/3.9.1")
         if self.options.with_pam == "openpam":
--- a/recipes/openssh/all/conanfile.py
+++ b/recipes/openssh/all/conanfile.py
@@ -52,7 +52,10 @@ class PackageConan(ConanFile):
     def requirements(self):
         self.requires("zlib/[>=1.2.11 <2]")
         if self.options.with_libcrypto == "openssl":
-            self.requires("openssl/[>=1.1 <4]")
+            if self.version == "9.1p1":
+              self.requires("openssl/[>=1.1 <3.2]")
+            else:
+              self.requires("openssl/[>=1.1 <4]")                
         elif self.options.with_libcrypto == "libressl":
             self.requires("libressl/3.9.1")
         if self.options.with_pam == "openpam":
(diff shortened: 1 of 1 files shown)


RubenRBS (dismissed):
Thanks!

recipes/openssh/all/conanfile.py
```diff
@@ -52,7 +52,7 @@ def layout(self):
     def requirements(self):
         self.requires("zlib/[>=1.2.11 <2]")
         if self.options.with_libcrypto == "openssl":
-            self.requires("openssl/[>=1.1 <=3.1]")
+            self.requires("openssl/[>=1.1 <3.2]")
```
jcar87, line 55:
```suggestion
            self.requires("openssl/[>=1.1 <4]")
```

Is there a reason this is capped was capped to 3.1 and now to 3.2? This should have been `[>= 1.1 <4]` as OpenSSL releases are presumed to be backwards compatible, and this is what we are using in all other recipes that use a version range for OpenSSL: 

klausholstjacobsen: I have only verified up to and including openssl 3.1.5!
I think a minor version bump requires verification.
When I get to that I will submit!

RubenRBS: I think this was always meant to be <4 as per https://github.com/conan-io/conan-center-index/pull/19031#discussion_r1508596311 whenI first raised this problem

klausholstjacobsen: I just committed the suggestion anyway :-)

jcar87: Yes, I think that would be the best course of action, to avoid having to continually bump this.  OpenSSL is API stable as per their guidelines and documentation - they also tend to not change behaviour.

I'm checking the documentation of OpenSSH and I can't find specific versions of OpenSSL/libcrypto being prescribed - however if that turns out to be the case, we could revisit. Otherwise, if you would like to locally ensure that the recipe doesn't cause "new" versions of OpenSSL to be used as they are released, I would recommend using lockfiles for this.

jcar87:
```
checking for openssl/opensslv.h... yes
checking OpenSSL header version... 30200010 (OpenSSL 3.2.1 30 Jan 2024)
checking for OpenSSL_version... yes
checking for OpenSSL_version_num... yes
checking OpenSSL library version... openssh/9.1p1: 
configure: error: Unknown/unsupported OpenSSL version ("30200010 (OpenSSL 3.2.1 30 Jan 2024)")
```

Interesting error here - if it is the code from here: https://github.com/openssh/openssh-portable/blob/1c0d81357921f8d3bab06841df649edac515ae5b/configure.ac#L2864-L2888
then I don't see why _any_ version of >3 would fail, other than `3.0.4` due to a vulnerability.

Will investigate!

klausholstjacobsen:
Added differentiated openssl requirements for 9.1p1 and 9.6p1

jcar87:
> Added differentiated openssl requirements for 9.1p1 and 9.6p1

It looks like they have recently added this: https://github.com/openssh/openssh-portable/commit/2eded551ba96e66bc3afbbcc883812c2eac02bd7

with

> OpenSSL has moved to 3.4 which we don't currently accept.  Based on
> the OpenSSL versioning policy[0] it looks like all of the 3.x versions
> should work with OpenSSH, so remove the distinction in configure and
> accept all of them.

I think this confirms that it's okay for the recipe to simply resolve to any version in the 3.x series.

Another solution for the version that is failing:
```
    def source(self):
        if self.version == "9.1p1":
            # See https://github.com/openssh/openssh-portable/commit/2eded551ba96e66bc3afbbcc883812c2eac02bd7
            replace_in_file(self, join(self.source_folder, "configure"), "300*", "30*")
```

jcar87 (approved):

recipes/openssh/all/conanfile.py
```diff
@@ -66,6 +66,10 @@ def validate(self):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        if self.version in ["9.1p1", "9.6p1"]:
```
jcar87, line 69: version 9.1p1 would have not accepted version 3.2,
and version 9.6p1 would have not accepted versions higher than than 3.2, so the recipe would've started failing once OpenSSL 3.3 or higher was released. Newer versions should already have the fix to accept any in the 3.x range, as per the OpenSSL versioning guidelines

RubenRBS (approved):

PR #23946, opened 2024-05-10
date: support gcc14 on test_package

toge:
Specify library name and version:  **date/***

In gcc14 and msvc, there are compilation errors on building test_package.

---


RubenRBS:
Hi @toge thanks a lot! Posting the before and after logs for gcc would be great, as we're not yet building that configuration to test it ourselves, so it would help a ton for future traceability, thanks!

AbrilRBS (approved):

AbrilRBS (dismissed):
Thanks!

jcar87 (changes requested):
it shouldn't be necessary to publish a new recipe revision for this! 

note to team: the pipeline should be able to handle fixing a test package without requiring to publish recipe revisions (or rebuilding packages from new revisions)

jcar87:
> note to team: the pipeline should be able to handle fixing a test package without requiring to publish recipe revisions (or rebuilding packages from new revisions)

note to self - fixing this asap ;)

AbrilRBS (approved):
The fix to support test_package only changes is now live, thanks!

jcar87 (approved):
Thanks!
I've modified the test package to minimise it as much as possible - ensuring that when there is a library to test, we check a symbol from that library, and nothing else. 
This helps with ongoing maintenance and hopefully will be more robust to compiler updates...
(I did test it with gcc-14 on Linux)

AbrilRBS (approved):

PR #23949, opened 2024-05-10
jpeg_compressor: refactor test package

ErniGH:


RubenRBS (approved):

recipes/jpeg-compressor/all/test_package/test_package.cpp
```diff
@@ -5,29 +5,21 @@
 
 int main(int argc, const char* argv[])
```
uilianries, line 6: It could be simplified even more:

```
int main() {
    const char* pSrc_filename = "non-real-file.jpg";
	const int req_comps = 4;
    int width = 0, height = 0, actual_comps = 0;
    jpgd::decompress_jpeg_image_from_file(pSrc_filename, &width, &height, &actual_comps, req_comps);

    return EXIT_SUCCESS;
}
```

uilianries (approved):
LGTM

PR #23951, opened 2024-05-10
thorvg: add version 0.13.4

toge:
Specify library name and version:  **thorvg/0.13.4**

https://github.com/thorvg/thorvg/compare/v0.13.2...v0.13.4

---

### DIFF
--- a/recipes/thorvg/all/conanfile.py
+++ b/recipes/thorvg/all/conanfile.py
@@ -152,7 +152,8 @@ class ThorvgConan(ConanFile):
             tc.project_options["simd"] = bool(self.options.with_simd)
         else:
             tc.project_options["vector"] = bool(self.options.with_vector)
-        tc.project_options["extra"] = str(self.options.with_extra) if self.options.get_safe("with_extra") else ''
+        if self.options.get_safe("with_extra") is not None:
+            tc.project_options["extra"] = str(self.options.with_extra) if self.options.with_extra else ''
         tc.generate()
         tc = PkgConfigDeps(self)
         tc.generate()
--- a/recipes/thorvg/all/conanfile.py
+++ b/recipes/thorvg/all/conanfile.py
@@ -108,6 +108,11 @@ class ThorvgConan(ConanFile):
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
 
+        if is_msvc(self) and self.settings.build_type == "Debug" and self.options.shared:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} doesn't support shared debug build on MSVC."
+            )
+
     def requirements(self):
         loaders_opt = str(self.options.with_loaders)
         if loaders_opt in ("all", "jpg"):
--- a/recipes/thorvg/all/conanfile.py
+++ b/recipes/thorvg/all/conanfile.py
@@ -108,9 +108,9 @@ class ThorvgConan(ConanFile):
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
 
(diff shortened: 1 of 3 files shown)


recipes/thorvg/all/conanfile.py
```diff
@@ -131,7 +137,8 @@ def generate(self):
             "threads": bool(self.options.with_threads),
             "examples": bool(self.options.with_examples),
             "tests": False,
-            "log": is_debug
+            "log": is_debug,
+            "static": True,
```
franramirez688, line 141: As far as I saw, this option is used [here](https://github.com/search?q=repo%3Athorvg%2Fthorvg%20get_option(%27static%27)&type=code). It would be great to clarify why this is always `true` because the upstream option description sounds a bit confusing.

franramirez688 (commented):
@toge Thanks a lot for adding this new version! The PR looks great, but I saw that it failed in Windows (Conan 1.x version). I can help with that if you need it indeed.
Only a small suggestion about the `static` option.

recipes/thorvg/all/conanfile.py
toge: Thanks a lot for your comment!
I'm sorry it is my debug code.
Removed.

recipes/thorvg/all/conanfile.py
```diff
@@ -102,6 +108,11 @@ def validate(self):
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
 
+        if is_msvc(self) and self.settings.build_type == "Debug":
+            raise ConanInvalidConfiguration(
+                f"{self.ref} doesn't support debug build on MSVC."
+            )
```
franramirez688, line 114: I think that we could report this because it looks like a bug. It was working for the previous versions and it's failing now.

franramirez688 (commented):
Looking great! Only one concern about the MSVC and debug configuration.

franramirez688 (dismissed):

recipes/thorvg/all/conanfile.py
```diff
@@ -141,6 +152,8 @@ def generate(self):
             tc.project_options["simd"] = bool(self.options.with_simd)
         else:
             tc.project_options["vector"] = bool(self.options.with_vector)
+        if "with_extra" in self.options:
+            tc.project_options["extra"] = str(self.options.with_extra) if self.options.with_extra else ''
```
RubenRBS, line 156:
```suggestion
        if self.options.get_safe("with_extra"):
            tc.project_options["extra"] = self.options.with_extra or ''
```

RubenRBS (commented):
Thanks! Minor change as this looks a tad better, otherwise looks good

recipes/thorvg/all/conanfile.py
franramirez688: I think that we could make it simpler:

```suggestion
        tc.project_options["extra"] = str(self.options.get_safe("with_extra", ""))
```

franramirez688 (dismissed):

recipes/thorvg/all/conanfile.py
```diff
@@ -141,6 +152,7 @@ def generate(self):
             tc.project_options["simd"] = bool(self.options.with_simd)
         else:
             tc.project_options["vector"] = bool(self.options.with_vector)
+        tc.project_options["extra"] = str(self.options.with_extra) if self.options.get_safe("with_extra") else ''
```
franramirez688, line 155: Sorry, @toge @RubenRBS I just updated your comment and I was wrong:

```suggestion
        if self.options.get_safe("with_extra"):
            tc.project_options["extra"] = str(self.options.with_extra) or ''
```

franramirez688 (approved):

RubenRBS (approved):

PR #23955, opened 2024-05-10
Add recipe for the bmx library provided by the BBC

irieger:
Specify library name and version:  **bmx/1.2** and **bmx/cci.20240510**

Currently I'm implementing something using the bmx library for handling video IO. Obviously, as I'm really settled on Conan now, I first created a package for the library which is so far missing.

As I'm looking into some options just recently added, besides the most recent 1.2 release I also added a snapshot in the usual form of cci.DATE for the current master branch.

---


recipes/bmx/all/conanfile.py
```diff
@@ -0,0 +1,131 @@
...
+        # Required libraries
+        self.requires("uriparser/0.9.8")
+        self.requires("expat/2.6.2")
+
+        if not (self.settings.os == 'Windows' or self.settings.os == 'Macos'):
+            self.requires('libuuid/1.0.3')
+
+        # Configuration dependent requirements
+        if self.options.with_libcurl:
+            self.requires("libcurl/8.6.0")
```
RubenRBS, line 54: This can probably use a version range, check https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges, but probably something like `[>=7.78 <9]` if it supports older libcurl releases, else the minimum one :)

recipes/bmx/all/conanfile.py
```diff
@@ -0,0 +1,131 @@
...
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, 11)
```
RubenRBS, line 58: We probably also want to comment out https://github.com/bbc/bmx/blob/main/CMakeLists.txt#L37 so that users get to properly choose the standard version used :)

RubenRBS (commented):
Thanks! I'll have to check with the team about our handling of the -O2 directives present in the upstream cmakelist, but other than that and the few comments I've left, this looks great so far, thanks! :)

recipes/bmx/all/conanfile.py
irieger: Done. Good catch. This showed me another problem - which I added a second patch for and prepare the upstream PR now. (Will add the link to the PR when I have it ready.)

recipes/bmx/all/conanfile.py
irieger: Sure, I think that should likely not be a problem. Just added it. Also added the range for expat where I saw ranges are also suggested in the linked page.

RubenRBS:
I'll regenerate the missing binaries now and restart the PR once they are ready

irieger:
@RubenRBS Again as in some other PRs, the result of the Conan 2 CI step doesn't get published here so I can't see the errors. At least locally I had working builds on Linux, Mac and Windows already so I'd like to see what fails.

RubenRBS:
@irieger sorry about that. Sometimes the bot is a bit too eager to post th results and ends up not waiting for the v2 pipeline if it takes too long to finish - We're moving away from this solution in the new piepeline we're currently implementing, so that will be less of an issue soon-ish, thanks for your patience meanwhile!

The errors taht would have been posted are here: https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23955/5-windows-msvc/bmx/1.2//summary.json

Nevertheless I've now restarted the CI, the missing binaries were built a few days back but I haven't had a chance to trigger the CI again, let's see if that fixes the v1 piepeline completely :)

RubenRBS:
The fact the uriparser is still missing tells us that there' something going on with some of the dependecies, will check on Monday, thanks for your patience :)

RubenRBS:
Same v2 errors: https://c3i.jfrog.io/c3i/misc-v2/summary.html?json=https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23955/7-windows-msvc/bmx/1.2//summary.json

irieger:
Somehow it tries to link to MXF.lib instead of MXF.dll when building shared. I'm trying to understand why that is but so far I haven't found any hint. The directory to the target is correct and the dll is there. Any idea where such a thing could come from?
I'm not really experienced with Windows development and its behaviour and on both macOS and Linux it also correctly solves to either the .a or .so/.dylib.

RubenRBS:
@irieger I'll get back to you on that one tomorrow once I bounce the idea with @uilianries :)

RubenRBS:
I'll investigate the uriparser missing binary tomorrow, this is not a case of it not being built :)

irieger:
@RubenRBS Raised a PR there: https://github.com/bbc/bmx/pull/73

irieger:
Also, any news on the missing Conan 1 dependency?

Also I think there has to be another issue leading to https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23955/7-windows-msvc/bmx/1.2//83e068769ccbb48d8b698dfe31ce3ba8709e5cae-build.txt (It is a shared build and for some reason it tries to link MXF.lib, while correctly MXF.dll was built.)

irieger:
Patched out the `-O2` flag instances in the 1.2 release and bumped the CCI version to include the latest changes on master integrating all the fixes (fixed CXX version, -O2 flag and C++20 compatibility code fixes). That reduces the needed CMake patching to a minimal modification on how the dependencies are searched and linked.

RubenRBS:
Hi @irieger thanks for your patience - I tried to reproduce the windows issue but I found an issue where the dependency graph contains strawberryperl which does not support non x86 archs, I'll try with a different system, or pinging @uilianries who I know has a x86_64 windows :)

RubenRBS:
Update, I've been able to reproduce this locally on my personal machine, will try to debug this tomorrow :)

irieger:
> Hi @irieger thanks for your patience - I tried to reproduce the windows issue but I found an issue where the dependency graph contains strawberryperl which does not support non x86 archs, I'll try with a different system, or pinging @uilianries who I know has a x86_64 windows :)

No worries. I have days where I'm a bit nervous but since I setup my local artifactory a few months ago and run it with my own branch of CCI to allow some small modifications etc., I also occasionally pull in some stuff that isn't merged in upstream yet and can easily use it still.

irieger:
I digged a bit deeper and think I found why building as a shared library on Windows fails. As I learned now, it seems that Windows creates a lib file also for shared libraries, unlike Linux or macOS where you either get an a or a dylib/so. As this project doesn't define any exports in either the files nor via the Cmake flag to export all symbols on Windows.

Locally I tried it with a patch adding `set_target_properties(MXF PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)` and equivalent. While it doesn't fail with the error of not finding the MXF.lib then, it fails to find some symbol. I'll add an issue tomorrow on the bmx project github. But from looking at things now after some reading I think it might have just not been tried on Windows as a shared lib so far.

So I'd suggest - if this works out in the CI now - that we merge it with a validation to prevent windows shared builds. And if it gets fixed in the upstream I'll integrate the fix later. If we agree on that I'd add a descriptive comment for possible users in the validate function with a link to the issue after I create it tomorrow.

RubenRBS (approved):
Thanks! Let's go with this for now until we figure out what's going on in Windows, sorry we couldn't find the issue either in our end :)

ErniGH (approved):

jcar87 (approved):

PR #23963, opened 2024-05-12
add wgpu-native 0.19.4.1

Sahnvour:
Specify library name and version:  **wgpu-native/0.19.4.1**

WebGPU, despite its name, is not only usable from web browsers. [wgpu-native](https://github.com/gfx-rs/wgpu-native) is Mozilla's native implementation in rust.
It's useful to various graphical native applications and games.

This recipe uses prebuilt binaries from the project, to avoid dependency on the rust toolchain.

I haven't tried with hooks enabled, since they don't seem to be supported by conan 2.x.

---

### DIFF
--- a/recipes/wgpu_native/all/conandata.yml
+++ b/recipes/wgpu_native/all/conandata.yml
@@ -1,0 +1,48 @@
+sources:
+  "0.19.4.1":
+    binaries:
+      Linux:
+        "x86_64":
+          Release:
+            url: "https://github.com/gfx-rs/wgpu-native/releases/download/v0.19.4.1/wgpu-linux-x86_64-release.zip"
+            sha256: "7d73bd7af2be60b632e5ab814996acb381d1b459975d6629f91c468049c8866a"
+          Debug:
+            url: "https://github.com/gfx-rs/wgpu-native/releases/download/v0.19.4.1/wgpu-linux-x86_64-debug.zip"
+            sha256: "1feb5419a84eb2a5710b5686b8f8c30c70f24cea5c39774853c2c5406a3857fd"
+        "armv8":
+          Release:
+            url: "https://github.com/gfx-rs/wgpu-native/releases/download/v0.19.4.1/wgpu-linux-aarch64-release.zip"
+            sha256: "6e53aa3f0aec4b2b65cb0d7635000cf39bddd672bcb6138a593bf8cb8134f621"
+          Debug:
+            url: "https://github.com/gfx-rs/wgpu-native/releases/download/v0.19.4.1/wgpu-linux-aarch64-debug.zip"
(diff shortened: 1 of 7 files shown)


RubenRBS:
Hi @Sahnvour thanks a lot for taking the time to create the recipe and submit the PR, we really appreciate it!

I have a general question before being able to review this: What's the general use-case of this recipe in the context of Conan? Do you have any insight on that? Thanks! :)

Sahnvour:
> Hi @Sahnvour thanks a lot for taking the time to create the recipe and submit the PR, we really appreciate it!
> 
> I have a general question before being able to review this: What's the general use-case of this recipe in the context of Conan? Do you have any insight on that? Thanks! :)

WebGPU is a cross-platform low-level graphics API to do rendering with GPUs.
Although it's currently used in recent/nightly builds of web browsers, it can be used to make games among other things. You can think of it as an alternative to OpenGL or Vulkan.

Having it in conan is useful because it isn't packaged with the OSes or drivers, and allows to benefit from conan-center and local conan cache (I'm thinking of CI in particular).

PS: I'm working through the errors as they come from the github actions.

Sahnvour:
Hope this is clear enough @RubenRBS.

I have two questions:
- the prebuilt archives from the project do not contain any license files (and you can see it's [dual-licensed](https://github.com/gfx-rs/wgpu-native)), but this fails a hook, what's the best way to add them to the package?
- it seems `test_package` using `meson` is not supported in the pipeline? Should I change it to cmake?

RubenRBS:
Hi @Sahnvour check https://github.com/conan-io/conan-center-index/pull/23963/commits/072a28e3aa4455b3b7d596b8b2ed7ebffdb06f03 for the license and https://github.com/conan-io/conan-center-index/pull/23963/commits/7a500ac3ada3758d536d84116486ff1dded49875 for how to use something other than cmake in the test_package (We usually prefer CMake because we assume it's installed in the agents, unlike other build tools)

The first commit seems to have a few linter issues, let me quickly fix it :)

Sahnvour:
I think I made progress but I'm not sure why the link fails, since the recipe adds the required linux libraries (I don't know about MacOS yet though). It also works both in static and shared on my machine. Do you have any idea @RubenRBS ?

Sahnvour:
Hi, I'm still interested in getting this merged, but was blocked at the time. Let me know how we can go further.

Sahnvour:
Hi @AbrilRBS, sorry to ping you directly, I'm just checking if you might have a clue or if I can do something about it.

Sahnvour:
I've done everything I could and I think it just needs a last notch to get it working (see previous comments).
With a working recipe I'd be able to add newer versions too.

ping @ErniGH for visibility

PR #23978, opened 2024-05-13
cpp-validator: add version 2.1.1

toge:
Specify library name and version:  **cpp-validator/2.1.1**

https://github.com/evgeniums/cpp-validator/compare/v2.0.3...v2.1.1

---


RubenRBS (commented):
I'm checking https://github.com/evgeniums/cpp-validator/blob/main/docs/index.md#building-and-installation (To see if it makes sense to open a PR upstream to mention Conan support) - and noticed the bit about fmt backend needing a define to properly work. Isn't this something that should be reflected in the recipe? Both as an option, and in the package_info? wdyt @toge?

memsharded:
> I'm checking https://github.com/evgeniums/cpp-validator/blob/main/docs/index.md#building-and-installation (To see if it makes sense to open a PR upstream to mention Conan support) - and noticed the bit about fmt backend needing a define to properly work. Isn't this something that should be reflected in the recipe? Both as an option, and in the package_info? wdyt @toge?

Maybe the recipe can hardcode at the moment the fmt backend, as it is recommended and preferred, but the ``HATN_VALIDATOR_FMT`` macro seems it should be defined in consumers, not sure why the ``test_package`` is not failing.

toge:
@AbrilRBS  @memsharded 

If HATN_VALIDATOR_FMT is not defined, the test_package will work fine because it just does not use fmt.

This is the description for this recipe because it follows the previous recipes.
I will wait for the conclusion in cpp-lazy and modify it because it depends on the policy of whether to provide options for the header only library, the same as I commented in cpp-lazy.
 https://github.com/conan-io/conan-center-index/pull/23236#discussion_r2009095053

PR #23992, opened 2024-05-14
tinygltf: refactor test package

ErniGH:


recipes/tinygltf/all/test_package/test_package.cpp
```diff
@@ -8,29 +8,13 @@
...
-  if (!err.empty()) {
-    printf("Err: %s\n", err.c_str());
-  }
-
-  if (!ret) {
-    printf("Failed to parse glTF\n");
-    return -1;
-  }
+  bool ret = loader.LoadBinaryFromFile(&model, &err, &warn, "non_existent_file.glb");
+  printf("Test %d\n", ret);
```
uilianries, line 17:
```suggestion
    loader.LoadBinaryFromFile(&model, &err, &warn, "non_existent_file.glb");
```

I would not check anything as it will always fail.

uilianries (approved):

toge (approved):

PR #23996, opened 2024-05-14
read-excel: refactor test package

ErniGH:


recipes/read-excel/all/test_v1_package/conanfile.py
```diff
@@ -14,5 +14,4 @@ def build(self):
     def test(self):
         if not tools.cross_building(self):
```
lucaskdc, line 15: You could improve this condition using ```can_run``` since the conan version required for this package is >= 1.5.0

reference: https://docs.conan.io/1/reference/conanfile/tools/build.html#conan-tools-build-can-run

lucaskdc (approved):

recipes/read-excel/all/test_v1_package/conanfile.py
RubenRBS: For the Conan v1 test packages we're actually keeping the `cross_building` check in place so that said flow is tested. The normal `test_package` will then test the `can_run` in both Conan 1 and 2 :)

RubenRBS (approved):

uilianries (approved):

PR #23997, opened 2024-05-14
mini: add recipe

toge:
Specify library name and version:  **mini/***

---

### DIFF
--- a/recipes/mini/all/test_package/test_package.cpp
+++ b/recipes/mini/all/test_package/test_package.cpp
@@ -1,0 +1,14 @@
+#include <iostream>
+
+#include "mini/ini.h"
+
+int main(void) {
+    mINI::INIFile file("test_package.ini");
+
+    mINI::INIStructure ini;
+
+    ini["things"]["chairs"] = "20";
+    ini["things"]["balloons"] = "100";
+
+    file.generate(ini);
+}
--- a/recipes/mini/all/test_package/test_package.cpp
+++ b/recipes/mini/all/test_package/test_package.cpp
@@ -5,10 +5,6 @@
 int main(void) {
     mINI::INIFile file("test_package.ini");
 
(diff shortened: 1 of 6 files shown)


recipes/mini/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+int main(void) {
+    mINI::INIFile file("test_package.ini");
+
+    mINI::INIStructure ini;
+
+    ini["things"]["chairs"] = "20";
+    ini["things"]["balloons"] = "100";
+
+    file.generate(ini);
```
RubenRBS, line 13: Let's try to avoid writing the a file in the test_package if possible :)

RubenRBS (commented):
Thanks! The recipe looks great, I only have a minor request in the test package :)

recipes/mini/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,14 @@
...
+}
```
jcar87, line 14:
```suggestion
#include <iostream>

#include "mini/ini.h"

int main(void) {
    mINI::INIFile file("test_package.ini");

    std::cout << "mini test successful \n";
    return 0;
}
```

recipes/mini/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,10 @@
+#include <iostream>
+
+#include "mini/ini.h"
+
+int main(void) {
+    mINI::INIFile file("test_package.ini");
```
jcar87, line 6: @RubenRBS  I've updated the test_package - its a header only. library, so no need to check for linkage, and the constructor does not write a file to disk (it does nothing).

RubenRBS (approved):

PR #24000, opened 2024-05-15
Add nifti_clib

Julianiolo:
Specify library name and version:  **nifti_clib/3.0.0**

This is a dependency of libvips, which is why I am adding it :)

---


Julianiolo:
One thing to note is the build seems to download extra data for testing, should this be disabled?

RubenRBS:
> One thing to note is the build seems to download extra data for testing, should this be disabled? (This can be done via setting `tc.variables["NIFTI_BUILD_TESTING"] = False` at line 83)
> 
> Edit: I disabled it for now.

@Julianiolo yes, recipes in CCI should not by default build their test targets :)

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
...
+        "shared": False,
+        "fPIC": True,
+        "use_nifti2": True,
+        "use_cifti": False,  # seems to be beta?
+        "use_fslio": False  # Note in CMakeLists.txt: "If OFF, The copyright of this code is questionable for inclusion with nifti."
+    }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
```
RubenRBS, line 38:
```suggestion
    def export_sources(self):
        export_conandata_patches(self)

```

Can be removed as there are no patches :)

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rm, rmdir
```
RubenRBS, line 4:
```suggestion
from conan.tools.files import copy, get, replace_in_file, rm, rmdir
```

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
...
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("zlib/1.3.1")
+        if self.options.use_cifti:
+            self.requires("expat/2.6.2")
```
RubenRBS, line 55:
```suggestion
        self.requires("zlib/[>=1.2.11 <2]")
        if self.options.use_cifti:
            self.requires("expat/[>=2.6.2 <3]")
```

These two libraries are some of the ones listed in https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#adherence-to-build-service so that version ranges are usable with them :)

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
...
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["NIFTI_INSTALL_NO_DOCS"] = True
+        tc.variables["USE_NIFTI2_CODE"] = self.options.use_nifti2
+        tc.variables["USE_CIFTI_CODE"] = self.options.use_cifti
+        tc.variables["USE_FSL_CODE"] = self.options.use_fslio
+        tc.variables["NIFTI_BUILD_TESTING"] = False  # maybe this should be false? It downloads extra test data
+        if is_msvc(self):
+            tc.variables["USE_MSVC_RUNTIME_LIBRARY_DLL"] = not is_msvc_static_runtime(self)
+            tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
```
RubenRBS, line 69: This is something we're trying to move away from.

Unless the library’s build-system is already using it, we should avoid using this CMake functionality because it interacts poorly with C++ templates and breaks certain compiler features. Libraries that don't provide a `.def` file and do not use `__declspec()` declarations simply do not support shared builds for Windows, and we should not be providing configurations that are not supported by upstream


Maybe upstream could be conviced to add proper Windows support for this configuration?

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
...
+            tc.preprocessor_definitions["_CRT_SECURE_NO_WARNINGS"] = 1
+        tc.generate()
+        tc = CMakeDeps(self)
+        tc.generate()
+        tc = VirtualBuildEnv(self)
+        tc.generate(scope="build")
```
RubenRBS, line 75: This is only necessary if your package contains build_requires :)
```suggestion
```

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+
+    def build(self):
+        self._patch_sources()
```
RubenRBS, line 81:
```suggestion
    def build(self):
```

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,160 @@
...
+
+        if is_msvc(self):  # patch in __declspec(dllimport) for a global variable that is used
+            to_replace = "extern char const * const inam[];"
+            fixed = "__declspec(dllimport) " + to_replace
+            replace_in_file(self, os.path.join(self.source_folder, "nifticdf", "nifticdf.h"), to_replace, fixed, strict=True)
```
RubenRBS, line 86: This is exactly the kind of changes we're trying to avoid - releated to my previous msvc comments (Either way, thanks for taking the time to try to "fix it", we appreciate it nontheless)

RubenRBS (commented):
Thanks a lot for taking the time to create the PR, we really appreciate it :)

I have some comments, but most of the PR looks great. Happy to help where needed :)

Julianiolo:
Uh is there a way to tell the test that mscv shared is not supported?

uilianries:
> Uh is there a way to tell the test that mscv shared is not supported?

@Julianiolo The test package should not be skipped for any configuration. Any produced package must be validated. In case some configuration does not work, you can avoid building it, but need a reason (e.g the upstream does not support it, you faced an error and does not know how to fix)

```
def validate(self):
        # INFO: Describe why it does not work, with upstream reference in case having one.
        raise ConanInvalidConfiguration(f"{self.ref} does not support -o {self.ref}:shared=True with MSVC compiler.") 
```

RubenRBS:
Hi @Julianiolo after talking with @uilianries we realized that this case is better suited to not model shared as an option for Windows, so that every configuration finds it as a static-library, sorry for the noise :)

Julianiolo:
What is the difference between the two?

Julianiolo:
@RubenRBS I'm not sure if this brings any advantages? I'm currently working on getting nifti_clib to support msvc shared builds.

RubenRBS:
The current one makes it so that dependencies that get built in shared mode do get built, instead of being validated out in this package validate() method. But, if you're working on supporting it, we cna go back to the old version and have it be valid once it's ready! SOrry for the noise :)

RubenRBS (dismissed):

danimtb (approved):

recipes/nifti_clib/all/conanfile.py
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def validate(self):
+        if is_msvc(self) and self.options.shared:
+            # not supported due to not having dllexport definitions
+            raise ConanInvalidConfiguration(f"{self.ref} does not support -o {self.ref}:shared=True with MSVC compiler.")
```
uilianries, line 43:
```suggestion
            raise ConanInvalidConfiguration(f"{self.ref} does not support -o {self.ref}:shared=True with MSVC compiler.")
        if not self.options.use_nifti2 and self.options.use_cifti:
            raise ConanInvalidConfiguration(f"{self.ref} -o '&:use_cifti=True' requires -o '&:use_nifti2=True'")
```

I see a strict relation between USE_CIFTI_CODE and USE_NIFTI2_CODE:

[NIFTI-Imaging/nifti_clib@v3.0.0/CMakeLists.txt#L170](https://github.com/NIFTI-Imaging/nifti_clib/blob/v3.0.0/CMakeLists.txt#L170)

The dependency means USE_CIFTI_CODE will be available only when USE_NIFTI2_CODE is enabled.

Testing this configuration locally, it builds/installs libnifticdf.a, but not libnifti2:

```
-- Install configuration: "Release"
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/lib/libznz.a
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/include/nifti/znzlib.h
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/lib/libniftiio.a
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/include/nifti/nifti1_io.h
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/include/nifti/nifti1.h
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/bin/nifti1_tool
-- Set runtime path of "/home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/bin/nifti1_tool" to "$ORIGIN/../lib"
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/lib/libnifticdf.a
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/include/nifti/nifticdf.h
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/bin/nifti_stats
-- Set runtime path of "/home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/bin/nifti_stats" to "$ORIGIN/../lib"
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/share/cmake/NIFTI/NIFTITargets.cmake
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/share/cmake/NIFTI/NIFTITargets-release.cmake
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/share/cmake/NIFTI/NIFTIConfig.cmake
-- Installing: /home/conan/.conan2/p/b/nifti2b784e1dfc2bd/p/share/cmake/NIFTI/NIFTIConfigVersion.cmake
```

Julianiolo:
@uilianries true! I missed that

RubenRBS (approved):
Thanks a lot for your work, we really appreciate it :)

uilianries (approved):
LGTM

PR #24010, opened 2024-05-16
hermes: add recipe

toge:
Specify library name and version:  **hermes/***

---


recipes/hermes/all/conandata.yml
```diff
@@ -0,0 +1,9 @@
+sources:
+  "0.12.0":
```
RubenRBS, line 2: Wonder if they'll ever release a new tagged version

recipes/hermes/all/conandata.yml
```diff
@@ -0,0 +1,9 @@
...
+    url: "https://github.com/facebook/hermes/archive/refs/tags/v0.12.0.tar.gz"
+    sha256: "bd8fd158381813483123eb1ab553ed08db68e4949f314c99ee8fa79fa8f3e7ed"
+patches:
+  "0.12.0":
+    - patch_file: "patches/0001-include-cstdint.patch"
```
RubenRBS, line 7: Has this been reported upstream? it's quite common for libraries not to import the necessary headers, but I think they'd appreciate the report :)

recipes/hermes/all/conanfile.py
```diff
@@ -0,0 +1,92 @@
...
+required_conan_version = ">=1.53.0"
+
+class HermesConan(ConanFile):
+    name = "hermes"
+    description = "A JavaScript engine optimized for running React Native."
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/facebook/hermes"
+    topics = ("javascript", "ahead-of-time", "react native")
+    settings = "os", "arch", "compiler", "build_type"
```
RubenRBS, line 18: missing `package_type` :)

recipes/hermes/all/conanfile.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("icu/74.2")
```
RubenRBS, line 57: Seems like this would require cpython? https://github.com/facebook/hermes/blob/8a272222bdd97435e1c4f2bae9ad207cff8369af/CMakeLists.txt#L81 or am I missing something?

RubenRBS (commented):
Thanks! Some initial comments :)

recipes/hermes/all/conandata.yml
toge: @RubenRBS 
I missed tagged versions.
I will update it

recipes/hermes/all/conandata.yml
toge: This is already fixed in upstream. Thanks!

toge:
@RubenRBS 
It seems to be same issue.
https://github.com/conan-io/conan-center-index/pull/24011#issuecomment-2117802290

RubenRBS:
We have now fixed this, relaunching! :)

recipes/hermes/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "2024-04-29":
```
RubenRBS, line 2: For Conan not to get confused, we'll need to remap this to `2024.04.29`, else it will think that this is version `2024` with prerelease version `04` which which has a prerelese `29` :)

RubenRBS, line 2:
```suggestion
  "2024.04.29":
```

recipes/hermes/config.yml
```diff
@@ -0,0 +1,3 @@
+versions:
+  "2024-04-29":
```
RubenRBS, line 2:
```suggestion
  "2024.04.29":
```

memsharded:
I have been trying to make this work, but it is being a bit of a challenge:

- First, I have tried to reduce to create only ``shared-library``, as it seems the most recommended way by the hermes project, but mainly because the project CMakeLists.txt was ignoring the recipe inputs and building always a shared library, irrespective all the CMakeToolchain variables.
- Then, I was having issues with the ``tool_requires`` to cpython, so I have replaced it with:
```diff
def build_requirements(self):
-        self.tool_requires("cpython/3.12.2")
+    def validate_build(self):
+        if getattr(sys, "frozen", False):
+            raise ConanInvalidConfiguration("This package cannot be built with a Conan self-contained executable, "
+                                            "needs full Python installation")
```
And it seems it works, both in Windows and Linux so far, so I think it is worth removing that moving piece.
- Then, I added this attempt to configure the toolchain:
```diff
-        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
-        tc.variables["HERMES_USE_STATIC_ICU"] = not self.dependencies["icu"].options.shared
-        tc.variables["HERMES_ENABLE_TEST_SUITE"] = False
+        tc.cache_variables["HERMES_USE_STATIC_ICU"] = not self.dependencies["icu"].options.shared
+        tc.cache_variables["HERMES_ENABLE_TEST_SUITE"] = False
+        tc.cache_variables["HERMES_IS_ANDROID"] = self.settings.os == "Android"
+        tc.cache_variables["HERMES_ENABLE_WIN10_ICU_FALLBACK"] = False
```
setting things like Android configuration.
- I also made a fix for Windows, because it was adding ``dl`` library always, which is incorrect

Still, my Windows build is failing in the ``test_package`` step, because failing to link the ``facebook::jsi``, which seems it can be a failure to export symbols in Windows.

I am going to push the changes, I think it might help a bit, but please let me know.

PR #24029, opened 2024-05-19
openssh: Added ability to control "--disable-strip" configure argument

klausholstjacobsen:
openssh/*

openssh: Added ability to control "--disable-strip" configure argument

---

### DIFF
--- a/recipes/openssh/all/conanfile.py
+++ b/recipes/openssh/all/conanfile.py
@@ -25,6 +25,7 @@ class PackageConan(ConanFile):
         "with_pam": [False, "openpam"],  # linux-pam and Solaris PAM are also supported
         "with_selinux": [True, False],
         "with_libedit": [True, False],
+        "with_strip": [True, False],
         "with_sandbox": [False, "auto", "capsicum", "darwin", "rlimit", "seccomp_filter", "systrace", "pledge"]
     }
     default_options = {
@@ -32,6 +33,7 @@ class PackageConan(ConanFile):
         "with_pam": False,
         "with_selinux": False,
         "with_libedit": False,
+        "with_strip": True,
         "with_sandbox": "auto"
     }
 
@@ -80,6 +82,9 @@ class PackageConan(ConanFile):
 
         tc = AutotoolsToolchain(self)
         tc.configure_args.append("--without-zlib-version-check")
+
+        if not self.options.with_strip:
+            tc.configure_args.append("--disable-strip")
 
         if self.options.with_selinux:
             tc.configure_args.append("--with-selinux")


RubenRBS (commented):
Thanks! As the CI does not build non default configurations by default, could you please provide successful compilation logs for `&:with_strip=False`? Thanks!

Do ping me once you provide them, else I might miss the notification :)

klausholstjacobsen:
@RubenRBS Here you go!

`conan create recipes/openssh/all/conanfile.py --version 9.1p1 --build openssh/* --user khj --channel testing -o:h openssh/*:with_strip=False`
[build_with_strip_false.log](https://github.com/conan-io/conan-center-index/files/15376376/build_with_strip_false.log)

`conan create recipes/openssh/all/conanfile.py --version 9.1p1 --build openssh/* --user khj --channel testing -o:h openssh/*:with_strip=True`
[build_with_strip_true.log](https://github.com/conan-io/conan-center-index/files/15376377/build_with_strip_true.log)

The difference is when install is called with or without `-s`.
With strip:
`/usr/bin/install -c -m 0755 -s ssh /home/khj/.conan2/p/b/opens59d9fc54f5913/p//bin/ssh`
without strip:
`/usr/bin/install -c -m 0755  ssh /home/khj/.conan2/p/b/opens9de2caf756580/p//bin/ssh`

RubenRBS (approved):

RubenRBS (approved):
Thanks a lot, really appreciated :)

RubenRBS (commented):
Hi @klausholstjacobsen I have a question about your specific strip needs in this case. In what case is controlling the disabling of stripping useful for you? I'd have guessed that embeded platforms would prefer to always set stripping on, but I'm looking for a bit more insight to understand, thanks!

toge (approved):

jcar87:
Apologies, but this should not have been merged without a use case being explained, as this can arguably be applied to any library built with autotools. 
For The 'Release' configuration, if strip is the default, I see no reason why this would be changed to anything else.

This PR may inadvertently cause the Debug binaries to be stripped of the debug symbols as well, by default, which would be undesirable.

@klausholstjacobsen any details as to why this needed would be great for our understanding, otherwise I think this needs to be reversed. Perhaps the most suitable build_type if you want to skip the stripping, is `DebWithRelInfo`

klausholstjacobsen:
@jcar87 sorry for my delayed response.
The with_strip options causes autotools to call strip with "-s".
I have not found a way to tell autotools which install command to use, so it uses the one installed on the build system.
This however fails when cross compiling due to missing shared libraries.
I am no expert in what install does, but apparently it needs to resolve shared libraries for the component being installed when called with the "-s" argument!

jcar87:
Thanks @klausholstjacobsen !!

is there an error we can use to reproduce?
I'd have thought the autotools scripts + install can handle situations where `-s` should not be called - it would be good to see if this is an issue specific to this library, or generalised to all autotools-based libraries!

PR #24033, opened 2024-05-20
glew: add option for glu

xyz1001:
Specify library name and version:  **glew/2.2.0**

glew with mesa-glu cause crash on my device. `glu/system` works well. 
Environment:
OS: Kubuntu 22.04
Hardware: Lenovo ThinkBook 14p G2 ACH with AMD 5800H

---


valgur:
@jwillikers

jwillikers:
@xyz1001 Thanks for submitting the PR regarding this issue! I'm more curious to understand exactly why it would crash at runtime. Are you running your application with the `conanrun.sh` shell script? That's important as it will put the runtime-loaded `libGLU` from the standard Conan package in `LD_LIBRARY_PATH` path.

xyz1001:
I am not running my application with `conanrun.sh`. I collect all so file from dependencies and copy them to the directory of executable file. And then I run my application with `LD_LIBRARY_PATH` set to the directory of executable file. 
image

It crash when call `glDrawArrays`. 
image

jwillikers:
@xyz1001 Thanks! That helps me understand the situation. I'm certain GLU is the issue yet. What happens if you delete `libglvnd` from the `.so` files you've copied and try re-running?

xyz1001:
@jwillikers It works well after I delete all so copy from `libglvnd`. It turn to load those so from system packages.

jwillikers:
> @jwillikers It works well after I delete all so copy from `libglvnd`. It turn to load those so from system packages.

@xyz1001 Thanks for checking. I think making `opengl/system` an alternative for `libglvnd` in `mesa/glu` would solve your problem. I'm not sure the issue is related to the `libglvnd` Conan package or just the fact that a lot of the dependencies like `glew` depend on `opengl/system` still instead of the `libglvnd` Conan package. To figure out what's really going would require a bigger deep dive into running with some different environment variables to debug the `libglvnd` loading and the Mesa drivers. I don't think that's really necessary at this point though.

xyz1001:
@jwillikers Yes, I am using the latest revision of the `libglvnd` recipe. The `conanfile.py` in my local conan cache is same as that in conan-center-index repo.

Issue #20082, opened 2023-09-22
[bug] arrow 13.0.0 not building for msvc debug profile


oOTimothyOo:
### Environment details

* Operating System+version: windows 11

* Compiler+version: 
[settings]
arch=x86_64
compiler=msvc
compiler.version=193
compiler.runtime=dynamic
build_type=Debug

* Conan version: 2.0.9
* Python version: 3.11.5

If I try to build arrow 13.0.0 for the profile above it will fail with cmake error 
![image](https://github.com/conan-io/conan/assets/62081758/7b4ad03b-3cb4-424b-b8fe-5c88f460910e)

I tried setting with_thrift on True but this did not work

### Steps to reproduce
make a conanfile with dependency  on arrow/13.0.0 with option parquet on True

PR #24011, opened 2024-05-16
arrow: add version 16.1.0

toge:
Specify library name and version:  **arrow/16.1.0**

https://github.com/apache/arrow/compare/apache-arrow-16.0.0...apache-arrow-16.1.0

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24044, opened 2024-05-20
arrow: fix recipe when arrow/*:parquet=True

robomics:
Specify library name and version:  **arrow/all**

When the arrow package is built with `arrow/*:parquet=True`, `arrow/*:with_thrift=True` is also required.
Thrift is found by Arrow by calling [FindThriftAlt.cmake](https://github.com/apache/arrow/blob/5809dafb0f25b4bbf65387b0206d0f94b391c0b9/cpp/cmake_modules/FindThriftAlt.cmake#L48-L64), which uses some custom logic to define `thrift::thrift` when the project is compiled on Windows.

This causes the `transitive_header=True` specified in Thrift's recipe to be ignored.
https://github.com/conan-io/conan-center-index/blob/b4b8a8dac707a71de0f0552970dfb809146a687f/recipes/thrift/all/conanfile.py#L69-L70

This PR patches [FindThriftAlt.cmake](https://github.com/apache/arrow/blob/5809dafb0f25b4bbf65387b0206d0f94b391c0b9/cpp/cmake_modules/FindThriftAlt.cmake#L48-L64) so that `find_package(Thrift)` is always called.

Fixes https://github.com/conan-io/conan-center-index/issues/20082 (note that without the patch I am unable to build the package using MSVC regardless of the `build_type`).

While debugging the above issue I also realized that `ARROW_WITH_RE2` was not explicitly set by the recipe, which on my machine causes compilation to fail for v11.0.0. This PR also addresses this issue.

Finally, I am bumping `thrift`, `xsimd`, and `zstd`to avoid version conflicts with other packages on cci.

---


valgur:
Thanks! Please also set `parquet=True` in default_options as well. It was only left disabled due to build issues in the last PR and this one should probably fix these.

robomics:
Another thing I noticed, is that arrow always uses a vendored version of mimalloc: https://github.com/apache/arrow/blob/b2e8c33c86c819b167a1cbca834da3c9047a9350/cpp/cmake_modules/ThirdpartyToolchain.cmake#L2197-L2204

Should I drop mimalloc from the requirement list (but leave the option to toggle mimalloc on or off)?

robomics:
@RubenRBS I think this is ready for review :).

The conan v2 pipeline failure seems to be the same described in https://github.com/conan-io/conan-center-index/pull/24011.

RubenRBS:
We're fixing the conflict from Conan 2 in a separate PR, will re-trigger once it gets fixed :)

robomics:
@RubenRBS any updates regarding the PR?

AbrilRBS:
Sorry @robomics the name change made me miss the notification.

The idea was to bump xsimd here too, but we ran into some issues there, which we're trying to address before pushing a PR out, sorry for the delay

Also, regarding
> Should I drop mimalloc from the requirement list (but leave the option to toggle mimalloc on or off)?

My main concern here is whether the minmalloc symbols are leaked to the dependencies of this library. If so, keeping mimalloc as a dependency might be a nice solution to avoid duplicated symbols!

robomics:
@AbrilRBS, @uilianries any updates on this?

I kindof need to build arrow with parquet support on windows in one of my projects.

I am more than happy to open other PRs to fix version conflicts with e.g. xsimd.
Just let me know how can I be of help :)

robomics:
uhm... It looks like the `thrift` package for v0.20.0 is missing (at least for Linux GCC).
Would you mind looking into this @AbrilRBS?

perseoGI:
@robomics building missing binaries now...

recipes/arrow/all/conandata.yml
```diff
@@ -48,24 +48,83 @@ sources:
...
+      patch_type: "conan"
+    - patch_file: "patches/8.0.0-0002-install-utils.patch"
       patch_description: "enable utils installation"
       patch_type: "conan"
-    - patch_file: "patches/8.0.0-0006-fix-cmake.patch"
+    - patch_file: "patches/8.0.0-0003-fix-cmake.patch"
       patch_description: "use cci package"
       patch_type: "conan"
+    - patch_file: "patches/8.0.0-0004-fix-msvc-build.patch"
+      patch_description: "see commit a53f2bda5dfa50f2d1dbfd9d7fbc98d8c9725564 from arrow's repo"
```
perseoGI, line 119: Maybe add a more descriptive description and a `patch_source` to the commit.
Or preferably the PR where this were introduced

recipes/arrow/all/patches/8.0.0-0002-install-utils.patch
```diff
@@ -1,8 +1,8 @@
 diff --git a/cpp/CMakeLists.txt b/cpp/CMakeLists.txt
-index aba18c8..bb463d0 100644
+index 5eec4ea72..301f0074c 100644
 --- a/cpp/CMakeLists.txt
 +++ b/cpp/CMakeLists.txt
-@@ -721,7 +721,7 @@ if(ARROW_WITH_BZ2)
+@@ -717,7 +717,7 @@ if(ARROW_WITH_BZ2)
```
perseoGI, line 5: Just for understanding, why did this line numbers change?

recipes/arrow/all/patches/11.0.0-0001-fix-cmake.patch
```diff
@@ -0,0 +1,42 @@
+diff --git a/cpp/cmake_modules/FindThriftAlt.cmake b/cpp/cmake_modules/FindThriftAlt.cmake
+index f3e49021d..95177c2a6 100644
+--- a/cpp/cmake_modules/FindThriftAlt.cmake
++++ b/cpp/cmake_modules/FindThriftAlt.cmake
+@@ -45,22 +45,21 @@ endif()
+ #   * https://github.com/apache/thrift/pull/2725
```
perseoGI, line 6: A good practice is to put this links in the `patch_source`, see https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conandata_yml_format.md#patch_source

robomics: None of these links are patch sources.
These are just some of the issues mentioning problems related to finding Thrift that I stumbled upon while looking for existing solutions to the build failure I was debugging.

perseoGI (approved):

perseoGI:
> Thanks @perseoGI! CI seems to be green now.
> 
> Can you please review this (and also #24934, #24935, and #24936 if appropriate)?

Could you refresh a lit my memory?
Those PRs are pointing to this one but in this PR we are not aiming to bump xsimd

robomics:
@perseoGI I am bumping `xsimd` here and in the PRs linked in my other post to address the conflict mentioned in https://github.com/conan-io/conan-center-index/pull/24382#pullrequestreview-2127713015.

perseoGI (approved):
LGTM

recipes/arrow/all/patches/16.0.0-0001-fix-cmake.patch
```diff
@@ -0,0 +1,62 @@
...
+diff --git a/cpp/src/parquet/CMakeLists.txt b/cpp/src/parquet/CMakeLists.txt
+index 93f2e72d8..e00f73f7d 100644
+--- a/cpp/src/parquet/CMakeLists.txt
++++ b/cpp/src/parquet/CMakeLists.txt
+@@ -262,11 +262,11 @@ if(NOT PARQUET_MINIMAL_DEPENDENCY)
+
+   # These are libraries that we will link privately with parquet_shared (as they
+   # do not need to be linked transitively by other linkers)
+-  list(APPEND PARQUET_SHARED_PRIVATE_LINK_LIBS thrift::thrift)
++  list(APPEND PARQUET_SHARED_PRIVATE_LINK_LIBS Boost::headers thrift::thrift)
```
perseoGI, line 54: Hi again, why do we need to link against `Boost::headers`?
Does boost need to be enabled if we want to use `thrift`?
In that case, we should add that check in the verify method!

robomics: Good catch!

robomics:
@perseoGI CI is green! :)

recipes/arrow/all/conanfile.py
```diff
@@ -261,6 +264,11 @@ def validate(self):
             if self.dependencies["jemalloc"].options.enable_cxx:
                 raise ConanInvalidConfiguration("jemmalloc.enable_cxx of a static jemalloc must be disabled")
 
+        if self.options.with_thrift and not self.options.with_zlib:
```
AbrilRBS, line 267: There are more checks in `ThirdpartyToolchain.cmake` that might be worth to reflect in the recipe, instead of relying on autodiscovery

recipes/arrow/all/conanfile.py
```diff
@@ -101,7 +101,7 @@ class ArrowConan(ConanFile):
         "with_glog": False,
         "with_grpc": False,
         "with_json": False,
-        "with_thrift": False,
+        "with_thrift": True,
```
AbrilRBS, line 104: Is there a good reason to have this set to true? Is this _just_ for testing the changes n the CI? We could then revert it now

recipes/arrow/all/conanfile.py
```diff
@@ -351,6 +360,7 @@ def generate(self):
         tc.variables["ARROW_WITH_THRIFT"] = bool(self.options.with_thrift)
         tc.variables["Thrift_SOURCE"] = "SYSTEM"
         if self.options.with_thrift:
+            tc.variables["ARROW_THRIFT"] = True
```
AbrilRBS, line 363: As far as I can see, this variable was not set, but maybe we would need to set `ARROW_WITH_THRIFT` instead? Am I missing something? Do you have any insight? Thanks!

AbrilRBS (approved):
Some comments from a review alongside @perseoGI

AbrilRBS (changes requested):
Didn't mean to approve just yet!

recipes/arrow/all/conanfile.py
robomics: Yes! `with_thrift` needs to be True in order to build Arrow with Parquet support (which I have enabled based on https://github.com/conan-io/conan-center-index/pull/24044#issuecomment-2121953837).

recipes/arrow/all/conanfile.py
robomics: I am not super familiar with Arrow's code base, so I am not 100% sure on what's the difference between `ARROW_WITH_THRIFT` and `ARROW_THRIFT`.

A quick search on Arrow's codebase suggests that `ARROW_THRIFT` is only referenced in

```console
$ find arrow -type f -exec grep -P "\bARROW_THRIFT\b" {} +

arrow/cpp/cmake_modules/ThirdpartyToolchain.cmake:if(ARROW_THRIFT)
arrow/python/build/lib.linux-x86_64-cpython-312/cmake_modules/ThirdpartyToolchain.cmake:if(ARROW_THRIFT)
```

Where it seems to be used only to turn on ZLIB support  - [link](https://github.com/apache/arrow/blob/817ccbcb05e958d67927a998cc186bfed0d8528f/cpp/cmake_modules/ThirdpartyToolchain.cmake#L389-L392)

```cmake
if(ARROW_THRIFT)
  set(ARROW_WITH_ZLIB ON)
endif()
```

So I think setting both `ARROW_WITH_THRIFT` and `ARROW_THRIFT` is ok (will push a commit to do so ASAP).

recipes/arrow/all/conanfile.py
robomics: That file is over 5000 lines long :D. I do not have the bandwidth to read through it to see what's going on.
If you have some specific checks in mind?

Otherwise I suggest we leave this to be addressed in another PR.

robomics:
Things are looking good now.
Can you please have a look @AbrilRBS, @perseoGI?

recipes/arrow/all/conanfile.py
AbrilRBS: This is the only one that stood out to me, so I'm happy with keeping that on a follow-up PR, yes :)

recipes/arrow/all/conanfile.py
AbrilRBS: Thanks!

AbrilRBS (approved):
This now looks good on my end, will ask @perseoGI to re-review for tomorrow so we can finally get this merged, thanks a lot for your patience @robomics and for taking the time to address our reviews, we appreciate it :)

robomics:
Awesome!
Thank you both for your comments and help throughout the review :)

PR #24046, opened 2024-05-20
thrift: add thrift v0.20.0 and bump deps

robomics:
Specify library name and version:  **thrift/all**

---

### DIFF
--- a/recipes/thrift/all/conandata.yml
+++ b/recipes/thrift/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.20.0":
+    url: "http://archive.apache.org/dist/thrift/0.20.0/thrift-0.20.0.tar.gz"
+    sha256: "b5d8311a779470e1502c027f428a1db542f5c051c8e1280ccd2163fa935ff2d6"
   "0.18.1":
     url: "http://archive.apache.org/dist/thrift/0.18.1/thrift-0.18.1.tar.gz"
     sha256: "04c6f10e5d788ca78e13ee2ef0d2152c7b070c0af55483d6b942e29cff296726"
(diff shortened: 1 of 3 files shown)


Croydon (approved):

CJCombrink (approved):

franramirez688 (approved):

CJCombrink:
@robomics Thanks for the effort, you saved me from doing it :D

CJCombrink:
@franramirez688 Thanks for the approval and merge

jcar87:
pleaase be aware that version bumping like this introduces a conflict in opentelemetry-cpp.

I would advise against bumping versions unnecessarily unless there is a clear reason - for example, a new feature in boost that is required.

CJCombrink:
@jcar87 I am very curious why adding a new thrift version have an affect on another package?
personally I think the packages should be independent of each other, except perhaps those that use ranges that the conan peaple have deemed can be ranged, like zlib and openssl when I last looked....

> clear reason 

0.20 is newer than 0.18.1, should be reason enough for a package manager

ps: thrift is pinned in [opentelemetry-cpp](https://github.com/conan-io/conan-center-index/blob/master/recipes/opentelemetry-cpp/all/conanfile.py#L139), that am I missing?
```
self.requires("thrift/0.17.0")
```

CJCombrink:
I [quickly looked at opentelemetry-cpp](https://github.com/search?q=repo%3Aopen-telemetry%2Fopentelemetry-cpp%20boost&type=code) and it seems like it does not require boost, thrift requires boost. So perhaps the correct thing is to remove boost from the opentelemetry-cpp recipe? 

(I am not familiar with opentelemetry-cpp)

CJCombrink:
Ok finally it also seems that opentelemetry-cpp dropped the thrift requirement in version [[1.10.0] 2023-07-11](https://github.com/open-telemetry/opentelemetry-cpp/blob/main/CHANGELOG.md#1100-2023-07-11)
> [REMOVAL] Remove the jaeger exporter https://github.com/open-telemetry/opentelemetry-cpp/pull/2031

So no need to worry about breaking it from the thrift side any more I guess

PR #24054, opened 2024-05-21
glib: refactor test package

ErniGH:

* Write the minimal test that tests the package (the compiler can find the correct includes, the static linker can resolve
* Do not perform heavy logic in CI.
* Remove binary blob - these have not been allowed anywhere in the repository for a while now, it's time to drop them where they exist


recipes/glib/all/test_package/conanfile.py
```diff
@@ -16,33 +15,6 @@ def layout(self):
...
-        virtual_run_env = VirtualRunEnv(self)
-        virtual_run_env.generate()
-
-        if self.settings.os == "Macos":
-            env = Environment()
-            # Avoid conflicts with system libiconv
-            # see: https://github.com/conan-io/conan-center-index/pull/17610#issuecomment-1552921286
-            env.define_path("DYLD_FALLBACK_LIBRARY_PATH", "$DYLD_LIBRARY_PATH")
-            env.define_path("DYLD_LIBRARY_PATH", "")
-            env.vars(self, scope="run").save_script("conanrun_macos_runtimepath")
```
ericLemanissier, line 35: it looks like this is still needed, cf https://github.com/ericLemanissier/proof-of-conan/actions/runs/9177037237/job/25233751761#step:8:2039

ericLemanissier (commented):
the PR description could be reworded to match the actual change: Remove pkg-config from the test, and use only CMake on all platforms
Because apart from that, the test is not more minimal (same test_package.c, and same executable is run), there is no suppression of either heavy logic or binary blob

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/9177037237/attempts/2)

uilianries (approved):
LGTM

PR #24059, opened 2024-05-21
glib: Depend on elfutils instead of libelf where possible

jwillikers:
The libelf library is antiquated.
The elfutils package is actively maintained and provides its own libelf implementation, which is the standard implementation on Linux. Since elfutils can not be used everywhere, only use it for supported configurations.

Requires:

- [x] #24061


ericLemanissier (approved):

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/9192178992/attempts/1)

ericLemanissier:
The build is now finished. You can take a look at the results [here](https://github.com/ericLemanissier/proof-of-conan/actions/runs/9192178992/attempts/2)

Issue #24058, opened 2024-05-21
[package] elf-utils/0.21: Fails to build with GCC 14


jwillikers:
### Description

A missing include causes an implicit function declaration in `elf-utils/0.21`. This is not an issue in version `0.22.5` which is packaged in CCI already, however version `0.21` is still used for `elfutils`.

```
/home/jordan/.conan2/p/b/gette3f107772e0f20/b/src/gettext-tools/src/locating-rule.c:150:7: error: implicit declaration of function ‘free’ [-Wimplicit-function-declaration]
  150 |       free (reduced);
      |       ^~~~
/home/jordan/.conan2/p/b/gette3f107772e0f20/b/src/gettext-tools/src/locating-rule.c:50:1: note: include ‘<stdlib.h>’ or provide a declaration of ‘free’
   49 | #include "xalloc.h"
  +++ |+#include <stdlib.h>
   50 |
```

### Package and Environment Details

* Package Name/Version: **elfutils/0.21**
* Operating System+version: **Fedora Linux 40**
* Compiler+version: **GCC 14.1.1**
* Conan version: **conan 2.3.1**
* Python version: **Python 3.12.3**

### Conan profile

Profile host:
[settings]
arch=x86_64
build_type=Debug
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux
os.libc=gnu
os.libc.version=2.39
mold/*:compiler.libcxx=libstdc++11
mesa/*:compiler.cppstd=17
[options]
wayland/*:shared=True
[conf]
tools.cmake.cmake_layout:build_folder_vars=['settings.arch', 'settings.compiler', 'options.platform']
tools.cmake.cmaketoolchain:find_package_prefer_config=True
tools.cmake.cmaketoolchain:generator=Ninja
tools.cmake.cmaketoolchain:user_toolchain=['/var/home/jordan/.conan2/toolchains/mold.cmake']
tools.env.virtualenv:powershell=False
tools.system.package_manager:mode=install
tools.system.package_manager:sudo=True
[buildenv]
LD=gcc
CC_LD=mold
CXX_LD=mold
avahi/*:LD=!
gdbm/*:LD=!
libalsa/*:LD=!
libffi/*:LD=!
libiconv/*:LD=!
libmount/*:LD=!
libtool/*:LD=!
libva/*:CC_LD=!
flex/*:LD=!
hwloc/*:LD=!
kmod/*:LD=!
[...]

PR #24060, opened 2024-05-21
elfutils: Bump gettext to version 0.22.5 to support GCC 14

jwillikers:
See issue #24058.


jcar87 (approved):
Should result in no conflicts (especially as this is a tool required)

PR #24061, opened 2024-05-21
elfutils: Correctly set pkg-config names for libraries

jwillikers:
The pkg-config names should be set as provided by the project. See https://sourceware.org/git/?p=elfutils.git;a=tree;f=config;hb=HEAD.


ericLemanissier (approved):

RubenRBS (approved):
Great catch, thanks a lot!

PR #24062, opened 2024-05-21
tcpcat library publish

ydcpp:
Specify library name and version:  **ydcpp-tcpcat/1.0.4**

Hi,

I'm the author of tcpcat library. tcpcat is a simple and easy to use C++ TCP Client and Server library.
Just released the project and wanted to publish package also in conan center.
Thanks in advance!

---


recipes/tcpcat/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+add_executable(test_package test_package.cpp)
+
+find_package(tcpcat CONFIG REQUIRED)
+target_link_libraries(test_package tcpcat::tcpcat)
```
RubenRBS, line 7: In https://github.com/ydcpp/tcpcat/blob/main/test/CMakeLists.txt#L26 you use `tcpcat` as the target name without the namespaces. What's your expected name?
If you prefer to have it as shown in the tests, we can acomplish that by adding a `self.cpp_info.set_property("cmake_target_name", "tcpcat")` line in the conanfile's `package_info()` method

recipes/tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        copy(self, "*", src=os.path.join(self.source_folder, "include"), dst=os.path.join(self.package_folder, "include"))
+        cmake = CMake(self)
+        cmake.install()
```
RubenRBS, line 77: It would be great if your upstream CMakeLists would properly install the headers as needed instead of relying on Conan's copy, just so it's easier for people just cloning your library to get set up :)

recipes/tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        copy(self, "*", src=os.path.join(self.source_folder, "include"), dst=os.path.join(self.package_folder, "include"))
```
RubenRBS, line 75: I see you have asio as a vendorized dependency
I know it's a somewhat large refactor, but if you could add an option to instead rely on a `find_package()` call for asio, it would ensure Conan users could override their asio dependency if needed/make it work with the current setup

As it stands right now there's little risk of odr violations, but better be safe than sorry :)

recipes/tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
```
RubenRBS, line 55: I've added some missing checks for Conan 1 - This can be dropped in the future but for now it's better to also have them

recipes/tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "gcc": "7",
+            "Visual Studio": "16",
+            "msvc": "192",
+            "clang": "7",
+            "apple-clang": "12"
+        }
```
RubenRBS, line 38: If you have any specific compiler limits, feel free to update this, I just used some common versions

recipes/tcpcat/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,11 @@
+#include <tcpcat/tcpcat.h>
+
+#include <iostream>
+#include <string>
+
+int main()
+{
+   auto session = tcpcat::TcpSession(nullptr, nullptr, 0);
```
RubenRBS, line 8: I've simplified the test package to just create a tcpsession which does not instantiate any socket, we try to stay clear from the network layer in the tests, and this seems to work ok :)

RubenRBS (commented):
Hi @ydcpp thanks a lot for taking the time to submit your recipe, we appreciate it :)

I've taken the liberty to update the recipe with some improvements, and I've also left some comments - happy to hear your insight on those when you have a chance!

recipes/tcpcat/all/test_package/CMakeLists.txt
ydcpp: thank you for the suggestion, yes we can add this to conanfile's `package_info()` method

recipes/tcpcat/all/conanfile.py
ydcpp: you're right, I will update upstream and make 1.0.1 release and will update the recipe for version here too.

recipes/tcpcat/all/conanfile.py
ydcpp: thank you very much, I'm new to publishing conan packages in conan-center so its great to have your assistance on these matters!

recipes/tcpcat/all/conanfile.py
ydcpp: any compiler version that supports std::optional in c++17 and above is alright to build tcpcat.

recipes/tcpcat/all/test_package/test_package.cpp
ydcpp: understandable, thank you!

recipes/tcpcat/all/conanfile.py
ydcpp: You're right, I will refactor upstream.

recipes/tcpcat/all/conanfile.py
RubenRBS: Great, thanks a lot! Feel free to ping me once the new version is in :)

ydcpp: I have released 1.0.1 version in upstream and fixed the issues with ASIO dependency and install() function in CMake. Also pushed a commit in this PR to use 1.0.1 version instead.
@RubenRBS please check again

ydcpp: Update: updated as 1.0.2

recipes/tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+class TcpcatConan(ConanFile):
+    name = "tcpcat"
+    package_type = "library"
+
+    # Optional metadata
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/ydcpp/tcpcat"
+    description = "Simple C++ TCP Server and Client library."
+    topics = ("network", "tcp", "tcp server", "tcp client")
```
valgur, line 19:
```suggestion
    topics = ("network", "tcp", "tcp-server", "tcp-client")
```
Not 100% sure but I think spaces in topics are not allowed.

recipes/tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,86 @@
...
+            check_min_cppstd(self, self._min_cppstd)
+
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def requirements(self):
+        self.requires("asio/[^1.30]", transitive_headers = True)
```
valgur, line 62:
```suggestion
        self.requires("asio/1.30.2", transitive_headers = True)
```
Version ranges are currently allowed for only a small number of dependencies on CCI. https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges

recipes/tcpcat/all/conanfile.py
ydcpp: updated, thanks.

recipes/tcpcat/all/conanfile.py
ydcpp: updated, thanks.

recipes/tcpcat/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,11 @@
+#include <tcpcat/tcpcat.h>
+
+#include <iostream>
+#include <string>
+
+int main()
+{
+   tcpcat::TcpSession session(nullptr, nullptr, 0);
```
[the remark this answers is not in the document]
RubenRBS: In fact, nothing is created when passing null pointers, this basically just executes the empty constructor and then calls a method that does a trivial attribute get :)

RubenRBS (commented):
Hi @ydcpp thanks a lot for updating the recipe :)
I have some points I'd like to mention after taking a look into the recipe and your upstream changes:

- You library's name has the unfortunate name collision with another older library, which as far as I can see comes in FreeBSD and is available in other Unix systems such as Ubuntu. Our general policy in these cases is to set the `name` attribute to `name = "ydcpp-tcpcat"` to avoid future confussions
- Instead of using `set(CMAKE_CXX_STANDARD 17)` it would be best if you used `target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)`, which tells CMake to ask for at least C++17 general support, but does not force C++17 specifically. This is helpful because in Conan users can set their desired cppstd flag and it's best if all packages listen to it, instead of silently changing the user's decisions (If you feel it's needed for your use-case upstream, we can always patch that line out, no problem)
- We'll need to patch out the find_program for `cppcheck` here, else you can add an option to enable/disable it and we can make sure to set it to False in the `configure` method
- Lastly, a super minor thing: When you tagged 1.0.3, you forgot to update the version in your CMakeLists 

Regarding the Windows failures, I'll submit a patch in a sec to fix it

Let me know if you need any help with any of the above, happy to help!

recipes/tcpcat/all/test_package/test_package.cpp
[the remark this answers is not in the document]
uilianries: Thank you!

ydcpp:
> Hi @ydcpp thanks a lot for updating the recipe :) I have some points I'd like to mention after taking a look into the recipe and your upstream changes:
> 
>     * You library's name has the unfortunate name collision with another older library, which as far as I can see comes in FreeBSD and is available in other Unix systems such as Ubuntu. Our general policy in these cases is to set the `name` attribute to `name = "ydcpp-tcpcat"` to avoid future confussions
> 
>     * Instead of using `set(CMAKE_CXX_STANDARD 17)` it would be best if you used `target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)`, which tells CMake to ask for at least C++17 general support, but does not force C++17 specifically. This is helpful because in Conan users can set their desired cppstd flag and it's best if all packages listen to it, instead of silently changing the user's decisions (If you feel it's needed for your use-case upstream, we can always patch that line out, no problem)
> 
>     * We'll need to patch out the find_program for `cppcheck` here, else you can add an option to enable/disable it and we can make sure to set it to False in the `configure` method
> 
>     * Lastly, a super minor thing: When you tagged 1.0.3, you forgot to update the version in your CMakeLists
> 
> 
> Regarding the Windows failures, I'll submit a patch in a sec to fix it
> 
> Let me know if you need any help with any of the above, happy to help!

thank you for the suggestions, I'll update upstream and let you know.

ydcpp:
@RubenRBS I updated upstream and library name, please check again. thanks in advance!

RubenRBS:
@ydcpp oh my, I missed the ping from 20 days ago, I'm so sorry for the delay, will get to this tomorrow first thing in the morning :)

RubenRBS:
Alas, I needed @uilianries to double check for me. The remaining failing in Conan 2 in Windows comes from the fact that no symbols are exported. Do you plan on supporting windows shared builds, or should we validate it out?

recipes/ydcpp-tcpcat/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+add_executable(test_package test_package.cpp)
+
+find_package(ydcpp-tcpcat CONFIG REQUIRED)
+target_link_libraries(test_package ydcpp-tcpcat)
```
uilianries, line 7:
```suggestion
target_link_libraries(test_package ydcpp-tcpcat)
target_compile_features(test_package PRIVATE cxx_std_17)
```

Need to pass C++17 standard to the compiler, because in Conan 1.x compiler.cppstd is not used by default.

RubenRBS:
Hi @ydcpp sorry for the delay: We have now pushed the last fixes needed to get the recipe working. Note that as I explained in https://github.com/conan-io/conan-center-index/pull/24062#issuecomment-2175741857, the current Windows shared builds have been disabled, as the current issue that no symbols are exported from your side in the library - let us know if you need any help with that, we can probably provide some hints as to how to better handle that on your side if needed :)

Other than that, we can merge the recipe after the current build succeeds, will keep an eye on it to ensure it gets thru to the finish line quickly. Again, thanks for your patience and follow-ups in this process

RubenRBS (dismissed):

recipes/ydcpp-tcpcat/all/conanfile.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
+        if not self.settings.get_safe("compiler.cppstd"):
+            tc.variables["CMAKE_CXX_STANDARD"] = self._min_cppstd
```
uilianries, line 79: I'm missing something, because the project already does it: https://github.com/ydcpp/tcpcat/blob/1.0.4/CMakeLists.txt#L16

RubenRBS: Oh I actually didn't catch that - let's just try with your test package changes, see if that fixes everything :)

RubenRBS (approved):

uilianries (approved):
LGTM, thank you @ydcpp !!!

Issue #23649, opened 2024-04-19
Consider using libglvnd from Conan instead of the system's via opengl/system


jcar87:
### What is your question?

Placeholder

See:
https://github.com/conan-io/conan-center-index/pull/23484
https://github.com/conan-io/conan-center-index/pull/23465
https://github.com/conan-io/conan-center-index/pull/23464
https://github.com/conan-io/conan-center-index/pull/23463
https://github.com/conan-io/conan-center-index/pull/23455
https://github.com/conan-io/conan-center-index/pull/23452
https://github.com/conan-io/conan-center-index/pull/23451
https://github.com/conan-io/conan-center-index/pull/23450
https://github.com/conan-io/conan-center-index/pull/23449
https://github.com/conan-io/conan-center-index/pull/23448
https://github.com/conan-io/conan-center-index/pull/23447
https://github.com/conan-io/conan-center-index/pull/23446
https://github.com/conan-io/conan-center-index/pull/23443
https://github.com/conan-io/conan-center-index/pull/23441
https://github.com/conan-io/conan-center-index/pull/23439
https://github.com/conan-io/conan-center-index/pull/23280
https://github.com/conan-io/conan-center-index/pull/23279
https://github.com/conan-io/conan-center-index/pull/23222
https://github.com/conan-io/conan-center-index/pull/22958

jwillikers:
@jcar87 You'll need to address the fact that `vulkan-loader` offers the same loading functionality for Vulkan graphics drivers as `libglvnd` does for GL drivers. Also, you'll want to address Wayland as well, which is much more important in my scenario than X11.

Also, the fact of the matter is, my company builds graphical applications, but what your saying sounds like go somewhere else if that's the case as that's not something CCI can handle. From your points, we shouldn't have a Qt Conan package that offers graphical components because we can't test them in CCI.

Fyi, vcpkg, HomeBrew, and Spack all offer Mesa packages.

jcar87:
> @jcar87 You'll need to address the fact that vulkan-loader offers the same loading functionality for Vulkan graphics drivers as libglvnd does for GL drivers. Also, you'll want to address Wayland as well, which is much more important in my scenario than X11.

[...]

PR #24067, opened 2024-05-22
mesa-glu: Add an option to choose the glvnd implementation

jwillikers:
Refer to PR #24033 and issue #23649. Default to the system implementation for now since that is consistent across most of the existing Conan packages.


maksim-petukhou:
@jcar87 is this PR expected to be closed?

Issue #24019, opened 2024-05-17
openssl: cannot find perl


TheClonerx:
despite openssl requiring perl in order to build, it is not specified in the recipe's tool_requirements.
while some distribution (such as debian & archlinux) do provide perl while installing cmake, gcc, and make, others such as fedora don't.
the docs mention that only cmake is assumed to be installed: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#build-requirements

to reproduce:
```sh
podman run --rm -it fedora
dnf install gcc g++ cmake conan
conan profile detect
# might need to add gcc 14 to ~/.conan2/settings.yml
conan install --requires "openssl/[*]" -b missing

# ...

openssl/3.2.1: Generating aggregated env files
openssl/3.2.1: Generated aggregated env files: ['conanbuild.sh', 'conanrun.sh']
openssl/3.2.1: Calling build()
openssl/3.2.1: apply_conandata_patches(): No patches defined in conandata
openssl/3.2.1: RUN: perl ./Configure "conan-Release-Linux-x86_64-gcc-14" no-shared --prefix=/ --libdir=lib --openssldir="/etc/ssl" no-unit-test threads PERL=perl no-tests --release -fPIC enable-fips no-md2 zlib --with-zlib-include="/root/.conan2/p/b/zlibaac6682b97598/p/include" --with-zlib-lib="/root/.conan2/p/b/zlibaac6682b97598/p/lib"
/bin/sh: line 1: perl: command not found

openssl/3.2.1: ERROR:
Package '2ead4b5224b22aa6d964443cf39c8659bea28efd' build failed
openssl/3.2.1: WARN: Build folder /root/.conan2/p/b/opens2a80818cf07f4/b/build-release
ERROR: openssl/3.2.1: Error in build() method, line 533
        self._make()
while calling '_make', line 520
        self.run("{perl} ./Configure {args}".format(perl=self._perl, args=args), env="conanbuild")
        ConanException: Error 127 while executing
```
--- 
of course could be easily solved by just installing perl through the system package manager, but in some situations one does not have root access to do such thing.

SpaceIm:
Indeed, but there is no `perl` recipe in conan-center (only `strawberryperl` recipe for Windows). One could create such recipe. Source code of 5.38.2: https://www.cpan.org/src/5.0/perl-5.38.2.tar.gz

PR #24078, opened 2024-05-23
perl: add recipe

### LINKED ISSUE #12257: [request] add perl/5.36@ recipe
### Package Details
  * Package Name/Version: **perl/5.36**
  * Website: **http://www.cpan.org/**
  * Source code: **http://www.cpan.org/src/**

Perl is a highly capable, feature-rich programming language with over 30 years of development. It could be used a build tool

### LINKED ISSUE #24027: [request] perl/5.38.2
### Package Name/Version

perl/5.38.2

### Webpage

https://www.perl.org

### Source code

https://www.cpan.org/src/5.0/perl-5.38.2.tar.gz

Perl is a general-purpose programming language originally developed for text manipulation and now used for a wide range of tasks including system administration, web development, network programming, GUI development, and more.

Ahajha:
Specify library name and version:  **perl/5.38.2**

Resolves #24027
Resolves #12257

---

### DIFF
--- a/recipes/perl/all/conanfile.py
+++ b/recipes/perl/all/conanfile.py
@@ -1,8 +1,8 @@
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
-from conan.tools.files import chdir, copy, get, rmdir
+from conan.tools.files import chdir, copy, get, replace_in_file, rmdir
 from conan.tools.gnu import Autotools, AutotoolsToolchain, AutotoolsDeps
 from conan.tools.layout import basic_layout
+from conan.tools.microsoft import msvs_toolset
 import os
 
 
@@ -18,6 +18,10 @@ class PerlConan(ConanFile):
     package_type = "application"
     settings = "os", "arch", "compiler", "build_type"
 
+    @property
+    def _settings_build(self):
+        return getattr(self, "settings_build", self.settings)
+
     def configure(self):
         self.settings.rm_safe("compiler.cppstd")
         self.settings.rm_safe("compiler.libcxx")
@@ -25,17 +29,18 @@ class PerlConan(ConanFile):
     def layout(self):
         basic_layout(self, src_folder="src")
 
-    def validate(self):
-        if self.settings.os == "Windows":
-            raise ConanInvalidConfiguration(f"{self.ref} is not supported on {self.settings.os}. "
-                                            "You may want to use strawberryperl instead.")
-
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def requirements(self):
         self.requires("bzip2/1.0.8")
         self.requires("zlib/[>=1.2.11 <2]")
+
+    def build_requirements(self):
+        if self._settings_build.os == "Windows":
+            self.win_bash = True
+            if not self.conf.get("tools.microsoft.bash:path", check_type=str):
+                self.tool_requires("msys2/cci.latest")
 
     def generate(self):
         tc = AutotoolsToolchain(self)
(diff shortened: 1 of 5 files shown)


recipes/perl/all/conanfile.py
```diff
@@ -0,0 +1,93 @@
...
+        tc = AutotoolsToolchain(self)
+        tc.generate()
+
+        deps = AutotoolsDeps(self)
+        deps.generate()
+    
+    @property
+    def _compiler(self):
+        compilers = self.conf.get("tools.build:compiler_executables", default={},
+                                                     check_type=dict)
```
valgur, line 50:
```suggestion
        compilers = self.conf.get("tools.build:compiler_executables", default={}, check_type=dict)
```
The indentation is a bit weird here.

recipes/perl/all/conanfile.py
```diff
@@ -0,0 +1,93 @@
...
+            return compilers["c"]
+        
+        # Otherwise it will use the system "cc" command - assuming the profile reflects this.
+        return None
+
+    def build(self):
+        # Can't use configure() since Perl uses "Configure" instead of "configure". Renaming the file does not seem to work.
+        with chdir(self, self.source_folder):
+            config_path = os.path.join(self.source_folder, 'Configure')
+            command = f"{config_path} -de -Dprefix={self.package_folder}"
```
valgur, line 61:
```suggestion
            command = f"Configure -de -Dprefix={self.package_folder}"
```
You have already chdir-ed inside the source folder.

recipes/perl/all/conanfile.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def package_info(self):
+        self.cpp_info.frameworkdirs = []
+        self.cpp_info.libdirs = []
+        self.cpp_info.resdirs = []
+        self.cpp_info.includedirs = []
+
+        if self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs = ["m", "rt"]
+
```
valgur, line 85:
```suggestion
```
There are no libraries to be linked, so the system_libs are unnecessary.

valgur (changes requested):
The system_libs should be dropped as they are unnecessary. Otherwise LGTM. Thanks!

recipes/perl/all/conanfile.py
```diff
@@ -0,0 +1,93 @@
...
+    def configure(self):
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def validate(self):
+        if self.settings.os == "Windows":
```
valgur, line 29: The recipe might work with MinGW. Maybe restrict it for MSVC only.

recipes/perl/all/conanfile.py
Ahajha: There are a few libraries in the package, I presume C extensions that link to the system libraries. The hooks complain, so this is mainly to satisfy that.

recipes/perl/all/conanfile.py
Ahajha: I sort of just assumed it didn't work, but I didn't actually try. I'll give it a shot and see how it goes.

Ahajha: Still working on this on the side - Seems they at least officially support MinGW which is neat. No surprise though that the Windows build of a Makefile-based project is difficult to work with.

jcar87:
Hi @Ahajha  - thanks for this PR, and apologies for the delay in getting back to this.

We are currently reviewing PRs that introduce new recipes and assessing the suitability for us to maintain and support and have come to the conclusion that it is currently not worth it to maintain a recipe for perl. Conan Center's goal has always been to focus on libraries, and in some specific instances standalone and well contained executables (e.g. ninja). Perl seems to be more aligned with it being part of a OS provided toolchain. 

For https://github.com/conan-io/conan-center-index/issues/24019 - perl is not only required by openssl, but also by any recipe that does `autoreconf` - as some of the autoconf scripts are actually perl scripts. So practical evidence would indicate that it is already present in most developer's systems where it is needed. On macOS it comes as standard (if xcode is installed, perl should also work), and on Linux we've found that it is typically installed as part of `build-essential` or equivalent. We have since updated our docs to reflect what we expect a typical system to have installed to work with Conan recipes: https://github.com/conan-io/conan-center-index/blob/master/docs/consuming_recipes.md#expected-environment - which hopefully clarifies this. The only instances we have seen where developers may not have complete build environments is those on super minimal Docker images. 

OpenSSL also requires specific perl modules that may not be part of minimal installation, so I'm not sure whether those are covered. 

if the situation changes or our understanding is wrong, we may revisit this in the future. Apologies if this is not the outcome you hoped for, and thank you for the time you have dedicated!

PR #24092, opened 2024-05-24
fix geograhiclib when tools option is False

### LINKED ISSUE #24053: Unable to use geographiclib/2.3
### Describe the bug

Unable to use geographiclib/2.3 in my project when "geographiclib:tools": False in default_option.

### How to reproduce it

```
default_options = {
    "geographiclib:tools": False,
}
```

>> conan create . 2.3@geographiclib/testing -pr x64d

zeeshancs07:
**geographiclib/2.3**
fixes #24053 
Were using geographiclib version 1.52 with the option "geographiclib:tools": False using conan version 1.x. Started getting errors when upgrading this lib to version 2.3

---


SpaceIm:
I'm pretty sure that this pattern is different in other versions.

I think a patch adding BUILD_TOOLS option in each version would be better, and it could be submitted upstream to avoid to maintain this patch forever.

Something like this for example (for version 2.3):

```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -452,9 +452,12 @@ if (WIN32)
 endif ()
 
 # The list of tools (to be installed into, e.g., /usr/local/bin)
+option (BUILD_TOOLS ON)
+if (BUILD_TOOLS)
 set (TOOLS CartConvert ConicProj GeodesicProj GeoConvert GeodSolve
   GeoidEval Gravity IntersectTool MagneticField Planimeter RhumbSolve
   TransverseMercatorProj)
+endif ()
 # The list of scripts (to be installed into, e.g., /usr/local/sbin)
 set (SCRIPTS geographiclib-get-geoids geographiclib-get-gravity
   geographiclib-get-magnetic)
@@ -464,7 +467,9 @@ set_property (GLOBAL PROPERTY USE_FOLDERS ON)
 # The list of subdirectories to process
 add_subdirectory (src)
 add_subdirectory (include/GeographicLib)
+if (BUILD_TOOLS)
 add_subdirectory (tools)
+endif ()
 add_subdirectory (man)
 add_subdirectory (doc)
 if (EXAMPLEDIR)
```

uilianries:
I just sent a PR to the upstream: https://github.com/geographiclib/geographiclib/pull/27

zeeshancs07:
@uilianries : Please verify the changes pushed, as after your PR (https://github.com/geographiclib/geographiclib/pull/27) gets completed, then we don't need patch and  CMakeList modification through conanfile.py.

recipes/geographiclib/all/conanfile.py
```diff
@@ -95,6 +95,7 @@ def generate(self):
         tc = CMakeToolchain(self)
         tc.variables["GEOGRAPHICLIB_LIB_TYPE"] = "SHARED" if self.options.shared else "STATIC"
         tc.variables["GEOGRAPHICLIB_PRECISION"] = self._cmake_option_precision
+        tc.variables["BUILD_TOOLS"] = self.options.tools
```
SpaceIm, line 98: Ok but you have to patch CMakeLists, otherwise it doesn't make sense.

uilianries:
@zeeshancs07 After speaking to the author, it will be better skipping the installation step only, probably will not have official support for skipping building tools. It was added in the very past, but was never supported by the upstream. I'll open a new PR to CCI changing it.

zeeshancs07:
handled tools building and packaging in https://github.com/conan-io/conan-center-index/pull/24121

PR #24103, opened 2024-05-26
add Qt 5.15.14

kambala-decapitator:
Specify library name and version:  **qt/5.15.14**

- adds Qt 5.15.14
- removes Android backtrace patch as the fix was upstreamed: https://code.qt.io/cgit/qt/qtbase.git/commit/?h=5.15&id=1f154eec608db6caa027db534856f40d2c7ccf87

---


ericLemanissier (approved):
Thanks!

ericLemanissier:
There are true positives, but it's not trivial to fix because it can change at each version of qt, and it requires diving into qt's `.pro` files. It'd be great if someone fixed them, but it should not block any PR to qt, because it somehow does not prevent consuming qt. These system libs must be indirectly pulled by other parts of the program.

RubenRBS (approved):

PR #24105, opened 2024-05-26
dfp: add version 1.0.3

toge:
Specify library name and version:  **dfp/1.0.3**

https://github.com/epam/DFP/compare/1.0.2...1.0.3

---

### DIFF
--- a/recipes/dfp/all/conandata.yml
+++ b/recipes/dfp/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.0.3":
+    url: "https://github.com/epam/DFP/releases/download/1.0.3/native-sources.tar.gz"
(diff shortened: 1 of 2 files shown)


RubenRBS:
It scares nme a bit that there's a 1.0.2 and 1.0.2 cxx but not for 1.0.3? I think we should double check with @agdavydov81 what the expected steps are :)

agdavydov81:
> It scares nme a bit that there's a 1.0.2 and 1.0.2 cxx but not for 1.0.3? I think we should double check with @agdavydov81 what the expected steps are :)

This is because version 1.0.2 without cxx and conan was published earlier with appropriate tag in git.
So, the version 1.0.3 support java, .net and cxx and no need the special suffix.

RubenRBS:
Thanks a lot :)

RubenRBS (approved):

PR #24110, opened 2024-05-26
bredis: add recipe

toge:
Specify library name and version:  **bredis/***

---

### DIFF
--- a/recipes/bredis/all/test_package/test_package.cpp
+++ b/recipes/bredis/all/test_package/test_package.cpp
@@ -1,28 +1,12 @@
 #include <cstdlib>
 #include <iostream>
 
-#include "boost/asio.hpp"
-#include "bredis/MarkerHelpers.hpp"
 #include "bredis/Connection.hpp"
+#include "bredis/MarkerHelpers.hpp"
 
-namespace r = bredis;
-namespace asio = boost::asio;
-
-int main(int argc, const char** argv) {
-  using socket_t = asio::ip::tcp::socket;
-  using next_layer_t = socket_t;
-
-  asio::io_context io_service;
-
-  auto ip_address = asio::ip::make_address("127.0.0.1");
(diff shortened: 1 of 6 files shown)


recipes/bredis/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,30 @@
...
+  using socket_t = asio::ip::tcp::socket;
+  using next_layer_t = socket_t;
+
+  asio::io_context io_service;
+  auto             ip_address = asio::ip::make_address("127.0.0.1");
+  auto             port       = boost::lexical_cast<std::uint16_t>("6379");
+  std::cout << "connecting to " << ip_address << "\n";
+  asio::ip::tcp::endpoint end_point(ip_address, port);
+  socket_t                socket(io_service, end_point.protocol());
+  socket.connect(end_point);
```
RubenRBS, line 21: Let's avoid reaching the network layer for test packages if possible :)

RubenRBS (commented):
Thanks! Looks great, but I have a request in the test package

recipes/bredis/all/test_package/test_package.cpp
toge: @RubenRBS 
These lines are in the dummy function which has been never called.
Is it not acceptable?

RubenRBS: We need funtions whose implementation lives in a .cpp to ensure those work too, and they need to be called for the compiler not to simply remove dead code, else it might just ignore it even if it's wrong :)

toge: Fixed.

perseoGI (approved):
Good job Toge! Thank you 💯

RubenRBS (approved):

ErniGH (approved):
❤️

jcar87 (approved):

PR #24121, opened 2024-05-27
[geographiclib] Skip packaging tools when not requested

### LINKED ISSUE #24053: Unable to use geographiclib/2.3

uilianries:
Specify library name and version:  **geographiclib/2.3**

After checking with the upstream about adding a native option to build or not tools (PR https://github.com/geographiclib/geographiclib/pull/27), it will be more comfortable only skipping the package step instead. Those generated tools require only small amount of time to be built and so far does not break in the CI.

closes #24053
closes #24092

/cc @zeeshancs07

---

### DIFF
--- a/recipes/geographiclib/all/conanfile.py
+++ b/recipes/geographiclib/all/conanfile.py
@@ -7,6 +7,7 @@ from conan.tools.files import (
     replace_in_file, rm, rmdir
 )
 from conan.tools.scm import Version
+from conan.tools.microsoft import is_msvc
 import os
 
 required_conan_version = ">=1.53.0"
@@ -137,7 +138,9 @@ class GeographiclibConan(ConanFile):
         self.cpp_info.set_property("cmake_file_name", "geographiclib")
         self.cpp_info.set_property("cmake_target_name", "GeographicLib::GeographicLib")
         self.cpp_info.set_property("pkg_config_name", "geographiclib")
-        self.cpp_info.libs = ["GeographicLib"] if Version(self.version) >= "2" else ["Geographic"]
+        suffix = "_d" if self.settings.build_type == "Debug" and is_msvc(self) else ""
+        suffix += "-i" if is_msvc(self) else ""
+        self.cpp_info.libs = [f"GeographicLib{suffix}"] if Version(self.version) >= "2" else [f"Geographic{suffix}"]
         self.cpp_info.defines.append("GEOGRAPHICLIB_SHARED_LIB={}".format("1" if self.options.shared else "0"))
 
         # TODO: to remove in conan v2 once cmake_find_package_* generators removed
--- a/recipes/geographiclib/all/conanfile.py
+++ b/recipes/geographiclib/all/conanfile.py
@@ -128,15 +128,16 @@ class GeographiclibConan(ConanFile):
             rmdir(self, os.path.join(os.path.join(self.package_folder, folder)))
         rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
         if not self.options.tools:
-            bin_files = os.listdir(os.path.join(self.package_folder, "bin"))
-            for bin_file in [it for it in bin_files if not it.endswith(".dll")]:
-                rm(self, bin_file, os.path.join(self.package_folder, "bin"))
+            rmdir(self, os.path.join(self.package_folder, "sbin"))
+            bin_files = [it for it in os.listdir(os.path.join(self.package_folder, "bin")) if not it.endswith(".dll")]
+            for it in bin_files:
+                rm(self, it, os.path.join(self.package_folder, "bin"))
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "geographiclib")
         self.cpp_info.set_property("cmake_target_name", "GeographicLib::GeographicLib")
         self.cpp_info.set_property("pkg_config_name", "geographiclib")
-        self.cpp_info.libs = ["GeographicLib"]
+        self.cpp_info.libs = ["GeographicLib"] if Version(self.version) >= "2" else ["Geographic"]
         self.cpp_info.defines.append("GEOGRAPHICLIB_SHARED_LIB={}".format("1" if self.options.shared else "0"))
 
         # TODO: to remove in conan v2 once cmake_find_package_* generators removed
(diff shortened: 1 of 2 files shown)


RubenRBS (dismissed):
Nice :)

recipes/geographiclib/all/conanfile.py
```diff
@@ -132,12 +127,17 @@ def package(self):
...
+            bin_files = [it for it in os.listdir(os.path.join(self.package_folder, "bin")) if not it.endswith(".dll")]
+            for it in bin_files:
+                rm(self, it, os.path.join(self.package_folder, "bin"))
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "geographiclib")
         self.cpp_info.set_property("cmake_target_name", "GeographicLib::GeographicLib")
         self.cpp_info.set_property("pkg_config_name", "geographiclib")
-        self.cpp_info.libs = collect_libs(self)
+        self.cpp_info.libs = ["GeographicLib"] if Version(self.version) >= "2" else ["Geographic"]
```
zeeshancs07, line 140: @uilianries : With conan 1.64.0 getting error 
`-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19045.
-- Conan: Target declared 'GeographicLib::GeographicLib'
CMake Error at build/generators/cmakedeps_macros.cmake:39 (message):
  Library 'GeographicLib' not found in package.  If 'GeographicLib' is a
  system library, declare it with 'cpp_info.system_libs' property
  build/generators/geographiclib-Target-debug.cmake:24 (conan_package_library_targets)
  build/generators/geographiclibTargets.cmake:26 (include)
  build/generators/geographiclib-config.cmake:16 (include)
  CMakeLists.txt:4 (find_package)`

But if I revert it then working fine. Haven't tested it with Conan 2.x.

uilianries:
@zeeshancs07 Thank you for reviewing. It worked before because was using `collect_libs()` but we avoid that because is unsafe in terms of respecting the link order. Plus, maintainers would not have idea about the expected library name. That's why I changed to the library.

Still, I only tested on Mac (where I'm running right now), but in Windows it uses another logic:

> -- Installing: C:/J/workspace/prod-v1/bsr/49569/abacc/.conan/data/geographiclib/1.51/_/_/package/36c0d435f1aed51582a09c4232131a376d19151c/lib/Geographic_d.lib

It adds a suffix related to Debug too. I'll update the PR to follow that rule.

recipes/geographiclib/all/conanfile.py
```diff
@@ -132,12 +128,19 @@ def package(self):
...
+            for it in bin_files:
+                rm(self, it, os.path.join(self.package_folder, "bin"))
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "geographiclib")
         self.cpp_info.set_property("cmake_target_name", "GeographicLib::GeographicLib")
         self.cpp_info.set_property("pkg_config_name", "geographiclib")
-        self.cpp_info.libs = collect_libs(self)
+        suffix = "_d" if self.settings.build_type == "Debug" and is_msvc(self) else ""
+        suffix += "-i" if is_msvc(self) else ""
```
zeeshancs07, line 142: Not sure about this suffix, as this is still causing the same issue on Windows.

uilianries: I'll enter in Windows to check it. Thank you for running it.

SpaceIm: According to upstream CMakeLists:

- Generic name is `GeographicLib` irrespective of major version.

- Postfix is `-i` if compiler is cl like (msvc & clang-cl) and lib is shared.

- Debug postfix is unpredictable if compiler is not cl-like, because there is a condition based on generator: https://github.com/geographiclib/geographiclib/blob/v2.3/CMakeLists.txt#L187-L193

I would keep `collect_libs()` in this recipe, or isolate this change in another PR. But you can expect regressions if you remove collect_libs, due to unpredictable naming and even difficulties to predict if CMake variable `MSVC` variable is true or not based on conan profile.

recipes/geographiclib/all/conanfile.py
```diff
@@ -132,12 +128,19 @@ def package(self):
         ]:
             rmdir(self, os.path.join(os.path.join(self.package_folder, folder)))
         rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
+        if not self.options.tools:
+            rmdir(self, os.path.join(self.package_folder, "sbin"))
+            bin_files = [it for it in os.listdir(os.path.join(self.package_folder, "bin")) if not it.endswith(".dll")]
+            for it in bin_files:
+                rm(self, it, os.path.join(self.package_folder, "bin"))
```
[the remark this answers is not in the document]
uilianries: Another possible is because is the past we usually added tools option for almost everything. But indeed need to re-check the initial commit.

zeeshancs07 (approved):

RubenRBS (approved):

zeeshancs07 (approved):

recipes/geographiclib/all/conanfile.py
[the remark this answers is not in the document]
ericLemanissier: @SpaceIm if you don't know why this option is here (https://github.com/conan-io/conan-center-index/pull/4295), I'd say we should ask @zeeshancs07 why they need it, what they expect it to do, and if possible remove the option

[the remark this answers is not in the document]
ericLemanissier: I guess the question is: would it be a breaking change if we removed this option, and always build and package the tools ?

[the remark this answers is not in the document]
uilianries: Do not need to remove, the upstream confirmed a way to avoid tools, so it's something officially supported: https://github.com/geographiclib/geographiclib/pull/27#issuecomment-2133322100

ericLemanissier (approved):

recipes/geographiclib/all/conanfile.py
SpaceIm: I'm confused. @uilianries did you read what I wrote here? Lib name is unpredictable for conan in some cases (generator dependent), and clang-cl is likely broken now.

PR #24142, opened 2024-05-28
strawberryperl:  Fix binaries not supporting non x86 archs

RubenRBS:
Found while trying to help in https://github.com/conan-io/conan-center-index/pull/23955


ericLemanissier (approved):
Out of curiosity, which practical case is excluded by this PR, Windows on Arm ?

RubenRBS:
Yes :)
I found this while trying to consume a graph from Parallels from my Macos m2 host

PR #24154, opened 2024-05-29
protobuf new versions

jcar87:
* Add versions 25 and 27, using the versioning for C++ described in https://protobuf.dev/support/version-support/#cpp 
* Add dependency on abseil for versions >22, and record actual dependencies on specific abseil components
* Expose `upb` option for versions 27 and higher. Set to False by default - but this might change in the future (being cautious as they are not currently used externally and they are unconditionally static)
* Require a minimum of C++14 for versions 22 and higher (since they require abseil)
* Refactor logic to handle defining the `protobuf::protoc` target, now it's done in a separate file without in-recipe patching
* define and expose `utf8` targets, ONLY when the library is built statically (or upb is required). These targets are unconditionally static and if protobuf is shared, they are private and hidden (and consumers must not link against them)


recipes/protobuf/all/conanfile.py
```diff
@@ -59,17 +67,34 @@ def configure(self):
...
+
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
         if self.options.with_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
 
+        if self._protobuf_release >= "22.0":
+            self.requires("abseil/20240116.2", transitive_headers=True)
```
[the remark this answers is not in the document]
jcar87: added this in the validate() method!

jcar87:
@coryan, if you could confirm that following this version scheme for the Conan recipes of protobuf is sound, that would be great!
Also removing older versions, after ensuring that most of the repo is on the most recent 3.21 we provided

recipes/protobuf/all/conanfile.py
```diff
@@ -59,17 +67,40 @@ def configure(self):
...
+
     def validate(self):
         if self.options.shared and is_msvc_static_runtime(self):
             raise ConanInvalidConfiguration("Protobuf can't be built with shared + MT(d) runtimes")
-
+        
+        if is_msvc(self) and self._protobuf_release >= "22" and self.options.shared and \
+            not self.dependencies["abseil"].options.shared:
```
uilianries, line 88: Is too risk enforce `self.options["abseil"].shared = True`? In Conan 1.x, we will not have packages for msvc + shared due mixed profiles.

recipes/protobuf/all/conandata.yml
```diff
@@ -1,62 +1,99 @@
 sources:
+  "5.27.0":
+    url: "https://github.com/protocolbuffers/protobuf/archive/refs/tags/v5.27.0.tar.gz"
+    sha256: "1611a03e550c27c526bde1da544c94f7aa65c10687bbc8e570537dfa94069e1a"
+  "4.25.3":
+    url: "https://github.com/protocolbuffers/protobuf/archive/refs/tags/v4.25.0.tar.gz"
```
coryan, line 6: Why the 4.25.0 tarball with the 4.25.3 version?

coryan:
> @coryan, if you could confirm that following this version scheme for the Conan recipes of protobuf is sound, that would be great! Also removing older versions, after ensuring that most of the repo is on the most recent 3.21 we provided

The major versions for 4.25.x and 5.27.x both look correct to me.  And thanks for cleaning up older versions.

recipes/protobuf/all/conandata.yml
jcar87: Very well spotted!! I tested several versions and this was a leftover. Thanks for noticing, will fix ASAP

jcar87:
> > @coryan, if you could confirm that following this version scheme for the Conan recipes of protobuf is sound, that would be great! Also removing older versions, after ensuring that most of the repo is on the most recent 3.21 we provided
> 
> The major versions for 4.25.x and 5.27.x both look correct to me. And thanks for cleaning up older versions.

Thanks so much!
After this is merged, will move on to gRPC and google-cloud-cpp - I have a feeling that there's chance finding a coherent version of `abseil` that works with all 3 may be tricky, already ran into issues building gRPC with the version of abseil in this PR :/

recipes/protobuf/all/conanfile.py
jcar87: Don't think it's worth it particularly - we know that shared works because of the Conan 2 checks, and in other recent cases we already decided to go the `validate()` way (and point the user to a more sensible choice), than to embed static inside shared (although on Windows it may be less of a risk).

RubenRBS (approved):

recipes/protobuf/all/conanfile.py
uilianries: > in other recent cases we already decided to go the validate() way (and point the user to a more sensible choice), than to embed static inside shared (although on Windows it may be less of a risk).

Good to know. Thanks!

uilianries (approved):
LGTM, good job.

saschasc:
@jcar87  @uilianries Thanks a lot for pushing this update. Much appreciated!

recipes/protobuf/all/test_package/CMakeLists.txt
```diff
@@ -17,3 +21,10 @@ if(TARGET protobuf::libprotoc)
     target_link_libraries(${PROJECT_NAME} PRIVATE protobuf::libprotoc)
 endif()
 
+if(NOT TARGET protobuf::protoc)
+    message(FATAL_ERROR "protoc executable should have been defined as part of find_package(protobuf)")
+endif()
+
+if(NOT COMMAND protobuf_generate)
+    message(FATAL_ERROR "protobuf_generate should have been defined as part of find_package(protobuf)")
+endif()
```
valgur, line 30: That's a very good compromise. Thanks! :slightly_smiling_face:

PR #24157, opened 2024-05-29
jsonifier: add recipe

toge:
Specify library name and version:  **jsonifier/***

https://github.com/RealTimeChris/Jsonifier

---

### DIFF
--- a/recipes/jsonifier/all/conanfile.py
+++ b/recipes/jsonifier/all/conanfile.py
@@ -1,0 +1,72 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.files import get, copy, rmdir
+from conan.tools.build import check_min_cppstd
+from conan.tools.scm import Version
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+import os
+
+
+required_conan_version = ">=1.53.0"
+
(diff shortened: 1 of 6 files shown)


RubenRBS (approved):
Thanks!

recipes/jsonifier/all/conanfile.py
```diff
@@ -0,0 +1,72 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
```
RubenRBS, line 62: Hi @toge I just realized that this library is calling cmake build but it's meant to be a header-only library. Could I get a bit more insight into why this is needed? Thanks!

RubenRBS (commented):
A question came up while I was showing @perseoGI around some PRs :)

recipes/jsonifier/all/conanfile.py
RubenRBS: No, I think this is ok, thanks a lot for checking, it was only a bit suprising :)

RubenRBS (approved):

perseoGI (approved):
Looks good!

PR #24159, opened 2024-05-29
[libprotobuf-mutator] Compatible with Conan 2.x

uilianries:
Specify library name and version:  **libprotobuf-mutator/1.1**

Ported libprotobuf-mutator to work with Conan 2.x, including:

- Dropped custom version support cci.xxx
- Added support for version 1.1 (Sep 2023)
- Added separated components for libprotobuf-mutator and libprotobuf-mutator-fuzzer
- Added fPIC and shared options
- Updated Protobuf to 3.21 as requirement
- Dropped support for legacy CMake generators cmake_find_package
- Simplified test package to not use protoc

NOTES:

- Versions 1.2 and 1.3 (latest) requires newer Protobuf versions (PR #24154) and Abseil.
- We can add new versions in a separated PR, so this one will be simple.
- Tested locally on Linux.

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24163, opened 2024-05-30
[libprotobuf-mutator] Compatible with Conan 2.x

uilianries:
Specify library name and version:  **libprotobuf-mutator/1.3**

The [libprotobuf-mutator](https://github.com/google/libprotobuf-mutator) has now official tags, which means we can drop the old custom version in CCI.

This PR is a follow-up of PR #24159. Between 1.1 and 1.3 some things changed and required more attention, including:

- Requires Protobuf release 25. The latest release 26 is incompatible, failed to build locally.
- It requires C++14, and it's enforced to C++14 only by set(CMAKE_CXX_STANDARD ...
- It gets a ride from Protobuf dependency and consumes abseil::StrCat method (fragile)
- Requires CMake 3.24 as minimal

Notes about this PR in concrete:

- I preferred using a CMake wrapper to fix protobuf and abseil link, because it's less invasive and I'll send a PR to the upstream, so we don't keep different patches. Abseil is used, but not directly linked by the project, it uses from Protobuf transitive dependency.
- Dropped Conan 1.x test package to validate cmake_find_package and so on
- Dropped custom cci version because was too old and current tags follow last commits in upstream
- Added shared and fPIC options. Both working on Linux (tested locally)
- Added components to follow same CMake targets provided by the upstream

**TODO**: Send a PR to the upstream fixing not only abseil usage, but cxx standard too.

---


uilianries:
The project does not expose symbols in .lib when building Windows + shared=True. Dropped shared option when building with MSVC. I'll send a PR to upstream trying to fix it.

jcar87 (approved):

PR #24038, opened 2024-05-20
Updated openblas from validate to validate_build

queltosh:
Specify library name and version:  openblas/*

fixes #24037

Validation method fails on cross compilation environment despite  native packages are in repo.  Apparently is because validate is run always, and this validations should be run only on build stage. Renaming the method name to validate_build looks better approach.

---

- [X] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [X] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24171, opened 2024-05-30
openblas: cross-compilation support

tttapa:
Specify library name and version:  **openblas/\***

This PR adds support for cross-compiling OpenBLAS. In previous versions of the recipe, cross-building was explicitly disabled.

When cross-compiling, OpenBLAS requires the `TARGET` option to be set, so this option has also been added to the Conan recipe. By default, it is set to `None` which has no effect. When cross-compiling, the default `TARGET` is determined based on `settings.arch`.

Native builds should not be affected by these changes.

---

Differences compared to other open pull requests for the OpenBLAS recipe:

- https://github.com/conan-io/conan-center-index/pull/24038:  
_Checking for cross-compilation support should be done in `validate_build()`, not `validate()`_.  
This change is included in the present PR as well, but the check is now different.

- https://github.com/conan-io/conan-center-index/pull/23207:  
Support for ARMv8 is also covered by the present PR, without the need for backporting patches.  
`#23207` uses `tc.cache_variables["TARGET"] = str(self.settings.arch).upper()`, which only works for some architectures, the dictionary in the present PR supports all common architectures, not just ARM.

---

### DIFF
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -141,6 +141,11 @@ class OpenblasConan(ConanFile):
                 raise ConanInvalidConfiguration(f'"{self.name}/*:build_relapack=True" option is only supported for GCC and Clang')
 
     def validate_build(self):
+        if Version(self.version) < "0.3.22" and cross_building(self, skip_x64_x86=True):
+            # OpenBLAS CMake builds did not support some of the cross-compilation targets in 0.3.20/21 and earlier.
+            # This was fixed in https://github.com/OpenMathLib/OpenBLAS/pull/3714 and https://github.com/OpenMathLib/OpenBLAS/pull/3958
+            raise ConanInvalidConfiguration(f"Cross-building is not supported for {self.name}/0.3.21 and earlier.")
+
         # If we're cross-compiling, and the user didn't provide the target, and
         # we couldn't infer the target from settings.arch, fail
         if cross_building(self, skip_x64_x86=True) and not self.options.target:
--- a/recipes/openblas/all/conanfile.py
+++ b/recipes/openblas/all/conanfile.py
@@ -11,6 +11,53 @@ import textwrap
 
 required_conan_version = ">=1.53.0"
 
+# Maps Conan's settings.arch to the corresponding OpenBLAS TARGET:
+conan_arch_to_openblas_target = {
+    "x86": "SANDYBRIDGE",  # Sandy bridge was discontinued in September of 2013,
+    "x86_64": "SANDYBRIDGE",  # supporting older CPUs is a performance trade-off
+    "ppc32be": None,  # TODO: OpenBLAS has POWER4, POWER5, POWER6, POWER7,
+    "ppc32": None,  #         POWER8, POWER9, POWER10, PPCG4, PPC970, PPC970MP,
+    "ppc64le": None,  #       PPC440, PPC440FP2, CELL
+    "ppc64": None,
+    "armv4": None,  # Not supported by OpenBLAS
+    "armv4i": None,  # Not supported by OpenBLAS
+    "armv5el": "ARMV5",
+    "armv5hf": "ARMV5",
+    "armv6": "ARMV6",
+    "armv7": "ARMV7",
+    "armv7hf": "ARMV7",
+    "armv7s": "ARMV7",
+    "armv7k": "ARMV7",
+    "armv8": "ARMV8",
+    "armv8_32": "ARMV7",  # No 32-bit ARMv8 TARGET in OpenBLAS
+    "armv8.3": "ARMV8",
+    "arm64ec": "ARMV8",
+    "sparc": None,  # TODO: OpenBLAS has SPARC, SPARCV7
+    "sparcv9": None,
+    "mips": None,  # TODO: OpenBLAS has P5600, MIPS1004K and MIPS24K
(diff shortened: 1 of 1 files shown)


elvisdukaj:
I was looking for this :)

recipes/openblas/all/conanfile.py
```diff
@@ -38,13 +86,15 @@ class OpenblasConan(ConanFile):
...
         "dynamic_arch": False,
+        "target": None,
     }
     options_description = {
         "build_lapack": "Build LAPACK and LAPACKE",
         "build_relapack": "Build with ReLAPACK (recursive implementation of several LAPACK functions on top of standard LAPACK)",
         "use_thread": "Enable threads support",
         "use_locking": "Use locks even in single-threaded builds to make them callable from multiple threads",
         "dynamic_arch": "Include support for multiple CPU targets, with automatic selection at runtime (x86/x86_64, aarch64 or ppc only)",
+        "target": "OpenBLAS TARGET variable (see TargetList.txt)",
```
samuel-emrys, line 97: Where is TargetList.txt?

tttapa: It's part of OpenBLAS itself: https://github.com/OpenMathLib/OpenBLAS/blob/4e9144b39faa6b0542e1b7c183e0dab1a8637a07/TargetList.txt

recipes/openblas/all/conanfile.py
```diff
@@ -68,23 +118,39 @@ def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
+        # When cross-compiling, OpenBLAS requires explicitly setting TARGET
+        if cross_building(self, skip_x64_x86=True) and not self.options.target:
+            # Try inferring the target from settings.arch
+            target = conan_arch_to_openblas_target.get(str(self.settings.arch))
+            if target:
+                self.output.warning(f'Setting OpenBLAS TARGET={target} based on settings.arch. This may result in suboptimal performance. Set the "{self.name}/*:target=XXX" option to silence this warning.')
+                self.options.target = target
```
jcar87, line 127: question here!

if we only set `self.options.target` when `cross_building`, that means that given a host profile, this recipe will have different package IDs depending on whether it was crossbuilt or not. This is not ideal - one may expect a "producer" such as CI to be able to cross-build, and a consumer on the "native" target system will not find that package because when not-crossbuild this option will default to `None` - unless I'm mistaken.

I wonder what values are mapped when NOT cross building - but if this file https://github.com/OpenMathLib/OpenBLAS/blob/442dec28df3c04b32e6f8630b40080d6a5715843/cmake/prebuild.cmake#L1402 is any indication, I have a suspicion that it is dynamically setting based on the features of the CPU of the machine doing the build? Then we're also in the same situation: for the same package ID, the contents may be different depending on which machine the build ran on, correct?

If this is the case - would it make sense to _always_ give `self.options.target` a value derived from the `arch` setting - regardless of whether or not it is cross-compiled, and point the user to set one that better suits their needs if necessary? I think that solves that particular concern

jcar87 (commented):
Thanks @tttapa, this is excellent!

I had a bit of trouble locating where the `TARGET` CMake variable was used in the upstream scripts - 

please see my comment above and thanks so much for solving this one!

recipes/openblas/all/conanfile.py
tttapa: Yes, I agree. The reason why I set `target=None` by default is to keep the existing behavior and not break things for current users. But as you said, this is a lie, since OpenBLAS will dynamically change the options in that case, and this _should_ result in different package IDs.

To complicate matters even further, there doesn't seem to be a way to disable the automatic detection in OpenBLAS, the logic you linked to is always executed `if (NOT CMAKE_CROSSCOMPILING)`, so we'd need to patch that (and hope nothing else breaks) if we want to explicitly set `TARGET` from Conan.

And then there's `DYNAMIC_ARCH`, which includes multiple versions of each function, and dispatches dynamically based on the target it's being executed on. I haven't tried this myself, and it does of course increase binary size and compile times.

Does Conan have any standard way to deal with different microarchitectures? I suspect that any library that uses SIMD or other modern instructions runs into this specific issue, so maybe this should be standardized (e.g. as a setting in the user's profile). E.g. Spack supports this: https://spack.readthedocs.io/en/latest/basic_usage.html#support-for-specific-microarchitectures. An alternative could be to adopt the System V ABI micro-architecture levels with coarser granularity: https://gitlab.com/x86-psABIs/x86-64-ABI

A quick google search only yields some old discussions without solutions like https://github.com/conan-io/conan/issues/847. This one about OpenBLAS's `TARGET` option is definitely relevant, though: https://github.com/conan-io/conan-center-index/pull/13478

elvisdukaj:
Maybe it’s worth considering creating an MR on Conan?

recipes/openblas/all/conanfile.py
jcar87: thanks @tttapa  - indeed those are the issues.

I think the current PR is fine:
- it keeps the current behaviour
- it gives an ability that wasn't there before

while I think it's worth fixing this issues, since doing so may result in changing the default option value - it's probably best to handle that in a follow-up PR

jcar87 (approved):

RubenRBS (approved):

perseoGI (approved):

Issue #24172, opened 2024-05-30
[package] geographiclib/2.3: Run unitests on GeographicLib Builds


FannoFlow1:
### Description

The current conanfile.py for geographiclib doesn't run unitests when building locally `--build=missing`
It likely should so downstream users know if there is an issue with geographiclib when compiling with different compiler settings. 

I think this requires just a `cmake.test()` addition to the build stage. 

### Package and Environment Details

* Package Name/Version: **zlib/1.2.8**
* Operating System+version: **Linux Ubuntu 18.04**
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

Build profile:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

### Steps to reproduce

conan install --requires=geographiclib/2.3 --build=*

### Logs

<details><summary>Click to expand log</summary>

```
geographiclib/2.3: Running CMake.build()
```

</details>

PR #24173, opened 2024-05-30
Update conanfile.py

FannoFlow1:
try adding cmake.test()

Specify library name and version:  **geographiclib/2.3**

resolves https://github.com/conan-io/conan-center-index/issues/24172
---

### DIFF
--- a/recipes/geographiclib/all/conanfile.py
+++ b/recipes/geographiclib/all/conanfile.py
@@ -115,6 +115,7 @@ class GeographiclibConan(ConanFile):
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
+        cmake.test()
 
     def package(self):
         copy(self, "LICENSE.txt", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))


jcar87:
Hi @FannoFlow1, thank you for your contribution.

We don't currently build or run tests during the package builds, in particular in our CI as the validity of the packages is a responsibility we place on the users who wish to test that. This is also in line with other package managers.

However let me check with the team, as it is still a useful feature that should be optional in the recipes.
We introduced the `tools.build:skip_test` configuration for this very purpose - this way the recipes could conditionally build and run tests.

FannoFlow1:
> Hi @FannoFlow1, thank you for your contribution.
> 
> We don't currently build or run tests during the package builds, in particular in our CI as the validity of the packages is a responsibility we place on the users who wish to test that. This is also in line with other package managers.
> 
> However let me check with the team, as it is still a useful feature that should be optional in the recipes. We introduced the `tools.build:skip_test` configuration for this very purpose - this way the recipes could conditionally build and run tests.

Ah interesting! For some reason I thought I was seeing unit tests in other packages being built, with this being an outlier. I must have been seeing something else in the logs since you're right the recipes I'm looking at don't run tests. 

I think it would be very desirable to have `cmake.test()` in the conan center recipes which leverages the  `tools.build:skip_test` property internally.

FannoFlow1:
I would also argue it would be very important to run the unit tests with the conancenter index packages because there are so many compiler variables. The compiler settings the package author tested with might be (or likely are) different than the conan center index builds. And I'm not sure its a safe assumption to argue that those differences dont matter.
Now most probably dont have an affect on most packages, but I'm not sure you can guarantee that for all packages in all builds without running the unit tests.

And same goes for the downstream user building from source. At best you'll get compiler / linker errors, at worst you have a silent runtime issue giving you the wrong answer on some math.

perseoGI:
Hi @FannoFlow1!

Thank you for your interest in Conan Center Index.
Running unit tests is not a CCI policy so we are closing this PR because we can't find the value in deviating from it.

Nevertheless, thanks a lot for taking the time to open the PR, we appreciate it 🐸

Issue #24180, opened 2024-06-01
[package] fltk/1.3.9: MSVC static runtime not respected


stgatilov:
### Description

Conan's setting `compiler.runtime = "static"` is ignored and package is always built against dynamic runtime.
That's because CMakeLists.txt of FLTK expect an option named `FLTK_MSVC_RUNTIME_DLL`, which conan recipe does not set.

To reproduce, build any program that uses FLTK on MSVC with runtime set to static.

### Package and Environment Details

* Package Name/Version: **fltk/1.3.9**
* Operating System+version: **Windows 10 x64**
* Compiler+version: **MSVC 2022**
* Conan version: **conan 2.3.1**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows

### Steps to reproduce

A simple test app using FLTK is here:
* https://gist.github.com/stgatilov/f8b8a1a488206fe2889f7e7e2f99438c

I built it with static runtime like this:
```
conan install . -of build -s compiler.runtime=static --build=missing
cd build
cmake .. -G "Visual Studio 17 2022" -DCMAKE_TOOLCHAIN_FILE=conan_toolchain.cmake  -DCMAKE_POLICY_DEFAULT_CMP0091=NEW
cmake --build . --config Release
```

### Logs

<details><summary>Click to expand log</summary>

```

D:\StevePrograms\temp2>conan install . -of build -s compiler.runtime=static --build=missing 

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=static
compiler.runtime_type=Release
compiler.version=193
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows


======== Computing dependency graph ========
Graph root
    conanfile.txt: D:\StevePrograms\temp2\conanfile.txt
Requirements
    fltk/1.3.9#5a6970da04dd7b2d7059177e45c434ef - Cache
    libjpeg/9e#a794b5e74808992b08a21b126e19eab4 - Cache
    libpng/1.6.43#c219d8f01983bac10c404fc613605eef - Cache
[...]

PR #24181, opened 2024-06-01
fltk: honor msvc static runtime

SpaceIm:
Since 1.3.9, there is an option `FLTK_MSVC_RUNTIME_DLL` which forces `CMAKE_MSVC_RUNTIME_LIBRARY` and therefore defeats conan toolchain. Therefore this `FLTK_MSVC_RUNTIME_DLL` option must be managed by recipe itself.
See https://github.com/fltk/fltk/blob/release-1.3.9/CMake/options.cmake#L76-L83

closes https://github.com/conan-io/conan-center-index/issues/24180

---

### DIFF
--- a/recipes/fltk/all/conanfile.py
+++ b/recipes/fltk/all/conanfile.py
@@ -4,6 +4,8 @@ from conan import ConanFile
 from conan.tools.apple import is_apple_os
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, collect_libs, copy, export_conandata_patches, get, rm, rmdir
+from conan.tools.microsoft import msvc_runtime_flag
+from conan.tools.scm import Version
 
 required_conan_version = ">=1.53.0"
 
@@ -35,6 +37,14 @@ class FltkConan(ConanFile):
         "with_gdiplus": True,
         "with_xft": False,
     }
+
+    @property
+    def _is_cl_like(self):
+        return self.settings.compiler.get_safe("runtime") is not None
+
+    @property
+    def _is_cl_like_static_runtime(self):
+        return self._is_cl_like and "MT" in msvc_runtime_flag(self)
 
     def export_sources(self):
         export_conandata_patches(self)
@@ -96,6 +106,9 @@ class FltkConan(ConanFile):
         tc.variables["OPTION_USE_SYSTEM_LIBJPEG"] = True
         tc.variables["OPTION_USE_SYSTEM_ZLIB"] = True
         tc.variables["OPTION_USE_SYSTEM_LIBPNG"] = True
+        if Version(self.version) >= "1.3.9":
+            if self._is_cl_like:
+                tc.variables["FLTK_MSVC_RUNTIME_DLL"] = not self._is_cl_like_static_runtime
 
         tc.generate()
         tc = CMakeDeps(self)


RubenRBS (commented):
Thanks! I'll recheck with the team the main reasons behind not pushing the cl_like helpers in Conan itself, will get back to you tomorrow :)

recipes/fltk/all/conanfile.py
```diff
@@ -96,6 +106,9 @@ def generate(self):
         tc.variables["OPTION_USE_SYSTEM_LIBJPEG"] = True
         tc.variables["OPTION_USE_SYSTEM_ZLIB"] = True
         tc.variables["OPTION_USE_SYSTEM_LIBPNG"] = True
+        if Version(self.version) >= "1.3.9":
+            if self._is_cl_like:
+                tc.variables["FLTK_MSVC_RUNTIME_DLL"] = not self._is_cl_like_static_runtime
```
jcar87, line 111: looking at the settings.yml in both Conan 1 and Conan 2.x, would the following not work to cover all cases in a single line?

```
tc.variables["FLTK_MSVC_RUNTIME_DLL"] =  self.settings.get_safe("compiler.runtime") in ["dynamic", "MD", "MDd"]
```

in Conan 2 one would simply do:
```
tc.variables["FLTK_MSVC_RUNTIME_DLL"] = self.settings.get_safe("compiler.runtime") == "dynamic"
```

SpaceIm: It seems more elegant to me to introduce a generic helper we can use in other recipes (branching based on `MSVC` CMake variable is quite common), which could be replaced by the equivalent from conan API if it's implemented later on.

jcar87: bear in mind that it is unlikely that there will be a helper for this at all.

The `msvc_xx()` or `is_msvc()`  helpers for the most part exist to abstract away the conan v1/v2 discrepancies around `Visual Studio` vs `msvc` as compilers, and are not intended to be generic long-term solutions beyond the migration path.

If we were supporting Conan 2.0 there would be no issues moving away from the helpers to just doing one-line comparisons:
* `is_msvc_static_runtime()`  -> `self.settings.get_safe("compiler.runtime") == "static"`
* `is_msvc()` -> `self.settings.compiler == "msvc"` 

In Conan 2.0, the check for this PR would simply be:
```
tc.variables["FLTK_MSVC_RUNTIME_DLL"] = self.settings.get_safe("compiler.runtime") == "dynamic"
```

and I suspect that would cover all cases. 

Furthermore, I find that the name `_is_cl_like()` is confusing in the context of this PR, because what we need to discern here is whether or not we are using the MSVC ABI, which is orthogonal to whether or not clang is invoked via clang-cl.exe or clang.exe.

jcar87 (approved):

recipes/fltk/all/conanfile.py
SpaceIm: I still think such helper would be very helpful since it's not obvious for most contributors how to deduce if a conan profile maps to a cl like compiler.

SpaceIm:
```
Furthermore, I find that the name _is_cl_like() is confusing in the context of this PR, because what we need to discern here is whether or not we are using the MSVC ABI, which is orthogonal to whether or not clang is invoked via clang-cl.exe or clang.exe.
```

I don't know whether it's confusing, but it's the terminology used in CMake documentation to describe the meaning of MSVC = ON, and it's what we are intersted of most of the time (several recipes change the name of their lib, or have branching depending on this variable, and conan recipes may have to predict whether this variable is ON OR OFF depending on conan profile).

RubenRBS (approved):

PR #24187, opened 2024-06-02
OpenImageIO: Add version 2.5.12.0

irieger:
Specify library name and version:  **openimageio/2.5.12.0**

Just adding a new version. The CMake patch needed a small adaptation as a few lines had been changed in the upstream CMake file (`src/cmake/externalpackages.cmake`). Otherwise it is exactly the same as the patch for the previous version.

Also there was a need to bump to **openexr/3.2.3** (from 3.2.1) as opencolorio - which is a dependency of openimageio was bumped to this version in the meantime so the Conan 2 pipeline failed.

---

### DIFF
--- a/recipes/openimageio/all/conanfile.py
+++ b/recipes/openimageio/all/conanfile.py
@@ -82,7 +82,7 @@ class OpenImageIOConan(ConanFile):
         self.requires("boost/1.84.0")
         self.requires("libtiff/4.6.0")
         self.requires("imath/3.1.9", transitive_headers=True)
-        self.requires("openexr/3.2.1")
+        self.requires("openexr/3.2.3")
         if self.options.with_libjpeg == "libjpeg":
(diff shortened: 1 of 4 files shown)


recipes/openimageio/all/conanfile.py
```diff
@@ -82,7 +82,7 @@ def requirements(self):
         self.requires("boost/1.84.0")
         self.requires("libtiff/4.6.0")
         self.requires("imath/3.1.9", transitive_headers=True)
-        self.requires("openexr/3.2.1")
+        self.requires("openexr/3.2.3")
```
RubenRBS, line 85: This dependency bump solves internal conflicts within the tree, approving 👍

RubenRBS (approved):
Thanks a lot :)

uilianries (approved):
LGTM

PR #24188, opened 2024-06-02
mpc: Unset LIBS before configuring

guy-david:
mpc/1.3.1

On systems that do not have `libmpfr.so.6` installed, I fail to configure this package:
```
configure:3919: checking whether we are cross compiling
configure:3927: clang-16 -o conftest  -m64 -O3 -I/local/users/guyd/conan/.conan/data/gmp/6.3.0/_/_/package/581209b46b4309cd0f7261e5f10bfe21d5d4fc08/include -I/local/users/guyd/conan/.conan/data/mpfr/4.2.0/_/_/package/e4297c5e065990fd51a06a706f890cb821c50fae/include  -DNDEBUG -I/local/users/guyd/conan/.conan/data/mpfr/4.2.0/_/_/package/e4297c5e065990fd51a06a706f890cb821c50fae/include -I/local/users/guyd/conan/.conan/data/gmp/6.3.0/_/_/package/581209b46b4309cd0f7261e5f10bfe21d5d4fc08/include -L/local/users/guyd/conan/.conan/data/gmp/6.3.0/_/_/package/581209b46b4309cd0f7261e5f10bfe21d5d4fc08/lib -L/local/users/guyd/conan/.conan/data/mpfr/4.2.0/_/_/package/e4297c5e065990fd51a06a706f890cb821c50fae/lib  -m64 -L/local/users/guyd/conan/.conan/data/mpfr/4.2.0/_/_/package/e4297c5e065990fd51a06a706f890cb821c50fae/lib -L/local/users/guyd/conan/.conan/data/gmp/6.3.0/_/_/package/581209b46b4309cd0f7261e5f10bfe21d5d4fc08/lib conftest.c -lm -lmpfr -lgmpxx -lgmp -lm >&5
configure:3931: $? = 0
configure:3938: ./conftest
./conftest: error while loading shared libraries: libmpfr.so.6: cannot open shared object file: No such file or directory
```

My theory is that the AutoToolsToolchain populates the `LIBS` environment variable with the package's dependencies (i.e. `-lgmp, -lmpfr`), MPC then adds these to the implicit list of libraries it should link (`LIBS        libraries to pass to the linker, e.g. -l<library>`). That works fine during linking but to actually run the resulting executable, one needs to set `LD_LIBRARY_PATH` or compile with an `RPATH` which I believe is undesirable.
I think that it's not MPC at fault in this case because it manually configures the paths to its dependencies, so therefore I limited the scope of this "fix" but I'd rather someone with more experience will take a look.

---

### DIFF
--- a/recipes/mpc/all/conanfile.py
+++ b/recipes/mpc/all/conanfile.py
@@ -79,6 +79,11 @@ class MpcConan(ConanFile):
         tc = AutotoolsToolchain(self)
         tc.configure_args.append(f'--with-gmp={unix_path(self, self.dependencies["gmp"].package_folder)}')
         tc.configure_args.append(f'--with-mpfr={unix_path(self, self.dependencies["mpfr"].package_folder)}')
+        # Unset LIBS which contains -lgmp, -lmpfr, etc. because mpc's configure
+        # script will incorporate it in its command-line when checking whether
+        # the compiler works and will in turn fail with:
+        #  ./conftest: error while loading shared libraries: libmpfr.so.6: cannot open shared object file: No such file or directory
+        tc.environment.unset('LIBS')
         tc.generate()
 
         tc = AutotoolsDeps(self)


jcar87:
Hi @guy-david, thank you for your contribution.

I believe something else might be happening that explains what you are experiencing - `libmpfr` is listed as a conan dependency, and I can see logic in the recipe to handle this appropriately.

Could you please provide details of your conan profile and which conan command you are invoking that results in this error so that we could troubleshoot this on our end? thanks!

jcar87:
Hi @guy-david  - closing this due to inactivity, however please let us know if this is still an issue and will happily look into it.

From what I can tell from the small log snippet:
when creating a package for `mpc/1.3.1` where `clang` is the compiler on Linux, the `./configure` script fails trying to run a test executable when it check whether it is cross-building.

>  On systems that do not have libmpfr.so.6 installed, I fail to configure this package:
`mpfr` is a dependency in this graph - so it should come from Conan regardless of whether there is one installed in the system.


>  That works fine during linking but to actually run the resulting executable,
Indeed this can be confirmed by the following in the logs you've provided:

```
-L/local/users/guyd/conan/.conan/data/mpfr/4.2.0/_/_/package/e4297c5e065990fd51a06a706f890cb821c50fae/lib  -lmpfr
```
this should cause the linker to link the mpfr coming from conan 

>  That works fine during linking but to actually run the resulting executable, one needs to set LD_LIBRARY_PATH 

indeed - these lines in the recipe https://github.com/conan-io/conan-center-index/blob/194b8da35ffd1cf218e151b8a2f5f010a2e59727/recipes/mpc/all/conanfile.py#L75 should achieve exactly the latter (`LD_LIBRARY_PATH`) - such that when the configure script runs, LD_LIBRARY_PATH contains a path to those folders.

I can only reproduce what you're reporting if I remove those lines - do you experience this issue with the most recent revision of this recipe?

On the other hand, I can see you're using clang - this can also be an issue when the compiler is not set to pass `--as-needed` to the linker - defining a `CFLAGS=-Wl,--as-needed` also works - although in essence it has the same effect as what you're proposing (dont link what you don't need).

please let us know if the issue still exists and we can have a look, thanks!

PR #24202, opened 2024-06-03
cmaes: Add 0.10.0

### LINKED ISSUE #24201: [request] cmaes/0.10.0
### Package Name/Version

cmaes/0.10.0

### Webpage

https://github.com/CMA-ES/libcmaes

### Source code

https://github.com/CMA-ES/libcmaes

From the Readme:

libcmaes is a multithreaded C++11 implementation (with Python bindings) of algorithms of the CMA-ES family for optimization of nonlinear non-convex 'blackbox' functions.

My reason for wanting this in conan:
I used it for my Code in physics and it is the only dependency not in conan. I already have a recipe in my fork and want to contribute it.

phbasler:
Specify library name and version:  **cmaes/0.10.0**

Add new package cmaes according to issue #24201 
Fixes #24201 

I am not the author of cmaes, but according to the authors in https://github.com/CMA-ES/libcmaes/issues/233 I can provide this recipe.

---- 

### Maintainers changes

- Dropped MSVC support. Upstream still has not accepted the patch. Originally, the project did not support MSVC and is delaying this PR. In case of accepting and merging, we can have a new PR in the future.
- Fixed both the generated CMake file and pkgconf (.pc) file name to follow libcmaes.

---


uilianries:
@phbasler Thank you for your PR! Please, take a look in you emails/accounts used in this PR: https://github.com/conan-io/conan-center-index/pull/24202#issuecomment-2144334275

The CLA bot listed 2 emails and it will not approve if both agrees to the CLA.

phbasler:
@uilianries 
Thanks.
The first commit was signed with an old e-mail adress which is not valid any longer.
I will have a look at it on how to change the assigned e-mail adress. Otherwise I will just redo the PR with the correct e-mail

phbasler:
I am a little confused here why the CI for linux with the profile 
```
[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.libcxx=libc++
compiler.version=13
os=Linux
[options]
cmaes:shared=False
[build_requires]
[env]
[conf]
tools.system.package_manager:sudo=True
```
fails.

If I run this locally with conan create and conan 1.64.1 then it works.

phbasler:
I am a little confused here.
Running `conan create` with the profile 
```
[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.libcxx=libc++
compiler.version=13
os=Linux
[options]
cmaes:shared=False
[build_requires]
[env]
CC=clang-13
CXX=clang++-13
```
and conan 1.64.1.
How can I fix the CI if this works locally?

recipes/cmaes/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+find_package(cmaes CONFIG REQUIRED)
+
+add_executable(example src/example.cpp)
+target_link_libraries(example libcmaes::cmaes)
+target_compile_features(example PUBLIC cxx_std_11)
```
phbasler, line 8: The package needs at least c++11 for the template headers.
Is this fine here or should this be moved to the `validate` function of the base profile to just exclude profiles with `cppstd<11`?

AbrilRBS: Both things actually! The best approach is to have the validate() method in the main recipe check for the min cppstd (Check how it is done in the [cmake_package](https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py) template), and also keep this so we know it works in the lowest valid cppstd :)

(Feel free to steal more ideas from that template if needed elsewhere too)

phbasler: I added the validate function

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+        self.requires("eigen/3.4.0", transitive_headers=True)
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["LIBCMAES_BUILD_EXAMPLES"] = False
+        tc.variables["LIBCMAES_BUILD_SHARED_LIBS"] = self.options.shared
+        tc.variables["LIBCMAES_USE_OPENMP"] = False
```
AbrilRBS, line 73: The current llvm-openmp package is not usable right now for cases like this, so for now we need to disable it

recipes/cmaes/all/patches/CompilerOptionsEigen.patch
```diff
@@ -0,0 +1,142 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 4a38212..dccfea4 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
```
AbrilRBS, line 4: I see this comes from https://github.com/CMA-ES/libcmaes/pull/234, do you have any insight into wether it will be merged eventually?

recipes/cmaes/all/patches/CompilerOptionsEigen.patch
```diff
@@ -0,0 +1,142 @@
...
+diff --git a/include/libcmaes/eigenmvn.h b/include/libcmaes/eigenmvn.h
+index fb8b11d..6a6943b 100644
+--- a/include/libcmaes/eigenmvn.h
++++ b/include/libcmaes/eigenmvn.h
+@@ -58,7 +58,20 @@ public:
+ 	static std::mt19937 rng;                        // The uniform pseudo-random algorithm
+ 	mutable std::normal_distribution<Scalar> norm; // gaussian combinator
+ 	
+-	EIGEN_EMPTY_STRUCT_CTOR(scalar_normal_dist_op)
++	scalar_normal_dist_op() = default;
```
AbrilRBS, line 126: I don't see this change referenced anywhere in the logs/upstream. Is this an issue on the library itself? If this is a backport, it'd be great to separate it to its own patch, so that we can crossreference back to it

recipes/cmaes/all/test_package/src/example.cpp
```diff
@@ -0,0 +1,61 @@
...
+  std::vector<double> xsol = bcand.get_x();
+
+  std::cout << "best solution: " << cmasols << std::endl;
+  std::cout << "optimization took " << cmasols.elapsed_time() / 1000.0 << " seconds\n";
+  return EXIT_SUCCESS;
+}
+
+int main()
+{
+  test1();
```
AbrilRBS, line 58: We could probably simplify this test_package, the less actual computation, the better and the more probable it is for the test package to stay valid for future versions

phbasler:
@AbrilRBS why did you remove the openmp option? Should the package not be used anymore?

AbrilRBS (commented):
Hi! Thanks for taking the time to create the PR, sorry that we didn't get back to you on a timely manner. I've left some comments on the recipe, and made a cleanup now that Conan 1 support is no longer needed :)

AbrilRBS:
> @AbrilRBS why did you remove the openmp option? Should the package not be used anymore?

Not that is should not be used per-se, it's more that its funcionality currently does not align with our expectations

recipes/cmaes/all/patches/CompilerOptionsEigen.patch
phbasler: I hope so. I will trigger @beniz again. If there is a new version with the PR, then I can adjust the conanfile here to not include this patch for the new version

recipes/cmaes/all/patches/CompilerOptionsEigen.patch
phbasler: This is removed in the current eigen master branch https://gitlab.com/libeigen/eigen/-/merge_requests/946 

I also got some problems because this additonally defines an empty copy ctr and we needed it to copy the members

recipes/cmaes/all/test_package/src/example.cpp
phbasler: I could remove the first test and just use the second test with the commented lines removed. Would that be ok?

phbasler: Done

phbasler:
@AbrilRBS Just a little bump. Do you need anything for the last two points?

phbasler:
@AbrilRBS @danimtb Anything to help with the review process?

recipes/cmaes/all/patches/CompilerOptionsEigen.patch
uilianries: In case it's from Eigen, then CMAES has to adapt itself to that change first. Please, check with the upstream on this case; the package manager should adopt this change without being solved by the upstream first.

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    homepage = "https://github.com/CMA-ES/libcmaes"
+    description = "libcmaes is a multithreaded C++11 library with Python bindings for high performance blackbox stochastic optimization using the CMA-ES algorithm for Covariance Matrix Adaptation Evolution Strategy"
+    topics = ("cmaes", "minimization")
+    package_type = "library"
+
+    settings = "os", "compiler", "build_type", "arch"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "surrog": [True, False],
```
uilianries, line 31:
```suggestion
```

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "surrog": True,
```
uilianries, line 36:
```suggestion
```

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["LIBCMAES_BUILD_EXAMPLES"] = False
+        tc.variables["LIBCMAES_BUILD_SHARED_LIBS"] = self.options.shared
+        tc.variables["LIBCMAES_USE_OPENMP"] = False
+        tc.variables["LIBCMAES_ENABLE_SURROG"] = self.options.surrog
```
uilianries, line 74:
```suggestion
```

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+        "fPIC": [True, False],
+        "surrog": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "surrog": True,
+    }
+
+    short_paths = True
```
uilianries, line 39:
```suggestion
```
Short paths are no longer used in Conan 2.x

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    export_conandata_patches,
+    get,
+    rmdir,
+    rm,
+    copy,
+    apply_conandata_patches,
+)
+from conan.tools.build import check_min_cppstd
+import os
+
```
uilianries, line 13:
```suggestion

required_conan_version = ">=2.0"
```

Make sure your new recipe is Conan 2.x compatible only. Conan Center Index no longer validate Conan 1.x

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    def validate(self):
+        check_min_cppstd(self, self._min_cppstd)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
+
+    def requirements(self):
+        self.requires("eigen/3.4.0", transitive_headers=True)
```
uilianries, line 64:
```suggestion
        # Transitive header: https://github.com/CMA-ES/libcmaes/blob/v0.10/include/libcmaes/eigenmvn.h#L36
        self.requires("eigen/3.4.0", transitive_headers=True)
```
Please, always give evidence when using transitive headers or transitive libs.

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rm(
+            self,
+            "*.cmake",
+            os.path.join(self.package_folder, "lib", "cmake", "libcmaes"),
+        )
```
uilianries, line 93:
```suggestion
        rmdir(self, os.path.join(self.package_folder, "lib"))
```

Remove the entire cmake folder, so it will not be empty when packaged.

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
+from conan import ConanFile
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout
```
uilianries, line 2:
```suggestion
from conan.tools.cmake import CMakeToolchain, CMakeDeps, CMake, cmake_layout
```

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+)
+from conan.tools.build import check_min_cppstd
+import os
+
+
+class CmaesConan(ConanFile):
+    name = "cmaes"
+
+    generators = "CMakeDeps"
+
```
uilianries, line 19:
```suggestion
```

Avoid mixing attributes when generate(), it may difficult when reviewing, and also may confuse users.

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,103 @@
...
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["LIBCMAES_BUILD_EXAMPLES"] = False
+        tc.variables["LIBCMAES_BUILD_SHARED_LIBS"] = self.options.shared
+        tc.variables["LIBCMAES_USE_OPENMP"] = False
+        tc.variables["LIBCMAES_ENABLE_SURROG"] = self.options.surrog
+        tc.variables["LIBCMAES_BUILD_PYTHON"] = False
+        tc.variables["LIBCMAES_BUILD_TESTS"] = False
+        tc.generate()
```
uilianries, line 77:
```suggestion
        tc.generate()
        deps = CMakeDeps()
        deps.generate()
```

uilianries (commented):
@phbasler Thank you for your time and effort in pushing this PR :smile: 

Just like @AbrilRBS, I'm concerned about these patches listed in this PR: We try to keep as close as possible to the upstream in terms of code to reflect the exact same behavior from there. We can accept patches, but usually as a last resort, we prefer to push a newer version with those fixes instead, so we do not need to "maintain" patches in the repository. You can obtain more information in [policy about patching](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#policy-about-patching) section. Also, when patching, please keep it as minimal as possible and well referenced from where it came, including using `patch_source` in the `conandata.yml` if possible, it helps all the community in terms of security and reliability. 

I also would keep this recipe as simple as possible, avoiding the option `surrog`. First, because the CI will not validate that option as False, so we need to build locally to check it. Second, because it can be configured via the Conan config `tools.cmake.cmaketoolchain:extra_variables` when needed from consumer side.

recipes/cmaes/all/conanfile.py
phbasler: The CI can build it with conan 2. Is this enough or are there some additional checks I can run?

uilianries: Sorry, I meant this attribute makes sure this recipe requires Conan 2.x at least: https://docs.conan.io/2/reference/conanfile/attributes.html#required-conan-version

recipes/cmaes/all/patches/CompilerOptionsEigen.patch
phbasler: I made these changes in https://github.com/CMA-ES/libcmaes/pull/234 but I am still waiting for a respone/merge there.
Once they merged it and made a new release I can happily make a new package release without patches.

As in the current case the latest release does not even build with cmake 4 without applying a patch

uilianries: Do you mean modifying the recipe, right? In that case, you can do it in this PR. 
Once the upstream merge your PR, you can open a second one to discuss the changes.

phbasler: Once they accept the PR and make a new release I can ditch the patch completly.
So it would just be the recipe + test_package then.

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,93 @@
...
+        copy(
+            self,
+            "COPYING",
+            self.source_folder,
+            os.path.join(self.package_folder, "licenses"),
+        )
+
+    def package_info(self):
+        self.cpp_info.libs = ["cmaes"]
+        self.cpp_info.set_property("cmake_target_name", "libcmaes::cmaes")
```
uilianries, line 93:
```suggestion
        self.cpp_info.set_property("cmake_target_name", "libcmaes::cmaes")
        self.cpp_info.set_property("cmake_file_name", "libcmaes")
```
Follow upstream's CMake file name: 
https://github.com/CMA-ES/libcmaes/blob/v0.10/CMakeLists.txt#L191C41-L191C49

recipes/cmaes/all/conanfile.py
```diff
@@ -0,0 +1,94 @@
...
+        self.cpp_info.set_property("cmake_file_name", "libcmaes")
```
uilianries, line 94:
```suggestion
        self.cpp_info.set_property("cmake_file_name", "libcmaes")
        self.cpp_info.set_property("pkg_config_name", "libcmaes")
```

https://github.com/CMA-ES/libcmaes/blob/v0.10/CMakeLists.txt#L156C46-L156C54

recipes/cmaes/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+find_package(cmaes CONFIG REQUIRED)
```
uilianries:
```suggestion
find_package(libcmaes CONFIG REQUIRED)
```

uilianries (approved):
LGTM. 

The linter accuses of using a bad URL for downloading, because it's a tag, not a proper release package. However, the upstream has no distributed release: https://github.com/CMA-ES/libcmaes/releases

phbasler:
@uilianries Thanks for adjusting this.
I will open a new PR once the PR on upstream is done.

PR #24203, opened 2024-06-03
civetweb: Add option to allow to set stack size per thread.

metalMajor:
Specify library name and version:  **civetweb/1.16**

When doing memory-intensive tasks in the thread that handles http requests, then it is needed to set the stack size of the thread. Default it is set to 102400 bytes, so that is low for some situations. This PR allows to set the existing CIVETWEB_THREAD_STACK_SIZE cmake value.

---

### DIFF
--- a/recipes/civetweb/all/conanfile.py
+++ b/recipes/civetweb/all/conanfile.py
@@ -22,6 +22,7 @@ class CivetwebConan(ConanFile):
         "fPIC": [True, False],
         "shared": [True, False],
         "ssl_dynamic_loading": [True, False],
+        "thread_stack_size": [None, "ANY"],
         "with_caching": [True, False],
         "with_cgi": [True, False],
         "with_cxx": [True, False],
@@ -39,6 +40,7 @@ class CivetwebConan(ConanFile):
         "fPIC": True,
         "shared": False,
         "ssl_dynamic_loading": False,
+        "thread_stack_size": None,
         "with_caching": True,
         "with_cgi": True,
         "with_cxx": True,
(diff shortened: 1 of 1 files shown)


RubenRBS (commented):
Thanks a lot! It think it would make sense to check in the `validate()` methid that the passed value can be parsed into an integer, something like
```
def validate(self):
    if self.options.thread_stack_size and not self.options.thread_stack_size.isdigit():
        raise ConanInvalidConfiguration("-o='civetweb/*:thread_stack_size' should be a positive integer")

**Comment by RubenRBS** [comment]
Thanks a lot for the followup and providing the compilation logs! :)

**Review by RubenRBS — APPROVED** [review]

**Review by uilianries — APPROVED** [review]
The log looks good!

**Comment by metalMajor** [comment]
Just a question: so when will this new option be visible in the conan center repo? I don't see it listed in the list of package options. How does it work?

**Comment by uilianries** [comment]
> Just a question: so when will this new option be visible in the conan center repo? I don't see it listed in the list of package options. How does it work?

Indeed that option is missing for Civetweb in Conan Center! I just triggered a new internal build to regenerate civetweb/1.16. Thank you for reporting!

**Comment by metalMajor** [comment]
Ah ok, thanks!

**Comment by uilianries** [comment]
@metalMajor The CI just finished and upload all new packages.

You now can check those new packages by running: `conan list -r conancenter "civetweb/1.16#fbdec252dc18b011fcd68d0663054ed8:*"`

**Comment by metalMajor** [comment]
I don't see the update it seems if you check the date:

```
$ conan list -r conancenter "civetweb/1.16#*"
conancenter
  civetweb
    civetweb/1.16
      revisions
        991b147a76abc500485606199e7c70aa (2023-04-17 09:49:35 UTC)
        6802d7300da7f20c7963e3c336d8593f (2023-05-06 08:18:08 UTC)
        193fde1a68fc376d0493d15c96aa888f (2023-09-11 22:34:02 UTC)
        d2291a8d994d00829d54545c95a077d2 (2023-10-08 12:41:13 UTC)
        fbdec252dc18b011fcd68d0663054ed8 (2024-06-04 09:31:00 UTC)
```

uilianries:
@metalMajor I just checked again and it's correct the current package. As its default value is `None`, is does not count in the package ID, but you can pass value to `thread_stack_size` option and this one should affect the package ID.


More info about options: https://docs.conan.io/2/reference/conanfile/attributes.html#options

PR #24204, opened 2024-06-03
qt: add fix for virtual can backend usage in multithreaded environment

lmife:
**qt/5.15.13 and qt/5.15.14**

This commit adds a patch for an issue in the QtSerialport virtual can backend.

The issue was identified by KDAB in one of our projects. KDAB has pushed a patch upstream, targeting Qt 6 (see https://codereview.qt-project.org/c/qt/qtserialbus/+/565329). As we are still using Qt 5, I'm providing a backported patch for Qt 5.15.13 and Qt 5.15.14 (sources are more or less unchanged, only some offset adaptions).

---

Also tested successfully with conan 1.64.1

### DIFF
--- a/recipes/qt/5.x.x/conandata.yml
+++ b/recipes/qt/5.x.x/conandata.yml
@@ -139,6 +139,11 @@ patches:
       "patch_file": "patches/5.15.12-fix-macos-cpp-lib-memory-resource.patch"
       "patch_source": "https://codereview.qt-project.org/c/qt/qtbase/+/482392"
       "patch_type": "portability"
+    - "base_path": "qt5/qtserialbus"
+      "patch_description": "Fix virtual can backend usage in multithreaded environment"
+      "patch_file": "patches/virtualcanback_mutex.patch"
+      "patch_source": "https://codereview.qt-project.org/c/qt/qtserialbus/+/565329"
+      "patch_type": "backport"
   "5.15.13":
     - "base_path": "qt5/qtbase"
       "patch_file": "patches/aa2a39dea5.diff"
@@ -162,6 +167,11 @@ patches:
       "patch_file": "patches/5.15.12-fix-macos-cpp-lib-memory-resource.patch"
       "patch_source": "https://codereview.qt-project.org/c/qt/qtbase/+/482392"
       "patch_type": "portability"
+    - "base_path": "qt5/qtserialbus"
+      "patch_description": "Fix virtual can backend usage in multithreaded environment"
+      "patch_file": "patches/virtualcanback_mutex.patch"
+      "patch_source": "https://codereview.qt-project.org/c/qt/qtserialbus/+/565329"
+      "patch_type": "backport"
   "5.15.12":
     - "base_path": "qt5/qtbase"
       "patch_file": "patches/aa2a39dea5.diff"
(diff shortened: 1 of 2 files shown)


RubenRBS:
Hi @lmife thanks a lot for taking the time to open the PR

I'm not against implementing this fix, but it should first be merged upstream for us to feel confident about adding it. Please ping me once it gets merged there so we can proceed, thanks!

jcar87:
HI @lmife - thank you for the update.

I have noticed that you are proposing to add this patch to versions 5.x, however, from the links you have provided this appears to be backported to 6.8, 6.7 and 6.5.

Is Qt also backporting this to Qt 5.15? Backporting from 6.8 all the way to 5.15 is risky for us, as we do not have a way to verify that the changes are correct - and we try to avoid as much as we can, maintaining fixes and backports that do not exist upstream (for 5.15), or where the original authors are unwilling to maintain - so the question is: is there evidence that Qt plans to support this for 5.15? If the answer is no, I don't see why we would either.

lmife:
> Is Qt also backporting this to Qt 5.15? Backporting from 6.8 all the way to 5.15 is risky for us, as we do not have a way to verify that the changes are correct - and we try to avoid as much as we can, maintaining fixes and backports that do not exist upstream (for 5.15), or where the original authors are unwilling to maintain - so the question is: is there evidence that Qt plans to support this for 5.15? If the answer is no, I don't see why we would either.

I do not think that there are plans from qt-project to cherry-pick this patch to 5.15.x as this is not a security issue. It is simply not possible to use the virtual can backend in a multithreaded environment without the patch.

But, the patch for Qt 6.x applies cleanly (only some offset changes) to Qt 5.15.x as the affected virtual can plug-in has not changed that much.

Although, I can understand if conan-io does not want to maintain such code. Please tell me how to proceed. Feel free to close the MR.

Additionally, we noticed that the current patch does not fully solve the problem. There will be a follow up patch upstream. (this might be another argument for you to not approve the MR :) )

jcar87:
> Additionally, we noticed that the current patch does not fully solve the problem. There will be a follow up patch upstream. (this might be another argument for you to not approve the MR :) )

Thanks for your reply. This is exactly our concern - the patch you are linking is for versions 6.x, but what is being proposed is for the much older 5.15.x. We ourselves have no way of verifying whether it works, beyond being able to say "it builds".

We provide libraries built from the source code from the canonical sources, "as is" - anything that goes beyond that, like backporting a patch further than what Qt own maintainers would do, is a concern:
A) makes us the sole maintainers  of this variation
B) breaks our promise of building from the canonical sources

For B),  there's mainly two concerns - one is functionality, and another is security. We don't know if the propose patch changes how Qt behaves vs what a user would expect, or if the fix actually works, and we don't have the ability or resources to validate that. For security, we also dont know if the patch introduces any security issues. A lot of users are performing security scanning based on the version number alone (e.g "5.15.x") - so we don't want to diverge: we want to avoid a situation where a security concern only exist in the Conan-provided recipe, and not upstream. I can appreciate that in this case, however, the origin of the patch is fully traceable, which is a good thing.

I'll close the PR on the grounds that:
- it seems unlikely qt themselves are backporting or validating this to 5.15
- it seems that it does not currently solve the problem

However, we'd like to help your case as much as we can. We know that there are a lot of users (in enterprise environments) that are happy maintaining either a fork of Conan Center, or a local repository from which to serve binaries to their teams. If there is anything we can do from our end to help this, including making it "easier" to add additional patches without modifying the recipe contents, we can certainly look into making this more seamless.

Please reply to this PR or open a new issue and we can help!

PR #23085, opened 2024-03-13
grpc : add version 1.64.2

sophieeihpos:
grpc/1.64.2

fixes #23084 
---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

© 2024 Morgan Stanley.
THIS SOFTWARE IS CONTRIBUTED SUBJECT TO THE TERMS OF THE Conan-io project Contributor License Agreement.
THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.

PR #24215, opened 2024-06-04
grpc add version 1.65

jcar87:
* Add grpc 1.65.0
* Refactor logic to handle component information in cpp_info to a separate file per version
* Use version range for abseil (only two versions match that range, should be compatible), and c-ares (C library with known backwards compatibility)
* Add logic to handle case when building `grpc` as a shared library on Linux and any of its dependencies are shared - this is due to the fact that the grpc build scripts invoke the just-built grpc_cpp_plugin. Not a problem on macOS due to rpath handling.
* Simplify test package


jcar87:
https://github.com/grpc/grpc/issues/36654

as we feared - very very sensitive to the exact version of protobuf, by the looks of it.

saschasc:
@jcar87 To move forward on this topic might it worth trying a version < 1.63.0 which still would resolve these two CVEs?

[CVE-2023-44487](https://nvd.nist.gov/vuln/detail/CVE-2023-44487), Severity: High
[CVE-2023-32731](https://nvd.nist.gov/vuln/detail/CVE-2023-32731), Severity: High

swaeberle:
This seems to be a duplicate to #23085, which adds v1.62.2 and v1.64.2.
Maybe both PRs can be merged?

AbrilRBS (approved):

danimtb (approved):

franramirez688 (approved):
Looks great 👏

jcar87:
> This seems to be a duplicate to #23085, which adds v1.62.2 and v1.64.2. Maybe both PRs can be merged?

thanks @swaeberle - have reviewed the PR - but it adds a rather complex (yet functional!) cmake parsing that we rather avoid maintaining. There are other changes in that PR worth considering (tangential to the new versions), so will reach out to the PR author.

Please comment or open an issue if you think it is still worth to add the older versions (1.62, 1.64) - I would advise to use 1.65 unless there's any specific reason not to.

swaeberle:
@jcar87 thanks for your explanation.

I think I can go on with v1.65.0. I just wanted to point out, that I stumbled upon another PR not yet linked to each other.

Thanks for your work :+1:

nextsilicon-itay-bookstein:
Build errors due to oversight in LD_LIBRARY_PATH logic under Linux has a proposed fix here: https://github.com/conan-io/conan-center-index/pull/24651

Presburger:
Why do you change the behavior of the old version when you add a new version? That's really stupid. FUCK

jcar87:
> Why do you change the behavior of the old version when you add a new version? That's really stupid. FUCK

Apologies if this has caused any issues - you are not providing any details, so unfortunately we cannot help. A couple of issues that have been reported to us, have been fixed as soon as we were made aware of them - and from what I can see only affected a subset of users rather than generalised. If there are any other issues that were not addressed by either https://github.com/conan-io/conan-center-index/pull/24651 or https://github.com/conan-io/conan-center-index/pull/24700 please provide details and we will look into it.

We strive to avoid causing disruption to any of our users. On top of this, Conan Center has revisions enabled (for Conan 1, and default behaviour in Conan 2) - so users have a variety of ways to ensure they dont automatically consume freshly published revisions  - chief among them being lockfiles. You can red about this here: https://docs.conan.io/2/devops/using_conancenter.html

That said, please consider editing your message and extending an apology. Further abuse will result in a ban.

PR #23469, opened 2024-04-10
openssl: add version `3.3.0`

gegles:
Specify library name and version:  **openssl/3.3.0**

* Normal entry addition for a new version
* Apparently, OpenSSL v3.3 now [added a CMake "Exporter"](https://github.com/openssl/openssl/commit/c768ccebc718ea0ed6afc5147fe4079fff632cd6#diff-ffa9c9947d71adb613fde0b54cc04eb4a74792b83e2e92a0addd1d8d7ae7a096), i.e. it generates an OpenSSLConfig.cmake. We thus need to delete this generated file as Conan does not want to rely upon it (See [KB-H016-CMAKE-MODULES-CONFIG-FILES](https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H016-CMAKE-MODULES-CONFIG-FILES)).

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24224, opened 2024-06-04
OpenSSL: Add 3.2.2, 3.1.6, 3.0.14 + removing old versions + small improvements

Croydon:
Building upon the work by @gegles in #23469 I also tried to add 3.3.x versions, but also couldn't get it to work. There is an encoding issue of file paths that I couldn't get to work with several different attempts of patching.

So this pull request contains improvements to the recipes applied along the way and adds new patch versions, but not 3.3.x versions for now.

GitHub is also going to be the only official download source, hence I removed the others. See https://openssl.org/blog/blog/2024/04/30/releases-distribution-changes/

---


gegles (approved):
Thanks @Croydon!

Wish you could also grab (or merge) what's in #23304 since it all builds well and was approved by 2 maintainers already). ;-) Thanks!

Croydon:
I understand that you are searching for a solution to your problem, but I am not going to merge the changes of #23304, sorry.  If the Conan team wants they could approve #23304 as it is in a mergeable state.

gegles:
> I understand that you are searching for a solution to your problem, but I am not going to merge the changes of #23304, sorry. If the Conan team wants they could approve #23304 as it is in a mergeable state.

Thanks. I think "my" problem, which is really more "How can Conan adapt to the non-standard way OpenSSL provides FIPS" problem, will actually be solved in a very nice way by what @memsharded did in https://github.com/conan-io/conan/pull/16132 (although that will require some change in the recipe as well).

I kept the PR open purely because it was still a worthy code cleanup to me (less hardcoded strings), but I am equally happy to close the PR if we don't see any value. thx!

Croydon:
> Thanks. I think "my" problem, which is really more "How can Conan adapt to the non-standard way OpenSSL provides FIPS" problem

Sure, did not mean this in a personal "you" way 🙂

RubenRBS (approved):
Thanks a lot for the cleanup both @Croydon and @gegles, we really appreciate it.
As per the 3.3.0 release, I anticipate we'll put some time apart to focus on trying to get a definitive answer as to how to advance on it this/next week, so thanks a lot for your patience

jcar87 (approved):
Thanks @Croydon  - yes, it's best to keep PRs separate, makes them a lot easier to review.

PR #24230, opened 2024-06-05
opentelemetry-cpp: Detect cppstd and CXX version

### LINKED ISSUE #24229: [package] opentelemetry-cpp/1.14.2: Fails to build with option "with_stl"
### Description

C++17 features are missing when building with STL (`-o with_stl=True`) and a profile, which doesn't specify `compiler.cppstd`.
C++14 is used.

Following options should be supported (see this breaking change):
https://github.com/open-telemetry/opentelemetry-cpp/blob/2535c70c4e4c1a92cb535413283b7c022c65554e/CHANGELOG.md?plain=1#L545

I'm going to create a pull request with improvements to solve this issue.

* Package Name/Version: **opentelemetry-cpp/1.14.2**
* Operating System+version: **Linux Debian 12**
* Compiler+version: **GCC 12**
* Docker image: none (built locally in wsl1)
* Conan version: **conan 1.64.0**
* Python version: **Python 3.11.2**

### Conan profile

[settings]
os=Linux
os_build=Linux
arch=x86_64
arch_build=x86_64
compiler=gcc
compiler.version=12
compiler.libcxx=libstdc++
build_type=Release
[options]
[build_requires]
[env]

### Steps to reproduce

**opentelemetry-cpp/1.14.2**
conan create recipes/opentelemetry-cpp/all/conanfile.py opentelemetry-cpp/1.14.2@ -o opentelemetry-cpp:shared=True -o opentelemetry-cpp:with_stl=True -o opentelemetry-cpp:with_abseil=False -o opentelemetry-cpp:with_otlp_grpc=False -o opentelemetry-cpp:with_otlp_http=False -o opentelemetry-cpp:with_zipkin=False

**opentelemetry-cpp/1.9.1**
conan create conan-center-index/recipes/opentelemetry-cpp/all/conanfile.py opentelemetry-cpp/1.9.1@ -o opentelemetry-cpp:shared=True -o opentelemetry-cpp:with_stl=True -o opentelemetry-cpp:with_abseil=False -o opentelemetry-cpp:with_otlp_grpc=False -o opentelemetry-cpp:with_otlp_http=False -o opentelemetry-cpp:with_zipkin=False -o opentelemetry-cpp:with_jaeger=False

### Logs

```
WARN: *** Conan 1 is legacy and on a deprecation path ***
WARN: *** Please upgrade to Conan 2 ***
Exporting package recipe
opentelemetry-cpp/1.14.2 exports: File 'conandata.yml' found. Exporting it...
opentelemetry-cpp/1.14.2 exports: Copied 1 '.yml' file: conandata.yml
opentelemetry-cpp/1.14.2: The stored package has not changed
opentelemetry-cpp/1.14.2: Using the exported files summary hash as the recipe revision: d317d01800bd89d59734f50e0e9d44f9
opentelemetry-cpp/1.14.2: Exported revision: d317d01800bd89d59734f50e0e9d44f9
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++
compiler.version=12
os=Linux
os_build=Linux
[options]
opentelemetry-cpp:shared=True
opentelemetry-cpp:with_abseil=False
opentelemetry-cpp:with_otlp_grpc=False
opentelemetry-cpp:with_otlp_http=False
opentelemetry-cpp:with_stl=True
opentelemetry-cpp:with_zipkin=False
[build_requires]
[env]

opentelemetry-cpp/1.14.2: Forced build from source
opentelemetry-cpp/1.14.2 (test package): Installing package
Requirements
    opentelemetry-cpp/1.14.2 from local cache - Cache
Packages
    opentelemetry-cpp/1.14.2:404724448056678c693431e88e9c16cf349104bf - Build

Installing (downloading, building) binaries...
opentelemetry-cpp/1.14.2: Copying sources to build folder
opentelemetry-cpp/1.14.2: Building your package in /home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf
opentelemetry-cpp/1.14.2: Generator txt created conanbuildinfo.txt
opentelemetry-cpp/1.14.2: Calling generate()
opentelemetry-cpp/1.14.2: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
opentelemetry-cpp/1.14.2: Preset 'release' added to CMakePresets.json. Invoke it manually using 'cmake --preset release'
opentelemetry-cpp/1.14.2: If your CMake version is not compatible with CMakePresets (<3.19) call cmake like: 'cmake <path> -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release/generators/conan_toolchain.cmake -DBUILD_TESTING=OFF -DBUILD_BENCHMARK=OFF -DWITH_EXAMPLES=OFF -DWITH_NO_DEPRECATED_CODE=OFF -DWITH_STL=ON -DWITH_GSL=OFF -DWITH_ABSEIL=OFF -DWITH_OTLP_GRPC=OFF -DWITH_OTLP_HTTP=OFF -DWITH_ZIPKIN=OFF -DWITH_PROMETHEUS=OFF -DWITH_ELASTICSEARCH=OFF -DWITH_ZPAGES=OFF -DWITH_JAEGER=OFF -DWITH_NO_GETENV=OFF -DWITH_ETW=OFF -DWITH_ASYNC_EXPORT_PREVIEW=OFF -DWITH_METRICS_EXEMPLAR_PREVIEW=OFF -DOPENTELEMETRY_INSTALL=ON -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=Release'
opentelemetry-cpp/1.14.2: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
opentelemetry-cpp/1.14.2: Aggregating env generators
opentelemetry-cpp/1.14.2: Calling build()
opentelemetry-cpp/1.14.2: CMake command: cmake -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE="/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/package/404724448056678c693431e88e9c16cf349104bf" -DBUILD_TESTING="OFF" -DBUILD_BENCHMARK="OFF" -DWITH_EXAMPLES="OFF" -DWITH_NO_DEPRECATED_CODE="OFF" -DWITH_STL="ON" -DWITH_GSL="OFF" -DWITH_ABSEIL="OFF" -DWITH_OTLP_GRPC="OFF" -DWITH_OTLP_HTTP="OFF" -DWITH_ZIPKIN="OFF" -DWITH_PROMETHEUS="OFF" -DWITH_ELASTICSEARCH="OFF" -DWITH_ZPAGES="OFF" -DWITH_JAEGER="OFF" -DWITH_NO_GETENV="OFF" -DWITH_ETW="OFF" -DWITH_ASYNC_EXPORT_PREVIEW="OFF" -DWITH_METRICS_EXEMPLAR_PREVIEW="OFF" -DOPENTELEMETRY_INSTALL="ON" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src"
-- Using Conan toolchain: /home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release/generators/conan_toolchain.cmake
-- Conan toolchain: Setting BUILD_SHARED_LIBS = ON
-- The C compiler identification is GNU 12.2.0
-- The CXX compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Building for architecture ARCH=x64
-- OPENTELEMETRY_ABI_VERSION_NO=1
-- OPENTELEMETRY_VERSION=1.14.2
-- Performing Test check_cxx_compiler_flag_-Wno-type-limits
-- Performing Test check_cxx_compiler_flag_-Wno-type-limits - Success
-- Performing Test check_cxx_compiler_flag_-Wno-deprecated-declarations
-- Performing Test check_cxx_compiler_flag_-Wno-deprecated-declarations - Success
-- Performing Test check_cxx_compiler_flag_-Wno-unused-parameter
-- Performing Test check_cxx_compiler_flag_-Wno-unused-parameter - Success
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Building WITH_STL=ON
-- Configuring done
-- Generating done
CMake Warning:
  Manually-specified variables were not used by the project:

    BUILD_BENCHMARK
    WITH_JAEGER
    WITH_ZPAGES


-- Build files have been written to: /home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release
opentelemetry-cpp/1.14.2: CMake command: cmake --build "/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release" '--' '-j8'
Scanning dependencies of target opentelemetry_common
Scanning dependencies of target opentelemetry_version
[  1%] Building CXX object sdk/src/version/CMakeFiles/opentelemetry_version.dir/version.cc.o
[  2%] Linking CXX shared library libopentelemetry_version.so
[  3%] Building CXX object sdk/src/common/CMakeFiles/opentelemetry_common.dir/random.cc.o
[  4%] Building CXX object sdk/src/common/CMakeFiles/opentelemetry_common.dir/core.cc.o
[  5%] Building CXX object sdk/src/common/CMakeFiles/opentelemetry_common.dir/global_log_handler.cc.o
[  6%] Building CXX object sdk/src/common/CMakeFiles/opentelemetry_common.dir/env_variables.cc.o
[  7%] Building CXX object sdk/src/common/CMakeFiles/opentelemetry_common.dir/base64.cc.o
[  8%] Building CXX object sdk/src/common/CMakeFiles/opentelemetry_common.dir/platform/fork_unix.cc.o
[  8%] Built target opentelemetry_version
In file included from /home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src/api/include/opentelemetry/nostd/variant.h:10,
                 from /home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src/sdk/src/common/core.cc:7:
/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src/api/include/opentelemetry/std/variant.h:17:12: error: ‘get_if’ has not been declared in ‘std’
   17 | using std::get_if;
      |            ^~~~~~
/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src/api/include/opentelemetry/std/variant.h:18:12: error: ‘monostate’ has not been declared in ‘std’
   18 | using std::monostate;
      |            ^~~~~~~~~
/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src/api/include/opentelemetry/std/variant.h:19:12: error: ‘variant_alternative_t’ has not been declared in ‘std’
   19 | using std::variant_alternative_t;
      |            ^~~~~~~~~~~~~~~~~~~~~
/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/src/api/include/opentelemetry/std/variant.h:23:22: error: ‘variant’ in namespace ‘std’ does not name a template type
   23 | using variant = std::variant<_Types...>;
      |                      ^~~~~~~
...
gmake[2]: *** [sdk/src/common/CMakeFiles/opentelemetry_common.dir/build.make:108: sdk/src/common/CMakeFiles/opentelemetry_common.dir/global_log_handler.cc.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:399: sdk/src/common/CMakeFiles/opentelemetry_common.dir/all] Error 2
gmake: *** [Makefile:149: all] Error 2
opentelemetry-cpp/1.14.2:
opentelemetry-cpp/1.14.2: ERROR: Package '404724448056678c693431e88e9c16cf349104bf' build failed
opentelemetry-cpp/1.14.2: WARN: Build folder /home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release
ERROR: opentelemetry-cpp/1.14.2: Error in build() method, line 257
        cmake.build()
        ConanException: Error 2 while executing cmake --build "/home/engel/.conan/data/opentelemetry-cpp/1.14.2/_/_/build/404724448056678c693431e88e9c16cf349104bf/build/Release" '--' '-j8'
```

</details>

steinerthomas:
Specify library name and version:  **opentelemetry-cpp/1.14.2**

Fixes #24229

---

### DIFF
--- a/recipes/opentelemetry-cpp/all/conanfile.py
+++ b/recipes/opentelemetry-cpp/all/conanfile.py
@@ -1,7 +1,7 @@
 from conan import ConanFile, conan_version
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.files import get, copy, rmdir, replace_in_file, save
-from conan.tools.build import check_min_cppstd
+from conan.tools.build import check_min_cppstd, default_cppstd
 from conan.tools.scm import Version
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
 from conan.tools.env import VirtualRunEnv, VirtualBuildEnv
@@ -24,7 +24,7 @@ class OpenTelemetryCppConan(ConanFile):
         "fPIC": [True, False],
         "shared": [True, False],
         "with_no_deprecated_code": [True, False],
-        "with_stl": [True, False],
+        "with_stl": [True, False, "default"],
         "with_gsl": [True, False],
         "with_abseil": [True, False],
         "with_otlp": ["deprecated", True, False],
@@ -69,9 +69,17 @@ class OpenTelemetryCppConan(ConanFile):
     def _min_cppstd(self):
         if self.options.with_abseil and Version(self.dependencies["abseil"].ref.version) >= "20230125":
             return 14
-        if self.options.with_stl:
-            return 17
         return 11
+
+    @property
+    def _default_cppstd(self):
+        return default_cppstd(self).replace("gnu", "")
+
+    @property
+    def _default_stl_version(self):
+        if self.options.with_stl == "default":
+            return f"CXX{self._default_cppstd}"
+        return self.options.with_stl
 
     @property
     def _compilers_minimum_version(self):
@@ -79,14 +87,6 @@ class OpenTelemetryCppConan(ConanFile):
             return {
                 "gcc": "6",
                 "clang": "5",
-                "apple-clang": "10",
-                "Visual Studio": "16",
-                "msvc": "192",
-            }
-        elif self._min_cppstd == 17:
-            return {
-                "gcc": "8",
-                "clang": "7",
                 "apple-clang": "10",
                 "Visual Studio": "16",
                 "msvc": "192",
@@ -218,7 +218,7 @@ class OpenTelemetryCppConan(ConanFile):
         tc.cache_variables["BUILD_BENCHMARK"] = False
         tc.cache_variables["WITH_EXAMPLES"] = False
         tc.cache_variables["WITH_NO_DEPRECATED_CODE"] = self.options.with_no_deprecated_code
-        tc.cache_variables["WITH_STL"] = self.options.with_stl
+        tc.cache_variables["WITH_STL"] = self._default_stl_version
         tc.cache_variables["WITH_GSL"] = self.options.with_gsl
         tc.cache_variables["WITH_ABSEIL"] = self.options.with_abseil
         if Version(self.version) < "1.10":
@@ -238,7 +238,7 @@ class OpenTelemetryCppConan(ConanFile):
         tc.cache_variables["WITH_METRICS_EXEMPLAR_PREVIEW"] = self.options.with_metrics_exemplar_preview
         tc.cache_variables["OPENTELEMETRY_INSTALL"] = True
         if not self.settings.compiler.cppstd:
-            tc.variables["CMAKE_CXX_STANDARD"] = self._min_cppstd
+            tc.variables["CMAKE_CXX_STANDARD"] = self._default_cppstd
         tc.generate()
 
         deps = CMakeDeps(self)
--- a/recipes/opentelemetry-cpp/all/conanfile.py
+++ b/recipes/opentelemetry-cpp/all/conanfile.py
@@ -69,6 +69,8 @@ class OpenTelemetryCppConan(ConanFile):
     def _min_cppstd(self):
         if self.options.with_abseil and Version(self.dependencies["abseil"].ref.version) >= "20230125":
             return 14
+        if self.options.with_stl:
+            return 17
         return 11
 
     @property
@@ -77,6 +79,14 @@ class OpenTelemetryCppConan(ConanFile):
             return {
                 "gcc": "6",
                 "clang": "5",
+                "apple-clang": "10",
+                "Visual Studio": "16",
+                "msvc": "192",
+            }
+        elif self._min_cppstd == 17:
+            return {
+                "gcc": "8",
+                "clang": "7",
                 "apple-clang": "10",
                 "Visual Studio": "16",
                 "msvc": "192",


recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -71,6 +71,16 @@ def _min_cppstd(self):
             return 14
         return 11
 
+    @property
+    def _default_cppstd(self):
+        return default_cppstd(self).replace("gnu", "")
```
steinerthomas, line 76: TODO: check why apple-clang fails to build with `default_cppstd` or add this:
```suggestion
        if self.options.with_stl == "default":
            return default_cppstd(self).replace("gnu", "")
        return self._min_cppstd
```

steinerthomas: @jcar87 `default_cppstd` for apple-clang13 profile returns `gnu98` - is this correct?

<details><summary>apple-clang13</summary>
<p>

```
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.libcxx=libc++
compiler.version=13.0
os=Macos
[options]
opentelemetry-cpp:shared=False
```

</p>
</details>

RubenRBS: Hi @steinerthomas, yes, this is correct. Unfortuntately apple-clang has not changed the default cppstd between versions, so the default cppstd for every version is still gnu98

steinerthomas: Hi @RubenRBS, thanks for your answer. Is there a utility where I can compare cppstd versions? This would be helpful here: if _min_cppstd greater than default_cppstd then return _min_cppstd  else return default_cppstd

AbrilRBS (dismissed):

RubenRBS:
We'll probably need to rebuild the missing binary before this can build :)

toge (approved):
Thanks a lot!

recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -71,6 +71,18 @@ def _min_cppstd(self):
             return 14
         return 11
 
+    @property
+    def _default_cppstd(self):
+        if self.options.with_stl == "default":
+            return default_cppstd(self).replace("gnu", "")
```
cynix, line 77: It seems this doesn’t match the `compiler.cppstd` setting.

recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -208,7 +220,7 @@ def generate(self):
         tc.cache_variables["BUILD_BENCHMARK"] = False
         tc.cache_variables["WITH_EXAMPLES"] = False
         tc.cache_variables["WITH_NO_DEPRECATED_CODE"] = self.options.with_no_deprecated_code
-        tc.cache_variables["WITH_STL"] = self.options.with_stl
+        tc.cache_variables["WITH_STL"] = self._default_stl_version
```
cynix, line 223: I think we need a `self.cpp_info.components[…].defines.append(f”OPENTELEMETRY_STL_VERSION=20{self._default_cppstd}”)` in `package_info()`

steinerthomas: Thanks I've added this line. Should we keep `self.cpp_info.components["opentelemetry_common"].defines.append("HAVE_CPP_STDLIB")` as it is or add a version check?

recipes/opentelemetry-cpp/all/conanfile.py
steinerthomas: Please explain.

steinerthomas:
I've appended the option `with_stl` so the standard can also be explicitly chosen and not only detected. Reason: Our profile defaulted to C++14 but our project where we use this package builds with C++20.

@AbrilRBS please review again. @franramirez688 please also review this PR.

steinerthomas:
> > I've appended the option with_stl so the standard can also be explicitly chosen and not only detected. Reason: Our profile defaulted to C++14 but our project where we use this package builds with C++20.
> 
> Shouldn’t it just use the `compiler.cppstd` setting (and make sure that’s included in the package ID)? If the package and the project using it are built with different #defines, there could be ODR issues?

Our profiles do not specify the cppstd setting (we are building many packages), because most of the time it doesn't matter which standard is used. Specifying the settings would mean that we have to rebuild all our packages AND build them for many different CXX versions...

cynix:
> Our profiles do not specify the cppstd setting (we are building many packages), because most of the time it doesn't matter which standard is used. Specifying the settings would mean that we have to rebuild all our packages AND build them for many different CXX versions...

That doesn’t sound right. If the cppstd doesn’t matter for a package, the recipe should delete it from `self.info.settings` when calculating the package ID so it doesn’t need to be rebuilt for different cppstd settings.

When you don’t specify a cppstd in your profile, presumably Conan is picking some default value based on your compiler or whatever, but that doesn’t mean it’s not specified — the package is still built with a particular cppstd value.

steinerthomas:
> That doesn’t sound right. If the cppstd doesn’t matter for a package, the recipe should delete it from `self.info.settings` when calculating the package ID so it doesn’t need to be rebuilt for different cppstd settings.
> 
> When you don’t specify a cppstd in your profile, presumably Conan is picking some default value based on your compiler or whatever, but that doesn’t mean it’s not specified — the package is still built with a particular cppstd value.
> 
I think you got me wrong. I know that there is a default value and since now it was ok for us to use the profiles default one. This is now our first library, where we have to specify the cppstd setting, which would mean we have to rebuild everything. Is this really intended?

> Actually I guess because you’re requiring the option to be set manually, this won’t happen. But still, there’s no good reason for it to differ from cppstd.
>
Exactly! This is why I also appended the option for explicitly specifying the CXX standard, not only detecting.

cynix:
> I think you got me wrong. I know that there is a default value and since now it was ok for us to use the profiles default one. This is now our first library, where we have to specify the cppstd setting, which would mean we have to rebuild everything. Is this really intended?

But why would you not use the default cppstd for this package too?

It still doesn’t feel like a good idea to allow packages built with different cppstd settings to be mixed.

> Exactly! This is why I also appended the option for explicitly specifying the CXX standard, not only detecting.

I don’t get why you’d ever set this option to a value that’s not the same as your cppstd setting, but I suppose you have your reasons. However, shouldn’t the default for this option be the same as the cppstd setting, for people who don’t need/want them to be different? That is surely the more common use case.

steinerthomas:
> > I think you got me wrong. I know that there is a default value and since now it was ok for us to use the profiles default one. This is now our first library, where we have to specify the cppstd setting, which would mean we have to rebuild everything. Is this really intended?
> 
> But why would you not use the default cppstd for this package too?
> 
> It still doesn’t feel like a good idea to allow packages built with different cppstd settings to be mixed.
> 
> > Exactly! This is why I also appended the option for explicitly specifying the CXX standard, not only detecting.
> 
> I don’t get why you’d ever set this option to a value that’s not the same as your cppstd setting, but I suppose you have your reasons. However, shouldn’t the default for this option be the same as the cppstd setting, for people who don’t need/want them to be different? That is surely the more common use case.

Sorry for the confusion you are right, it was just the missing definition `OPENTELEMETRY_STL_VERSION` which caused our problem.
I revered the previous commit, thanks for your review!

recipes/opentelemetry-cpp/all/conanfile.py
cynix: `default_cppstd()` returns some hardcoded default cppstd value based on the compiler version. See:

https://github.com/conan-io/conan/blob/51731ef38732b07a0b17f1523790b71d4d24839e/conan/tools/build/cppstd.py#L68

Which eventually calls:

https://github.com/conan-io/conan/blob/51731ef38732b07a0b17f1523790b71d4d24839e/conan/internal/api/detect_api.py#L299

cynix:
> Sorry for the confusion you are right, it was just the missing definition `OPENTELEMETRY_STL_VERSION` which caused our problem. I revered the previous commit, thanks for your review!

Seems I can’t “unresolve” a code comment. Please see an additional reply here: https://github.com/conan-io/conan-center-index/pull/24230/files#r1740952937

recipes/opentelemetry-cpp/all/conanfile.py
steinerthomas: Now I'm returning `compiler.cppstd` if set in profile.

steinerthomas: Should I remove the "default" value and return a Version specific value?

```
@property
    def _default_cppstd(self):
            return str(self.settings.compiler.cppstd).replace("gnu", "")
        else:
            return default_cppstd(self).replace("gnu", "")

    @property
    def _default_stl_version(self):
        if self.options.with_stl and Version(self.version) >= "1.12":
            return f"CXX{self._default_cppstd}"
        return self.options.with_stl
```

recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -71,6 +71,21 @@ def _min_cppstd(self):
             return 14
         return 11
 
+    @property
+    def _default_cppstd(self):
+        if self.options.with_stl == "default":
+            if self.settings.compiler.cppstd:
```
cynix, line 77: Might be better to use `self.settings.get_safe(“compiler.cppstd”)`?

steinerthomas: Ok, then we should also use it there?

https://github.com/conan-io/conan-center-index/blob/7848c1c8b0ab4097819cf4f93e901c01d66fddd8/recipes/opentelemetry-cpp/all/conanfile.py#L147

https://github.com/conan-io/conan-center-index/blob/7848c1c8b0ab4097819cf4f93e901c01d66fddd8/recipes/opentelemetry-cpp/all/conanfile.py#L230

recipes/opentelemetry-cpp/all/conanfile.py
steinerthomas: @cynix whats your opinion?

cynix: Do you mean just make the option True/False, and return the cppstd value when set to True? I think that’s fine too :)

AbrilRBS:
@steinerthomas could you check if the current approach works for you? I've simplified the approach as now we don't have to support Conan 1 anymore

jcar87:
Bear in mind that the original issue is to support Conan 1x  https://github.com/conan-io/conan-center-index/issues/24229
but as of this week it is no longer supported

recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -185,7 +184,7 @@ def _required_boost_components(self):
         return ["locale"] if self.options.get_safe("with_jaeger") else []
 
     def validate(self):
-        if self.settings.compiler.cppstd:
+        if self.settings.compiler.get_safe("cppstd"):
             check_min_cppstd(self, self._min_cppstd)
```
AbrilRBS, line 188:
```suggestion
        check_min_cppstd(self, self._min_cppstd)
```

AbrilRBS:
In fact, seems like we need a newer compiler in case of `with_stl=True`, as seen in https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-24230/11/package_build_logs/build_log_opentelemetry-cpp_1_17_0_e4a27418867a8a563233363e2c102cd1_0a22d92ade802089eff409be2e737bce98bbc7b3.txt

AbrilRBS (approved):

recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -72,6 +71,9 @@ class OpenTelemetryCppConan(ConanFile):
 
     @property
     def _min_cppstd(self):
+        if Version(self.version) < "1.12.0" and self.options.with_stl:
+            # Old versions require C++17 with STL
```
AbrilRBS, line 75: Forgot to mention, this comes from https://github.com/open-telemetry/opentelemetry-cpp/blob/v1.9.1/CMakeLists.txt#L128

steinerthomas:
> Bear in mind that the original issue is to support Conan 1x #24229 but as of this week it is no longer supported

It seems that with conan2 it's mandatory to specify the `compiler.cppstd`.

@jcar87 / @AbrilRBS Is this correct? Where can I find this in the docs?

jcar87:
> > Bear in mind that the original issue is to support Conan 1x #24229 but as of this week it is no longer supported
> 
> It seems that with conan2 it's mandatory to specify the `compiler.cppstd`.
> 
> @jcar87 / @AbrilRBS Is this correct? Where can I find this in the docs?

Hi @steinerthomas. Conan 2 autodetected profiles will set `compiler.cppstd`, and in Conan Center we assume that they follow this practice and are set. Outside of Conan Center, irrespective of the version of Conan, I'd say no setting is really mandatory - they're an arbitrary convention.

chausner:
We ran into this same issue (`OPENTELEMETRY_STL_VERSION` not being set with the current recipe). Any chance of getting this merged?

steinerthomas:
> We ran into this same issue (`OPENTELEMETRY_STL_VERSION` not being set with the current recipe). Any chance of getting this merged?

I resolved the merge conflict. Please review.

jcar87:
Please accept our apologies for the delay in reviewing this. 
This was pending a team review, and we were a bit confused about what bug this was fixing.

We have merged https://github.com/conan-io/conan-center-index/pull/28325 that includes a fix for the segfault when `with_stl=True`. 
There were additional issues: the relevant compiler flag was not being propagated by the same target as upstream, so we have tried to align the recipe with the most recent version as it is upstream.

Please note that since we now only support Conan 2, we assume that `compiler.cppstd` is defined.

> Our profiles do not specify the cppstd setting (we are building many packages), because most of the time it doesn't matter which standard is used. Specifying the settings would mean that we have to rebuild all our packages AND build them for many different CXX versions...
@steinerthomas - I would say that @cynix  comments were spot-on. This recipe is unusual in that it generates ABI-incompatible binaries depending on which c++ standard is used - so in Conan 1 this would've been a problem.

Please note that specifying the `compiler.cppstd` setting should *NOT* mean rebuilding all your packages for multiple CXX versions. Conan 2's default behaviour is to rely on the compatibility plugin (an implementation of which is bundled and setup upon installing Conan) - such that if you are building a project with say, `compiler.cppstd=20`, and have a dependency built with c++14, it will consider and use that binary - Conan will not force you to rebuild everything when suitably compatible binaries exist (although it should force you when binaries are incompatible across the value of compiler.cppstd=20 - it is the case with abseil and anything that depends on it, unsure if it is the case with opentelemetry-cpp if the compile definition is propagated correctly).

PR #24251, opened 2024-06-06
add recipe for dbus-cxx

ErikHons:
**dbus-cxx/2.5.1**

At NI we are using conan as our C++ packaging tool. We dbus-cxx for native C++ bindings to DBus with a path to Windows support. 

---

### DIFF
--- a/recipes/dbus-cxx/2.x.x/conanfile.py
+++ b/recipes/dbus-cxx/2.x.x/conanfile.py
@@ -1,0 +1,89 @@
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import cmake_layout, CMake, CMakeToolchain
+from conan.tools.gnu import PkgConfigDeps
+from conan.tools.files import get
+from conan.tools.scm import Version
+
+class DbusCXX(ConanFile):
+    name = "dbus-cxx"
+    license = "LGPL-3.0-only", "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "http://dbus-cxx.github.io"
+    description = "DBus-cxx provides an object-oriented interface to DBus"
+    topics = "bus", "interprocess", "message"
+    package_type = "static-library"
+    settings = "os", "compiler", "build_type", "arch"
+
+    options = {
+        "fPIC": [True, False],
+        "with_glib": [True, False],
+        "with_qt": [True, False],
+        "with_uv": [True, False],
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 17
+
+    # Default to a relocatable static library with only the default
+    # standalone dispatcher
+    default_options = {
+        "fPIC": True,
+        "with_glib": False,
+        "with_qt": False,
+        "with_uv": False,
+   }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+        if Version(self.version) < "2.5.0":
+            del self.options.with_uv
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+
+    def requirements(self):
+        self.requires("libsigcpp/3.0.7", transitive_headers=True)
+        self.requires("expat/[>=2.6.2 <3]")
+
+        if self.options.get_safe("with_uv"):
+            self.requires("libuv/[>=1.0 <2.0]")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["ENABLE_GLIB_SUPPORT"] = self.options.with_glib
+        tc.cache_variables["ENABLE_QT_SUPPORT"] = self.options.with_qt
+        if Version(self.version) >= "2.5.0":
+            tc.cache_variables["ENABLE_UV_SUPPORT"] = self.options.with_uv
+        tc.generate()
+        deps = PkgConfigDeps(self)
+        deps.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        if self.options.with_glib:
+            self.cpp_info.libs.append("dbus-cxx-glib")
+        if self.options.with_qt:
+            self.cpp_info.libs.append("dbus-cxx-qt")
+        if self.options.get_safe("with_uv"):
+            self.cpp_info.libs.append("dbus-cxx-uv")
+
+        self.cpp_info.libs.append("dbus-cxx")
+        self.cpp_info.includedirs = ['include/dbus-cxx-2.0']
--- a/recipes/dbus-cxx/2.x.x/test_package/CMakeLists.txt
+++ b/recipes/dbus-cxx/2.x.x/test_package/CMakeLists.txt
@@ -1,0 +1,8 @@
+cmake_minimum_required (VERSION 3.15)
+project (dbus_cxx_package_test CXX)
+
+find_package (dbus-cxx REQUIRED CONFIG)
+
+add_executable (test test.cpp)
+target_link_libraries (test dbus-cxx::dbus-cxx)
+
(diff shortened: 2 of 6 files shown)


RubenRBS:
Hi @ErikHons thanks a lot for taking the time to submit the recipe, we appreciate it :)

I was going to submit some cleanups for the recipe, but I realized that you didn't grant us permission to directly push, so here is the patch (Have no time to open as PR, sorry!)
```
diff --git a/recipes/dbus-cxx/2.x.x/conanfile.py b/recipes/dbus-cxx/2.x.x/conanfile.py
index 50d6b6324..ec1f5e99c 100644
--- a/recipes/dbus-cxx/2.x.x/conanfile.py
+++ b/recipes/dbus-cxx/2.x.x/conanfile.py
@@ -1,13 +1,14 @@
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
-from conan.tools.cmake import cmake_layout, CMake, CMakeToolchain
+from conan.tools.cmake import cmake_layout, CMake, CMakeToolchain, CMakeDeps
+from conan.tools.files import get
-import conan.tools.files
+
 
 class DbusCXX(ConanFile):
     name = "dbus-cxx"
-    license = "LGPL-3.0-only, BSD-3-Clause"
-    url = "https://github.com/dbus-cxx/dbus-cxx"
+    license = "LGPL-3.0-only", "BSD-3-Clause"
     homepage = "http://dbus-cxx.github.io"
     description = "DBus-cxx provides an object-oriented interface to DBus"
     topics = "bus", "interprocess", "message"
@@ -28,30 +29,34 @@ class DbusCXX(ConanFile):
         "glib_support": False,
         "qt_support": False,
         "uv_support": False,
-   }
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 17
 
     def config_options(self):
             del self.options.fPIC
+        if Version(self.version) < 2.5.0:
+            del self.options.uv_support
 
     def validate(self):
-        if self.options.uv_support and Version(self.version) < "2.5.0":
-            raise ConanInvalidConfiguration("UV support requires verion >= 2.5.0")
 
     def requirements(self):
-        self.requires("libsigcpp/[^3.0.7]", transitive_headers=True)
-        self.requires("expat/[^2.5.0]")
+        self.requires("libsigcpp/3.0.7", transitive_headers=True)
+        self.requires("expat/[>=2.6.2 <3]")
 
         if self.options.uv_support:
             self.requires("libuv/[>=1.0 <2.0]")
 
     def source(self):
-        conan.tools.files.get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def layout(self):
-
-    generators = "PkgConfigDeps"
+        cmake_layout(self)
 
     def generate(self):
         tc = CMakeToolchain(self)
@@ -60,6 +65,8 @@ class DbusCXX(ConanFile):
         if Version(self.version) >= "2.5.0":
             tc.cache_variables["ENABLE_UV_SUPPORT"] = self.options.uv_support
         tc.generate()
+        deps.generate()
 
     def build(self):
         cmake = CMake(self)

```

Some comments: 
* I've removed the `generators = "PkgConfigDeps"` and used `CMakeDeps` in the generate method, but this stumped me a bit. Why do you need pkgconfigdeps? Any insight on that?
* As per https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges I've pinned and changed the version ranges you added, except for libuv, which we're currently discussing on whether to allow it as version range or not, will let you know once we come to a conclusion
* Instead of validating bad versions depending on the version, it's better if we outright delete the option for those versions where it does not exist - We only need to take care of guarding with the same logic its access in the rest of the recipe, or using `self.options.get_safe("uv_support")

Also some more question:
- You have a mismatched config.yml and conandata.yml files, both files in this case should contain the same versions
- I see the qt_support option being defined, but not used in the recipe
- usually for options that toggle optional dependencies, the convention in CCI has been to call them `with_xxx`, so `uv_support` would be `with_uv`

ErikHons:
> Hi @ErikHons thanks a lot for taking the time to submit the recipe, we appreciate it :)

I love conan. Glad to give back!

> I was going to submit some cleanups for the recipe, but I realized that you didn't grant us permission to directly push, so here is the patch (Have no time to open as PR, sorry!)

Yeah. We run our corporate builds off that fork. I'm happy to apply the patch.

> Some comments:
> 
> * I've removed the `generators = "PkgConfigDeps"` and used `CMakeDeps` in the generate method, but this stumped me a bit. Why do you need pkgconfigdeps? Any insight on that?

The library uses pkg_check_modules() by default. It's supposed to use find_package() when a toolchain is specified bit that wasn't working for me. I'll take another look.

> I've pinned and changed the version ranges you added, except for libuv, which we're currently discussing on whether to allow it as version range or not, will let you know once we come to a conclusion

I don't grok how that all works. My main concern is that libuv 1.45 drops support for an old version of RedHat I have to build for still. I think I can override all that in the client build though right?

> Instead of validating bad versions depending on the version, it's better if we outright delete the option for those versions where it does not exist 

Makes sense!

> You have a mismatched config.yml and conandata.yml files, both files in this case should contain the same versions

Oopise. I'm working off a branch of 2.4.0 (we patch this recipe in our fork of the index). I added 2.5.1 for this PR and just missed this step.

> I see the qt_support option being defined, but not used in the recipe

That option sets the ENABLE_QT_SUPPORT cmake variable and adds the dbus-cxx-qt library to self.cpp_info.libs

> usually for options that toggle optional dependencies, the convention in CCI has been to call them `with_xxx`, so `uv_support` would be `with_uv`

Nice. I'll change it.

Thanks for the quick response!

ErikHons:
@RubenRBS the dbus-cxx top-level cmake contains stuff like:

```
if( CMAKE_TOOLCHAIN_FILE )
find_package( sigc++-3 REQUIRED )
else()
pkg_check_modules( sigc REQUIRED sigc++-3.0 )
endif( CMAKE_TOOLCHAIN_FILE )
```
Conan passes the `-DCMAKE_TOOLCHAIN_FILE="..."` parameter that should define that cmake variable but cmake keeps using the pkg-config path. I'm still looking at it.

recipes/dbus-cxx/2.x.x/conanfile.py
```diff
@@ -0,0 +1,89 @@
...
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+
+    def requirements(self):
+        self.requires("libsigcpp/3.0.7", transitive_headers=True)
+        self.requires("expat/[>=2.6.2 <3]")
+
+        if self.options.get_safe("with_uv"):
+            self.requires("libuv/[>=1.0 <2.0]")
```
RubenRBS, line 53: Allowing this after checking with the team

cf https://github.com/conan-io/conan-center-index/pull/24257

ErikHons:
Hey @RubenRBS, I see that the "Linux, GCC-legacy" action is failing on the gcc 7 build but not why. When I first created a docker image to test that case it failed because the pkg-config tools weren't installed. After adding the pkgconf package the build works.

Is it possible that your pipelines don't have the pkgconf package installed? If so, what do you recommend going forward: update CCI pipelines, patch dbus-cxx in the recipe,  something else?

recipes/dbus-cxx/2.x.x/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required (VERSION 3.16)
```
jcar87, line 1:
```suggestion
cmake_minimum_required (VERSION 3.15)
```

recipes/dbus-cxx/2.x.x/conanfile.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def layout(self):
+        cmake_layout(self)
```
AbrilRBS, line 59:
```suggestion
        cmake_layout(self, src_folder="src")
```

ErikHons:
All checks have passed!!!!!!! Woo!!!!

recipes/dbus-cxx/2.x.x/conanfile.py
```diff
@@ -0,0 +1,89 @@
...
+from conan.tools.scm import Version
+
+class DbusCXX(ConanFile):
+    name = "dbus-cxx"
+    license = "LGPL-3.0-only", "BSD-3-Clause"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "http://dbus-cxx.github.io"
+    description = "DBus-cxx provides an object-oriented interface to DBus"
+    topics = "bus", "interprocess", "message"
+    package_type = "static-library"
```
jcar87, line 15: is this correct? this seems to be unconditionally shared
https://github.com/dbus-cxx/dbus-cxx/blob/master/CMakeLists.txt#L250

franramirez688:
Superseded by https://github.com/conan-io/conan-center-index/pull/27142 (preserved all the original commits)

PR #24237, opened 2024-06-06
[boost] Add check for coroutine support for Cobalt build

uilianries:
Specify library name and version:  **boost/1.84.0**

Since Boost 1.84.0, the new library Cobalt was introduced and requires not only C++20, but also concepts and coroutines:

- https://github.com/boostorg/cobalt/blob/boost-1.85.0/build/Jamfile#L54
- https://github.com/boostorg/cobalt/blob/boost-1.85.0/include/boost/cobalt/concepts.hpp#L8

This PR adds a new check to validate the compiler version if has support to coroutines, based on https://en.cppreference.com/w/cpp/compiler_support#cpp20

fixes #24216

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24273, opened 2024-06-10
[boost] Add -fcoroutines for GCC10 only

uilianries:
Specify library name and version:  **boost/1.85.0**

The PR #24237 fixed a compatibility situation when using compiler that don't have C++20 fully supported with coroutines, so people can use Boost Cobalt as well.

When using GCC10, the coroutines support is available, but it requires passing `-fcoroutines`, as documented in https://gcc.gnu.org/gcc-10/changes.html. For GCC11 and later it's no longer required.

There was a mistake in the logic to add or not `-fcoroutines`, it uses `or`, but should `and`, to only be applied for GCC10.

The Clang project supports coroutines since version 14, and does not require `-fcoroutines`. 

This bug was originally reported in Discord: https://discord.com/channels/400588936151433218/764433875640123402/1249600810439147531

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -1395,7 +1395,7 @@ class BoostConan(ConanFile):
             cxx_flags.append(f"-DBOOST_STACKTRACE_ADDR2LINE_LOCATION={self.options.addr2line_location}")
 
         if not self.options.get_safe('without_cobalt', True) and \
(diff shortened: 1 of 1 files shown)


RubenRBS (dismissed):
Agh, sorry I did not catch it in my original review :)

perseoGI (approved):
LGTM thanks!

danimtb (approved):

PR #24292, opened 2024-06-12
octo-logger-cpp: add version 1.12.0, update dependencies

toge:
Specify library name and version:  **octo-logger-cpp/***

https://github.com/ofiriluz/octo-logger-cpp/compare/v1.11.0...v1.12.0

---

### DIFF
--- a/recipes/octo-logger-cpp/all/conandata.yml
+++ b/recipes/octo-logger-cpp/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.12.0":
+    url: "https://github.com/ofiriluz/octo-logger-cpp/archive/v1.12.0.tar.gz"
+    sha256: "bcf5fffebd478e9d1f192591dca44642f3ed3335b0d2ec4ea982ab75fbee62d7"
   "1.11.0":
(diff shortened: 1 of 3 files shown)


perseoGI (approved):
LGTM! Approving the PR even though it has bump dependencies on it which they seem to be not mandatory from upstream. But those new versions are more used on CCI so we cross fingers not to generate version conflicts to our users (͡° ͜ʖ ͡°)

jcar87:
> LGTM! Approving the PR even though it has bump dependencies on it which they seem to be not mandatory from upstream. But those new versions are more used on CCI so we cross fingers not to generate version conflicts to our users (͡° ͜ʖ ͡°)

Please note that this PR would actually introduce conflicts here:
https://github.com/conan-io/conan-center-index/blob/master/recipes/octo-keygen-cpp/all/conanfile.py#L41-L45

This is not a blocker for this PR - we can open a follow up PR to align `octo-keygen-cpp` around the fmt version such that it does not conflict.

jcar87 (approved):
LGTM, thanks @toge 

Will follow up with a PR to align fmt versions in `octo-encryption-cpp`

PR #24294, opened 2024-06-12
mold: add version 2.32.0

toge:
Specify library name and version:  **mold/2.32.0**

https://github.com/rui314/mold/compare/v2.31.0...v2.32.0

---

### DIFF
--- a/recipes/mold/all/conandata.yml
+++ b/recipes/mold/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "2.32.0":
+    url: "https://github.com/rui314/mold/archive/refs/tags/v2.32.0.tar.gz"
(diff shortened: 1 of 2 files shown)


perseoGI (commented):
Hi @toge, I've been researching a lit and found mold is compatible with arm64 -> apple sillicon so there is a check in the `validate` method that is not neccesary. Maybe we can remove that check before merging.

jcar87 (approved):

jcar87:
> Hi @toge, I've been researching a lit and found mold is compatible with arm64 -> apple sillicon so there is a check in the `validate` method that is not neccesary. Maybe we can remove that check before merging.

can be handled in a separate PR if newer versions of mold support apple silicon

perseoGI (approved):
Approving this PR, we will create another to give support for ARM64 architecture

PR #24310, opened 2024-06-13
quill: add version 4.4.0, remove older versions

toge:
Specify library name and version:  **quill/4.4.0**

https://github.com/odygrd/quill/compare/v4.3.0...v4.4.0

---

### DIFF
--- a/recipes/quill/all/conandata.yml
+++ b/recipes/quill/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "4.4.0":
+    url: "https://github.com/odygrd/quill/archive/v4.4.0.tar.gz"
+    sha256: "be166da93210df53aab34ece1bddd14b08506588186723b504ffeeec651bbf7c"
   "4.3.0":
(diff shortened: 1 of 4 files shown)


perseoGI (approved):
Thanks for the clean up Toge! ❤️

RubenRBS (commented):
Hi @toge thanks a lot!
Going by the diff, seems like newer versions of the library support building as shared. is this something you tried? Or should we move this to a different PR? Would be totally fair not to do that extra work now, you have done more than plenty enough already! :)

toge:
@RubenRBS 
I think it is an example to make quill a shared lib.
It seems not to be a part of library.

RubenRBS:
> I think it is an example to make quill a shared lib. 

But theh that configuration would be supported if I'm not misunderstanding something right?

Either way, that can way for a follow up PR where we add support for it if we end up wanting it, thanks a lot for adding the new version in the first place :)

RubenRBS (approved):

PR #24317, opened 2024-06-13
llvm-core 19.1.7

planetmarshall:
### Summary
Changes to recipe:  **llvm-core/19.1.7**

Depends on https://github.com/conan-io/conan-center-index/pull/22997

#### Motivation

Add The latest LLVM release

#### Details

* Add LLVM ~~19.1.6~~ 19.1.7
* Use graphviz parsing rather than CMake file parsing to resolve components, as suggested by comments in #22997

---


recipes/llvm-core/all/conanfile.py
```diff
@@ -1,429 +1,512 @@
...
-                alias_targets
-            )
+                    data["requires"].append(component)
+            return data
 
-            self._create_cmake_module_alias_targets(
-                os.path.join(self.package_folder, self._old_alias_module_file_rel_path),
-                old_alias_targets
-            )
+        cmake_exports = self.load(Path(self.package_folder) / "lib" / "cmake" / "llvm" / "LLVMExports.cmake")
```
valgur, line 383:
```suggestion
        cmake_exports = Path(self.package_folder, "lib", "cmake", "llvm", "LLVMExports.cmake").read_text("utf8")
```
`load(self, ...)` does not convert `\r\n` to `\n` on Windows, which can cause the regular expressions to fail unexpectedly.

I was bitten by this when applying a similar approach in the VTK recipe: https://github.com/conan-io/conan-center-index/blob/26a09d5b9c589206551669e75ebec365dff6368c/recipes/vtk/all/conanfile.py#L849-L853

recipes/llvm-core/all/conandata.yml
```diff
@@ -1,4 +1,14 @@
...
+  "19.1.7":
+    "llvm":
+      url: https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.7/llvm-19.1.7.src.tar.xz
+      sha256: 96f833c6ad99a3e8e1d9aca5f439b8fd2c7efdcf83b664e0af1c0712c5315910
+    "cmake":
+      url: https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.7/cmake-19.1.7.src.tar.xz
+      sha256: 11c5a28f90053b0c43d0dec3d0ad579347fc277199c005206b963c19aae514e3
+    "third-party":
+      url: https://github.com/llvm/llvm-project/releases/download/llvmorg-19.1.7/third-party-19.1.7.src.tar.xz
+      sha256: b96deca1d3097c7ffd4ff2bb904a50bdd56bec7ed1413ffb0d1d01af87b72c12
```
uilianries, line 11:
```suggestion
```

What is the case needed third-party? It only contains unittest and benchmark folders. Both are unrelated when consuming or packaging the project in most cases. The benchmark headers can be disabled by using the CMake definitions LLVM_INCLUDE_BENCHMARKS=False

recipes/llvm-core/all/conanfile.py
```diff
@@ -52,13 +53,75 @@
     "XCore"
 }
 
+def components_from_dotfile(dotfile):
```
uilianries, line 56: Why is it required only now? I saw your feature request in Conan client for Dot (Graphviz), still, this is a bit complex to maintain, so I would avoid it if possible. Boost is a good example of how opaque the recipe becomes after receiving customizations.

recipes/llvm-core/all/patches/19x/0000-cmake-dependencies.patch
```diff
@@ -0,0 +1,93 @@
...
++    list(APPEND CMAKE_REQUIRED_LIBRARIES ${ZLIB_LIBRARIES})
+     check_symbol_exists(compress2 zlib.h HAVE_ZLIB)
+     cmake_pop_check_state()
+     if(LLVM_ENABLE_ZLIB STREQUAL FORCE_ON AND NOT HAVE_ZLIB)
+@@ -232,11 +232,11 @@ if(NOT LLVM_USE_SANITIZER MATCHES "Memory.*")
+     # Skip libedit if using ASan as it contains memory leaks.
+     if (LLVM_ENABLE_LIBEDIT AND NOT LLVM_USE_SANITIZER MATCHES ".*Address.*")
+       if(LLVM_ENABLE_LIBEDIT STREQUAL FORCE_ON)
+-        find_package(LibEdit REQUIRED)
++        find_package(editline REQUIRED)
```
uilianries, line 19: Patching editline can be avoided with CMakeDeps directly: 

```
deps = CMakeDeps(self)
deps.set_property("editline", "cmake_file_name", "LibEdit")
deps.set_property("editline", "cmake_target_name", "LibEdit::LibEdit")
deps.set_property("editline", "cmake_additional_variables_prefixes", "LibEdit")
```

recipes/llvm-core/all/conanfile.py
```diff
@@ -137,6 +211,11 @@ def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
             del self.options.with_libedit  # not supported on windows
+        if Version(self.version) < 18:
```
uilianries, line 214: Do you have a reference about zstd inclusion? Could please add a comment in the recipe, I did not find a reference in https://releases.llvm.org/18.1.0/projects/libcxx/docs/ReleaseNotes/18.html

recipes/llvm-core/all/conanfile.py
```diff
@@ -261,11 +348,19 @@ def generate(self):
...
+        else:
+            cmake_variables["LLVM_ENABLE_ZSTD"] = "FORCE_ON" if self.options.get_safe("with_zstd") else False
+
         if self.options.targets != "all":
             cmake_variables["LLVM_TARGETS_TO_BUILD"] = self.options.targets
 
+        if is_msvc(self):
+            # https://discourse.llvm.org/t/failing-to-compile-llvm-on-windows-with-clang
+            cmake_variables["MSVC"] = True
+
```
uilianries, line 363:
```suggestion
```

I would not enforce this native cmake variable. Using clang on Windows has several flavours, msvc is only one case: https://blog.conan.io/2022/10/13/Different-flavors-Clang-compiler-Windows.html

uilianries (commented):
@planetmarshall Thank you for your PR, updating the llvm-core project.

I see this is not a simple bump version, but has several changes and misses details about the motivation behind these changes. Could please add more details, as this recipe is really complex and hard to maintain? Thank you!

recipes/llvm-core/all/conanfile.py
planetmarshall: See [this comment](https://github.com/conan-io/conan-center-index/pull/22997#discussion_r1725129187) from #22997 for the context behind this decision.

In summary, parsing CMake Graphviz output was how the original (Conan 1.x) LLVM recipe generated component dependencies - I removed this in favour of parsing CMake output, however in the above referenced comment it was suggested that Graphviz was the "least-worst" option, so I've added it back in here.

recipes/llvm-core/all/conandata.yml
planetmarshall: Removed.

recipes/llvm-core/all/patches/19x/0000-cmake-dependencies.patch
planetmarshall: Done

recipes/llvm-core/all/conanfile.py
planetmarshall: It's not explicit in the release notes. It's available since LLVM 15 - see https://reviews.llvm.org/D128465

recipes/llvm-core/all/conanfile.py
planetmarshall: Removed. This was fixed in https://github.com/llvm/llvm-project/pull/95505

recipes/llvm-core/all/conanfile.py
```diff
@@ -199,7 +281,12 @@ def validate_build(self):
                 raise ConanInvalidConfiguration("Shared Debug build is not supported on CCI due to resource limitations")
 
     def source(self):
-        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        sources = self.conan_data["sources"][self.version]
+        if Version(self.version) < 18:
```
uilianries, line 285:
```suggestion
        if Version(self.version) < 15:
```

recipes/llvm-core/all/conanfile.py
uilianries: Could you please add more detailed comments in the recipe, including the method description, and what is happening in the key steps? As I commented before, it not only increases the recipe's complexity but also makes its maintenance harder. More information will be welcome to understand it in future PRs.

recipes/llvm-core/all/patches/11x/0000-cmake-dependencies.patch
uilianries, line 1: In case needing to touch those patches because CMakeDeps.set_property was not able to fit the expected native names by llvm, then I would avoid using it. My intention was to avoid patching anything, using CMakeDeps to match what is expected by llvm project without using any patch.

uilianries (commented):
@planetmarshall Thank you for keeping this PR, I saw your last changes and CI is passing again. Still, I have some comments.

uilianries (commented):
AS this PR has several changes, I did a diff between the package llvm-core/13.0.0 in Conan Center and the produced by this PR:

[llvm-core-13.0.0.diff.txt](https://github.com/user-attachments/files/19484467/llvm-core-13.0.0.diff.txt)

The artifacts are still there, no missing library/header. Some configuration changed, or just moved the position, which is fine IMO.

uilianries (approved):
@planetmarshall THank you for updating this recipe, it's more clear now how the dot file is working to list llvm dependencies.

It took a while to be reviewed, most because this recipe not updates a version, but also changes the way of processing the llvm dependencies. Please, consider doing separated PRs in the future, when having more than one topic in the changes, so it may be faster to have merged. Regards!

jcar87:
Merging this as to not delay this any further - but I don't understand why `PurePosixPath` is being used compared to what it was before, for example here:

```
    def _package_folder_path(self):
        return PurePosixPath(self.package_folder)
```

Please sympathise with the fact that we have a high volume of pull requests to review. This recipe is bringing back an implementation that was removed here https://github.com/conan-io/conan-center-index/pull/22997 -  and both versions of this implementation are far from trivial and difficult to review.

I would very strongly advise that if there is any sense of urgency in merging a PR, to limit the scope of the changes to only what's necessary to achieve the goal of the PR (e.g. adding the new version) - and if there _are_ changes that are not trivial to a casual reviewer (e.g PosixPath), please explain them. We tend to limit the use of "pure python" imports to make sure recipes remain compatible irrespective of version of Python.

PR #24326, opened 2024-06-14
[config] Enable github status checks for conan v2 profiles

danimtb:
### Summary
This enables the status checks for conan v2 and disables the conan v1 one.

#### Motivation
As the conan v2 builds gain importance, it is better feedback to the user to show the conan v2 builds in the status checks than the ones in conan 1. As in conan 2 there are less configurations built than in conan 1, this will also help on reducing the noise in the PRs.

#### Details
See this as an example of a PR with a recipe that has just one version:
image

---

### DIFF
--- a/.c3i/config_v1.yml
+++ b/.c3i/config_v1.yml
@@ -54,7 +54,7 @@ tasks:
   cci:
     conan_v2_run_export: false
     write_comments: true
-    detailed_status_checks: true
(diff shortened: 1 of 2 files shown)


RubenRBS (approved):

uilianries (approved):
LGTM. Plus, it will result in less noise as we only have few profiles in Conan 2.x

PR #17609, opened 2023-05-18
[APR] Support for cross-building using cache file

swebb2066:
The APR configure.in script makes extensive (30 instances) use of AC_TRY_RUN (to determine system capabilities) and checks for /dev/zero. These runtime checks only work when run on the host so the configure script will fail when cross compiling unless the relevant configuration variables  are provided in a cache file or on the command line or in an environment variable.

The configuration variable values are most easily determined by running the configure script on a host system using:
         `./configure --cache-file={gnu_host_triplet}.cache`
 and then removing the 'precious' autoconf variables:
          `ac_cv_env_*`

The generated cache file can be repeatedly used to cross-compile to the targeted host system by including it with the recipe data.
The default targeted host system name can be overriden in the profile using (for example).
 ```
[conf]
tools.gnu:host_triplet=arm-linux-gnueabihf
```
This PR assumes the generated cache file is stored in the 'cross_build' directory.

Using the command:
`conan create . apr/1.7.4@ -pr:h arm-linux-gnueabihf -pr:b default`
on the *master* branch conanfile.py with the validate() removed results in:
```
Exporting package recipe
apr/1.7.4 exports: File 'conandata.yml' found. Exporting it...
apr/1.7.4 exports: Copied 1 '.yml' file: conandata.yml
apr/1.7.4: Calling export_sources()
apr/1.7.4: A new conanfile.py version was exported
apr/1.7.4: Folder: /home/stephen/.conan/data/apr/1.7.4/_/_/export
apr/1.7.4: Source folder is corrupted, forcing removal
apr/1.7.4: Exported revision: 8068d417833d1980a50813734ce2fac8
Configuration (profile_host):
[settings]
arch=armv7
arch_build=x86_64
build_type=Release
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=8
os=Linux
os_build=Linux
[options]
boost:bzip2=False
boost:shared=False
boost:zlib=False
log4cxx:shared=True
yaml-cpp:shared=False
[build_requires]
[env]
AR=arm-linux-gnueabihf-ar
AS=arm-linux-gnueabihf-as
CC=arm-linux-gnueabihf-gcc
CFLAGS=-I/opt/xtools/arm-linux-gnueabihf/usr/include
CONAN_CFLAGS=-I/opt/xtools/arm-linux-gnueabihf/usr/include -DLINUX -D_GNU_SOURCE
[...]

PR #24329, opened 2024-06-14
[POC][APR] Support for cross-building using configuration variables

franramirez688:
### Summary
Changes to recipe:  **apr/[*]**

#### Motivation
Coming from: https://github.com/conan-io/conan-center-index/pull/17609

This PR does not provide pre-built cached files, it offers some common flags to let the machine configure the APR library when cross-building.

#### Details

* Added configuration flags to bypass some common errors trying to cross-compile APR.
* Added user conf variable `user.apr:cache_file` to pass any pre-built cached file.

cc @swebb2066

### DIFF
--- a/recipes/apr/all/conanfile.py
+++ b/recipes/apr/all/conanfile.py
@@ -70,11 +70,11 @@ class AprConan(ConanFile):
 
     def validate_build(self):
         if cross_building(self) and not is_msvc(self):
-            # Conan provide for Windows and Linux a simple "hack" to avoid entering a pre-built cached file
-            if self.settings.os not in ("Windows", "Linux") and self.conf.get("user.apr:cache_file") is None:
+            # Conan provides for Linux some configuration flags to avoid entering a pre-built cached file
+            if self.settings.os != "Linux" and self.conf.get("user.apr:cache_file") is None:
                 raise ConanInvalidConfiguration("apr recipe doesn't support cross-build for all the platforms"
                                                 " due to runtime checks in autoconf. You can provide"
-                                                " a cached file via Conan conf: \n"
+                                                " a pre-built cached file via Conan conf: \n"
                                                 " [conf]\nuser.apr:cache_file=/path/to/cache_file to try it.")
 
     def build_requirements(self):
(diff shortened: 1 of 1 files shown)


recipes/apr/all/conanfile.py
```diff
@@ -83,6 +89,56 @@ def build_requirements(self):
...
+        configure_args = []
+        user_cache_file = self.conf.get("user.apr:cache_file", check_type=str)
+        if user_cache_file:
+            configure_args.append(f"--cache-file={user_cache_file}")
+            return configure_args
+
+        self.output.warning("Trying to set some configuration arguments, but it"
+                            " could fail. The best approach is to provide a"
+                            " pre-built cached file.")
+        # Let's try this hack (tested on Linux ARM and Intel)
```
jcar87, line 125: not sure I'd call this a hack,  I think passing these variables to the configure script is valid and for example is documented in the changelog for this very purpose, but i could be wrong:
https://github.com/apache/apr/commit/d484a5ba10d4ea4e623b3e3969d3fdf54791e7e6

recipes/apr/all/conanfile.py
```diff
@@ -83,6 +89,56 @@ def build_requirements(self):
...
+        if self.settings.os == "Linux":
+            # The following are known to be true in modern Linux
+            configure_args.extend(["apr_cv_mutex_robust_shared=yes",
+                                   "ac_cv_file__dev_zero=yes",
+                                   "ac_cv_mmap__dev_zero=yes",
+                                   "ac_cv_define_PTHREAD_PROCESS_SHARED=yes",
+                                   "apr_cv_process_shared_works=yes",
+                                   "apr_cv_tcp_nodelay_with_cork=yes"])
+        elif self.settings.os == "Windows":
+            configure_args.extend(["apr_cv_mutex_robust_shared=no"
```
jcar87, line 135: I'm not 100% I would include windows in this - these values are true for mingw, but I'm not sure they're true for other scenarios (like pure msvc runtime) - for example, I think Windows+gcc may support posix threads but Windows+msvc may not  (not entirely familiar with the specifics).

recipes/apr/all/conanfile.py
```diff
@@ -83,6 +89,56 @@ def build_requirements(self):
...
+        """
+        The APR configure.in script makes extensive (30 instances) use
+        Check warning on line 73 in recipes/apr/all/conanfile.py
+
+        String statement has no effect of AC_TRY_RUN (to determine system capabilities) and
+        checks for /dev/zero. These runtime checks only work when run on the host so
+        the configure script will fail when cross compiling unless the relevant
+        configuration variables are provided in a cache file or on the command line or in an
+        environment variable. The configuration variable values are most easily determined by
+        running the configure script on a host system using:
```
jcar87, line 102: I think the issue is that apr build scripts perform some checks that don't have a fallback when cross-building, and the values need to provided at ./configure time with knowledge of the target system.  I'd try to make this a bit more "compact" of an explanation, but the functionality is good!

recipes/apr/all/conanfile.py
```diff
@@ -83,6 +89,56 @@ def build_requirements(self):
...
+        self.output.warning("Trying to set some configuration arguments, but it"
+                            " could fail. The best approach is to provide a"
+                            " pre-built cached file.")
+        # Let's try this hack (tested on Linux ARM and Intel)
+        if self.settings.os == "Linux":
+            # The following are known to be true in modern Linux
+            configure_args.extend(["apr_cv_mutex_robust_shared=yes",
+                                   "ac_cv_file__dev_zero=yes",
+                                   "ac_cv_mmap__dev_zero=yes",
+                                   "ac_cv_define_PTHREAD_PROCESS_SHARED=yes",
```
jcar87, line 131: Surprised this check needs any runtime anything, fairly certain it just checks that the `PTHREAD_PROCESS_SHARED` macro is defined in pthread.h - this seems part of Posix and I can't find examples of this not being defined on Linux. Very much not the case on windows though

recipes/apr/all/conanfile.py
```diff
@@ -83,6 +89,56 @@ def build_requirements(self):
...
+                                   "apr_cv_process_shared_works=yes",
+                                   "apr_cv_tcp_nodelay_with_cork=yes"])
```
jcar87, line 133: tcpcork is available on Linux since kernel 2.2 released in 1999
https://linux.die.net/man/7/tcp

So I think we can also assume that this is available - the check is more useful keeping in mind that apr is quite old and aims to support things like solaris, sunOS, etc

jcar87 (approved):

uilianries (approved):

RubenRBS (approved):

PR #24336, opened 2024-06-14
Exposing the exceptions feature of tomlplusplus

### LINKED ISSUE #24304: [request] tomlplusplus options
### Package Name/Version

tomlplusplus/3.4.0

### Changelog

https://github.com/marzer/tomlplusplus/releases/tag/v3.4.0

I'd like the `tomlplusplus` recipe to support the pre-processor options available in tomlplusplus.

xoorath:
### Summary
Changes to recipe: Expose the exceptions feature of tomlplusplus as [discussed with the author](https://github.com/marzer/tomlplusplus/issues/227).

Resolves #24304

#### Motivation
Users of tomlplusplus may wish to disable exception support on the library without disabling it on their project as a whole.

#### Details
When the `exceptions` options is set to `True` or `False` that will define `TOML_EXCEPTIONS` to 1 or 0. When it is set to `None` (the default) no define is set.

---


recipes/tomlplusplus/all/conanfile.py
```diff
@@ -73,3 +79,6 @@ def package_info(self):
         self.cpp_info.set_property("cmake_target_name", "tomlplusplus::tomlplusplus")
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
+        if self.options.exceptions.value is not None:
+            define_value = "1" if self.options.exceptions is True else "0"
+            self.cpp_info.defines.append(f"TOML_EXCEPTIONS={define_value}")
```
RubenRBS, line 84: Seems like upstream nicely handles the case where this is hardcoded to 1 but the compiler does not have exception support enabled :)

RubenRBS (approved):
Thanks a lot for taking the time to add this option, we appreciate it :)

memsharded (approved):

elliotcmorris:
Just FYI for anyone after me, I believe this change may cause some test breakages, as it did with [ours](https://github.com/OpenAssetIO/OpenAssetIO/actions/runs/9561321694/job/26355268327), as tomlplusplus seemed not to be throwing anymore.
Not too much of a problem, can now use this option to turn exceptions back on.

memsharded:
Hi @elliotcmorris

Maybe you delete one comment? I was checking it from an email notification, but I couldn't find it here. I have double checked the changes in this PR, in theory it shouldn't be breaking, please let me know otherwise.

elliotcmorris:
Yeah I did, I was wavering a bit in my confidence on my diagnosis.

It seems to us that this update has caused exceptions to be disabled regardless of provided argument. Will follow with a link to some test results shortly.

Not quite sure why this is, perhaps something to do with checking the option for identity (ie using `.value` in one case and not in the other?)

memsharded:
I see what you mean, it is definitely possible that using ``if self.options.get_safe("exceptions") is not None:`` would be a better approach and could avoid those issues. 

We'll do a PR to fix this asap. Thanks very much for checking this!

elliotcmorris:
That's awesome.

Just for clarity, we checked all three inputs, `None`, `True` and `False`, and couldn't get exceptions to be enabled in any case.

RubenRBS:
I've submitted https://github.com/conan-io/conan-center-index/pull/24385 to fix this issue, sorry about the issues it caused, we'll work on automated checks for these situations not to arise in the future :)

(@elliotcmorris I see you had to pin the recipe revision directly in your recipe, and I had one question for some further insight: Is your non-use of a lockfile an explicit design decission? if so I'd appreciate any insight into why if you could, thanks!)

xoorath:
Thanks all! I don't check github daily. I appreciate identifying and fixing the issue with checking None here! 

I did a bunch of testing locally and did not experience that same problem - I don't know why, but this seems like a more reliable way to do it anyways.

elliotcmorris:
@RubenRBS Not really explicit no. We don't run lockfiles generally due to a vague desire to keep on the bleeding edge of changes. We'd rather deal with issues like this every few months than let compatibility issues build.

As for why I didn't use a lockfile to do the mitigation in the short-term, that's honestly just because I knew how to do the pin and not how to make a lockfile.

Issue #24327, opened 2024-06-14
Prioritization of new recipe PRs


ErniGH:
The objective of this issue is to make the new recipe PRs in Conan Center Index visible with the aim of prioritising them for resolution. The criteria for this order are somewhat arbitrary, as they only take into account the project's stars.

|Name|PRs|Stars|Additional Info|
|----|---|-----|---------------|
|git|#23828|50700|It is an app|
|xgboost|#23673|25700||
|libphonenumber|#23689|16100||
|hermes|#24010|9500||
|colmap|#23827|7000||
|osrm|#23734|6100||
|wxwidgets|#21353|5800||
|matplotlib-cpp|#22256|4200||
|minhook|#22165|4100||
|matplotplusplus|#22257|4000||
|imgui-node-editor|#21650|3400|The Pr have two libs|
|dxc|#18931|3000||
|g2o|#22337|3000||
|optick|#20546|2900||
|compute_library|#19958|2700||
|pangolin|#23782|2300||
|userver|#20712|2300||
|blis|#23800|2200||
|nanobind|#20297|2100||
|simd|#20909|2000||
|criterion|#16711|1900||
|perl|#24078|1900|It is an app|
|audiowaveform|#19695|1800||
|picohttpparser|#19034|1800||
|tiledb|#19381|1800||
|usearch|#22931|1800||
|casadi|#22468|1600||
|vlfeat|#23214|1600||
|kompute|#20085|1500||
|libtomcrypt|#22113|1500||
|lapack|#23798|1400||
|coin-ipopt|#22467|1300||
|imguifiledialog|#21650|1200|The Pr have two libs|
|libvncserver|#20460|1100||
|nvidia-vaapi-driver|#23089|1100|The PR have four libs|
|suitesparse-amd|#23541|1100||
|suitesparse-btf|#23542|1100||
|suitesparse-camd|#23543|1100||
|suitesparse-ccolamd|#23544|1100||
|suitesparse-cholmod|#23553|1100||
|suitesparse-colamd|#23545|1100||
|suitesparse-config|#23540|1100||
|suitesparse-cxsparse|#23546|1100||
|suitesparse-graphblas|#23320|1100||
|suitesparse-klu|#23547|1100||
|suitesparse-lagraph|#23548|1100||
|suitesparse-ldl|#23549|1100||
|suitesparse-mongoose|#23550|1100||
|suitesparse-paru|#23555|1100||
|suitesparse-rbio|#23551|1100||
[...]

PR #24339, opened 2024-06-15
thorvg: add version 0.13.8, fix homepage

toge:
### Summary
Changes to recipe:  **thorvg/0.13.8**

#### Motivation

- thorvg/0.13.8 has several improvements.
- fix url of homage

#### Details

https://github.com/thorvg/thorvg/compare/v0.13.7...v0.13.8

---

### DIFF
--- a/recipes/thorvg/all/conanfile.py
+++ b/recipes/thorvg/all/conanfile.py
@@ -20,7 +20,7 @@ class ThorvgConan(ConanFile):
     description = "ThorVG is a platform-independent portable library that allows for drawing vector-based scenes and animations."
     license = "MIT"
     url = "https://github.com/conan-io/conan-center-index"
-    homepage = "https://www.thorvg.org/"
+    homepage = "https://github.com/thorvg/thorvg"
     topics = ("svg", "animation", "tvg")
     package_type = "library"
     settings = "os", "arch", "compiler", "build_type"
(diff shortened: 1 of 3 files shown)


recipes/thorvg/all/conanfile.py
```diff
@@ -20,7 +20,7 @@ class ThorvgConan(ConanFile):
     description = "ThorVG is a platform-independent portable library that allows for drawing vector-based scenes and animations."
     license = "MIT"
     url = "https://github.com/conan-io/conan-center-index"
-    homepage = "https://github.com/thorvg/thorvg"
+    homepage = "https://www.thorvg.org/"
```
RubenRBS, line 23: By having the homepage be the git(hub) repository, automated tools can be used to check for new releases/information about the PRs/recipe in a easier way, I'd recommend keeping it to point to the repositories where possible. 

Wdyt @toge?

See for example https://github.com/conan-io/conan-center-index/issues/24327 which would have been pretty hard to generate without it

RubenRBS (commented):
Thanks! Minor comment before approving

recipes/thorvg/all/conanfile.py
toge: @RubenRBS 
I agree with you.
I think it is very important to be able to automatically detect the version upgrades of a huge number of recipes.


By the way, as a reverse of this PR process, is it necessary to change recipes whose homepage is not github?

RubenRBS (approved):

Issue #24321, opened 2024-06-13
[package] <readline>/<8.2>: Need to disable examples in build


paigeadelethompson:
### Description

Readline's make refers to an example source file that doesn't exist: 

```
--- install ---
install: /opt/me/home/.conan2/p/b/readla0dd35098c4a2/b/src/examples/rltest2.c: No such file or directory
*** [install] Error code 71

make[1]: stopped in /opt/me/home/.conan2/p/b/readla0dd35098c4a2/b/build-release/examples
1 error

make[1]: stopped in /opt/me/home/.conan2/p/b/readla0dd35098c4a2/b/build-release/examples
--- install-pc ---
/usr/bin/install -c -m 644 /opt/me/home/.conan2/p/b/readla0dd35098c4a2/b/build-release/history.pc /opt/me/home/.conan2/p/b/readla0dd35098c4a2/p//lib/pkgconfig/history.pc

ERROR: readline/8.2: Error in package() method, line 91
	autotools.install()
	ConanException: Error 6 while executing
```
from: `https://git.savannah.gnu.org/git/readline.git`
```
git checkout readline-8.2
M	configure
Note: switching to 'readline-8.2'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at f7a382f readline-8.2 distribution sources and documentation
[me@me ~/readline]$ grep -r "rltest2.c"
./examples/Makefile.in:		rltest2.c rl-callbacktest.c hist_erasedups.c hist_purgecmd.c \
./examples/Makefile.in:rltest2.o: rltest2.c
./examples/Makefile:		rltest2.c rl-callbacktest.c hist_erasedups.c hist_purgecmd.c \
./examples/Makefile:rltest2.o: rltest2.c
[me@me ~/readline]$ find . -name "rltest2.c"
[me@me ~/readline]$ 

```

there's an option for disabling examples: 

```
 ./configure --help 2>&1 | grep examples
  --disable-install-examples
                          don't install examples [[default=install]]
```

Maybe add that to the `conanfile.py`: 
```
[...]

PR #24322, opened 2024-06-14
Add examples option to readline build (default disables installation of examples, fixes conan-io/conan-center-index/issues/24321)

paigeadelethompson:
### Summary
Changes to recipe:  **readline/[all]**

#### Motivation
Described in https://github.com/conan-io/conan-center-index/issues/24321

#### Details
Adds an option (defaults to True) for installation on examples (currently broken in https://git.savannah.gnu.org/git/readline.git master as well as readline-8.2 which is the latest used by Conan)

---
- https://github.com/conan-io/conan-center-index/issues/24321

PR #24350, opened 2024-06-16
readline: Disable example installation

RubenRBS:
### Summary
Changes to recipe:  **readline/[*]**

#### Motivation

The examples were always installed, but newer versions have a typo that made them fail.
Closes https://github.com/conan-io/conan-center-index/issues/24321

#### Details

This is a follow-up/replacement of https://github.com/conan-io/conan-center-index/pull/24322 with express authorization from the author https://github.com/conan-io/conan-center-index/pull/24322#issuecomment-2166999621 - The changes are different so no cherry-pick/CLA signing is necessary on their part.

I would appreciate if I could get some extra insight into whether keeping the examples could be useful for someone - afaik it does not install any heads/libs, so it's purely documentation on our end, (which unless there's a good reason, keeping those is not what we've been doing with the rest of libraries!) @paigeadelethompson

### DIFF
--- a/recipes/readline/all/conanfile.py
+++ b/recipes/readline/all/conanfile.py
@@ -11,6 +11,7 @@ from conan.errors import ConanInvalidConfiguration
 
 required_conan_version = ">=1.53.0"
 
+
 class ReadLineConan(ConanFile):
     name = "readline"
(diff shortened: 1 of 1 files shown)


uilianries (approved):
LGTM.

perseoGI (approved):
LGTM 🚀

PR #24364, opened 2024-06-17
[beauty] Fix Openssl dependency management 

### LINKED ISSUE #24363: [bug] Openssl enable problem in beauty/1.0.3 via Conan(inconistent conan files)
### Describe the bug

Hello,

To use openssl in the newer versions of beauty,

"BEAUTY_ENABLE_OPENSSL" flag needs to be set.

As a default, in conanfile(https://github.com/dfleury2/beauty/blob/1.0.3/conanfile.py) it is already set as "True"

but,
when i try to use via conan it never sets, also if you try to set it manually it gives the error at the below,

ERROR: beauty/1.0.3: option 'openssl' doesn't exist Possible options are ['shared', 'fPIC'] CMake Error at cmake/conan.cmake:826 (message): Conan install failed='1'

When i deeply investigated i realized that "openssl" option is not exists in conan-center conan file.

https://github.com/conan-io/conan-center-index/blob/master/recipes/beauty/all/conanfile.py

So you can never enable openssl in beauty if you use it from conan-center.

There is an incosistency between original conan file and conan-center conan file.

Best Regards,

### How to reproduce it

_No response_

uilianries:
### Summary
Changes to recipe:  **beauty/1.0.3**

#### Motivation

The current recipe uses OpenSSL, but does pass the option to activate the feature.

#### Details

- Added version 1.0.4 (Support for Boost 1.85.0): https://github.com/dfleury2/beauty/issues/30
- Downgraded required C++ version to 17: https://github.com/dfleury2/beauty/blob/master/CMakeLists.txt#L9
- Add option for OpenSSL: https://github.com/dfleury2/beauty/blob/master/CMakeLists.txt#L28

/cc @barbacar

fixes #24363

---

### DIFF
--- a/recipes/beauty/all/conandata.yml
+++ b/recipes/beauty/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.0.4":
+    url: "https://github.com/dfleury2/beauty/archive/refs/tags/1.0.4.tar.gz"
+    sha256: "8bfd3718470bdbed05b6b9253543d6ab9b33e21e25028e53fa513c67893a39f3"
   "1.0.3":
     url: "https://github.com/dfleury2/beauty/archive/refs/tags/1.0.3.tar.gz"
     sha256: "1707b0974537cfd8d1d2fd0b5accd5a662c0b408042aa01384cf5840ecc43594"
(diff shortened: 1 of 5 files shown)


RubenRBS (approved):
Greast work @uilianries :)

PR #21951, opened 2023-12-26
libpq: bump versions

valgur:
https://www.postgresql.org/about/news/postgresql-161-155-1410-1313-1217-and-1122-released-2749/

PR #24380, opened 2024-06-18
libpq: Update to latest release libpq/16.3 (PostgreSQL)

gsantner:
### Summary
Changes to recipe:  **libpq** - added version 16.3

#### Motivation
Make latest major version of libpq available at Conan.

#### Details
Added the latest release download. Also noticed the mingw patch. Tried to manually apply it with `patch` and it applied the diff without problem here (patch is by Conan and not upstreamed). Didn't work at Conan somehow so exported the patch for 16.3 as own patch file again.

---

### DIFF
--- a/recipes/libpq/all/conandata.yml
+++ b/recipes/libpq/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "16.3":
+    url: "https://ftp.postgresql.org/pub/source/v16.3/postgresql-16.3.tar.bz2"
+    sha256: "331963d5d3dc4caf4216a049fa40b66d6bcb8c730615859411b9518764e60585"
   "15.5":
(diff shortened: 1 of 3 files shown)


gsantner:
Found also PR <https://github.com/conan-io/conan-center-index/pull/21951> albeit it adds a older version (16.2). Maybe additional fixes are required and it's easier to get that PR into shape. However might just as well be that the CI builds succeed right away for 16.3, let's see 😄 .

gsantner:
>checking for icu-uc icu-i18n... no
> configure: error: ICU library not found
> If you have ICU already installed, see config.log for details on the
> failure.  It is possible the compiler isn't looking in the proper directory.
> Use --without-icu to disable ICU support.

Closing as I understand now why the other PR  #21951 has so much additional changes that I didn't consider needed. They are as it not build successful otherwise/or requires explicit changes.

Issue #23421, opened 2024-04-08
[question] Shared building issues when linking on Linux (due to runpath/LD_LIBRARY_PATH issue)?


irieger:
### What is your question?

Hey,

I some time ago run into a problem building several shared libraries of larger projects building more than one artifact. The problem I run into - after hours of debugging - I think I condensed down to a core issue, that I'm trying to understand.

So the question is basically: How does shared linking and LD_LIBRARY_PATH / CMAKE_LIBRARY_PATH work?

To describe the problem on a concrete example, I can reference this PR: https://github.com/conan-io/conan-center-index/pull/23112

Here I try to give a brief example CMakeLists.txt excerpt, that roughly I think describes the problem:
```
...
find_package(minizip-ng)
...

add_library(a ...)
target_link_libraries(a PRIVATE  minizip-ng)

add_executable(exe ...)
target_link_libraries(exe  PRIVATE a)
```

Running `conan create ...` with this CMake results in a problem when linking target `exe`. Here is the corresponding output:
```
/usr/bin/ld: /root/.conan2/p/b/miniz3c27d6cd524ad/p/lib/libminizip-ng.so.4: undefined reference to `libiconv_open'
/usr/bin/ld: /root/.conan2/p/b/miniz3c27d6cd524ad/p/lib/libminizip-ng.so.4: undefined reference to `libiconv_close'
/usr/bin/ld: /root/.conan2/p/b/miniz3c27d6cd524ad/p/lib/libminizip-ng.so.4: undefined reference to `libiconv'
```

Doing some analysis with `ldd`, `readelf -d` etc. I found that liba.so (in the real case `libOpenColorIO.so.2.3.2`) finds all it dependencies but libiconv.so.2. Output of ldd:
```
        linux-vdso.so.1 (0x00007fffb1bde000)
        libexpat.so.1 => /root/.conan2/p/b/expatb2443c2409e5b/p/lib/libexpat.so.1 (0x0000772c128db000)
        libImath-3_1.so.29 => /root/.conan2/p/b/imath93646a610b69b/p/lib/libImath-3_1.so.29 (0x0000772c12888000)
        libpystring.so => /root/.conan2/p/b/pystread4a2c558974/p/lib/libpystring.so (0x0000772c12873000)
        libyaml-cpp.so.0.8 => /root/.conan2/p/b/yaml-8766cd17dd26d/p/lib/libyaml-cpp.so.0.8 (0x0000772c127ee000)
[...]

PR #24395, opened 2024-06-19
OpenColorIO: Fix validation & package_info

irieger:
### Summary
Changes to recipe:  **OpenColorIO/all**

#### Motivation
Ensure more configurations options are properly build. The validation basically prevents all shared builds right now.

#### Details
* Remove the validation preventing shared building: I synced with the maintainer a bit in the projects Slack channel and he definitely confirmed, that there is no intention from the project side that things should only work with static linking
* Add required libraries to package_info: Not sure if that is required, but most projects seem to list all dependencies

---


irieger:
Taking out the relevant parts of #23112 without the build hacks done there by changing linking to public.

RubenRBS:
Let me know if you need help fixing the gcc compilation issue/ping me once you figure it out, thanks! :)

irieger:
> Let me know if you need help fixing the gcc compilation issue/ping me once you figure it out, thanks! :)

Hey @RubenRBS , thanks for offering help. Just wanted to have a more clean MR as documentation compared to the old one, but the underlying problem is discussed in #23421 and still is true. Looks like it works with gold, I tried mold and kept it for now in my profile and it works well so far. I have no clue what the proper fix for this might be. It just looks like the default linker is requireing to know where to find each indirectly needed so while mold and gold are fine with linking just the direct dependencies. (Populating the LD_LIBRARY_PATH before build makes default linking work too.)

recipes/opencolorio/all/conanfile.py
```diff
@@ -189,6 +186,20 @@ def package_info(self):
 
         self.cpp_info.libs = ["OpenColorIO"]
 
+        self.cpp_info.requires.append("expat::expat")
```
AbrilRBS, line 189: These requirements are only necesary when some of them are optional, or where not the whole package is needed. If no requirements are specified, all available ones coming from the direct dependencies are used

So I'd say we can revert this bit, as that's what this code is doing, adding all available ones. We can let Conan handle it in this case :)

AbrilRBS (commented):
Thanks! Comment a about the requires

recipes/opencolorio/all/patches/2.3.1-0001-fix-cmake-source-dir-and-targets.patch
```diff
@@ -68,7 +68,7 @@ index 26b4bb4c..d54cff3d 100755
          "$<BUILD_INTERFACE:xxHash>"
 -        ${YAML_CPP_LIBRARIES}
 -        MINIZIP::minizip-ng
-+        yaml-cpp
++        PUBLIC yaml-cpp
```
memsharded, line 71: I am bringing this back, as it is related to a limitation of CMakeDeps, sorry for the noise.
There is already some work in a new improved ``CMakeDeps`` generator that will make this unnecessary.

AbrilRBS (approved):

memsharded (approved):

PR #17032, opened 2023-04-14
cmake: Use a combined recipe that can download binaries or build from sources

datalogics-kam:
Specify library name and version:  **cmake/***

Fixes #16607

Having a separate recipe for CMake from sources isn't compatible with placing CMake recipes into the same remote repo. The recipe revisions keep overwriting each other. Therefore, I combined the recipes.

This solves our problem for Solaris. We have it in our fork, and I'm offering it to the community.

- By default, uses binaries on macOS, Windows and Linux.
- `from_sources` option can be used to build from sources on other platforms, or to force sources on Windows or Linux.
- The ability to bootstrap CMake was removed. If you need a CMake to build CMake from sources, either use a CMake that is already in Conan, or build a CMake with autotools by hand.

I do get
```
[HOOK - conan-center.py] pre_export(): ERROR: [CONANDATA.YML FORMAT (KB-H030)] First level entries ['binaries'] not allowed. Use only first level entries ['sources', 'patches'] in conandata.yml (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H030-CONANDATA.YML-FORMAT)
```

Suggestions welcome on how best to segregate data for sources and binaries in `conandata.yml` to conform to the requirements of the hooks.

I can also reorganize the recipe folder as requested.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24404, opened 2024-06-20
cmake: re-add from source builds as `cci.` versions

Croydon:
I think that the decision to re-package binary downloads was a mistake to begin with. Most of the named arguments are also outdated today as they are solved by version ranges.

This PR makes the CMake recipe, that builds from source, again available from the Conan Center remote, under the `cci.` version prefix.

---

Might be of interest to @datalogics-kam (#17032)

### DIFF
--- a/recipes/cmake/3.x.x/conandata.yml
+++ b/recipes/cmake/3.x.x/conandata.yml
@@ -1,28 +1,4 @@
 sources:
-  "3.19.8":
-    url: https://github.com/Kitware/CMake/releases/download/v3.19.8/cmake-3.19.8.tar.gz
-    sha256: 09b4fa4837aae55c75fb170f6a6e2b44818deba48335d1969deddfbb34e30369
-  "3.20.6":
-    url: https://github.com/Kitware/CMake/releases/download/v3.20.6/cmake-3.20.6.tar.gz
-    sha256: a0bd485e1a38dd13c0baec89d5f4adbf61c7fd32fddb38eabc69a75bc0b65d72
-  "3.21.7":
-    url: https://github.com/Kitware/CMake/releases/download/v3.21.7/cmake-3.21.7.tar.gz
-    sha256: 3523c4a5afc61ac3d7c92835301cdf092129c9b672a6ee17e68c92e928c1375a
-  "3.22.6":
-    url: https://github.com/Kitware/CMake/releases/download/v3.22.6/cmake-3.22.6.tar.gz
-    sha256: 73933163670ea4ea95c231549007b0c7243282293506a2cf4443714826ad5ec3
-  "3.23.5":
-    url: "https://github.com/Kitware/CMake/releases/download/v3.23.5/cmake-3.23.5.tar.gz"
-    sha256: "f2944cde7a140b992ba5ccea2009a987a92413762250de22ebbace2319a0f47d"
-  "3.24.3":
-    url: "https://github.com/Kitware/CMake/releases/download/v3.24.3/cmake-3.24.3.tar.gz"
-    sha256: "b53aa10fa82bff84ccdb59065927b72d3bee49f4d86261249fc0984b3b367291"
-  "3.25.2":
-    url: "https://github.com/Kitware/CMake/releases/download/v3.25.2/cmake-3.25.2.tar.gz"
-    sha256: "c026f22cb931dd532f648f087d587f07a1843c6e66a3dfca4fb0ea21944ed33c"
-  "3.26.4":
-    url: "https://github.com/Kitware/CMake/releases/download/v3.26.4/cmake-3.26.4.tar.gz"
-    sha256: "313b6880c291bd4fe31c0aa51d6e62659282a521e695f30d5cc0d25abbd5c208"
   "cci.3.29.6":
     url: "https://github.com/Kitware/CMake/releases/download/v3.29.6/cmake-3.29.6.tar.gz"
     sha256: "1391313003b83d48e2ab115a8b525a557f78d8c1544618b48d1d90184a10f0af"
--- a/recipes/cmake/3.x.x/conandata.yml
+++ b/recipes/cmake/3.x.x/conandata.yml
@@ -1,4 +1,4 @@
 sources:
-  "cci.3.29.6":
-    url: "https://github.com/Kitware/CMake/releases/download/v3.29.6/cmake-3.29.6.tar.gz"
-    sha256: "1391313003b83d48e2ab115a8b525a557f78d8c1544618b48d1d90184a10f0af"
+  "cci.3.26.4":
+    url: "https://github.com/Kitware/CMake/releases/download/v3.26.4/cmake-3.26.4.tar.gz"
+    sha256: "313b6880c291bd4fe31c0aa51d6e62659282a521e695f30d5cc0d25abbd5c208"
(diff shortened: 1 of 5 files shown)


memsharded:
Quick feedback regarding ``cci`` versions: ConanCenter will no longer be accepting these as a major digit when existing numeric versions exists, because they break the version ordering, the version ranges and other things like ``conan graph outdated``.

It will be possible with a new versioning approach that adds ``cci`` as a minor digit, like ``1.2.3.cci.DDDD`` to add cci versions that happen after ``1.2.3`` for example, because the repo stopped doing numbered releases. This is not the case for this repo, so the ``cci`` versioning approach wouldn't be valid for this case.

jcar87:
The current binary CMake recipe still provides several benefits over the source variant and the binaries we would be generating for it.

* While the recipe does remove the compiler setting from the package ID, if OpenSSL and zlib are linked statically (which they would be with default settings), the resulting package ID would include theirs, so the compiler version is reflected in the package ID in an indirect way - this means that unless users are using exactly the same configuration as the one we publish binaries with, they will experience missing binaries that they have to build from source. Not currently the case with the binary recipe.
* There is the added problem of our Conan 2 builds using gcc11, with the dynamic libstdc++ runtime - it won’t work on older systems (like Ubuntu 20.04). We can’t currently statically link the runtime, in case the user explicitly wants cmake to depend on openssl and zlib dynamically (those would require loading libstdc++.so).

The above points are true for all “tool” recipes currently in Conan Center and built for Conan 2 - however  we need to consider the usage of the cmake recipe is much more widespread, so we ought to be more careful.

* OpenSSL and ZLIB as requirements of CMake means that they _can_ cause conflicts in the graph of the build context, if other things in the build context also require openssl and zlib. This should now be mitigated almost completely in Conan Center as we have moved to require version ranges whenever OpenSSL and ZLIB are required - but we have no control over what the users do, and there may be some recipes that still don’t work with OpenSSL 3. 
* The versions of OpenSSL and zlib would be reflected in the package ID of CMake - since we use versions range, whenever new versions are published, new package IDs will be required, even for a given version of CMake for which we already had binaries:
    * If openssl and zlib are required statically, ANY new version published will cause this. Both openssl and zlib get relatively frequent updates.
    * If openssl and and zlib are required dynamically, only new major and minor versions will cause this (this is more frequent for OpenSSL, less frequent for zlib).

The points above mean that if the current source recipe were to be the default, users would still experience missing binaries quite frequently, having to build them from source, not only for new versions of CMake, but also whenever there are updates to zlib or openssl. We want to avoid this - if we can save users time, we should.

* On Windows, CMake requires CMake to be built - whereas the current binary recipe has no requirement. This may change the expectations of some users, if they have chosen to not have CMake pre-installed. Even if they do, building CMake on Windows in my experience takes a bit longer than on other platforms.

We do not want users to have to be building CMake from source unnecessarily - our focus is currently on the user experience of recipe/package consumers, and for the most part the current binary recipe has been working well for this - albeit with the major caveat of not working outside of supported configurations.


There are also other issues with the source variants that we have seen:
* If users choose to force OpenSSL and zlib to be shared in the build context, it will be exposed via LD_LIBRARY_PATH, not only to satisfy the dependency for CMake, but also any other process that CMake launches. This can cause issues if CMake launches system-managed executables that require the system libraries (e.g. git requires OpenSSL, and python requires zlib), and cause the linker to load incompatible versions. There is no easy fix for this - mileage will vary depending on what processes are launched from CMake (via execute process). 


As well as recipe maintainability issues:
* The bootstrap vs assume-we-have-cmake building mechanisms means a lot of duplicated logic, and only one path was ever tested by CI - the one that assumes a version of CMake is already present in the system anyway. I imagine there are cases where users do what to actually bootstrap, but that oath was never exercised by CI.
* The errors that you are currently getting on CI have to do with the fact that OpenSSL 3 depends on ZLIB, which interact with this: https://github.dev/Kitware/CMake/blob/e2be23a2b39f4380f32fe65ba770addc154579c7/Source/Modules/CMakeBuildUtilities.cmake#L136 - because zlib must now be a system library if OpenSSL depends on it. I remember testing that the source recipe was compatible with OpenSSL 3, but this was tested on macOS. Have you tested this locally on Linux?


While I appreciate the PR as it stands adds it as a new version that would fall outside the version range that is expressed elsewhere in Conan Center, we will not be adding the source versions back just yet. In particular, having two competing versioning schemes has downsides - we very much want users to be able to require the “latest”, e.g. `cmake/[*]`, and `cci.` would evaluate as newer - we’re already having issues with libraries that have gone from cci to something else, or from versions to `cci.` - we need to rethink the strategy for those cases moving forward.


Will otherwise accept fixes to the “source” recipe, such that it actually builds on Linux, or to add new versions, but will not be building this on CI just yet or publishing binaries for it.

This is  still very much in our radar - once we deploy our new CI pipeline, as well as allowing recipes that support Conan 2 only, we will have look into bringing this back - we want a user experience that is equivalent to what the binary recipe provides (binaries that work without conflicts or missing binaries), as well as being able to build it from source in other platforms.

But in the meantime, using the binary recipes has “protected” users from a myriad of problems while we focused our efforts on migrating recipes to support Conan 2.0. Down the line we will also ensure that we have the ability to validate the recipe with different options (so that the bootstrap doesn’t diverge), as well as using an old enough gcc or linking the runtime statically.

As for alternatives in the meantime, if for any reason users do not want to use the Conan-center provided versions for CMake, there are ways to handle this without modifying any other recipes:
* `[platform_requires]` in the profile ([docs](https://docs.conan.io/2/reference/config_files/profiles.html#platform-requires)) - can be used to tell Conan that the system available cmake should be used instead - on recent enough systems this tends to be the case.
* `[replace_tool_requires]` can be used to redirect `cmake/...` to a different dependency, for example if users have ones they have built from source (perhaps with a user/channel)
* users that wish to have their own cmake recipe and packages, and bypass the one from Conan Center, can prevent Conan from considering the Conan Center one - I've tried this quickly, the following should prevent CMake from being satisfied by the conancenter remote.
```
conan remote update --allowed-packages='~cmake/*' conancenter
```

datalogics-kam:
Hi! My duties have shifted somewhat, and I'm not working directly with Conan at the moment, but I will make the appropriate teams at Datalogics aware of this development.

A few comments in recap:

* Primarily, Datalogics is interested in building CMake for platforms other than Mac/Windows/Linux. When built for other operating systems, at least at DL, concerns about static linking and libc versions are not relevant.
* The work in #17032 revolved mainly around being able to use binaries when available, but source when not, without having to use an alternate package name or recipe revision, to avoid collisions in Conanfile or in the remote repo.
* In fact, we expected to even use the same package version.
* The way I last left our internal Curated Conan Center Index project, we were using a `bootstrap_cmake` package to build CMake. This was a known good CMake package that was stored under a special name that we introduced as a `tool_requires` for building CMake. Or, we could use a CMake that was built by hand (tediously, but only once) to be able to make latest CMake from a Conan recipe.
* Because of that, the bootstrapping logic was left out of #17032. I agree with the Conan team's assessment that it was a maintenance headache.

So that's what I remember. I'll drop a link to this PR into our company Slack to see if it is of interest to those currently working with Conan.

Thanks for bringing this to my attention, and thanks for Conan!

Croydon:
Thank you for the extensive feedback @memsharded, @jcar87 and @datalogics-kam.

It might be possible to bootstrap CMake on Windows with msys2 as a dependency. In a second step, I would have liked to try that and if it would have worked, modifying the recipe to always bootstrap, hence eliminating the requirement of a pre-existing CMake binary in path and simplifying the recipe.

But the other points would still stand, so I am going to close this for now.

PR #24408, opened 2024-06-20
pcre2: Add support for PCRE2_LINK_SIZE

keef-cognitiv:
### Summary
Changes to recipe:  **pcre2/10.XX**

#### Motivation
Add support for passing a configurable PCRE2_LINK_SIZE. 

This directly translates to LINK_SIZE here https://github.com/PCRE2Project/pcre2/blob/9a51f31da1c2d45e1f31d6a0d6b2f62975fed373/src/config.h.generic#L167.

In short, the default size of 2 creates a hard limit of 65535 bytes to be used by the compiled regular expression. This creates a hard upper limit for the size of regular expressions that can be compiled. Using a link size of 4 expands this to 4GB. For most use cases, 64KB is more than enough. But in some edge cases, it's not:

https://github.com/PCRE2Project/pcre2/issues/271

https://github.com/PCRE2Project/pcre2/issues/406

This adds support for the flag in the conan package to support the users in those edge cases.

#### Details
This just adds an option (`link_size` to the package that gets passed through as `PCRE2_LINK_SIZE` to cmake which passes `LINK_SIZE` to the compiler.

I am not well versed on the CI system that conan-center-index uses to know whether I'm missing a step to add this to all of the test builds of the package, and where to add it.

---

### DIFF
--- a/recipes/pcre2/all/conanfile.py
+++ b/recipes/pcre2/all/conanfile.py
@@ -29,6 +29,7 @@ class PCRE2Conan(ConanFile):
         "with_bzip2": [True, False],
         "support_jit": [True, False],
         "grep_support_callout_fork": [True, False],
+        "link_size": [2, 3, 4],
     }
     default_options = {
         "shared": False,
@@ -41,6 +42,7 @@ class PCRE2Conan(ConanFile):
         "with_bzip2": True,
         "support_jit": False,
         "grep_support_callout_fork": True,
+        "link_size": 2,
     }
 
     def export_sources(self):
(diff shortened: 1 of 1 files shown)


recipes/pcre2/all/conanfile.py
```diff
@@ -29,6 +29,7 @@ class PCRE2Conan(ConanFile):
         "with_bzip2": [True, False],
         "support_jit": [True, False],
         "grep_support_callout_fork": [True, False],
+        "link_size": [2, 3, 4],
```
RubenRBS, line 32: Sorry for the little experience in prce2, have to ask: is 4 the max just because having more than 4GB of resulting compiled expression would be overkill, or is it a hard limit in the codebase?

RubenRBS (commented):
Thanks! Minor question to get a bit more insight before approving :)

recipes/pcre2/all/conanfile.py
keef-cognitiv: The number ultimately represents the number of bytes to use to represent offsets in the compiled expression. Which sounds like branch sizes in assembly but I think is more for the state machine opcodes to be able to say "how far to jump in the event of some matching state in order to find the next proper state." So if that number is only 2 bytes (16 bits), the next state can't be more than 65535 bytes away. And if it's a 2MB expression, that's going to often be way too far to represent. The trade off though is that there's more "fluff" in the state machine opcodes so more cache thrashing => worse performance where those larger jumps weren't needed.

That said, why it can't go up to 8 for example, I don't know. Having looked through the code, it seems like it maybe could. But I might be missing some place where a value ultimately is read into a uint32_t or something. In either case, the autconf based build strictly enforces that it's 2, 3, or 4. I don't think the cmake one does though.

RubenRBS: Thanks for the insight! :)

RubenRBS (approved):
Than

perseoGI (approved):
LGTM thanks for contributing to Conan!

PR #24423, opened 2024-06-23
bgfx: add new version and consolidated recipe

RazielXYZ:
### Summary
Changes to recipe:  **bgfx**

#### Motivation
This adds a consolidated recipe for bgfx going forward. This is versioned using bgfx' versioning scheme (https://github.com/bkaradzic/bgfx/blob/93961afcfd45bdcbdb91bddfb133a621d340f136/src/bgfx.cpp#L3566), and integrates bx and bimg into this package. This was done after discussions about whether having bx and bimg separate is valuable, and how due to their interdependence and reliance on specific versions, having them separate makes updates on CCI much slower and more cumbersome.

#### Details
The recipe is still based on the previous bgfx recipe, but with quite a few changes and additions.
It adds all things required for bimg and bx to be integrated into this recipe, such as defines, libs, and their headers.
It also adds an option for rtti, due to the fact that bgfx' (and bx', and bimg's) build scripts are hardcoded to build with rtti disabled. Packages on CCI are rtti enabled by default, and mixing these can cause linker issues, especially on linux in my experience.
There is also improved support for mingw and android, changes to make other things more consistent, and so on.
The new recipe has been locally tested on windows 11 msvc194 and mingw, linux gcc, and android.

Some questions I was pondering:
Would there be value in adding an option for overriding bx' CRT detection in order for this to build with something like musl?
Would there be value in exposing options for opengl version etc. and other compile-time options that bgfx has?

---

### DIFF
--- a/recipes/bgfx/config.yml
+++ b/recipes/bgfx/config.yml
@@ -1,3 +1,5 @@
 versions:
+  "1.127.8771":
+    folder: consolidated
   "cci.20230216":
     folder: all
--- a/recipes/bgfx/consolidated/conandata.yml
+++ b/recipes/bgfx/consolidated/conandata.yml
@@ -1,0 +1,11 @@
+sources:
+  "1.127.8771":
+    "bgfx":
+      url: "https://github.com/bkaradzic/bgfx/archive/06d0e2af2f50fee3e8e84e0c3b407073692a678c.tar.gz"
+      sha256: "0869F9D18C63AB8D804918FEFCC53D8568626A26FCAE7273DF1A70D4C8E7AFE3"
+    "bimg":
+      url: "https://github.com/bkaradzic/bimg/archive/2afa64c14c1e3dd5d28412ee03bee0dfe7242f03.tar.gz"
+      sha256: "376E17EA4A70CEE72E6006DD942CB2F2FE3DB047F524A3F203A15A8BBA1BCD90"
+    "bx":
+      url: "https://github.com/bkaradzic/bx/archive/e64b7c097f30d33134fe8fa522b3c95fc461c3d3.tar.gz"
+      sha256: "2A6E438AC6250CD6052CBBABF2906482DFA1585E537064CB5DA47E2EC728FE92"
--- a/recipes/bgfx/consolidated/conanfile.py
+++ b/recipes/bgfx/consolidated/conanfile.py
@@ -1,0 +1,368 @@
+from conan import ConanFile
+from conan.tools.files import copy, get, rename, replace_in_file
+from conan.tools.build import check_min_cppstd
+from conan.tools.layout import basic_layout
+from conan.tools.microsoft import is_msvc, check_min_vs, is_msvc_static_runtime
+from conan.tools.apple import fix_apple_shared_install_name, is_apple_os
+from conan.tools.scm import Version
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.microsoft import MSBuild, VCVars
+from conan.tools.gnu import Autotools, AutotoolsToolchain
+from conan.tools.env import VirtualBuildEnv
+from pathlib import Path
+import os
+
+required_conan_version = ">=1.50.0"
+
+class bgfxConan(ConanFile):
+    name = "bgfx"
+    license = "BSD-2-Clause"
+    homepage = "https://github.com/bkaradzic/bgfx"
+    url = "https://github.com/conan-io/conan-center-index"
+    description = "Cross-platform, graphics API agnostic, \"Bring Your Own Engine/Framework\" style rendering library."
+    topics = ("rendering", "graphics", "gamedev")
+    settings = "os", "compiler", "arch", "build_type"
+    options = {"fPIC": [True, False], "shared": [True, False], "tools": [True, False], "rtti": [True, False]}
+    default_options = {"fPIC": True, "shared": False, "tools": False, "rtti": True}
+
+    @property
+    def _bx_folder(self):
+        return "bx"
+
+    @property
+    def _bimg_folder(self):
+        return "bimg"
+    
+    @property
+    def _bgfx_folder(self):
+        return "bgfx"
+
+    @property
+    def _bgfx_path(self):
+        return os.path.join(self.source_folder, self._bgfx_folder)
+
+    @property
+    def _genie_extra(self):
+        genie_extra = ""
+        if is_msvc(self) and not is_msvc_static_runtime(self):
+            genie_extra += " --with-dynamic-runtime"
+        if self.options.shared:
+            genie_extra += " --with-shared-lib"
+        # generate tools projects regardless of tools option because that also generates bimg_encode
+        genie_extra += " --with-tools"
+        return genie_extra
+
+    @property
+    def _lib_target_prefix(self):
+        if is_msvc(self):
+            return "libs\\"
+        else:
+            return ""
+
+    @property
+    def _tool_target_prefix(self):
+        if is_msvc(self):
+            return "tools\\"
+        else:
+            return ""
+        
+    @property
+    def _shaderc_target_prefix(self):
+        if is_msvc(self):
+            return "shaderc\\"
+        else:
+            return ""
+        
+    @property
+    def _tools(self):
+        return ["texturec", "texturev", "geometryc", "geometryv", "shaderc"]
+
+    @property
+    def _projs(self):
+        projs = [f"{self._lib_target_prefix}bx", f"{self._lib_target_prefix}bimg", f"{self._lib_target_prefix}bimg_decode", f"{self._lib_target_prefix}bimg_encode"]
+        if self.options.shared:
+            projs.extend([f"{self._lib_target_prefix}bgfx-shared-lib"])
+        else:
+            projs.extend([f"{self._lib_target_prefix}bgfx"])
+        if self.options.tools:
+            for tool in self._tools:
+                if "shaderc" in tool:
+                    projs.extend([f"{self._tool_target_prefix}{self._shaderc_target_prefix}{tool}"])
+                else:
+                    projs.extend([f"{self._tool_target_prefix}{tool}"])
+        return projs
+
+    @property
+    def _compiler_required(self):
+        return {
+            "gcc": "8",
+            "clang": "11",
+            "apple-clang": "12", #to keep CCI compiling on osx 11.0 or higher, for now
+        }
+
+    @property
+    def _settings_build(self):
+        return getattr(self, "settings_build", self.settings)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("opengl/system")
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, 17)
+        check_min_vs(self, 192)
+        if not is_msvc(self):
+            try:
+                minimum_required_compiler_version = self._compiler_required[str(self.settings.compiler)]
+                if Version(self.settings.compiler.version) < minimum_required_compiler_version:
+                    raise ConanInvalidConfiguration("This package requires C++17 support. The current compiler does not support it.")
+            except KeyError:
+                self.output.warn("This recipe has no checking for the current compiler. Please consider adding it.")
+
+    def build_requirements(self):
+        self.tool_requires("genie/1170")
+        if not is_msvc(self) and self._settings_build.os == "Windows":
+            if self.settings.os == "Windows": # building for windows mingw
+                # self.win_bash = True
+                # if not self.conf.get("tools.microsoft.bash:path", check_type=str):
+                #     self.tool_requires("msys2/cci.latest")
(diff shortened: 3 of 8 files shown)


RazielXYZ:
This needs osx >= 13, how do we limit it to that? Looks like it's trying to use 11.3

recipes/bgfx/config.yml
```diff
@@ -1,3 +1,5 @@
 versions:
+  "1.127.8771":
+    folder: consolidated
   "cci.20230216":
     folder: all
```
xoorath, line 5: I think it might be better to put the new recipe in all since it will be all moving forward, and to move the old recipe to another folder name like `cci.2023`

recipes/bgfx/consolidated/conandata.yml
```diff
@@ -0,0 +1,11 @@
...
+  "1.127.8771":
+    "bgfx":
+      url: "https://github.com/bkaradzic/bgfx/archive/06d0e2af2f50fee3e8e84e0c3b407073692a678c.tar.gz"
+      sha256: "0869F9D18C63AB8D804918FEFCC53D8568626A26FCAE7273DF1A70D4C8E7AFE3"
+    "bimg":
+      url: "https://github.com/bkaradzic/bimg/archive/2afa64c14c1e3dd5d28412ee03bee0dfe7242f03.tar.gz"
+      sha256: "376E17EA4A70CEE72E6006DD942CB2F2FE3DB047F524A3F203A15A8BBA1BCD90"
+    "bx":
+      url: "https://github.com/bkaradzic/bx/archive/e64b7c097f30d33134fe8fa522b3c95fc461c3d3.tar.gz"
+      sha256: "2A6E438AC6250CD6052CBBABF2906482DFA1585E537064CB5DA47E2EC728FE92"
```
xoorath, line 11: The version of bx you're using here is a couple commits newer than the version of bgfx.

The following are a match for the version of bgfx you're using.
```
"sha256": "0fc705c7bb7ae01f2a2e1ef2be2426b877407e4cdab1b3ca5d4f77eb814716df"
"url": "https://github.com/bkaradzic/bx/archive/c5b7b4629961ed7ac882cd3f64f28e1b25910e55.tar.gz"
```

xoorath (changes requested):
Recommending a new recipe folder name + updating the bx package to a different version so it's in sync with bgfx.

recipes/bgfx/config.yml
RubenRBS: We can move forward with this, I also think it's cleaner

recipes/bgfx/consolidated/conandata.yml
RubenRBS: How could we know this exactly for future versions? Just that de dates match? Or is there a file specifying them?

recipes/bgfx/consolidated/conanfile.py
```diff
@@ -0,0 +1,368 @@
...
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("opengl/system")
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, 17)
+        check_min_vs(self, 192)
```
RubenRBS, line 120:
```suggestion
        check_min_vs(self, 192)
        if self.settings.os == "Macos" and self.settings.get_safe("os.sdk_version") and self.settings.get_safe("os.sdk_version") < "13.0":
            raise ConanInvalidConfiguration(f"{self.ref} requires macos sdk version >= 13")
```

recipes/bgfx/consolidated/conanfile.py
```diff
@@ -0,0 +1,368 @@
...
+        if not is_msvc(self):
+            try:
+                minimum_required_compiler_version = self._compiler_required[str(self.settings.compiler)]
+                if Version(self.settings.compiler.version) < minimum_required_compiler_version:
+                    raise ConanInvalidConfiguration("This package requires C++17 support. The current compiler does not support it.")
+            except KeyError:
+                self.output.warn("This recipe has no checking for the current compiler. Please consider adding it.")
```
RubenRBS, line 127:
```suggestion
                self.output.warning(f"{self.ref} does no checking for the current compiler. Assuming it works, please consider adding it.")
```

recipes/bgfx/consolidated/conandata.yml
RazielXYZ: Pretty much just the dates, yes. But generally, the latest commit of any of these at any point in time should work with the latest on all others, barring some very short term situations where one repo has been committed but the others have not and need to be to work with the changes.

recipes/bgfx/config.yml
RazielXYZ: By "this" we mean moving the new stuff to "all"?

RubenRBS: Yes :) and the old recipe to a folder with a clarifying name, I'll let you come up with it

RazielXYZ:
On conan 1, it still doesn't seem to use C++17 even with setting tc.cppstd = "-std=c++17" manually. I'm guessing tc.generate() overwrites it from self.settings?
On mac, I guess we might need to skip macos builds on cci for now if the agents' macos sdk is too old.

recipes/bgfx/all/conanfile.py
```diff
@@ -48,119 +48,133 @@ def _genie_extra(self):
...
         if is_msvc(self):
-            tc = VCVars(self)
+            tc = MSBuildToolchain(self)
+            if not self.settings.get_safe("compiler.cppstd"):
+                tc.cppstd = "c++17"
             tc.generate()
         else:
             tc = AutotoolsToolchain(self)
+            if not self.settings.get_safe("compiler.cppstd"):
+                tc.cppstd = "-std=c++17"
```
RubenRBS, line 165: Filed as https://github.com/conan-io/conan/issues/16533

recipes/bgfx/consolidated/conandata.yml
RubenRBS: Thanks for the info!

recipes/bgfx/all/conanfile.py
xoorath: @AbrilRBS What's the fate of this following the closure of #16533 ?

xoorath (approved):
Looks good. I double checked the commits and hashes.

xoorath:
Hey @AbrilRBS, sorry to bother you but is there any way you're aware of to escalate this topic for assessment? Or any workarounds?

recipes/bgfx/config.yml
AbrilRBS: Circling back to this, is there any actual reason to keep the cci version around? I think it would be ok to just keep the released versions from now on

recipes/bgfx/all/conanfile.py
```diff
@@ -48,119 +49,121 @@ def _genie_extra(self):
...
-                if Version(self.settings.compiler.version) < minimum_required_compiler_version:
-                    raise ConanInvalidConfiguration("This package requires C++14 support. The current compiler does not support it.")
-            except KeyError:
-                self.output.warn("This recipe has no checking for the current compiler. Please consider adding it.")
+        check_min_cppstd(self, 17)
+        if self.settings.os == "Macos":
+            if self.settings.get_safe("os.sdk_version") and self.settings.get_safe("os.sdk_version") < "11.0":
+                raise ConanInvalidConfiguration(f"{self.ref} requires macos sdk version >= 11")
+            if (not self.settings.get_safe("os.sdk_version") and
+                    self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "14.0"):
```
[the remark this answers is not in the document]
RazielXYZ: That seems fine to me!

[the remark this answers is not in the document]
xoorath: Thank you for your work on this. I'm excited to make use of this package again.

recipes/bgfx/all/conanfile.py
```diff
@@ -191,27 +195,39 @@ def build(self):
...
             conf = build_type_to_make_config[str(self.settings.build_type)]
             if os_to_use_make_config_suffix[str(self.settings.os)]:
                 conf += arch_to_make_config_suffix[str(self.settings.arch)]
             if self.settings.os == "Windows":
-                mingw = "MINGW=$MINGW_PREFIX"
-                proj_path = proj_path.replace("\\", "/") # Fix path for msys...
+                if "mingw-builds" in self.dependencies.build:
+                    # self.run("if [ ! -d /mingw64 ]; then mkdir /mingw64; fi")
+                    # self.run("pacman -Sy mingw-w64-x86_64-gcc --needed --noconfirm")
+                    # mingw = "MINGW=$MSYS_ROOT/mingw64"
```
memsharded, line 226: These lines seems unrelated or at least confusing, maybe should be removed.

I am not convinced about the check ``if "mingw-builds" in self.dependencies.build:`` for a specific dependency name in the dependencies, cause if someone uses another ``mingw_compiler`` package it won't work. Ideally the package itself could define the ``MINGW`` env-var or something like that.

But I also think we can move forward with it, and this can be discussed later.

memsharded (approved):
I have reviewed the changes and recipe in code, and things are looking good.
But I haven't done any extra builds or testing, so please @AbrilRBS and @jcar87 let me know if this is enough, or would you prefer that I do some further builds.

recipes/bgfx/all/conanfile.py
memsharded: Sure, no problem, leave it as-is, I think we can figure this out later, after merge.

RazielXYZ:
@memsharded considering it's been a bit, should we also update this to the latest bgfx and others before merging (and get rid of those commented out attempts at mingw working)? Or merge it as is first then get a new version added afterwards?

AbrilRBS:
>  should we also update this to the latest bgfx and others before merging

If the new version is easily added, then by all means yes! I'll do some local testing once you update and we can merge, but if you are ok with the old versions, also happy with it

RazielXYZ:
@AbrilRBS awesome - just did that, and did a couple local builds too before pushing it.

RazielXYZ:
Can't say I've seen that error before (nor can I seem to reproduce it locally)...

RazielXYZ:
Would a newer gcc on CCI have a newer glibc, or is this a runner issue that we have to bypass somehow else?

jcar87:
Thanks @RazielXYZ  - this is certainly a step in the right direction. Apologies it took to get back to this one.

We have moved forward with https://github.com/conan-io/conan-center-index/pull/28338  - which does consolidate all in a single recipe, but using CMake from https://github.com/bkaradzic/bgfx.cmake/
This makes it easier to maintain. Please let us know if this achieves the same goals or if there's anything missing.

RazielXYZ:
@jcar87 thanks for letting me know! bgfx.cmake is not actually maintained by the bgfx author, although it is basically endorsed, as long as there's people who maintain it, so it should be fine.
The only relatively important thing missing is the profiler option. I'm not sure if bgfx.cmake has support for that or not, as I've always used the default genie system for bgfx. If bgfx.cmake has or gets support for it, I'd assume we can add it.

jcar87:
> @jcar87 thanks for letting me know! bgfx.cmake is not actually maintained by the bgfx author, although it is basically endorsed, as long as there's people who maintain it, so it should be fine. The only relatively important thing missing is the profiler option. I'm not sure if bgfx.cmake has support for that or not, as I've always used the default genie system for bgfx. If bgfx.cmake has or gets support for it, I'd assume we can add it.

Is the profiler an executable tool? we can look into adding it if it's needed.

RazielXYZ:
@jcar87 no, the profiler is a a compile-time option for bgfx. It defines BGFX_CONFIG_PROFILER=1 which enables and exposes an interface for users to integrate profiler callbacks and sections, and adds profiler sections to some places in bgfx internally.

uilianries:
@RazielXYZ Hello! 

Unfortunately, [I did not find an externalized option for profiler](https://github.com/search?q=repo%3Abkaradzic%2Fbgfx.cmake%20profiler&type=code) on https://github.com/bkaradzic/bgfx.cmake, but reading the upstream build script, it looks resulting in 2 defines:

https://github.com/bkaradzic/bgfx/blob/5aaf3c62e4fe1a4a7567ff739cbbca8b41217616/scripts/genie.lua#L197

```lua
if _OPTIONS["with-profiler"] then
	defines {
		"ENTRY_CONFIG_PROFILER=1",
		"BGFX_CONFIG_PROFILER=1",
	}
end
```

As a workaround, you can inject those defines directly into your profile, like:

```ini
...

[conf]
bgfx/1.129.8930-495:tools.build:defines=['ENTRY_CONFIG_PROFILER=1', 'BGFX_CONFIG_PROFILER=1']
bgfx/1.129.8930-495:tools.info.package_id:confs=['tools.build:defines']
```

This will inject those definitions into your build and change your package ID as well. 

Still, I think it's worth trying to push to upstream as well, because it's a feature maintained in the official build script.

Regards.

RazielXYZ:
The ENTRY_CONFIG_PROFILER one is not necessary (unless using the Entry framework in your own projects, or wanting to profile the examples or tools I suppose), but that sounds good. Thanks!

PR #24426, opened 2024-06-24
libpcap: add explicit dpdk option

ReRandom:
### Summary
Changes to recipe:  **libpcap**

#### Motivation
libpcap has [compilation errors when compiling with dpdk](https://github.com/the-tcpdump-group/libpcap/issues/1159) lib higher then v21.11-rc1. So explicitly setting this flag will allow you to control this.

#### Details
I ran into such a problem on a machine where dpdk was installed, although I did not need the compatibility of these libraries. It was possible to solve the problem by isolating the environments, but still.

To reproduce the problem, you must first install dpdk, then try to install libpcap.

---

### DIFF
--- a/recipes/libpcap/all/conanfile.py
+++ b/recipes/libpcap/all/conanfile.py
@@ -30,12 +30,14 @@ class LibPcapConan(ConanFile):
         "fPIC": [True, False],
         "enable_libusb": [True, False],
         "with_snf": [True, False],
+        "with_dpdk": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "enable_libusb": False,
         "with_snf": False,
+        "with_dpdk": False,
     }
 
     # TODO: Add dbus-glib when available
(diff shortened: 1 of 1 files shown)


recipes/libpcap/all/conanfile.py
```diff
@@ -30,12 +30,14 @@ class LibPcapConan(ConanFile):
...
         "enable_libusb": [True, False],
         "with_snf": [True, False],
+        "with_dpdk": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "enable_libusb": False,
         "with_snf": False,
+        "with_dpdk": False,
```
RubenRBS, line 40: You mention that without this option, the recipe does not build, which leads me to belive that this is True in upstream's scripts when nothing is said.

If this makes all builds fail, we could make an expcetion here and diverge from upstream, but the CCI policy is to follow Upstream choices as closely as possible. Any insight into the proper default value? Thanks!

RubenRBS (commented):
Thanks! I have some questions with the default value, othwise looks good :)

recipes/libpcap/all/conanfile.py
ReRandom: I'm sorry, I probably misled you, errors occur with the `--with-dpdk` flag, which is enabled by default in libpcap. But the problem appears only if dpdk lib is installed. To avoid errors, you need the `--without-dpdk` flag, which is set by `"with_dpdk": False`, which I set by default.

valgur:
Can probably add a `dpdk` requirement as well once #24817 is merged.

ReRandom:
I don't think that's necessary. Moreover, since the assembly problem has been around for a long time, I'm afraid no one uses it.

PR #24427, opened 2024-06-24
[dbus] Use default meson values when options are None

uilianries:
### Summary
Changes to recipe:  **dbus/1.15.8**

#### Motivation

This bug was reported by @garethsb. Thank you! 

Using option value as `None` is converted to `"None"`, resulting in misconfiguration when consuming the project.

Plus, `with_glib` option is deprecated, and now removed.

#### Details

The upstream defines specific paths when not configured: https://gitlab.freedesktop.org/dbus/dbus/-/blob/dbus-1.15.8/meson.build?ref_type=tags#L956

It will be better only passing values when the option is configured.

---

### DIFF
--- a/recipes/dbus/1.x.x/conanfile.py
+++ b/recipes/dbus/1.x.x/conanfile.py
@@ -32,10 +32,9 @@ class DbusConan(ConanFile):
         "system_socket": [None, "ANY"],
         "system_pid_file": [None, "ANY"],
         "with_x11": [True, False],
-        "with_glib": ["deprecated", True, False],
         "with_systemd": [True, False],
         "with_selinux": [True, False],
-        "session_socket_dir": ["ANY"],
+        "session_socket_dir": [None, "ANY"],
     }
     default_options = {
         "shared": False,
@@ -45,7 +44,6 @@ class DbusConan(ConanFile):
         "system_socket": None,
         "system_pid_file": None,
         "with_x11": False,
-        "with_glib": "deprecated",
         "with_systemd": False,
         "with_selinux": False,
         "session_socket_dir": "/tmp",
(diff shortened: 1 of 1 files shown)


garethsb (approved):
LGTM, thanks!

RubenRBS (approved):
Good catch :)

recipes/dbus/1.x.x/conanfile.py
```diff
@@ -134,8 +130,10 @@ def generate(self):
             tc.project_options["message_bus"] = bool(self.options.message_bus)
         if self.options.get_safe("dbus_user"):
             tc.project_options["dbus_user"] = str(self.options.dbus_user)
-        tc.project_options["system_pid_file"] = str(self.options.get_safe("system_pid_file", ""))
-        tc.project_options["system_socket"] = str(self.options.get_safe("system_socket", ""))
+        if self.options.system_pid_file:
```
garethsb, line 133: @uilianries I think maybe these ifs should all use `.get_safe()` to handle the case that the key is not present in options at all, or can that never happen?

RubenRBS: That can never happen, as it's not being deleted in the config_options/configure methods :)

recipes/dbus/1.x.x/conanfile.py
```diff
@@ -134,8 +130,10 @@ def generate(self):
...
+            tc.project_options["system_pid_file"] = self.options.system_pid_file
```
franramirez688, line 134:
```suggestion
            tc.project_options["system_pid_file"] = str(self.options.system_pid_file)
```

recipes/dbus/1.x.x/conanfile.py
```diff
@@ -134,8 +130,10 @@ def generate(self):
...
+        if self.options.system_socket:
+            tc.project_options["system_socket"] = self.options.system_socket
```
franramirez688, line 136:
```suggestion
            tc.project_options["system_socket"] = str(self.options.system_socket)
```

franramirez688 (approved):

RubenRBS (approved):

PR #24448, opened 2024-06-26
dtl: support gcc 14

toge:
### Summary
Changes to recipe:  **dtl/***

#### Motivation
To support gcc 14, make `enableTrivial` mutable.
https://github.com/cubicdaiya/dtl/pull/18

#### Details

---

### DIFF
--- a/recipes/dtl/all/conanfile.py
+++ b/recipes/dtl/all/conanfile.py
@@ -17,6 +17,7 @@ class DtlConan(ConanFile):
 
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
+    no_copy_source = True
 
     def layout(self):
         basic_layout(self, src_folder="src")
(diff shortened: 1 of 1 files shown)


recipes/dtl/all/conanfile.py
```diff
@@ -18,7 +17,6 @@ class DtlConan(ConanFile):
 
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
-    no_copy_source = True
```
AbrilRBS, line 21: @toge we can keep this if we move the replace_in_file to the source() method, as every configuration will need it, and we save some space :)

AbrilRBS (commented):
Approving this as an executive decision, this will not impact the generated code, it was wrong before! If removing const, at most we would lose on optimizations, but that's something good. Waiting on @toge to address my suggestion before merging :)

recipes/dtl/all/conanfile.py
toge: Thank you for your review!
I agree with you. Fixed.

perseoGI (approved):
Approving as an exception

AbrilRBS (approved):

PR #24450, opened 2024-06-26
qt: fix setting cppstd

SvenRoederer:
### Summary

#### Motivation
"current_cpp_std" is of type str, so the list elements that are indexed by it should also be of this type

---
- [ X ] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
- [ X ] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)
- [ X ] Tested locally with at least one configuration using a recent version of Conan

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -613,16 +613,16 @@ class QtConan(ConanFile):
         current_cpp_std = self.settings.get_safe("compiler.cppstd", default_cppstd(self))
         current_cpp_std = str(current_cpp_std).replace("gnu", "")
         cpp_std_map = {
-            11: "FEATURE_cxx11",
-            14: "FEATURE_cxx14",
-            17: "FEATURE_cxx17",
-            20: "FEATURE_cxx20"
+            "11": "FEATURE_cxx11",
+            "14": "FEATURE_cxx14",
+            "17": "FEATURE_cxx17",
+            "20": "FEATURE_cxx20"
         }
         if Version(self.version) >= "6.5.0":
-            cpp_std_map[23] = "FEATURE_cxx2b"
+            cpp_std_map["23"] = "FEATURE_cxx2b"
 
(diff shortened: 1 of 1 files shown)


recipes/qt/6.x.x/conanfile.py
```diff
@@ -613,16 +613,16 @@ def generate(self):
...
+            "14": "FEATURE_cxx14",
+            "17": "FEATURE_cxx17",
+            "20": "FEATURE_cxx20"
         }
         if Version(self.version) >= "6.5.0":
-            cpp_std_map[23] = "FEATURE_cxx2b"
+            cpp_std_map["23"] = "FEATURE_cxx2b"
 
         for std, feature in cpp_std_map.items():
-            tc.variables[feature] = "ON" if int(current_cpp_std) >= std else "OFF"
```
jcar87, line 625: the comparison as it currently is, is correct, if I'm not mistaken

`current_cpp_std` is a string, and it is being cast to an `int`, and is being compared with an int (each of the `keys` in the `cpp_std_map` dictionary.

Testing this locally seems to behave as expected.

jcar87:
From what I can see, this is merely stylistic and the code is behaving as expected.

With `compiler.cppstd=gnu17`, I get:
```
FEATURE_cxx11=ON
FEATURE_cxx14=ON
FEATURE_cxx17=ON
FEATURE_cxx20 =OFF
FEATURE_cxx2b=OFF
```

which seems correct - I believe the comparison is being perform as intended.


unfortunately, given the size of the library and the number of versions we simultaneously maintain, I don't think it is worth the CI efforts given that it does not alter the behaviour, and it seems stylistic only. 

Closing this since functionally this does not change the recipe logic at all. If there is an actual logic error, please reopen the issue or provide more details, and we'll be happy to look into this. Apologies it took long to get to this PR.

PR #24459, opened 2024-06-27
llnl-units: Add 0.9.1 to conan center index

MridulS:
### Summary
Adds a new recipe for https://github.com/LLNL/units

#### Motivation

We maintain a local copy of this conanfile.py, it would be nice to use the index so we don't need to keep maitaining a local copy.

#### Details

This is tested with conan 1.x locally, not with conan2

---

### DIFF
--- a/recipes/llnl-units/all/conanfile.py
+++ b/recipes/llnl-units/all/conanfile.py
@@ -29,19 +29,15 @@ class UnitsConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
-        "base_type": ["uint32_t", "uint64_t"],
-        "namespace": [None, "ANY"],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
-        "base_type": "uint32_t",
-        "namespace": None,
     }
 
     @property
     def _min_cppstd(self):
-        return 17
+        return 14
 
     @property
     def _compilers_minimum_version(self):
@@ -97,9 +93,10 @@ class UnitsConan(ConanFile):
 
     def package_info(self):
         self.cpp_info.libs = ["units"]
-        units_namespace = self.options.get_safe("namespace")
-        self.cpp_info.defines = [f"UNITS_BASE_TYPE={self.options.base_type}"]
-        if units_namespace:
+        namespace = self.conf.get("user.llnl-units:namespace", check_type=str)
+        base_type = self.conf.get("user.llnl-units:base_type", check_type=str, default="uint32_t")
+        self.cpp_info.defines = [f"UNITS_BASE_TYPE={base_type}"]
+        if namespace:
             self.cpp_info.defines.append(f"UNITS_NAMESPACE={units_namespace}")
 
         self.cpp_info.set_property("cmake_file_name", "units")
--- a/recipes/llnl-units/all/conanfile.py
+++ b/recipes/llnl-units/all/conanfile.py
@@ -68,7 +68,6 @@ class UnitsConan(ConanFile):
         tc = CMakeToolchain(self)
         tc.preprocessor_definitions["UNITS_CMAKE_PROJECT_NAME"] = "LLNL-UNITS"
         tc.preprocessor_definitions["UNITS_ENABLE_TESTS"] = "OFF"
-        tc.preprocessor_definitions["UNITS_BASE_TYPE"] = self.options.base_type
         tc.preprocessor_definitions["UNITS_BUILD_SHARED_LIBRARY"] = self.options.shared
         tc.preprocessor_definitions[
             "UNITS_BUILD_STATIC_LIBRARY"
--- a/recipes/llnl-units/all/conanfile.py
+++ b/recipes/llnl-units/all/conanfile.py
@@ -1,0 +1,98 @@
+import os
+
+from conans import ConanFile, tools
+from conan.tools.cmake import CMake, CMakeToolchain
+from conan.tools.files import copy, rm, rmdir
+
+
+class UnitsConan(ConanFile):
+    name = "llnl-units"
+    license = "BSD-3"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://units.readthedocs.io"
+    description = (
+        "A run-time C++ library for working with units "
+        "of measurement and conversions between them "
+        "and with string representations of units "
+        "and measurements"
+    )
+    topics = (
+        "units",
+        "dimensions",
+        "quantities",
+        "physical-units",
+        "dimensional-analysis",
+        "run-time",
+    )
+    settings = "os", "compiler", "build_type", "arch"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "base_type": ["uint32_t", "uint64_t"],
+        "namespace": "ANY",
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "base_type": "uint32_t",
+        "namespace": None,
+    }
+    @property
+    def _min_cppstd(self):
+        return 14
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "10",
+            "clang": "7",
+            "gcc": "7",
+            "msvc": "191",
+            "Visual Studio": "15",
+        }
+    def source(self):
+        tools.get(**self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["UNITS_ENABLE_TESTS"] = "OFF"
+        tc.variables["UNITS_BASE_TYPE"] = self.options.base_type
+        if units_namespace := self.options.get_safe("namespace"):
+            tc.variables["UNITS_NAMESPACE"] = units_namespace
+        if self.options["shared"]:
+            tc.variables["UNITS_BUILD_SHARED_LIBRARY"] = "ON"
+            tc.variables["UNITS_BUILD_STATIC_LIBRARY"] = "OFF"
+        tc.variables["UNITS_CMAKE_PROJECT_NAME"] = "LLNL-UNITS"
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        self.copy("*.hpp", dst="include/units", src="units/units")
+        self.copy("*units.lib", dst="lib", keep_path=False)
+        self.copy("*.dll", dst="bin", keep_path=False)
+        self.copy("*.so", dst="lib", keep_path=False)
+        self.copy("*.dylib", dst="lib", keep_path=False)
+        self.copy("*.a", dst="lib", keep_path=False)
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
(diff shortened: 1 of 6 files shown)


recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    topics = (
+        "units",
+        "dimensions",
+        "quantities",
+        "physical-units",
+        "dimensional-analysis",
+        "run-time",
+    )
+    settings = "os", "compiler", "build_type", "arch"
```
RubenRBS, line 27:
```suggestion
    settings = "os", "compiler", "build_type", "arch"
    package_type = "library"
```

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,98 @@
...
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "base_type": ["uint32_t", "uint64_t"],
+        "namespace": "ANY",
```
RubenRBS, line 32:
```suggestion
        "namespace": [None, "ANY"],
```

Else not defining namespace will still give it a value

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,98 @@
...
+        }
+    def source(self):
+        tools.get(**self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["UNITS_ENABLE_TESTS"] = "OFF"
+        tc.variables["UNITS_BASE_TYPE"] = self.options.base_type
+        if units_namespace := self.options.get_safe("namespace"):
+            tc.variables["UNITS_NAMESPACE"] = units_namespace
```
RubenRBS, line 61: CCI recipes need to run in Python 3.6, so syntax like this is not available :/

```suggestion
        units_namespace = str(self.options.get_safe("namespace"))
        if units_namespace != "None":
            tc.variables["UNITS_NAMESPACE"] = units_namespace
```

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,98 @@
...
+        if self.options["shared"]:
+            tc.variables["UNITS_BUILD_SHARED_LIBRARY"] = "ON"
+            tc.variables["UNITS_BUILD_STATIC_LIBRARY"] = "OFF"
```
RubenRBS, line 64:
```suggestion
        tc.variables["UNITS_BUILD_SHARED_LIBRARY"] = self.options.shared
        tc.variables["UNITS_BUILD_STATIC_LIBRARY"] = not self.options.shared
```

So if upstream ever changes the defaults we don't need to modify thke recipe :)

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,98 @@
...
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        self.copy("*.hpp", dst="include/units", src="units/units")
+        self.copy("*units.lib", dst="lib", keep_path=False)
+        self.copy("*.dll", dst="bin", keep_path=False)
+        self.copy("*.so", dst="lib", keep_path=False)
+        self.copy("*.dylib", dst="lib", keep_path=False)
+        self.copy("*.a", dst="lib", keep_path=False)
```
RubenRBS, line 79: Is this not handled by the cmake install call? Does upstream not provide one?

Also, the `dest` paths in here should be of the form with os.path.join(self.package_folder, ...)`

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,98 @@
...
+    settings = "os", "compiler", "build_type", "arch"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "base_type": ["uint32_t", "uint64_t"],
+        "namespace": "ANY",
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
```
RubenRBS, line 36: Please check [this methods implemented in the cmake-project template](https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L61-L67), they are important to properly handle fpic behaviour in the different configurations

RubenRBS (commented):
Thanks a lot for taking the time to submit this to CCI, we appreciate it. I've added some comments about the recipe, but mostly it looks great :)

AbrilRBS (dismissed):
Thanks a lot! :)

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
+import os
+
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeToolchain
+from conan.tools.files import copy, rm, rmdir, get
+
+
+class UnitsConan(ConanFile):
+    name = "llnl-units"
+    license = "BSD-3"
```
uilianries, line 10:
```suggestion
    license = "BSD-3-Clause"
```

Please, always use the short for listed by SPDX: https://spdx.org/licenses/BSD-3-Clause.html

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "base_type": "uint32_t",
+        "namespace": None,
+    }
+
+    @property
+    def _min_cppstd(self):
+        return 17
```
uilianries, line 44: The project refers as C++14 here: https://github.com/LLNL/units/blob/v0.9.1/CMakeLists.txt#L35

Am I missing something?

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.la", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
+
+    def package_info(self):
+        self.cpp_info.libs = ["units"]
+        units_namespace = self.options.get_safe("namespace")
```
uilianries, line 100: It should not be an option, because it does not affect the package content, it will be same even if you pass a different value. I would suggest using a configuration instead:

```python
namespace = self.conf.get("user.llnl-units:namespace", check_type=str):
if namespace:
    self.cpp_info.defines.append(f"UNITS_NAMESPACE={namespace}")
```

uilianries (commented):
Thank you for your contribution! Some notes:

- I see there is support for header-only in the project: https://github.com/LLNL/units/blob/v0.9.1/CMakeLists.txt#L115 It can be a future improvement in the recipe as a new option. It's totally fine the current state of the recipe
- The third party is only consumed by tests, so the project has no dependencies: https://github.com/LLNL/units/tree/v0.9.1/ThirdParty

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+        self.cpp_info.defines = [f"UNITS_BASE_TYPE={self.options.base_type}"]
```
uilianries, line 101: Same for `base_type`, it can be a configuration instead.

recipes/llnl-units/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+        if units_namespace:
+            self.cpp_info.defines.append(f"UNITS_NAMESPACE={units_namespace}")
+
+        self.cpp_info.set_property("cmake_file_name", "llnl-units")
+        self.cpp_info.set_property("cmake_target_name", "llnl-units::units")
```
uilianries, line 106:
```suggestion
        self.cpp_info.set_property("cmake_file_name", "units")
        self.cpp_info.set_property("cmake_target_name", "units::units")
```


Tested locally, cmake file and target are `units` only:

```
cmake -P cmake_install.cmake
-- Install configuration: ""
-- Installing: /tmp/units/build/install/lib/libunits.a
-- Installing: /tmp/units/build/install/include/units/units.hpp
-- Installing: /tmp/units/build/install/include/units/units_decl.hpp
-- Installing: /tmp/units/build/install/include/units/unit_definitions.hpp
-- Installing: /tmp/units/build/install/include/units/units_util.hpp
-- Installing: /tmp/units/build/install/include/units/units_conversion_maps.hpp
-- Installing: /tmp/units/build/install/include/units/units_math.hpp
-- Installing: /tmp/units/build/install/include/units/commodity_definitions.hpp
-- Installing: /tmp/units/build/install/bin/units_convert
-- Installing: /tmp/units/build/install/lib/cmake/units/unitsTargets.cmake
-- Installing: /tmp/units/build/install/lib/cmake/units/unitsTargets-noconfig.cmake
-- Installing: /tmp/units/build/install/lib/cmake/units/unitsConfigVersion.cmake
-- Installing: /tmp/units/build/install/lib/cmake/units/unitsConfig.cmake
```

```
grep :: /tmp/units/build/install/lib/cmake/units/unitsTargets.cmake                                                                                                                                                                               
foreach(_cmake_expected_target IN ITEMS units::units units::header_only units::units_convert units::compile_flags_target)
# Create imported target units::units
add_library(units::units STATIC IMPORTED)
set_target_properties(units::units PROPERTIES
  INTERFACE_LINK_LIBRARIES "\$<LINK_ONLY:units::compile_flags_target>"
```

recipes/llnl-units/all/conanfile.py
AbrilRBS: Good catch!

recipes/llnl-units/all/conanfile.py
MridulS: Thanks for the review!

This allows us to use nested namespaces something like `llnl::units`.

It's a nice to have when there are so many "units" libraries out there :)

uilianries: I guess you are referring about that review, right? https://github.com/conan-io/conan-center-index/pull/24459/files/61b135395d3f8be52aeb4b2de35e607113bdacd5#diff-170923aa2ee0d9ef9d7b6df1188a6b29b070ce5e731cb984e06a90fe94de9c6e

I see, but in ConanCenterIndex package as closer are possible to the upstream. In case you see a possible name conflict, I would suggest asking to the upstream changing it first. Thank you explaining your situation 😄

recipes/llnl-units/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.15)
+
+project(test_package LANGUAGES CXX)
+
+find_package(llnl-units REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} test_package.cpp)
+
+target_link_libraries(${PROJECT_NAME} PRIVATE llnl-units::units)
```
uilianries, line 9:
```suggestion
find_package(units REQUIRED CONFIG)

add_executable(${PROJECT_NAME} test_package.cpp)

target_link_libraries(${PROJECT_NAME} PRIVATE units::units)
```

Need to update tests too.

uilianries (approved):
LGTM, thank you! 😁

czoido (approved):

AbrilRBS (approved):
Thanks!

PR #24465, opened 2024-06-27
Qt5: fix incorrectly added qtaudio_alsa and gstcamerabin plugin libs

valgur:
### Summary
Changes to recipe:  **qt/5.x.x**

#### Motivation
`qt/5.15.14` with default options and `*/*:shared=False` fails during `find_package()` in consuming projects due to `qtaudio_alsa` and `gstcamerabin` library files not being found.

These two plugins are always added on Linux in the conanfile. However, they are only enabled in the Qt5 project files if `libalsa` and `gstreamer` are available: https://github.com/qt/qtmultimedia/blob/v5.14.2/src/plugins/plugins.pro#L47-L59

#### Details
Only add these plugins when `with_libalsa` and `with_gstreamer` are enabled, respectively, as a fix.

This has probably not been caught by the CI due to the `"essential_modules": not os.getenv('CONAN_CENTER_BUILD_SERVICE')` default, which disables `qtmultimedia`.

---


RubenRBS (approved):
Thanks!

franramirez688 (approved):

Issue #24368, opened 2024-06-17
[package] raylib/5.0: Missing rcamera.h header


Ezbob:
### Description

I've tried to run the 3d camera examples from the RayLib package, but it seems that the `rcamera.h` header is missing from the package's include path. I've managed to find it in the RayLib's source so it seems like it still around. Furthermore it seems that the `raudio.h` header is also missing.

### Package and Environment Details

* Package Name/Version: **raylib/5.0**
* Operating System+version: **Windows 10 19045.4529**
* Compiler+version: **MSVC 17**
* Conan version: **conan 2.4.1**
* Python version: **Python 3.12.0**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.version=193
os=Windows

[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

Put:
```
[requires]
raylib/5.0

[generators]
CMakeDeps
CMakeToolchain

[layout]
cmake_layout
``` 
in a `conanfile.txt`.
Run `conan install . -pr:h=default -pr:b=default --build=missing` to install depedencies.

Try and compile this program with the using the `conan_toolchain.cmake`:
```
#include "raylib.h"
#include "rcamera.h"

int main(void)
{
  return 0;
}
```

### Logs

<details><summary>Click to expand log</summary>

```
MSBuild version 17.7.2+d6990bcfa for .NET Framework

  Checking Build System
  Building Custom Rule H:/c++/pewpew/CMakeLists.txt
  main.c
H:\c++\pewpew\src\main.c(15,10): fatal  error C1083: Cannot open include file: 'rcamera.h': No such file or directory [H:\c++\pewpew\build\pewpew.vcxproj]
```

</details>

RubenRBS:
Hi @Ezbob thanks a lot for taking the time to report your issue, we appreciate it :)

I don't think this is a recipe issue, as per the upstream cmakelist in https://github.com/raysan5/raylib/blob/master/src/CMakeLists.txt#L24, you can see that they set a minimal collection of headers as part of their public API, so it's not clear to me if the expectations are to be able to use such headers externally

As you mention, to confuse things even more, they do show up in the examples, so maybe we need to ask upstream first?

Ezbob:
Hello, and thanks for your quick response, and sorry for my slow one.
Hmm... Yes I can see that it looks like they don't expose `rcamera.h` as a public header...
[...]

PR #24472, opened 2024-06-29
raylib: Add some useful headers to res folder

AbrilRBS:
### Summary
Changes to recipe:  **raylib/[*]**

#### Motivation

As explained in https://github.com/conan-io/conan-center-index/issues/24368, the library does not package some useful files that users might want to use as part of their build. This is similar to the imgui situation, so the solution is also similar: Add those files to the res folder, where users can then copy them from

#### Details

The linked issue and subsequent upstream issue have all the relevant info

As a draft until I get a chance to update the hook that will fail

cc @Ezbob

### DIFF
--- a/recipes/raylib/all/conanfile.py
+++ b/recipes/raylib/all/conanfile.py
@@ -91,6 +91,13 @@ class RaylibConan(ConanFile):
         rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
 
+        res_include_folder = os.path.join(self.package_folder, "res", "include")
+        copy(self, pattern="rcamera.h", src=os.path.join(self.source_folder, "src"), dst=res_include_folder, keep_path=False)
+        copy(self, pattern="rgestures.h", src=os.path.join(self.source_folder, "src"), dst=res_include_folder, keep_path=False)
+        copy(self, pattern="rlgl.h", src=os.path.join(self.source_folder, "src"), dst=res_include_folder, keep_path=False)
+        copy(self, pattern="raymath.h", src=os.path.join(self.source_folder, "src"), dst=res_include_folder, keep_path=False)
+        copy(self, pattern="raudio.h", src=os.path.join(self.source_folder, "src"), dst=res_include_folder, keep_path=False)
+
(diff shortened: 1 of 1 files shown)


Ezbob:
Hey @AbrilRBS , big thanks for this. 
I have a noob question though, how do you use/copy these headers in your conan file from a consumers perspective?

recipes/raylib/all/conanfile.py
```diff
@@ -127,6 +134,9 @@ def package_info(self):
         elif self.settings.os == "Windows":
             self.cpp_info.system_libs.append("winmm")
 
+        # Some useful files are not packaged by default
+        self.cpp_info.resdirs = ["res/include"]
```
uilianries, line 138:
```suggestion
        self.cpp_info.resdirs = ["res/include"]
        if self.conf.get("user.raylib:include_res", , default=False, check_type=bool):
            self.cpp_info.includedirs.append("res/include")
```

So users can use res folder directly without affecting the package ID.

AbrilRBS:
Hi @Ezbob, not a noob question at all, I should have been more clear as to how to proceed, sorry about that!

We're going to make this available thru a conf as @uilianries suggested - I'll post a how-to after I commit the changes :)

AbrilRBS:
With the current approach (pending dismissal from someone on the team if they don't like it) would be for you to compile your package as normal, but add the `user.raylib:include_res=True` conf, either in your CLI with `-c="user.raylib:include_res=True"`, or in your profile:
```
[conf]
user.raylib:include_res=True
```

This will make it so the recipe adds the new include path to your project, giving you access to those files.

The alternative approach if you don't need everything, or only need certain files, would be NOT to define the conf, and instead in your recipe's `generate()` method, have:

```py
def generate(self):
    # ... The rest of your generate method ...
    copy(self, pattern="rcamera.h", src=self.dependencies["raylib"].cpp_info.resdirs[0], dst=os.path.join(self.build_folder, "path-where-you-want-thefiles")
```

This would only copy rcamera.h and would give you the liberty of moving it to a place of your liking - hope this helps :)

Ezbob:
@AbrilRBS Oh that look nice! Huge thanks for this!

AbrilRBS:
Superseeded by https://github.com/conan-io/conan-center-index/pull/24585

PR #24476, opened 2024-07-01
diligent-core: add package_type, clean up a bit

valgur:
### Summary
Changes to recipe:  **diligent-core/***

#### Motivation
Includes some changes from #21522 that did not get merged in favor of the superseding PR.

#### Details
- Adds `package_type`.
- Uses `CMAKE_PROJECT_DiligentCore_INCLUDE=conan_deps.cmake` instead of patching.
- Bumps xxhash and xkcbcommon to better align with the versions used in other recipes on CCI.
- Adds `self.cpp_info.resdirs = ["res"]`
- Unvendor GLEW.
- Don't remove the installed licenses for vendored libraries.
- Install `BuildUtils.cmake`, so `diligent-tools` and `diligent-fx` don't need to include it with their recipes (https://github.com/conan-io/conan-center-index/blob/master/recipes/diligent-tools/all/BuildUtils.cmake).
- The rest of the changes are purely cosmetic.

---


RubenRBS (approved):
Thanks a lot for taking the time to clean this recipe up @valgur, really appreciated :)

Ignore the unexpected CI issue, we're working on it, will retrigger the ci once it's ready

recipes/diligent-core/all/conanfile.py
```diff
@@ -45,36 +48,41 @@ def _minimum_compilers_version(self):
...
     def export_sources(self):
-        copy(self, "conan_deps.cmake", src=self.recipe_folder, dst=os.path.join(self.export_sources_folder, "src"), keep_path=False)
+        copy(self, "conan_deps.cmake", self.recipe_folder, os.path.join(self.export_sources_folder, "src"))
         export_conandata_patches(self)
-        
+
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def package_id(self):
```
MikeLankamp, line 76: Hi @valgur, since you're touching this:

I've also mentioned this in #23347, but the override of `self.info.settings.compiler.runtime` as done here seems to cause problems with the compatible package finder. When using diligent-core with a `compiler.cppstd=17` profile, conan complains:
```
diligent-core/2.5.2: Compatible package ID 0b6ce91bc61db9be4c90e8138b36d3ae166dc8be equal to the default package ID: Skipping it.
diligent-core/2.5.2: Checking 3 compatible configurations
diligent-core/2.5.2: Compatible configurations not found in cache, checking servers
diligent-core/2.5.2: 'f9de1b468d6419cb25174c1f3ce2c2b9866a247f': compiler.cppstd=14
diligent-core/2.5.2: Main binary package '0b6ce91bc61db9be4c90e8138b36d3ae166dc8be' missing. Using compatible package 'f9de1b468d6419cb25174c1f3ce2c2b9866a247f': compiler.cppstd=14
ERROR: Invalid setting 'MD/MDd' is not a valid 'settings.compiler.runtime' value.
Possible values are ['static', 'dynamic']
Read "http://docs.conan.io/2/knowledge/faq.html#error-invalid-setting"
```
Apparently Conan re-validates the info from package_id when processing compatible packages.

This makes diligent-core unusable right now unless your profile _exactly_ targets C++14. I'm not sure if the problem is with this recipe (why combine release/debug for package_id?), or with Conan (since `self.info` [isn't supposed to be validated](https://docs.conan.io/2/reference/conanfile/methods/package_id.html#partial-information-erasure)?).

If the problem's with the recipe, I hope you could fix that with this PR as well.

RubenRBS: Looking into this from a Client perspective, thanks for the heads-up

RubenRBS: Tracking in https://github.com/conan-io/conan/pull/16575

AbrilRBS: Hi @valgur @MikeLankamp this turned out to indeed be a bug in the Conan client - we're releasing a fix in the 2.5.0 release, due early this week, again, thanks for the heads-up, we appreciate it :)

As for the changes in the recipe itself, @memsharded and I have some concerns over how the runtime discarding debug info is even valid, we'll talk with the team tomorrow and let you know once we know more!

MikeLankamp: Many thanks for fixing it, this was blocking my workflow. I almost considered rewriting my code using Diligent to use Vulkan because of this 😅 Looking forward to the 2.5.0 release!

AbrilRBS: Just a heads-up @MikeLankamp, Conan 2.5.0 has now been released. Let me know if using it fixes your issue! :)

MikeLankamp: Hi @AbrilRBS, it did indeed! Many thanks!

recipes/diligent-core/all/conanfile.py
```diff
@@ -45,36 +48,41 @@ def _minimum_compilers_version(self):
...
-        if visual.is_msvc(self.info):
-            if visual.is_msvc_static_runtime(self.info):
+        if is_msvc(self.info):
+            if is_msvc_static_runtime(self.info):
                 self.info.settings.compiler.runtime = "MT/MTd"
             else:
                 self.info.settings.compiler.runtime = "MD/MDd"
```
AbrilRBS, line 81: I think this might be problematic. Do you have any extra insight into why this was needed in the first place in https://github.com/conan-io/conan-center-index/pull/7804? Or how can this library work when discarding such runtime info? If we're not sure about this, we should remove this erasure - at worst we only need to compile a few more binaries

PR #24480, opened 2024-07-01
Feature/add instinct cpp

RobinQu:
### Summary
Changes to recipe:  **instinct-cpp/0.1.5**

#### Motivation
Make instinct-cpp available on conan center

#### Details
First recipe of instinct-cpp  v0.1.5

---

### DIFF
--- a/recipes/instinct-cpp/all/conanfile.py
+++ b/recipes/instinct-cpp/all/conanfile.py
@@ -18,7 +18,7 @@ class InstinctCppConan(ConanFile):
                    "and built-in utilities for developing AI Agent applications (RAG, Chatbot, Code interpreter) powered by language models. "
                    "Call it langchain.cpp if you like.")
     license = "Apache-2.0"
-    url = "https://github.com/RobinQu/instinct.cpp"
+    url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/RobinQu/instinct.cpp"
     topics = ("llm", "agent", "genai")
     package_type = "library"
--- a/recipes/instinct-cpp/all/conanfile.py
+++ b/recipes/instinct-cpp/all/conanfile.py
@@ -1,0 +1,202 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rm, rmdir
+from conan.tools.microsoft import check_min_vs, is_msvc, is_msvc_static_runtime
+from conan.tools.scm import Version
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class PackageConan(ConanFile):
+    name = "instinct-cpp"
+    version = "0.1.5"
+    description = ("instinct.cpp provides ready to use alternatives to OpenAI Assistant API a"
+                   "nd built-in utilities for developing AI Agent applications (RAG, Chatbot, Code interpreter) powered by language models. "
+                   "Call it langchain.cpp if you like.")
+    license = "Apache-2.0"
+    url = "https://github.com/RobinQu/instinct.cpp"
+    homepage = "https://github.com/RobinQu/instinct.cpp"
+    topics = ("LLM", "Agent", "GenAI")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_duckdb": [True, False],
+        "with_exprtk": [True, False],
+        "with_pdfium": [True, False],
+        "with_duckx": [True, False]
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_duckdb": True,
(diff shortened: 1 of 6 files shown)


RobinQu:
I think the access request is approved. How should we proced this PR?

AbrilRBS:
@RobinQu Sorry for the delay on the review: I'll push some cleanup changes to your recipe and comment on them in a bit :)

recipes/instinct-cpp/all/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+        if self.options.with_duckdb:
+            self.requires("duckdb/0.10.2", transitive_headers=True)
+
+        self.requires("uriparser/0.9.7", transitive_headers=True)
+        self.requires("crossguid/0.2.2", transitive_headers=True)
+        self.requires("protobuf/5.27.0", transitive_headers=True)
+        self.requires("reactiveplusplus/2.1.1", transitive_headers=True)
+        self.requires("icu/74.1")
+        self.requires("tsl-ordered-map/1.1.0", transitive_headers=True)
+        self.requires("fmt/10.2.1", transitive_headers=True)
```
AbrilRBS, line 86: The `transitive_headers` would only be needed if those includes were part of the public API of the project, usually there are a few, but it's a bit suprising that every dependency has it. Does every dependency leak to the consumers?

recipes/instinct-cpp/all/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+        
+        def _ensure_dep_has_option(dep, option, value):
+            if self.dependencies[dep].options[option] != value:
+                raise ConanInvalidConfiguration(f'{self.ref} needs -o="{dep}/*:{option}={value}"')
+
+        _ensure_dep_has_option("duckdb", "with_httpfs", self.default_options["duckdb/*:with_httpfs"])
```
AbrilRBS, line 121: This is bad news if it's not the default for those packages: It will make it so nothing gets built in our CI, as every configuration would still not find these packages - cci only builds the default options for each package (minus shared/header_only options where we combine them). If these are not the defaults, we'll need successful compilation logs of a few different configurations before we're confident about merging this

I'll provide apple-clang ones once the recipe is ready to go to save you some time :)

recipes/instinct-cpp/all/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+
+        _ensure_dep_has_option("duckdb", "with_httpfs", self.default_options["duckdb/*:with_httpfs"])       
+        _ensure_dep_has_option("icu", "with_extras", self.default_options["icu/*:with_extras"])
+        _ensure_dep_has_option("icu", "data_packaging", self.default_options["icu/*:data_packaging"])
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
```
AbrilRBS, line 129: A bit suprising for a library with so many dependencies to have these little amount of options, it's great! :)

recipes/instinct-cpp/all/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if self.options.with_duckx:
+            self.requires("duckx/1.2.2", transitive_headers=True)
+
+        if self.options.with_exprtk:
+            self.requires("exprtk/0.0.2", transitive_headers=True)
+
+        self.requires("llama-cpp/b3040", transitive_headers=True)
+        self.requires("cpp-httplib/0.15.3", transitive_headers=True)
+        self.requires("cli11/2.4.1")
```
AbrilRBS, line 106: I had to remove the gtest test_require from here:

If we want to give users the ability to run the tests, we have a way to control that with the `skip_tests` conf, but usually CCI packages simply just don't build/run them, so no need to add test_requirements :)

recipes/instinct-cpp/all/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+    @property
+    def _min_cppstd(self):
+        return 20
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "15",
+            "clang": "17",
+            "gcc": "12"
```
AbrilRBS, line 55: Has this library been tested on Windows? we would add a min msvc if it's possible to run there :)

recipes/instinct-cpp/all/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+        "with_pdfium": [True, False],
+        "with_duckx": [True, False]
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_duckdb": True,
+        "with_exprtk": True,
+        "with_pdfium": True,
+        "with_duckx": True,
```
AbrilRBS, line 40: It would also be great to have successful compilation logs when turning off some or all of these options, just to check that it works :)

AbrilRBS (commented):
Thanks! I've now pushed the changes and made some comments, let me know what you think :)

RobinQu:
@AbrilRBS Thanks for your response. I will look into these issues after a 4-day vocation.

AbrilRBS:
Hope you have a good one :) Ping me once you address them/if you need any help!

PR #24506, opened 2024-07-03
Feature/add open usd

EstebanDugueperoux2:
### Summary

#### Details

---


AbrilRBS:
Building the missing binaries now, will let you know once they are ready :)

AbrilRBS:
The binaries should be available for the next build :)

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,526 @@
...
+                self.options.build_embree_plugin = False
+        if self.options.build_renderman_plugin:
+            if not self.options.build_imaging:
+                self.options.build_renderman_plugin = False
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        if self.settings.os == "Windows":
+             self.options["openssl"].shared = True
```
AbrilRBS, line 143: you will need to validate this option in the validate() method if this is a hard requirement, as users can override recipe choices

AbrilRBS: Check here for an example of a nice message to give to users https://github.com/conan-io/conan-center-index/pull/24480/files/985651bff18949278eb0da17f8ee32710c20c377#diff-d0428b639a2f228893f252d9582c8726cb6d7dffb159fea13ddbac1ac3772f65R119 (plus an explanation, if necessary)

EstebanDugueperoux2: Hi, I have removed this part.
I leave this PR in draft as it is not yet ready for review.
Sorry.

EstebanDugueperoux2: Ready for review.

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+        if self.settings.os == "Macos" and not self.options["opensubdiv"].with_metal:
```
AbrilRBS, line 173:
```suggestion
        if self.settings.os == "Macos" and not self.dependencies["opensubdiv"].options.with_metal:
```

EstebanDugueperoux2:
Hi @AbrilRBS , is it ok for you now?

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rm, rmdir
+from conan.tools.scm import Version
+import os
+
+required_conan_version = ">=1.53.0"
+
+class OpenUSDConan(ConanFile):
+    name = "openusd"
+    description = "Universal Scene Description"
+    license = "LicenseRef-LICENSE.txt"
```
AbrilRBS, line 16:
```suggestion
    license = "DocumentRef-LICENSE.txt:LicenseRef-Modified-Apache-2.0-License"
```

EstebanDugueperoux2:
Hi @AbrilRBS , is it ok for you now?

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+        if self.options.build_usd_imaging and not self.options.build_imaging:
+            self.options.build_usd_imaging = False
+        self.options.build_gpu_support = self.options.enable_gl_support or is_apple_os(self) or self.options.enable_vulkan_support
+        self.options["opensubdiv"].with_opengl = self.options.enable_gl_support
```
AbrilRBS, line 96: Is this a hard requirement? It should be checked in the validate() method if so

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+        "enable_hdf5_support": [True, False],
+        "build_draco_plugin": [True, False],
+        "enable_osl_support": [True, False],
+        "build_animx_tests": [True, False],
+        "enable_python_support": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": False,
+        "build_usd_tools": True,
```
AbrilRBS, line 49: These are quite a lot of options! We would appreciate if you could submit successful compilation logs some of the most important versions being turned on, as to have traceability over what currently works at this point, thanks!

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+                self.requires("opengl/system")
+            if self.options.enable_ptex_support:
+                self.requires("ptex/2.4.2")
+            if self.options.enable_openvdb_support:
+                self.requires("openvdb/11.0.0")
+            if self.options.build_embree_plugin:
+                self.requires("embree3/3.13.5")
+        if self.options.build_renderman_plugin:
+            # TODO: add a recipe for renderman
+            self.requires("renderman/x.y.z")
```
AbrilRBS, line 147: Better to have this commented until the recipe is present

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+        if self.options.build_draco_plugin:
+            self.requires("draco/1.5.6")
+        if self.options.enable_materialx_support:
+            self.requires("materialx/1.38.10", transitive_headers=True)
+        if self.options.enable_osl_support:
+           # TODO: add osl to conan center (https://github.com/AcademySoftwareFoundation/OpenShadingLanguage)
+            self.requires("openshadinglanguage/1.13.8.0")
+        if self.options.build_animx_tests:
+           # TODO: add animx to conan center (https://github.com/Autodesk/animx/)
+            self.requires("animx/x.y.z")
```
AbrilRBS, line 161: Same here

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+        tc.variables["PXR_ENABLE_OSL_SUPPORT"] = self.options.enable_osl_support
+        tc.variables["PXR_BUILD_ANIMX_TESTS"] = self.options.build_animx_tests
+
+        tc.generate()
+
+        tc = CMakeDeps(self)
+        tc.generate()
+
+        tc = VirtualBuildEnv(self)
+        tc.generate(scope="build")
```
AbrilRBS, line 236: build could be skipped here, it's the default for `VirtualBuildEnv` unless I'm missing something

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,490 @@
...
+
+    def _patch_sources(self):
+        rmdir(self, os.path.join(self.source_folder, "cmake", "modules"))
+                              
+        if self.options.enable_materialx_support:
```
AbrilRBS, line 241: As far as I understand, these are patches to use the target names that conan is producing so that the recipe consumes the same ones, righ?

Maybe patching could be avoided by modifying the `cmake_file_name`/`cmake_target_name` in the `CMakeDeps` generator for these dependencies?

AbrilRBS (commented):
Thanks! Some more comments about the changes, the recipe itself mostly looks good :)

AbrilRBS:
@EstebanDugueperoux2 the bot will retrigger the PR automatically once the maintenance mode is over - we decided to keep it around this weekend until we fix some issues we found. Note that closing and reopening PRs is discouraged, if not for anything other than GitHub automatically sending them to the Done state in our internal project instead of how we tagged it, which can then make it so that we forget to circle back to it in a timely manner :)

EstebanDugueperoux2:
@AbrilRBS  ok.
Any news since?

EstebanDugueperoux2:
@AbrilRBS I don't understand why I have a conflict on boost only with conanv2 then I have created another PR to bump this openvdb requirement https://github.com/conan-io/conan-center-index/pull/24711 but I don't understand why its CI is red.

EstebanDugueperoux2:
I leave this PR in draft because it remains an issue with static build of openusd where plugin seems cause segfault.

EstebanDugueperoux2:
Hi @AbrilRBS,

The KB-H013 hook (https://github.com/conan-io/hooks/blob/master/hooks/conan-center.py#L1086) doesn't considers folders defined dynamically in self.cpp_info.libdirs and in the case of this recipe when some plugins options are enabled, their associated shared libraries are installed in plugin folder hence the "self.cpp_info.libdirs.append("plugin/usd/")" call in https://github.com/EstebanDugueperoux2/conan-center-index/blob/feature/AddOpenUSD/recipes/openusd/all/conanfile.py#L290.

Could it be possible to have this hook rewritten to be based on post_package_info to consider self.cpp_info.libdirs dynamically?


Regards.

EstebanDugueperoux2:
Hi @AbrilRBS ,

To fix this recipe, I need to integrate https://github.com/conan-io/conan-center-index/pull/25532 but its pipelines seems blocked.
Could you unlock it and make a review?

Thanks in advance.

jcar87:
> In [build logs](https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-24506/6/package_build_logs/build_log_openusd_25_02_f280184f5f3b7791009f90372e9db462_7b82da7ea4cc5f6a1ee7b904b93e072be6f83f52.success.txt), we see that hioAvif.so is correctly installed.

I think it's because the library is called `hioAvif.so` instead of `libhioAvif.so` - pinging @franramirez688  in case he has more insights (I was under the impression we would fall back to looking for the file without a prefix as well - but I could be wrong)

franramirez688:
> I think it's because the library is called hioAvif.so instead of libhioAvif.so - pinging @franramirez688 in case he has more insights (I was under the impression we would fall back to looking for the file without a prefix as well - but I could be wrong)

Hi @EstebanDugueperoux2, as @jcar87 said above, the problem is related to that name. For instance, this works on my side:

```python
            if self.settings.os == "Linux" and self.options.shared:
                self.cpp_info.components["hioAvif"].libs = ["hioAvif.so"]
            else:
                self.cpp_info.components["hioAvif"].libs = ["hioAvif"]
```
The problem is that `CMakeDeps` is using `find_library` under the hood so it needs the full name because of the missing prefix `lib`.

I also recommend trying the new incubating [CMakeConfigDeps](https://docs.conan.io/2/incubating.html#new-cmakeconfigdeps-generator). It is under development, but any feedback is more than welcome 😁

EstebanDugueperoux2:
Hi @franramirez688,

I have done a first test locally on this recipe by replacing CMakeDeps usage by CMakeConfigDeps one, and I have following error:

`
openusd/25.02: WARN: Using the new CMakeConfigDeps generator, behind the 'tools.cmake.cmakedeps:new' gate conf. This conf will change next release, breaking, so use it only for testing and dev
ERROR: openusd/25.02: Error in generate() method, line 280
	tc.generate()
	AttributeError: 'NoneType' object has no attribute 'package_type'

`

Regards.

franramirez688:
> openusd/25.02: WARN: Using the new CMakeConfigDeps generator, behind the 'tools.cmake.cmakedeps:new' gate conf. This conf will change next release, breaking, so use it only for testing and dev
ERROR: openusd/25.02: Error in generate() method, line 280
tc.generate()
AttributeError: 'NoneType' object has no attribute 'package_type'

Thanks for the feedback @EstebanDugueperoux2
I'll look into that 😉

EstebanDugueperoux2:
Hi @jcar87 @AbrilRBS this MR is ready for review.

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,524 @@
...
+        "enable_gl_support": [True, False],
+        "build_gpu_support": [True, False],
+        "enable_ptex_support": [True, False],
+        "enable_openvdb_support": [True, False],
+        "build_alembic_plugin": [True, False],
+        "enable_hdf5_support": [True, False],
+        "build_draco_plugin": [True, False],
+        "enable_osl_support": [True, False],
+        "build_animx_tests": [True, False],
+        "enable_python_support": [True, False],
```
ErniGH: This recipe has too many options. We usually prefer to keep recipes minimal, providing only the essential options needed to build the library. Additional options can then be added iteratively based on how the library is used. Having so many options increases maintenance costs and poses a significant risk, as not all possible combinations are tested in the CI.

Please reduce these options to the minimum necessary to build the library.

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,524 @@
...
+                self.options.build_usdview = False
+            elif not self.options.enable_python_support:
+                self.options.build_usdview = False
+            elif not self.options.build_gpu_support:
+                self.options.build_usdview = False
+        if self.options.build_embree_plugin:
+            if not self.options.build_imaging:
+                self.options.build_embree_plugin = False
+            elif not self.options.build_gpu_support:
+                self.options.build_embree_plugin = False
```
ErniGH: The `config_options` should not modify user-defined options on the fly, as this behavior can be unexpected for end users. Options that are incompatible or dependent on each other should be handled in the `validate` method instead.

Please move any necessary incompatibility checks to the `validate` method.

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,524 @@
...
+        if self.options.enable_gl_support:
+            self.options["opensubdiv/*"].with_opengl = self.options.enable_gl_support
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.options.enable_python_support:
+            self.requires("boost/1.87.0")
+        # openusd doesn't support yet recent release of onetbb, see https://github.com/PixarAnimationStudios/OpenUSD/issues/1471
```
ErniGH: Now supports 2021.9, we have it in conan-center-index

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,524 @@
...
+                self.requires("embree3/3.13.5")
+        if self.options.build_alembic_plugin:
+            self.requires("alembic/1.8.6")
+            if self.options.enable_hdf5_support:
+                self.requires("hdf5/1.14.5")
+        if self.options.build_draco_plugin:
+            self.requires("draco/1.5.6")
+        if self.options.enable_materialx_support:
+            self.requires("materialx/1.39.1", transitive_headers=True)
+        # if self.options.enable_osl_support:
```
ErniGH: The same issue occurs here: supporting so many options from the beginning adds a great deal of complexity to the recipe. Looking at the upstream README, it seems that the library doesn't actually require so many dependencies for common usage.

Please simplify the requirements.

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,524 @@
...
+            tc.set_property("materialx::MaterialXCore", "cmake_target_name", "MaterialXCore")
+            tc.set_property("materialx::MaterialXFormat", "cmake_target_name", "MaterialXFormat")
+            tc.set_property("materialx::MaterialXGenShader", "cmake_target_name", "MaterialXGenShader")
+            tc.set_property("materialx::MaterialXRender", "cmake_target_name", "MaterialXRender")
+            tc.set_property("materialx::MaterialXGenGlsl", "cmake_target_name", "MaterialXGenGlsl")
+            
+        tc.generate()
+
+        tc = VirtualBuildEnv(self)
+        tc.generate()
```
ErniGH: This is no longer necessary in Conan2

ErniGH (changes requested):
It seems that the validation checks are preventing the CI from running. Please add build logs.

EstebanDugueperoux2:
Hi @ErniGH  thanks for the review, I see no build because of missing build of opensubdiv dependency.
OpenUSD works only with opensubdiv having onetbb and opengl enabled.
I have enabled opengl for opensubdiv in https://github.com/conan-io/conan-center-index/pull/25532/files but it seems opensubdiv:with_tbb should also be enabled.

ErniGH:
@EstebanDugueperoux2 

In Conan Center, the full matrix with all the different option combinations for each library is not calculated and stored. The only configuration that is built is shared.

In this case, the library requires dependencies with specific options that are not enabled by default:
```
self.options["opensubdiv/*"].with_tbb = True  
self.options["opensubdiv/*"].with_opengl = True
```
As a result, those packages will never be found in Conan Center.
That’s why we usually include logs for the different option combinations to ensure everything builds correctly.

On another note, these options shouldn’t be modified on the fly — it’s better to remove them from here and instead add two messages to the validate() method:
```
openusd requires -o opensubdiv/*:with_tbb=True
openusd requires -o opensubdiv/*:with_opengl=True
```
This validation will prevent any version from being built unless those options are explicitly set, so the build logs will still be necessary.

EstebanDugueperoux2:
Hi @ErniGH,

this is the logs of a Linux build of shared libraries 
[build_shared_debug.log](https://github.com/user-attachments/files/20909726/build_shared_debug.log)

Regards.

recipes/openusd/all/conanfile.py
```diff
@@ -0,0 +1,299 @@
...
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder=".")
```
ErniGH:
```suggestion
        cmake_layout(self, src_folder="src")
```

ErniGH (changes requested):
Hi @EstebanDugueperoux2 , I tested locally and I can build the library, but the test is failing. I tried on MacOS, Windows, and Linux, none of them work. 

I think it would be better to try creating a simpler test that doesn't use system paths or files that could cause errors. On the other hand, the test doesn't need to verify the library's functionality, but rather check that it has been built and linked correctly.

[mac-openusd.log](https://github.com/user-attachments/files/20983227/mac-openusd.log)

EstebanDugueperoux2:
Hi @ErniGH , I have the "error while loading shared libraries: libtbbmalloc_debug.so.2: cannot open shared object file: No such file or directory" error only with static build on test_package step.
onetbb conan recipe is typed as shared-library, perhaps it is the origin of the issue.
What do you think?

Regards.

ErniGH:
Hi @EstebanDugueperoux2, I would need to see the full log. Could you share it?

EstebanDugueperoux2:
[build_shared_release_linux.log](https://github.com/user-attachments/files/21203884/build_shared_release_linux.log)
Hello @ErniGH,

I have updated the recipe to forbid static build because of onetbb recipe not allowing it (https://github.com/conan-io/conan/issues/15040#issuecomment-1790753531) and onetbb is a hard requirement of OpenUSD.
I see an attempt to restore it in https://github.com/conan-io/conan-center-index/pull/23296.
Do conan to override this package_type in profile for example?

See in attachment the logs for shared build on Linux.

P.S.: for information OpenUSD could replace its homemade python script build by conanfile (https://github.com/PixarAnimationStudios/OpenUSD/issues/2262).

Regards.

EstebanDugueperoux2:
Hi @perseoGI @AbrilRBS @jcar87 ,

could you review this PR before it is automatically closed?

Regards.

Latios96:
@EstebanDugueperoux2 thank you so much for your effort, I hope this gets reviewed and merged at some point

EstebanDugueperoux2:
Hi @Latios96,

Are you from conanio team?
Can you make a review?
What I remember from this PR is that as onetbb doesn't supports static build, openusd cannot be build statically with conan.

Regards.

Latios96:
I'm sorry, I'm not from the conanio team, I just wanted to thank you and express my interest in this PR

EstebanDugueperoux2:
Hi @ErniGH @AbrilRBS @jcar87 could I have a review of this PR?
One topic to discuss is the ability to have static build, especially of onetbb, even if not officially supported by onetbb team.

ErniGH:
@EstebanDugueperoux2 

I've reviewed this recipe again, but I can’t get it to work on macOS. Additionally, on Windows there are two validates that prevent it from building. On Linux, I can get it to build, but the test package fails locally.

Are you able to build the library for these systems? The CI isn’t building anything, and we need working build logs to verify that the recipe is working correctly.

EstebanDugueperoux2:
dHi @ErniGH,

Sorry, since the openusd release upgrade, I have forget to check the changelog and a lib has been removed.
Now on Linux it is ok with `conan create . --version 25.11 -o "openusd/*:shared=True" -o opensubdiv/*:with_tbb=True` command.

To ease review of build failures, I propose to enable onetbb requirement for opensubdiv by default, as done in upstream (https://github.com/conan-io/conan-center-index/pull/28883).
This is not a breaking change since the single impact is to have osdcpu library in addition.
Opensubdiv seems only used by this PR in conan-center-index, 

Regards.

EstebanDugueperoux2:
Hi @ErniGH @AbrilRBS,

How can I help in the progress of this PR?

Regards.

tuccio:
I tried to build it myself on macOS and bumped into this error while linking `libusd_hdSt`

```
  "OpenSubdiv::v3_7_0::Osd::MTLPatchShaderSource::GetPatchBasisShaderSource()", referenced from:
      pxrInternal_v0_25_11__pxrReserved__::_GetOSDPatchBasisShaderSource() in codeGen.cpp.o
  "OpenSubdiv::v3_7_0::Osd::MTLPatchShaderSource::GetPatchDrawingShaderSource()", referenced from:
      pxrInternal_v0_25_11__pxrReserved__::HdSt_CodeGen::Compile(pxrInternal_v0_25_11__pxrReserved__::HdStResourceRegistry*) in codeGen.cpp.o
c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

My fix was adding adding `osdgpu` to `OPENSUBDIV_LIBRARIES`, i.e.

```
tc.variables["OPENSUBDIV_LIBRARIES"] = "OpenSubdiv::osdcpu;OpenSubdiv::osdgpu"
```

and adding the respective aliases.

That fixed the error on macOS at least.

EstebanDugueperoux2:
Hi @tuccio,

I have been notified of a comment of you but it seems you have removed it.
I have updated this PR about your remark to build correctly on Mac.

Regards.

EstebanDugueperoux2:
OpenUSD is available in vcpkg (https://vcpkg.io/en/package/usd), it would be nice to have it also through conan-center-index.

EstebanDugueperoux2:
ping @AbrilRBS @ErniGH

PR #24516, opened 2024-07-04
plutovg: add version 0.0.1, remove pre release versions

toge:
### Summary
Changes to recipe:  **plutovg/0.0.1**

#### Motivation
This is the initial release.

#### Details
While PlutoVG start to use meson instead of CMake, I created new `all` directory nad remove old-`all` direvtory to `prerelease`.

---


AbrilRBS (commented):
If the old versions are pre-releases, I'd also be fine with outright removing them, the binaries will still be around, but it will be less maintenance overhead in the future :) wdyt @toge?

toge:
@AbrilRBS 
Several pre-release versions are used by lunasvg.
I think that if we don't remove the old version dependencies from lunasvg, it will cause inconvenience. 
Can we remove these versions without that fix?

AbrilRBS:
> I think that if we don't remove the old version dependencies from lunasvg, it will cause inconvenience.
Can we remove these versions without that fix?

@toge sorry for the delay! We don't necessarily need to keep the version if lunasvg depends on them, as the binaries will remain, and I don't expect bugfixes in the recipe for such an old version to be expected. We can update lunasvg in a follow-up PR to bump its dependency :)

toge:
@AbrilRBS 
Thank you for your comment!
I delete pre release versions.

perseoGI (dismissed):
LGTM, thanks for cleaning up pre-releases, IMHO they did not make much sense to keep them (as Abril said, binaries will remain in Artifactory)

AbrilRBS (approved):

recipes/plutovg/all/conanfile.py
```diff
@@ -24,46 +28,65 @@ class PlutoVGConan(ConanFile):
...
-        self.settings.rm_safe("compiler.libcxx")
         self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
 
     def layout(self):
-        cmake_layout(self, src_folder="src")
+        basic_layout(self, src_folder="src")
+
+    def build_requirements(self):
+        self.tool_requires("meson/1.2.3")
```
AbrilRBS, line 48:
```suggestion
        self.tool_requires("meson/[>=1.2.3 <2]")
```

perseoGI (approved):

PR #24520, opened 2024-07-04
Add SVT-JPEG-XS package

irieger:
### Summary
Changes to recipe:  **libsvtjpegxs/0.9.0**

#### Motivation
I'm currently experimenting with state of the art image/video compression algorithms and thus JPEG-XS with its current only open implementation is of interest to me and Intel I think has a good track record of providing optimized implementation of image algorithms.

#### Details
Add recipe for libsvtjpegxs (https://github.com/OpenVisualCloud/SVT-JPEG-XS). Naming following the name of SVT-AV1 which is provided as libsvtav1 and uses the same scheme for the actual compiled libraries.

---


uilianries (commented):
@irieger Thank you for your PR! I did some changes over your PR:

- The license should follow SPDX name: https://spdx.org/licenses/BSD-2-Clause-Patent.html
- Added a patch to let Conan manage fPIC
- Set Conan 2.x as minimal, so we can use implements feature
- CMake required by the upstream is 3.5, so no need to tool requires
- I did not find any dependency in the project, so I removed requirements
- Skipped building apps. In case you need them we can update the PR
- Rename the recipe the project name. I did not find any other package manager supporting this project
- Updated the test package to consume a method, so we can prove the linkage
- Exposed the definition DEF_DLL when consuming DLL on windows (did not check locally) 
- Fixed the pc config file name according to the upstream
- Adapted the bin folder to point to lib, so we can find the installed DLL
- The library is pure-C files, so does not need cppstd neither libcxx

uilianries (dismissed):
LGTM. Reviewed build logs.

irieger:
> @irieger Thank you for your PR! I did some changes over your PR:
> 
>     * The license should follow SPDX name: https://spdx.org/licenses/BSD-2-Clause-Patent.html
> 
>     * Added a patch to let Conan manage fPIC
> 
>     * Set Conan 2.x as minimal, so we can use implements feature
> 
>     * CMake required by the upstream is 3.5, so no need to tool requires
> 
>     * I did not find any dependency in the project, so I removed requirements
> 
>     * Skipped building apps. In case you need them we can update the PR
> 
>     * Rename the recipe the project name. I did not find any other package manager supporting this project
> 
>     * Updated the test package to consume a method, so we can prove the linkage
> 
>     * Exposed the definition DEF_DLL when consuming DLL on windows (did not check locally)
> 
>     * Fixed the pc config file name according to the upstream
> 
>     * Adapted the bin folder to point to lib, so we can find the installed DLL
> 
>     * The library is pure-C files, so does not need cppstd neither libcxx

Hey @uilianries,

thanks for picking this up. I played a bit but quickly realized that the project has some drawbacks like only supporting AVX so no macOS arm support etc. and looks sadly quite abandoned, so I didn't move forward with it. But if you want we can try to finalize it.

Currently it looks like somehow the CI is somehow stuck?

uilianries:
> Currently it looks like somehow the CI is somehow stuck?

Yes, the CI has been updated some weeks ago, and may have few bugs: https://github.com/conan-io/conan-center-index/discussions/25772

I'll be checking asap.

> project has some drawbacks like only supporting AVX so no macOS arm support etc

True, I was able to cross-compile on Mac M1 to Mac intel only. The recipe looks good now, only CI is the blocker.

uilianries:
Fixed, the CI was not able to export the recipe on MSVC, but exited unexpected instead of reporting the error. The PR is good for reviewing now.

uilianries (approved):
LGTM

recipes/svtjpegxs/all/patches/0.9.0-0001-manage-fpic.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index e7cdc8d..6a88a89 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -88,7 +88,6 @@ else()
+     set_property(CACHE SANITIZER PROPERTY STRINGS Address Memory Thread Undefined Integer)
+ endif()
+ 
+-set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
AbrilRBS, line 9: This is set twice in the CMakeLists, maybe they really do need this to properly compile as shared?

recipes/svtjpegxs/all/conanfile.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_APPS"] = False
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
```
AbrilRBS, line 63: In the logs, I see cpuinfo being vendorized. Maybe we can try to use CCI's package? If it turns out to be too cumbersome we can look into whether keeping the vendorized does make sense though

AbrilRBS (commented):
Thanks a lot! Some minor comments, otherwise the recipe looks good

recipes/svtjpegxs/all/conanfile.py
uilianries: Nice idea, I'll build locally and post the result.

uilianries: Just checked the upstream code is bit more complicated as it uses a custom build for cpuinfo and it built as module. I would keep the current vendorized version.

uilianries:
The fPIC needs to be enforced, because the project use OBJECT targets and those are not managed with fPIC in case building a shared library, by CMake.

The vendorized cpuinfo is more complicated to be unvedorized, I would keep as is right now.

AbrilRBS (approved):

PR #22195, opened 2024-01-07
ffmpeg: add v6.1.1, drop old versions

valgur:
Also cleaned up package_info() a bit.

PR #24522, opened 2024-07-04
ffmpeg: allow msvc shared with conan v1 & ffmeg < 6.1.1

SpaceIm:
### Summary
Changes to recipe:  **ffmpeg/all**

#### Motivation

Allow ffmpeg shared again with msvc & conan v1 when version < 6.1.1 (regression from https://github.com/conan-io/conan-center-index/pull/22195)

#### Details

Build of ffmpeg shared with msvc & conan v1 and version < 6.1.1 was working fine, and actually in https://github.com/conan-io/conan-center-index/pull/22195 there was no issue in CI, just for 6.1.1, so no reason to prevent shared build for all versions.

Moreover short_paths was an attempt for fix this in this PR, but there was no improvement when you look at CI logs, so no reason to keep it.

---


AbrilRBS (dismissed):
Thanks! We should have realized that old versions were still ok to build as shared, thanks for taking the time to revert it for those versions :)

AbrilRBS:
The build issues seem to be related to what @uilianries and I are looking into for this recipe, will update you once that's addressed :)

Drllap:
@SpaceIm @AbrilRBS @uilianries What is the status here? Do I understand it correctly that the shared lib version on msvc only fails on v6.1.1 and v6.1. If I create a MR that disables shared builds for only those two could that me merged?

uilianries:
I rebased the branch over master. Locally, I can build using ffmpeg:shared=True for 6.1 (Conan 1.x)

recipes/ffmpeg/all/conanfile.py
```diff
@@ -345,7 +343,7 @@ def validate(self):
                 raise ConanInvalidConfiguration("FFmpeg '{}' option requires '{}' option to be enabled".format(
                     dependency, "' or '".join(features)))
 
-        if conan_version.major == 1 and is_msvc(self) and self.options.shared:
+        if Version(self.version) >= "6.1.1" and conan_version.major == 1 and is_msvc(self) and self.options.shared:
```
uilianries, line 346:
```suggestion
        if Version(self.version) >= "6.1" and conan_version.major == 1 and is_msvc(self) and self.options.shared:
```

uilianries (approved):
LGTM.

PR #24532, opened 2024-07-05
qhull: add `libqhullcpp` component, add v8.1-alpha4

### LINKED ISSUE #13955: [request] qhull/8.0.2
### Package Details
  * Package Name/Version: **qhull/8.0.2**
  * Changelog: **https://github.com/qhull/qhull/releases/tag/v8.0.2**

The above mentioned version is newly released by the upstream project and not yet available as a recipe. Please add this version.
CMakeLists.txt has been changed a lot, so patches need to be updated.
Also I've noticed that qhullcpp isn't present in recipe.

I've started bumping version but it's not clear to me what exactly should be patched in CMakeLists.txt for different targets.
https://github.com/conan-io/conan-center-index/compare/master...Nekto89:conan-center-index:qhull/8.0.2

### LINKED ISSUE #16321: [question] Why is libqhullcpp excluded from all qhull packages?
### What is your question?

Hi!

I'm trying to switch a large C++ codebase over to using conan.
One dependency is libqhullcpp, which isn't included in the conan qhull recipe.

The qhull 8.0.1 recipe builds libqhullcpp, but doesn't package it.
Patch [0003-fix-cmake-8.0.x.patch](https://github.com/conan-io/conan-center-index/blob/master/recipes/qhull/all/patches/0003-fix-cmake-8.0.x.patch#L39) explicitly excludes the qhull_CPP libraries and headers from cmake.install().
The recipe doesn't seem to expose an option to include/exclude the C++ files.

What would be the reasoning to exclude the C++ libaries?

How would I install them anyway using conan? Is there a "local patch" option?

Best regards 

---

**Expected behavior**: libqhullcpp is included in qhull/8.0.1

**Actual behavior**: libqhullcpp is explicitly excluded in the abovementioned patch

**OS (tried)**: MacOS 13.1, latest Windows 10, CentOS 7.2

**Conan version**: 1.59.0

**Compiler**: apple-clang 14.0.0_clang-1400.0.29.202, msvc, gcc 8, gcc 11

valgur:
### Summary
Changes to recipe:  **qhull/cci.20231130**

#### Motivation
The `qhull` recipe currently does not export the `libqhullcpp` component, which is required by:
- cilantro: #24528
- Open3D: https://github.com/isl-org/Open3D

I also added the latest head commit version to the recipe, since the development version includes a few useful changes, such as https://github.com/qhull/qhull/pull/76 (thanks, @SpaceIm!).

Dropped v8.0.0, which is redundant to v8.0.1 and not used in any recipes.

#### Details
The cpp component is only available with the newly added version since I did not want to modify the existing patches.

Like the `reentrant` option, the `cpp` component is always built, but only exposed in `package_info()` when the option is enabled.

I had to re-enable the building and installing of the non-reentrant shared C library for backwards compatibility in the latest version.

- Resolves #16321
- Resolves #13955 (presumably)

---


valgur:
`cilantro` builds fine locally with an unvendored `libqhullcpp` library after these changes: https://github.com/valgur/conan-center-index/commit/b73a6bc18f10aeeb980fb8c962b5bb857a46f13a

recipes/qhull/all/conandata.yml
```diff
@@ -1,27 +1,27 @@
 sources:
+  "cci.20231130":
```
AbrilRBS, line 2: We're changing the `cci.YYYYMMDD` scheme to ensure it works in version ranges, I'll submit the suggestion once we get final consensus

SpaceIm:
Would it be possible to ask upstream a release in order to avoid a custom cci version?

AbrilRBS:
> Would it be possible to ask upstream a release in order to avoid a custom cci version?

This would be ideal @SpaceIm @valgur :)

perseoGI:
> Would it be possible to ask upstream a release in order to avoid a custom cci version?

@valgur @SpaceIm did you have some time to request for a semantic release schema at upstream? https://github.com/qhull/qhull/issues?q=sort%3Aupdated-desc+is%3Aissue+is%3Aopen

shkwon98:
@jcar87 @uilianries @AbrilRBS 
I'm really looking forward to it being merged...!

perseoGI:
I've just launched the new CI for this PR. Once CI is green, I'll approve! 🐸

recipes/qhull/all/conanfile.py
```diff
@@ -42,14 +42,10 @@ def config_options(self):
...
 
-    @property
-    def _is_v8_1(self):
-        return self.version == "8.1-alpha4" or Version(self.version) >= "8.1"
-
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        if self._is_v8_1:
+        if Version(self.version) >= "8.1-alpha4":
```
valgur, line 48:
```suggestion
        if Version(self.version, True) >= "8.1":
```
@perseoGI to make it work with both the pre-release and a final `8.1` release.

shkwon98:
Is there any progress?

recipes/qhull/all/conandata.yml
```diff
@@ -1,27 +1,27 @@
 sources:
+  "8.1-alpha4":
+    url: "https://github.com/qhull/qhull/archive/refs/tags/v8.1-alpha4.tar.gz"
```
uilianries, line 3:
```suggestion
  "8.1.alpha4":
    url: "https://github.com/qhull/qhull/archive/refs/tags/v8.1.alpha4.tar.gz"
```

https://github.com/qhull/qhull/tags

There is a duplication over alpha4 tag. The latest uses a dot, not a hyphen.

Usually alpha/beta are not attractive to be added in CCI, but it's clear the alpha state is pretty long, 4 years since last stable release, so it's fine.

recipes/qhull/all/conanfile.py
```diff
@@ -1,60 +1,75 @@
...
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "cpp": [True, False],
         "reentrant": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "cpp": False,
```
uilianries, line 33: The upstream does not really support an option for that, instead, users can decide what they will consume based on cmake targets. I would suggest removing `cpp` option, it only increases recipe complexity and is not clear about the benefit, as there is no related issue.

uilianries: it does need that logic, again, only increases complexity. It could be decided in `package_info` if is installed or not. Besides that, it's an alpha version (long period, but still alpha) and may change.

valgur: Anyway, if you meant dropping the cpp component entirely, then I'm definitely against that. There's even an unrelated issue requesting support for it.

uilianries: > I would actually enable the cpp component by default

Agreed.

shkwon98:
Aren't all the problems that have come up so far solved? I think if the CI job just runs, it seems like it will be merged directly..!
There are PRs that can be reviewed only after this PR is merged. I hope it will be reviewed as soon as possible.

sunbubble:
Also looking forward to this being merged :)

PR #24545, opened 2024-07-07
Simplify the `ffmpeg` test_package

ErniGH:
Simplify the `ffmpeg` test_package by removing all dependencies


valgur:
It's not doing anything particularly complicated and that info serves as a useful sanity check, imo.

SpaceIm:
I agree with @valgur, and I don't even understand the rational. This test package doesn't have any external dependency, it checks include & link of several components of ffmpeg in a very minimalistic way, which is a good thing.

uilianries (approved):

ErniGH:
It seems that the current test works well enough. I close this PR

PR #24569, opened 2024-07-09
mold: Add README

perseoGI:
### Summary
Changes to recipe:  **mold**

#### Motivation
Add mold README

---

### DIFF
--- a/recipes/mold/README.md
+++ b/recipes/mold/README.md
@@ -15,7 +15,6 @@ Add the following section to your _host_ profile targeting Linux:
 [tool_requires]
 *:mold/[*]
 [conf]
-# The following config will only work with clang or gcc >= 12
 tools.build:exelinkflags=['-fuse-ld=mold']
 tools.build:sharedlinkflags=['-fuse-ld=mold']
 ```
--- a/recipes/mold/README.md
+++ b/recipes/mold/README.md
@@ -5,9 +5,9 @@ times quicker than the LLVM lld linker, the second-fastest open-source linker.
 mold aims to enhance developer productivity by minimizing build time, 
 particularly in rapid debug-edit-rebuild cycles.
 
-You can configure [Conan](https://github.com/conan-io) to download the latest
-version of `mold` and use it as the linker when building your dependencies and
-projects from source.
+You can configure Conan to download the latest version of `mold` and use it as the linker 
+when building your dependencies and projects from source. Currently only supported
+when targeting Linux as the platform.
 
 Add the following section to your _host_ profile targeting Linux:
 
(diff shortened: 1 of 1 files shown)


recipes/mold/README.md
```diff
@@ -0,0 +1,22 @@
+# mold: A Modern Linker
+
+mold is a faster drop-in replacement for existing Unix linkers. It is several
+times quicker than the LLVM lld linker, the second-fastest open-source linker,
+which I initially developed a few years ago. mold aims to enhance developer
+productivity by minimizing build time, particularly in rapid debug-edit-rebuild
+cycles.
```
jcar87, line 7:
```suggestion
mold is a faster drop-in replacement for existing Unix linkers. It is several
times quicker than the LLVM lld linker, the second-fastest open-source linker.
mold aims to enhance developer productivity by minimizing build time, 
particularly in rapid debug-edit-rebuild cycles.
```

recipes/mold/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+You can configure [Conan](https://github.com/conan-io) to download the latest
+version of `mold` and use it as the linker when building your dependencies and
+projects from source.
```
jcar87, line 11:
```suggestion
You can configure Conan to download the latest version of `mold` and use it as the linker 
when building your dependencies and projects from source. Currently only supported
when targeting Linux as the platform.
```

recipes/mold/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+Add the following section to your _host_ profile targeting Linux:
```
jcar87, line 13:
```suggestion
To use mold automatically as the linker, you can add the following section to your
_host_ profile that targets Linux. When using gcc, please note that the following
flags require gcc 12.1 or greater.
```

recipes/mold/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+```
+[tool_requires]
+*:mold/[*]
+[conf]
+# The following config will only work with clang or gcc >= 12
```
jcar87, line 19:
```suggestion
```

jcar87:
Side note: we need to double check, but technically we _could_ make it work with gcc "older" than 12.1 - we would need the recipe's package_info to either pass `-B <package_folder>/libexec/bin` as a compiler flag - or consider the `GCC_EXEC_PREFIX` environment variable. It would work, assuming there's an executable called `ld` inside the package (rather than mold), but Im not 100% convinced it's the best course of action (it supplants ld)

davidsanfal (approved):

jcar87 (approved):

PR #24566, opened 2024-07-09
[ffmpeg] Fix Android build by disabling dependencies

uilianries:
### Summary
Changes to recipe:  **ffmpeg/4.4.4**

#### Motivation

Several issues involving ffmpeg and Android build: #23724 #23729 #23721 #18474 #23728 #23718 #20511 

The `ffmpeg` works fine when building to Android, the problem are some dependencies that are not well supported on Android and are enabled by default.

This PR remove those projects that are not supported on Android or have known bug in the upstream. I removed `libfdk_aac` and `libx265` as dependencies, both does not work on Android.

The build log is available here: https://gist.github.com/uilianries/dc9906dfb1b28f3bbfb25aefecb7dbaf#file-android-clang17-build-log

The openh264 has a separated PR to fix Android support: https://github.com/conan-io/conan-center-index/pull/24564

#### Details

The `libfdk_aac` fails when build using Android NDK: https://gist.github.com/uilianries/dc9906dfb1b28f3bbfb25aefecb7dbaf#file-libfdk_aac-android-build-log

It's a known bug: 

- https://github.com/mstorsjo/fdk-aac/issues/150
- https://github.com/mstorsjo/fdk-aac/issues/124

The `libx265` also fails when building for Android: https://gist.github.com/uilianries/dc9906dfb1b28f3bbfb25aefecb7dbaf#file-libx265-android-build-log

The upstream has related issues:

- https://bitbucket.org/multicoreware/x265_git/issues/603/aarch64-assembly-code-failing-when-cross
- Found similar errors in other package managers. 

fixes #23729
fixes #18474
fixes #23728
fixes #20511

As ffmpeg has several dependencies that are affecting its Android build, these PRs should be merged first:

- #24564
- #24627
- #24587
- #24583

---
- [x] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
- [x] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)
- [x] Tested locally with at least one configuration using a recent version of Conan

Issue #20511, opened 2023-10-11
[package] ffmpeg/6.0: Error when build for Android


paulocoutinhox:
### Description

When i try build for Android, i get this error:

```
[ 66%] Building CXX object CMakeFiles/fdk-aac.dir/libSBRdec/src/lpp_tran.cpp.o
/Users/paulo/.conan/data/libfdk_aac/2.0.2/_/_/build/36795b10a36cc186da0ea4af22f48240bd009d62/src/libSBRdec/src/lpp_tran.cpp:122:10: fatal error: 'log/log.h' file not found
#include "log/log.h"
         ^~~~~~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/fdk-aac.dir/libSBRdec/src/lpp_tran.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/fdk-aac.dir/all] Error 2
make: *** [all] Error 2
libfdk_aac/2.0.2: 
libfdk_aac/2.0.2: ERROR: Package '36795b10a36cc186da0ea4af22f48240bd009d62' build failed
libfdk_aac/2.0.2: WARN: Build folder /Users/paulo/.conan/data/libfdk_aac/2.0.2/_/_/build/36795b10a36cc186da0ea4af22f48240bd009d62/build/RelWithDebInfo
ERROR: libfdk_aac/2.0.2: Error in build() method, line 86
	cmake.build()
	ConanException: Error 2 while executing cmake --build "/Users/paulo/.conan/data/libfdk_aac/2.0.2/_/_/build/36795b10a36cc186da0ea4af22f48240bd009d62/build/RelWithDebInfo" '--' '-j10'
COMMAND: conan install /Users/paulo/Developer/workspaces/cpp/ai-kit -pr:b default -pr:h /Users/paulo/Developer/workspaces/cpp/ai-kit/conan/profiles/xplpc_android_profile -s:h build_type=RelWithDebInfo -s:h arch=armv7 -s:h os.api_level=19 -o xplpc_enable_serializer_for_json=True --build=missing --update
WORKING DIR: /Users/paulo/Developer/workspaces/cpp/ai-kit/build/conan/kotlin-android/armeabi-v7a
[ERROR] Command execution has failed
```

### Package and Environment Details

* Package Name/Version: **ffmpeg/6.0**
* Operating System+version: **macos ventura**
* Compiler+version: **clang 15**
[...]

PR #24583, opened 2024-07-11
libfdk_acc: Do not support building for Android when using NDK

uilianries:
### Summary
Changes to recipe:  **libfdk_acc/2.0.4**

#### Motivation

The issue #20511 points an error when building `ffmpeg` for Android:

```
[ 66%] Building CXX object CMakeFiles/fdk-aac.dir/libSBRdec/src/lpp_tran.cpp.o
/Users/paulo/.conan/data/libfdk_aac/2.0.2/_/_/build/36795b10a36cc186da0ea4af22f48240bd009d62/src/libSBRdec/src/lpp_tran.cpp:122:10: fatal error: 'log/log.h' file not found
#include "log/log.h"
         ^~~~~~~~~~~
1 error generated.
```

That error comes from `fdk-aac`. It's missing a header named as `log.h`, the same error is  reported in the upstream already:

- https://github.com/mstorsjo/fdk-aac/issues/124
- https://github.com/mstorsjo/fdk-aac/issues/150
- https://github.com/mstorsjo/fdk-aac/issues/144

The author is clear about not supporting NDK here: https://github.com/mstorsjo/fdk-aac/issues/124

#### Details

It's possible to fix the build error by adding a log file, as suggested here: https://github.com/mstorsjo/fdk-aac/issues/150#issuecomment-1560944481

Does it build? Yes, I tested locally.

Does it work? No idea, plus, the author did not recommend an official way to support, so I preferred to just drop that support.

Related to the PR #24566 

---

### DIFF
--- a/recipes/libfdk_aac/all/conanfile.py
+++ b/recipes/libfdk_aac/all/conanfile.py
@@ -48,7 +48,7 @@ class LibFDKAACConan(ConanFile):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
-    def validate(self):
+    def validate_build(self):
         if cross_building(self) and self.settings.os == "Android":
             # https://github.com/mstorsjo/fdk-aac/issues/124#issuecomment-653473956
             # INFO: It's possible to inject a log.h to fix the error, but there is no official support.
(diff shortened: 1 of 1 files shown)


recipes/libfdk_aac/all/conanfile.py
```diff
@@ -46,6 +48,12 @@ def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
+    def validate(self):
```
AbrilRBS, line 51: Should this not be a validate_build?

uilianries, line 51:
```suggestion
    def validate_build(self):
```

Following @AbrilRBS 's suggestion. Indeed anyone could consume a package produced by Android. The restriction is when cross-building, using Android NDK only.

uilianries: Done, applied.

AbrilRBS (approved):

AbrilRBS (approved):

jcar87 (approved):

PR #24585, opened 2024-07-11
raylib: add some common build settings to options

Julianiolo:
### Summary
Changes to recipe:  **raylib/***

#### Motivation
raylib defines a lot of build settings in `config.h`. I added some of the most common/important ones to the recipe options

#### Details
We just patch the `config.h` file to match the options. This seems to be the easiest way to do this, and it makes the `config.h` file mirror the current configuration (although that file is currently not exported; I guess this is relevant to that: #24472).

---


recipes/raylib/all/conanfile.py
```diff
@@ -54,6 +67,10 @@ def requirements(self):
         if self.settings.os == "Linux":
             self.requires("xorg/system")
 
+    def validate(self):
+        if self.options.module_rtext and not self.options.module_rtextures:
+            raise ConanInvalidConfiguration("Cannot build rtext without rtextures")
```
AbrilRBS, line 72:
```suggestion
            raise ConanInvalidConfiguration('-o="raylib/*:module_rtext=True" needs -o="raylib/*:module_rtextures=True"')
```

AbrilRBS (approved):
Thanks @Julianiolo for taking the time to add the options :)

uilianries:
As this configuration is not exposed as project option, I would suggest using Conan configuration to pass any custom definition to the compiler instead:

You can add it to your profile, or use via command line directly:

```
conan create all --version=5.0 -c 'tools.build:defines=["SUPPORT_MODULE_RTEXTURES", "SUPPORT_MODULE_RAUDIO"]'
```

This configuration will pass `-DSUPPORT_MODULE_RTEXTURES -DSUPPORT_MODULE_RAUDIO` directly to the compiler. The `-D` is automatically added by CMake, Conan passes those definition to the build tool actually.

As it may affect the package ID, in case changing the result of the library behavior, you still can inject that configuration as part of the package ID as well:

```
conan create all --version=5.0 -c 'tools.build:defines=["SUPPORT_MODULE_RTEXTURES", "SUPPORT_MODULE_RAUDIO"]' -c 'tools.info.package_id:confs=["tools.build:defines"]'
```

So, in summary `tools.info.package_id:confs` will change the package ID according to defines listed in the `tools.build:defines`, in case you change that values, it will result in a different package ID as well.

Julianiolo:
Oh WOW, how did I not see that lol, I was looking at the wrong CMakeLists.txt....

Julianiolo:
I just saw [this](https://github.com/raysan5/raylib/blob/ae50bfa2cc569c0f8d5bc4315d39db64005b1b08/CMakeOptions.txt#L38), I guess this might just be the solution to #24472? Unless there are more relevant files.

uilianries:
@Julianiolo That could work, but need to test first. Thank for spotting it.

recipes/raylib/all/conanfile.py
```diff
@@ -1,6 +1,7 @@
 from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
-from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir, save
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir, save, replace_in_file
```
uilianries, line 4:
```suggestion
from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir, save
```
Not used import

recipes/raylib/all/conanfile.py
```diff
@@ -72,14 +109,40 @@ def generate(self):
             tc.variables["USE_EXTERNAL_GLFW"] = "ON"
             tc.variables["OPENGL_VERSION"] = "OFF" if not self.options.opengl_version else self.options.opengl_version
         tc.variables["WITH_PIC"] = self.options.get_safe("fPIC", True)
+
+        tc.variables["CUSTOMIZE_BUILD"] = True
+        true_false = lambda x: True if x else False
+        tc.variables["SUPPORT_MODULE_RSHAPES"]   = true_false(self.options.module_rshapes)
```
uilianries, line 115:
```suggestion
        tc.variables["SUPPORT_MODULE_RSHAPES"]   = self.options.module_rshapes
```

The OptionValue is converted to boolean already, you should not need another helper function.

recipes/raylib/all/conanfile.py
```diff
@@ -22,11 +23,43 @@ class RaylibConan(ConanFile):
...
+        "compression_api": [True, False],
+        "events_automation": [True, False],
+        "custom_frame_control": [True, False]
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "opengl_version": None,
+
+        "module_rshapes": True,
```
uilianries, line 48: I would suggest adding only option that you really need for now, otherwise, it will increase the recipe maintenance. We can not validate each combination, which means there is a risk breaking things depending the combination.

I also would request a build log, if possible, using the options that you are adding, I mean, using the non-default option value, to make sure it's not broken. I asking it because we found recipes with custom options that are not working and the CI really does not check it (only shared, fPIC and header_only are checked), so any user opens an issue months later reporting about that option is broken and recipe/upstream is bugged.

Julianiolo:
@uilianries  I just saw this: https://github.com/raysan5/raylib/commit/307c998495a769092a8587c91a3efdb526de909c

So I guess maybe the patching approach is actually more future safe?

uilianries:
> On second thought, why does it parse the config.h to then produce cmake options, when it could just include the config.h in the code??

@Julianiolo because we avoid patching anything as we will become the maintainers of that patch.

The project offers transparent options via CMake, and it reflects to the Conan generator CMaketoolchain, that's is well prepared and integrated to work with the project, and we have been using it for any other project. So, using CMake will consume less maintenance from our side.

Julianiolo:
@uilianries No, with that commit, it does not support options via cmake anymore (or am I reading this wrong?)

Julianiolo:
Ok, turns out I was reading that completely wrong, lol
It generates the cmake options from `config.h`, everything should work as before

Julianiolo:
So now I removed some options, and added a `on_off` converter, as I had some problems just passing the booleans to cmake.

@uilianries  Do you mean by build logs just the output from `conan create all/conanfile.py ....` ? I can certainly do that :)

Julianiolo:
@uilianries so what is the status?

uilianries:
@Julianiolo like I commented before (https://github.com/conan-io/conan-center-index/pull/24585#discussion_r1678811483) I would recommend only adding those options that you really need.

Plus, add a full build log with those options values (e.g module_rshapes is True, then we need a build with False). As you can see, as more option, worst is tracking and maintaining. Regards.

Julianiolo:
build log:
`conan create all\conanfile.py --version=5.0 -o raylib/*:events_waiting=True -o raylib/*:custom_frame_control=True > build.txt 2>&1`
[build2.txt](https://github.com/user-attachments/files/17043478/build2.txt)

recipes/raylib/all/conanfile.py
```diff
@@ -22,11 +23,21 @@ class RaylibConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "opengl_version": [None, "4.3", "3.3", "2.1", "1.1", "ES-2.0"],
+
+        "module_raudio": [True, False],
+
+        "events_waiting": [True, False],
+        "custom_frame_control": [True, False]
```
uilianries, line 30:
```suggestion
        "customize_build": [True, False],
        "module_raudio": [True, False],
        "events_waiting": [True, False],
        "custom_frame_control": [True, False]
```

Add customize_build as option, as it may affect other options too.

recipes/raylib/all/conanfile.py
```diff
@@ -22,11 +23,21 @@ class RaylibConan(ConanFile):
...
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "opengl_version": None,
+
+        "module_raudio": True,
+
+        "events_waiting": False,
+        "custom_frame_control": False
```
uilianries, line 40:
```suggestion
        "customize_build": False,
        "module_raudio": True,
        "events_waiting": False,
        "custom_frame_control": False
```

Follow the upstream: https://github.com/raysan5/raylib/blob/5.0/CMakeOptions.txt#L11

recipes/raylib/all/conanfile.py
```diff
@@ -72,14 +83,29 @@ def generate(self):
...
+        tc.variables["SUPPORT_MODULE_RAUDIO"]    = on_off(self.options.module_raudio)
+
+        # this makes it include the headers rcamera.h, rgesture.h and rprand.h
+        tc.variables["SUPPORT_CAMERA_SYSTEM"]    = "ON"
+        tc.variables["SUPPORT_GESTURES_SYSTEM"]  = "ON"
+        tc.variables["SUPPORT_RPRAND_GENERATOR"] = "ON"
+
+        tc.variables["SUPPORT_EVENTS_WAITING"]       = on_off(self.options.events_waiting)
+        tc.variables["SUPPORT_CUSTOM_FRAME_CONTROL"] = on_off(self.options.custom_frame_control)
+
```
uilianries, line 98:
```suggestion

        tc.variables["CUSTOMIZE_BUILD"] = self.options.customize_build
        tc.variables["SUPPORT_MODULE_RAUDIO"] = self.options.get_safe("module_raudio")
        tc.variables["SUPPORT_EVENTS_WAITING"] = self.options.get_safe("events_waiting")
        tc.variables["SUPPORT_CUSTOM_FRAME_CONTROL"] = self.options.get_safe("custom_frame_control")
```

I would ask to remove this forced customization and simplify it instead:

- The upstream goes to the opposite way by default: https://github.com/raysan5/raylib/blob/5.0/CMakeOptions.txt#L11. The `CUSTOMIZE_BUILD` disabled, same for camera, gestures and rprand
- Not even others package managers are customizing it, real example:
  - AUR: https://gitlab.archlinux.org/archlinux/packaging/packages/raylib/-/blob/main/PKGBUILD?ref_type=heads
  - Fedora: https://src.fedoraproject.org/rpms/raylib/blob/rawhide/f/raylib.spec
  - Homebrew: https://github.com/Homebrew/homebrew-core/blob/79b0412c280d8a1283a2f38f444cad436261e1c9/Formula/r/raylib.rb

recipes/raylib/all/conanfile.py
```diff
@@ -1,4 +1,5 @@
 from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
```
uilianries, line 2:
```suggestion
```

No longer used.

uilianries (commented):
@Julianiolo Thank you for updating your PR.

There are still points that need few changes:

- The raylib project does not use customize build by default, neither other package managers that I visited are using it by default, so I don't recommend doing it by default as will affect every user.
- In case needing to extra cmake definitions, you still can use Conan config, e.g:

```
conan install ... -c -c tools.cmake.cmaketoolchain:extra_variables='{"SUPPORT_CAMERA_SYSTEM": "ON", "SUPPORT_GESTURES_SYSTEM": "ON"}'
```
The same can be part of your profile. You can obtain more information in the [CMakeToolchain](https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#conf) and [Config](https://docs.conan.io/2/reference/commands/config.html#conan-config-list) docs page.

Still, it will not affect you package ID directly, so you can enforce it by using the `tools.info.package_id:confs`:

    -c tools.info.package_id:confs='["tools.cmake.cmaketoolchain:extra_variables"]'
    
This will tell Conan to use extra_variables as part of the Package ID.

- The options `module_raudio`, `events_waiting`, and `custom_frame_control` are directly affected by `customize_build`, so I would suggest removing them in configure, as you can not change them in case customize_build is False:

```
def configure(self):
    if not self.customize_build:
        del module_raudio
        del events_waiting
        del custom_frame_control
```

- You do not need to parse to "ON"/"OFF" those options values, the boolean values work perfectly with CMake, we have being using it since many years and is compatible with both CMake 2.8 and +3.x
```

**Comment by uilianries** [comment]
@Julianiolo Thank you for pointing that detail! I'll discuss with Abril about it!

> Btw, does turning on customize build by itself even do anything?

Yes, the upstream uses `cmake_dependent_option` to enable a bunch of options in case customize_build is active. For instance: https://github.com/raysan5/raylib/blob/5.0/CMakeOptions.txt#L31

**Comment by Julianiolo** [comment]
@uilianries my question is if enabling customize build actually changes anything in the result. Shouldn't it just produce the exact same thing if no other options are changed?

**Comment by uilianries** [comment]
@Julianiolo No, `cmake_dependent_option` only enables the target option, in case the second condition is true as well. Example:

https://github.com/raysan5/raylib/blob/5.0/CMakeOptions.txt#L38

`cmake_dependent_option(SUPPORT_CAMERA_SYSTEM "Provide camera module (rcamera.h) with multiple predefined cameras: free, 1st/3rd person, orbital" ON CUSTOMIZE_BUILD ON)`

The option `SUPPORT_CAMERA_SYSTEM` is only available and configured as `ON` by default, in case `CUSTOMIZE_BUILD` is `ON`. In case `CUSTOMIZE_BUILD` is `OFF`, then, the `SUPPORT_CAMERA_SYSTEM` is `OFF` and hidden for usage. 

More information: https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

**Comment by Julianiolo** [comment]
@uilianries But shouldn't it still be ON if customize build is OFF, due to <force> being ON? So in that case customize build does nothing besides making that option available to the user, right?

**Comment by uilianries** [comment]
@Julianiolo Correct, but you may consider https://github.com/raysan5/raylib/blob/5.0/cmake/CompileDefinitions.cmake#L12

Without `CUSTOMIZE_BUILD=ON`, no other option will be used and does not matter their default value. 

----

After talking with @AbrilRBS, we considered exposing those 3 options `SUPPORT_CAMERA_SYSTEM`, `SUPPORT_GESTURES_SYSTEM` and `SUPPORT_RPRAND_GENERATOR` in self.options, so users can control that. Still, they need `customize_build=True` to be available. Please, update the recipe according.

Speaking about the PR #24472, that PR is copying those extra headers, but the upstream does not copy and will not (It was asked directly to the upstream). So, do you need those headers as well? In case positive, will need to copy them according to the active option.

**Comment by Julianiolo** [comment]
@uilianries I thought that enabling `SUPPORT_CAMERA_SYSTEM` (or simmilar) would just enable you to include them from your code?

**Comment by uilianries** [comment]
@Julianiolo Sorry, I did not understand what you mean. What I can say: Raylib will not copy camera headers even SUPPORT_CAMERA_SYSTEM is enabled, you need to copy them manually as #24472

**Comment by Julianiolo** [comment]
Huh so I thought that enabling for example `SUPPORT_CAMERA_SYSTEM` would just copy the header files, enabling you to just include them in your program. It seems it is not done that way (which is kinda weird?). It seems that enabling `SUPPORT_CAMERA_SYSTEM` just makes it include the implementation of `rcamera.h` in the library (see [this](https://github.com/raysan5/raylib/blob/ae50bfa2cc569c0f8d5bc4315d39db64005b1b08/src/rcore.c#L111))

I currently don't need those headers. But maybe we should indeed then use a solution like outlined in #24472.

**Comment by Julianiolo** [comment]
So this is my current state, everything seems to work except that the copy in the test conanfile.py fails (the one that is supposed to copy the rcamera.h). See:
```
ERROR: raylib/5.0 (test package): Error in generate() method, line 20
        copy(self, pattern="rcamera.h", src=self.dependencies["raylib"].cpp_info.resdirs[0], dst=os.path.join(self.build_folder, "raylib_stuff"))
        IndexError: list index out of range
```
I checked and the rcamera.h and others are properly copied into the res folder of the packege so that should not be the issue.

I guess the `self.dependencies["raylib"].cpp_info.resdirs[0]` might be the problem, I just copied that from [here](https://github.com/conan-io/conan-center-index/pull/24472#issuecomment-2209559918).

**Comment by Julianiolo** [comment]
@uilianries Well for my current use case they aren't needed (neither being build into the lib nor being accessible to include), but as there is already a PR for it, and it is probably a somewhat common need, we should probably at least make the headers accessible. The building into the lib is kinda not nescessary i guess, as you can just do the implementation in your own c/cpp file.

**Comment by uilianries** [comment]
@Julianiolo thank you for your quick answer! I'll check your failed test package. 😄

**Comment by Julianiolo** [comment]
@uilianries I found the issue, now the remaining question is, how do you add your custom copy location to the include directories?

**Comment by uilianries** [comment]
> @uilianries I found the issue, now the remaining question is, how do you add your custom copy location to the include directories?

@Julianiolo You need to change the `includedirs` in the `self.cpp_info`: https://docs.conan.io/2/tutorial/creating_packages/define_package_information.html#define-information-for-consumers-the-package-info-method

**Comment by Julianiolo** [comment]
@uilianries does that also work for the test package? I tried it and it does not seem to work.

**Comment by uilianries** [comment]
> does that also work for the test package? I tried it and it does not seem to work.

@Julianiolo What do you mean exactly? The test package should receive the include dirs from Conan generators, like CMakeDeps. In the test pacakge, you should not need to configure dependencies include dirs, only consume them.

**Comment by Julianiolo** [comment]
@uilianries well we copy the headers in the test package ( the ` def generate(self):
        copy(self, pattern="rcamera.h", src=self.dependencies["raylib"].cpp_info.resdirs[0], dst=os.path.join(self.build_folder, "raylib_stuff"))`), but then how do we add that directory to the include directories?

**Comment by Julianiolo** [comment]
@uilianries any idea on this?

**Comment by Julianiolo** [comment]
@uilianries ah sorry then, have a nice vacation :)

**Comment by uilianries** [comment]
@Julianiolo Hello again! I reviewed this PR again and tested locally. Please, consider the PR https://github.com/Julianiolo/conan-center-index/pull/1 as my review/fixing.

File: recipes/raylib/all/conanfile.py [hunk 9]
```diff
@@ -72,14 +111,29 @@ def generate(self):
...
+
+            # this makes it include the headers rcamera.h, rgesture.h and rprand.h
+            tc.variables["SUPPORT_CAMERA_SYSTEM"]    = self.options.camera_system
+            tc.variables["SUPPORT_GESTURES_SYSTEM"]  = self.options.gestures_system
+            if self._support_rprand_generator:
+                tc.variables["SUPPORT_RPRAND_GENERATOR"] = self.options.rprand_generator
+
         # Due to a specific logic of cmakedeps_macros.cmake used by CMakeDeps to try to locate shared libs on Windows
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0054"] = "NEW"
+
```
uilianries, line 131:
```suggestion
        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0091"] = "NEW"
```

https://cmake.org/cmake/help/latest/policy/CMP0091.html
https://c3i.jfrog.io/c3i/misc-v2/logs/pr/24585/14-windows-msvc/raylib/3.5.0//824321e7368fdc0501873cfce5f6e99adc7a95b8-build.txt

Julianiolo:
@uilianries how did this faul again?

Also, why does it even suddenly need that policy?

uilianries:
@Julianiolo You have to revert the commit https://github.com/conan-io/conan-center-index/pull/24585/commits/08dc263d06c8ce4258a9846ebad195e61204f07f

This recipe has patches and they are needed! https://github.com/conan-io/conan-center-index/tree/master/recipes/raylib/all/patches

Julianiolo:
@uilianries OH, you are right...

The Recipe linter gave me an error about them being unused [see](https://github.com/conan-io/conan-center-index/runs/32331736222), I guess that is a bug...

uilianries:
> The Recipe linter gave me an error about them being unused ([see](https://github.com/conan-io/conan-center-index/runs/32331736222)), I guess that is a bug...

@Julianiolo You are right, it's a bug! Good catch. It's because the version 5.0 does not have any patch. Will open an issue reporting the case, thank you!

uilianries (approved):
LGTM

AbrilRBS (approved):
Thanks!

PR #24590, opened 2024-07-11
msys2: conditionally prepend pkg_config_path

jcar87:
### Summary
Changes to recipe:  **msys2/all**

#### Motivation
Windows users building ffmpeg are reporting issues with the `PKG_CONFIG_PATH` variable due to a spurious `::` empty path that may interfere with pkgconf.exe or msys2 path conversions. See: 
https://github.com/conan-io/conan-center-index/issues/23309

Based on user troubleshooting, the problem goes away if there's no empty entry in the path list.

#### Details
Use shell parameter expansion such that, if `PKG_CONFIG_PATH` already has a value, we append to it - otherwise, do noting. This avoids an empty `:` when `PKG_CONFIG_PATH` is unset.
see
https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

---

### DIFF
--- a/recipes/msys2/all/conanfile.py
+++ b/recipes/msys2/all/conanfile.py
@@ -156,8 +156,9 @@ class MSYS2Conan(ConanFile):
             os.utime(tmp_name, None)
 
         # Prepend the PKG_CONFIG_PATH environment variable with an eventual PKG_CONFIG_PATH environment variable
+        # Note: this is no longer needed when we exclusively support Conan 2 integrations
         replace_in_file(self, os.path.join(self._msys_dir, "etc", "profile"),
-                              'PKG_CONFIG_PATH="', 'PKG_CONFIG_PATH="$PKG_CONFIG_PATH:')
+                              'PKG_CONFIG_PATH="', 'PKG_CONFIG_PATH="${PKG_CONFIG_PATH:+${PKG_CONFIG_PATH}:}')
 
     def package(self):
         excludes = None


AbrilRBS (approved):

elvisdukaj:
I will test this locally and tell you if it is working!

irieger:
I just rerun my failing `conan install` command that requires ffmpeg which failed before (#23309) with a clean conan home and the same profile etc. that failed before but this branch as the `local-recipes-index` index and now everything build.

danimtb (approved):

jcar87:
> I just rerun my failing `conan install` command that requires ffmpeg which failed before (#23309) with a clean conan home and the same profile etc. that failed before but this branch as the `local-recipes-index` index and now everything build.

Thanks so much for testing this!

elvisdukaj:
I confirm it works for me as well! I was close to the solution but missing the big picture! Thank you for this fix! It is unblocking many packages with this issue on Windows.

What strikes me odd is that `pkgconf` should be able to correctly handle this. Indeed the unit tests are successful. I don't understand this though.

jcar87:
I have a couple ideas about that:

*  it could be related to the implicit path type conversions that msys2 performs: https://www.msys2.org/docs/filesystem-paths/, and path lists in this case are ambiguous - and maybe something on your system causes it to convert a path when it shouldn't - for example, if you have disk drive units beyond the typical `C:/` ( just an option, I have no idea!)
* it depends on how `pkgconf.exe` handles an "empty" entry - for programs that parse a "list of paths" via an environment variable, the "empty" entry is not really well defined, in some cases it actually means "the current working directory", rather than "something to skip"

PR #24591, opened 2024-07-11
glib: support python 3.12

ErniGH:
Changes in this PR:
* Backport: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4133 to support python 3.12 (which does not have distutils). Also compatible with earlier versions of python.
* Use newer version of meson (at least 1.2.3 is required to support python 3.12, otherwise it complains about missing distutils, see [here](https://mesonbuild.com/meson-python/reference/meson-compatibility.html#cmdoption-arg-1.2.3))

### DIFF
--- a/recipes/glib/all/conandata.yml
+++ b/recipes/glib/all/conandata.yml
@@ -27,13 +27,56 @@ sources:
     url: "https://download.gnome.org/sources/glib/2.75/glib-2.75.3.tar.xz"
     sha256: "7c517d0aff456c35a039bce8a8df7a08ce95a8285b09d1849f8865f633f7f871"
 patches:
+  "2.78.3":
+    - patch_file: "patches/remove-distutils-2.77.0.patch"
+      patch_type: backport
+      patch_description: remove distutils
+      patch_source: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4133
+  "2.78.1":
+    - patch_file: "patches/remove-distutils-2.77.0.patch"
+      patch_type: backport
+      patch_description: remove distutils
+      patch_source: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4133
+  "2.78.0":
+    - patch_file: "patches/remove-distutils-2.77.0.patch"
+      patch_type: backport
+      patch_description: remove distutils
+      patch_source: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4133
+  "2.77.3":
+    - patch_file: "patches/remove-distutils-2.77.0.patch"
+      patch_type: backport
+      patch_description: remove distutils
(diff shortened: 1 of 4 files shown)


recipes/glib/all/conandata.yml
```diff
@@ -27,13 +27,56 @@ sources:
...
+    - patch_file: "patches/remove-distutils-2.77.0.patch"
+      patch_type: bugfix
+      patch_description: remove distutils
+      patch_source: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4133
   "2.76.3":
     - patch_file: "patches/libintl-discovery.patch"
       patch_type: bugfix
       patch_description: fix libintl discovery
       patch_source: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3352
+    - patch_file: "patches/remove-distutils.patch"
```
AbrilRBS, line 70: Can this file also be versioned as 2.75 just like above? It wil make it less likely that it ends up dangling in the future :)

recipes/glib/all/patches/remove-distutils-2.77.0.patch
```diff
@@ -0,0 +1,51 @@
+diff --git a/gio/gdbus-2.0/codegen/utils.py b/gio/gdbus-2.0/codegen/utils.py
+index 0204610..f8d758c 100644
+--- a/gio/gdbus-2.0/codegen/utils.py
++++ b/gio/gdbus-2.0/codegen/utils.py
+@@ -19,7 +19,7 @@
+ #
+ # Author: David Zeuthen <davidz@redhat.com>
```
AbrilRBS, line 7: :verymad: The original patch is yours, they stole your credit! hahaha 😄

recipes/glib/all/conanfile.py
```diff
@@ -81,7 +81,7 @@ def requirements(self):
             self.requires("libiconv/1.17")
 
     def build_requirements(self):
-        self.tool_requires("meson/1.2.2")
+        self.tool_requires("meson/[>=1.2.3]")
```
AbrilRBS, line 84:
```suggestion
        self.tool_requires("meson/[>=1.2.3 <2]")
```
if we are doing this it should be capped to the next major version :)

AbrilRBS (commented):
Great job! Some minor comments, otherwise looks great :)

recipes/glib/all/conandata.yml
jcar87: I wouldn't worry too much - we can use tooling to find dangling patches regardless of the filename, and they are not critical

jwillikers (approved):

jwillikers:
@ErniGH Thanks for fixing this!

jcar87 (approved):

PR #24601, opened 2024-07-12
opendis6: Deprecate in favor of libsersi

uilianries:
### Summary
Changes to recipe:  **opendis6/0.1.0**

#### Motivation

The author decided to rename the project to avoid name collision.

#### Details

The CI have limitations when renaming packages, so we will first mark this one as deprecated, then, open a second PR only adding a the new recipe.

/cc @crhowell3

Related to #24487

---

### DIFF
--- a/recipes/opendis6/all/conandata.yml
+++ b/recipes/opendis6/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
   "0.1.0":
     url: "https://github.com/crhowell3/libdis6/archive/dc1cd3bffdffcabf13b237f270d92cfebf45717c.tar.gz"
-    sha256: "3079f3bafa8f90c563ea2a0796cfff9516237a0f014d137f5bcd47c6d51136ce"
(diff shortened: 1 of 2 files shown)


AbrilRBS (dismissed):
Thanks! :)

uilianries:
Needed to update the download url to preserve the package url. Otherwise, it would fail, as the project points to another name now to the tag 0.1.0.

jcar87:
> Needed to update the download url to preserve the package url. Otherwise, it would fail, as the project points to another name now to the tag 0.1.0.

have we confirmed the contents are the same?

uilianries:
> have we confirmed the contents are the same?


Yes. The package content is the same as before, the new tag points to the new library name, but repository is the "same", only renamed and continued with new changes. 

/cc @crhowell3

AbrilRBS (approved):

AbrilRBS (changes requested):
Waiting for new name from upstream

PR #24609, opened 2024-07-12
dbus: Rename 1.x.x directory to all per standard conventions

jwillikers:
### Summary
Changes to recipe:  **dbus/***

#### Motivation

The only directory under `recipes/dbus` is `1.x.x`. This should be `all` instead per standard conventions. This makes it easier to navigate to the files in this directory when the subdirectory uses a consistent name.

---


AbrilRBS (approved):

AbrilRBS (approved):
Honestly it does not bother me at all that these folders are not called all, it even gives the recipe a bit of history showing that at some point they have had different recipes - approving nontheless as it does not create new recipe revisions

toge (approved):

PR #24625, opened 2024-07-14
atomic_queue: add recipe

luizfeldmann:
### Summary
Specify library name and version: **atomic_queue/1.6.3**

#### Details
C++14 multiple-producer-multiple-consumer lock-free queues based on circular buffer and [std::atomic](https://en.cppreference.com/w/cpp/atomic/atomic). Designed with a goal to minimize the latency between one thread pushing an element into a queue and another thread popping it from the queue.

---


recipes/atomic_queue/all/test_v1_package/CMakeLists.txt
uilianries, line 1: The entire `test_v1_package` folder can be removed. It's no longer needed for new recipes in CCI.

uilianries (commented):
Saudações e obrigado pelo seu PR! 😄 

The `test_v1_package` is used to validate Conan 1.x legacy generators like `cmake_find_package`, but we are dropping that support for new recipes, so you can remove that folder from your branch.

All other things like default CMake target name and requiring C++14 as standard are looking good. We usually impose a list of required compilers x versions to validate it with Conan 1.x in case cppstd is not declared, but GCC 5 passed without errors, so I'm fine skipping it.

Thank you.

uilianries (approved):
LGTM.

AbrilRBS (approved):
Thanks! :)

PR #24632, opened 2024-07-15
icu: change recipe to work for windows arm64

datalogics-rgailiunas:
### Summary
Changes to recipe:  **icu**

#### Motivation
ICU fails to build on Windows ARM64 as Autotools doesn't seem to recognize win_bash being set to true. 

#### Details
Autotools doesn't pass the glovb variable of win_bash, because of this ICU runs the build stage without actually building anything. This PR adds explicit calls to the make function with ```win_bash = True``` to force the package to build. Windows ARM64 uses an experimental version of clang in which all unixy tools, like bash, go through x64 emulation, which seems to be part of the issue.

---

### DIFF
--- a/recipes/icu/all/conanfile.py
+++ b/recipes/icu/all/conanfile.py
@@ -8,7 +8,7 @@ from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import is_apple_os
 from conan.tools.build import cross_building, stdcpp_library
 from conan.tools.env import Environment, VirtualBuildEnv
-from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, mkdir, rename, replace_in_file, rm, rmdir, save
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, mkdir, rename, replace_in_file, rm, rmdir, save, chdir
 from conan.tools.gnu import Autotools, AutotoolsToolchain
 from conan.tools.layout import basic_layout
 from conan.tools.microsoft import check_min_vs, is_msvc, unix_path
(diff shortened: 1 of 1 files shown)


jcar87:
Hi @datalogics-rgailiunas - thanks for your PR.

Could you provide more information as to the profile settings (host and build profiles), the msys2 installation and error you are getting?

On Windows arm64 I'm able to successfully build icu as it is (before this PR), with the following command:
```
conan create . --version=74.2 -s:b msys2/*:arch=x86_64 -c tools.gnu:host_triplet=aarch64-w64-mingw32 
```

and the following profiles:
```
Profile host:
[settings]
arch=armv8
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.gnu:host_triplet=aarch64-w64-mingw32

Profile build:
[settings]
arch=armv8
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
msys2/*:arch=x86_64
```

Not an ideal situation since the current msys2 package in Conan Center does not support arm64, but my understanding is that even today's [preliminary support](https://www.msys2.org/wiki/arm64/), the tooling we require to be provided by msys2 is still x86_64 (and emulated when running on arm64)

datalogics-rgailiunas:
Hi @jcar87,

Thank you for your reply. It appears I am able to build the package with the changes to my profiles.

What I was using before was:
``` 
Profile host:
[settings]
arch=armv8
arch_build=armv8
build_type=Release
compiler=Visual Studio
compiler.cppstd=14
compiler.runtime=MT
compiler.version=17
os=Windows
os_build=Windows
[conf]
tools.microsoft.bash:subsystem=msys2
tools.microsoft.bash:path="C:\msys64\clangarm64.exe"

Profile build:
[settings]
arch=armv8
arch_build=armv8
build_type=Release
compiler=Visual Studio
compiler.cppstd=14
compiler.runtime=MT
compiler.version=17
os=Windows
os_build=Windows
msys2/*:arch=x86_64```

It seems my mistake was with my host configurations, I thought what I needed to do was set the bash because it wasn't finding the correct one initially and those host settings didn't cause issues with other packages I had been using.
Once again, thank you for your help!

jcar87:
HI @datalogics-rgailiunas, thanks for checking, glad to hear it's working well for you.

When `tools.microsoft.bash:path` is defined in the profile (as you had initially), the recipe will no longer depend on `msys2` as a tool requires, so I think it could be related to that. If there is an issue, i'd expect it to be related to specifying that conf externally outside of the msys2 recipe, I'll double check - either way I think it's not related to arm64 :D 

WIll close this PR but if there are any issues please let us know! Happy to help with windows arm64 :)

Issue #24406, opened 2024-06-20
[SDL2] <sdl_ttf, sdl_image>/<>: Version discrepancy


rdbruce:
### Description

I have issues downloading the SDL2 packages sdl, sdl_ttf, and sdl_image due to a conflict with the libpng dependency. 
![image](https://github.com/conan-io/conan-center-index/assets/29234483/5df75c5e-7557-453e-8ef2-0be2f9c0777a)
I decided to use the SDL versions in this post just to check. 
https://blog.conan.io/2023/07/20/introduction-to-game-dev-with-sdl2.html
I think there is a discrepancy between the recipes causing the issue.
In the sdl_image recipe we have `self.requires("libpng/1.6.40")` [here on 105](https://github.com/conan-io/conan-center-index/blob/master/recipes/sdl_image/all/conanfile.py) however in the sdl_ttf freetype dependency recipe we have `self.requires("libpng/[>=1.6 <2]")` [here on 61](https://github.com/conan-io/conan-center-index/blob/master/recipes/freetype/all/conanfile.py).
I think this is causing a version mismatch.

I am sorry if this post was a waste of time. I am a novice programmer.

### Package and Environment Details

* Package Name/Version: **zlib/1.2.8**
* Operating System+version: **Linux Ubuntu 18.04**
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

my conanfile.py

from conan import ConanFile
from conan.tools.cmake import cmake_layout

class ExampleRecipe(ConanFile):
    settings = "os", "compiler", "build_type", "arch"
    generators = "CMakeDeps", "CMakeToolchain"

def requirements(self):
    # self.requires("sdl/[~2.30]")
    self.requires("sdl_ttf/2.0.18")
    self.requires("sdl_image/2.0.5")

def layout(self):
    cmake_layout(self)

### Steps to reproduce

conan install .

### Logs

robert@junk:~/code/Cstuff/SDL/final_pendulum$ conan install .

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
[...]

PR #24641, opened 2024-07-16
Use version range for libpng in sdl_image recipe

czoido:
Closes: https://github.com/conan-io/conan-center-index/issues/24406

Uses the same version range for `sdl_image` for `libpng` [that the one used by freetype](https://github.com/conan-io/conan-center-index/blob/739cf667d599a4d2dcd6a7b53bcc29dc019e38c0/recipes/freetype/all/conanfile.py#L61) that is a requirement of `sdl_ttf` so that we avoid conflicts as much as possible as these libraries are frequently used together.


AbrilRBS (approved):
👍

uilianries (approved):

davidsanfal (approved):

PR #24644, opened 2024-07-16
qt: permit usage of system libjpeg/libjpeg-turbo

mrjoel:
### Summary
Changes to recipe:  **qt/6.x.x** and **qt/5.x.x**

#### Motivation
This enables usage, particularly on Linux, of using the system native libjpeg equivalent library, whether the system provides libjpeg or libjpeg-turbo.

#### Details
Adding `system` as an allowed value for the `with_libjpeg` option enables subsequent processing and enabling of the Qt JPEG support, without having to build a specific library within Conan (and presumably vendor it with the application). This approach allows building using the native distribution provided jpeg variant and having it as a system dependency.

The result is not incurring a Conan requires, while still allowing e.g., `FEATURE_system_jpeg` to be enabled and have Qt configuration report jpeg support.

---

### DIFF
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -52,7 +52,7 @@ class QtConan(ConanFile):
         "with_fontconfig": [True, False],
         "with_icu": [True, False],
         "with_harfbuzz": [True, False],
-        "with_libjpeg": ["libjpeg", "libjpeg-turbo", False],
+        "with_libjpeg": ["libjpeg", "libjpeg-turbo", "system", False],
         "with_libpng": [True, False],
         "with_sqlite3": [True, False],
         "with_mysql": [True, False],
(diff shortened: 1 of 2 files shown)


mrjoel:
For Windows, this leaves the door open for no system package being available. It does the same on Linux as well, if neither `-dev{,el}` package is installed. I considered adding a block in `config_options()` disallowing the system option on Windows, but decided not to. If there's a better way to handle this in general I'd certainly be open to it.

jcar87:
Hi @mrjoel  - as mentioned above, have you tried the suggested https://docs.conan.io/2.5/reference/config_files/profiles.html#platform-requires ?

jcar87:
Closing this due to inactivity, and the fact that this is not a pattern anywhere else in Conan Center.

we do advise to try https://docs.conan.io/2.5/reference/config_files/profiles.html#platform-requires and provide feedback if there's anything to fix on our end, thanks!

PR #24651, opened 2024-07-17
grpc: Fix recipe to prepend to LD_LIBRARY_PATH rather than clobber

nextsilicon-itay-bookstein:
In the presence of an ambient LD_LIBRARY_PATH we want to prepend rather than clobber, same as in the case of Mac. Otherwise we may end up missing library paths required by the user.

### Summary
Changes to recipe:  **grpc**

#### Motivation

Recipe modifies cmake in a way that invokes protoc with a modified LD_LIBRARY_PATH under Linux. However, rather than prepending, it assigns, causing build errors when LD_LIBRARY_PATH is necessary.

#### Details

The PR fixes the recipe to prepend to the existing environment variable.

---


jcar87:
Hi @nextsilicon-itay-bookstein , thanks for opening this PR. We are currently in maintenance, but will get to this shortly.

I would imagine that this is something specific to your case, rather than something general? that is, you have en overridden LD_LIBRARY_PATH to support the system runtime (glibc, libstdc++, libm, pthread, etc), correct? Otherwise this should not cause errors

nextsilicon-itay-bookstein:
Yes, for example easybuild/lmod foreign package manager, where loading packages sets this environment variable. Note that the Mac flow just a few lines beforehand does prepend instead of assign.

nextsilicon-itay-bookstein:
Actually, upon further inspection, the Mac flow only propagates, doesn't preprend. Odd.

recipes/grpc/all/conanfile.py
```diff
@@ -222,7 +222,7 @@ def _patch_sources(self):
             # Note: if protobuf used RPATH instead of RUNPATH this is not needed
             replace_in_file(self, cmakelists,
                             "COMMAND ${_gRPC_PROTOBUF_PROTOC_EXECUTABLE}",
-                            'COMMAND ${CMAKE_COMMAND} -E env "LD_LIBRARY_PATH=$<JOIN:${CMAKE_LIBRARY_PATH},:>" ${_gRPC_PROTOBUF_PROTOC_EXECUTABLE}')
+                            'COMMAND ${CMAKE_COMMAND} -E env "LD_LIBRARY_PATH=$<JOIN:${CMAKE_LIBRARY_PATH};$ENV{LD_LIBRARY_PATH},:>" ${_gRPC_PROTOBUF_PROTOC_EXECUTABLE}')
```
jcar87, line 225:
```suggestion
                            'COMMAND ${CMAKE_COMMAND} -E env "LD_LIBRARY_PATH=$<JOIN:${CMAKE_LIBRARY_PATH},:>:$ENV{LD_LIBRARY_PATH}" ${_gRPC_PROTOBUF_PROTOC_EXECUTABLE}')
```

jcar87 (approved):

AbrilRBS (approved):

PR #24665, opened 2024-07-19
Update conandata_yml_format.md

cuppajoeman:
When reading the documentation here I misread that it was the hash of the file rather than the url, looking back this is an obvious mistake because we care about the files matching not the URL's matching. 

I decided to make it even more obvious for anyone else who might have gotten confused in the moment, I also added in a sample command which can generate the hash.


perseoGI (approved):
LGTM! Thanks for improving the docs!
But please, @cuppajoeman you need to request access

cuppajoeman:
> LGTM! Thanks for improving the docs! But please, @cuppajoeman you need to request access

I already requested access: 

https://github.com/conan-io/conan-center-index/issues/4#issuecomment-2237858633

Just waiting until I get added : )

perseoGI (approved):
LGTM

uilianries (approved):
LGTM, I've been using this second option for years.

PR #24670, opened 2024-07-19
joltphysics: add version 5.0.0

cuppajoeman:
### Summary
Changes to recipe:  **joltphysics/[5.0.0]**
These changes allow for users to install jolt5.0.0 with conan

#### Motivation
Many new features were added in jolt 5.0.0, these pertain to soft body, vehicles, character, constraint, collision detection, and can be read more about here: https://github.com/jrouwe/JoltPhysics/releases/tag/v5.0.0

#### Details
I am still not an authorized user, so I don't think this will be able to be merged in until I am, but I thought I would start the merge request a bit earlier, because it's my first time doing this and thought that there might be things that I have to change after review, by the time it's correct I'd probably have authorization.

**WARNING**: This merge request is most likely incorrect because I don't fully understand what the previous patch was doing and so I didn't include a patch in this commit, even though running `conan . create --version=5.0.0` works just fine.

So far I have:
* Followed the conan2 guide on consuming packages and creating packages
* Gone through: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/README.md#two-creating-a-package

#### Questions

* I am the least familiar with patching, notice that in my commit I have added the following line to the `CMakeLists.txt` in `test_package`: 
```
add_definitions(-DJPH_PROFILE_ENABLED=0 -DJPH_DEBUG_RENDERER=0)
```

I added this because when it was not there and I tried to run `conan create . --version=5.0.0` I got the error: 
```
joltphysics/5.0.0 (test package): Running test()
joltphysics/5.0.0 (test package): RUN: ./test_package
Version mismatch, make sure you compile the client code with the same Jolt version and compiler definitions!
Mismatching define JPH_PROFILE_ENABLED.
Mismatching define JPH_DEBUG_RENDERER.
```

From what I understand this is because we build the `joltphysics` package in isolation from the test package, since they both use jolt, they must have matching compiler definitions or else the code is not compatible. When I inspect the previous version patches they have lines like this: 

https://github.com/conan-io/conan-center-index/blob/1bbf24347d4528ecdcd1c13dd63258fb03c96ca7/recipes/joltphysics/all/patches/3.0.1-0001-fix-cmake.patch#L104-L108

Is my method of changing the `CMakeLists.txt` incorrect? Should I be using patching for this? [Also I have read the corresponding section on [patches](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#policy-about-patching)]

* Another question I had comes from the previous patch that looks like this: 

https://github.com/conan-io/conan-center-index/blob/1bbf24347d4528ecdcd1c13dd63258fb03c96ca7/recipes/joltphysics/all/patches/3.0.1-0001-fix-cmake.patch#L87-L102

What is the purpose of this modification? Do I need create a patch that does that in my version here?

@toge I know that you created the previous patches, would you be able to help answer a few of my questions?

---

Apologies for asking so many questions, but I really want to help contribute to conan center index. I think once I complete my first merge request I will have far less questions.

---

### DIFF
--- a/recipes/joltphysics/all/test_package/CMakeLists.txt
+++ b/recipes/joltphysics/all/test_package/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.8)
 project(test_package LANGUAGES CXX)
 
+add_definitions(-DJPH_PROFILE_ENABLED=0 -DJPH_DEBUG_RENDERER=0)
+
 find_package(joltphysics REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
--- a/recipes/joltphysics/all/conandata.yml
+++ b/recipes/joltphysics/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "5.0.0":
+    url: "https://github.com/jrouwe/JoltPhysics/archive/refs/tags/v5.0.0.tar.gz"
+    sha256: "5231953d1b1d5b9cb617facf86341a11337e1cd04456949af6911b917a1646cb"
   "3.0.1":
     url: "https://github.com/jrouwe/JoltPhysics/archive/refs/tags/v3.0.1.tar.gz"
     sha256: "7ebb40bf2dddbcf0515984582aaa197ddd06e97581fd55b98cb64f91b243b8a6"
--- a/recipes/joltphysics/config.yml
+++ b/recipes/joltphysics/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "5.0.0":
+    folder: all
   "3.0.1":
     folder: all
   "2.0.1":


AbrilRBS:
Hi @cuppajoeman  thanks for taking the time to add the new version. To be able to answer your questions, we'll need to check the previous PRs, see if some info was discussed there. I'll try to take a look tomorrow, see if we can come up with some answers! :)

shiena:
@cuppajoeman 
It looks like that patch was first made over here.
https://github.com/conan-io/conan-center-index/pull/14573 (@SpaceIm )

shiena:
@cuppajoeman @AbrilRBS 
I would like to add the following changes for mac.
When building with `build_type=Release` in apple-clang, `INTERPROCEDURAL_OPTIMIZATION` is turned on and binaries with bitcode enabled are generated.
Since bitcode has been deprecated since Xcode14 and is now causing trouble, I want to disable it by default for mac.
https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes#Deprecations

```diff
diff --git a/recipes/joltphysics/all/conanfile.py b/recipes/joltphysics/all/conanfile.py
index ef2cbdfc7..bdd5f9e77 100644
--- a/recipes/joltphysics/all/conanfile.py
+++ b/recipes/joltphysics/all/conanfile.py
@@ -1,5 +1,6 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import is_apple_os
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get
@@ -29,6 +30,7 @@ class JoltPhysicsConan(ConanFile):
         "simd": ["sse", "sse41", "sse42", "avx", "avx2", "avx512"],
         "debug_renderer": [True, False],
         "profile": [True, False],
+        "lto": [True, False],
     }
     default_options = {
         "shared": False,
@@ -36,6 +38,7 @@ class JoltPhysicsConan(ConanFile):
         "simd": "sse42",
         "debug_renderer": False,
         "profile": False,
+        "lto": True,
     }
 
     @property
@@ -80,6 +83,8 @@ class JoltPhysicsConan(ConanFile):
             del self.options.fPIC
         if self.settings.arch not in ("x86", "x86_64"):
             del self.options.simd
+        if is_apple_os(self):
+            self.options.lto = False
 
     def configure(self):
         if self.options.shared:
@@ -131,6 +136,7 @@ class JoltPhysicsConan(ConanFile):
             tc.variables["USE_STATIC_MSVC_RUNTIME_LIBRARY"] = is_msvc_static_runtime(self)
         tc.variables["JPH_DEBUG_RENDERER"] = self.options.debug_renderer
         tc.variables["JPH_PROFILE_ENABLED"] = self.options.profile
+        tc.variables["INTERPROCEDURAL_OPTIMIZATION"] = self.options.lto
         if Version(self.version) >= "3.0.0":
             tc.variables["ENABLE_ALL_WARNINGS"] = False
         tc.generate()
```

SpaceIm:
Please disable lto if it's manually managed in upstream CMakeList, and don't add such option in conancenter recipe. Lto can be enabled by users through conan config.

shiena:
@SpaceIm 
Perhaps because of the upstream definition, if you build for mac with `build_type=Release` , lto is enabled. 
Since the purpose is to disable it, it is sufficient to add `tc.variables[“INTERPROCEDURAL_OPTIMIZATION”] = False`.

uilianries:
@shiena Hello and thank you commenting. Have you discussed with the upstream this situation involving lto, instead? It look pretty particular, and it's configured already to be always ON: https://github.com/jrouwe/JoltPhysics/blob/f95ad217f2c2797081cbc33b6d3463dd0ef65f84/Build/CMakeLists.txt#L29 Maybe the upstream could change the behavior when using OSX to avoid the current error.

recipes/joltphysics/all/test_package/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.8)
 project(test_package LANGUAGES CXX)
 
+add_definitions(-DJPH_PROFILE_ENABLED=0 -DJPH_DEBUG_RENDERER=0)
```
uilianries, line 4: What's the objective of these definitions, why not exporting via cpp_info?

SpaceIm:
> @cuppajoeman It looks like that patch was first made over here. #14573 (@SpaceIm )

I've made this patch for joltphysics 2.0.1 (first version of joltphysics packaged in conancenter), I don't know if the logic is still relevant for other versions.

For 2.0.1, patch was fixing this:
- C++17 was hardcoded (raw injection of `-std=c++17` flag), so it was removed and replaced by target_compile_defintions cxx_std_17, which allows to honor compiler.cppstd from conan profile if above C++17.
- hardcoded LTO was removed (conancenter recipes provide non-LTO build by default, otherwise it can lead to binary compatibility issues across minor versions of some compilers).
- Werror/WX removed (warnings as errors can lead to compilation errors for recent compilers)
- Upstream CMakeLists was only creating static lib, not shared, so `STATIC` was removed from add_library call, allowing to honor shared option from recipe.
- There was no install target in upstream CMakeLists, so it has been added
- In Upstream CMakeLists, `JPH_DEBUG_RENDERER` and `JPH_PROFILE_ENABLED` were injected based on build type. Patch replaced this behavior by an option, so that users can decide to enable these features based on recipe option.

recipes/joltphysics/all/test_package/CMakeLists.txt
uilianries, line 4:
```suggestion
```

The current recipe is exposing those definitions already: https://github.com/conan-io/conan-center-index/blob/master/recipes/joltphysics/all/conanfile.py#L162

Please, revert this change or explain what is failing.

uilianries:
Just read the issue https://github.com/jrouwe/JoltPhysics/issues/418. I would pass `tc.variables[“INTERPROCEDURAL_OPTIMIZATION”] = False` by default. In case someone want that 10% of extra performance, is still possible to override it via the configuration `tools.cmake.cmaketoolchain:extra_variables`.

shiena:
@uilianries Thank you for all your advice. I will withdraw my request and investigate the upstream topic.

recipes/joltphysics/all/test_package/CMakeLists.txt
cuppajoeman: I kept getting an error after trying to run the executable saying that the library had been built with different definitions and so I hardcoded those in to fix it, I'll remove this so it can be done properly

cuppajoeman:
Hi there, I'm going to make some comments based on @SpaceIm said about the previous patches and what does or does not need to be carried through to the new 5.0.0 and 5.1.0 versions

>     C++17 was hardcoded (raw injection of -std=c++17 flag), so it was removed and replaced by target_compile_defintions cxx_std_17, which allows to honor compiler.cppstd from conan profile if above C++17.

It looks like we won't have to apply this patch anymore as now it looks like in upstream they don't do that anymore:

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Build/CMakeLists.txt#L125

> hardcoded LTO was removed (conancenter recipes provide non-LTO build by default, otherwise it can lead to binary compatibility issues across minor versions of some compilers).

Looking at the  patches, I didn't see anything about removing the LTO stuff, but it might just be because I don't know the syntax for turning those options on too much in cmake what I did find was this: 

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Build/CMakeLists.txt#L29

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Build/CMakeLists.txt#L249C1-L267C35 

>     Werror/WX removed (warnings as errors can lead to compilation errors for recent compilers)

Do we still want to do this? I think now jolt has an option for that:

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Build/CMakeLists.txt#L186-L188

>     Upstream CMakeLists was only creating static lib, not shared, so STATIC was removed from add_library call, allowing to honor shared option from recipe.

This should no longer need to be applied due to this line: 

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Jolt/Jolt.cmake#L475
    
>     There was no install target in upstream CMakeLists, so it has been added

This no longer needs to be patched because now there is an install target:

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Build/CMakeLists.txt#L279-L300

>     In Upstream CMakeLists, JPH_DEBUG_RENDERER and JPH_PROFILE_ENABLED were injected based on build type. Patch replaced this behavior by an option, so that users can decide to enable these features based on recipe option.

Now there are options to decide whether or not those definitions are on or not, so we probably no longer need to patch it, here is the relevant lines: 

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Jolt/Jolt.cmake#L575-L586

---

Based on the above data, this probably means that we don't need to patch anything anymore? I'd love to hear @SpaceIm's thoughts on this or anyone else who has more experience with changes that need to occur on upstream packages.

--- 

Also something that I don't quite understand with conan packages is how a conanfile.py works against different versions, for example right now there are certain options specified in the current conanfile.py:

https://github.com/conan-io/conan-center-index/blob/73992eea0ab56086fd74bfe5e3998cee5b232448/recipes/joltphysics/all/conanfile.py#L13-L39

Now if I add new options to the conanfile.py for version 5.0.0 and 5.1.0 how can we make sure that those options can't be enabled for older versions?

--- 

Should I now start working on adding in the new options specified in here? 

https://github.com/jrouwe/JoltPhysics/blob/778cbb2d5ba418df99be5b7ef48431a0480240a7/Build/CMakeLists.txt#L5-L95

Thanks.

shiena:
There is a bug that some header files are not copied during installation.
This has already been fixed in upstream, but the following patch for 5.1.0 is recommended.
https://github.com/jrouwe/JoltPhysics/commit/304c71912de0944c499d30e8324f3490c86076df

```diff
diff --git a/Jolt/Jolt.cmake b/Jolt/Jolt.cmake
index 78a4149a..0bc7a80b 100644
--- a/Jolt/Jolt.cmake
+++ b/Jolt/Jolt.cmake
@@ -13,6 +13,7 @@ set(JOLT_PHYSICS_SRC_FILES
 	${JOLT_PHYSICS_ROOT}/AABBTree/AABBTreeToBuffer.h
 	${JOLT_PHYSICS_ROOT}/AABBTree/NodeCodec/NodeCodecQuadTreeHalfFloat.h
 	${JOLT_PHYSICS_ROOT}/AABBTree/TriangleCodec/TriangleCodecIndexed8BitPackSOA4Flags.h
+	${JOLT_PHYSICS_ROOT}/ConfigurationString.h
 	${JOLT_PHYSICS_ROOT}/Core/ARMNeon.h
 	${JOLT_PHYSICS_ROOT}/Core/Array.h
 	${JOLT_PHYSICS_ROOT}/Core/Atomics.h
@@ -365,6 +366,7 @@ set(JOLT_PHYSICS_SRC_FILES
 	${JOLT_PHYSICS_ROOT}/Physics/SoftBody/SoftBodyShape.h
 	${JOLT_PHYSICS_ROOT}/Physics/SoftBody/SoftBodySharedSettings.cpp
 	${JOLT_PHYSICS_ROOT}/Physics/SoftBody/SoftBodySharedSettings.h
+	${JOLT_PHYSICS_ROOT}/Physics/SoftBody/SoftBodyUpdateContext.h
 	${JOLT_PHYSICS_ROOT}/Physics/SoftBody/SoftBodyVertex.h
 	${JOLT_PHYSICS_ROOT}/Physics/StateRecorder.h
 	${JOLT_PHYSICS_ROOT}/Physics/StateRecorderImpl.cpp
```

noizex:
Hi, I'm very interested in Jolt v5.x availability in Conan - is there something I can help with to get this recipe updated?

cuppajoeman:
> Hi, I'm very interested in Jolt v5.x availability in Conan - is there something I can help with to get this recipe updated?

I'm planning on getting it done this coming week

noizex:
> Also something that I don't quite understand with conan packages is how a conanfile.py works against different versions, for example right now there are certain options specified in the current conanfile.py:

This is actually an interesting question - most of the packages in Conan repository state only a few recent versions, often just a couple of minor versions within single major version. I browsed existing recipes looking for a jump from a very basic version like 1.0 to 3.0 or higher, something that would also experience a change in make configs, addition of new flags and options - and couldn't find anything. Most of packages just have a static set of options. 

If there was such a huge change between v3 and v5 of Jolt  and it introduced a lot of new options compared to the older versions, I actually started to wonder if this does not justify a separate recipe for newer version? But couldn't find such use cases either. 

I was able to find some hints in Conan docs for options [here]
(https://docs.conan.io/2/tutorial/creating_packages/configure_options_settings.html) 

It notes some way to dynamically remove options based on some conditional checks, so I think this could be used to check against the `version` and remove options if it's older than some set threshold. To make it nicer maybe there could be a `list` of options for v5 and if version is less than that, they're removed. 

> `configure()`: use this method to configure which options or settings of the recipe are available. For example, in this case, we delete the fPIC option, because it should only be True if we are building the library as shared (in fact, some build systems will add this flag automatically when building a shared library).
> 
> `config_options():` This method is used to constrain the available options in a package before they take a value. If a value is assigned to a setting or option that is deleted inside this method, Conan will raise an error. In this case we are deleting the fPIC option in Windows because that option does not exist for that operating system. Note that this method is executed before the configure() method.

noizex:
> I added this because when it was not there and I tried to run `conan create . --version=5.0.0` I got the error:
> 
> ```
> ======== Testing the package: Executing test ========
> joltphysics/5.0.0 (test package): Running test()
> joltphysics/5.0.0 (test package): RUN: ./test_package
> Version mismatch, make sure you compile the client code with the same Jolt version and compiler definitions!
> Mismatching define JPH_PROFILE_ENABLED.
> Mismatching define JPH_DEBUG_RENDERER.
> ```
> 
> From what I understand this is because we build the `joltphysics` package in isolation from the test package, since they both use jolt, they must have matching compiler definitions or else the code is not compatible. When I inspect the previous version patches they have lines like this:

This is true, but in my opinion package built with `default` options should have exactly the same configuration & definitions as a client built with no additional defines using that package. I think I know why this is causing error (and it is erroring on my local test currently too, as I went on and started reviewing this package). 

Here is how options are defined in Conan for `JPH_PROFILE_ENABLED`:

```python
        "profiler": False, # so long as the following two options are false, this variable comes into effect,
                                  # this is added as a custom conan option so that you can have custom profiler
                                  # behavior
        "profiler_in_debug_and_release": True,
        "profiler_in_distribution": False,
  
        <...>        

        tc.variables["PROFILER_IN_DEBUG_AND_RELEASE"] = self.options.profiler_in_debug_and_release
        tc.variables["PROFILER_IN_DISTRIBUTION"] = self.options.profiler_in_distribution
   
       <...>
       # if both options are false
       if (not self.options.profiler_in_debug_and_release and not self.options.profiler_in_distribution):
            # but you want custom behavior
            if (self.options.profiler):
                self.cpp_info.defines.append("JPH_PROFILE_ENABLED")
```

This looks correct to me as these are the defaults from Jolt CMake config as well:

```cmake
# Enable the profiler in Debug and Release builds. Note that PROFILER_IN_DISTRIBUTION will override this setting.
option(PROFILER_IN_DEBUG_AND_RELEASE "Enable the profiler in Debug and Release builds" ON)

# Enable the profiler in all builds.
# Note that enabling this reduces the performance of the library.
option(PROFILER_IN_DISTRIBUTION "Enable the profiler in all builds" OFF)
```

There is also this note: **PROFILER_IN_DISTRIBUTION will override this setting** which sort of indicates that `PROFILER_IN_DISTRIBUTION` takes precedence and if it's False then profiler WILL be disabled. That's what recipe is based on as it does default `profiler_in_debug_and_release` to `True` but `profiler_in_distribution` is False, so should disable ALL profiling. 

But here is how it's handled in Jolt's CMake:
```cmake
# Enable the profiler
if (PROFILER_IN_DISTRIBUTION)
	target_compile_definitions(Jolt PUBLIC "JPH_PROFILE_ENABLED")
elseif (PROFILER_IN_DEBUG_AND_RELEASE)
	target_compile_definitions(Jolt PUBLIC "$<$<CONFIG:Debug,Release,ReleaseASAN,ReleaseUBSAN>:JPH_PROFILE_ENABLED>")
endif()
```

This condition does not actually make `PROFILER_IN_DISTRIBUTION` guard a definitive one. If `PROFILER_IN_DISTRIBUTION` is false (which it is in our case) but `PROFILER_IN_DEBUG_AND_RELEASE` is true, the `elseif` will be followed and it will enable `JPH_PROFILE_ENABLED` anyway, despite `PROFILER_IN_DISTRIBUTION` being false.

I think this is mistake in upstream on how it handles this condition, at least based on the above **Note** that tells us that the more general setting is "overriding". I think it either needs to set `PROFILER_IN_DEBUG_AND_RELEASE` to false (which will make the default test work, but it will now diverge from original Jolt's config assumptions) or make change in the upstream to be more like:

```cmake
# Enable the profiler
if (PROFILER_IN_DISTRIBUTION)
        if (PROFILER_IN_DEBUG_AND_RELEASE)
	       target_compile_definitions(Jolt PUBLIC "$<$<CONFIG:Debug,Release,ReleaseASAN,ReleaseUBSAN>:JPH_PROFILE_ENABLED>")
        else()
        	target_compile_definitions(Jolt PUBLIC "JPH_PROFILE_ENABLED")
        endif()
endif()
```
which would I think match what the intention behind this flag is (again, assuming this from the **Note** that the more general setting should override the more specific one - so `PROFILE_IN_DISTRIBUTION` being false means there is no profiling and `PROFILER_IN_DEBUG_AND_RELEASE` can't have effect.

cuppajoeman:
Hi @noizex I looked over what you wrote here and I agree with what was said. I have some notes and a question:

I did some reading of the docs and found this out: https://docs.conan.io/2/tutorial/versioning/revisions.html which can be summed up by: 

> This sections introduces how doing modifications to a given recipe or source code without explicitly creating new versions, will still internally track those changes with a mechanism called revisions.

So as I understand it, it might not be necessary to conditionally load in specific options because for each version, the most recent revision is used for that version, though that would almost nullify the reason to check for version which might explain why you're not seeing many examples of it, though there may be cases that it's used, but as I mentioned I don't have enough experience with conan to know those cases yet.

---

As you mentioned the current status of this package is that we get mistmached defines, and I don't see where the discrepancy comes from, we have:

```py
        "profiler": False, # so long as the following two options are false, this variable comes into effect,
                                  # this is added as a custom conan option so that you can have custom profiler
                                  # behavior
        "profiler_in_debug_and_release": True,
        "profiler_in_distribution": False,
  
        <...>        

        tc.variables["PROFILER_IN_DEBUG_AND_RELEASE"] = self.options.profiler_in_debug_and_release
        tc.variables["PROFILER_IN_DISTRIBUTION"] = self.options.profiler_in_distribution
```

which would equate to setting: 

```
# Enable the profiler in Debug and Release builds. Note that PROFILER_IN_DISTRIBUTION will override this setting.
option(PROFILER_IN_DEBUG_AND_RELEASE "Enable the profiler in Debug and Release builds" ON)

# Enable the profiler in all builds.
# Note that enabling this reduces the performance of the library.
option(PROFILER_IN_DISTRIBUTION "Enable the profiler in all builds" OFF)
```

So when we build the package using `conan create . --version=5.0.0` wouldn't the execution path through the mentioned if statement be the same in jolt's cmake file when compared against the execution path when the project is being build in the test? (even though, as you mentioned the override logic is not right)

cuppajoeman:
Hi @noizex I saw your slack thread where it was mentioned that you would want to directly append the the definition, but I'm still curious we have lines like this: 

```python
        tc.variables["DEBUG_RENDERER_IN_DEBUG_AND_RELEASE"] = self.options.debug_renderer_in_debug_and_release
```

and then 
```cmake
# Enable the debug renderer
if (DEBUG_RENDERER_IN_DISTRIBUTION)
	target_compile_definitions(Jolt PUBLIC "JPH_DEBUG_RENDERER")
elseif (DEBUG_RENDERER_IN_DEBUG_AND_RELEASE)
	target_compile_definitions(Jolt PUBLIC "$<$<CONFIG:Debug,Release,ReleaseASAN,ReleaseUBSAN>:JPH_DEBUG_RENDERER>")
endif()
```

From what I understand when you set tc.variables you're defining those options to be true?

noizex:
> Hi @noizex I saw your slack thread where it was mentioned that you would want to directly append the the definition, but I'm still curious we have lines like this in conanfile.py:
> 
> ```python
>         tc.variables["DEBUG_RENDERER_IN_DEBUG_AND_RELEASE"] = self.options.debug_renderer_in_debug_and_release
> ```
> 
> and then in jolt.cmake:
> 
> ```cmake
> # Enable the debug renderer
> if (DEBUG_RENDERER_IN_DISTRIBUTION)
> 	target_compile_definitions(Jolt PUBLIC "JPH_DEBUG_RENDERER")
> elseif (DEBUG_RENDERER_IN_DEBUG_AND_RELEASE)
> 	target_compile_definitions(Jolt PUBLIC "$<$<CONFIG:Debug,Release,ReleaseASAN,ReleaseUBSAN>:JPH_DEBUG_RENDERER>")
> endif()
> ```
> 
> From what I understand when you set tc.variables you're defining those options to be true which would then allow one of these branches to execute in turn creating the compiler definition?

I think that explains why I had this lingering thought that appending to defines is not something that solves every situation. I suppose this would require literally recreating this logic from library's CMakeFiles to Conan logic in `package_info`. So we probably should be checking for build_type there? Instead using options it would probably be better to just copy defines that were used for this particular run of package build, but not sure this is even possible :/

noizex:
> From what I understand when you set tc.variables you're defining those options to be true which would then allow one of these branches to execute in turn creating the compiler definition?

Yes, this passes that flag to CMake and then CMake adds the compile definition based on this. But to make it reach the consumer of the package, based on that Slack conversation, we need to add this to `package_info` which feels a bit redundant and requires us to recreate similar conditionals that are used via CMake :/

cuppajoeman:
@noizex I have an idea on how we can do this something like this: 

```py
conan_option_to_cmake_option_name = {
    "use_asserts": "USE_ASSERTS",
    "double_precision": "DOUBLE_PRECISION",
    "generate_debug_symbols": "GENERATE_DEBUG_SYMBOLS",
    "override_cxx_flags": "OVERRIDE_CXX_FLAGS",
    "cross_platform_deterministic": "CROSS_PLATFORM_DETERMINISTIC",
    "cross_compile_arm": "CROSS_COMPILE_ARM",
    "build_shared_libs": "BUILD_SHARED_LIBS",
    "interprocedural_optimization": "INTERPROCEDURAL_OPTIMIZATION",
    "floating_point_exceptions_enabled": "FLOATING_POINT_EXCEPTIONS_ENABLED",
    "cpp_exceptions_enabled": "CPP_EXCEPTIONS_ENABLED",
    "cpp_rtti_enabled": "CPP_RTTI_ENABLED",
    "object_layer_bits": "OBJECT_LAYER_BITS",

    # X86 processor features
    "use_sse4_1": "USE_SSE4_1",
    "use_sse4_2": "USE_SSE4_2",
    "use_avx": "USE_AVX",
    "use_avx2": "USE_AVX2",
    "use_avx512": "USE_AVX512",
    "use_lzcnt": "USE_LZCNT",
    "use_tzcnt": "USE_TZCNT",
    "use_f16c": "USE_F16C",
    "use_fmadd": "USE_FMADD",

    # WebAssembly SIMD
    "use_wasm_simd": "USE_WASM_SIMD",

    # Warnings and Stats
    "enable_all_warnings": "ENABLE_ALL_WARNINGS",
    "track_broadphase_stats": "TRACK_BROADPHASE_STATS",
    "track_narrowphase_stats": "TRACK_NARROWPHASE_STATS",

    # Debug Renderer
    "debug_renderer_in_debug_and_release": "DEBUG_RENDERER_IN_DEBUG_AND_RELEASE",
    "debug_renderer_in_distribution": "DEBUG_RENDERER_IN_DISTRIBUTION",

    # Profiler
    "profiler_in_debug_and_release": "PROFILER_IN_DEBUG_AND_RELEASE",
    "profiler_in_distribution": "PROFILER_IN_DISTRIBUTION",

    # Allocator and Vector
    "disable_custom_allocator": "DISABLE_CUSTOM_ALLOCATOR",
    "use_std_vector": "USE_STD_VECTOR",

    # Object Stream and Install
    "enable_object_stream": "ENABLE_OBJECT_STREAM",
    "enable_install": "ENABLE_INSTALL"
}

this pretty much would copy the settings for both the 
def custom_profiler_check(options):
    return not options.profiler_in_debug_and_release and not options.profiler_in_distribution and options.profiler

# Map for custom evaluation logic
custom_evaluation_map = {
    "JPH_PROFILE_ENABLED": custom_profiler_check
}

def generate(self):
    for conan_option, cmake_option in conan_option_to_cmake_option_name.items():
        tc.variables[cmake_option] = getattr(self.options, conan_option)

def package_info(self):
    for conan_option, cmake_option in conan_option_to_cmake_option_name.items():
        if getattr(self.options, conan_option):
            self.cpp_info.defines.append(cmake_option)
    

    for cmake_option, custom_func in custom_evaluation_map.items():
        if custom_func(self.options):
            self.cpp_info.defines.append(cmake_option)
```

This would copy the settings for both the package and the consumer, let me know your thoughts on this, if you think it's alright I can go ahead with it.

TariqYoussef:
Hello @cuppajoeman @AbrilRBS , 
Are there any updates regarding this issue ?
Regards,
Tariq

cuppajoeman:
> Hello @cuppajoeman @AbrilRBS @noizex , Are there any updates regarding this issue ? Regards, Tariq

At the moment I need some direction on how to handle the problem last specfied by @noizex when we have a clear direction on how to fix this then I think it can be finished pretty fast

TariqYoussef:
I'm willing to help as well.

cuppajoeman:
> I'm willing to help as well.

Honestly I think if you could try to get the attention of someone who is more experienced on the slack server to look over what we have that would be really useful, I've tried a few times but haven't had any luck

shiena:
[JoltPhysics 5.2.0](https://github.com/jrouwe/JoltPhysics/releases/tag/v5.2.0) released.

cuppajoeman:
Hey @franramirez688 lmk if you have any questions about this PR I can try to help out

TariqYoussef:
Hello @franramirez688 Hope you are well
Do you have any idea when will this be finished ?

uilianries:
@cuppajoeman Hello! We are going to review this PR on this Friday (22/11) and check possible solution involving the build. Thank you for your PR again!

uilianries:
First, @cuppajoeman Thank for bringing us this new major version of joltphysics in this PR 💚 

@TariqYoussef @noizex @shiena Thank for your comments in this PR, it's clear the project has several options and can result in different configurations, as those options are dependent of each other.

After discussing about this PR with Conan team, we idealized of reducing this PR, following the minimal of options possible, as the CI can not test all those combinations and may result in future issues. The current configured options for 5.x are to disable examples, tests, profiler and the renderer.

In case you really need a specific option, we could discuss in a future issue after this PR, so we would avoid new delays.

----

Still, Conan 2.x offers a safe way to inject cmake definitions without changing any line of the recipe, the configuration `tools.cmake.cmaketoolchain:extra_variables`, so you can enabled/disable features, for instance:

```
-c tools.cmake.cmaketoolchain:extra_variables='{"PROFILER_IN_DEBUG_AND_RELEASE": "ON"}'
```

This override the value for `PROFILER_IN_DEBUG_AND_RELEASE`. It can be extended to any variable present in the upstream's CMakeLists.txt.

This configuration does not affect the package ID, so it will override your local cache with a new package that does not follow the recipe in Conan Center. In order affect the package ID, by any configuration used, you can use configuration `tools.info.package_id:confs`:

```
-c tools.info.package_id:confs='["tools.cmake.cmaketoolchain:extra_variables"]'
```

Reference: https://docs.conan.io/2/reference/config_files/global_conf.html

These configuration can be added to your profile as well, so you do not need to copy/paste for every command line. That's a fast way to unblock you in case needing a new option not available in the recipe yet.

uilianries:
The CI was not able to process the current state of this PR. It's an internal bug and is listed to be fixed. Sorry for the inconvenience.

cuppajoeman:
> The CI was not able to process the current state of this PR. It's an internal bug and is listed to be fixed. Sorry for the inconvenience.

Is there any idea on the progress of the internal bug or an expected time the bug blocking this will get fixed? Is there an issue we can look at to see the progress of it?

cuppajoeman:
@uilianries Hi, I'm just asking again about any more information or a timeline we can expect the internal bug which is blocking this to be fixed

jcar87:
Hi @cuppajoeman  - the bug has been fixed for a while, apologies the job for this has not been relaunched. Please see the Checks tab for CI build results (as there are some failures)

uilianries:
The current error seems to be related to Apple Clang 13 because locally I can build using Apple Clang 16. 

As I didn't find a related issue on the upstream, so I opened an issue asking for help: https://github.com/jrouwe/JoltPhysics/issues/1482

Once we have the minimal Apple Clang version required, we can add it `validate()`, so the CI will skip Apple builds in case not fitting the required compiler version, but will build for Linux and Windows.

uilianries:
The upstream kindly provided a workaround for Apple Clang 13, and the same is merged in the master branch already. I pushed the very same patch to this new recipe version, on the commit [4c278ee](https://github.com/conan-io/conan-center-index/pull/24670/commits/4c278eef29ca50ba4c63e8cf4ae68ce8d1c830bf)

uilianries (approved):
LGTM.

- The package_info follows the generated cmake values: [JoltConfig.cmake.txt](https://github.com/user-attachments/files/18598636/JoltConfig.cmake.txt)
- License is MIT: https://github.com/jrouwe/JoltPhysics/blob/master/LICENSE
- C++ 17 is required: https://github.com/jrouwe/JoltPhysics/blob/v5.2.0/Jolt/Jolt.cmake#L2
- CMake 3.20 is mandatory: https://github.com/jrouwe/JoltPhysics/blob/v5.2.0/Build/CMakeLists.txt#L1

AbrilRBS (approved):

uilianries (approved):

cuppajoeman:
@uilianries  I don't believe that the supplied `test_package.cpp` is sufficient, as when we expand the package test to the full [`HelloWorld.cpp`](https://github.com/jrouwe/JoltPhysics/blob/92aa4c2d02e1483edcf30bd8671a02f4e9ce4776/HelloWorld/HelloWorld.cpp) then the package [fails to work properly on windows. ](https://github.com/conan-io/conan-center-index/issues/26561)

uilianries:
@cuppajoeman Thank you for reporting the case in a new issue! It should be investigated during next week!

noizex:
Yeah I also ran into this issue. The suggested fix works for now! 

What I wonder is how we can change the different options for this recipe like:
```     
        tc.cache_variables["TARGET_UNIT_TESTS"] = False
        tc.cache_variables["TARGET_HELLO_WORLD"] = False
        tc.cache_variables["TARGET_PERFORMANCE_TEST"] = False
        tc.cache_variables["TARGET_SAMPLES"] = False
        tc.cache_variables["TARGET_VIEWER"] = False
        tc.cache_variables["CROSS_PLATFORM_DETERMINISTIC"] = False
        tc.cache_variables["INTERPROCEDURAL_OPTIMIZATION"] = False
        tc.cache_variables["GENERATE_DEBUG_SYMBOLS"] = False
        tc.cache_variables["ENABLE_ALL_WARNINGS"] = False
        tc.cache_variables["OVERRIDE_CXX_FLAGS"] = False
        tc.cache_variables["DEBUG_RENDERER_IN_DEBUG_AND_RELEASE"] = False
        tc.cache_variables["PROFILER_IN_DEBUG_AND_RELEASE"] = False
```        
via conanfile.txt? It seems like these options are sort of hardcoded and some we may want to have changed for our builds? I noticed I had this
```
joltphysics/*:disable_custom_allocator=True
```
in my [options] section of my older Jolt recipe but not sure this ever worked.

uilianries:
> Yeah I also ran into this issue. The suggested fix works for now!
> 
> What I wonder is how we can change the different options for this recipe like:
> 
> ```
>         tc.cache_variables["TARGET_UNIT_TESTS"] = False
>         tc.cache_variables["TARGET_HELLO_WORLD"] = False
>         tc.cache_variables["TARGET_PERFORMANCE_TEST"] = False
>         tc.cache_variables["TARGET_SAMPLES"] = False
>         tc.cache_variables["TARGET_VIEWER"] = False
>         tc.cache_variables["CROSS_PLATFORM_DETERMINISTIC"] = False
>         tc.cache_variables["INTERPROCEDURAL_OPTIMIZATION"] = False
>         tc.cache_variables["GENERATE_DEBUG_SYMBOLS"] = False
>         tc.cache_variables["ENABLE_ALL_WARNINGS"] = False
>         tc.cache_variables["OVERRIDE_CXX_FLAGS"] = False
>         tc.cache_variables["DEBUG_RENDERER_IN_DEBUG_AND_RELEASE"] = False
>         tc.cache_variables["PROFILER_IN_DEBUG_AND_RELEASE"] = False
> ```
> 
> via conanfile.txt? It seems like these options are sort of hardcoded and some we may want to have changed for our builds? I noticed I had this
> 
> ```
> joltphysics/*:disable_custom_allocator=True
> ```
> 
> in my [options] section of my older Jolt recipe but not sure this ever worked.

Hello, @noizex! For those exposed options, you can use the Conan 2.x [config feature](https://docs.conan.io/2/reference/config_files/global_conf.html#global-conf) to inject variables directly to CMake when running Conan. For instance:

`-c 'tools.cmake.cmaketoolchain:extra_variables={"DISABLE_CUSTOM_ALLOCATOR": "OFF"}'`

You can add this configuration to your Conan command line, as represented here, or you can add it to your [Conan profile](https://docs.conan.io/2/reference/config_files/profiles.html#conf). It will not work in case it is being added to your conanfile.txt. 

Still, this configuration alone does not affect your [package ID](https://docs.conan.io/2/tutorial/consuming_packages/different_configurations.html#introducing-the-concept-of-package-id), so you will need to add it in another configuration, so Conan will consider any change in `extra_variables` as part of the package ID:

`-c 'tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]'`

The same can be added to your profile as well. Regards.

cuppajoeman:
> > Yeah I also ran into this issue. The suggested fix works for now!
> > What I wonder is how we can change the different options for this recipe like:
> > ```
> >         tc.cache_variables["TARGET_UNIT_TESTS"] = False
> >         tc.cache_variables["TARGET_HELLO_WORLD"] = False
> >         tc.cache_variables["TARGET_PERFORMANCE_TEST"] = False
> >         tc.cache_variables["TARGET_SAMPLES"] = False
> >         tc.cache_variables["TARGET_VIEWER"] = False
> >         tc.cache_variables["CROSS_PLATFORM_DETERMINISTIC"] = False
> >         tc.cache_variables["INTERPROCEDURAL_OPTIMIZATION"] = False
> >         tc.cache_variables["GENERATE_DEBUG_SYMBOLS"] = False
> >         tc.cache_variables["ENABLE_ALL_WARNINGS"] = False
> >         tc.cache_variables["OVERRIDE_CXX_FLAGS"] = False
> >         tc.cache_variables["DEBUG_RENDERER_IN_DEBUG_AND_RELEASE"] = False
> >         tc.cache_variables["PROFILER_IN_DEBUG_AND_RELEASE"] = False
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > via conanfile.txt? It seems like these options are sort of hardcoded and some we may want to have changed for our builds? I noticed I had this
> > ```
> > joltphysics/*:disable_custom_allocator=True
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in my [options] section of my older Jolt recipe but not sure this ever worked.
> 
> Hello, @noizex! For those exposed options, you can use the Conan 2.x [config feature](https://docs.conan.io/2/reference/config_files/global_conf.html#global-conf) to inject variables directly to CMake when running Conan. For instance:
> 
> `-c 'tools.cmake.cmaketoolchain:extra_variables={"DISABLE_CUSTOM_ALLOCATOR": "OFF"}'`
> 
> You can add this configuration to your Conan command line, as represented here, or you can add it to your [Conan profile](https://docs.conan.io/2/reference/config_files/profiles.html#conf). It will not work in case it is being added to your conanfile.txt.
> 
> Still, this configuration alone does not affect your [package ID](https://docs.conan.io/2/tutorial/consuming_packages/different_configurations.html#introducing-the-concept-of-package-id), so you will need to add it in another configuration, so Conan will consider any change in `extra_variables` as part of the package ID:
> 
> `-c 'tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]'`
> 
> The same can be added to your profile as well. Regards.

This might already be obvious but something I didn't realize for a while was that if you specify the extra config options, it doesn't by default cause the library to rebuild, you have to pass the `--build=jolt...` option as well to rebuild it with the new option. 

I think when you allow for extra variables go be taken into account for the package id that it will automatically trigger the rebuild, and I think once I had to delete my build folder to get it to work. Things to keep in mind for anyone else who visits this.

noizex:
By the way is this the way that Conan standards suggest for such toggles? It feels a bit counterintuitive to have this injected outside of conanfile which is the source of truth for me on what specific toggles we want to have on that library?

jcar87:
> This might already be obvious but something I didn't realize for a while was that if you specify the extra config options, it doesn't by default cause the library to rebuild, you have to pass the --build=jolt... option as well to rebuild it with the new option.

The change in options needs to cause the package_id to change - if we're injecting via a conf (e.g. `tools.cmake.cmaketoolchain:extra_variables` - then you also want to tell Conan that this conf affects the package ID (with `-c 'tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]'`). If both are in place, and the package ID didn't previously exist in your local Conan cache, then it should be built from source.

> By the way is this the way that Conan standards suggest for such toggles? It feels a bit counterintuitive to have this injected outside of conanfile which is the source of truth for me on what specific toggles we want to have on that library?

It's a recipe maintainability compromise in Conan Center, rather than an enforced standard one way or the other. Historically in Conan Center, recipes have reflected the options exposed in their build systems, so we've had to "match" them in the recipes. However, this can cause the space of possible package IDs to be very large - take a look at the OpenSSL recipe for example: https://github.com/conan-io/conan-center-index/blob/264d0cd6e12de96bf7b796e456cc96a9e9444cce/recipes/openssl/3.x.x/conanfile.py#L28-L90 - the True/False toggles alone would give way to 2^62 combinations of the package, and the reality is obvious: these combinations are *not* tested, a lot of them plainly don't work, and despite best efforts, new versions may add/remove options to the point where one may want a catch-all anyway: https://github.com/conan-io/conan-center-index/pull/27083.

This practice in Conan Center (Reflecting **all options** in the recipe), is not matched by any package manager that we know of - so we are in the process of considering whether the "comprehensive" options approach is the correct one from a practical point of view. What we want instead is to expose the options that are most likely to be useful to most users, have a way to inject build toggles without the recipe having to be modified - and wait for feedback to see if this approach reaches a good compromise between the ability to customise binaries from recipes, and our ability to maintain the recipes themselves.

noizex:
I can't really get it to work. I played with the profiles entry and it does seem to pick that variable but it tries to apply it to all the packages and I feel it forces them all to recompile when doing 'conan install'. Feels like this can't be _neatly_ applied to some specific package instead of all via profile? 

I'm still not convinced that official package won't support things like debug drawing in Debug build for instance - because it's off by default in every build. It makes it super difficult to use the library and probably just forces people to clone their own recipe, change it the way it suits them and not use the `conan-io` repo. It probably also depends on the library but unfortunately Jolt author decided to put debug drawing capabilities behind compile-time define rather than runtime check for some renderer instance being registered, and once it's compiled without it there is no way to have debug rendering on. I'm sure someone else could be missing something else, but I'm almost certain most people would expect debug rendering to work at least in debug build.

noizex:
Also this was available with Jolt 3.0 as can be seen here:
https://github.com/conan-io/conan-center-index/blob/master/recipes/joltphysics/3.x/conanfile.py#L132

so not having it in 5.0 is a bit of regression. Would it make sense to create PR that at least re-adds profiling and debug rendering options because without them this recipe is not very usable in typical case. 

Btw. I was not able to leverage the injection method suggested above, it either breaks completely or affects all packages within my conanfile.txt and tries to recompile them. Maybe I'm missing something as this feature is not well documented and seems to have different versions (some are using file with the defines, while the suggestion was to pass a string dict-like param)

PR #17594, opened 2023-05-17
mysql-connector-cpp:8.3.0

samuaz:
Specify library name and version:  **mysql-connector-cpp/8.0.33**

Hi guys, this is to include conan support for mysql-connector-cpp, conan already have the  mysql-connector-c but is missing the cpp ones.

---

- [ X] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [X ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [X ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

Issue #16312, opened 2023-02-27
[request] mysql_connector_cpp/8.0.32


ahmed192a:
### Package Name/Version

mysql_connector_cpp/8.0.32

### Webpage

https://dev.mysql.com/doc/connector-cpp/8.0/en/

### Source code

https://github.com/mysql/mysql-connector-cpp

### Description of the library/tool

This manual describes how to install and configure MySQL Connector/C++ 8.0, which provides C++ and plain C interfaces for communicating with MySQL servers, and how to use Connector/C++ to develop database applications.

Connector/C++ 8.0 is highly recommended for use with MySQL Server 8.0 and 5.7. Please upgrade to Connector/C++ 8.0.

For notes detailing the changes in each release of Connector/C++, see [MySQL Connector/C++ Release Notes](https://dev.mysql.com/doc/relnotes/connector-cpp/en/).

For legal information, see the [Legal Notices](https://dev.mysql.com/doc/connector-cpp/8.0/en/preface.html#legalnotice).

For help with using MySQL, please visit the [MySQL Forums](http://forums.mysql.com/), where you can discuss your issues with other MySQL users.

Licensing information.  This product may include third-party software, used under license. If you are using a Commercial release of MySQL Connector/C++, see the [MySQL Connector/C++ Commercial Release License Information User Manual](https://downloads.mysql.com/docs/licenses/connector-cpp-8.0-com-en.pdf) for licensing information, including licensing information relating to third-party software that may be included in this Commercial release. If you are using a Community release of MySQL Connector/C++, see the [MySQL Connector/C++ Community Release License Information User Manual](https://downloads.mysql.com/docs/licenses/connector-cpp-8.0-gpl-en.pdf) for licensing information, including licensing information relating to third-party software that may be included in this Community release.

Document generated on: 2023-01-19 (revision: 74871)

Issue #374, opened 2019-11-22
[request] mysql-connector-c/8.0


andre-cunha:
### Package Details
  * Package Name/Version: **mysql-connector-c/8.0**
  * website: https://dev.mysql.com/doc/connector-cpp/8.0/en/

The above mentioned version is newly released by the upstream project and not yet available as a recipe. Please add this version.

ericLemanissier:
this is a new package, MySQL Connector/C++ is not the same thing as MySQL Connector/C or libmysqlclient

andre-cunha:
What i am looking for is MySQL Connector/C++. So is this the right package?

ericLemanissier:
Yes, you should just update the issue title and edit your first message to make it clear.

andre-cunha:
done!

hwhsu1231:
May I ask, when will this package be available from ConanCenter?

Croydon:
Whenever someone does the work and opens a pull request. Contributions are always welcome 🙂

PR #20699, opened 2023-10-20
[mysql-connector-cpp] Add new package for mysql-connector-cpp/8.0.33

uilianries:
Specify library name and version:  **mysql-connector-cpp/8.0.33**

This PR is a continuation over the PR #17594 with the changes requested on https://github.com/samuaz/conan-center-index/pull/2. 

Some considerations:

- The MySQL Connector C++ has all dependencies listed internally, but offers using external too.
- Using external dependencies is an adaptation only, we still need to follow internal rules, like using CMake variables and custom CMake methods. It resulted in several patches.
- There is a current linkage (I guess) error when building on Windows, but I'm not inclined to spend more time on it now, pushing the current state: Working on *nix.

closes https://github.com/samuaz/conan-center-index/pull/2
closes #16312
closes #374
closes #17594

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24678, opened 2024-07-21
mysqlcpp: add recipe

husitawi:
### Summary
Add new recipe: mysqlcpp/9.0.0

#### Motivation
No C++ MySQL connector recipe available.

#### Details
No changes. New recipe with test package.

closes https://github.com/samuaz/conan-center-index/pull/2
closes https://github.com/conan-io/conan-center-index/issues/16312
closes https://github.com/conan-io/conan-center-index/issues/374
closes https://github.com/conan-io/conan-center-index/pull/17594

---


AbrilRBS:
Having issues compiling locally for apple-clang 15, will try to submit a fix

husitawi:
> Having issues compiling locally for apple-clang 15, will try to submit a fix asap

Well I was testing locally with an M1 chip, is your machine x86 ? That's what the CI is having problems with too on apple-clang13

husitawi:
Hey @AbrilRBS 👋 Any updates over this ?

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+               "fPIC": [True, False],
+               }
+
+    default_options = { "shared": False, "fPIC": True }
+
+    def validate(self):
+        check_min_cppstd(self, "17")
+
+    def requirements(self):
+        self.requires("openssl/[>=1.1 <4]", headers=True, libs=True)
```
valgur, line 40:
```suggestion
        self.requires("openssl/[>=1.1 <4]")
```
These are True by default.

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+    settings = "os", "compiler", "build_type", "arch"
+    options = {
+               "shared": [True, False],
+               "fPIC": [True, False],
+               }
+
+    default_options = { "shared": False, "fPIC": True }
+
+    def validate(self):
+        check_min_cppstd(self, "17")
```
valgur, line 37:
```suggestion
        if self.settings.compiler.cppstd:
            check_min_cppstd(self, "17")
```
`compiler.cppstd` setting is not guaranteed to be available in general.

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+    def requirements(self):
+        self.requires("openssl/[>=1.1 <4]", headers=True, libs=True)
+        self.requires("boost/1.85.0")
+        self.requires("zlib/1.3.1")
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.24 <4]")
+
+        if not is_msvc(self):
+            self.tool_requires("ninja/[>=1.10 <2]")
```
valgur, line 48: Is this a requirement or more of a nice to have?

husitawi: I noticed a significant performance improvement in Linux sub-systems on windows. Actually Make just hoards so much memory that the build crashes.

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+        replace_in_file(self, os.path.join(self.source_folder, "cdk", "extra", "zstd", "CMakeLists.txt"),
+                            "enable_pic()",
+                            "enable_pic()\n"\
+                            "add_compile_definitions(ZSTD_DISABLE_ASM)",
+                            strict=False)
+
+        # Apple patches
+        if is_apple_os(self) and cross_building(self):
+            print(f"Building for {str(self.settings.arch)}")
+            patch = f"set(CMAKE_OSX_ARCHITECTURES \"{str(self.settings.arch)}\" CACHE INTERNAL \"\" FORCE)\n"
```
valgur, line 123:
```suggestion
            patch = f"set(CMAKE_OSX_ARCHITECTURES \"{self.settings.arch}\" CACHE INTERNAL \"\" FORCE)\n"
```
Should use `self.output.info()` instead of `print()` in recipes, but that was not adding much anyway.

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+                vs_version = "unknown"
+
+            self.output.info(f"VS Version: {vs_version}")
+            self.vs_version = vs_version
+            self.vs = f"vs{vs_version}"
+        else:
+            self.vs = ""
+
+        template_dirs = ["lib64", "lib64/debug", "lib", "lib/debug"]
+        self.cpp_info.libdirs = template_dirs if not self.vs else [f"{lib}/{self.vs}" for lib in template_dirs]
```
valgur, line 173:
```suggestion
    @property
    def _vs_version(self):
        msvc_toolset_version = self.settings.get_safe("compiler.toolset")
        msvc_version = self.settings.get_safe("compiler.version")

        if msvc_toolset_version:
            vs_version = msvc_toolset_version[1:3]
        elif msvc_version:
            vs_version = msvc_version[:2]
            if vs_version == "18":
                vs_version = "12"
            elif vs_version == "19":
                vs_version = "14"
        else:
            self.output.warn("MSVC_TOOLSET_VERSION or MSVC_VERSION not defined")
            vs_version = "unknown"

        self.output.info(f"VS Version: {vs_version}")
        return f"vs{vs_version}"

    def package_info(self):
        template_dirs = ["lib64", os.path.join("lib64", "debug"), "lib", os.path.join("lib", "debug")]
        if is_msvc(self):
            vs_version = self._vs_version
            self.cpp_info.libdirs = [os.path.join(lib, vs_version) for lib in template_dirs]
        else:
            self.cpp_info.libdirs = template_dirs
```
I suggest moving `self.vs` logic into a separate function and using `os.path.join()`.

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,207 @@
...
+            lib += "-mt"
+        self.cpp_info.libs = [lib]
+
+        if not self.options.shared:
+            self.cpp_info.defines = ["MYSQL_STATIC"]
+            self.cpp_info.defines = ["STATIC_CONCPP"]
+
+        self.cpp_info.set_property("cmake_find_package", "mysql-concpp")
+        self.cpp_info.set_property("cmake_find_package_multi", "mysql-concpp")
+        self.cpp_info.set_property("cmake_config_file", "mysql-concpp-config.cmake")
```
valgur, line 207:
```suggestion
```
These won't work and should be `self.cpp_info.names["cmake_find_package"] = "..."` instead. Would need extra handling of mismatched package and namespace names. Not worth the effort. I'd simply drop any legacy generators support.

husitawi:
@valgur Hey, just wanted to point out the abnormal number of commits, sorry for that, I wanted to get apple-clang13 working but I guess I'm gonna have to let it go. It's impossible to replicate the duplicate OpenSSL definitions (static builds) on my machine (M1 Max) at link time, that's the only configuration that keeps failing in the CI.

recipes/mysql-concpp/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.25)
```
valgur, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15)
```
Why not the default of CMake 3.15?

recipes/mysql-concpp/all/test_package/src/example.cpp
valgur, line 1: ConanCenter has been trying to move away from complex test_package .c/.cpp files in recent months. You should replace it with a bare minimum one or so function calls against the library to just test that includes and linking works.

valgur (changes requested):
Some fairly minor suggestions. Looks good overall. Thanks!

recipes/mysql-concpp/all/test_package/src/example.cpp
husitawi: WIll do, I'll cout some macro then, I pulled the example from the official docs, felt like a proper test.

recipes/mysql-concpp/all/conanfile.py
valgur: Oh, wow, ok. Interesting.

valgur:
While matching the name used for `find_package()` (`mysql-concpp`) is a good idea, I would suggest using `mysql-connector-cpp` as the Conan package name instead for consistency with the existing `mysql-connector-c`. That's also the name of the GitHub repo for the project, so it's not completely arbitrary.

recipes/mysql-concpp/all/test_package/src/example.cpp
valgur, line 1: Thanks for the simplifcation! Can be slimmed down even more aggressively, though:
```cpp
#include <mysqlx/xdevapi.h>

int main()
{
    mysqlx::SessionSettings from_url("mysqlx://user:pwd@127.0.0.1:1234/db?ssl-mode=required");
    return 0;
}
```
CCI tries to avoid creating network connections and filesystem accesses in test_package, if possible.

recipes/mysql-concpp/all/test_package/conanfile.py
```diff
@@ -0,0 +1,29 @@
...
+class TestPackageConan(ConanFile):
+    settings = "os", "compiler", "build_type", "arch"
+    generators = "CMakeDeps", "CMakeToolchain"
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.15 <4]")
+
```
valgur, line 17:
```suggestion
```
Can be avoided since Conan v2 assumes and requires that CMake 3.15+ is present on the system.

recipes/mysql-concpp/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+project(PackageTest CXX)
+
+find_package(mysql-concpp REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} src/example.cpp)
+target_link_libraries(${PROJECT_NAME} PRIVATE mysql::concpp)
+
```
valgur, line 8:
```suggestion
target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
```
Might help with the Conan v1 issues.

recipes/mysql-concpp/all/conanfile.py
```diff
@@ -0,0 +1,240 @@
...
+required_conan_version = ">=1.64.1"
+
+class MysqlCppConnRecipe(ConanFile):
+    name = "mysql-concpp"
+    package_type = "library"
+    short_paths = True
+    version= "9.0.0"
+
+    # Optional metadata
+    license = "GPL-2.0"
```
valgur, line 21:
```suggestion
    license = "GPL-2.0-only"
```
Can be made more specific.

perseoGI:
Hi @husitawi!
Thank you very much for your contribution. 
First of all, have you taken a look to the existing PRs that tries to introduce this library?
- https://github.com/conan-io/conan-center-index/pull/17594 
- https://github.com/conan-io/conan-center-index/pull/20699 
- https://github.com/conan-io/conan-center-index/pull/21949

Having taken a look to these PRs yours look the most up to date, but it will be ideal to "merge the knowledge" of these PRs into a single one.

husitawi:
Hi @AbrilRBS can you please explain some of the changes being done?

perseoGI:
Hi @husitawi !
Here is the explanation:
- Remove the fixed version on the recipe as in CCI we delegate versioning to the CI which will look in the `conandata.yml` file to generate all possibilities.
- If there is not specific and clear reason to use a very old `openssl` version, prefer use version range: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md
- Use packaged dependencies instead of relaying on third party libs in upstream
- Avoid relaying on `ninja`, let consumers decide which generator to use. Also, we do not understand which criteria were chosen to use `ninja` in every os but windows
- In the `generate` method, use packaged versions instead of compiling upstream third party dependencies

Take a look at this template https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py, that may help you with some ordering and other commented stuff!

husitawi:
> Hi @husitawi ! Here is the explanation:
> 
> * Remove the fixed version on the recipe as in CCI we delegate versioning to the CI which will look in the `conandata.yml` file to generate all possibilities.

This will require more patching done on the package, it already includes weird mechanisms to find compression libraries and protobuf.

> * If there is not specific and clear reason to use a very old `openssl` version, prefer use version range: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md

Newer versions kept creating duplicate symbols for some reason, reverting back into an older version fixed it.

> * Use packaged dependencies instead of relaying on third party libs in upstream
> * Avoid relaying on `ninja`, let consumers decide which generator to use. Also, we do not understand which criteria were chosen to use `ninja` in every os but windows

Ninja was memory efficient on Linux subsystems in Winodws, otherwise the build broke and ran out of memory.

> * In the `generate` method, use packaged versions instead of compiling upstream third party dependencies

Again, needs more patches on the package itself, won't work on its own

> Take a look at this template https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py, that may help you with some ordering and other commented stuff!

I appreciate the help, but please suggest the changes instead of committing them directly, it only makes it harder to understand the edits and make them work.

valgur:
@husitawi I looked into the recipe for the project and found the two culprits for most of the issues you are encountering. See #25170.

recipes/mysql-connector-cpp/all/conanfile.py
```diff
@@ -0,0 +1,258 @@
...
+        tc.cache_variables["WITH_LZ4"] = self._package_folder_dep("lz4")
+        # ZLIB patches
+        tc.cache_variables["WITH_ZLIB"] = self._package_folder_dep("zlib")
+        # ZSTD patches
+        tc.cache_variables["WITH_ZSTD"] = self._package_folder_dep("zstd")
+        # Build patches
+        tc.cache_variables["BUILD_STATIC"] = not self.options.shared
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
+        # Disable Boost, only legacy JDBC connector needs it
+        tc.cache_variables["BOOST_DIR"] = "FALSE"
```
samuaz, line 116: why not support legacy JDBC? in this PR
https://github.com/conan-io/conan-center-index/pull/17594
i added support to it. 

i think there is still useful to support the old JDBC connector for services that are using mysql server 6x

samuaz:
I really advocate supporting the JDBC connector as I was including in my PR

https://github.com/conan-io/conan-center-index/pull/17594

There are still a lot of projects using it.

It is one of the reasons I was trying to add this mysql package to Conan also

perseoGI:
@samuaz I consider it will be better for all if we joint or work in a same PR.
Would you mind to add the `JDBC` support to this PR and try to make the CI pass?

valgur:
@perseoGI #25170

samuaz:
> @samuaz I consider it will be better for all if we joint or work in a same PR. Would you mind to add the `JDBC` support to this PR and try to make the CI pass?

guys, add the JDBC support here is basically retake my PR.

my PR was really close to be done, i make it pass the CI many times, not offense but really dont understand why redo the same things many times. my PR was there from february 23

i dont have problem to work on this.. .but is basically cherry pick my changes so....

PR #24694, opened 2024-07-23
Add new versions for llama-cpp and also adding a couple of options

czoido:
- Removing old b2038 version
- Adding options for building with curl (can use autodownload from hugginface repos), cuda (pending testing) and metal.
- Add options to build examples, I think it's worth adding as there are some that are very interesting and serve to test the options for curl, cuda, etc.
- Had to patch how curl is added but will try to open a PR to upstream

Leaving in draft until I can test the cuda part (some pieces missing)


AbrilRBS (approved):
Looks great :)

danimtb (approved):

PR #24709, opened 2024-07-24
openssl: add `use_validated_fips` option

gegles:
### Summary
Changes to recipe:  **openss/***

#### Motivation

The end goal here is for the openssl package to be able to build and provide the validated FIPS module from another version of itself. This takes advantage of [this recent change in conan](https://github.com/conan-io/conan/pull/16132). 

This recipe change is following @jcar87's suggestion [here](https://github.com/conan-io/conan-center-index/issues/22796#issuecomment-2079830251).

#### Details

* Add the `use_validated_fips` boolean option. `False` by default
* If `use_validated_fips` is `True` then there is no need to build the current version of the FIPS module so set `no_fips = True`
* If `use_validated_fips` is `True` then add a hidden dependency to the latest FIPS validated version of `openssl`, i.e. `openssl/3.0.9`
* Depending on whether either `not no_fips` or `use_validated_fips` is true, we copy the appropriate `fips` library into the final pakcage folder.

I did try to also add a programatic check in the `test_package` (following the `digest_legacy.c` example, but, for the FIPS provider to load properly at all (in order to test the version), it requires the `fipsmodule.cnf` to be properly generated via the `openssl fipsinstall` command ... All this seems maybe more complexity than desired...

FWIW,

I've tested this as part of our software depending on this which does the proper `fipsinstall` command:
```
OPENSSL_CONF=/etc/faspio-gateway/openssl.cnf OPENSSL_MODULES=/usr/local/lib/ossl-modules /usr/local/sbin/openssl@faspio-gateway list -providers
Providers:
  base
    name: OpenSSL Base Provider
    version: 3.2.2
    status: active
  fips
    name: OpenSSL FIPS Provider
    version: 3.0.9
    status: active
```

As you can see, the default provider is `3.2.2` (from the latest version of this package I used as a dependency in my project whereas the FIPS provider is `3.0.9` as per the result of using `use_validated_fips` option.

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -118,7 +118,7 @@ class OpenSSLConan(ConanFile):
         if self.options.use_validated_fips:
             # As of version 3.3.1, the FIPS module is validated for the following versions
             # see https://openssl-library.org/source/ (excluding ancient 3.0.0)
-            versions = ['3.0.8', '3.0.9']
+            versions = ['3.0.8', '3.0.9', '3.1.2']
             versions = sorted([Version(v) for v in versions], reverse=True)
 
             # Find the closest version that is less than or equal to the current version
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -1,7 +1,10 @@
+import re
+from io import StringIO
+
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
+from conan.errors import ConanException, ConanInvalidConfiguration
 from conan.tools.apple import fix_apple_shared_install_name, is_apple_os, XCRun
-from conan.tools.build import build_jobs
+from conan.tools.build import build_jobs, can_run
 from conan.tools.files import chdir, copy, get, replace_in_file, rm, rmdir, save
 from conan.tools.gnu import AutotoolsToolchain
 from conan.tools.layout import basic_layout
@@ -89,6 +92,7 @@ class OpenSSLConan(ConanFile):
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
         "tls_security_level": [None, 0, 1, 2, 3, 4, 5],
+        "use_validated_fips": [True, False],
     }
     default_options = {key: False for key in options.keys()}
     default_options["fPIC"] = True
@@ -108,6 +112,37 @@ class OpenSSLConan(ConanFile):
     @property
     def _use_nmake(self):
         return self._is_clang_cl or is_msvc(self)
+
+    @property
+    def _fips_version(self):
+        if self.options.use_validated_fips:
+            # As of version 3.3.1, the FIPS module is validated for the following versions
+            # see https://openssl-library.org/source/ (excluding ancient 3.0.0)
+            versions = ['3.0.8', '3.0.9']
+            versions = sorted([Version(v) for v in versions], reverse=True)
+
+            # Find the closest version that is less than or equal to the current version
+            fips_validated_version = next((v for v in versions if v <= Version(self.version)), None)
+            return fips_validated_version
+        else:
+            return self.version
+
+    @property
+    def _is_fips_enabled(self):
+        return not self.options.no_fips or self.options.use_validated_fips
+
+    @property
+    def _provides_validated_fips(self):
+        return self.options.use_validated_fips and self.version == self._fips_version
+
+    @property
+    def _fips_provider_dir(self):
+        if self.options.use_validated_fips and not self._provides_validated_fips:
+            return self.dependencies["openssl"].runenv_info.vars(self)["OPENSSL_MODULES"]
+        elif not self.options.no_fips:
+            return os.path.join(self.source_folder, "providers")
+        else:
+            return None
 
     def config_options(self):
         if self.settings.os != "Windows":
@@ -129,9 +164,24 @@ class OpenSSLConan(ConanFile):
         if not self.options.no_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
 
+        if self.options.use_validated_fips:
+            if not self._provides_validated_fips:
+                self.output.info(f"Using validated FIPS module from openssl/{self._fips_version}")
+                self.requires(f"openssl/{self._fips_version}", visible=False, libs=False, headers=False, run=False, options={'no_fips': False})
+            else:
+                self.output.info(f"Using validated FIPS module from self (i.e. {self._fips_version})")
+
     def validate(self):
         if self.settings.os == "iOS" and self.options.shared:
             raise ConanInvalidConfiguration("OpenSSL 3 does not support building shared libraries for iOS")
+
+        if self.options.use_validated_fips:
+            if self._fips_version is None:
+                raise ConanInvalidConfiguration(f"OpenSSL {self.version} - no compatible FIPS validated version found")
+            if self.options.no_fips:
+                raise ConanInvalidConfiguration("FIPS support is requested, but no_fips is set to True")
+            elif not self._provides_validated_fips and self.dependencies["openssl"].options.no_fips:
+                raise ConanInvalidConfiguration(f"In order to use FIPS module from openssl/{self._fips_version}, it needs to be built with `no_fips` option set to False")
 
     def build_requirements(self):
         if self.settings_build.os == "Windows":
@@ -366,7 +416,12 @@ class OpenSSLConan(ConanFile):
         else:
             args.append("-fPIC" if self.options.get_safe("fPIC", True) else "no-pic")
 
-        args.append("no-fips" if self.options.get_safe("no_fips", True) else "enable-fips")
+        # pass no-fips to the current build if:
+        # - use_validated_fips is enabled and using the fips module from a different version
+        # - user requested no-fips
+        no_fips = self.options.use_validated_fips and not self._provides_validated_fips or self.options.no_fips
+        args.append("no-fips" if no_fips else "enable-fips")
+
         args.append("no-md2" if self.options.get_safe("no_md2", True) else "enable-md2")
         if str(self.options.tls_security_level) != "None":
             args.append(f"-DOPENSSL_TLS_SECURITY_LEVEL={self.options.tls_security_level}")
@@ -397,7 +452,7 @@ class OpenSSLConan(ConanFile):
             ])
 
         for option_name in self.default_options.keys():
-            if self.options.get_safe(option_name, False) and option_name not in ("shared", "fPIC", "openssldir", "tls_security_level", "capieng_dialog", "enable_capieng", "zlib", "no_fips", "no_md2"):
+            if self.options.get_safe(option_name, False) and option_name not in ("shared", "fPIC", "openssldir", "tls_security_level", "capieng_dialog", "enable_capieng", "zlib", "no_fips", "no_md2", "use_validated_fips"):
                 self.output.info(f"Activated option: {option_name}")
                 args.append(option_name.replace("_", "-"))
         return args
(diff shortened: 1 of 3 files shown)


gegles:
FYI @jcar87 @bemehiser @Croydon @memsharded @kulkarniamit @hoyhoy

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -551,8 +569,8 @@ def package(self):
                 if file.endswith(".a"):
                     os.unlink(os.path.join(libdir, file))
 
-        if not self.options.no_fips:
-            provdir = os.path.join(self.source_folder, "providers")
+        if self._is_fips_enabled:
+            provdir = self._fips_provider_dir
```
kulkarniamit, line 573: Works as expected 👍🏼 
Just a minor thing, does it help to add a `self.output.info()` to print the `provdir` path ?
It will make it easier to confirm that the certified fips module has been copied from 3.0.9 directory.

gegles: Good point @kulkarniamit, I'll see what I can do... 

Actually, ideally, I was trying to add some testing logic in the `test_package` code to be able to properly verify the version of the `fips` module... For example, that `version == 3.0.9` when `use_validated_fips == True` or `version == tested_package_version` otherwwise, but both programmatically or simply using the `openssl list -providers` command require a properly `fipsinstall`ed module before loading it to check the version...

Do you know of any ways to check the version of the `fips` module without it needing to be fully installed?

kulkarniamit: As you mentioned, printing buildinfo related information for fips provider requires loading the provider. Loading FIPS provider requires having a configuration file with `module-mac` to enable self tests of the module.

If we don't prefer checking the integration with `openssl` command, we are left with two options:
* Use `strings` command (or any other equivalent) and search for presence of version number in module. Need to ensure we've a consistent command or python string search that platform independent
  ```bash
  # Linux
  $ strings <path_to_fips_module> | grep -i -A1 buildinfo
  buildinfo
  3.0.9
  ```
* Print the fips module source and destination directory information

Though its extra work, running `openssl fipsinstall` to generate `fipsmodule.cnf` and loading the provider programmatically in `test_package.c` would be the ideal future-proof platform-independent approach to printing version information.

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -136,6 +140,8 @@ def layout(self):
     def requirements(self):
         if not self.options.no_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
+        if self.options.use_validated_fips:
+            self.requires("openssl/3.0.9", visible=False, libs=False, headers=False, run=False)
```
Nekto89, line 144: this version should probably added again to config.yml and conandata.yml

jcar87:
This looks really good @gegles ! Thanks for this.

Some comments from me:

I would probably keep, in the recipe, a list of "fips validated" versions (as per the OpenSSL documentation), and have a logic such that if `use_validated_fips` is `True`, then we either:
    - find the closest version, lower than the current one, that is fips-validated
    - if the current version is already fips validated, _don't_ depend on it (avoid a recipe depending on itself with the same version) - the current version should be good enough on its own, correct?

This might look something like:
```
    @property
    def _fips_validated_version(self):
        # As of version 3.3.1, the FIPS module is validated for the following versions
        # see https://openssl-library.org/source/   
        versions = ['3.0.0', '3.0.8', '3.0.9']
        versions = sorted([Version(v) for v in versions], reverse=True)

        # Find the closest version that is less than or equal to the current version
        fips_validated_version = next((v for v in versions if v <= Version(self.version)), None)
        return fips_validated_version

    def requirements(self):
            # ...
            fips_version = self._fips_validated_version
            if fips_version and fips_version != self.version:
                self.output.info(f"Using validated FIPS module from openssl version:{fips_version}")
                self.requires(f"openssl/{fips_version}", visible=False, libs=False, headers=False, run=False, options={'no_fips': False})
```


I would probably add some additional checks in the `validate()` method as there are a few cases that would be invalid:
* there is no earlier version that is fips validated (not the case since eventually 3.0.0 was validated)
* if we are depending on another version for fips validation, but _that_ dependency is being built with `no_fips=True`

This may look like:
```
        if self.options.use_validated_fips:
            fips_version = self._fips_validated_version
            if fips_version is None:
                raise ConanInvalidConfiguration(f"OpenSSL {self.version} - no compatible FIPS validated version")
            if self.options.no_fips:
                raise ConanInvalidConfiguration(f"Fips support is requested, but no_fips is set to True")
            elif fips_version != self.version and self.dependencies["openssl"].options.no_fips:
                raise ConanInvalidConfiguration(f"In order to use FIPS module from openssl/{fips_version}, it needs to be built with `no_fips` option set to False")
```


Note that I've seen in the PR that this is done:
```

        if self.options.use_validated_fips == True:
            self.options.no_fips = True
```

however I would err in the side of caution - bear in mind that if externally a user does `-o "openssl/*:no_fips=True" -o "openssl/*:use_validated_fips=True"` - it will apply to both versions of openssl in the graph. 

What I would suggest is just rely on the validation - assume that if users pass these, it's a contradiction. And to avoid building the fips-module when it is being provided by an earlier version, I would do in the configure args method:

```
        # pass no-fips to the current build if:
        # - use_validated_fips is enabled and using the fips module from a different version
        # - user requested no-fips
        no_fips = self.options.use_validated_fips and self._fips_validated_version != self.version or self.options.no_fips
        args.append("no-fips" if no_fips else "enable-fips")
```


The above are all very rough - but gives an idea. 

I also agree with @Nekto89  - we should re-add 3.0.9 to the list of maintained versions.  Perhaps if we do this, we may choose to keep the list in the recipe constrained to just `3.0.9`

As for the installation, the openssl documentation does specify that the self-tests should be run - personally, if a `use_validated_fips` is meant to offer "strong" expectation that we are using a validated fips, I would test this not just in the test package, but in the recipe's build method itself, at the very end of the build. This would have to be guarded with a `can_run` - but I dont have enough experience with the validation process here - and I'm not sure where the resulting config file should reside

recipes/openssl/3.x.x/conanfile.py
gegles: Good call, added it back in [3c52995](https://github.com/conan-io/conan-center-index/pull/24709/commits/3c5299580b24848f22b4da856e4ea3ed38ec9210)

gegles:
> h experience with the validation process here - and I'm not sure where the resulting config file should reside

Thanks for the review and all the feedback @jcar87!

I agree with most everything and will see what I can do.

I had already just re-added `3.0.9` based on @Nekto89's feedback.

I am seriously questioning whether it would make any sense for anybody to use anything but the latest validated version (i.e. `v3.0.9`).... I would think it's probably best to instead encourage folks to only use the latest and thus only specify `3.0.9` ...
Because of that I would rather not offer a list, but instead just offer the latest validated version... LMK.

Thanks!

gegles:
> Some comments from me:
> 
> I would probably keep, in the recipe, a list of "fips validated" versions (as per the OpenSSL documentation), and have a logic such that if `use_validated_fips` is `True`, then we either: - find the closest version, lower than the current one, that is fips-validated - if the current version is already fips validated, _don't_ depend on it (avoid a recipe depending on itself with the same version) - the current version should be good enough on its own, correct?
> 
> This might look something like:
> 
> ```
>     @property
>     def _fips_validated_version(self):
>         # As of version 3.3.1, the FIPS module is validated for the following versions
>         # see https://openssl-library.org/source/   
>         versions = ['3.0.0', '3.0.8', '3.0.9']
>         versions = sorted([Version(v) for v in versions], reverse=True)
> 
>         # Find the closest version that is less than or equal to the current version
>         fips_validated_version = next((v for v in versions if v <= Version(self.version)), None)
>         return fips_validated_version
> 
>     def requirements(self):
>             # ...
>             fips_version = self._fips_validated_version
>             if fips_version and fips_version != self.version:
>                 self.output.info(f"Using validated FIPS module from openssl version:{fips_version}")
>                 self.requires(f"openssl/{fips_version}", visible=False, libs=False, headers=False, run=False, options={'no_fips': False})
> ```
> 
> I would probably add some additional checks in the `validate()` method as there are a few cases that would be invalid:
> 
> * there is no earlier version that is fips validated (not the case since eventually 3.0.0 was validated)
> * if we are depending on another version for fips validation, but _that_ dependency is being built with `no_fips=True`
> 
> This may look like:
> 
> ```
>         if self.options.use_validated_fips:
>             fips_version = self._fips_validated_version
>             if fips_version is None:
>                 raise ConanInvalidConfiguration(f"OpenSSL {self.version} - no compatible FIPS validated version")
>             if self.options.no_fips:
>                 raise ConanInvalidConfiguration(f"Fips support is requested, but no_fips is set to True")
>             elif fips_version != self.version and self.dependencies["openssl"].options.no_fips:
>                 raise ConanInvalidConfiguration(f"In order to use FIPS module from openssl/{fips_version}, it needs to be built with `no_fips` option set to False")
> ```
> 
> Note that I've seen in the PR that this is done:
> 
> ```
> 
>         if self.options.use_validated_fips == True:
>             self.options.no_fips = True
> ```
> 
> however I would err in the side of caution - bear in mind that if externally a user does `-o "openssl/*:no_fips=True" -o "openssl/*:use_validated_fips=True"` - it will apply to both versions of openssl in the graph.
> 
> What I would suggest is just rely on the validation - assume that if users pass these, it's a contradiction. And to avoid building the fips-module when it is being provided by an earlier version, I would do in the configure args method:
> 
> ```
>         # pass no-fips to the current build if:
>         # - use_validated_fips is enabled and using the fips module from a different version
>         # - user requested no-fips
>         no_fips = self.options.use_validated_fips and self._fips_validated_version != self.version or self.options.no_fips
>         args.append("no-fips" if no_fips else "enable-fips")
> ```
> 
> The above are all very rough - but gives an idea.

@jcar87, with [f50c238](https://github.com/conan-io/conan-center-index/pull/24709/commits/f50c2386163bc01f73a28906c149e03dd3dc3f49), I've now incorporate most of your suggestions/feedback. 
I've restricted to only `3.0.8` and `3.0.9`. I've left `3.0.0` behind as it would not even work and it's ancient.

I've manually tested the various error conditions:
```
conan create -b missing --version 3.2.2 -o "&:use_validated_fips=True" -o "&:no_fips=True" 3.x.x

ERROR: There are invalid packages:
openssl/3.2.2: Invalid: FIPS support is requested, but no_fips is set to True
```

```
conan create -b missing --version 3.2.2 -o "&:use_validated_fips=True" -o "openssl/3.0.9:no_fips=True" 3.x.x
ERROR: There are invalid packages:
openssl/3.2.2: Invalid: In order to use FIPS module from openssl/3.0.9, it needs to be built with `no_fips` option set to False
```

```
conan create -b missing --version 3.0.7 -o "&:use_validated_fips=True" 3.x.x

ERROR: There are invalid packages:
openssl/3.0.7: Invalid: OpenSSL 3.0.7 - no compatible FIPS validated version found
```

gegles:
> As for the installation, the openssl documentation does specify that the self-tests should be run - personally, if a `use_validated_fips` is meant to offer "strong" expectation that we are using a validated fips, I would test this not just in the test package, but in the recipe's build method itself, at the very end of the build. This would have to be guarded with a `can_run` - but I dont have enough experience with the validation process here - and I'm not sure where the resulting config file should reside

@jcar87 take a look at my latest changes... I've now added a `fipsinstall` and `fipsinstall -verify` steps in the `package()` phase...

I got caught by a linter issue (re using `stderr`)... I guess it's not available on the min conan version used on conan center... So I'll need to refactor/revisit.

Anyway let me know what you think and if I am the right track...

Cheers. G.

kulkarniamit:
> - find the closest version, lower than the current one, that is fips-validated

> ```
> conan create -b missing --version 3.0.7 -o "&:use_validated_fips=True" 3.x.x
> ERROR: There are invalid packages:
> openssl/3.0.7: Invalid: OpenSSL 3.0.7 - no compatible FIPS validated version found
>```

According to OpenSSL maintainers: https://github.com/openssl/openssl/discussions/25048#discussioncomment-10195512
> "_The ABI is deliberately set so any 3.x fips module works with any 3.x.y in general._"

I ran a few commands to verify that OpenSSL 3.0.7 binary can work with 3.0.9 FIPS provider
```
$ ./bin/openssl version -v
OpenSSL 3.0.7 1 Nov 2022 (Library: OpenSSL 3.0.7 1 Nov 2022)

$ OPENSSL_MODULES=../openssl-3.0.9/lib64/ossl-modules ./bin/openssl list -providers
Providers:
  base
    name: OpenSSL Base Provider
    version: 3.0.7
    status: active
  fips
    name: OpenSSL FIPS Provider
    version: 3.0.9
    status: active

$ export OPENSSL_MODULES=../openssl-3.0.9/lib64/ossl-modules 

$ echo "helloworld" | ./bin/openssl dgst -sha1
SHA1(stdin)= e7509a8c032f3bc2a8df1df476f8ef03436185fa

$ echo "helloworld" | ./bin/openssl dgst -sha256
SHA2-256(stdin)= 8cd07f3a5ff98f2a78cfc366c13fb123eb8d29c1ca37c79df190425d5b9e424d

$ echo "helloworld" | ./bin/openssl dgst -md5
Error setting digest
40876E37977F0000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:373:Global default library context, Algorithm (MD5 : 102), Properties ()
40876E37977F0000:error:03000086:digital envelope routines:evp_md_init_internal:initialization error:crypto/evp/digest.c:254:
```

@gegles @jcar87 Any particular reason why we're limiting lower versions of OpenSSL 3.0.x (x < 9) to not work with 3.0.8 or 3.0.9 FIPS provider?

jcar87:
> @gegles @jcar87 Any particular reason why we're limiting lower versions of OpenSSL 3.0.x (x < 9) to not work with 3.0.8 or 3.0.9 FIPS provider?

none really! we didn't have that context.
Although bear in mind that we are unlikely to maintain up-to-date recipes for versions dating that far back - considering that OpenSSL 3.x itself is backwards compatible

szigetics:
@gegles : Great update! Thank you for this! 🙏

I see that for this updated recipe to work, the changes from https://github.com/conan-io/conan/pull/16132 are required.
So a conan version update is necessary for this updated recipe to work.

I also see that the there is an official conan release which has the necessary update : https://github.com/conan-io/conan/releases/tag/2.4.0 .
So it's definitely available for conan2.

However : I cannot find a conan1 release with this change.

So my question : **could the above recipe update work with conan1 as well? Which conan1 version have to be used?**

gegles:
> So my question : **could the above recipe update work with conan1 as well? Which conan1 version have to be used?**

@szigetics, I may be wrong, but my understanding is that this feature would only work with Conan v2 and especially with the https://github.com/conan-io/conan/pull/16132 changeset as you pointed out.

gegles:
I know it's not the preferred way, but, in order to resolve conflicts and rebase onto the latest `master`, I ended up squashing and rebasing the branch. thx!

jcar87:
> I know it's not the preferred way, but, in order to resolve conflicts and rebase onto the latest `master`, I ended up squashing and rebasing the branch. thx!

Oh don't worry about it! Personally I prefer squash/rebase - The issue here is the github interface when a PR has already been reviewed and the next change is rebase, its hard to see what changed vs the previous review - but rebasing tends to be a lot safer in general

gegles:
Based on [this latest blogpost](https://openssl-library.org/post/2025-03-11-fips-140-3/index.html), I've now also added `3.1.2` as the latest FIPS validated version.

@jcar87 any chance we could make progress toward merging this in? Not sure what else needs to be done... thx!

jcar87:
> Based on [this latest blogpost](https://openssl-library.org/post/2025-03-11-fips-140-3/index.html), I've now also added `3.1.2` as the latest FIPS validated version.
> 
> @jcar87 any chance we could make progress toward merging this in? Not sure what else needs to be done... thx!

Working on this - I have rebased the branch on top of the most recent changes in `master` - you'll have to resync your local branch before pushing new commits - aiming to have this completed today, hopefully

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -536,6 +591,58 @@ def _replace_runtime_in_file(self, filename):
...
+        copy(self, module_filename, src=provdir, dst=modules_dir)
+
+        self.output.info("Testing FIPS module (via fipsinstall & fipsinstall -verify)")
+        if can_run(self):
+            openssl_bin = os.path.join(self.package_folder, "bin", "openssl")
+            fips_module = os.path.join(modules_dir, module_filename)
+            fips_cnf = os.path.join(self.build_folder, "fips.cnf")
+
+            # fipsinstall
+            fipsinstall_command = [openssl_bin, "fipsinstall", f"-module {fips_module}", f"-out {fips_cnf}"]
```
jcar87, line 607: A couple of questions:
* How does this work when cross building an `openssl_bin` is of a foreign architecture that won't run locally? (may be a generic OpenSSL question rather than a Conan specific one)


From the openssl [documentation](https://github.com/openssl/openssl/blob/master/README-FIPS.md#installing-the-fips-provider):

> The FIPS module must have the self tests run, and the FIPS module config file output generated on every machine that it is to be used on. For OpenSSL 3.0, you must not copy the FIPS module config file output data from one machine to another.

I can see the latter is correct (the config file is not being copied from one machine to another) - but should the self-tests and the .cnf file be run on any machine where openssl is installed (and might run?) - this may be a good candidate for the `finalize()` method https://blog.conan.io/2024/09/03/New-finalize-method.html 

My concern is here is perhaps more "semantical" - the `use_validated_fips` name of the option would sugesst that it is "validated", when in reality and as per the docs, there's additional responsibility on the consumer side when it comes to self-test and configuration, so the option is more "use_fips_module_from_a_validated_version", rather than offering stronger guarantees.

jcar87 (commented):
Added a couple of comments (non exhaustive review), and an additional question

`openssl` uses a config file (e.g. `openssl.cnf`) that I'm assuming is also used by the libraries itself, correct?

The recipe seems to be hardcoding this altogether (which makes me question what happens when you build on one machine and run in another), but also on Linux this is set to the systems' openssl:

https://github.com/conan-io/conan-center-index/blob/master/recipes/openssl/3.x.x/conanfile.py#L332-L335


but it looks like for fips, the configuration needs to explicitly reference the generated file after the self-validation

https://github.com/openssl/openssl/blob/85cabd94958303859b1551364a609d4ff40b67a5/doc/man7/fips_module.pod#L73-L97

So how would this work on Linux? Or any other system?  Or should we default to the other cases listed in the docs? https://docs.openssl.org/master/man7/fips_module/#making-all-applications-use-the-fips-module-by-default

Issue #23277, opened 2024-03-26
[request] version bump for common offenders


valgur:
These are some of the most common offenders for version conflicts currently, in my experience. I propose doing a version bump for the listed recipes, about a 100 packages in total. I would add Boost to this list as well, but that would double the number.

I'm willing to create the PRs (at a reasonable rate) if I get greenlit for that. @jcar87, I think you advised against doing it unprompted it in one of the PRs by @mayeut, so I'm pinging you personally.

Alternatively, maybe some of the listed dependencies could be added to the permitted version ranges list.

# expat -> `[>=2.6.2 <3]`

Version ranges for expat are allowed and recommended now: https://github.com/conan-io/conan-center-index/pull/23510#discussion_r1566001446

2.6.1 and below are vulnerable
https://nvd.nist.gov/vuln/detail/CVE-2024-28757
https://repology.org/project/expat/information

- [x] [aaf](https://github.com/conan-io/conan-center-index/blob/master/recipes/aaf/all/conanfile.py)	2.5.0
- [x] [apr-util](https://github.com/conan-io/conan-center-index/blob/master/recipes/apr-util/all/conanfile.py)	2.5.0
- [x] [avahi](https://github.com/conan-io/conan-center-index/blob/master/recipes/avahi/all/conanfile.py)	2.5.0
- [x] [cairo](https://github.com/conan-io/conan-center-index/blob/master/recipes/cairo/meson/conanfile.py)	2.5.0
- [x] [cpython](https://github.com/conan-io/conan-center-index/blob/master/recipes/cpython/all/conanfile.py)	2.6.0
- [x] [dbus](https://github.com/conan-io/conan-center-index/blob/master/recipes/dbus/all/conanfile.py)	2.6.0
- [x] [exiv2](https://github.com/conan-io/conan-center-index/blob/master/recipes/exiv2/all/conanfile.py)	2.5.0
- [x] [fontconfig](https://github.com/conan-io/conan-center-index/blob/master/recipes/fontconfig/all/conanfile.py)	2.6.0
- [x] [freexl](https://github.com/conan-io/conan-center-index/blob/master/recipes/freexl/all/conanfile.py)	2.5.0
- [ ] [gdal](https://github.com/conan-io/conan-center-index/blob/master/recipes/gdal/all/conanfile.py)	2.6.2
- [x] [gdcm](https://github.com/conan-io/conan-center-index/blob/master/recipes/gdcm/all/conanfile.py)	2.5.0
[...]

PR #24711, opened 2024-07-25
bump openvdb boost requirement

EstebanDugueperoux2:
### Summary
Changes to recipe:  **openvdb/***

#### Motivation
Upgrade boost dependency to avoid conflict on another PR.


recipes/openvdb/all/conanfile.py
```diff
@@ -130,10 +130,10 @@ def package_id(self):
 
     def requirements(self):
         # https://github.com/AcademySoftwareFoundation/openvdb/blob/v10.0.1/doc/dependencies.txt#L36-L84
-        self.requires("boost/1.84.0", transitive_headers=True)
+        self.requires("boost/1.85.0", transitive_headers=True)
         self.requires("onetbb/2021.10.0", transitive_headers=True, transitive_libs=True)
         if self.options.use_imath_half:
-            self.requires("imath/3.1.9", transitive_headers=True, transitive_libs=True)
+            self.requires("imath/3.1.11", transitive_headers=True, transitive_libs=True)
```
valgur, line 136: Most recipes on CCI still rely on `3.1.9` so I would suggest sticking to that one here and in the OpenUSD recipe to avoid conflicts, unless you have a specific reason to use the newer version. See https://github.com/conan-io/conan-center-index/issues/23277

EstebanDugueperoux2: Thanks for your feedback.

PR #24713, opened 2024-07-25
qt6: reduce version burden

jcar87:
### Summary
qt6: reduce number of maintained versions

#### Motivation
* we are maintaining too many versions unnecessarily (qt is backwards compatible and all patch versions have strong compatibility guarantees)
* this will improve CI turnaround for all future Qt PRs

#### Details
* All minor releases of Qt 6 (6.y) are source and binary compatible with the previous versions - that it, it should be safe for users to use the newer versions
* Patch releases (6.y.z) are backwards and forwards compatible - Conan already models this well 
* Keeping 6.5 as that is the current LTS

We would recommend users to depend on a version range rather than a specific version, and will be moving forward with this in other recipes in Conan Center that depend on Qt 6.

Note: intentionally leaving all logic and patches pertaining to the no longer maintained versions, in case there are strong reasons to revert.

More details:
* https://wiki.qt.io/Qt-Version-Compatibility
* 
---


uilianries (approved):
Less is more.

AbrilRBS (approved):

PR #24714, opened 2024-07-25
wasmtime: add version 27.0.0

toge:
### Summary
Changes to recipe:  **wasmtime/27.0.0**

#### Motivation
Major version up.

#### Details
https://github.com/bytecodealliance/wasmtime/compare/v21.0.0...v27.0.0

---

### DIFF
--- a/recipes/wasmtime/all/conandata.yml
+++ b/recipes/wasmtime/all/conandata.yml
@@ -1,34 +1,34 @@
 sources:
-  "23.0.1":
+  "25.0.0":
     Windows:
       "x86_64":
-        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v23.0.1/wasmtime-v23.0.1-x86_64-windows-c-api.zip"
-        sha256: "5278da4f3c0aec05c380afcfa8bd5736e88e6a1609036864a16449fa86914cf2"
+        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v25.0.0/wasmtime-v25.0.0-x86_64-windows-c-api.zip"
+        sha256: "870788d74a5b779ba6c3338bef82f814fe2cbafe23e24311572a1daa79365080"
     MinGW:
       "x86_64":
-        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v23.0.1/wasmtime-v23.0.1-x86_64-mingw-c-api.zip"
-        sha256: "ecb440151a060bcdf6b862a619c428f2b355a1d09c1f6881c6474425f8169572"
+        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v25.0.0/wasmtime-v25.0.0-x86_64-mingw-c-api.zip"
+        sha256: "3730db3ab5899fa5a20141fc1a455f2c15b051464b6723893b35264d94db2485"
     Linux:
       "x86_64":
-        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v23.0.1/wasmtime-v23.0.1-x86_64-linux-c-api.tar.xz"
-        sha256: "3b9e34e074123383cc9598eee52ef1fa4ee2401733276262ba172bcf964173de"
+        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v25.0.0/wasmtime-v25.0.0-x86_64-linux-c-api.tar.xz"
+        sha256: "e4dc50edbc432ff7fe12ff02137bd7889194305a74a28a7f2742856ced2d90aa"
       "armv8":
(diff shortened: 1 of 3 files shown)


perseoGI (approved):
LGTM

recipes/wasmtime/all/conanfile.py
```diff
@@ -68,6 +74,8 @@ def validate(self):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
+        if self._is_glibc_older_than_2_28:
+            raise ConanInvalidConfiguration(f"{self.ref} requires glibc 2.28 or later.")
```
uilianries, line 78: Is glib 2.28 required for any older version too, or only 23.0.1? I guess something is missing, because this package was working before.

toge: @uilianries 
Thank you for your review.
You're right.
wasmtime 23.0.1 only requires glibc 2.28.
I fixed it.

toge:
@uilianries 
`pr-merge` in CI is a last piece.
Do you have any way to solve this problem?

perseoGI (approved):
LGTM

ErniGH (approved):

perseoGI (approved):
LGTM

AbrilRBS (approved):

AbrilRBS (commented):
@toge can you remove the 6.0.1 from the config.yml? There is no corresponding entry in the conandata.yml file :)

toge:
@AbrilRBS 
Thank you for your review!
I fixed.

AbrilRBS (approved):

recipes/wasmtime/all/conandata.yml
```diff
@@ -1,4 +1,34 @@
...
+    Macos:
+      "x86_64":
+        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v25.0.2/wasmtime-v25.0.2-x86_64-macos-c-api.tar.xz"
+        sha256: "a0b33054fd3b5cf9707d1a39a68a80d11600e2fcdf6074ba4263b9c339bfe64b"
+      "armv8":
+        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v25.0.2/wasmtime-v25.0.2-aarch64-macos-c-api.tar.xz"
+        sha256: "8cc68c43559c78e103bebf63ba16b768f7bb9c0ba164649279c1a09be690b416"
+    Android:
+      "armv8":
+        url: "https://github.com/bytecodealliance/wasmtime/releases/download/v25.0.2/wasmtime-v25.0.2-aarch64-linux-c-api.tar.xz"
```
AbrilRBS, line 30: This is the linux release, but there seems to be a typo? The release contains Android files. Is this duplicatin in pourpose?

AbrilRBS (changes requested):
blocking while we check

jcar87:
As @AbrilRBS  mentions, we just ran this PR in our new pipeline (which is still internal) and detected two issues:

Screenshot 2024-10-10 at 10 24 31

The first one is that the config.yaml and the conandata.yaml seem to have an inconsistency (there's a version in `config.yml` that is not present in conandata. This is not detected by the current linter, and it is not detected when the `source()` method is executed for the missing version - because that version is not built due to invalid configuration.   

I suspect this issue predates this PR - the new pipeline would have stopped this at the PR that introduced the change.

On the other hand, there is a second issue where there are two duplicated checksums - this is already reported by the linter that we enabled yesterday (results here https://github.com/conan-io/conan-center-index/pull/24714/checks?check_run_id=31342639853)

The warning message needs improving - however it appears to be correct. The entry for android is the same as for Linux, but it seems upstream does have a separate package for Android: https://github.com/bytecodealliance/wasmtime/releases/ 
It's still within the realm of possibilities that since Android _is_ Linux at the low level, the "c-package" is the same for Android and Linux, but it's worth checking this.

(the other warnings by the linter can be disregarded, we are working on improving the detection of mismatches between the version and the URL)

toge:
@AbrilRBS 
There is an error on Linux CI pipeline.
```
ERROR: There are invalid packages:
wasmtime/27.0.0: Invalid: wasmtime/27.0.0 requires glibc 2.28 or later.
```

It is caused by `ConvanInvalidConfiguration` in `validate()`.
```python
        if Version(self.version) >= "23.0.0" and self._is_glibc_older_than_2_28:
            raise ConanInvalidConfiguration(f"{self.ref} requires glibc 2.28 or later.")
```

I think it is a valid way to disable building the recipe on target environment.
Please tell me how to fix this problem.

toge:
temporary closed

PR #24720, opened 2024-07-26
[sentry-native] Add 0.7.8

MartinDelille:
### Summary

Add version : **sentry-native/0.7.7**

#### Details

---

### DIFF
--- a/recipes/sentry-native/all/conanfile.py
+++ b/recipes/sentry-native/all/conanfile.py
@@ -51,7 +51,7 @@ class SentryNativeConan(ConanFile):
 
     @property
     def _min_cppstd(self):
-        return "17" if Version(self.version) < "0.7.7" else "20"
+        return "17"
 
     @property
     def _minimum_compilers_version(self):
@@ -63,7 +63,7 @@ class SentryNativeConan(ConanFile):
             "msvc": "191",
             "gcc": minimum_gcc_version,
             "clang": "3.4",
-            "apple-clang": "10.0",
+            "apple-clang": "5.1",
         }
 
     def config_options(self):
@@ -128,6 +128,8 @@ class SentryNativeConan(ConanFile):
             )
         if self.options.transport == "winhttp" and self.settings.os != "Windows":
             raise ConanInvalidConfiguration("The winhttp transport is only supported on Windows")
+        if self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "10.0":
+            raise ConanInvalidConfiguration("apple-clang < 10.0 not supported")
 
     def build_requirements(self):
         if self.settings.os == "Windows":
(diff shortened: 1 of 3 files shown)


MartinDelille:
Hi @uilianries ! Do you have any idea why the build fails because of the range library ? Requiring C++ 20 for 0.7.7 and upper is not enough ?

MartinDelille:
@ErniGH Do you have a clue to help me understand what is going on with these error ? 0.7.9 is out. Shall I push a new version ?

recipes/sentry-native/all/conanfile.py
```diff
@@ -46,12 +46,12 @@ class SentryNativeConan(ConanFile):
...
         "crashpad_with_tls": "openssl",
         "with_breakpad": "sentry",
-        "wer": False
+        "wer": False,
     }
 
     @property
     def _min_cppstd(self):
-        return "17"
+        return "17" if Version(self.version) < "0.7.7" else "20"
```
ErniGH, line 54: Hi @MartinDelille , I've been testing the PR on my machine without using this version requirement, and both on Mac and Linux, version 0.7.8 works for me. I've checked upstream, and it seems that the version is 17.

https://github.com/getsentry/sentry-native/blob/master/CMakeLists.txt#L33

Maybe we can just undo this change and see how it works.

As for the other change from clang 5.1 to 10.0, I can't find a reason for it. Could you provide any justification for this change?


Sorry for the delay :(

uilianries:
There is no newer Apple Clang version than 13.0 in CCI, but testing locally with Apple Clang 15.0 is working fine:

[sentry-native-0.7.0-apple-clang-15.log](https://github.com/user-attachments/files/16931973/sentry-native-0.7.0-apple-clang-15.log)

The Cpp Reference says Apple Clang 14.0 should be enough for C++20 bit_cast:  
https://en.cppreference.com/w/cpp/compiler_support/20

ErniGH (approved):
Hey @MartinDelille , it looks like it's finally working. It seems that the new version indeed needed C++ 20 to support `Concepts` and `bit_cast`, @uilianries saw it.

Many thanks to both of you. ❤️

uilianries (approved):

mishander:
@MartinDelille Can you please explain to me why we have a comment about C++20 here:
```
if self.options.get_safe("with_crashpad") == "sentry":
# Sentry-native 0.7.8 requires C++20: Concepts and bit_cast
#https://github.com/chromium/mini_chromium/blob/e49947ad445c4ed4bc1bb4ed60bbe0fe17efe6ec/base/numerics/byte_conversions.h#L88
```
            
To me, at least at a first glance, it seems like Sentry is using its version of crashpad which is C++17?

https://github.com/getsentry/mini_chromium/blob/fd75723d16e99fb1e6d1487c4f278a5ce618fdf0/base/numerics/byte_conversions.h

shouldn't it be 
`if self.options.get_safe("with_crashpad") == "google"`
or something like that?

PR #18979, opened 2023-07-31
libfabric: migrate to Conan v2, add missing dependencies

valgur:
Requires, optionally
- #18978

closes #19361

PR #24725, opened 2024-07-26
rdma-core: add v52.0, enable all modules

valgur:
### Summary
Changes to recipe:  **rdma-core/52.0**

#### Motivation
#18979 requires `build_librdmacm=True`. Enabled the remaining `build_libmlx5` as well, since it does not add any dependencies or other notable overhead.

#### Details
https://github.com/linux-rdma/rdma-core/releases/tag/v52.0

---

### DIFF
--- a/recipes/rdma-core/all/conanfile.py
+++ b/recipes/rdma-core/all/conanfile.py
@@ -36,8 +36,8 @@ class PackageConan(ConanFile):
         "build_libibnetdisc": True,
         "build_libmana": True,
         "build_libmlx4": True,
-        "build_libmlx5": False,
-        "build_librdmacm": False,
+        "build_libmlx5": True,
+        "build_librdmacm": True,
     }
 
     def configure(self):
--- a/recipes/rdma-core/all/conanfile.py
+++ b/recipes/rdma-core/all/conanfile.py
@@ -49,8 +49,7 @@ class PackageConan(ConanFile):
 
     def requirements(self):
         self.requires("libnl/3.8.0")
-        if self.settings.os in ["Linux", "FreeBSD"]:
-            self.requires("libudev/system")
+        self.requires("libudev/system")
 
     def validate(self):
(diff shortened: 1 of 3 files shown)


valgur:
@perseoGI

perseoGI:
There is a compilation issue with `librdmacm` in conan 1 pipeline. Could you take a look?
I'm agree to enable the remaining options by default, but, couldn't that be done from `libfabrick`?

valgur:
> couldn't that be done from libfabrick?

Not on CCI itself, since overriding an option would require the dependency to be rebuilt and the CI does not use `--build-missing`.

perseoGI (approved):
LGTM now, thanks! And thanks for the clarification

recipes/rdma-core/all/conanfile.py
```diff
@@ -73,6 +72,10 @@ def generate(self):
         # tc.variables["ENABLE_STATIC"] = not self.options.shared
         tc.variables["NO_PYVERBS"] = True
         tc.variables["NO_MAN_PAGES"] = True
+        # Otherwise get set to ${install_prefix}/car/run and the paths in
+        # https://github.com/linux-rdma/rdma-core/blob/v52.0/buildlib/config.h.in
+        # can exceed the 108-character limit for socket paths on Linux
+        tc.variables["CMAKE_INSTALL_RUNDIR"] = "/var/run"
```
uilianries, line 78:
```suggestion
        if "CMAKE_INSTALL_RUNDIR" not in self.conf.get("tools.cmake.cmaketoolchain:extra_variables", check_type=dict, default={}):
            tc.variables["CMAKE_INSTALL_RUNDIR"] = self."/var/run"
```

I would keep it more flexible as we have issues when declaring hardcoded system path in a recipe.

valgur: Thanks! That's a much better solution.

PR #15995, opened 2023-02-13
[outcome/2.2.4] Add multi-file outcome library version

calebkiage:
Add v2.2.4

Specify library name and version:  **outcome/2.2.4**

Add multi-file header version of outcome lib. Needed by upcoming llfio recipe

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24727, opened 2024-07-26
[outcome] Add option single_header

jeremy-coulon:
### Summary
Changes to recipe:  **outcome/***

#### Motivation
Add `single_header` option. Default is `True` for backward compatibility.

#### Details
I need `single_header=False` for my own recipe of [LLFIO](https://github.com/ned14/llfio)
I found a duplicate PR which is closed here https://github.com/conan-io/conan-center-index/pull/15995
My solution seems much simpler.

---

### DIFF
--- a/recipes/outcome/all/conanfile.py
+++ b/recipes/outcome/all/conanfile.py
@@ -20,6 +20,8 @@ class OutcomeConan(ConanFile):
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
     no_copy_source = True
+    options = { "single_header": [True, False] }
+    default_options = { "single_header": True }
 
     @property
     def _min_cppstd(self):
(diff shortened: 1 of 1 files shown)


jeremy-coulon:
`outcome` also depends on a recent version of `status-code` when `single_header=False`

Waiting for this PR to be merged: https://github.com/conan-io/conan-center-index/pull/24728

uilianries:
@jeremy-coulon Hello! Thank you for your PR, and sorry for the delay in reviewing it.

Some comments about your PR:

The upstream indeed documents about using single-header or multiple headers for Outcome:
https://github.com/ned14/outcome#usage-as-a-single-header-file

When consuming external status-code from a package manager, a definition should be exported as well: https://github.com/ned14/outcome/blob/develop/CMakeLists.txt#L172C34-L172C64. It should be added to `package_info`.

The non-packaged project llfio is consuming outcome as pointed: https://github.com/ned14/llfio/blob/06ea607de0346a1d77a807fd37efc2fdfb005de5/programs/CMakeLists.txt#L15C26-L15C33 And requires non single-header variant: https://github.com/ned14/llfio/blob/06ea607de0346a1d77a807fd37efc2fdfb005de5/include/llfio/v2.0/multiplex.hpp#L30

This PR makes sense in terms of supporting llfio, but it will not work when adding llfio in ConanCenterIndex, because the package Outcome with the single_header=False will not exist in Conan Center, resulting in the missing headers.

PR #24736, opened 2024-07-27
gpu.cpp: new recipe

valgur:
### Summary
Changes to recipe:  **gpu.cpp/cci.20240726**

#### Motivation
gpu.cpp is a lightweight library that makes portable GPU compute with C++ simple.

It focuses on general purpose native GPU computation, leveraging the WebGPU specification as a portable low-level GPU interface. This means we can drop in GPU code in C++ projects and have it run on Nvidia, Intel, AMD, and other GPUs. The same C++ code can work on a wide variety of laptops, workstations, mobile devices or virtually any hardware with Vulkan, Metal, or DirectX support.

https://gpucpp.answer.ai/

#### Details
The library is not yet very stable in terms of the build system, it seems. It currently does not install any headers or export a CMake config. I opted to place the installed headers under an unofficial `gpu.cpp/` prefix and chose to use the `gpu.cpp` CMake file name and `gpu.cpp::gpu.cpp` CMake target implicitly generated by Conan.

I'll probably have to simplify `test_package.cpp`, but I'd like to see whether the more complex one passes for all CCI configurations first.

- Requires #24735

/cc @AnswerDotAI @austinvhuang

---

### DIFF
--- a/recipes/gpu.cpp/all/conandata.yml
+++ b/recipes/gpu.cpp/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "cci.20240726":
+    url: "https://github.com/AnswerDotAI/gpu.cpp/archive/0a2d73474a5fa858fbe2021c33a82bd21bfcef38.tar.gz"
+    sha256: "07dee5c15c9612c627ca03c3a46ff5d92323d646ab54fbd2aaf9e58fcd4b54b2"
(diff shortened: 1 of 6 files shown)


austinvhuang:
@valgur nice work, would love to see a dawn recipe available. FYI over the weekend, we've pushed some changes to update the dawn commit we're pointing to and simplify our build dependencies to use dawn directly.

PR #24748, opened 2024-07-28
librdkafka: add version 2.5.0

toge:
### Summary
Changes to recipe:  **librdkafka/2.5.0**

#### Motivation
There are several new features in 2.5.0.

#### Details
https://github.com/confluentinc/librdkafka/compare/v2.4.0...v2.5.0#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a

---

### DIFF
--- a/recipes/librdkafka/all/conanfile.py
+++ b/recipes/librdkafka/all/conanfile.py
@@ -4,6 +4,7 @@ from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
 from conan.tools.gnu import PkgConfigDeps
 from conan.tools.scm import Version
+from conan.tools.microsoft import is_msvc
 import os
 
 required_conan_version = ">=1.55.0"
@@ -54,6 +55,8 @@ class LibrdkafkaConan(ConanFile):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+        if is_msvc(self):
+            del self.options.syslog
 
     def configure(self):
         if self.options.shared:
--- a/recipes/librdkafka/all/conanfile.py
+++ b/recipes/librdkafka/all/conanfile.py
@@ -101,6 +101,7 @@ class LibrdkafkaConan(ConanFile):
         tc.variables["WITH_SASL"] = self.options.sasl
         tc.variables["ENABLE_LZ4_EXT"] = True
         tc.variables["WITH_CURL"] = self.options.curl
+        tc.variables["WITH_SNAPPY"] = True
         tc.preprocessor_definitions["WITH_SYSLOG"] = "1" if self.options.syslog else "0"
         tc.generate()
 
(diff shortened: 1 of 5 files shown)


dotfield (commented):
Under options (around lines 24-31) there is no option for "snappy". This has been a problem in 2.3.0 and 2.4.0 too. It should be added with the others and have WITH_SNAPPY set to the value.

In reality you should also make SYSLOG configurarable too with these options although we do not use this option. Still, it is an option of the librdkafka build and should be configurable.

toge:
@dotfield 
Thank you for your comment.
I try to fix your point.

toge:
@dotfield 
I found `WITH_SNAPPY` is alwasy set 1 by `config.h` and snappy.h and snappy.c are included upstream repository.

https://github.com/confluentinc/librdkafka/blob/v2.5.0/packaging/cmake/config.h.in#L35

Do you want to add `snappy` option to disable snappy?

dotfield:
> @dotfield I found `WITH_SNAPPY` is alwasy set 1 by `config.h` and snappy.h and snappy.c are included upstream repository.
> 
> https://github.com/confluentinc/librdkafka/blob/v2.5.0/packaging/cmake/config.h.in#L35
> 
> Do you want to add `snappy` option to disable snappy?

It is probably better to set the flag by default to True. That way it should not "break" anyone who has been using it before. Although you have put WITH_SNAPPY=1 in the config.h file it is still set to 0 in the cmake options created which leads to a build warning. It should be consistent, and although it appears that config.h is picked up by rd.h and therefore everything before it hits code checks like in rdkafka_msgset_reader.c, it is still better this is set as an option and is consistent.

The alternative would be to set it to True in conanfile.py (without a configurable option) so it gets picked up by cmake too.

toge:
@dotfield 
Thanks for your advice.
I will add a definition to conanfile.py that sets WITH_SNAPPY to True.

Since it would require modification of config.h to disable the use of SNAPPY depending on the option, I think that it is appropriate to modify on upstream.

perseoGI (approved):
LGTM

uilianries (commented):
Thank you for your PR! Some important points related to this version.

- Google snappy is vendorized in the project, it's possible to decouple, but it would not worth now. The same is available since version 1.9.2, which will not break users using that version: https://github.com/confluentinc/librdkafka/blob/v1.9.2/src/snappy.c

- Would suggest removing syslog option for msvc in config_options. The syslog is available in unix system, but not in Windows. It may work with msys, but probably will fail with visual studio.

toge:
@uilianries 
Thank you for your advice.
I disabled syslog options on msvc.

recipes/librdkafka/all/conanfile.py
```diff
@@ -98,6 +104,8 @@ def generate(self):
         tc.variables["WITH_SASL"] = self.options.sasl
         tc.variables["ENABLE_LZ4_EXT"] = True
         tc.variables["WITH_CURL"] = self.options.curl
+        tc.variables["WITH_SNAPPY"] = True
+        tc.preprocessor_definitions["WITH_SYSLOG"] = "1" if self.options.syslog else "0"
```
uilianries, line 108:
```suggestion
        tc.preprocessor_definitions["WITH_SYSLOG"] = "1" if self.options.get_safe("syslog") else "0"
```

uilianries (approved):
LGTM. I tested with syslog active on Linux, and built without errors. Plus, it does not forward an ifdef WITH_SYSLOG in public headers. Full build log:

[librdkafka-2.5.0-syslog-linux.log](https://github.com/user-attachments/files/16466278/librdkafka-2.5.0-syslog-linux.log)

Nice job!

AbrilRBS (approved):
Thankss!

perseoGI (approved):

Issue #24729, opened 2024-07-26
[package] qt/6.7.1: cannot build without system installed libclang


weatherhead99:
### Description

I can no longer build the qt recipe without installing libclang on my system. The submodule "qttools" depends on libclang, 
but passing `-oqt/*:qttools=False` does not prevent building of the qttools module, because it is marked as "essential" (https://github.com/conan-io/conan-center-index/issues/24481). 

I understand there is not a conan recipe that contains libclang, fair enough, but there should be some way to build this without an installed libclang. I do not believe qttools actually is "essential" especially because one could also disable qtdoc and qttranslations which depend on it.

Furthermore, if a system libclang is going to be required, it should be actually in the dependencies of the conan file (i.e. via a "libclang/system" package or similar), and not implicitly assumed.

### Package and Environment Details

* Package Name/Version: **qt/6.7.1**
* Operating System+version: **Debian 12**
* Compiler+version: **GCC 12.2**
* Conan version: **conan 2.5.0**
* Python version: **Python 3.10.12 (anaconda)**

### Conan profile

Host profile:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
[options]
*/*:shared=True
qt/*:with_x11=False
[conf]
tools.gnu:pkg_config=/usr/bin/pkg-config

Build profile:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
[options]
*/*:shared=True
qt/*:with_x11=False
[conf]
tools.gnu:pkg_config=/usr/bin/pkg-config

### Steps to reproduce

conan install qt/6.7./1@ --build=missing -oqt/*:qttools=False -oqt/*:qttranslations=False -oqt/*:qtdoc=False

### Logs

======== Computing dependency graph ========
ERROR: qt/6.7.1: Error in configure() method, line 230
	_enablemodule(module_name)
while calling '_enablemodule', line 220
	_enablemodule(req)
while calling '_enablemodule', line 218
	setattr(self.options, mod, True)
	ConanException: Incorrect attempt to modify option 'qttools' from 'False' to 'True'

planetmarshall:
#22997 tangentially related as it makes a libclang recipe possible,however that PR has been green for months and I'm becoming a bit disillusioned with making further contributions.

[...]

PR #24753, opened 2024-07-29
qt6: dont use system libclang

jcar87:
### Summary
Changes to recipe:  qt

#### Motivation
The build scripts look for libclang in the system "optionally" - this can cause errors in some systems where the development files for libclang are not fully present. Otherwise it would succeeds if it finds it, or if it doesn't at all.

See https://github.com/conan-io/conan-center-index/issues/24729

#### Details
* Make the recipe for reproducible and do not rely on whether or not libclang is present in the system
* This essentially disables building the QDoc utility, but it would seem that tool is more useful for Qt developers to generate the Qt documentation itself. We will consider other options if it turns out this tool is widely used.

Additionally:
* on a trial basis, use version ranges for libjpeg turbo and sqlite3 - which are known to be stable libraries with a backwards compatible C interface


uilianries (approved):

perseoGI (approved):
LGTM!

PR #24759, opened 2024-07-30
libtorch: new recipe

### LINKED ISSUE #6861: [request] pytorch/1.9
### Package Details
  * Package Name/Version: **pytorch/1.9**
  * Website: **https://pytorch.org/**
  * Source code: **https://github.com/pytorch/pytorch**

  * Tensor computation (like NumPy) with strong GPU acceleration
  * Deep neural networks built on a tape-based autograd system

valgur:
### Summary
Changes to recipe:  **libtorch/2.4.0**

#### Motivation
Tensors and Dynamic neural networks in Python with strong GPU acceleration.

https://github.com/pytorch/pytorch

[Packaging status](https://repology.org/project/libtorch/versions)

#### Details
Continues from #5100 by @SpaceIm.

CUDA, HIP and SYCL backends are currently disabled since the PR is complex enough already and these can be addressed in a follow-up PR. Vulkan and Metal (TODO) should be usable as GPU backends currently.

Distributed feature is disabled as well to limit the scope and due to `openmpi` not yet being available (#18980).

Android and iOS builds are probably broken and need testing.

Non-OpenBLAS BLAS backends are probably not usable due to OpenBLAS being required for LAPACK. A separate LAPACK recipe would be required to fix that (such as #23798).

Closes #6861.

TODO:
- [x] Export missing CMake variables.
- [ ] Test with Metal on macOS.
- [ ] Submit bugfix patches upstream.
- [ ] Create a recipe for `pocketfft` and unvendor.

---


hasB4K:
Hello @valgur, thanks for this amazing PR. Do you plan to continue working on it? 🤞Having libtorch in Conan would be so neat. Since OpenMPI is now available, do you plan to let the user to enable the distributed feature?

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+        tc.variables["USE_CCACHE"] = False
+        tc.variables["DEBUG_CUDA"] = self.options.get_safe("debug_cuda", False)
+        tc.variables["USE_COREML_DELEGATE"] = self.options.coreml_delegate
+        tc.variables["BUILD_LAZY_TS_BACKEND"] = self.options.build_lazy_ts_backend
+        tc.variables["USE_MIMALLOC"] = self.options.with_mimalloc
+
+        tc.variables["BLAS"] = self._blas_cmake_option_value
+
+        tc.variables["MSVC_Z7_OVERRIDE"] = False
+
```
keef-cognitiv, line 436: Incidentally, this also needs 
```
tc.variables["CMAKE_CXX_EXTENSIONS"] = True
```

Tested this while running a build that uses a compiler.cppstd. If it is using a non-gnu standard (which for other packages it must be) ATen breaks with the same error as: https://github.com/pytorch/QNNPACK/issues/67

This converts cppstd=c++17 for example to cppstd=gnu++17

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+                lib_folder = os.path.join(self.package_folder, "lib")
+                if is_apple_os(self):
+                    lib_fullpath = os.path.join(lib_folder, f"lib{libname}.a")
+                    whole_archive = f"-Wl,-force_load,{lib_fullpath}"
+                elif is_msvc(self):
+                    lib_fullpath = os.path.join(lib_folder, libname)
+                    whole_archive = f"-WHOLEARCHIVE:{lib_fullpath}"
+                else:
+                    lib_fullpath = os.path.join(lib_folder, f"lib{libname}.a")
+                    whole_archive = f"-Wl,--whole-archive,{lib_fullpath},--no-whole-archive"
```
lia-viam, line 529: Thanks for your work on this PR--I am not using this library directly but found it through following some github issues on whole archive linking.

For this line--I wonder if we would want to do this with `-Wl,--push-state,--pop-state`? See eg
https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_LINK_LIBRARY_USING_FEATURE.html#loading-a-whole-static-library

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+        if self._is_mobile_os:
+            self.options.blas = "eigen"
+            self.options.build_lazy_ts_backend = False
+            del self.options.with_fbgemm
+            # del self.options.distributed
+        if self.settings.arch not in ["x86", "x86_64", "armv8"]:
+            self.options.rm_safe("with_mkldnn")
+        if not is_apple_os(self) or self.settings.os not in ["Linux", "Android"]:
+            del self.options.with_nnpack
+        self.options.with_itt = self.settings.arch in ["x86", "x86_64"]
```
joda01, line 199: This line overrides the manual taken settings in `with_itt`. So even if set to `False` this line sets it to `True` on x86, x86_64 which may not be the expected behaviour.

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+            "openblas": "OpenBLAS",
+            "mkl": "MKL",
+            "veclib": "vecLib",
+            "flame": "FLAME",
+            "generic": "Generic"
+        }[str(self.options.blas)]
+
+    @property
+    def _use_nnpack_family(self):
+        return any(self.options.get_safe(f"with_{name}") for name in ["nnpack", "qnnpack", "xnnpack"])
```
joda01, line 253: `with_xnnpack` can not be deleted because it is unsafe used on line 284 `if self.options.with_xnnpack:`

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+        if self.options.with_fbgemm:
+            self.requires("fbgemm/0.8.0", transitive_headers=True, transitive_libs=True)
+        if self.options.with_gflags:
+            self.requires("gflags/2.2.2")
+        if self.options.with_glog:
+            self.requires("glog/0.7.1", transitive_headers=True, transitive_libs=True)
+        if self.options.get_safe("with_qnnpack"):
+            self.requires("fxdiv/cci.20200417")
+            self.requires("psimd/cci.20200517")
+        if self.options.with_xnnpack:
```
joda01, line 284:
```suggestion
        if self.options.get_safe("with_xnnpack", False):
```

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+            self.requires("libnuma/2.0.16")
+        if self.options.with_opencl:
+            self.requires("opencl-headers/2023.12.14")
+            self.requires("opencl-icd-loader/2023.12.14")
+        if self.options.with_vulkan:
+            self.requires("vulkan-headers/1.3.268.0")
+            self.requires("vulkan-loader/1.3.268.0")
+        if self.options.with_mimalloc:
+            self.requires("mimalloc/2.1.7")
+
```
joda01, line 300: Build on MacOS needs pybind
```suggestion
        if is_apple_os(self):
            self.requires("pybind11/2.13.6")
```

recipes/libtorch/all/conanfile.py
```diff
@@ -0,0 +1,667 @@
...
+        self.tool_requires("cpython/[~3.12]")
+        if self._depends_on_flatbuffers:
+            self.tool_requires("flatbuffers/<host_version>")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+        # Keep only a restricted set of vendored dependencies.
+        # Do it before build() to limit the amount of files to copy.
+        allowed = ["pocketfft", "kineto", "miniz-2.1.0"]
```
joda01, line 364: For MacOs build some more 3rd party libs are neeeded
```suggestion
        allowed = ["pocketfft", "kineto", "miniz-2.1.0"]
        if self.is_mac_os == True:
            allowed = ["pocketfft", "kineto", "miniz-2.1.0", "opentelemetry-cpp", "protobuf"]
```

recipes/libtorch/all/conanfile.py
valgur: Thanks for the comment, but I don't see `with_xnnpack` option being deleted anywhere? This specific line is querying the value, not removing the option.

recipes/libtorch/all/conanfile.py
valgur: Are you sure it's required specifically on macOS? I might have missed it on Linux due to having it available on my system.

recipes/libtorch/all/conanfile.py
valgur: Hmm... I would prefer to unvendor these and use Conan versions. I'll have to do more testing on macOS.

recipes/libtorch/all/conanfile.py
joda01: Sorry for the missing context. with_xnnpack must be deleted for MacOS build since this is not yet supported under Mac. For that cases it might be  problem.
I managed to build libtorch under linux and MacOS ARM with some modifications based on @valgur's work.
If you are interested you can see the actual verison:
https://github.com/joda01/imagec-recipes/actions/runs/12983296316

recipes/libtorch/all/conanfile.py
joda01: I tried it without but had no success. You can have a look to 
https://github.com/joda01/imagec-recipes/actions/runs/12983296316

recipes/libtorch/all/conanfile.py
joda01: I total agree with you! It's a fast workaround.

joda01:
@valgur Thank's a lot for your initial work providing a libtorch recipe!
Do you have some experience with Windows build. In my pipeline I get Python not found form cmake even Pyhton is installed an in path!?

recipes/libtorch/all/conanfile.py
valgur: Thanks! I'll definitely take a look.

PR #24768, opened 2024-07-30
qt5: fix check for python2 version

jcar87:
### Summary
Changes to recipe:  qt/5.x.x

#### Motivation
* building qtwebengine requires python2 (still), and the current check in the `validate_build()` method does not work since it checks for stdout

#### Details
* check the python version by passing `-c "import platform;print(platform.python_version())"` instead, which is correctly printed to stdout - avoids other alternative solutions like pipe redirection or calling the underlying process popen


jcar87:
tested locally with
```
conan create . --version=5.15.14 -o "*:shared=True" -o "qt/*:qtwebengine=True" -o "qt/*:gui=True" -o "qt/*:qtlocation=True" -o "qt/*:qtwebchannel=True"
```

on Windows

AbrilRBS (approved):

uilianries (approved):
Nice, much safer than using pipes.

PR #24769, opened 2024-07-30
pocketfft: new recipe

valgur:
### Summary
Changes to recipe:  **pocketfft/cci.20240710**

#### Motivation
Adds PocketFFT: a header-only heavily modified implementation of FFTPack.

https://github.com/mreineck/pocketfft

A dependency of libtorch (#24759).

[Packaging status](https://repology.org/project/pocketfft/versions)

#### Details
The library is only configurable through preprocessor defines. I exposed all available options with their default values kept intact.

---


recipes/pocketfft/all/conandata.yml
```diff
@@ -0,0 +1,9 @@
+sources:
+  "cci.20240710":
```
AbrilRBS, line 2: Same thing about the new version scheme for new recipes. If they have not released any version, we can go for `0.0.0.cci.20240710`

AbrilRBS (commented):
Thanks! Same thing about the version scheme, otherwise looks good

recipes/pocketfft/all/conandata.yml
valgur: Ok, sounds good.
Just to clarify - the version prefix should be the latest officially released version in general, right?

AbrilRBS: Yes, so that if a new version is ever released, there's little chance of ours being considered newer than the released tag :) (And 0.0.0 for libraries without versions!)

AbrilRBS (approved):
Thanks!

danimtb (approved):

Issue #24732, opened 2024-07-27
[request] subunit/1.4.4


vsbogd:
### Package Name/Version

subunit/1.4.4

### Changelog

https://github.com/testing-cabal/subunit/releases/tag/1.4.4

### Context about the new update

The above-mentioned version is newly released by the upstream project and not yet available as a recipe.
Please add this version.

vsbogd:
I already prepared change https://github.com/vsbogd/conan-center-index/tree/package/subunit, waiting for access to raise the PR.

PR #24773, opened 2024-07-31
subunit: add version 1.4.4

vsbogd:
### Summary
Changes to recipe:  **lib/1.4.4**. Fixes: #24732

#### Motivation
Version 1.4.0 which is the current subunit version in Conan repo is not compatible with Python 3.12 (see [subunit bug #2052948](https://bugs.launchpad.net/subunit/+bug/2052948)). Version 1.4.4 has compatibility issues fixed.

#### Details
Changes in PR:
- add new version and add autoreconf() step because new version requires it https://github.com/conan-io/conan-center-index/commit/f0329525508be55f0f838ffe3ea7e1523cab1dc5
- replace `ConanFile.env_info` (which is deprecated) by `ConanFile.runenv_info` https://github.com/conan-io/conan-center-index/commit/37a10d45757b568216fa801923e55ece9e55a9be
Changed are tested with both 1.4.0 and 1.4.4 version using both Conan 1 and 2.

---

### DIFF
--- a/recipes/subunit/all/conanfile.py
+++ b/recipes/subunit/all/conanfile.py
@@ -137,6 +137,7 @@ class SubunitConan(ConanFile):
         apply_conandata_patches(self)
         with chdir(self, self.source_folder):
             autotools = Autotools(self)
+            autotools.autoreconf()
             autotools.configure()
             autotools.make()
 
--- a/recipes/subunit/all/conanfile.py
+++ b/recipes/subunit/all/conanfile.py
@@ -165,4 +165,4 @@ class SubunitConan(ConanFile):
         self.cpp_info.components["libcppunit_subunit"].set_property("pkg_config_name", "libcppunit_subunit")
 
         bin_path = os.path.join(self.package_folder, "bin")
-        self.env_info.PATH.append(bin_path)
+        self.runenv_info.append_path("PATH", bin_path)
--- a/recipes/subunit/all/conandata.yml
+++ b/recipes/subunit/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.4.4":
+    url: "https://github.com/testing-cabal/subunit/archive/refs/tags/1.4.4.tar.gz"
+    sha256: "395787f52b8b36830c331faca546ea6c248b6cbd8de95989c2f64d432cc90531"
   "1.4.0":
     url: "https://launchpad.net/subunit/trunk/1.4.0/+download/subunit-1.4.0.tar.gz"
     sha256: "00792096e96cbf1d5f9394dadd0ea73309a4f8b2214822949f1888ce5c2cf1c0"
--- a/recipes/subunit/config.yml
+++ b/recipes/subunit/config.yml
@@ -1,3 +1,5 @@
 versions:
+  "1.4.4":
+    folder: "all"
   "1.4.0":
     folder: "all"


vsbogd:
Not sure why build fails under MacOSX and cannot debug it without access to the files on the build machine. Found some similar issue on StackOverflow https://stackoverflow.com/questions/14465667/errors-when-installing-a-perl-module-using-make-mac-osx-10-7. The answer says that XCode is probably configured incorrectly.

recipes/subunit/all/conanfile.py
```diff
@@ -73,8 +73,15 @@ def build_requirements(self):
...
             if not self.conf.get("tools.microsoft.bash:path", check_type=str):
                 self.tool_requires("msys2/cci.latest")
-        if is_msvc(self):
+        if self.version == "1.4.0":
+            if is_msvc(self):
+                self.tool_requires("automake/1.16.5")
+        else:
+            # version >=1.4.4 is kept on GitHub without configuration built
+            self.tool_requires("libtool/2.4.7")
+            self.tool_requires("pkgconf/2.2.0")
```
valgur, line 82:
```suggestion
```
There is no `PkgConfDeps` generator used in the recipe. Is the `pkgconf` tool dependency really being used?

valgur:
I would make the recipe simpler and always run `autoreconf`, even for the existing version. Not sure if others agree, though.

recipes/subunit/all/conanfile.py
vsbogd: `configure.ac` uses `PKG_CHECK_...` which fails if `pkgconf` is not installed: https://github.com/testing-cabal/subunit/blob/98aa1076f8aed666262022634ddccdf9e345b13f/configure.ac#L64-L66 
I don't know `autoconf` well to understand whether it can be fixed by another way.

vsbogd:
> I would make the recipe simpler and always run autoreconf, even for the existing version. Not sure if others agree, though.

Yeah it was my first intention. I added this condition because code of the version `1.4.0` which is downloaded from `lunchpad` is preconfigured and I saw it built well previously. I guessed that current failure on MacOSX armv8 can be caused by this reconfiguration. But adding condition doesn't fix build failure, so I would revert last commit. But I would prefer doing it after build failure is fixed.

vsbogd:
> But adding condition doesn't fix build failure, so I would revert last commit. 

Hm, it looks like in fact this commit fixes build for Conan 2.0 completely (see difference between https://github.com/conan-io/conan-center-index/pull/24773#issuecomment-2259909799 and https://github.com/conan-io/conan-center-index/pull/24773#issuecomment-2306716928). Thus I think it is better to leave it as is.

ErniGH:
Hi, first of all, I'm really sorry for the delay. I was looking at this new version and it feels strange to me. The source of the tar.gz isn't the same as before — it's in a different repo, but I can't figure out the connection between the previous one and the new repository either. Can you help me?

vsbogd:
Hi, I found the GitHub repo when I was looking for a fix for [Python 3.12 compatibility issue](https://bugs.launchpad.net/subunit/+bug/2052948) I posted on Launchpad. I don't remember but probably I looked at the PyPi [python-subunit page](https://pypi.org/project/python-subunit/) which points to the [Launchpad page](http://launchpad.net/subunit) as home page of the project and to the [GitHub repo](https://github.com/testing-cabal/subunit) as a source code page. Then I have found that version that is compatible with Python 3.12 is published as a [GitHub release of version 1.4.4](https://github.com/testing-cabal/subunit/releases/tag/1.4.4).

Later I have found that `python-subunit` package in Arch Linux repo https://archlinux.org/packages/extra/any/python-subunit/ is also built from the GitHub repo release, see [package build recipe](https://gitlab.archlinux.org/archlinux/packaging/packages/python-subunit/-/blob/main/PKGBUILD?ref_type=heads)

Most people from [Launchpad team list](https://launchpad.net/~subunit) are also among main [GitHub repo contributors](https://github.com/testing-cabal/subunit/graphs/contributors). Thus I think that team decided move the source code to the GitHub but unfortunately I don't see any mention of this decision on either Launchpad or GitHub page.

vsbogd:
Actually [Launchpad project page](https://code.launchpad.net/subunit) also states that code is hosted on GitHub: 
image

ErniGH (approved):
LGTM it is easier to keep just the last version 👍 

Thank you all

vsbogd:
Are there any other blockers for the merge?

PR #24800, opened 2024-08-02
add new recipe for libMeshb

EstebanDugueperoux2:
### Summary

#### Details

---


recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
```
valgur, line 3: Remove the unused imports, please.

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if is_msvc(self) and self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
```
valgur, line 51: A comment on the reason why would be useful here. I'm assuming the symbols are not exported?

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+    def package(self):
+        copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.la", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "lib"))
+        rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
```
valgur, line 83:
```suggestion
        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
```

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def package_info(self):
+        self.cpp_info.libs = ["Meshb.7"]
```
valgur, line 86:
```suggestion
        self.cpp_info.set_property("cmake_file_name", "libMeshb")
        self.cpp_info.set_property("cmake_target_name", "libMeshb::Meshb.7")

        self.cpp_info.libs = ["Meshb.7"]
```
See https://github.com/LoicMarechal/libMeshb/blob/v7.80/sources/CMakeLists.txt

This also ensures that matching CMake variables are created for the ones exported by the project, like `libMeshb_INCLUDE_DIRS`.

recipes/libmeshb/all/test_package/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.15)
+
+project(test_package LANGUAGES C)
+
+find_package(libMeshb REQUIRED CONFIG)
+
+add_executable(${PROJECT_NAME} test_package.c)
+target_link_libraries(${PROJECT_NAME} PRIVATE libmeshb::libmeshb)
```
valgur, line 8:
```suggestion
target_link_libraries(${PROJECT_NAME} PRIVATE libMeshb::Meshb.7)
```

recipes/libmeshb/all/test_package/conanfile.py
```diff
@@ -0,0 +1,26 @@
+from conan import ConanFile
+from conan.tools.build import can_run
+from conan.tools.cmake import cmake_layout, CMake
+from conan.tools.files import copy
```
valgur, line 4:
```suggestion
```

recipes/libmeshb/all/test_package/test_package.c
```diff
@@ -0,0 +1,29 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include "libmeshb7.h"
```
valgur, line 3:
```suggestion
#include <libmeshb7.h>

#include <stdio.h>
#include <stdlib.h>
```
Not critical, but it's useful to list the project include first to test that no standard headers are missing from the project includes.

recipes/libmeshb/all/conandata.yml
```diff
@@ -0,0 +1,5 @@
+sources:
+  "7.8.0":
+    url:
+    - "https://github.com/LoicMarechal/libMeshb/archive/refs/tags/v7.80.tar.gz"
```
valgur, line 4:
```suggestion
    url: "https://github.com/LoicMarechal/libMeshb/archive/refs/tags/v7.80.tar.gz"
```

valgur (changes requested):
The correct CMake file and target names are missing.

recipes/libmeshb/all/conanfile.py
EstebanDugueperoux2: Hi @valgur,

I don't know very well C++ specific on Windows.
If I understand correctly to do shared lib on Windows code must include export macro (https://learn.microsoft.com/en-us/cpp/build/reference/export-exports-a-function?view=msvc-170)?
When looking at https://github.com/LoicMarechal/libMeshb/blob/master/sources/libmeshb7.h I don't see such export macro.
It seems that this can be managed by CMake (https://www.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/).

I have removed this valide method but now CI fails on Windows with shared option at true.

I'll try "tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True" in generate() method to fix that.

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rm, rmdir
```
valgur, line 4:
```suggestion
from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, replace_in_file, rmdir
```

recipes/libmeshb/all/test_package/test_package.c
valgur, line 1: Can you modify the example code to not rely on a binary file? There has been a push to avoid them and keep the test_package snippets simpler as well. I'm assuming it's mostly to avoid any security issues akin to the xz_utils one where test binaries were used to smuggle in a backdoor.

Since it's C, you can probably simply pass an invalid path and ignore the return code.

valgur (changes requested):
LGTM, but perhaps the test_package can be made simpler.

recipes/libmeshb/all/test_package/test_package.c
EstebanDugueperoux2: I don't know how to do that.
Would it be possible from test_package's conanfile.py reference source_folder of parent conanfile.py because sources of lbMeshb contains some samples files that I could pass in input of this test_package?

valgur: Try
```cpp
#include <libmeshb7.h>

#include <stdio.h>

int main() {
   int64_t OutMsh;
   if(!(OutMsh = GmfOpenMesh("test.meshb", GmfWrite, 2, 3)))
      return 1;
   GmfCloseMesh(OutMsh);
   puts("Success! Created test.meshb");
   return 0;
}
```

valgur (approved):
Thanks!

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    def export_sources(self):
+        export_conandata_patches(self)
```
AbrilRBS, line 29: No patches, it can be removed

AbrilRBS (commented):
Minor thing after a first look, othwewise looks good :)

recipes/libmeshb/all/conanfile.py
uilianries: > I'll try "tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True" in generate() method to fix that.

Please, don't. We try to avoid adding feature not supported by the upstream. Plus, `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS` is fragile, it exposes all symbols, including private and is not capable to export static variables, causing linkage errors. In the upstream does not support shared on Windows, neither do we.

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
```
uilianries, line 35:
```suggestion
        self.settings.rm_safe("compiler.libcxx")
        if is_msvc(self):
            self.package_type = "static-library"
            del self.options.shared
```

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def validate(self):
+        # Windows shared build doesn't seems supported because code doesn't include dllexport
+        # See https://www.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/
+        if is_msvc(self) and self.options.shared:
+            raise ConanInvalidConfiguration(f"{self.ref} can not be built as shared on Visual Studio and msvc.")
```
uilianries, line 51:
```suggestion
```

uilianries (commented):
Removing shared for msvc give the opportunity of consuming this package with other shared libraries with Conan 2.x, otherwise, it will always raise an invalid configuration:

    conan install  ... -o "*/*:shared=True"

recipes/libmeshb/all/test_package/test_package.c
```diff
@@ -0,0 +1,12 @@
+#include <libmeshb7.h>
+
+#include <stdio.h>
+
+int main() {
+   int64_t OutMsh;
+   if(!(OutMsh = GmfOpenMesh("test.meshb", GmfWrite, 2, 3)))
+      return 1;
+   GmfCloseMesh(OutMsh);
+   puts("Success! Created test.meshb");
```
uilianries, line 10:
```suggestion
   GmfCloseMesh(NULL);
```

I would prefer not touching the disk if possible. The test package only needs to validate the library, do not need to be functional.

jcar87:
Just relaunched the jobs

recipes/libmeshb/all/test_package/test_package.c
EstebanDugueperoux2: Hi @uilianries doing that make the test_package failed with error code 139.
I don't know how do other IO library.
Perhaps we can skip this test_package validation?

recipes/libmeshb/all/test_package/test_package.c
```diff
@@ -0,0 +1,8 @@
+#include <libmeshb7.h>
+
+#include <stdio.h>
+
+int main() {
+   GmfCloseMesh(NULL);
+   return 0;
+}
```
uilianries, line 8:
```suggestion
#include <libmeshb7.h>

int main() {
   int ver, dim;
   int64_t LibIdx = GmfOpenMesh("/dev/null", GmfRead, &ver, &dim);
   return 0;
}

```
Let's try something that does not raise a signal. I tested locally, in case not be able to read the file (permission or does not exist), it will return -1. As I commented before, we just want to validate the package (header include and library linkage), the test package does not need be functional.

valgur: Reading `/dev/null` probably won't work on Windows, though. Might as well try `""` probably.

uilianries: No problem at all. Like I commented above, missing file will result in failure as well. This current example fails in Linux already, we just do not check anything. :smiling_face_with_tear:

uilianries: Also tested using a non-existing file in Linux, like `/foo/bar` and resulted in the same.

recipes/libmeshb/all/test_package/test_package.c
EstebanDugueperoux2: Thanks it works.

recipes/libmeshb/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_gmf_asio": False
+    }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
```
uilianries, line 30:
```suggestion
            del self.options.fPIC
            del self.options.with_gmf_asio
```

The option `gmf_asio` will require `aio.h` which is available only for Unix systems.

https://github.com/LoicMarechal/libMeshb/blob/v7.80/sources/libmeshb7.c#L114

uilianries (commented):
Only a final detail, otherwise, looking good :smile:

uilianries (approved):
LGTM. Thank you for your PR!

valgur (approved):

AbrilRBS (approved):

PR #24807, opened 2024-08-03
Add release 1.4.4 of AMGCL

EstebanDugueperoux2:
### Summary

#### Details

---

### DIFF
--- a/recipes/amgcl/all/conandata.yml
+++ b/recipes/amgcl/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.4.4":
+    url: "https://github.com/ddemidov/amgcl/archive/1.4.4.tar.gz"
+    sha256: "02fd5418e14d669422f65fc739ce72bf9516ced2d8942574d4b8caa05dda9d8c"
   "1.4.3":
     url: "https://github.com/ddemidov/amgcl/archive/1.4.3.tar.gz"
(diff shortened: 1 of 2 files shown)


perseoGI (commented):
Added new option `with_boost` as this library now works with out it.
Even though, default option will be set to True to avoid issues if users decide to bump their dependencies as AMGCL without boost is more limited.

https://github.com/ddemidov/amgcl/compare/1.4.3...1.4.4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR51

recipes/amgcl/all/conanfile.py
```diff
@@ -20,12 +21,19 @@ class AmgclConan(ConanFile):
...
+        if Version(self.version) < "1.4.4":
+            del self.options.with_boost
 
     def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.81.0")
+        if self.options.get_safe("with_boost"):
+            self.requires("boost/1.81.0")
```
AbrilRBS, line 36:
```suggestion
            self.requires("boost/1.85.0")
```

Less conflicts. As you suggested over on slack, a range would be a solution if it wasn't boost, it's too big to confidentely add ranges to it unconditionally. Maybe for header-only libraries it would be a valid approach?

AbrilRBS (approved):

perseoGI (approved):

PR #24809, opened 2024-08-03
libjxl: disable linking static libgcc and libstdc++

toge:
### Summary
Changes to recipe:  **libjxl/***

#### Motivation
In some environments, there are compilation errors in static build.
It seems to be caused by linking static libgcc and libcstdc++.
This PR tries to fix it.

#### Details
Compilation error message are following:
```
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - no
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
CMake Error at cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Threads (missing: Threads_FOUND)
  cmake-3.28/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  cmake-3.28/Modules/FindThreads.cmake:226 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:170 (find_package)
```

After applying this PR.
```
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Using -lpthread as --whole-archive
```

---

### DIFF
--- a/recipes/libjxl/all/conandata.yml
+++ b/recipes/libjxl/all/conandata.yml
@@ -8,6 +8,3 @@ sources:
   "0.8.2":
     url: "https://github.com/libjxl/libjxl/archive/v0.8.2.zip"
     sha256: "1f2ccc06f07c4f6cf4aa6c7763ba0598f12a7544d597f02beb07f615eb08ccf0"
-  "0.6.1":
-    url: "https://github.com/libjxl/libjxl/archive/v0.6.1.zip"
-    sha256: "3e4877daef07724aa6f490bf80c45ada804f35fe3cce59c27e89c5ae3099535a"
--- a/recipes/libjxl/all/conanfile.py
+++ b/recipes/libjxl/all/conanfile.py
@@ -82,10 +82,10 @@ class LibjxlConan(ConanFile):
             # Fails with a missing DLL error in test_package
             raise ConanInvalidConfiguration(f"{self.ref} does not support shared builds with MSVC")
 
-    # def build_requirements(self):
-    #     # Require newer CMake, which allows INCLUDE_DIRECTORIES to be set on INTERFACE targets
-    #     # Also, v0.9+ require CMake 3.16
-    #     self.tool_requires("cmake/[>=3.19 <4]")
+    def build_requirements(self):
+        # Require newer CMake, which allows INCLUDE_DIRECTORIES to be set on INTERFACE targets
+        # Also, v0.9+ require CMake 3.16
+        self.tool_requires("cmake/[>=3.19 <4]")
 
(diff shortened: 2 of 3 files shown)


valgur (approved):
Thanks!

jcar87:
Hi @toge - thanks for your contribution.

> In some environments, there are compilation errors in static build.

When this recipe was last built on CI (Linux, gcc11, shared=False), [build log here](https://c3i.jfrog.io/c3i/misc-v2/logs/pr/18812/4-linux-gcc/libjxl/0.10.2//4ce43b2cd05eb16eaa16c4e1031d383741ed92f2-build.txt)  there were no failures related to this. Could you provide more details as to which environments are problematic?


I suspect the problem with this recipe is actually this line:
https://github.com/conan-io/conan-center-index/blob/d8515e82deeb27522b67549ba3237e99c6f55287/recipes/libjxl/all/conanfile.py#L95

which should always be set to off. That variable applies to the tools (the executables), rather than the libraries: https://github.com/libjxl/libjxl/blob/main/CMakeLists.txt#L151-L152

Setting this to off would prevent this:
```
-- Using -lpthread as --whole-archive
```

The variable that controls static/shared libraries in CMake's conventional `BUILD_SHARED_LIBS`, as per https://github.com/libjxl/libjxl/blob/main/CMakeLists.txt#L97

jcar87 (changes requested):
Please set `JPEGXL_STATIC` to `OFF` instead, which applies to the executables and only refers to the runtime. The upstream project sets this to off by default, and it is taken to mean something like "use static runtime in tool executables", and is unrelated to whether the libraries are being built shared or statically.

toge:
@jcar87 
Thank you for your detail investigation!
I don't understand the meaning of JPEGXL_STATIC.
Thank you very much.

jcar87 (dismissed):

AbrilRBS (dismissed):

AbrilRBS (approved):
I've removed the old 0.6.1 instead of removing the extra libraries it was creating so that we could remove some noise in the recipe :) Thanks a lot @toge for addressing this one quickly :)

jcar87 (approved):

PR #24810, opened 2024-08-04
cigi-cl: new recipe

luizfeldmann:
### Summary
Specify library name and version: cigi-cl/4.0.6a

#### Motivation
From[ Wikipedia](https://en.m.wikipedia.org/wiki/Common_Image_Generator_Interface):

> The Common Image Generator Interface (CIGI) (pronounced sig-ee), is an on-the-wire data protocol that allows communication between an Image Generator and its host simulation. The interface is designed to promote a standard way for a host device to communicate with an image generator (IG) within the industry.

> In the past most image generators provided their own proprietary interface; every host had to implement that interface making changing image generators a costly ordeal. CIGI was created to standardize the interface between the host and the image generator so that little modification would be needed to switch image generators. The CIGI initiative was largely spearheaded by **The Boeing Company** during the early 21st century.

> The latest version of CIGI (CIGI 4.0) was developed by the Simulation Interoperability Standards Organization (SISO) in the form of [SISO-STD-013-2014](https://cdn.ymaws.com/www.sisostandards.org/resource/resmgr/standards_products/siso-std-013-2014_standard_f.pdf)

> **CIGI Class Library**: The **CCL** is an object-oriented software interface that automatically handles message composition and decomposition (i.e. packing, unpacking and byte swapping to the ICD specification) on both the Host and Image Generator sides of the interface. The CCL interprets Host or Image Generator messages based on compile time parameters. It also performs error handling and translation between different versions of CIGI.

#### Details
This is a recipe for the reference implementation of the _CIGI_ protocol.
The library is hosted on [SourceForge](https://sourceforge.net/projects/cigi/), and unfortunately is a bad case of migration to CMake, having some issues like: using Windows' path separators, not properly declaring `cmake_minimum_required` and also declaring two separate targets for the shared and static libraries instead of relying on `BUILD_SHARED_LIBS`.
This demanded some use of the `replace_in_file` to patch the _CMakeLists.txt_ and leave it in a usable state (it might have worked as-is in CMake 2.x but not anymore...).

Despite being somewhat old (latest standard from 2014), CIGI is highly relevant in the niche of flight simulators.
Having this in _conancenter_ would greatly help the use case of _conan_ in the training & simulation industry.

---


luizfeldmann:
Apparently the `_NP` in `PTHREAD_RECURSIVE_MUTEX_NP` means a *non-portable* extension, which is not supported by Apple as stated [here](https://stackoverflow.com/questions/53075327/use-of-undeclared-identifier-pthread-recursive-mutex-initializer-np/53075997#53075997). An inspection of the [code](https://opensource.apple.com/source/libpthread/libpthread-454.80.2/include/pthread/pthread.h.auto.html) shows it's indeed not provided. Therefore, updated recipe so that `validate()` will report Apple is not supported.

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+import os
+
+required_conan_version = ">=1.53.0"
+
+class CigiClConan(ConanFile):
+    name = "cigi-cl"
+    description = "Industry standard communication with compliant image generators"
+    license = "LGPL-2.1-only"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://sourceforge.net/projects/cigi/"
```
uilianries, line 17:
```suggestion
    homepage = "https://cigi.sourceforge.io/product_ccl.php"
```

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, replace_in_file, rm, rmdir
+from conan.tools.microsoft import is_msvc, is_msvc_static_runtime
+from conan.tools.apple import is_apple_os
+import os
+
+required_conan_version = ">=1.53.0"
+
+class CigiClConan(ConanFile):
+    name = "cigi-cl"
```
uilianries, line 13:
```suggestion
    name = "cigi-ccl"
```

I would suggest renaming this project, including the folder to `cigi-ccl`

Repology and Ubuntu are using this name already: https://repology.org/project/ccl-cigi-class-library/versions

There is a "mirror" in github using cigi-ccl as well: https://github.com/Hadron/cigi-ccl

The official web package also mention the ccl: https://cigi.sourceforge.io/product_ccl.php

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+        else:
+            cmake.build(target="cigicl-static")
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+
+        copy(self, "license.html", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.pc", os.path.join(self.package_folder, "lib"), recursive=True)
```
uilianries, line 93:
```suggestion
        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
```
Avoid keeping empty folder.

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+        if self.options.shared:
+            cmake.build(target="cigicl-shared")
+        else:
+            cmake.build(target="cigicl-static")
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+
+        copy(self, "license.html", self.source_folder, os.path.join(self.package_folder, "licenses"))
```
uilianries, line 91:
```suggestion
        copy(self, "license.html", self.source_folder, os.path.join(self.package_folder, "licenses"))
        copy(self, "COPYING", self.source_folder, os.path.join(self.package_folder, "licenses"))
```

Copy copyright as well.

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.pc", os.path.join(self.package_folder, "lib"), recursive=True)
+
+    def package_info(self):
```
uilianries, line 95:
```suggestion
    def package_info(self):
        self.cpp_info.set_property("pkg_config_name", "cigicl")
```

When building the package, I have: 
```
-- Installing: /home/conan/.conan2/p/b/cigi-eea3cb2353ba1/p/lib/pkgconfig/cigicl.pc
```

uilianries (commented):
@luizfeldmann Thank you for submitting this new recipe. Please, take a look in my review.

recipes/cigi-cl/all/conanfile.py
luizfeldmann: Because `CCL` expands to `Cigi-CL`, calling it `Cigi-CCL` has the word `Cigi` twice in the name.
I don't particularly like this, but I guess it's OK to be consistent with the other packages.

recipes/cigi-cl/all/conanfile.py
luizfeldmann: Is this relevant since we are removing the file anyhow with `rm(self, "*.pc", os.path.join(self.package_folder, "lib"), recursive=True)` ?

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if is_apple_os(self):
```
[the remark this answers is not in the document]
luizfeldmann: This is the error while trying to build for *MacOS*:
```
CigiOutgoingMsg.cpp:196:38: error: use of undeclared identifier 'PTHREAD_MUTEX_RECURSIVE_NP'
pthread_mutexattr_settype( &attr, PTHREAD_MUTEX_RECURSIVE_NP );
```

As I mentioned [here](https://github.com/conan-io/conan-center-index/pull/24810#issuecomment-2267611084):
> Apparently the `_NP` in `PTHREAD_RECURSIVE_MUTEX_NP` means a _non-portable_ extension, which is not supported by Apple as stated [here](https://stackoverflow.com/questions/53075327/use-of-undeclared-identifier-pthread-recursive-mutex-initializer-np/53075997#53075997). An inspection of the [code](https://opensource.apple.com/source/libpthread/libpthread-454.80.2/include/pthread/pthread.h.auto.html) shows it's indeed not provided. Therefore, updated recipe so that `validate()` will report Apple is not supported.

I don't think it's an official limitation. Probably just wasn't properly tested.

recipes/cigi-cl/all/conanfile.py
```diff
@@ -0,0 +1,107 @@
...
+        copy(self, "license.html", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rm(self, "*.pc", os.path.join(self.package_folder, "lib"), recursive=True)
+
+    def package_info(self):
+        build_type_suffix = ""
+        if self.settings.build_type == "Debug" and self.settings.os == "Windows":
+            build_type_suffix = "D"
+
+        if not is_msvc(self):
```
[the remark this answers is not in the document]
luizfeldmann: In the source code, it's defined as follows:
```
#ifdef _WIN32
   InitializeCriticalSection( &m_msgLock );
#else
   pthread_mutexattr_t attr;
   ...
```

recipes/cigi-cl/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,32 @@
...
+    int status = msgIn.ProcessIncomingMsg(rawMsgData, msgLength);
+
+    // Cleanup
+    msgOut.FreeMsg();
+    delete pIgSession;
+    delete pHostSession;
+
+    // Test results
+    return (status == CIGI_SUCCESS) ? 0 : -1;
+}
```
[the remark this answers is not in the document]
luizfeldmann: Ok, except for the `auto`.

recipes/cigi-ccl/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def validate(self):
+        if is_apple_os(self):
+            raise ConanInvalidConfiguration(f"{self.settings.os} is not supported")
```
uilianries, line 42:
```suggestion
            # CigiOutgoingMsg.cpp:196:38: error: use of undeclared identifier 'PTHREAD_MUTEX_RECURSIVE_NP'
            raise ConanInvalidConfiguration(f"{self.settings.os} Conan recipe is not supported on Apple. Contributions are welcome")
```

uilianries (approved):
LGTM. Ótimo, muito obrigado!

recipes/cigi-cl/all/conanfile.py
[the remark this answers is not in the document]
uilianries: That's totally okay. I updated the recipe to clarify it better, so anyone can understand why that invalid configuration is there.

recipes/cigi-cl/all/conanfile.py
uilianries: >  Better not to use any of that code for reference.

Totally agree.

AbrilRBS (approved):
Thanks!

Issue #24813, opened 2024-08-04
[question] Rationale of default_options


pasbi:
### What is your question?

When building one of my projects (which uses opencv) using conan, I faced the issue that the opencv-conan-package doesn't support the PFM image codec.

Took me quite a while to figure out that PFM codec is turned off by default: [conanfile.py:200](https://github.com/conan-io/conan-center-index/blob/master/recipes/opencv/4.x/conanfile.py#L200C1-L200C36).

In [OpenCV/CMakeLists.txt:476](https://github.com/opencv/opencv/blob/4.x/CMakeLists.txt#L476), the codec is available by default.
PFM is super simple, it should be available pretty much anywhere with zero dependencies and it's size is probably negligible.
Hence, I see no reason to set it to `OFF` by default.

I think it's crucial to keep the default close to what people expect.
OpenCV has a ton of options and re-building its package takes forever.

Is there a general aim to keep the conanfile.py defaults close to the defaults defined in the package's original CMakeLists.txt (if applicable)?

Is there a particular reason for the `with_imgcodec_pfm`-option defaulting to `False`? Can we change it to `True`?

valgur:
The question is better suited for https://github.com/conan-io/conan-center-index

>  Is there a general aim to keep the conanfile.py defaults close to the defaults defined in the package's original CMakeLists.txt (if applicable)?

The recipes generally try to stick to the default values set by the project, yes.

Regarding the PFM option, if you look at the PR that added it (https://github.com/conan-io/conan-center-index/pull/6802), the rationale was to keep the existing behavior of having them disabled. Since adding features is generally not considered a breaking change, enabling them by default should be acceptable, imo. Consider opening a PR on CCI to continue the discussion there.

/cc @SpaceIm

jcar87:
Closed via https://github.com/conan-io/conan-center-index/pull/24814

As a note on `default_options` -  I think there's different approaches. Traditionally Conan Center has been following "the defaults from the library maintainers" - however sometimes the libraries maintainers will expose too many options, or will default to "depend on everything if found on the system", even if it is for some optional functionality that may not be in the critical path for a library to fulfil its purpose.

We are likely to revisit our guidance in the future - the goal should be for the defaults to provide a library that fulfils its core functionality, finding a good balance between defaults provided, recipe complexity, and number of dependencies (which would increase build times for users with configurations not provided by Conan Center). 

[...]

PR #24814, opened 2024-08-04
Align opencv image codec option default values to upstream defaults

pasbi:
The OpenCV projects sets the default of this value to true, there is no reason to change this. If anything, not enabling this by default can lead to unexpected fails at runtime which can be hard to debug.

### Summary
Changes to recipe:  **lib/opencv**

#### Motivation
The OpenCV projects sets the default of this value to true, there is no reason to change this.
If anything, not enabling this by default can lead to unexpected fails at runtime which can be hard to debug.
See #24813.

#### Details
Changing the default value for option `with_imgcodec_pfm` from `False` to `True`.

---

### DIFF
--- a/recipes/opencv/4.x/conanfile.py
+++ b/recipes/opencv/4.x/conanfile.py
@@ -191,7 +191,7 @@ class OpenCVConan(ConanFile):
         "with_jpeg": "libjpeg",
         "with_png": True,
         "with_tiff": True,
-        "with_jpeg2000": "jasper",
+        "with_jpeg2000": "openjpeg",
         "with_openexr": True,
         "with_webp": True,
         "with_gdal": False,
@@ -358,6 +358,9 @@ class OpenCVConan(ConanFile):
             # in a big dependency graph
             if not self._has_with_wayland_option:
                 self.options.with_gtk = True
+
+        if Version(self.version) < "4.3.0":
+            self.options.with_jpeg2000 = "jasper"
 
     @property
     def _opencv_modules(self):
@@ -1212,6 +1215,9 @@ class OpenCVConan(ConanFile):
             raise ConanInvalidConfiguration(
                 "viz module can't be enabled yet. It requires VTK which is not available in conan-center."
             )
+        if self.options.with_jpeg2000 == "openjpeg" and Version(self.version) < "4.3.0":
+            raise ConanInvalidConfiguration("openjpeg is ot available for OpenCV before 4.3.0")
+
 
     def build_requirements(self):
         if self.options.get_safe("with_protobuf"):
--- a/recipes/opencv/4.x/conanfile.py
+++ b/recipes/opencv/4.x/conanfile.py
@@ -196,10 +196,10 @@ class OpenCVConan(ConanFile):
         "with_webp": True,
         "with_gdal": False,
         "with_gdcm": False,
-        "with_imgcodec_hdr": False,
+        "with_imgcodec_hdr": True,
         "with_imgcodec_pfm": True,
-        "with_imgcodec_pxm": False,
-        "with_imgcodec_sunraster": False,
+        "with_imgcodec_pxm": True,
+        "with_imgcodec_sunraster": True,
         "with_msmf": True,
         "with_msmf_dxva": True,
         # objdetect module options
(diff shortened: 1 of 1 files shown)


valgur:
Should probably enable the other imgcodecs as well if they are enabled by default in OpenCV (I have not checked).

pasbi:
> Should probably enable the other imgcodecs as well if they are enabled by default in OpenCV (I have not checked).

I thought about that, too.
Unfortunately, it's not so simple and I'm afraid I need help to do that.
Some options are not one-to-one translated from CMake to conan.
E.g., whether `WITH_OPENEXR` is enabled by default depends on the platform and more.
In conan you cannot set the `with_jpeg2000`-option to both `jasper` and `openjpeg` at the same time, however, CMake enables both `WITH_JASPER` and `WITH_OPENJPEG` by default.

Here is a quickly compiled overview (didn't double check it, take with care).

conan option | conan default | OpenCV default
-- | -- | --
with_avif | False, | OFF
with_jpeg | "libjpeg", | WITH_JPED=ON
with_png | True, | ON
with_tiff | True, | ON
with_jpeg2000 | "jasper", | WITH_JASPER = ON, WITH_OPENJPEG=ON
with_openexr | True, | ((WIN32 OR ANDROID OR APPLE) OR BUILD_OPENEXR) OR NOT CMAKE_CROSSCOMPILING
with_webp | True, | ON
with_gdal | False, | OFF
with_gdcm | False, | OFF
with_imgcodec_hdr | False, | ON
with_imgcodec_pfm | False, | ON
with_imgcodec_pxm | False, | ON
with_imgcodec_sunraster | False, | ON
with_msmf | True, | NOT MINGW
with_msmf_dxva | True, | WITH_MSMF

SpaceIm:
> In conan you cannot set the `with_jpeg2000`-option to both `jasper` and `openjpeg` at the same time, however, CMake enables both `WITH_JASPER` and `WITH_OPENJPEG` by default.

Indeed, these 2 options are enabled by default in CMakeLists of OpenCV, BUT in cmake/OpenCVFindLibsGrfmt.cmake, there is a logic where openjpeg has precedence over jasper when both options are enabled (jasper branch is disabled when openjpeg is found).
So technically openjpeg should be the default value in conanfile of opencv. But in the meantime, in old versions jasper was the default jpeg2000 backend.

SpaceIm:
And yes regarding all these with_imgcodec options, you don't see any reason to not change their default value. There default values in conanfile come from https://github.com/conan-io/conan-center-index/pull/6802

pasbi:
Thanks @SpaceIm for your answer.

I'm not sure if I understand correctly.
The intention of this pull request is to change the default of the `with_imgcodec_pfm` option to `True`.

I have explained why I think this would be a better default exhaustively. 
I don't understand how you can say that you don't see a reason, I have just presented it.
Please explain why you think it's not worth changing them.

In addition to that, @valgur proposed to align the default values of the other (image codec) options with the upstream `CMakeLists.txt`.
That's why I did that short survey.
I have no particular interest in changing any of the other `with_imgcodec_*`, `with_jpeg*` or other default values (though I agree it could make sense).

pasbi:
@SpaceIm  please have a look again.

recipes/opencv/4.x/conanfile.py
```diff
@@ -1212,6 +1215,9 @@ def validate(self):
             raise ConanInvalidConfiguration(
                 "viz module can't be enabled yet. It requires VTK which is not available in conan-center."
             )
+        if self.options.with_jpeg2000 == "openjpeg" and Version(self.version) < "4.3.0":
+            raise ConanInvalidConfiguration("openjpeg is ot available for OpenCV before 4.3.0")
```
valgur, line 1219:
```suggestion
            raise ConanInvalidConfiguration("openjpeg is not available for OpenCV before 4.3.0")
```

pasbi:
@jcar87 thanks for answering.

My intention was just to update the default of `with_imgcodec_pfm`.
Though I think it makes sense to apply the other suggested changes in the same PR.

To me merging this PR is not urgent, so I'm fine.
I have updated the PR title to reflect the more thorough changes.

AbrilRBS (approved):
Thanks!

SpaceIm (approved):

jcar87 (approved):

mologie:
Unfortunately this PR breaks building OpenCV on iOS completely, because option with_jpeg2000 is accessed unconditionally on line 1218, but is deleted when the target is iOS, which this crashes in verify(). I will open a PR to address this shortly.

mologie:
No worries, thanks for helping out with Conan either way, was a quick fix after all o/

Our CI at @authenticvision happens to catch iOS-specific bugs in Conan recipes all the time since there unfortunately is no official CI that builds for mobile systems, which makes these bugs quite hard to spot during review.

PR #24825, opened 2024-08-05
daw_utf_range: update daw_header_libraries/2.110.0

toge:
### Summary
Changes to recipe:  **daw_utf_range/2.2.5**

#### Motivation
daw_header_libraries/2.110.0 fixed several bugs.
I want to apply these fixes to daw_utf_range and daw_json_link.

#### Details

---

### DIFF
--- a/recipes/daw_utf_range/all/conandata.yml
+++ b/recipes/daw_utf_range/all/conandata.yml
@@ -15,7 +15,7 @@ sources:
     url: "https://github.com/beached/utf_range/archive/v2.2.0.tar.gz"
     sha256: "00f60360736062403c8a7a94dd07c750366958f20d1864578faecf2e09d84265"
 daw_headers_mapping:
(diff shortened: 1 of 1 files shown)


AbrilRBS:
I'm thinking whether it would make sense to allow version ranges for these dependencies among eachother. Do you know if the range in which they are compatible with eachother is well defined somewhere? We could then use that to change the dependencies so that we have no need to update as new versions are released. Wdyt @toge?

toge:
@AbrilRBS 
Unfortunately, I don't think there is a clear information for each libraries version.
The daw_json_link manages the two libraries that depend on it in [extern/CMakeLists.txt](https://github.com/beached/daw_json_link/blob/release/extern/CMakeLists.txt), but since it uses master branch, it is not helpful in conan.

AbrilRBS (approved):
Thanks!

AbrilRBS (approved):

Issue #24831, opened 2024-08-05
[package] aws recipes: Refactor aws dependency versions according to aws-sdk-cpp


danimtb:
### Description

I have recently taken a look at all AWS-related PRs and realized that the group of these recipes are commonly used together usually requiring the aws-sdk-cpp package from the consumer recipe.

The problem right now is that this recipes had version-bumped its requirements without apparent criteria which has led to the situation of having different issues:
- Version conflicts.
- Many conditional requirements are hard to debug for the number of versions for each recipe.
- Versions that are not clear to work with each other.
- Outdated version of the aws-sdk-cpp.

So to overcome this situation we have worked to establish a working flow to keep a sane dependency graph for the group of aws-sdk-cpp recipes with the following rules:
- Versions in CCI of the AWS packages should be defined by the aws-sdk-cpp library (_**other versions should be removed if they are not used by a version of aws-sdk-cpp package as they are not guaranteed to work together**_).
- Requirements should have fixed versions (avoid conditional version requirements that makes it difficult to update).
- The versions of any AWS package are bumped whenever there is a version update of the aws-sdk-cpp package, as this is the main entry point for users.
- The process for this should start with an issue similar to the second part of this one, specifying the dependency version and opening the related PRs.

This should fix many of the version conflicts and, in general, the usability of AWS recipes.

-------
 As mentioned above, the criteria to update and require any version of an AWS recipe should be defined by the aws-sdk-cpp library. In this case, I have found that from some versions of ago in the aws-sdk-cpp library, there is a script to fetch its dependencies that has the commits of the dependencies needed to compile it: https://github.com/aws/aws-sdk-cpp/blob/1.9.234/prefetch_crt_dependency.sh

Going through that script led to the following versions for its requirements (direct and transitive) for **aws-sdk-cpp/1.9.234**:
- **aws-c-common**: https://github.com/awslabs/aws-c-common/commit/79e77e4238dc545da3d37992d4f7731206e29e16 --> Version: 0.6.11
- **aws-c-compression**: https://github.com/awslabs/aws-c-compression/commit/5fab8bc5ab5321d86f6d153b06062419080820ec --> Version: 0.2.14
- **aws-c-sdkutils** (not required in this aws-sdk-cpp version - aws-crt-cpp & aws-c-auth dependency -)
[...]

PR #24843, opened 2024-08-06
[aws refactor] aws-c-compression/0.2.14: require aws-c-common/0.6.11

danimtb:
Related to https://github.com/conan-io/conan-center-index/issues/24831


recipes/aws-c-compression/all/conanfile.py
```diff
@@ -40,10 +40,10 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        if Version(self.version) <= "0.2.15":
-            self.requires("aws-c-common/0.8.2", transitive_headers=True, transitive_libs=True)
-        else:
-            self.requires("aws-c-common/0.9.6", transitive_headers=True, transitive_libs=True)
+        if self.version == "0.2.18":
```
AbrilRBS, line 43: Can we move this to the `conandata.yml` so that changing this in the future does not change the revision of unrelated recipes?

AbrilRBS (commented):
Minor comment, thanks!

recipes/aws-c-compression/all/conanfile.py
danimtb: Given the number of aws recipes and the different versions of each one, I believe it is more straightforward to keep the conditional dependencies directly in the recipe, even if changes in the future produce a new recipe revision. However, we would need to reach an agreement here cc/ @czoido @perseoGI

AbrilRBS (approved):

perseoGI (approved):

czoido (approved):

PR #24857, opened 2024-08-06
New recipe: reflectcpp; intended to replace recipe reflect-cpp, which is named incorrectly

liuzicheng1987:
### Summary
Adds a new recipe:  **reflectcpp/0.14.0**. The recipe "reflect-cpp" should be deprecated or removed, because its name is wrong.

#### Motivation
As I explained to @AbrilRBS and @uilianries  in https://github.com/conan-io/conan-center-index/pull/24819, the problem is the following:

I am the main author of reflectcpp, but I didn't upload any of the previous Conan packages for reflectcpp. Other people did that without informing me. I found out about its existence, because users were complaining to me about the Conan package.

In principle, I am happy about people uploading my library to Conan, however they made a mistake: The have called they package "reflect-cpp", but in our CMakeLists.txt, we have always referred to the package as "reflectcpp":

https://github.com/getml/reflect-cpp/blob/main/CMakeLists.txt

But a requirement of the Conan package is that the name of the installed package be the same as the name of the package in the CMakeLists.txt (which makes a lot of sense). 

To be honest, I may have contributed to the confusion, because I often refer to the package as "reflect-cpp". However, it is the name in the CMakeLists.txt that matters for our purposes and that name is "reflectcpp".

This didn't matter as long as the library was treated as header-only. But this was a workaround from the get-go and in any case it won't work for any of the newer versions.

The name of the package on Conan is wrong, plain and simple. If you stick with the old name you won't be able to host any of the newer versions of reflectcpp and will be stuck with an outdated version for all eternity. It needs to be fixed.

#### Details
I have added a build pipeline. I do not use the bundled dependencies, but install them via Conan instead. JSON support is activated by default. Since you only want very simple tests, not comprehensive functional testing (https://docs.conan.io/2/tutorial/creating_packages/test_conan_packages.html), I have done just that.

---


liuzicheng1987:
Can someone give me some advice? This package only support C++20, but the pipeline keeps on trying to build it for earlier versions. What do I do here? This can't be the first package facing this problem.

toge:
@liuzicheng1987 
I have written several PR updating reflect-cpp recipes.
Now I see that reflectcpp is the official name, not reflect-cpp.
I had no doubts about the name of reflect-cpp because of your reddit post and the name of repository.
I apologize for the unnecessary hassle this has caused you.

You need to implement `validate()` to make CI work only for compilers that support C++20.
Could you please refer me to the existing reflect-cpp recipe or to the conan-center-index template?

https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/header_only/all/conanfile.py#L63-L71

uilianries (commented):
@liuzicheng1987 Please, keep the original test package from https://github.com/conan-io/conan-center-index/tree/master/recipes/reflect-cpp/all/test_package

The test package is used to validate those packaged artifacts, like headers and libraries, not including the headers, but also linking libraries. Plus, we do not expose rules of settings in test_package.

liuzicheng1987:
@toge, absolutely no problem. Like I said, I probably contributed to the confusion, because I often refer to it as "reflect-cpp" myself.

I am grateful there are so many people who are willing to contribute to the library.

uilianries:
@liuzicheng1987 Thank you again for your effort on this PR. I opened a PR to suggest all changes requested by me, instead of doing directly in the PR. It would faster than asking with suggestions here. Please, take a look: https://github.com/liuzicheng1987/conan-center-index/pull/1

liuzicheng1987:
@uilianries , sounds great. Thank you so much for your help. I will take a closer look tonight.

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,154 @@
...
+        minimum_version = self._compilers_minimum_version.get(
+            str(self.settings.compiler), False
+        )
+        if (
+            minimum_version
+            and Version(self.settings.compiler.version) < minimum_version
+        ):
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
```
uilianries, line 154:
```suggestion
```
Ops! It's duplicated.

liuzicheng1987: Yeah, already fixed it

valgur:
While the incorrect name is unfortunate, I would personally vote to keep the existing `reflect-cpp` recipe and add a
```py
self.cpp_info.set_property("cmake_file_name", "reflectcpp")
self.cpp_info.set_property("cmake_target_name", "reflectcpp")
```
to its `package_info()` to fix the incorrect CMake names.

Deprecating the old package does not remove it from CCI and having it linger side-by-side with the correctly-named one only adds to the confusion instead of making things clearer.

liuzicheng1987:
It appears that the Conan Center Index Build Service has been under maintenance for the last 2-3 days, so I can't build the pipeline. Does anyone know how long that is going to continue?

Honestly, I am not too opposed to @valgur's suggestion. I simply didn't know that this was possible until he pointed it out. What does everyone else think about this?

liuzicheng1987:
@uilianries , it took a bit longer than anticipated and I had to make a minor upstream change, but it works now. Could you give me some feedback?

Also, how are we going to proceed about @valgur 's proposal?

uilianries:
@liuzicheng1987 Sorry the delay, after discussion with Conan team we will add this new recipe and mark the old one as deprecated. In the future (few months) we can remove it from Conan Center as well, but because it's your particular case, removing recipes is not something common.

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,152 @@
...
+            self.options.rm_safe("fPIC")
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def requirements(self):
+        if not self.options.shared:
+            self.requires("ctre/3.9.0", transitive_headers=True)
+            self.requires("yyjson/0.8.0", transitive_headers=True)
```
uilianries, line 67:
```suggestion
        if not self.options.shared:
            self.requires("ctre/3.9.0", transitive_headers=True)
            self.requires("yyjson/0.8.0", transitive_headers=True)
```

Why does adding those requirements only when static?

uilianries:
I just opened the PR https://github.com/conan-io/conan-center-index/pull/25137 to deprecate the reflect-cpp.

liuzicheng1987:
@uilianries , OK. Is there anything I still need to do?

uilianries:
@liuzicheng1987 Yes, please, check my question above: https://github.com/conan-io/conan-center-index/pull/24857#pullrequestreview-2277764954

liuzicheng1987:
@uilianries , I responded to your question. Sorry it took me such a long time to see this.

But now the jenkins/pr-merge and it won't tell me why. Do you have any idea what went wrong here?

uilianries:
@liuzicheng1987 Sorry, those "details" link goes to ConanCenterIndex internal Jenkins, that's forbidden for the public for security reasons. What it does is posting the build results as Github comments, as you can see in this PR. However, the last build failed internally, and was not capable to post the results. I triggered a new build just now, thank you for pinging.

recipes/reflectcpp/all/conanfile.py
liuzicheng1987: If I set them to False, where would YYJSON come from? Like I said it can't come from the external libraries if they are linked statically but the library is linked dynamically.

toge:
@uilianries 
Please review it again.
Several people including me want to use latest vension.

liuzicheng1987:
@uilianries , I have now also added support for version 0.15.0, which I just released.

Could you please review this again? It appears there is quite a bit of demand for this...

recipes/reflectcpp/all/conanfile.py
uilianries: > If I set them to False, where would YYJSON come from?

It will be provided by Conan, using VirtualRunEnv generator:

https://docs.conan.io/2/reference/tools/env/virtualbuildenv.html

This generator is automatically executed. The test package is the example of it, it find all transitive dependencies by the configured environment variable (e.g LD_LIBRARY on Linux).

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,152 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        env = VirtualBuildEnv(self)
+        env.generate()
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["REFLECTCPP_USE_BUNDLED_DEPENDENCIES"] = self.options.shared
```
uilianries, line 111:
```suggestion
        tc.cache_variables["REFLECTCPP_USE_BUNDLED_DEPENDENCIES"] = False
```
Use dependencies only provided by Conan

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,152 @@
...
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        env = VirtualBuildEnv(self)
+        env.generate()
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIBS"] = self.options.shared
```
uilianries, line 110:
```suggestion
        tc.cache_variables["REFLECTCPP_BUILD_SHARED"] = self.options.shared
```

Reflectcpp has its own shared option.

uilianries (commented):
The new minor looks good in terms of no breaking the current recipe state: https://github.com/getml/reflect-cpp/compare/v0.14.1...v0.15.0

recipes/reflectcpp/all/conanfile.py
liuzicheng1987: @uilianries , if I do this, the tests won't pass for reasons I have explained above. I have tried my best to recreate the problem building the library without Conan, but to no avail. It must be something that is somehow specific to Conan. My theory is that the option `shared` is not passed through to YYJSON and that is why the linking fails.

uilianries: It's failing because of yyjson is not linked to shared, only static. As I commented above, you need to remove that condition and link both yyjson and ctre always.

danimtb (approved):

liuzicheng1987:
@uilianries , I think that I now have resolved all of the issues and this is ready to be merged.

If you agree, I will maintain the Conan package from now on and also add more comprehensive testing on my side to make sure that the Conan options we provide all work as expected.

The issue with the shared build was a bit more complicated than we both anticipated, but the good news is that I got it resolved.

Fixing it required some changes on our side. That is why the shared library version of reflectcpp-0.14.0 and reflectcpp-0.15.0 won't build. But I have included the necessary changes in the newest release,  reflectcpp-0.16.0.

The other change necessary was to add this line to the CMakeLists.txt:

```
SET(CMAKE_EXE_LINKER_FLAGS  "${CMAKE_EXE_LINKER_FLAGS} -Wl,--copy-dt-needed-entries")
```

Passing this linker flag is necessary whenever, the following three statements hold true:

- You are linking a shared library
- The shared library links to another shared library that is written in a different language
- You also require direct access to the other library

Because YYJSON is written in C (as well as some of the other libraries we use), these three statements hold true when `shared=True`.

liuzicheng1987:
@uilianries, yes, the changes are incorporated upstream.

The error message is the following:

DSO missing from command line

I would refer you to this website which at greater detail how this error occurs.

_The DSO missing from command line message will be shown when the linker can’t find the required symbol with its normal search, but the symbol is available in one of the dependencies of a directly specified dynamic library.

In the past, the linker thought that symbols in languages that depended on other languages were available. But in a later version, that changed, and now the linker has a stricter view of what is available. So, the message is meant to help with this change._

https://linuxpip.org/how-to-fix-dso-missing-from-command-line/?utm_content=cmp-true

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+        if self.settings.os == "Windows":
+            self.options.rm_safe("fPIC")
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def requirements(self):
+        self.requires("ctre/3.9.0", transitive_headers=True)
+        self.requires("yyjson/0.10.0", transitive_headers=True)
```
uilianries, line 68:
```suggestion
        # INFO: include/rfl/json/Writer.hpp includes yyjson.h
        # INFO: Transitive lib needed to avoid undefined reference to symbol 'yyjson_mut_doc_new'
        self.requires("yyjson/0.10.0", transitive_headers=True, transitive_libs=True)
```

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,148 @@
...
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support.")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
```
uilianries, line 98:
```suggestion
        get(self, **self.conan_data["sources"][self.version], strip_root=True)
        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "set(CMAKE_CXX_STANDARD 20", "")
```

Let Conan configure the C++ standard based on settings.compiler.cppstd. Otherwise, will use always C++20.

uilianries (commented):
@liuzicheng1987 You current error is because yyjson's method `yyjson_mut_doc_new` is exposed directly in write.hpp, and Conan 2.x does not export dependencies libraries: 

https://github.com/getml/reflect-cpp/blob/main/include/rfl/json/write.hpp#L44

In order to fix that, you have to explicit tell to Conan to make yyjson available to the upstream.

recipes/reflectcpp/all/conanfile.py
```diff
@@ -0,0 +1,151 @@
...
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "set(CMAKE_CXX_STANDARD 20)", "")
+
+    def generate(self):
+        env = VirtualBuildEnv(self)
+        env.generate()
+        deps = CMakeDeps(self)
```
uilianries, line 106:
```suggestion
        deps = CMakeDeps(self)
        if self.options.with_flatbuffers:
            deps.set_property("flatbuffers", "cmake_target_name", "flatbuffers::flatbuffers")
```

Because Conan generates the CMake target `flatbuffers::libflatbuffers`. This change fixes the follow error:


```
-- Configuring done (1.2s)
CMake Error at CMakeLists.txt:131 (target_link_libraries):
  Target "reflectcpp" links to:

    flatbuffers::flatbuffers

    * An ALIAS target is missing.



-- Generating done (0.1s)
CMake Generate step failed.  Build files cannot be regenerated correctly.

reflectcpp/0.16.0: ERROR: 
Package 'e71c47e4a8170b93577b9cb78554080767eaafb9' build failed
```

liuzicheng1987:
@uilianries, if there is anything I can do to help, please let me know.

uilianries:
It requires MSVC version 193, Update 8. However, CCI does not use Update in settings. I'll check with the team.

uilianries:
@liuzicheng1987 I did few changes in your PR:

- Moved your PR to folder reflect-cpp, so it's no longer a "new" entire recipe, but a change over the current reflect-cpp
- Added features related to Conan 2.x. The Conan 2.x is mandatory now in CCI: https://github.com/conan-io/conan-center-index/discussions/25461
- The CI is running MSVC 19.3 update 6, but your project requires update 8. The settings.yml supports the update number as part of the profile, but we do not use it, so we can not validate it in the recipe as well. The CI should be updated to use the latest update version available, but it takes time and we don't want to delay you even more, so the minimal MSVC required is 194 in the recipe, but should be changed as soon as we have the CI fixed.

Thank you again for your big effort over this recipe. I built locally again, using dependencies and it looks good to me now.

uilianries (approved):
LGTM

liuzicheng1987:
@uilianries, thank you so much for your comments as well!

AbrilRBS (approved):

PR #24868, opened 2024-08-08
Set PNG_INTEL_SSE to 'off' if arch doesn't support intel intrinsics

jasal82:
### Summary
Changes to recipe:  **libpng/x**

#### Motivation
The libpng `CMakeLists.txt` sets the default for `PNG_INTEL_SSE` to `on`, so the logic in the `conanfile.py` generate method should be to set it to `off` if the architecture doesn't support intel sse intrinsics. However, it is instead setting it to `on` if the architecture **does** support sse, so the path where arch has no sse instructions currently leaves the flag set to `on`, making CMake compile the intel specific files and breaking builds downstream.

#### Details
To quickfix the issue I added an `else` branch to the `if` block which sets the variable to `off` if no support is detected (which is the safest solution in my opinion, because the CMakeLists.txt could change its default semantics in future versions, so a simple inversion of the condition is not robust enough).

There might be other variables that should be handled the same.

---

### DIFF
--- a/recipes/libpng/all/conanfile.py
+++ b/recipes/libpng/all/conanfile.py
@@ -116,6 +116,8 @@ class LibpngConan(ConanFile):
             tc.variables["PNG_MIPS_MSA"] = self._neon_msa_sse_vsx_mapping[str(self.options.msa)]
         if self._has_sse_support:
             tc.variables["PNG_INTEL_SSE"] = self._neon_msa_sse_vsx_mapping[str(self.options.sse)]
+        else:
+            tc.variables["PNG_INTEL_SSE"] = "off"
         if self._has_vsx_support:
             tc.variables["PNG_POWERPC_VSX"] = self._neon_msa_sse_vsx_mapping[str(self.options.vsx)]
         if Version(self.version) >= "1.6.41":


jcar87:
Thanks @jasal82 ! Good catch.

jcar87:
Hi @jasal82  - thanks for your contribution.

I was bitten by this as well, although I think there's something else at play.

If the recipe _doesn't_ set `PNG_INTEL_SSE` at all because we are not in x86 or x86_64, it is because the CMakeLists.txt scritps of libpng themselves don't define this as an option in exactly the same cases

For 1.6.32:

https://github.com/pnggroup/libpng/blob/v1.6.32/CMakeLists.txt#L140-L145

```
if(CMAKE_SYSTEM_PROCESSOR MATCHES "^i?86" OR
	CMAKE_SYSTEM_PROCESSOR MATCHES "^x86_64*" )
  set(PNG_INTEL_SSE_POSSIBLE_VALUES on off)
  set(PNG_INTEL_SSE "on" CACHE STRING "Enable INTEL_SSE optimizations:
     off: disable the optimizations")
```

and for newer versions:
```
if (APPLE AND CMAKE_OSX_ARCHITECTURES)
  set(TARGET_ARCH ${CMAKE_OSX_ARCHITECTURES})
else()
  set(TARGET_ARCH ${CMAKE_SYSTEM_PROCESSOR})
endif()

...

# Set definitions and sources for Intel.
if(TARGET_ARCH MATCHES "^(i[3-6]86|x86|AMD64)")
  set(PNG_INTEL_SSE_POSSIBLE_VALUES on off)
  set(PNG_INTEL_SSE "on"
      CACHE STRING "Enable INTEL_SSE optimizations: on|off; on is default")
```

> so the path where arch has no sse instructions currently leaves the flag set to on, making CMake compile the intel specific files and breaking builds downstream.

From the above, what you describe can only happen if libpng is detecting `CMAKE_SYSTEM_PROCESSOR` as an intel one, but it actually isn't.

I have seen that happen when cross-building and the toolchain does not define the `CMAKE_SYSTEM_PROCESSOR`,.. seems like cmake just uses the one for the build machine.

Is it the case that you are cross-compiling? I would make sure to pass the `tools.cmake.cmaketoolchain:system_processor` configuration (or define it in the profile), so that CMake actually defines the correct architecture.

On the other hand I don't mind explicitly defining the options regardless, and setting them to off when we know we're not in that architecture.
At most we would get warnings from CMake, but I don't think this a big deal and the recipe would look cleaner
```
CMake Warning:
  Manually-specified variables were not used by the project:

    PNG_INTEL_SSE
    PNG_MIPS_MSA
    PNG_POWERPC_VSX
```

jasal82:
Hi @jcar87 ,

you are right. We already stumbled across the underlying problem yesterday, which is indeed related to `CMAKE_SYSTEM_PROCESSOR`. We found that the value of the variable is reset to x86 during build, although we set it correctly from the user toolchain file. However, it seems that Conan expects the user toolchain file to set all three variables `CMAKE_SYSTEM_NAME`, `CMAKE_SYSTEM_VERSION` and `CMAKE_SYSTEM_PROCESSOR`, which is not documented anywhere. We only had `CMAKE_SYSTEM_PROCESSOR` in that file, under the assumption that the toolchain generator would provide sane defaults for the given Conan settings, but obviously it is setting it to `None` in this case.

jcar87:
Hi @jasal82 - 

I've checked the logic that deals with this, and CMakeToolchain should set those 3 variables when cross-building, _except_ when `tools.cmake.cmaketoolchain:user_toolchain` is set - in that case it expects that file to take that responsibility.

This is implied in the documentation here - although admittedly it could be clarified: https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#cross-building 

On the other hand - from what you are commenting, I don't think it is Conan that is resetting the value - I suspect it is CMake.
The documentation for `CMAKE_SYSTEM_PROCESSOR` states:

> When not cross-compiling, this variable has the same value as the [CMAKE_HOST_SYSTEM_PROCESSOR](https://cmake.org/cmake/help/latest/variable/CMAKE_HOST_SYSTEM_PROCESSOR.html#variable:CMAKE_HOST_SYSTEM_PROCESSOR) variable.

The key is the first bit - "when not cross-compiling" . CMake only assumes it is cross-compiling if `CMAKE_SYSTEM_NAME` is explicitly set in a toolchain file.

what I believe is happening:
- CMakeToolchain is not setting any of those 3 variables because a `tools.cmake.cmaketoolchain:user_toolchain` is provided
- The `user_toolchain` does not set `CMAKE_SYSTEM_NAME`, but it may set the `CMAKE_SYSTEM_PROCESSOR`
- CMake will then just assume it's not cross-building and override the value of `CMAKE_SYSTEM_PROCESSOR`

I've opened https://github.com/conan-io/conan/issues/16807 to see if we can improve this - but in the meantime I suspect you can either;
* define `CMAKE_SYSTEM_NAME` and `CMAKE_SYSTEM_PROCESSOR` in your user toolchain file
or define the confs that will cause the right values to be set in the conan-generated toolchain:
* `tools.cmake.cmaketoolchain:system_processor`
* `tools.cmake.cmaketoolchain:system_name`

jasal82:
Thanks for the explanation. We now set `CMAKE_SYSTEM_NAME` in the user toolchain file. That makes the build select the expected source files.

jasal82:
Now we're stuck further down in the build because it seems that Conan/CMake is appending a `--sysroot=...` argument to the assembler invocation, which is not allowed in our embedded toolchain. Any idea where this could be coming from and how to disable it?

jcar87:
> Now we're stuck further down in the build because it seems that Conan/CMake is appending a `--sysroot=...` argument to the assembler invocation, which is not allowed in our embedded toolchain. Any idea where this could be coming from and how to disable it?

Ha! Your timing is impeccable, as I faced this just a couple days ago.
I'm assuming you are setting the assembler executable in `tools.build:compiler_executables`?  I had success removing the assembler from that list - then cmake uses the _compiler_ as the assembler, and then it works. Most (emphasis on _most_) other build systems (e.g. autotools) _should_ be clever enough to deduce the compiler executable just based on the name of the compiler executable.

I'm not sure if this is a CMake bug when one sets the assembler executable directly, where CMake appends `--sysroot` then the assembler executable is specified. I can confirm Conan is _not_ doing it.

jasal82:
Yep, that did the trick! But the reason why we added these `compiler_executables` settings in the first place was some third party builds not being able to detect the compiler correctly (for example boost b2). I need to verify that these builds still work after that change.

jcar87:
> Yep, that did the trick! But the reason why we added these `compiler_executables` settings in the first place was some third party builds not being able to detect the compiler correctly (for example boost b2). I need to verify that these builds still work after that change.

The `c` and `cpp` compilers are probably good to keep - it's the `asm` one that may be different - although I need to double check, I think passing `--sysroot` might be a cmake issue. On the other hand most examples of CMake toolchains I can find, set the asm compiler to the same executable as the C compiler

jasal82:
And I think CMake **always** appends the `--sysroot` flag. If you remove the `asm` setting in `compiler_executables`, CMake will just use `gcc` as assembler executable, which of course accepts `--sysroot`. Omg, I hate CMake.

jasal82:
So I guess it should also work to set `compiler_executables.asm` to the respective gcc binary, e.g. `aarch64-oe-linux-gcc`.

jasal82:
Confirmed. This is what we go with now in our Yocto toolchain wrappers:
```
c = env["CC"].split(" ")[0]
cpp = env["CXX"].split(" ")[0]
self.conf_info.define("tools.build:compiler_executables", {
    "asm": c,
    "c": c,
    "cpp": cpp
})
```
Hopefully that won't break any other build.

jasal82:
I think it's ok to keep the current logic, since my specific error was just a follow-up error of another problem. Thanks for your help.

Issue #23567, opened 2024-04-15
[package] qt/5.x: Recipe validation errors are hard to diagnose


planetmarshall:
### Description

I realise this is probably not a particularly high priority, but it's difficult to diagnose:

```
ERROR: conanfile.py (qt/5.15.13): Error in configure() method, line 271
	_enablemodule(module)
while calling '_enablemodule', line 260
	setattr(self.options, mod, True)
	ConanException: Incorrect attempt to modify option 'qtmultimedia' from 'False' to 'True'
```

As meaning that I have to set `qt/*:essential_modules=False` in order for the recipe to validate. I can probably fix this myself but it might be a quick fix for someone more familiar with the recipe.

### Package and Environment Details

* Package Name/Version: **qt/5.15.13**
* Conan version: **conan 2.2.2**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=13
os=Linux

### Steps to reproduce

```
conan create recipes/qt/5.x.x/ --version 5.15.13 -o 'qt/*:qtmultimedia=False' --build missing
```

### Logs

```
======== Computing dependency graph ========
ERROR: conanfile.py (qt/5.15.13): Error in configure() method, line 271
	_enablemodule(module)
while calling '_enablemodule', line 260
	setattr(self.options, mod, True)
	ConanException: Incorrect attempt to modify option 'qtmultimedia' from 'False' to 'True'

```

ericLemanissier:
@planetmarshall can you try with https://github.com/conan-io/conan-center-index/pull/23571 ?

planetmarshall:
LGTM, thanks.

PR #23171, opened 2024-03-19
qt: add options to enable several modules at once

ericLemanissier:
Specify library name and version:  **qt/***

categories of modules which can be enabled: essential, addon, deprecated and preview

also, remove a dangling patch

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #24876, opened 2024-08-08
qt refactor option handling and error reporting

jcar87:
Recipes: qt5 and qt6

## Summary
- Restore module option defaults  to what they were before https://github.com/conan-io/conan-center-index/pull/23171, that is, set `essential_modules` is False by default in all cases
- Avoid using different option defaults on CI than outside of it - to ensure that users can properly troubleshoot missing binaries with `conan graph explain` (see https://github.com/conan-io/conan/issues/16215)
- Add debug calls and more information - this can help explain which options are enabled and why (conan needs to be invoked with `-vv`)
- Improve handling of errors when options are contradictory:
    - Instead of `"ConanException: Incorrect attempt to modify option 'qtmultimedia' from 'False' to 'True'"`
    - Now we get: `ERROR: qt/5.15.14: Invalid configuration: Modules ['qtgraphicaleffects'] are explicitly disabled, but are required by ['qtquickcontrols2'], enabled by other options`
  

Close https://github.com/conan-io/conan/issues/16215
Close https://github.com/conan-io/conan-center-index/issues/23567

## Details

* This affects the "module" options that control whether specific submodules are built or not.
* Options can be passed either via command line `-o`, inside a profile, or propagated from another recipe that requires this
* In this recipe, these module options *do not have a default value*

### Default behaviour when no options are passed
* None of the submodule options are enabled, and the are all `False` by default
* `gui` and `widgets` are built by default 
* The two above are the baseline and restore the behaviour that the qt recipes had before https://github.com/conan-io/conan-center-index/pull/23171

### When specific submodules are set set to `=True`
* They are enabled explicitly
* Their dependencies will be set to `=True` as well, even if they were not enabled explicitly

### When any submodule is explicitly set to `=False`
* Typically not needed, since they are all False by default
* If a module is set to False is a requirement of another one set to `True`, this will now display a more informative error

### Notes convenience options, e.g `essential_modules=True`
* These were introduced in https://github.com/conan-io/conan-center-index/pull/23171 and provide a "shortcut" to enable multiple options at once, with the behaviour described below:
* All of this modules  in the essential category will be enabled and set to `=True`, _as if_ they had been explicitly requested individually - as per the logic above
* But if ANY module (in a category or a requirement) is set to False, this is respected
* If `essential_modules=True`, but some modules are explicitly set to `False`:
     - This is okay if the modules set to False are not a dependency of something else that is enabled
     - Otherwise, display an informative error

> [!WARNING]  
>  There are some issues with the granularity of `essential_modules=True` and would discourage its use
> Namely: the recipe expresses modules at the level of the Qt repositories, however, as per [Qt's website](https://doc.qt.io/qt.html)) these are expressed directly as modules. We have observed some issues:
>    - `qtdoc` is marked as "essential" by the recipe - but this generates the Qt documentation, which is unlikely to be essential to the recipe consumers
>    - similar case with `qttools` - some tools are arguably not essential, but they're all lumped together
>    - the dependencies are expressed between repositories, but not between actual modules -so for example, the logic in the dependencies are not enough to get a successful configuration of options when `qtwebengine` is enabled

### DIFF
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -123,9 +123,10 @@ class QtConan(ConanFile):
         "sysroot": None,
         "config": None,
         "multiconfiguration": False,
-        "essential_modules": not os.getenv('CONAN_CENTER_BUILD_SERVICE')
     }
-    default_options.update({f"{status}_modules": False for status in _module_statuses if status != "essential"})
+    # essential_modules, addon_modules, deprecated_modules, preview_modules:
+    #    these are only provided for convenience, set to False by default
+    default_options.update({f"{status}_modules": False for status in _module_statuses})
 
     no_copy_source = True
     short_paths = True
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
debug needs to be changed for Conan 1, otherwise this looks great!

jcar87:
Example outputs:

Qt5:
```
conan create . --version=5.15.14 -vv  -o "qt/*:essential_modules=True" -o "qt/*:qtmultimedia=False"  -o "qt/*:qtgraphicaleffects=False"
```

Screenshot 2024-08-09 at 12 23 30



Qt6:
```
conan create . --version=6.7.1 -vv -o "qt/*:essential_modules=True" -o "qt/*:qtdeclarative=False"
```
Screenshot 2024-08-09 at 12 22 35

perseoGI (approved):
Good job!

ErniGH (approved):

PR #24878, opened 2024-08-09
libpointmatcher: replace existing broken version for new one and tweaked version

perseoGI:
### Summary
Changes to recipe:  **libpointmatcher/1.4.3**

#### Motivation

Old `v1.3.1` version had a graph conflict and several compile issues

#### Details

- Removed support for `v1.3.1`
- Added newest version `v1.4.3`
- Improved recipe with several protections and enhencements

---


recipes/libpointmatcher/all/conanfile.py
```diff
@@ -41,15 +51,19 @@ def config_options(self):
...
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
         self.requires("boost/1.81.0", transitive_headers=True)
-        # FIXME: eigen 3.4.0 is not compatible yet (see https://github.com/ethz-asl/libpointmatcher/issues/496)
-        self.requires("eigen/3.3.9", transitive_headers=True)
+        self.requires("eigen/3.4.0", transitive_headers=True)
         self.requires("libnabo/1.0.7")
+        self.requires("yaml-cpp/0.7.0", transitive_headers=True)
```
valgur, line 66:
```suggestion
        self.requires("yaml-cpp/0.8.0", transitive_headers=True)
```
Unless the v0.7.0 is to avoid a version conflict, v0.8.0 is newer and more common on CCI.

perseoGI: It is the required version upstream. 
I'll check on Monday if 0.8.0 works. 
If it do, we can bump!
Thanks 😃

recipes/libpointmatcher/all/conanfile.py
```diff
@@ -26,10 +32,21 @@ class LibpointmatcherConan(ConanFile):
...
+        return "17"
+
+    @property
+    def _compilers_minimum_version(self):
+        # https://github.com/norlab-ulaval/libpointmatcher/blob/1.4.3/CMakeLists.txt#L240
+        return {
+            "gcc": "9",
+            "Visual Studio": "19.0.23506",
+            "clang": "5",
+            "apple-clang": "7.0",
```
valgur, line 48:
```suggestion
            "gcc": "9",
            "clang": "5",
            "apple-clang": "7",
            "msvc": "190",
            "Visual Studio": "14",
```

perseoGI: Thanks!

valgur:
The failure due to a missing cast from `double` to `float` can probably be fixed with this patch:
```patch
--- a/pointmatcher/PointCloudGenerator.cpp
+++ b/pointmatcher/PointCloudGenerator.cpp
@@ -170,7 +170,7 @@
     {
         // Sample random values of theta and phi.
         const ScalarType phi{ 2.0f * pi * uniformDistribution(randomNumberGenerator) };
-        const ScalarType radiusSample{ radius * sqrt(uniformDistribution(randomNumberGenerator)) };
+        const ScalarType radiusSample{ radius * std::sqrt(uniformDistribution(randomNumberGenerator)) };
 
         // Pre-compute values, such as sine and cosine of phi and theta.
         const ScalarType sinPhi{ std::sin(phi) };
```
All other values in that calculation are of `ScalarType`, so the non-std `sqrt()` is likely the problem there.

AbrilRBS (approved):
Thanks @perseoGI great job :)

uilianries (approved):

PR #24899, opened 2024-08-12
[google-cloud-cpp] add v2.28.0 version

dbolduc:
**google-cloud-cpp/2.28.0**

- Add version 2.28.0 of `google-cloud-cpp`.
- Remove version 2.12.0, and clean up code only relevant to 2.12.0.
- Simplify handling of some legacy proto libraries from https://github.com/googleapis/google-cloud-cpp/issues/8022
- Drop the MSVC patch from previous versions which was fixed upstream in https://github.com/googleapis/google-cloud-cpp/pull/13777
- Individual compute_${foo}_protos libraries were removed in https://github.com/googleapis/google-cloud-cpp/pull/13836 so mirror that here.

Tested with the following on linux:
```
for v in "2.28.0" "2.19.0" "2.15.1"; do
  env CC=gcc-11 CXX=g++-11 conan create --build missing  --build-require  --version ${v} recipes/google-cloud-cpp/2.x
done
```

---


dbolduc:
@jcar87 - Hey, I could use some help. It looks like there is not a `grpc/1.54.3` package with `compiler.cppstd=17`. What should I do about this?

Is it possible to build a gRPC package in this configuration and retry the build? Should I change something in `google-cloud-cpp`.

```
ERROR: Missing prebuilt package for 'grpc/1.54.3'. You can try:
    - List all available packages using 'conan list "grpc/1.54.3:*" -r=remote'
    - Explain missing binaries: replace 'conan install ...' with 'conan graph explain ...'
    - Try to build locally from sources using the '--build=grpc/1.54.3' argument

```

saschasc:
@dbolduc  Why are there 3 versions in the CI pipeline and not just google-cloud-cpp/2.28.0?

jcar87:
> @jcar87 - Hey, I could use some help. It looks like there is not a prebuilt `grpc/1.54.3` package with `compiler.cppstd=17`. What should I do about this?
> 
> Is it possible to build a gRPC package in this configuration and retry the build? Should I change something in `google-cloud-cpp`.
> 
> ```
> ERROR: Missing prebuilt package for 'grpc/1.54.3'. You can try:
>     - List all available packages using 'conan list "grpc/1.54.3:*" -r=remote'
>     - Explain missing binaries: replace 'conan install ...' with 'conan graph explain ...'
>     - Try to build locally from sources using the '--build=grpc/1.54.3' argument
> 
> More Info at 'https://docs.conan.io/2/knowledge/faq.html#error-missing-prebuilt-package'
> ```

We're looking into this! in this case it is because of the version of abseil being used in this recipe, you could probably expand it to cover the same range as the grpc recipe: 

`"abseil/20230125.3"` -> `"abseil/[>=20230125.3 <=20230802.1]"`

> @dbolduc Why are there 3 versions in the CI pipeline and not just google-cloud-cpp/2.28.0?

the recipe file is shared amongst multiple versions, so if its modified, a new recipe revision needs to be published for the earlier versions as well

saschasc:
@jcar87 Thanks for the insight, I wasn't aware of this.

dbolduc:
> you could probably expand it to cover the same range as the grpc recipe:

Worth a try. Thanks for looking! :crossed_fingers:

recipes/google-cloud-cpp/2.x/extract_dependencies.py
```diff
@@ -35,13 +37,15 @@
 
 # Used in _generate_proto_requires(): the *.deps files are generated from
 # Bazel and contain a few targets that have incorrect names for CMake.
+#
+# This list maps onto google_cloud_cpp_load_protodeps()'s `targets_substitutions`.
```
AbrilRBS, line 41: Thanks a lot for taking the time to add some more context in here! We really appreciate it :)

AbrilRBS (approved):
Looks great to me, thanks a lot for taking the time to get this thru to the finish line!

dbolduc:
ping? we are so close...

franramirez688 (approved):

Issue #24045, opened 2024-05-20
[package] qt/6.7.0: Build failure on macOS with vulkan=true


neobrain:
### Description

Qt uses MoltenVK on macOS when Vulkan support is enabled. Currently, this configuration fails to build with Conan, since some of the Qt sources can't find the MoltenVK headers. A similar issue was reported in #6812 and was thought to have been fixed in #6835.

### Package and Environment Details

* Package Name/Version: **qt/6.7.0**
* Operating System+version: **macOS 14.5**
* Compiler+version: **clang 15**
* Conan version: **conan 2.3.1**
* Python version: **Python 3.12.3**

### Conan profile

[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=15
os=Macos

### Steps to reproduce

Due to #23684, a conanfile to force `vulkan-headers` to a specific version is required to reproduce this issue. Depending on your compiler version, you will also need to use the fixed MoltenVK recipe from #23866 ~~and the fixed Qt recipe from #24013~~.

```sh
cat <<EOF > conanfile.py
from conan import ConanFile

class BasicConanfile(ConanFile):
    name = "pkg"

    def requirements(self):
        self.requires("qt/6.7.0")
        self.requires("vulkan-headers/1.3.268.0", force=True) # See https://github.com/conan-io/conan-center-index/issues/23684
        pass

    def configure(self):
        self.options["qt"].with_vulkan = True;

        # Disable some options to speed up the build
        self.options["qt"].with_sqlite3 = False;
        self.options["qt"].with_freetype = False;
        self.options["qt"].with_harfbuzz = False;
        self.options["qt"].with_brotli = False;
        self.options["qt"].with_md4c = False;
        self.options["qt"].with_obdc = False;
        self.options["qt"].with_libpng = False;
        self.options["qt"].openssl = False;
        self.options["qt"].opengl = "no";
        pass
EOF

conan install . --build=missing
```

### Logs

Relevant excerpt of the logs:

```
FAILED: qtbase/src/plugins/platforms/cocoa/CMakeFiles/QCocoaIntegrationPlugin.dir/main.mm.o 
[...]

Issue #23684, opened 2024-04-20
[package] qt/6.7.0: Version conflict in vulkan-headers dependency with vulkan=true


neobrain:
### Description

Qt pulls in multiple versions of vulkan-headers (`qt/6.7.0`->`vulkan-loader/1.3.268.0`->`vulkan-headers/1.3.268.0` and `qt/6.7.0`->`moltenvk/1.2.2`->`vulkan-headers/1.3.239.0`, leading to a conflict:

```
Version conflict: Conflict between vulkan-headers/1.3.239.0 and vulkan-headers/1.3.268.0 in the graph.
Conflict originates from moltenvk/1.2.2
```

### Package and Environment Details

* Package Name/Version: **qt/6.7.0**
* Operating System+version: **macOS Sonoma 14.4.1**
* Compiler+version: **clang 15**
* Conan version: **conan 2.2.3**
* Python version: **Python 3.12.3**

### Conan profile

[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=15
os=Macos

### Steps to reproduce

conan install --requires qt/6.7.0 -o 'qt/*:with_vulkan'=True

### Logs

_

jcar87:
Thanks @neobrain  - the issue here in the Qt recipe is that the dependencies cannot have versions independently, as they have very strict compatibility. I've fixed this in https://github.com/conan-io/conan-center-index/pull/24903 and added a note in the recipe so that versions aren't bumped without checking first.

neobrain:
Thanks! I can confirm with the PR I no longer need any workarounds to resolve the conflicts :)

PR #24903, opened 2024-08-12
Fix locating moltenvk headers when with_vulkan=True

jcar87:
### Summary
Changes to recipe:  qt/6

* Fix issue on macOS where building qt fails when `with_vulkan=True` because it does not find the moltenvk headers
* Record direct dependency on vulkan-headers (these are searched _directly_ by Qt, see [here](https://github.com/qt/qtbase/blob/v6.7.1/cmake/FindWrapVulkanHeaders.cmake))
* Record `transitive_headers` for the vulkan-headers when `with_vulkan=True`, these are used here: https://github.com/qt/qtbase/blob/v6.7.1/src/gui/vulkan/qvulkaninstance.h#L19-L21 - if these are not found, including `<QVulkanWindow>` does not properly include the vulkan headers, leading to include errors
* Fix version conflict when `with_vulkan=True` - moltenvk and vulkan have strict version requirements and need to be paired with the right versions

#### Motivation
* Fix version conflict
* Fix compiler error:

```
In file included from /xxx/.conan2/p/qt0cc42066e0980/s/src/qtbase/src/plugins/platforms/cocoa/qcocoaapplicationdelegate.mm:41:
In file included from /xxx/.conan2/p/qt0cc42066e0980/s/src/qtbase/src/plugins/platforms/cocoa/qcocoaintegration.h:8:
/xxx.conan2/p/qt0cc42066e0980/s/src/qtbase/src/plugins/platforms/cocoa/qcocoawindow.h:19:10: fatal error: 'MoltenVK/mvk_vulkan.h' file not found
#include <MoltenVK/mvk_vulkan.h>
         ^~~~~~~~~~~~~~~~~~~~~~~
```

#### Details
* patch the qt cmake file that handles vulkan headers, to propagate the include path of moltenvk as well

Close https://github.com/conan-io/conan-center-index/issues/24045
Close https://github.com/conan-io/conan-center-index/issues/23684

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -366,7 +366,11 @@ class QtConan(ConanFile):
         if self.options.with_pcre2:
             self.requires("pcre2/10.42")
         if self.options.get_safe("with_vulkan"):
-            self.requires("vulkan-loader/1.3.268.0")
+            # Note: the versions of vulkan-loader and moltenvk
+            #       must be exactly part of the same Vulkan SDK version
+            #       do not update either without checking both
+            #       require exactly the same version of vulkan-headers
+            self.requires("vulkan-loader/1.3.239.0")
             if is_apple_os(self):
                 self.requires("moltenvk/1.2.2")
         if self.options.with_glib:
@@ -720,6 +724,13 @@ class QtConan(ConanFile):
                         os.path.join(self.source_folder, "qtbase", "cmake", "QtAutoDetect.cmake" if Version(self.version) < "6.6.2" else "QtAutoDetectHelpers.cmake"),
                         "qt_auto_detect_vcpkg()",
                         "# qt_auto_detect_vcpkg()")
+
+        # Handle locating moltenvk headers when vulkan is enabled on macOS
+        replace_in_file(self, os.path.join(self.source_folder, "qtbase", "cmake", "FindWrapVulkanHeaders.cmake"),
+        "if(APPLE)", "if(APPLE)\n"
+                    " find_package(moltenvk REQUIRED QUIET)\n"
+                    " target_include_directories(WrapVulkanHeaders::WrapVulkanHeaders INTERFACE ${moltenvk_INCLUDE_DIR})"
+        )
 
     def _xplatform(self):
         if self.settings.os == "Linux":
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -304,9 +304,6 @@ class QtConan(ConanFile):
         if Version(self.version) >= "6.6.1" and self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "13.1":
             raise ConanInvalidConfiguration("apple-clang >= 13.1 is required by qt >= 6.6.1 cf QTBUG-119490")
 
-        if self.settings.os == "Macos" and self.dependencies["double-conversion"].options.shared:
-            raise ConanInvalidConfiguration("Test recipe fails because of Macos' SIP. Contributions are welcome.")
-
         if self.options.get_safe("qtwebengine"):
             if not self.options.shared:
                 raise ConanInvalidConfiguration("Static builds of Qt WebEngine are not supported")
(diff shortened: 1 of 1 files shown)


jcar87:
Tested on macOS and we now get a successful build, but have not tested this on the consumer side yet - so this may not be fully sufficient

jcar87:
> Tested on macOS and we now get a successful build, but have not tested this on the consumer side yet - so this may not be fully sufficient

have managed to test this against the sample vulkan project here: https://github.com/qt/qtbase/tree/v6.7.1/examples/vulkan/hellovulkancubes


Screenshot 2024-08-12 at 14 24 28

so far have only managed to make it work with the shared variant, built with the following options:
```
-o "qt/*:with_vulkan=True" -o "*:shared=True" -o "spirv-tools/*:shared=False" "-o spirv-cross/*:shared=False" 
```

otherwise the application might build but not run correctly (for the static version, even if I link the plugins statically, it fails to initialise vulkan, that would require further investigation).

jcar87:
@nebrain@neobrain - apologies it took so long to fix this - if you have the chance to test these fixes, please let us know!

recipes/qt/6.x.x/conanfile.py
```diff
@@ -304,9 +304,6 @@ def validate(self):
         if Version(self.version) >= "6.6.1" and self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "13.1":
             raise ConanInvalidConfiguration("apple-clang >= 13.1 is required by qt >= 6.6.1 cf QTBUG-119490")
 
-        if self.settings.os == "Macos" and self.dependencies["double-conversion"].options.shared:
```
jcar87, line 307: This restriction is artificial and not related to Qt itself, but rather Conan or Conan Center - lifting it. Have managed to reproduce the original problems, if issues are reported will propose a more robust fix (otherwise, this is not currently causing issues on CI).

AbrilRBS (approved):
👍

neobrain:
> @nebrain - apologies it took so long to fix this - if you have the chance to test these fixes, please let us know!

Thanks for working on this! Indeed Qt builds fine without any workarounds for me now.

Sadly I haven't been able to get my actual application running though, similarly to the example you tried. It's failing to pick up the VkInstance extensions required to use MoltenVK. I haven't been able to look more deeply into it yet however. With my custom Qt recipe I previously got it working while also pointing `VK_ICD_FILENAMES=<path to MoltenVK_icd.json>` and `QT_VULKAN_LIB=<path to libMoltenVK.dylib>`, but this doesn't seem to work for this PR.

ErniGH (approved):

jcar87:
> > @nebrain - apologies it took so long to fix this - if you have the chance to test these fixes, please let us know!
> 
> Thanks for working on this! Indeed Qt builds fine without any workarounds for me now.
> 
> Sadly I haven't been able to get my actual application running though, similarly to the example you tried. It's failing to pick up the VkInstance extensions required to use MoltenVK. I haven't been able to look more deeply into it yet however. With my custom Qt recipe I previously got it working while also pointing `VK_ICD_FILENAMES=<path to MoltenVK_icd.json>` and `QT_VULKAN_LIB=<path to libMoltenVK.dylib>`, but this doesn't seem to work for this PR.

Thanks for trying this out @neobrain!
If there is a reproducible example to report, we can look into this.

The example from https://github.com/qt/qtbase/tree/v6.7.1/examples/vulkan/hellovulkancubes does seem to work without modifications to either the example or the recipe, but only when Qt is shared. 

Are you using shared Qt or static builds? Any info would help, and we can open an issue to track this, thanks!

neobrain:
> Are you using shared Qt or static builds? Any info would help, and we can open an issue to track this, thanks!

Good news: I've finally got around to clean up my application code and look into this again, and it seems that everything is working fine now!

For reference, my setup is the following:
* Shared Qt with Vulkan (i.e. MoltenVK) enabled
* `-DVULKAN_HPP_DISPATCH_LOADER_DYNAMIC=0` used globally
* `-DVK_USE_PLATFORM_MACOS_MVK=1` used globally

To run the application, conanrun.sh must still be sourced (but I don't need to override VK_ICD_FILENAMES and QT_VULKAN_LIB like before).

jcar87:
Hi @neobrain !! Happy to hear it is working now, thanks for taking the time to test this out and report back!

PR #24928, opened 2024-08-13
Fix linking problems when using CUDA with llama-cpp in static

czoido:
This PR adds a build module for consumers to add the CUDA targets when using CUDA with llama-cpp in static.

### DIFF
--- a/recipes/llama-cpp/all/cmake/llama-cpp-cuda-static.cmake
+++ b/recipes/llama-cpp/all/cmake/llama-cpp-cuda-static.cmake
@@ -1,4 +1,4 @@
-find_package(CUDAToolkit REQUIRED)
+find_dependency(CUDAToolkit REQUIRED)
 if (WIN32)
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):
Good catch :)

recipes/llama-cpp/all/cmake/llama-cpp-cuda-static.cmake
```diff
@@ -0,0 +1,7 @@
+find_package(CUDAToolkit REQUIRED)
```
valgur, line 1:
```suggestion
include(CMakeFindDependencyMacro)
find_dependency(CUDAToolkit REQUIRED)
```
`find_dependency()` is preferable to `find_package()` in CMake config modules: https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html

AbrilRBS (dismissed):

AbrilRBS (approved):

valgur (approved):

PR #24966, opened 2024-08-16
cpython: drop deprecated nis module, add cross-compilation support

valgur:
### Summary
Changes to recipe:  **cpython/[*]**

#### Motivation
CPython from Conan fails to link as a dependency on Ubuntu 24.04 due to an invalid `nsl` system dep at https://github.com/conan-io/conan-center-index/blob/ab3333f3ba1baf38c3a11fceff73d1db385bd317/recipes/cpython/all/conanfile.py#L886

The `nsl` library is no longer included with glibc since 2018 or so and the legacy `nis` module in CPython has been deprecated since v3.11 and is to be removed in v3.13 as a result.

https://bugs.python.org/issue32521
https://docs.python.org/3/library/nis.html

#### Details
I dropped the unsupported `with_nis` option and added patches to ensure that the `nis` module and the `nsl` library are consistently ignored, even if it is available on older Linux systems.

While this nominally a breaking change, anyone still relying on the `nis` module in 2024 seems like a quite remote possibility. There has been no visible interest in adding a package for `libnis` on CCI either. I suppose backwards compatibility could be restored if someone complains?

---


valgur:
The patches now work correctly and a `'nis.cpython-311-x86_64-linux-gnu.so' links to system library 'nsl' but it is not in cpp_info.system_libs` warning is no longer produced. :slightly_smiling_face:

recipes/cpython/all/patches/3.10/3.10.14-0002-disable-nis.patch
```diff
@@ -0,0 +1,11 @@
+--- a/setup.py
++++ b/setup.py
+@@ -2594,7 +2594,7 @@
+             ))
+ 
+     def detect_nis(self):
+-        if MS_WINDOWS or CYGWIN or HOST_PLATFORM == 'qnx6':
++        if True:
```
Ahajha, line 8: I'm guessing this is to prevent all installations from detecting nis, instead of having some builds with and some without?

Ahajha (approved):

recipes/cpython/all/patches/3.10/3.10.14-0002-disable-nis.patch
valgur: Yes, exactly.

Issue #20574, opened 2023-10-15
[package] qt/6.6.0: missing __qt_internal_cmake_apple_support_files_path


tamaskenezlego:
### Description

In the original Qt installation `lib/CMake/Qt6/Qt6Config.cmake` sets the `__qt_internal_cmake_apple_support_files_path` CMake variable. It's used in `QtPublicAppleHelpers.cmake/_qt_internal_copy_info_plist()` to locate the `lib/CMake/Qt6/macos/Info.plist.app.in` file.

In conan-built Qt `Qt6Config.cmake` is conan-generated and does not set `__qt_internal_cmake_apple_support_files_path`. 

The result is that configuring a Qt/MacOS bundle executable (`qt_add_executable(... MACOSX_BUNDLE ...)`) prints the following error:

```
CMake Error: File /Info.plist.app.in does not exist.
```

### Package and Environment Details

* Package Name/Version: **qt/6.6.0**
* Operating System+version: **MacOS**
* Conan version: **conan 1.61.0**

### Conan profile

[settings]
os=Macos
os_build=Macos
arch=armv8
arch_build=armv8
compiler=apple-clang
compiler.version=15
compiler.libcxx=libc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

Run cmake-configure on a helloworld-qt application (MacOS) where the executable is added with:

```
qt_add_executable(target_name MACOSX_BUNDLE main.cpp)
```

### Logs

The conan-log is not relevant.

DoomHammer:
Still present in Qt 6.7.1

DoomHammer:
@jcar87 @ericLemanissier do you know where is this generated? I could try to fix it, but looking at conanfile.py I have no idea where to start.

uilianries:
Hello @tamaskenezlego and @DoomHammer !! The PR #24642 has been merged and a hotfix is not available in the Conan Center. Please, update your local copy:

    conan install -r conancenter --update --requires-"qt/<version>"

DoomHammer:
Thanks!

For some reason, though, I'm still getting:

```
CMake Error: File /Users/doomhammer/.conan2/p/b/qtd17264e33cd40/p/res/macos/Info.plist.app.in does not exist.
```

`/Users/doomhammer/.conan2/p/b/qtd17264e33cd40/p/lib/cmake/Qt6/macos/Info.plist.app.in` is, however, present.

PR #24971, opened 2024-08-16
Fix Mac installation

DoomHammer:
### Summary
Changes to recipe:  **qt/6.x.x**

#### Motivation
Previous PR: https://github.com/conan-io/conan-center-index/pull/24642 did not fix this issue https://github.com/conan-io/conan-center-index/issues/20574

The path is still wrong and doesn't work with downstream projects.

#### Details

* Set the correct path to the `Info.plist.app.in` folder.
* Removed the previous copy of this file.
* Added extra simple check to test_package

---

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -851,8 +851,6 @@ class QtConan(ConanFile):
         if self.settings.os == "Macos":
             save(self, ".qmake.stash", "")
             save(self, ".qmake.super", "")
-            copy(self, "Info.plist.app.in", src=os.path.join(self.package_folder, "lib", "cmake", "Qt6", "macos"),
-                                            dst=os.path.join(self.package_folder, "res", "macos"))
         cmake = CMake(self)
         cmake.install()
         copy(self, "*LICENSE*", self.source_folder, os.path.join(self.package_folder, "licenses"),
@@ -888,7 +886,7 @@ class QtConan(ConanFile):
         filecontents += f"set(QT_VERSION_MINOR {ver.minor})\n"
         filecontents += f"set(QT_VERSION_PATCH {ver.patch})\n"
         if self.settings.os == "Macos":
-            filecontents += f'set(__qt_internal_cmake_apple_support_files_path "${{CMAKE_CURRENT_LIST_DIR}}/../../../lib/cmake/Qt6/macos")\n'
+            filecontents += 'set(__qt_internal_cmake_apple_support_files_path "${CMAKE_CURRENT_LIST_DIR}/../../../lib/cmake/Qt6/macos")\n'
         targets = ["moc", "rcc", "tracegen", "cmake_automoc_parser", "qlalr", "qmake"]
         if self.options.with_dbus:
             targets.extend(["qdbuscpp2xml", "qdbusxml2cpp"])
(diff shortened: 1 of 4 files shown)


franramirez688 (commented):
Hi @DoomHammer 

Thanks for this PR 👏 

I just saw the PR where this line was introduced: https://github.com/conan-io/conan-center-index/pull/24642/files#diff-c548829da9d38ae15a8cb4447140bc50ac3f38250d1f348decd9569dc3298b24R842-R843

I was wondering if we could also remove the lines from the `package()` method:

```python
            copy(self, "Info.plist.app.in", src=os.path.join(self.package_folder, "lib", "cmake", "Qt6", "macos"),
                                            dst=os.path.join(self.package_folder, "res", "macos"))
```
I just tried it locally and it worked by applying the diff suggested by @uilianries in the PR (https://github.com/conan-io/conan-center-index/pull/24642) changelog.

DoomHammer:
I'm not sure I understand, can you rephrase?

uilianries:
@DoomHammer Could you please specify the exactly error that you are trying to fix, with a reproducible scenario? I see you comment [here](https://github.com/conan-io/conan-center-index/issues/20574#issuecomment-2255647141), but still not enough to have your scenario, we need more details like: your Conan profile, Conan version and full build log, including command line used.

The PR #24642 was exactly to fix the MACOSX BUNDLE with CMake, so I'm afraid your PR will bring back that same bug.

franramirez688:
> I'm not sure I understand, can you rephrase?

@DoomHammer Yes, let me rephrase it. This [PR](https://github.com/conan-io/conan-center-index/pull/24642) from @uilianries introduced this piece of code:

```python
    def package(self):
        if self.settings.os == "Macos":
            # ...
            copy(self, "Info.plist.app.in", src=os.path.join(self.package_folder, "lib", "cmake", "Qt6", "macos"),
                                            dst=os.path.join(self.package_folder, "res", "macos"))
        # ...
        if self.settings.os == "Macos":
            filecontents += f'set(__qt_internal_cmake_apple_support_files_path 
                                              "${{CMAKE_CURRENT_LIST_DIR}}/../../../res/macos")\n'
```
Then, what I'm saying is to change it as you're suggesting here, but also remove the `copy()` line: 
```python
    def package(self):
        if self.settings.os == "Macos":
            # ...
            # copy(self, "Info.plist.app.in", src=os.path.join(self.package_folder, "lib", "cmake", "Qt6", "macos"),
            #                                dst=os.path.join(self.package_folder, "res", "macos"))
        # ...
        if self.settings.os == "Macos":
            filecontents += f'set(__qt_internal_cmake_apple_support_files_path 
                                              "${{CMAKE_CURRENT_LIST_DIR}}/../../../lib/cmake/Qt6/macos")\n'
```

> The PR https://github.com/conan-io/conan-center-index/pull/24642 was exactly to fix the MACOSX BUNDLE with CMake, so I'm afraid your PR will bring back that same bug.

@uilianries I tested it locally and that `res` folder is not there. I applied the changes above and it's working fine (modifying the test package for sure).

franramirez688 (commented):
@DoomHammer 

I added a commit adding some extra checks to the `test_package` to test this special case and removing that useless `copy`. Waiting for @jcar87 feedback before approving.

jcar87 (approved):

czoido (approved):

franramirez688 (approved):

PR #25045, opened 2024-08-25
OpenJPH: Add recipe for new library

irieger:
### Summary
Changes to recipe:  **openjph/0.15.0**

#### Motivation
I'm currently looking into a multitude of higher-end image compression algorithms and for this I started to look into OpenJPH which isn't yet available through Conan. It is an open source implementation of the High Throughput extension for JPEG 2000. So to try & use it, I created a recipe I want to contribute back.

#### Details
Adding the library.

---


recipes/openjph/all/conandata.yml
```diff
@@ -2,3 +2,8 @@ sources:
   "0.15.0":
     url: "https://github.com/aous72/OpenJPH/archive/0.15.0.tar.gz"
     sha256: "36601fbd3b4e1fe54eef5e6fa51ac0eca7be94b2a3d7c0967e3c8da66687ff2c"
+patches:
+  "0.15.0":
+    - patch_file: "patches/0.15.0_cmake-remove-libversion-suffix-win.patch"
+      patch_description: "Ensure that the library isn't suffixed with the version name for better compatibility"
```
SpaceIm, line 8: You should follow upstream convention instead of patching the name. Add suffix in package info by relying on is_msvc.

irieger: Thank you for the feedback. Adapted it

recipes/openjph/all/conanfile.py
```diff
@@ -0,0 +1,106 @@
...
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "openjph")
+        self.cpp_info.set_property("cmake_target_name", "openjph::openjph")
+        self.cpp_info.set_property("pkg_config_name", "openjph")
+
+        version_suffix = ""
+        if is_msvc(self):
+            v = Version(self.version)
+            version_suffix = f".{v.major}.{v.minor}"
+            print('Lib Version', version_suffix)
```
AbrilRBS, line 100:
```suggestion
```

recipes/openjph/all/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.options.with_tiff:
+            self.requires("libtiff/4.6.0")
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, 14)
```
AbrilRBS, line 52: As per https://github.com/aous72/OpenJPH/blob/master/CMakeLists.txt#L95, this should be 11 for the library itself right?

Also, we should remove the hardcoding of `CMAKE_CXX_STANDARD` in the following line too

recipes/openjph/all/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_tiff": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_tiff": True,
```
AbrilRBS, line 29: would an option for `OJPH_BUILD_EXECUTABLES` be useful here?

recipes/openjph/all/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+    description = "Open-source implementation of JPEG2000 Part-15 (or JPH or HTJ2K)"
+    license = "BSD-2-Clause"
+    homepage = "https://github.com/aous72/OpenJPH"
+    url = "https://github.com/conan-io/conan-center-index"
+    topics = ("ht-j2k", "jpeg2000", "jp2", "openjph", "image", "multimedia", "format", "graphics")
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_tiff": [True, False],
```
AbrilRBS, line 24: What happens when a CI that supports some SIMD options builds this with the default options here for all extensions https://github.com/aous72/OpenJPH/blob/master/CMakeLists.txt#L29, but a consumer without that support tries to run the same package? The package id does not change

recipes/openjph/all/conanfile.py
```diff
@@ -0,0 +1,105 @@
...
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, 14)
+
+        if self.settings.compiler == "gcc" and \
+            Version(self.settings.compiler.version) < "6.0":
+            raise ConanInvalidConfiguration(f"{self.ref} requires gcc >= 6.0")
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.11 <4]")
```
AbrilRBS, line 59: If cmake < 3.15 is needed, we don't add a tool_requires here, and instead rely on users already having them on the system (Or adding it thru profiles in the worst case scenario)

```suggestion
```

AbrilRBS (commented):
Thanks! Some comments after a first look :)

recipes/openjph/all/conanfile.py
irieger: Just added an option for it and also the third executable that isn't build by default but via an option.

recipes/openjph/all/conanfile.py
irieger: Oh, I read over the comment to be honest. Just fixed it. Would it be good practice to make the tests also avaialble? To my understanding it isn't really normally as conan is for providing libraries, not directly develop them from CCI recipes or do I see that wrong?

irieger:
@AbrilRBS Could you have a look again? Maybe you have an idea for the failure.

I looked a while at the failing build and can't make any sense of. The code works with Conan 2 on macOS but not on Conan 1. Otherwise compiler version, C++ standard etc. match so I don't really get what might cause the failure. Also looking at the code it looks fine to me so I'm not having any clue why it fails.

recipes/openjph/all/patches/0.15.0_cmake-cxx-standard.patch
```diff
@@ -0,0 +1,26 @@
...
+diff --git src/apps/ojph_stream_expand/CMakeLists.txt src/apps/ojph_stream_expand/CMakeLists.txt
+index 5aea342..e21bdd9 100644
+--- src/apps/ojph_stream_expand/CMakeLists.txt
++++ src/apps/ojph_stream_expand/CMakeLists.txt
+@@ -5,7 +5,7 @@ if (OJPH_BUILD_STREAM_EXPAND)
+ 
+   include_directories(../common)
+   include_directories(../../core/common)
+-  set(CMAKE_CXX_STANDARD 14)
++  #set(CMAKE_CXX_STANDARD 14)
```
jcar87, line 23: @AbrilRBS, @irieger 

The issue in Conan 1.x on macOS is happening because the Conan 1 profiles do not define `compiler.cppstd`, and the `appel-clang` compiler defaults to `gnu98` on the command line.
Fully commenting out this line will cause these Conan 1 builds to _not_ pass any standard whatsoever, and default to whatever the compiler defaults to - clearly something 11 or greter on all others except for apple clang.

The recipe probably needs something like
``` 
if not self.settings.get_safe("compiler.cppstd"):
    tc.cache_variables["CMAKE_CXX_STANDARD"] = xx # 11 or 14 depending on some condition
```

alternatively maybe tweaking the patch to make it a cache variable _might_ work :
* default to that value in Conan 1.x when cppstd is not defined
* obey the value in the `conan_toolchain.cmake` if provided, ignoring the hardcoded value in the cmakelists

but this latter alternative would need to be tested

irieger: Ok, your first suggestion did work.

irieger: Thank you!

AbrilRBS (approved):
Great job, thanks a lot for taking the time to also add the new version, and sorry about the delay! :)

PR #25046, opened 2024-08-26
fix wxwidget build error with build_type=RelWithDebInfo in Windows

xyz1001:
### Summary
Changes to recipe:  **wxwidget/3.2.5**

#### Motivation

Error occurred when build wxwidgets with build_type=RelWithDebInfo. The reason is that wxwidgets limits `CMAKE_CONFIGURATION_TYPES` to `Debug;Release` if this variable is not defined.
See https://github.com/wxWidgets/wxWidgets/issues/24779

#### Details

Define `CMAKE_CONFIGURATION_TYPES` to `Debug;Release;RelWithDebInfo`.

---


valgur (approved):

recipes/wxwidgets/all/conanfile.py
```diff
@@ -240,6 +240,7 @@ def generate(self):
         tc.variables["wxUSE_URL"] = self.options.protocol
         tc.variables["wxUSE_PROTOCOL"] = self.options.protocol
         tc.variables["wxUSE_FS_INET"] = self.options.fs_inet
+        tc.variables["CMAKE_CONFIGURATION_TYPES"] = "Debug;Release;RelWithDebInfo"
```
jcar87, line 243: I would err on the side of caution and leave this variable _undefined_ when using a single config generator. Defining it here like this will do it for _all_ generators (single or multi config), rather than only for multiconfig as intended here: https://github.com/wxWidgets/wxWidgets/blob/master/CMakeLists.txt#L12-L18
`CMAKE_CONFIGURATION_TYPES` is one of those variables that projects may check for "is this a multiconfig generator" because it is assumed to _only_ be defined for those. small risk, but rather not take the chance

I think this is one of those instances where a patch may be preferable.

ErniGH, line 243:
```suggestion
        tc.cache_variables["CMAKE_CONFIGURATION_TYPES"] = "Debug;Release;RelWithDebInfo;MinSizeRel"
```

AbrilRBS (dismissed):

jcar87 (approved):

jcar87:
Thanks for reporting @xyz1001  and for fixing, sorry it took a while to get to this.

PR #25076, opened 2024-08-29
zlib: Fix shared library on Windows MSVC

alessiosacco:
### Summary
Changes to recipe:  **zlib/1.3.1**

#### Motivation
When option.shared = True and compiler is MSVC, the CMakeLists.txt creates an archive called zlib.lib, but the dll is called zdll.dll.

This configuration is not standard and is not supported by cmakedeps_macro.cmake module when using find_package(ZLIB).

The source patch now creates zlib.lib and zlib.dll.

#### Details
Tested using Windows, MSVC v142 and v143, debug and release with option.shared both false and true, Debian Linux GCC 10.

---


jcar87 (changes requested):
Renaming the DLL on Windows will break compatibility with pre-existing binaries and cause widespread breakages so I'm afraid we cannot accept this change.

When we say the name is "not standard" why do we mean by that? There is no prerequisite whatsoever for the lib and DLL files to have the same "name", as far as I know, even if that's typically the expected.

alessiosacco:
The cmakedeps_macro.cmake generated by CMakeDeps, in the function `conan_package_library_targets` is not able to find the shared library because it expects the name of the implib and the runtime library to be the same, which is not what happens when zlib is compiled under windows using the old conanfile and patch.
The function uses the CMake `find_library` function with the name argument `_LIBRARY_NAME ` that is extracted from the implib filename in order to search for the runtime library. Since the runtime does not have the same name as the implib a STATUS message is shown during CMake configure: `message(STATUS "Cannot locate shared library: ${_LIBRARY_NAME}")`
The problem is also reflected in how the target is created, because it will be an `UNKNOWN IMPORTED` library instead of a `SHARED IMPORTED` library.
With the current `self.cpp_info` specified in package_info there is no way to specify that the runtime library has a different name from the implib, see [docs](https://docs.conan.io/2/reference/conanfile/methods/package_info.html#cpp-info-library-and-build-information).

jcar87:
Hi @alessiosacco - thanks for providing more details.

I have opened this issue in the Conan repository: https://github.com/conan-io/conan/issues/16926 - as this is general and indeed a limitation. 
The "warning" being a warning is correct - CMake only needs the import library `.lib`  - although the logic in CMakeDeps that sets the imported target type to `UNKNOWN`  instead of `SHARED` as a result of this, may have other side effects (see https://github.com/conan-io/conan/issues/12654) 

Are you experiencing any errors or issues other than the warning?

That said, I'm afraid we can't accept this PR as it would cause breakages across the board on Windows when using shared libraries. The Conan model takes into account that if the version hasn't changed and a library is depended on as a shared library, downstream dependees don't need to be rebuilt. If ABI or the shared library names change _for the same version_ in a new recipe revision, existing binaries of dependees that rely on zlib shared would not be rebuilt BUT they would stop working at runtime as the new package does not provide the DLL name expected.  A _possible_ workaround would be to simply duplicate the DLLs in the new revision when making this sort of change (provide the old dll name and the new dll name). However we tend to be very conservative here: we should keep the exact names used upstream, unless there's as a very strong and compelling justification. Arguably the zlib recipe should have never renamed the DLL in the existing patch in the same place - would need to check the historical of the recipe to understand why this was done.

alessiosacco:
Hi @jcar87,
I have run into this issue when trying to upgrade the version of zlib that is packaged inside the project I am building. The zlib library was previously prebuilt and copyied to the install folder of shared library that I am distributing. The prebuilt zlib was installed using the CMake install using `install(IMPORTED_RUNTIME_ARTIFACTS)`. CMake requires the targets that are used for this call to declare the `IMPORTED_LOCATION_<CONFIG>` property and the targets need to be `IMPORTED SHARED` for this to work in the case of `RUNTIME` in Windows and `LIBRARY` in Linux. The libraries that are provided by the cmakedeps_macros module do not satisfy these requirements because the .dll is not declared in the property and the target is `IMPORTED UNKNOWN`. I will follow the issue you have opened.

jcar87:
Hi @alessiosacco - thank you for providing additional information, this is very useful. Indeed this would be a limitation as a result of https://github.com/conan-io/conan/issues/16926.

For getting CMake to bring/copy the DLLs you depend on, you may have better luck with the following example, where we can try for CMake to recursively find the DLLs (and it should find them, as it relies on information external to the targets)
https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#conan-runtime-lib-dirs

jcar87:
Closing since we cannot rename a shared library without causing widespread breakages.

In the future, the newer version of CMakeDeps (https://github.com/conan-io/conan/pull/17296) can handle this - by implementing the `location` and `link_location` as separate attributes in `package_info()` 

We can assist with `IMPORTED_RUNTIME_ARTIFACTS` or equivalent if necessary

PR #25090, opened 2024-08-31
gtk: migrate, split into `3.x` and `4.x` subfolders, add all deps and components, introspection support

valgur:
### Summary
Changes to recipe:  **gtk/[*]**

#### Motivation
The GTK library is a quite significant one, but is yet to be migrated to Conan 2.x.

The recipe details for GTK 4 have diverged from GTK 3 enough that squeezing them into a single recipe file hurts more than helps. Even more so, GTK 3 seems to be more or less frozen on version 3.24.x while 4.x is getting frequent updates.

#### Details
This PR:
- Migrates the GTK recipe to 2.x.
- Adds the latest 3.x and 4.x versions and drops old ones.
- Carefully reproduces the .pc components and their details so they match the logic in the Meson build files. Avoids overlinking by using the exact dependency components.
- Adds an option for building the Broadway (HTML5) GTK backend.
- Adds support for all optional dependencies, which is actually not that many.
  - Dependencies that are not yet available on CCI or are currently incompatible (GStreamer) are provided as system dependencies.
- Deprecates the `with_pango` option and sets it to always enabled. The Pango dependency is mostly assumed by the project to be available and disentangling it did not seem to be worth the effort.
- Deprecates `with_cloudproviders` which never had any effect.
- Added optional `gobject-introspection` support.

This PR depends on fixes and updates to a number of other recipes:
- #25015
- #25091
- #25092
- #25093
- #25094
- #25095
- #25096
- #25097

All of the fixes have been already applied in my personal fork, if you're interested in testing or using the recipe before all of the related ones have been merged: https://github.com/valgur/conan-center-index

---
- [x] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
- [x] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)
- [x] Tested locally with at least one configuration using a recent version of Conan

PR #25092, opened 2024-08-31
pango: add v1.54.0, enable libxft, freetype and fontconfig  by default, add introspection support

valgur:
### Summary
Changes to recipe:  **pango/[*]**

#### Motivation
Related to the GTK PR #25090.

- v1.54.0 is required by the latest GTK 4.x release.
- The `pangoxft` component is required by the GTK X11 backend.
- Freetype and fontconfig are already enabled by default in Cairo, which Pango depends on, so disabling them does not have much benefit.
- Pango introspection info is required to build the GTK introspection info in GTK 4.x.

#### Details
Closes #22970 - merged the changes from it.

---


recipes/pango/all/conanfile.py
```diff
@@ -31,50 +31,48 @@ class PangoConan(ConanFile):
...
+            self.options.with_freetype = False
+            self.options.with_fontconfig = False
 
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
         self.settings.rm_safe("compiler.libcxx")
         self.settings.rm_safe("compiler.cppstd")
+        self.options["cairo"].with_freetype = self.options.with_freetype
+        self.options["cairo"].with_fontconfig = self.options.with_fontconfig
```
ErniGH, line 62: Could you explain why are you adding this? In the previous revision it was working with cairo using both `freetype` and `fontconfig` enabled. 

Do you have any reference from upstream? 

I am trying to fix the missing binaries issues :(

ErniGH: Well, how do you feel about reverting the change to maintain the previous version from Conan Center Index? The idea is to avoid the missing binaries issues associated with this change.

valgur: > The idea is to avoid the missing binaries issues associated with this change.

Both the `cairo` and `pango` would have the `with_freetype` and `with_fontconfig` options set to `True` as a result of this PR, so I don't see how it would cause issues in the default case at least.

I'll just revert it since I don't care about the change too much.

AbrilRBS:
@valgur we have made some changes to the recipe based on the latest meson.build. I'm actually not quite sure if all of them represent the old versions too, we'll need to check, but feel free to take a look at the changes, let me know what you think! 

The only issue I was not able to track down is that if I call `conan create . --version=1.54.0 -b=missing  -o="&:with_freetype=False"` , I still get `Run-time dependency freetype2 found: YES 26.1.20` in the meson output, which is suprising to me, but maybe you have a clearer idea of what's going on?

valgur:
Thank you so much @AbrilRBS for investing the tangled mess of Freetype and FontConfig options in this project! The changes look good to me.
I only converted the meson.build references to full URLs, since that also includes the version of the file it is referring to, which is difficult to figure out at a later date.

AbrilRBS (approved):

ErniGH (approved):

PR #25098, opened 2024-08-31
sdl: handle pulseaudio not having 'shared' option in v17.0

valgur:
### Summary
Changes to recipe:  **sdl/[*]**

#### Motivation
Currently `self.dependencies["pulseaudio"].options.shared` breaks when used with `pulseaudio/17.0` on Windows.

#### Details

---


uilianries (commented):
Hello @valgur! Thank you for suggesting this PR over SDL project. I did a small investigation, and something looks been missing to understand your scenario. Here is my conclusion:

The Pulseaudio 17.0, packaged by Conan, uses Meson to be built only and only provides shared libraries, so the option `shared` does not exist in that recipe. Also, the upstream no longer gives autotools support like previous versions, and it only builds shared libs.

References:

- https://github.com/conan-io/conan-center-index/blob/master/recipes/pulseaudio/meson/conanfile.py#L22
- https://gitlab.freedesktop.org/search?search=library(&nav_source=navbar&project_id=810&group_id=1283&search_code=true&repository_ref=v17.0

In previous versions supported in CCI, they have the shared option and Autotools:

- https://github.com/conan-io/conan-center-index/blob/master/recipes/pulseaudio/all/conanfile.py#L25

So the case of having shared option from pulseaudio available or not, at first sight, makes sense.

------

However, this PR targets Windows OS, which is **not** supported by the pulseaudio recipe:

- https://github.com/conan-io/conan-center-index/blob/master/recipes/pulseaudio/all/conanfile.py#L79
- https://github.com/conan-io/conan-center-index/blob/master/recipes/pulseaudio/meson/conanfile.py#L68

Plus, the SDL recipe does not allow using Pulseaudio in case not building for Linux:

- https://github.com/conan-io/conan-center-index/blob/master/recipes/sdl/all/conanfile.py#L116

Finally, the current SDL recipe uses **only** pulseaudio 14.2 only, which has the option shared available always:

```
$ conan list -r conancenter "pulseaudio/14.2#latest:*" -fo "&:shared=True"
Connecting to remote 'conancenter' anonymously
conancenter
  pulseaudio
    pulseaudio/14.2
      revisions
        30afeff38a7770cb9cda17a66da49a19 (2024-06-28 06:54:18 UTC)
          packages
            e98bf87c38f17cdb2bf1b257f0172ac1466c38a0
              info
                settings
                  arch: x86_64
                  build_type: Release
                  compiler: gcc
                  compiler.version: 11
                  os: Linux
                options
                  shared: True
                  with_alsa: True
                  with_dbus: False
                  with_glib: False
                  with_openssl: True
                  with_x11: True
                requires
                  libalsa/1.2.Z
                  libcap/2.69.Z
                  libiconv/1.17.Z
                  libsndfile/1.2.Z
                  libtool/2.4.Z
                  openssl/3.3.Z
                  xorg/system
            0c52722c38d6292e6bfd6802692e38b12dfde536
              info
                settings
                  arch: x86_64
                  build_type: Release
                  compiler: gcc
                  compiler.version: 11
                  os: Linux
                options
                  shared: True
                  with_alsa: True
                  with_dbus: False
                  with_glib: False
                  with_openssl: True
                  with_x11: True
                requires
                  libalsa/1.2.Z
                  libcap/2.69.Z
                  libiconv/1.17.Z
                  libsndfile/1.2.Z
                  libtool/2.4.Z
                  openssl/3.2.Z
                  xorg/system

```

So, before this "bug" occurs, Conan will raise a `ConanInvalidConfiguration` on Windows, not allowing being built. The same reflects the current build, where pulseaudio is not active when building SDL:

https://c3i.jfrog.io/c3i/misc-v2/logs/pr/25098/1-windows-msvc/sdl/2.30.5//8e40a03e74608ac8499063f4621a06fc3d479112-build.txt

> --   SDL_PULSEAUDIO              (Wanted: OFF): OFF
> --   SDL_PULSEAUDIO_SHARED       (Wanted: OFF): OFF

And for Linux it should not be an issue, unless the pulseaudio version is overridden to 17.0.

Thus, before accepting or not this change, it would be better to clarify your scenario with more details, which can be used to reproduce your case. I see something is missing to reproduce your case.

valgur:
Sorry for the noise, I must have mixed something up in the PR description. The Windows bit is indeed irrelevant, but pulseaudio/17.0 not having a shared option and breaking the self.dependencies["pulseaudio"].options.shared as a result still applies, I think?

uilianries (approved):
LGTM.

The case makes sense but only as overprotection in case an user is overriding pulseaudio version to 17.x.

This error does not occur right now in the CCI because we are using 14.x.

I didn't find any reported case in CCI: `https://github.com/conan-io/conan-center-index/issues?q=sort%3Aupdated-desc is%3Aissue is%3Aopen sdl`

PR #25122, opened 2024-09-03
Various Qt plugin targets and populate libs property

planetmarshall:
### Summary
Changes to recipe:  **qt/5.x**

#### Motivation

Adds some missing plugin targets, bringing the available CMake targets in the conan recipe closer to the official Qt CMake targets.

#### Details

Adds some missing plugin targets for CMake, notably

* `Qt5::QOffscreenIntegrationPlugin`
* Various Wayland platform and client plugins

Also populates the `libs` property even for shared builds. This is useful for CMake install scripts for Qt applications as the location of the plugin libraries can be extracted by examining the targets. Previously this was not possible as the `libs` property was only populated for static builds.

Note that this is still rather complicated due to https://github.com/conan-io/conan/issues/12654, but something like the following will work:

```.cmake
get_property(plugin_libs TARGET Qt5::QJpegPlugin PROPERTY INTERFACE_LINK_LIBRARIES)
list(GET plugin_libs 1 plugin_libs_1)
# the link libraries are generator expressions so just pass them through to
# the install script
install(CODE
        "
  file(GET_RUNTIME_DEPENDENCIES
    RESOLVED_DEPENDENCIES_VAR _deps
    LIBRARIES $<TARGET_FILE:${plugin_libs_1}>
  )
  file(INSTALL $<TARGET_FILE:${plugin_libs_1} \${_deps}>
        DESTINATION \${CMAKE_INSTALL_PREFIX}/lib
        FOLLOW_SYMLINK_CHAIN
  )
  "
)
```

---


jcar87 (changes requested):
Thanks for your PR! May I ask to please discard any formatting changes and limit PR to the minimal changes required for the goal? I appreciate some legacy code may look like it can be subjectively improved, but we do have a high amount of PRs to review - thanks!

PR #25143, opened 2024-09-04
Add Gliese

Blackflame576:
GitHub: https://github.com/DeepForge-Tech/Gliese

### DIFF
--- a/recipes/Gliese/config.yml
+++ b/recipes/Gliese/config.yml
@@ -1,0 +1,6 @@
+versions:
+  # It's encouraged to add new versions on the top
+  "2.1.0":
+    folder: all
+  "2.0.0":
+    folder: all


jcar87 (changes requested):
Hi @Blackflame576  - thanks for your contribution.

Recipes must all be lower case, including the subfolder in `recipes`, and git submodules are not allowed (the contents of the recipes should be in the repository).

Please also note that the upstream repository seems very recent, has no description of what this library is (no readme, no  docs).  
Security and due diligence are very important to us and our users, and without additional information, this will not be considered for Conan Center.

jcar87:
@DeepForge-Tech - let us know if you still want us to consider this library for inclusion in Conan Center. We are very happy to include to libraries that are _new_ to the community, in particular because their inclusion in Conan Center can aid with adoption for users. In most cases in Conan Center, the libraries we provide come from canonical sources and there is a library description, a readme, a test suite, etc. While libraries are provided as-is, we want to understand _what_ is being provided as well.

Please let us know if the upstream repository for Gliese includes this information, or if there is any other source of us to consider

Issue #5521, opened 2021-05-12
[package] gsl-lite/0.38.1: Contract violation usability issue


coffee-lord:
### Package and Environment Details
  * Package Name/Version: **gsl-lite/0.38.1**

The version of the [conanfile.py](https://github.com/conan-io/conan-center-index/blob/a836ab2cad1e57719f31f132076182056393ac5a/recipes/gsl-lite/all/conanfile.py) in this repo differs quite drastically from the [upstream version](https://github.com/gsl-lite/gsl-lite/blob/e1c381746c2625a76227255f999ae9f14a062208/conanfile.py), specifically we found that we can't use the conan-center version because it requires a user to choose a contract violation behavior upfront whereas the upstream version does not. Our build system needs to be able to decide the violation policy for itself at a later stage because we reuse conan dependencies for all submodules and some submodules require a different violation policy from the rest.

I propose we add `None` to the list of valid values for the `on_contract_violation` option and make it so that `conanfile.py` doesn't add any defines when `on_contract_violation` is `None`.

SpaceIm:
This option should never have been added in the recipe. Actually it should be completly removed.

coffee-lord:
That would work for us too. In fact, as a workaround, we are packaging the upstream conanfile in our private repo and it works exactly as we hoped it would. I was quite puzzled when I discovered that conan-center's conanfile deviates so much from the upstream version to the point where we can't use it in our project.

coffee-lord:
Also, a minor nitpick:
https://github.com/conan-io/conan-center-index/blob/a836ab2cad1e57719f31f132076182056393ac5a/recipes/gsl-lite/all/conanfile.py#L12

`("GSL")` is not a tuple, `("GSL",)` is.

perseoGI:
Hi there!
Sorry to catch up so late.
I've created a PR https://github.com/conan-io/conan-center-index/pull/25157 which aims to address your old request.

PR #25157, opened 2024-09-05
gsl-lite: add None option for on_contract_violation

perseoGI:
### Summary
Changes to recipe:  **gsl-lite/all**

#### Motivation

In this very old PR https://github.com/conan-io/conan-center-index/issues/5521, user request to have the possibility of not defining a concrete `on_contract_violation` policy. Instead, let downstream decide.

#### Details

Close https://github.com/conan-io/conan-center-index/issues/5521

---


recipes/gsl-lite/all/conanfile.py
```diff
@@ -24,7 +24,7 @@ class GslLiteConan(ConanFile):
     #  3. GSL_UNENFORCED_ON_CONTRACT_VIOLATION: nothing happens
     #
     options = {
-        "on_contract_violation": ["terminate", "throw", "unenforced"]
+        "on_contract_violation": ["terminate", "throw", "unenforced", "None"]
```
AbrilRBS, line 27: None is not a good name in this case, it gets too close to options being able to not be defined

jcar87:
~In terms of implementation, easiest in these cases, rather than have `None` as a valid value, is to simply remove it from the `default_options`, and then access it with `get_safe` -  it will remain undefined unless it is done explicitly. Has the same effect but it's less code (and arguably, a better reflection of the situation).~

secondary  to this, changing the behaviour of a default option will impact users that may already be relying on the previous default behaviour. we may break them or cause their own apps to behave differently. We need to evaluate the risk in this case (and any cases where we change a default). I would try and determine that before merging this PR.

And more generally - we should re-evalaute the pattern in Conan Center for header only libraries to propagate compiler flags to consumer via Conan options. The typical use _outside of Conan_ would be for the consumers to do a `#define` before doing an `#include`. I would only consider the "option" approach if the upstream project is able to generate targets in the `-config.cmake` files where the imported targets propagate it as a compiler flag, and if this is configurable at build time. So reflect how the library may be used outside of Conan Center, rather than invent a new pattern.

recipes/gsl-lite/all/conanfile.py
```diff
@@ -23,11 +24,12 @@ class GslLiteConan(ConanFile):
     #  2. GSL_THROW_ON_CONTRACT_VIOLATION: a gsl::fail_fast exception will be thrown
     #  3. GSL_UNENFORCED_ON_CONTRACT_VIOLATION: nothing happens
     #
-    options = {
-        "on_contract_violation": ["terminate", "throw", "unenforced"]
-    }
+    options = {"on_contract_violation": [None, "terminate", "throw", "unenforced"]}
     default_options = {
-        "on_contract_violation": "terminate",
+        "on_contract_violation": None,
```
jcar87, line 29: has this pattern not come up before, where we decide to leave the value undefined instead? I think it achieves the same effect in the package ID ?

jcar87: ignore this - this is a header only library so it shouldn't be in the package ID anyway

AbrilRBS (approved):

PR #25170, opened 2024-09-06
mysql-connector-cpp: new recipe

valgur:
### Summary
Changes to recipe:  **mysql-connector-cpp/9.0.0**

#### Motivation
MySQL Connector/C++ is a MySQL database connector for C++. It lets you develop C++ and C applications that connect to MySQL Server.

https://github.com/mysql/mysql-connector-cpp

[Packaging status](https://repology.org/project/mysql-connector-c++/versions)

#### Details
The project is not doing anything particularly too complex or special in its CMake files overall, but chooses to do some things in quite convoluted and non-standard ways:
- [`boostrap()`](https://github.com/mysql/mysql-connector-cpp/blob/9.0.0/cdk/cmake/bootstrap.cmake#L55), which is done by default before `project()` call. This breaks many Conan variables and targets and does not appear to really be necessary (it's disabled for Ninja, for example).
- [`merge_libraries()`](https://github.com/mysql/mysql-connector-cpp/blob/9.0.0/cmake/libutils.cmake#L145). This assumes that all external deps have been built as shared. Static external deps get merged into the final archive file, which leads to "duplicate symbol" linker errors on some compilers, such as apple-clang. I replaced it with the much simpler and more common approach of using `OBJECT`-type libs for the internal libraries.

Based on
- #21949
- #24678

Closes
- closes #24678
- closes https://github.com/conan-io/conan-center-index/issues/16312
- closes https://github.com/conan-io/conan-center-index/issues/374

---


recipes/mysql-connector-cpp/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,7 @@
+#include <mysqlx/xdevapi.h>
+
+int main()
+{
+    mysqlx::SessionSettings from_url("mysqlx://user:pwd@127.0.0.1:1234/db?ssl-mode=required");
```
perseoGI, line 5: Hi!
Is this constructor accessing the network layer? 
In that case, new policies in Conan enforce not to access network layer from test_packages for security reasons so another library test could be done.

recipes/mysql-connector-cpp/all/conanfile.py
```diff
@@ -0,0 +1,169 @@
...
+        lib = "mysqlcppconnx"
+        if not self.options.shared:
+            lib += "-static"
+            if is_msvc_static_runtime(self):
+                lib += "-mt"
+        self.cpp_info.libs = [lib]
+
+        if self.settings.os == "Windows":
+            self.cpp_info.system_libs.append("dnsapi")
+            self.cpp_info.system_libs.append("ws2_32")
```
perseoGI, line 155:
```suggestion
            self.cpp_info.system_libs.extend(["dnsapi", "ws2_32"])
```

recipes/mysql-connector-cpp/all/conanfile.py
```diff
@@ -0,0 +1,169 @@
...
+        elif self.settings.os != "FreeBSD":
+            self.cpp_info.system_libs.append("resolv")
+        if self.settings.os == "SunOS":
+            self.cpp_info.system_libs.append("socket")
+            self.cpp_info.system_libs.append("nsl")
```
perseoGI, line 160:
```suggestion
            self.cpp_info.system_libs.append(["socket", "nsl"])
```

perseoGI (commented):
Looking good

recipes/mysql-connector-cpp/all/test_package/test_package.cpp
valgur: It deliberately constructs a `SessionSettings` object instead of a `Session` one to avoid the creation of a network connection.

recipes/mysql-connector-cpp/all/conanfile.py
```diff
@@ -0,0 +1,167 @@
...
+            self.cpp_info.system_libs.append(["socket", "nsl"])
+
+        if not self.options.shared:
+            self.cpp_info.defines = ["MYSQL_STATIC", "STATIC_CONCPP"]
+
+
+        if is_apple_os(self) or self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs.extend(["resolv"])
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                self.cpp_info.system_libs.extend(["m", "pthread"])
```
uilianries, line 167:
```suggestion
                self.cpp_info.system_libs.extend(["m", "pthread", "dl"])
```

https://github.com/mysql/mysql-connector-cpp/blob/9.0.0/mysql-concpp-config.cmake.in#L463

recipes/mysql-connector-cpp/all/conandata.yml
```diff
@@ -0,0 +1,21 @@
+sources:
+  "9.0.0":
+    url: "https://github.com/mysql/mysql-connector-cpp/archive/refs/tags/9.0.0.zip"
+    sha256: "9d4b90a4efe8861e821136fb3024074d14875749d8b69d1821361b2f2b8bfe55"
+patches:
+  "9.0.0":
+    - patch_file: "patches/9.0.0/0001-override-cmake-policy-version.patch"
+      patch_description: "Set CMake policy version to 3.15, move project() statement"
+      patch_type: "conan"
```
uilianries, line 9:
```suggestion
```

I really don't see an error when not using this patch. I would avoid if possible.

recipes/mysql-connector-cpp/all/conandata.yml
```diff
@@ -0,0 +1,21 @@
...
+    - patch_file: "patches/9.0.0/0002-rename-find_dependency.patch"
```
uilianries, line 10: I would prefer having an issue pointed in the upstream about this. I understand the risk of name collision, but keeping it only here is not good as well.

recipes/mysql-connector-cpp/all/test_package/test_package.cpp
perseoGI: Okey, thanks for the clarification!
It may be interesting to put a comment above to clarify this "non network layer access behavior" for future reviews!

recipes/mysql-connector-cpp/all/patches/9.0.0/0006-replace-merge_libraries-with-object-libs.patch
uilianries, line 1: I could not find any other package manager (ubuntu, aur, gentoo and conda) using similar approach. I understand the risk of having duplicated symbols and causing a mess, but this case should be reported to the upstream as well.

uilianries (commented):
Thank you for your PR! This project is one that we tried to package many time, but now, with this PR, we are much closer to achieve it. 

I have some concerns about those patches, they are related to the upstream directly, and I don't find any other package manager doing the same. I would suggest opening issues to the upstream reporting about find_dependency and merge_library, otherwise, we will need to keep updating the same patch always.

recipes/mysql-connector-cpp/all/conandata.yml
valgur: It's not just a potential risk. The CMake configuration step consistently failed for me due to `find_package()` calls importing `CMakeFindDependencyMacro`.

recipes/mysql-connector-cpp/all/patches/9.0.0/0006-replace-merge_libraries-with-object-libs.patch
valgur: Don't be overly reliant on the system-wide package managers as a reference. The issue precisely stems from the upstream project assuming that all of the external libraries are system-wide shared libraries rather than static libs.

It's not just a risk either. The apple-clang compiler refuses to link `test_package` due to duplicate symbols from OpenSSL and the mysql-connector-cpp lib when OpenSSL is built as static, for example.

I'm not going to reach out to upstream regarding these issues, sorry.

recipes/mysql-connector-cpp/all/conandata.yml
valgur: Yeah, looks like the issues were caused by `bootstrap()` rather than the policy version or `project()` call location.

husitawi:
Despite that hot mess of CMAKE abstraction files and function, you actually did it, nice finds and well done! 🎉

recipes/mysql-connector-cpp/all/patches/9.0.0/0006-replace-merge_libraries-with-object-libs.patch
husitawi: Concerning duplicate symbols and OpenSSL, do you think it's worth trying my approach in the latest commits? In short, remove the OpenSSL dependency from mysqlx and only link it in the foundation target.

In their 'docs' they do explain that an external openssl lib can be included via a new common target.

perseoGI:
Would you mind  @valgur to add @samuaz `with_jdbc` option to your recipe?

valgur:
> Would you mind  @valgur to add @samuaz `with_jdbc` option to your recipe?

Sure, gladly, but not right now. Maybe after a couple of weeks.

valgur:
Thanks, @perseoGI. I'm in favor of a simpler solution as well, if it works.

perseoGI (approved):
Thank you @valgur, we are merging this finally!

PR #25173, opened 2024-09-07
qt: Add UIC test

DoomHammer:
### Summary
Changes to recipe:  **qt/6.x.x**

#### Motivation
UIC seems broken on Linux. It cannot find ICU so even though ICU is installed by Conan. These tests should catch this.

#### Details
Ideally, a fix for Linux should also come in here, but for now I have no idea how to handle it.

---

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -116,6 +116,7 @@ class QtConan(ConanFile):
 
         "gui": True,
         "widgets": True,
+        "qttools": True,
 
         "device": None,
         "cross_compile": None,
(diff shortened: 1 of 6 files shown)


DoomHammer:
FYI: @uilianries @franramirez688

DoomHammer:
Oh, the test code comes from Qt examples

DoomHammer:
Now this is weird. It was supposed to fail on Linux... :/

DoomHammer:
OK, it seems shared ICU was the problem here. Still, this PR might be useful maybe?

jcar87:
> OK, it seems shared ICU was the problem here. Still, this PR might be useful maybe?

we tend to keep the tests as simple as possible - 
CI doesn't currently build Qt on Linux due to a limitation in our infrastructure, so we wouldn't be able to test it on CI.

For things like this it's best to open an issue with the error logs so that we can best advise - thanks!

jcar87:
Closing this due to inactivity, as we don't have a bug to reproduce or even build logs as to what the error is. However, if there's something we can reproduce or inspect, we'll  be happy to assist to ensure 1) it's not broken 2) there are no regressions.

DoomHammer:
Does anybody even use qt6 with Conan? The last time I tried it, it was useless without working UIC.

jcar87:
> Does anybody even use qt6 with Conan? The last time I tried it, it was useless without working UIC.

What we've tested seems to work - do you have a reproducible example we can look at?

PR #25179, opened 2024-09-07
[zulu-openjdk/11.0.24] Add Zulu JDK version 11.0.24 for platforms of interest.

datalogics-robb:
### Summary
Changes to recipe:  **zulu-openjdk/11.0.24**

#### Motivation
Adding the binaries for zulu for these less popular platforms allows us to switch to using this for a JNI interface, where in the past we have had multiple JDK Conan packages in use.

#### Details

Add the Zulu jdk version 11.0.24 for x86, x86_64, armv8 and SPARC v9 (64-bit) platforms.

---


datalogics-robb:
> Hooks produced the following warnings for commit [661a486](https://github.com/conan-io/conan-center-index/commit/661a486e008866dac03e9b5da2d62e8e33538d67)

This warning would be produced by every version - and should be - but we would NOT want to put all those into system libs because end users might not use every library in the package.  This *should* be tagged with `Bump version` but probably isn't because of the warning.

ErniGH (approved):
Hi @datalogics-robb, the PR looks good thanks. Sorry for the hook noise :(

AbrilRBS (approved):

PR #25181, opened 2024-09-08
readstat: add new recipe of readstat/1.1.9

shun2wang:
### Summary
Add recipe:  **readstat/1.1.9**
https://github.com/WizardMac/ReadStat
#### Motivation
Add new library of readstat library to conancenter.

#### Details
ReadStat is a command-line tool (+ C library) for  read/write and converting SAS, Stata, and SPSS files.

---


recipes/readstat/all/conanfile.py
```diff
@@ -0,0 +1,113 @@
...
+            msbuild = MSBuild(self)
+            msbuild.build_type = self._msbuild_configuration
+            msbuild.build(sln=os.path.join("VS17", "ReadStat.sln"))
+        else:
+            apply_conandata_patches(self)
+            autotools = Autotools(self)
+            autotools.configure()
+            autotools.make()
+            
+    def package(self):
```
uilianries, line 96:
```suggestion
    def package(self):
        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
```

uilianries (commented):
@shun2wang Thank you for your PR! Please, take a look in the error messages, including in the hook messages. Right now, it's failing because the license is not copied to the licenses folder.

recipes/readstat/all/conanfile.py
shun2wang: Thanks for your review!
btw: I have problems with msvc build not because I dont have that locally env.

shun2wang:
hmmm, why it pending not start build?

uilianries:
I would suggest keeping sources indexed by OS in conandata.yml instead. You can obtain several examples in CCI, for instance: https://github.com/conan-io/conan-center-index/blob/master/recipes/wasmtime/all/conandata.yml

shun2wang:
NEED HELP:

Anyone can help me to build it on Windows? with thanks.

shun2wang:
Hi @uilianries Could you please help me with it? with thanks.

uilianries:
@shun2wang Unfortunately I have no access to Windows machine right now, but I'll have by this week. I'll be back to this PR this week.

uilianries:
@shun2wang I tried to build Readstat on Windows but it failed as well and looks be a lack of support from the upstream. 

Their CI is configured only for mingw on Windows: https://github.com/WizardMac/ReadStat/blob/dev/.github/workflows/build.yml

The MSVC project file supported there is old, and will need patch to work as well.

Are you really using Readstat on Windows? If Yes, how did you build it? In case you could not build locally, I would suggest only supporting Linux and Mac for now. 

You could ask for CMake support to the upstream, I believe it could not fix the Windows build, but give better Linux support as well.

shun2wang:
>Are you really using Readstat on Windows?

Yes, I build it by msys2 env on my local. I actually only want to use this on Windows and Mac, and you know Linux builds are quite easy

shun2wang:
Is there a way in conan to build these libraries for Windows using the msys environment?

uilianries:
> Is there a way in conan to build these libraries for Windows using the msys environment?

@shun2wang Yes, please read https://github.com/conan-io/conan-center-index/pull/26024#issuecomment-2503286710

Both ReadStat and libRdata look be in the same case. You could try the same approach for both packages. Tell me if you can not configure your recipe to work with MSYS2. Regards.

shun2wang:
>Both ReadStat and libRdata look be in the same case. You could try the same approach for both packages. Tell me if you can not configure your recipe to work with MSYS2.

Yes, I tried many many times for build with MSYS2 but failed, please help me. with thanks.

uilianries:
Took me a while, but now I'm able to build ReadStat on Windows:

Using MSYS + CL compiler is not possible, the project is not prepared for that. It would require refactoring the build files. For instance, it uses compiler flags for GCC directly on CL, and of course, it does not work.

Building with MinGW works without much change, I needed to incorporate an official patch to fix a build as well. Here are my build logs:

- [readstat-1.1.9-windows-mingw-static.log](https://github.com/user-attachments/files/20821326/readstat-1.1.9-windows-mingw-static.log)
- [readstat-1.1.9-windows-mingw-shared.log](https://github.com/user-attachments/files/20821328/readstat-1.1.9-windows-mingw-shared.log)

For MSVC, the upstream comes with .sln file and they are using in the CI. Doing some changes I'm capable to make it work too:

- [readstat-1.1.9-windows-msvc-static.log](https://github.com/user-attachments/files/20821340/readstat-1.1.9-windows-msvc-static.log)

shun2wang:
thanks a lot!

uilianries (approved):
@shun2wang Thank you for all your effort over this PR! 👋 

We finally managed to make it work 😄 

Some comments about this PR:

- The upstream provide a release .tar, but it misses the license file and VS project files. So downloading the source is necessary for msvc.
- When building with MinGW, I have the very same error as reported on the issue https://github.com/WizardMac/ReadStat/issues/300. The official patch worked like a charm
- There is no presence of `DynamicLibrary` and `__declspec(dllexport)` in the project, so no shared=True for MSVC build
- The Werror is configured by default and some warning was breaking Unix build. I needed to suppress Werror
- Only configuring `tools.compilation:verbosity` is not enough, needed to pass `V=1` to make command
- In order to make it work with Visual Studio, I needed to use a dynamic patch for the solution and project files. It can not be moved to `source()`, because I need the toolset value from settings (Conan does not allow accessing settings under source())


Regards!

shun2wang:
the "failed" label was not be removed.is it influenced to merge?

uilianries:
> the "failed" label was not be removed.is it influenced to merge?

No. The current CI ignores those labels, it's based on the check results instead. So don't worry, it was from the past CI environment.

shun2wang:
Should we merge this?

shun2wang:
Bump. can we merge this?

shun2wang:
No, This is ready for merge.

shun2wang:
@uilianries @AbrilRBS Please have a look.

shun2wang:
@uilianries @AbrilRBS @jcar87 Please have a look. with thanks.

PR #25182, opened 2024-09-08
seq: add recipe

toge:
### Summary
Changes to recipe:  **seq/1.1.0**

#### Motivation
Seq is a stable container library.

#### Details
https://github.com/Thermadiag/seq/

---

### DIFF
--- a/recipes/seq/all/conanfile.py
+++ b/recipes/seq/all/conanfile.py
@@ -1,0 +1,77 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get, rmdir
+from conan.tools.scm import Version
+import os
+
+required_conan_version = ">=1.53.0"
+
+class PackageConan(ConanFile):
+    name = "seq"
+    description = "a collection of original C++14 STL-like containers and related tools"
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/Thermadiag/seq"
+    topics = ("formatting", "hashmap", "radix", "header-only")
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
+
+    @property
+    def _min_cppstd(self):
+        return 14
+
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "10",
+            "clang": "7",
+            "gcc": "7",
+            "msvc": "191",
+            "Visual Studio": "15",
+        }
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def package_id(self):
+        self.info.clear()
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["HEADER_ONLY"] = True
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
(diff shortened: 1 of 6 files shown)


recipes/seq/all/conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["HEADER_ONLY"] = True
+        tc.generate()
+        venv = VirtualBuildEnv(self)
+        venv.generate(scope="build")
+
+    def build(self):
+        replace_in_file(self, os.path.join(self.source_folder, "seq", "type_traits.hpp"), "#if defined(__GNUG__) && (__GNUC__ < 5)", "#if !defined(__clang__) && defined(__GNUG__) && (__GNUC__ < 5)")
```
AbrilRBS, line 65: What's the reasonale behind this?
Also, did you report the issue to upstream? It would be great for tracking :)

recipes/seq/all/conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+    package_type = "header-library"
+    settings = "os", "arch", "compiler", "build_type"
+    short_paths = True
+
+    @property
+    def _min_cppstd(self):
+        return 14
+
+    @property
+    def _compilers_minimum_version(self):
```
AbrilRBS, line 28: Note that this method is no longer useful in Conan 2 (as we always have cppstd!), and that specific incompatibilities with old compilers should be expressed directly in the validate method.

So for example, if `gcc 5` supports the min_cppstd, but it has a bug that prevents it from compiling this library until gcc 6, then just add that in the validate method, like

```
def validate(self)
    check_min_cppstd(self, 14)
    if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "6":
        raise ConanInvalidConfiguration("Specific reason for each one")
```

instead of relying on a general method, this way the user will have more info :)

Thanks!

recipes/seq/all/conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+        self.tool_requires("cmake/[>=3.16 <4]")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["HEADER_ONLY"] = True
+        tc.generate()
+        venv = VirtualBuildEnv(self)
```
AbrilRBS, line 61: Note that this is implicit in Conan 2

AbrilRBS (commented):
Thanks! Some comments, looks good for the most part :)

recipes/seq/all/conanfile.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.16 <4]")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["HEADER_ONLY"] = True
```
AbrilRBS, line 59: Is it too hard to support the build version?

recipes/seq/all/conanfile.py
toge: I has been merged to upstream.
https://github.com/Thermadiag/seq/pull/4

I will created a separeted patch file.

toge: Fixed.

recipes/seq/all/conanfile.py
toge: I modified this PR to support shared build.
(1.1 supports shared build only.)

recipes/seq/all/conanfile.py
toge: @AbrilRBS 
Do you mean HEADER_ONLY set automatically with `header_only=True`?
I can't check it works fine.

recipes/seq/all/conanfile.py
toge: Thank you!
Fixed.

recipes/seq/all/conanfile.py
AbrilRBS: No, `VirtualBuildEnv` is automatically injected

toge: @AbrilRBS 
Thanks!
Fixed.

recipes/seq/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        "header_only": [True, False],
+    }
+    default_options = {
+        "header_only": False,
+    }
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def config_options(self):
```
uilianries, line 29:
```suggestion
    def configure(self):
```

The `config_options` is too early, it executes before evaluating any option value. The `configure()` has option values computed already.

uilianries (changes requested):
The recipe looks good, but the CI did not generate packages for header_only=False. Need to check the CI, blocking to avoid merging without the missing configuration.

AbrilRBS:
Hi @toge we have a fix ready for this, but we're holding the update for a bit. You don't need to do anything else for this PR, we'll retrigger it once the issue has been fixed on our side, thanks for your patience! :)

AbrilRBS:
@toge could you add 1.3 instead? Upstream answered positively to our questions :)

PR #25014, opened 2024-08-22
xz-utils: support armv8 and clang-cl on Windows

jcar87:
### Summary
* Support armv8 on Windows
* Support clang from Visual Studio installation

Was not previously working.

PR #25197, opened 2024-09-09
xz_utils: honor static runtime for clang-cl

SpaceIm:
### Summary

#### Motivation
In case of a profile representing clang-cl with static runtime, runtime is not honored. This PR fixes it.

#### Details
see https://github.com/conan-io/conan-center-index/pull/25014#issuecomment-2317511419

---

### DIFF
--- a/recipes/xz_utils/all/conanfile.py
+++ b/recipes/xz_utils/all/conanfile.py
@@ -4,7 +4,7 @@ from conan.tools.env import VirtualBuildEnv
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rename, replace_in_file, rm, rmdir, save
 from conan.tools.gnu import Autotools, AutotoolsToolchain
 from conan.tools.layout import basic_layout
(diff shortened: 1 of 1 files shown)


jcar87:
>  runtime is not honored. 

For this profile and _without_ changes from this PR:

```
Profile host:
[settings]
arch=armv8
build_type=Release
compiler=clang
compiler.cppstd=17
compiler.runtime=static
compiler.runtime_type=Release
compiler.runtime_version=v144
compiler.version=17
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Visual Studio 17
```

I can se the `/MT` flag being passed correctly (and the test package passing, otherwise I imagine it would fail with the mismatch)

The `/MT` flag is correctly set by the Conan-generated `.props` - so while it is building the pre-existing configuration that doesn't have `MT` in the name - it is actually correctly obeying the Conan-passed settings - may be that the solution here is to just disregard that block altogether

jcar87:
Closing this due to inactivity, and the fact that we have confirmed that the runtime is correctly honored so unsure what this is fixing. A cleanup may be due - but not a priority unless it's causing issues.

PR #25201, opened 2024-09-10
vswhere: new recipe

valgur:
### Summary
Changes to recipe:  **vswhere/3.1.7**

#### Motivation
CLI tool to locate Visual Studio 2017 and newer installations. https://github.com/microsoft/vswhere
[Packaging status](https://repology.org/project/vswhere/versions)

Used by NSS to configure MSVC support in #19262: https://github.com/nss-dev/nss/blob/NSS_3_104_RTM/coreconf/msvc.sh#L47-L48

While the dependency in NSS could probably be patched around, the tool is simple to package and possibly useful in other contexts as well.

#### Details

---

### DIFF
--- a/recipes/vswhere/all/conandata.yml
+++ b/recipes/vswhere/all/conandata.yml
@@ -1,0 +1,8 @@
+sources:
+  "3.1.7":
+    exe:
+      url: "https://github.com/microsoft/vswhere/releases/download/3.1.7/vswhere.exe"
+      sha256: "c54f3b7c9164ea9a0db8641e81ecdda80c2664ef5a47c4191406f848cc07c662"
+    license:
+      url: "https://raw.githubusercontent.com/microsoft/vswhere/3.1.7/LICENSE.txt"
+      sha256: "a1e41a01dd80dbe3f8ad5edebfd746e64ccff9c2aae3d8f47922746cf83204ef"
--- a/recipes/vswhere/all/conanfile.py
+++ b/recipes/vswhere/all/conanfile.py
@@ -1,0 +1,51 @@
+import os
+
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.files import copy, download
+from conan.tools.microsoft import is_msvc
+
+required_conan_version = ">=1.47.0"
+
+
+class VswhereConan(ConanFile):
+    name = "vswhere"
+    description = "Locate Visual Studio 2017 and newer installations"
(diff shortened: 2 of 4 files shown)


Nekto89:
Is it possible to not have this tool at all on machines with VS installed?
```
vswhere is included with the installer as of Visual Studio 2017 version 15.2 and later, and can be found at the following location: %ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe
```

valgur:
> Is it possible to not have this tool at all on machines with VS installed?
> 
> ```
> vswhere is included with the installer as of Visual Studio 2017 version 15.2 and later, and can be found at the following location: %ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe
> ```

Ah, good point. I missed that fact. It was not found on PATH when building NSS, though, despite VCVars being activated. Might need additional setup.

valgur:
I would still add it as a CCI package. `vswhere` is guaranteed to be installed only on VS 2017 / 15.2. It's normally not found on PATH, though. The recipes will need to include `env.append_path("PATH", r"C:\Program Files (x86)\Microsoft Visual Studio\Installer")` for it to be found. And event then it feels a bit fragile.

jcar87:
Agreeing with @Nekto89  here - note that Conan 2 supports VS2017 and newer - and that vswhere is unconditionally installed by the installer in that location (irrespective of where visual studio itself is installed) - don't think it's worth it. 

I'm not aware of any installation process of vswhere where it ends up in PATH - so it may be worth checking with the nss developers, https://github.com/nss-dev/nss/blob/NSS_3_104_RTM/coreconf/msvc.sh#L47-L48 - i suspect this should refer to the file with its full path

On the other hand, looking at the NSS scripts - it is using vswhere to manually set up the variables that the vcvars script would otherwise do - this could cause two issues:
- if vcvars is already "loaded" (that is, LIB, INCLUDE vars is already defined) - these variables will either be overwritten or appended/prepended 
- it picks up the "latest" that vswhere finds -  instead of the one specified by the user in the Conan profile.


It may be safest to let Conan bring up the vcvars, and simply define `GYP_MSVS_VERSION`, `VSPATH` and `GYP_MSVS_OVERRIDE_PATH` in the recipe - and don't let the nss script invoke vswhere at all

valgur:
Sounds reasonable. I'll look into getting rid of the vswhere dep in nss.

Issue #25206, opened 2024-09-10
[request] libtiff with latest zstd version


stevn:
### Package Name/Version

libtiff/*

### Changelog

<https://github.com/facebook/zstd/releases>

### Context about the new update

libtiff conanfile.py currently hard-codes the zstd version to v1.5.5:

<https://github.com/conan-io/conan-center-index/blob/master/recipes/libtiff/all/conanfile.py#L80>

However, I would like to use the newer version: zstd/1.5.6.

<https://conan.io/center/recipes/zstd?version=1.5.6>

I guess zstd/1.5.6 should be compatible as well with libtiff, at least I cannot find any specific version requirements.

<https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/ZSTDCodec.cmake?ref_type=heads>

stevn:
Could someone please explain the problem? Why was https://github.com/conan-io/conan-center-index/pull/25208 closed? Is it possible to use zstd/1.5.6 with libtiff? Is that too difficult? It looks like it breaks many if not all dependers of libtiff? Is this a conan-specific problem or is it related to these projects explicitly requiring zstd/1.5.5? Is there any upgrade path?

uilianries:
Hello! This issue is now fixed by #27878. Libtiff is using version ranges for the zstd dependency.

Please, update your local copy to the latest version available:

    conan install -r conancenter --update --requires=libtiff/4.7.1

Regards!

PR #25208, opened 2024-09-10
libtiff: bump deps & use version range for xz_utils & zstd

SpaceIm:
### Summary
Changes to recipe:  **libtiff/all**

#### Motivation
Use version range for xz_utils & zstd dependencies since it's now allowed, and bump some other deps in the meantime.

closes https://github.com/conan-io/conan-center-index/issues/25206

#### Details

---

### DIFF
--- a/recipes/libtiff/all/conanfile.py
+++ b/recipes/libtiff/all/conanfile.py
@@ -65,21 +65,21 @@ class LibtiffConan(ConanFile):
         if self.options.zlib:
             self.requires("zlib/[>=1.2.11 <2]")
         if self.options.libdeflate:
-            self.requires("libdeflate/1.19")
+            self.requires("libdeflate/1.21")
         if self.options.lzma:
-            self.requires("xz_utils/5.4.5")
+            self.requires("xz_utils/[>=5.4.5 <6]")
         if self.options.jpeg == "libjpeg":
-            self.requires("libjpeg/9e")
+            self.requires("libjpeg/9f")
         elif self.options.jpeg == "libjpeg-turbo":
-            self.requires("libjpeg-turbo/3.0.2")
+            self.requires("libjpeg-turbo/3.0.3")
         elif self.options.jpeg == "mozjpeg":
             self.requires("mozjpeg/4.1.5")
         if self.options.jbig:
             self.requires("jbig/20160605")
         if self.options.zstd:
-            self.requires("zstd/1.5.5")
+            self.requires("zstd/[>=1.5 <1.6]")
         if self.options.webp:
-            self.requires("libwebp/1.3.2")
+            self.requires("libwebp/1.4.0")
 
     def validate(self):
         if self.options.libdeflate and not self.options.zlib:


uilianries:
This PR results in future conflicts, the `libtiff` is largely used in CCI and any bump dependency may break other packages. For instance, we can quickly list several recipes using libtiff:

```
find ../ -name conanfile.py -exec grep libtiff/ {} + 
../podofo/all/conanfile.py:            self.requires("libtiff/4.6.0")
../gtk/all/conanfile.py:            self.requires("libtiff/4.3.0")
../openmvg/all/conanfile.py:        self.requires("libtiff/4.5.1")
../libvips/all/conanfile.py:            self.requires("libtiff/4.6.0")
../itk/all/conanfile.py:        self.requires("libtiff/4.6.0")
../gdk-pixbuf/all/conanfile.py:            self.requires("libtiff/4.6.0")
../osgearth/all/conanfile.py:        self.requires("libtiff/4.5.1")  # override
../opencv/3.x/conanfile.py:            self.requires("libtiff/4.6.0")
../opencv/4.x/conanfile.py:            self.requires("libtiff/4.6.0")
../opencv/2.x/conanfile.py:            self.requires("libtiff/4.6.0")
../librasterlite/all/conanfile.py:        self.requires("libtiff/4.5.1")
../ncbi-cxx-toolkit-public/26/conanfile.py:        "TIFF":         "libtiff/4.3.0",
../dcmtk/all/conanfile.py:            self.requires("libtiff/4.6.0")
../openslide/all/conanfile.py:        self.requires("libtiff/4.6.0")
../libgd/all/conanfile.py:            self.requires("libtiff/4.6.0")
../imagemagick/all/conanfile.py:            self.requires("libtiff/4.3.0")
../leptonica/all/conanfile.py:            self.requires("libtiff/4.6.0")
../djvulibre/all/conanfile.py:        self.requires("libtiff/4.6.0")
../tesseract/all/conanfile.py:            self.requires("libtiff/4.6.0")
../libcvd/all/conanfile.py:            self.requires("libtiff/4.6.0")
../vigra/all/conanfile.py:        self.requires("libtiff/4.6.0")
../freeimage/all/conanfile.py:            self.requires("libtiff/4.6.0")
../librasterlite2/all/conanfile.py:        self.requires("libtiff/4.5.1")
../wxwidgets/all/conanfile.py:        self.requires("libtiff/4.6.0")
../poppler/all/conanfile.py:            self.requires("libtiff/4.3.0")
../sail/all/conanfile.py:            self.requires("libtiff/4.6.0")
../pdf-writer/all/conanfile.py:            self.requires("libtiff/4.6.0")
../sdl_image/all/conanfile.py:            self.requires("libtiff/4.6.0")
../cimg/all/conanfile.py:            self.requires("libtiff/4.6.0")
../openimageio/all/conanfile.py:        self.requires("libtiff/4.6.0")
../gdal/pre_3.5.0/conanfile.py:        self.requires("libtiff/4.6.0")
../gdal/post_3.5.0/conanfile.py:        self.requires("libtiff/4.6.0")
../diligent-tools/all/conanfile.py:        self.requires("libtiff/4.3.0")
../proj/all/conanfile.py:            self.requires("libtiff/4.6.0")
../openscenegraph/all/conanfile.py:            self.requires("libtiff/4.6.0")
../libgeotiff/all/conanfile.py:        self.requires("libtiff/4.6.0", transitive_headers=True, transitive_libs=True)
```

In case adopting this PR, by building libtiff with those new dependencies, we will have those conflicts later:

Using the profile Linux GCC-11 Release static only (same as in CCI):

```
conan install --requires=podofo/0.9.7
ERROR: Version conflict: Conflict between libjpeg/9f and libjpeg/9e in the graph.
Conflict originates from libtiff/4.6.0

Run 'conan graph info ... --format=html > graph.html' and open 'graph.html' to inspect the conflict graphically.
```

```
conan install --requires=libvips/8.15.2

ERROR: Version conflict: Conflict between libjpeg/9f and libjpeg/9e in the graph.
Conflict originates from libtiff/4.6.0

Run 'conan graph info ... --format=html > graph.html' and open 'graph.html' to inspect the conflict graphically.
```

```
conan install --requires=itk/5.1.2
ERROR: Version conflict: Conflict between libjpeg/9f and libjpeg/9e in the graph.
Conflict originates from libtiff/4.6.0

Run 'conan graph info ... --format=html > graph.html' and open 'graph.html' to inspect the conflict graphically
```

stevn:
@AbrilRBS Could you please explain the problem? Why was this closed? Is it possible to use zstd/1.5.6 with libtiff? Is that too difficult? It looks like it breaks many if not all dependers of libtiff? Is this a conan-specific problem or is it related to these projects explicitly requiring zstd/1.5.5? Is there any upgrade path? Would it be possible to do this change for the next libtiff release for example without touching the older libtiff versions?

uilianries:
Hello @stevn!

This PR is not adding a new version, neither fixing a registered bug in issues. Are experience an error, like missing package or version conflicts?

The PR was closed due the massive dependencies graph affected by libtiff, as I commented previously. It can be updated yes, but need extra effort as the CI need follow a specific build order to avoid conflicts and missing binaries. This kind of operation is reversed for more important events, like security failure in a version, or a new version that affects many packages.

To mitigate such situation, CCI is adopting version ranges, but not all package are using it yet, as the team need to check each package for compatibility. There is a documentation about what package are supported with version ranges in CCI so far: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#adherence-to-build-service

> Would it be possible to do this change for the next libtiff release for example without touching the older libtiff versions?

When update libtiff version we could take the opportunity not only to re-visit the case, but also using more version ranges in CCI.

valgur:
@uilianries Wouldn't adding the version ranges for
```python
self.requires("xz_utils/[>=5.4.5 <6]")
self.requires("zstd/[>=1.5 <1.6]")
```
be justified to avoid conflicts in other libraries? I would keep the rest as is.

stevn:
@uilianries Thanks for the quick reply! This is valuable information that lets me understand the implications of such a change. I really appreciate your effort in examining such version dependencies, conflicts etc.

I have created a test project here which demonstrates my problem:

https://github.com/stevn/conan_test_tiff

Since the libtiff v4.7.0 conan package hardcodes its requirement to zstd v1.5.5, I tried using `override=True` to resolve the conan package version conflict when using zstd/1.5.6 in my project. However, then I get compiler errors (zstd.h header can no longer be found). This can be examined by commenting / uncommenting the various self.requires(zstd...) lines in my root conanfile.py.

I also read this note in the Conan documentation pages:

> Best practices
> The force and override traits to solve conflicts are not recommended as a general versioning solution, just as a temporary workaround to solve a version conflict. Its usage should be avoided whenever possible, and updating versions or version ranges in the graph to avoid the conflicts without overrides and forces is the recommended approach.

See https://docs.conan.io/2.0/reference/conanfile/methods/requirements.html

So I am currently not sure how to proceed.

AbrilRBS:
> be justified to avoid conflicts in other libraries? I would keep the rest as is.

Yes, but the reason we're not allowing them either right now is to avoid the proliferation of PRs that _just_ add this change which would collapse our ability to focus on more important PRs and the rest of the current backlog. As the new CI pipeline is quite close, we prefer to keep the pain for a bit more, and then once we can open 1 PR with all the changes, bring all the recipes that need version ranges up-to-date with eachother.

I hope this helps to understand the situation a bit better :)

AbrilRBS:
@stevn 
> However, then I get compiler errors (zstd.h header can no longer be found). This can be examined by commenting / uncommenting the various self.requires(zstd...) lines in my root conanfile.py.

This is because the `override=True` trait does _not_ generate a direct requirement to the current recipe, it only means "if you see this library, override its version, but do not add a requirement to myself". You might be looking for the `force=True` trait which does add the requirement.

But the best approach currently in Conan 2 to temporaly solve conflicts and try new versions graph-wise which have no control over, is to use the `[replace_requires]` section of profiles, [documented here](https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires)

I hope this helps, if you have any other question feel free to open an issue either here if its recipe-specific, or in the main Conan repo, we're always happy to help :)

stevn:
@AbrilRBS 

> override=True trait does not generate a direct requirement

Oh, I see. Sorry, my mistake. I updated the test project to use `force=True` and it works OK now.

> use the [replace_requires]

I'll give that a try as well, thanks!

PR #25213, opened 2024-09-10
libarchive: bump deps & use version range for xz_utils & zstd + use correct dependency for with_pcreposix option

SpaceIm:
### Summary

#### Motivation
Use version range for xz_utils & zstd dependencies since it's now allowed, and bump some other deps in the meantime.
While bumping versions, I've realized that pcre2 was duplicated in requirements: with_pcreposix actually requires pcre, not pcre2 (and with_pcre2 requires pcre2), so it's fixed in the meantime.

#### Details
Proof that `with_pcreposix` (which maps to `ENABLE_PCREPOSIX` CMake option of upstream CMakeLists) requires pcre, not pcre2:
https://github.com/libarchive/libarchive/blob/853bf656ef6d294a4fedc9afc74b262c3c0a6501/CMakeLists.txt#L1306-L1307
https://github.com/libarchive/libarchive/blob/853bf656ef6d294a4fedc9afc74b262c3c0a6501/build/cmake/FindPCREPOSIX.cmake

---

### DIFF
--- a/recipes/libarchive/all/conanfile.py
+++ b/recipes/libarchive/all/conanfile.py
@@ -93,7 +93,7 @@ class LibarchiveConan(ConanFile):
         if self.options.with_iconv:
             self.requires("libiconv/1.17")
         if self.options.with_pcreposix:
-            self.requires("pcre2/10.42")
+            self.requires("pcre2/10.44")
         if self.options.with_nettle:
             self.requires("nettle/3.9.1")
         if self.options.with_openssl:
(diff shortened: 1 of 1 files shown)


AbrilRBS:
Reopening this one as it's also a bugfix

recipes/libarchive/all/conanfile.py
```diff
@@ -93,25 +93,25 @@ def requirements(self):
...
+            self.requires("pcre/8.45")
         if self.options.with_nettle:
             self.requires("nettle/3.9.1")
         if self.options.with_openssl:
             self.requires("openssl/[>=1.1 <4]")
         if self.options.with_libb2:
             self.requires("libb2/20190723")
         if self.options.with_lz4:
-            self.requires("lz4/1.9.4")
+            self.requires("lz4/1.10.0")
```
AbrilRBS, line 104:
```suggestion
            self.requires("lz4/1.9.4")
```

It would conflict elsewhere, let's keep this for now, maybe a good version range target for the future?

AbrilRBS (commented):
Thanks for the logs showing that pcre should be used, quite helpful :)

AbrilRBS (approved):
Thanks!

AbrilRBS (approved):

uilianries (approved):
Good catch with pcre. Indeed that option is False by default, so we didn't see at first review. Thank you.

PR #25221, opened 2024-09-11
libcurl: add version 8.10.0

toge:
### Summary
Changes to recipe:  **libcurl/8.10.0**

#### Motivation
There are several improvements in 8.10.0.

#### Details
https://github.com/curl/curl/compare/curl-8_9_1...curl-8_10_0

---


jmalopoy:
The curl team modified the root CMakeLists.txt, so this PR will require changes in the recipe. I attached the ones I found. Hopefully they should help to it
[recipe_8.10.0.patch](https://github.com/user-attachments/files/16975831/recipe_8.10.0.patch)

toge:
@jmalopoy 
Thanks you for your advice and contribution.
Sorry. I have little time in this weekend.
Please wait for several days.

ErniGH (approved):
LGTM! Thanks

AbrilRBS (approved):

recipes/libcurl/all/conanfile.py
```diff
@@ -339,33 +339,53 @@ def _patch_cmake(self):
...
         replace_in_file(self, cmakelists, "find_package(Zstd REQUIRED)", "find_package(Zstd REQUIRED MODULE)")
-        replace_in_file(self, os.path.join(self.source_folder, "CMake", "FindZstd.cmake"), "if(UNIX)", "if(0)")
+        if Version(self.version) < "8.10.0":
+            replace_in_file(self, os.path.join(self.source_folder, "CMake", "FindZstd.cmake"), "if(UNIX)", "if(0)")
 
         # c-ares
-        replace_in_file(self, cmakelists, "find_package(CARES REQUIRED)", "find_package(c-ares REQUIRED CONFIG)")
-        replace_in_file(self, cmakelists, "${CARES_LIBRARY}", "c-ares::cares")
+        if Version(self.version) < "8.10.0":
+            replace_in_file(self, cmakelists, "find_package(CARES REQUIRED)", "find_package(c-ares REQUIRED CONFIG)")
```
jcar87, line 347: @ErniGH, @AbrilRBS 
did we consider doing the overrides in the generate() method to avoid this patching?
https://docs.conan.io/2/reference/tools/cmake/cmakedeps.html#overwrite-properties-from-the-consumer-side-using-cmakedeps-set-property 

note that we may be able to also set `CARES_LIBRARY` via `tc.cache_variables` instead of patching the anything

I suspect the same could be done for a lot of the other dependencies

Alternative consider following the pattern of injecting a cmake file during `project()` so that we locate and define the relevant variables ahead of time? we could clean this up by A LOT :D

not critical...

PR #25272, opened 2024-09-15
add package/faiss

simshi:
### Summary
Add new recipe:  **faiss** (https://github.com/facebookresearch/faiss)

#### Motivation
Faiss is a library for efficient similarity search and clustering of dense vectors.

#### Limitations
1. [x] Because there is no available `CUDAToolkit/system` and `OpenMP/system`, consumers must express dependencies on those dependencies by themself. Please refer to `test_package/CMakeLists.txt` as example.
   - Solved cuda dependency, add a cmake build module file, by @valgur
   - add '-fopenmp' related flags into `cpp_info`
2. [ ] Faiss prefers MKL, but no conan package, now, using OpenBLAS without an option to choose.
3. [ ] OpenBLAS recipe lacks of options like `USE_OPENMP`, `NUM_THREADS`, faiss openmp code calls into openblas without openmp supporting may cause issues. 

#### Verison
- 1.11.0

---


recipes/faiss/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("openblas/0.3.27")
+        self.test_requires("gtest/1.12.1")
```
valgur, line 74: `self.test_requires()` be placed under `build_requirements()` instead.

Besides that, following the value of the `tools.build:skip_test` config option on CCI is discouraged as it can lead to unexpected test failures for consumers and CCI maintainers try to avoid the additional potential burden from bug reports related to these.

recipes/faiss/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+        if not (
+            self.options.with_opt_generic or
+            self.options.get_safe("with_opt_avx2", False) or
+            self.options.get_safe("with_opt_avx512", False)
+        ):
+            raise ConanInvalidConfiguration("""Must configure at least one of following optimizations:
+    - generic: no optimization
+    - avx2: with x86 AVX2 instruction set
+    - avx512 (only availible >=1.8.0): with x86 AVX512 instruction set
+""")
```
valgur, line 88: I would suggest keeping the recipe slightly simpler and always enabling `FAISS_WITH_OPT_GENERIC` for non-debug builds instead. What do you think?

recipes/faiss/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+        if not self.dependencies["openblas"].options.get_safe("use_openmp"):
+            self.output.warning("Prefers openbals with use_openmp")
```
valgur, line 90:
```suggestion
            self.output.warning("-o openblas/*:use_openmp is recommended but is not available.")
```

recipes/faiss/all/patches/fix-cmake-1.7.4.patch
```diff
@@ -0,0 +1,252 @@
...
+-  target_compile_options(faiss_avx2 PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-mavx2 -mfma -mf16c -mpopcnt>)
+-else()
+-  # MSVC enables FMA with /arch:AVX2; no separate flags for F16C, POPCNT
+-  # Ref. FMA (under /arch:AVX2): https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64
+-  # Ref. F16C (2nd paragraph): https://walbourn.github.io/directxmath-avx2/
+-  # Ref. POPCNT: https://docs.microsoft.com/en-us/cpp/intrinsics/popcnt16-popcnt-popcnt64
+-  target_compile_options(faiss_avx2 PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/arch:AVX2>)
+-  # we need bigobj for the swig wrapper
+-  add_compile_options(/bigobj)
++find_package(OpenMP REQUIRED)
```
valgur, line 48: This will fail on AppleClang and on Clang provided by most Linux package managers.

Related:
- #24584
- #22360

recipes/faiss/config.yml
```diff
@@ -0,0 +1,5 @@
+versions:
+  "1.8.0":
+    folder: all
+  "1.7.4":
+    folder: all
```
valgur, line 5: I would only keep the latest version unless you specifically need the older one for some reason.

valgur:
> Since v1.8.0, faiss has a separated gpu library, linked to cpu lib by whole-archive, but conan itself has no solution for it

The libtorch recipe had a similar issue and has a workaround for it: #24759. Probably useful here as well.

valgur (changes requested):
Needs some work, but looks good overall. Thank you!

valgur:
I simplified the recipe a bit by avoiding most of the patching, improved OpenMP and CUDA support and filled in a few other gaps on my personal fork: https://github.com/valgur/conan-center-index/blob/dev/recipes/faiss/all/conanfile.py

simshi:
Thank you, that's great!

valgur:
Opened a PR for OpenMP support in OpenBLAS: #25344. It's already included in my personal fork as well.

Issue #25263, opened 2024-09-14
[request] ann/1.1.2


SpaceIm:
### Package Name/Version

ann/1.1.2

### Webpage

https://www.cs.umd.edu/~mount/ANN

### Source code

https://www.cs.umd.edu/~mount/ANN/Files/1.1.2/ann_1.1.2.tar.gz

### Description of the library/tool

ANN is a library written in C++, which supports data structures and algorithms for both exact and approximate nearest neighbor searching in arbitrarily high dimensions.

PR #25273, opened 2024-09-15
add new recipe ann/1.1.2

SpaceIm:
### Summary
Changes to recipe: **ann/1.1.2**

#### Motivation
Add ANN library to conancenter.

#### Details

- Upstream there are some raw Makefiles and a vcproj file (only for DLL, not static). I've written a simple CMakeLists instead (library is very simple to compile), so that it can be more portable.
- A small fix is needed for static lib on windows in order to avoid dllexport/dllimport. So a `ANN_STATIC` definition is introduced, it must be defined at build & consume time if windows & static.
- Another fix is needed for compilation with C++17 or above: there are `register` keyword in front of several variables in source code (several cpp files & private headers, no fix required in public headers). It's useless and not supported at all since C++17.
- The latest version is from 2010, so I don't hope any backport of these patches upstream.

Closes https://github.com/conan-io/conan-center-index/issues/25263

---

### DIFF
--- a/recipes/ann/all/CMakeLists.txt
+++ b/recipes/ann/all/CMakeLists.txt
@@ -1,0 +1,29 @@
+cmake_minimum_required(VERSION 3.15)
+project(ANN LANGUAGES CXX)
+
+include(GNUInstallDirs)
+
+file(GLOB ANN_SRC_FILES ${ANN_SRC_DIR}/src/*.cpp)
+
+add_library(ANN ${ANN_SRC_FILES})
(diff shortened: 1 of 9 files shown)


SpaceIm:
FYI, it's an optional dependency of `Graphviz` (it's required by `libmingle` component of `Graphviz`).

AbrilRBS (approved):
Looks good to me, thanks!

danimtb (approved):

PR #25302, opened 2024-09-18
libraw: fix inconsistencies with upstream project

valgur:
### Summary
Changes to recipe:  **libraw/[*]**

#### Motivation
The recipe uses a custom CMakelists.txt, which does not completely reproduce the logic in the project build scripts:
- The project builds two separate libraries - raw and raw_r. The only difference for the latter is that it's built with `-pthread`. I added it as an optional component to the recipe to avoid linking against both by accident in consuming recipes.
- The .pc files installed by the project include a `-Iinclude/libraw` compile flag not set by the recipe.
- The library provides a pure-C interface, but it requires linking against the C++ stdlib in that case.

The missing include dir path caused a compilation failure in `imlib2`.

#### Details

---


AbrilRBS (approved):
Thanks a lot for taking the time to fix the recipe usage, and the detailed PR description, we appreciate it :)

The changes look good to me

perseoGI (approved):
LGTM

PR #21073, opened 2023-11-11
ogre: migrate to Conan v2, bump to latest version

valgur:
Ogre switched to semver when transitioning from v1.12 to v13.x: https://www.ogre3d.org/2021/08/14/ogre-13-released

PR #25325, opened 2024-09-19
imgui: make backends available as optional components, convert -docking versions into an option

valgur:
### Summary
Changes to recipe:  **imgui/[*]**

#### Motivation
There are two major issues with the current approach of building only the core `imgui.cpp` and no backends or other optional components:
- Difficult to use. The consuming projects will need to inject the `<package_folder>/res/...` source files into the build system in order to use the optional components. This is manageable but very inconvenient compared to the alternative of linking against a suitable component target.
- Tends to break shared builds. Imgui uses a `IMGUI_API` export macro for shared builds and visibility handling. The users need to take care that the core library symbols provided by the imgui recipe and the optional components built by the consuming projects have the correct `IMGUI_API` corresponding the static/shared build of either project. This is tricky to manage and easy to mess up.

In addition to that, the current recipe does not provide a way to modify [`imconfig.h`](https://github.com/ocornut/imgui/blob/master/imconfig.h). These values cannot be changed after the core library has already been built.

Another minor benefit is that the Conan and system dependencies for the optional components are handled by `package_info()` in the imgui recipe and don't need to be duplicated in the consuming projects.

#### Details
The PR aims to be backwards-compatible. `imgui::imgui` CMake target still links against the core library only. All backends can be disabled altogether by setting `build_backends=False` for the old behavior.

These changes are not without their downsides:
- Significantly more complex recipe. Imgui famously not providing any kind of build system outside of examples does not help in this regard.
- Potential for option conflicts. The CCI recipe will likely need to build with most if not all backends enabled (or at the very least OpenGL and Vulkan ones) to avoid conflicts on CCI itself.
- Recipes listing `imgui::imgui` as a `cpp_info.requires` either explicitly or implicitly no longer link against just the core lib but against all the optional components as well.

Given these downsides, I was initially hesitant to open this PR and shelved it for a while, but after hitting a wall in the Ogre PR (#21073) when trying to unvendor ImGui there, this PR started to look a lot more reasonable. That is, Ogre expects the optional Freetype component to be available and building it as part of the Ogre recipe hits the shared build issues listed above.

Test build with nearly all options enabled (only DirectX 11 and DirectX 12 disabled): https://github.com/conan-io/conan-center-index/pull/25325#issuecomment-2370959006

---


valgur:
The recipe is ready for review.

recipes/imgui/all/conanfile.py
```diff
@@ -1,104 +1,345 @@
-from conan import ConanFile
-from conan.tools.files import get, copy, replace_in_file
-from conan.tools.scm import Version
-from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 import os
 import re
+from pathlib import Path
```
valgur, line 3: Not related to the recipe, but the linter incorrectly flags this line as
> conan-import-disallowed:	Private modules are disallowed, import tools from 'from conan.tools.xxxx import yyyyy'

PR #25348, opened 2024-09-21
PcapPlusPlus: add version 24.09

seladb:
### Summary
Changes to recipe:  **pcapplusplus/24.09**

#### Motivation
This is a new version of PcapPlusPlus, released a year after the previous version. It contains a lot of updates and changes

#### Details
High level summary of the changes: https://github.com/seladb/PcapPlusPlus/releases/tag/v24.09
Full Changelog: [`v23.09...v24.09`](https://github.com/seladb/PcapPlusPlus/compare/v23.09...v24.09)

---


recipes/pcapplusplus/cmake/conanfile.py
```diff
@@ -39,6 +40,10 @@ def validate(self):
         if self.settings.os not in ("FreeBSD", "Linux", "Macos", "Windows"):
             raise ConanInvalidConfiguration(f"{self.settings.os} is not supported")
 
+        compiler_version = Version(self.settings.compiler.version)
+        if self.settings.compiler == "gcc" and compiler_version < "5.1":
+            raise ConanInvalidConfiguration("PcapPlusPlus requires GCC >= 5.1")
```
seladb, line 45: I added this because the build with GCC 5.0 fails:
https://c3i.jfrog.io/c3i/misc/logs/pr/25348/1-linux-gcc-legacy/pcapplusplus/24.09//67b8b30a48b5f4d20d090acd959766840b9443a5-build.txt
https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/25348/1-linux-gcc-legacy/pcapplusplus/24.09//summary.json

I'm not sure why initializing an `std::array` with an initialization list doesn't work, but [this thread](https://stackoverflow.com/a/8192275) suggests it might be related to a bug in old GCC versions

This is new code that was added in v24.09 I can add this validation only for v24.09, but I figured that GCC 5 is pretty old so this validation can apply to v23.09 as well. Please let me know what you think

seladb:
@AbrilRBS can you please take a look at this PR?

recipes/pcapplusplus/cmake/conanfile.py
```diff
@@ -36,22 +46,41 @@ def requirements(self):
...
+
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
     def generate(self):
         tc = CMakeToolchain(self)
         tc.variables["PCAPPP_BUILD_TESTS"] = False
         tc.variables["PCAPPP_BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_SHARED_LIBS"] = self.options.shared
```
valgur, line 67:
```suggestion
```
This should not be necessary - set automatically by Conan and the project is using a high enough CMake policy version to not be affected by CMP0077 issues either.

AbrilRBS (dismissed):
This should work now, sorry for the noise!

seladb:
@AbrilRBS unfortunately CI still fails 😢

AbrilRBS:
This seems like a real compilation issue when compiling with msvc as a shared library. Do you have any insight into those actual errors?

seladb:
> This seems like a real compilation issue when compiling with msvc as a shared library. Do you have any insight into those actual errors?

Yes, compiling with MSVC as a shared library doesn't work and it's a known issue: https://github.com/seladb/PcapPlusPlus/issues/1396
We should not allow it in the Conan recipe

AbrilRBS:
@seladb I've pushed some changes in [`cf6a752` (#25348)](https://github.com/conan-io/conan-center-index/pull/25348/commits/cf6a7523396807b954499ce7919f8a44d4a30e13) that tell Conan to not compile for that configuration.

I'm being conservative and just added the check for the current versions, so when we add a new version in the future, the same compilation issue will pop up and we'll need to update the recipe. This is on pourpose so that once/if you fix this upstream, we don't inadvertedly miss on out on it :)

AbrilRBS (approved):
Looks good now, thanks a lot for your patience :)

AbrilRBS (approved):

seladb:
> Looks good now, thanks a lot for your patience :)

Thank you so much @AbrilRBS for your help, much appreciated! 🙏

czoido (approved):

PR #25361, opened 2024-09-23
onetbb: disable tbbind for Apple non-macOS

kambala-decapitator:
hwloc can't be built for these platforms

### Summary
Changes to recipe:  **onetbb/***

#### Motivation

to use the latest version of tbb for iOS, one must explicitly disable tbbind option which is inconvenient

#### Details

hwloc can only be built for macOS among Apple platforms

---

### DIFF
--- a/recipes/onetbb/all/conanfile.py
+++ b/recipes/onetbb/all/conanfile.py
@@ -54,7 +54,7 @@ class OneTBBConan(ConanFile):
     @property
     def _tbbbind_supported(self):
         if is_apple_os(self):
-            return return self.settings.os == "Macos" and Version(self.version) >= "2021.11.0"
+            return self.settings.os == "Macos" and Version(self.version) >= "2021.11.0"
         return True
 
     @property
--- a/recipes/onetbb/all/conanfile.py
+++ b/recipes/onetbb/all/conanfile.py
@@ -1,5 +1,6 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
+from conan.tools.apple import is_apple_os
 from conan.tools.build import cross_building
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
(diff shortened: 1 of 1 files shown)


recipes/onetbb/all/conanfile.py
```diff
@@ -52,7 +53,9 @@ def _tbbbind_hwloc_version(self):
 
     @property
     def _tbbbind_supported(self):
-        return self.settings.os != "Macos" or Version(self.version) >= "2021.11.0"
+        if not is_apple_os(self):
+            return True
+        return self.settings.os == "Macos" and Version(self.version) >= "2021.11.0"
```
valgur, line 58:
```suggestion
        if is_apple_os(self):
            return return self.settings.os == "Macos" and Version(self.version) >= "2021.11.0"
        return True
```
Seems easier to read to me. Just a suggestion.

AbrilRBS (dismissed):
Great catch, thanks a lot for taking the time to fix the issue :)

valgur (approved):
Thanks!

recipes/onetbb/all/conanfile.py
```diff
@@ -52,7 +53,9 @@ def _tbbbind_hwloc_version(self):
 
     @property
     def _tbbbind_supported(self):
-        return self.settings.os != "Macos" or Version(self.version) >= "2021.11.0"
+        if is_apple_os(self):
+            return return self.settings.os == "Macos" and Version(self.version) >= "2021.11.0"
+        return True
```
valgur, line 58:
```suggestion
        if is_apple_os(self):
            return self.settings.os == "Macos" and Version(self.version) >= "2021.11.0"
        return True
```

valgur (changes requested):
Typo

valgur (approved):

AbrilRBS (approved):
Thanks!

PR #25364, opened 2024-09-23
libxml2: add 2.11.9, 2.12.9 and remove unused versions

Nekto89:
### Summary
Changes to recipe:  **libxml2/***

#### Motivation
2.11.9 and 2.12.9 contain fix for [CVE-2024-40896] Fix XXE protection in downstream code
https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.11.9
https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.12.9

#### Details

I've also removed bugfix versions that aren't present in current master

---

### DIFF
--- a/recipes/libxml2/all/conandata.yml
+++ b/recipes/libxml2/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "2.12.9":
+    url: "https://download.gnome.org/sources/libxml2/2.12/libxml2-2.12.9.tar.xz"
+    sha256: "59912db536ab56a3996489ea0299768c7bcffe57169f0235e7f962a91f483590"
   "2.12.7":
     url: "https://download.gnome.org/sources/libxml2/2.12/libxml2-2.12.7.tar.xz"
     sha256: "24ae78ff1363a973e6d8beba941a7945da2ac056e19b53956aeb6927fd6cfb56"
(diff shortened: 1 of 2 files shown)


gsantner:
Related to this, there is also my PR #25322 open that adds the 2.13 series

AbrilRBS:
We'll also merge this PR this Wednesday alongside the linked PR, thanks a lot for taking the time to add the new patch versions, we appreciate it :)

jcar87:
> I don't know how to safely choose versions that can be removed. I just used text search but maybe it's not enough? There are multiple recipes with version ranges "libxml2/[>=2.12.5 <3]"

generally if they are not used by any other recipe in the same revision of the repository, it may be safe to delete if there are newer versions available and in use (or via a version range). However on the other hand, for when there are many major.minor.patch - we would always always keep the most recent .patch.

As an aside, we'd like to update the language used when "removing" versions - the Conan Center remote will always serve all versions ever published, there is no removal process (IIRC the only actual remove we've had was xz-utils due to the malicious nature of the release). dropping a version from the git repository has an affect of: "stop publishing new recipe revisions or packages for older versions" - which is a bit longer than 'remove', but a tad less misleading :P

AbrilRBS (approved):

uilianries (approved):
LGTM. I didn't find a removed version that is affecting CCI.

PR #25367, opened 2024-09-23
openjdk license: combine the 2 expressions into one for SBOM generator compatibility

johan-boule:
### Summary
Changes to recipe:  openjdk all versions, licence attribute

#### Motivation
When using the Conan SBOM generator extension, the CycloneDX library rejects the licence information because it's using SPDX expressions, and in this case, there can be only one element.

#### Details
To allow the license to be valid when using Conan's SBOM generator extension, we have to use only one element when using SPDX expressions. I'm unsure whether multiple WITH operators are valid, so I propose to use a AND operator. WITH has precedence over AND, so no parenthesis is needed.

---

NOTICE: I didn't read any of the CLA text I signed, nor did I read the contributing guidelines.

### DIFF
--- a/recipes/openjdk/all/conanfile.py
+++ b/recipes/openjdk/all/conanfile.py
@@ -13,7 +13,7 @@ class OpenJDK(ConanFile):
     url = "https://github.com/conan-io/conan-center-index/"
     description = "Java Development Kit builds, from Oracle"
     homepage = "https://jdk.java.net"
-    license = "GPL-2.0-only WITH Classpath-exception-2.0", "GPL-2.0-only WITH OpenJDK-assembly-exception-1.0"
+    license = "GPL-2.0-only WITH Classpath-exception-2.0 AND GPL-2.0-only WITH OpenJDK-assembly-exception-1.0"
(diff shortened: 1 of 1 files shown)


valgur (approved):

AbrilRBS:
Hi @johan-boule thanks a lot for taking the time to create this, we appreciate it.

Conan documents the license field to be a tuple object (usually only one element is present, but internally it gets converted to one!), so I think the best approach would actually be to improve the sbom generator to handle these cases, and have the tool make the join as needed

Happy to help with that if needed, thanks!

valgur:
@AbrilRBS I think that change is still useful, though. The SPDX licenses tuple often is ambiguous in terms of whether the list is to be interpreted as being joined by AND or OR.

johan-boule:
I wonder if there's a generic way to combine the tuple elements. Should it be combined with AND or with OR ?
In the particular case of openjdk, I think AND makes sense, but could there be cases where it should be OR ?

johan-boule:
Can you please move and reopen this issue to the conan-extensions project ?

PR #25378, opened 2024-09-24
magic_enum: disable building of tests and examples

valgur:
### Summary
Changes to recipe:  **magic_enum/[*]**

#### Motivation
My initial goal was to disable the building of tests and examples, but I found the CMake build to be doing nothing useful besides installing the headers (i.e. no config header being generated or something like that). I replaced the build step with a simple copying of the headers.

#### Details

---


AbrilRBS (approved):

jcar87 (dismissed):
Thanks @valgur.

We'd prefer to continue using the CMake scripts as they are - there is nothing wrong with them and they are the method used by other public package repositories as far as I can see. Please also note that in the future (however distant this may be), we may be allow packaging the generated cmake files. 

However feel free to define the necessary variables to disable building examples or tests.

jcar87:
For reference: https://github.com/conan-io/conan-center-index/pull/24894/files - using `copy()` instead of `cmake.install()` is what caused the conan center package to have a bug in the first place, resulting in the package for version 0.9.6 to have the files _in the wrong folder_, which was was not intended by the upstream maintainers and would have not been a problem had the recipe been using cmake all along. `cmake --install` allows separating where the headers live in the sources folder, vs where the maintainers want to have them installed. 
This is also the reason why the recipe _does_ duplicate the headers for the specific versions, so as to not cause disruption to any users which used the headers in the subfolder instead. Thus, this would be a regression and undoing an earlier fix in the recipe. It would have allowed this: https://github.com/Neargye/magic_enum/commit/eb867ec9baaabc5b60816861923fdef53972cba2#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a to be completely transparent.

valgur:
Thanks. I missed that.

> We'd prefer to continue using the CMake scripts as they are - there is nothing wrong with them and they are the method used by other public package repositories as far as I can see. Please also note that in the future (however distant this may be), we may be allow packaging the generated cmake files.

That's wildly incongruent with how nearly all header-only package recipes are set up on CCI and with the recipe templates as well, though. I'm fine either way, but I'd appreciate if you could refrain from making up new rules/conventions on the fly in every other PR.

jcar87:
> That's wildly incongruent with how nearly all header-only package recipes 

Oh dear! If there are header libraries that do have a canonical build system that we are not using, it would be useful to locate those so that we can plan to rectify this in the future.

I can see we have at least 50+ recipes for header only libraries that _do_ use CMakeToolchain - not sure I'd say it is the exception. I'd expect quite a lot of header-only libraries are provided "as is", or have a build system that is purely internal, without "install" logic - hence the use of copying in Conan recipes - but if they have a build system with install logic, it may be useful to avoid precisely what happened with the magic enum headers (as it allows upstream maintainers to abstract their folder structures).

AbrilRBS (approved):

ErniGH (approved):
LGTM

PR #25381, opened 2024-09-24
fcl: add missing transitive_headers=True, CMP0077

valgur:
### Summary
Changes to recipe:  **fcl/[*]**

#### Details

---

### DIFF
--- a/recipes/fcl/all/conanfile.py
+++ b/recipes/fcl/all/conanfile.py
@@ -85,7 +85,6 @@ class FclConan(ConanFile):
             tc.cache_variables["OCTOMAP_PATCH_VERSION"] = str(octomap_version.patch)
         tc.cache_variables["BUILD_TESTING"] = False
         tc.cache_variables["FCL_NO_DEFAULT_RPATH"] = False
-        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
         tc.generate()
 
         cd = CMakeDeps(self)
--- a/recipes/fcl/all/conanfile.py
+++ b/recipes/fcl/all/conanfile.py
@@ -47,10 +47,13 @@ class FclConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("eigen/3.4.0")
-        self.requires("libccd/2.1")
+        # Used in fcl/common/types.h public header
+        self.requires("eigen/3.4.0", transitive_headers=True, transitive_libs=True)
+        # Used in fcl/narrowphase/detail/convexity_based_algorithm/support.h
+        self.requires("libccd/2.1", transitive_headers=True, transitive_libs=True)
         if self.options.with_octomap:
-            self.requires("octomap/1.9.7")
+            # Used in fcl/geometry/octree/octree.h
+            self.requires("octomap/1.9.7", transitive_headers=True, transitive_libs=True)
 
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):
@@ -82,6 +85,7 @@ class FclConan(ConanFile):
             tc.variables["OCTOMAP_PATCH_VERSION"] = octomap_version.patch
         tc.variables["BUILD_TESTING"] = False
         tc.variables["FCL_NO_DEFAULT_RPATH"] = False
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
         tc.generate()
 
         cd = CMakeDeps(self)
--- a/recipes/fcl/all/conanfile.py
+++ b/recipes/fcl/all/conanfile.py
@@ -77,12 +77,12 @@ class FclConan(ConanFile):
         tc.cache_variables["FCL_COVERALLS_UPLOAD"] = False
         tc.cache_variables["FCL_WITH_OCTOMAP"] = self.options.with_octomap
         if self.options.with_octomap:
-            octomap_version_str = self.dependencies["octomap"].ref.version
+            octomap_version_str = str(self.dependencies["octomap"].ref.version)
             tc.cache_variables["OCTOMAP_VERSION"] = octomap_version_str
             octomap_version = Version(octomap_version_str)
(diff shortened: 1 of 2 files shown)


recipes/fcl/all/conanfile.py
```diff
@@ -82,6 +85,7 @@ def generate(self):
             tc.variables["OCTOMAP_PATCH_VERSION"] = octomap_version.patch
         tc.variables["BUILD_TESTING"] = False
         tc.variables["FCL_NO_DEFAULT_RPATH"] = False
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
jcar87, line 88: Please use `cache_variables` as per both CMake's and Conan documentation

recipes/fcl/all/conanfile.py
```diff
@@ -47,10 +47,13 @@ def layout(self):
...
-        self.requires("eigen/3.4.0")
-        self.requires("libccd/2.1")
+        # Used in fcl/common/types.h public header
+        self.requires("eigen/3.4.0", transitive_headers=True, transitive_libs=True)
+        # Used in fcl/narrowphase/detail/convexity_based_algorithm/support.h
+        self.requires("libccd/2.1", transitive_headers=True, transitive_libs=True)
         if self.options.with_octomap:
-            self.requires("octomap/1.9.7")
+            # Used in fcl/geometry/octree/octree.h
+            self.requires("octomap/1.9.7", transitive_headers=True, transitive_libs=True)
```
jcar87, line 56: none of these seem to require `transitive_libs` - eigen is a hader only library and for the other two headers mention, only types /macros seem to be required, but should cause no symbols to be _directly_ required by the translation unit including this header - at least, not "indirectly".

jcar87 (dismissed):

recipes/fcl/all/conanfile.py
valgur: It's generally not worth the time and effort to dig into the finer points of transitive_libs=True/False for every minor library, IMO, although I don't mind setting it to False for clear header-only libs like Eigen or octree.
It's very easy to miss uses of externally-defined symbols in the code and test_package usually tests a too narrow slice of the library to catch issues caused by transitive_libs=False.

Regarding Eigen specifically, though, it does have support for an optional SuiteSparse backend that could be well added in the future as an optional dependency, which is why I decided to keep it as True.

Anyway, I'll change it to False for Eigen and octree.

jcar87: > It's generally not worth the time and effort to dig into the finer points of transitive_libs=True/False for every minor library,

Thanks! We don't expect you to do it - we are very happy to do so as part of the review process so that we can orient this better to suit the recipe.

recipes/fcl/all/conanfile.py
```diff
@@ -63,25 +66,26 @@ def source(self):
...
             octomap_version_str = self.dependencies["octomap"].ref.version
-            tc.variables["OCTOMAP_VERSION"] = octomap_version_str
+            tc.cache_variables["OCTOMAP_VERSION"] = octomap_version_str
             octomap_version = Version(octomap_version_str)
-            tc.variables["OCTOMAP_MAJOR_VERSION"] = octomap_version.major
-            tc.variables["OCTOMAP_MINOR_VERSION"] = octomap_version.minor
-            tc.variables["OCTOMAP_PATCH_VERSION"] = octomap_version.patch
-        tc.variables["BUILD_TESTING"] = False
-        tc.variables["FCL_NO_DEFAULT_RPATH"] = False
+            tc.cache_variables["OCTOMAP_MAJOR_VERSION"] = octomap_version.major
```
jcar87, line 83: I suspect these may need `str()` - the requirement to be json serializable may come from the fact that `cache_variables` end up in the generated presets file.

recipes/fcl/all/conanfile.py
```diff
@@ -47,10 +47,13 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("eigen/3.4.0")
-        self.requires("libccd/2.1")
+        # Used in fcl/common/types.h public header
+        self.requires("eigen/3.4.0", transitive_headers=True)
+        # Used in fcl/narrowphase/detail/convexity_based_algorithm/support.h
+        self.requires("libccd/2.1", transitive_headers=True, transitive_libs=True)
```
jcar87, line 53: As far as I can see - this still does not require transitive_libs to be unconditionally set to true, has this been verified?

recipes/fcl/all/conanfile.py
valgur: Thanks!

> We can probably lint/statically detect the correct cases here - I'll make sure we have hooks that can advise better.

That would be absolutely incredible, but I imagine it's probably next to impossible to handle all the possible #ifdef configurations in the headers. Still, just knowing that transitive headers and possibly libs are definitely required with the baseline config would be very useful already.

AbrilRBS (dismissed):
Thanks!

recipes/fcl/all/conanfile.py
```diff
@@ -63,25 +66,26 @@ def source(self):
...
-            tc.variables["OCTOMAP_MINOR_VERSION"] = octomap_version.minor
-            tc.variables["OCTOMAP_PATCH_VERSION"] = octomap_version.patch
-        tc.variables["BUILD_TESTING"] = False
-        tc.variables["FCL_NO_DEFAULT_RPATH"] = False
+            tc.cache_variables["OCTOMAP_MAJOR_VERSION"] = str(octomap_version.major)
+            tc.cache_variables["OCTOMAP_MINOR_VERSION"] = str(octomap_version.minor)
+            tc.cache_variables["OCTOMAP_PATCH_VERSION"] = str(octomap_version.patch)
+        tc.cache_variables["BUILD_TESTING"] = False
+        tc.cache_variables["FCL_NO_DEFAULT_RPATH"] = False
+        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
```
jcar87, line 88:
```suggestion
```

AbrilRBS (approved):

jcar87 (approved):

PR #25388, opened 2024-09-24
boost v1.86.0: add process library

robomics:
### Summary
Changes to recipe:  **boost/1.86.0**

In Boost v1.86.0 `boos.process` v2 was turned into a compiled library (see [changelog](https://www.boost.org/users/history/version_1_86_0.html)).

This PR adds `process` to the list of dependencies and adds tests for the new library component.

#### Details

---


robomics:
Uhm... I guess the link error could be fixed with something like

```cmake
if (WIN32)
    target_link_libraries(executable ntdll)
endif()
```

But I don't know whether this error is a symptom of another issue happening upstream (e.g. a link flag missing in Boost.Asio).

Unfortunately I do not daily run Windows and do very little development on it, so I am not sure what's the best way forward here.

uilianries:
@robomics Thank you for providing this hotfix. I was about to check it! Tell me if you need any help. :smile:

recipes/boost/all/conanfile.py
```diff
@@ -2034,6 +2035,14 @@ def filter_transform_module_libraries(names):
 
                 self.cpp_info.components[f"numpy{pyversion.major}{pyversion.minor}"].requires = ["numpy"]
 
+            if not self.options.get_safe("without_process") and self.settings.os == "Windows":
+                self.cpp_info.components["process"].system_libs.extend(["ntdll", "shell32", "Advapi32", "user32"])
+                if self._shared:
```
Nekto89, line 2040: This is probably already handled by recipe through "BOOST_ALL_DYN_LINK"

robomics: That is what I would expect as well, but then I ran into errors complaining about missing symbols on Windows (see [here](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/25388/3-windows-visual_studio/boost/1.86.0//summary.json)) and noticed that in boost::process CMakeLists.txt linking to those libraries is just wrapped around an `if(WIN32)` check - [link](https://github.com/boostorg/process/blob/e8b59f66e6243aa84f28e984705fed5b120f6524/CMakeLists.txt#L59-L61).

So I believe explicitly link to those libs in the `conanfile.py` when building boost::process is not optional/redundant.

robomics:
@uilianries I have a good news and a bad news. The good news is that I figured out the issue with the Windows builds. The bad news is that the solution I found is not in line with what is done by upstream.

Upstream defines `BOOST_PROCESS_SOURCE=1` as private on the boost::process target (see [here](https://github.com/boostorg/process/blob/e8b59f66e6243aa84f28e984705fed5b120f6524/CMakeLists.txt#L49-L51)). However to make things work, I have to add it to the list of defines with public visibility.

Defining `BOOST_PROCESS_SOURCE` leads to the definition of `BOOST_PROCESS_V2_DECL` as `BOOST_SYMBOL_EXPORT` instead of `BOOST_SYMBOL_IMPORT` (see [here](https://github.com/boostorg/process/blob/e8b59f66e6243aa84f28e984705fed5b120f6524/include/boost/process/v2/detail/config.hpp#L156-L166); on Windows, `BOOST_SYMBOL_IMPORT` is defined as `__declspec(dllimport)` while `BOOST_SYMBOL_EXPORT` is defined as `__declspec(dllexport)`).

To me it looks like the `#define` logic used in boost::process may be incorrect (but I am not 100% sure).

uilianries:
@robomics Thank you for your deep investigation! Really interesting case! 

I see you used those CMakeList.txt files as reference. Please, understand that we do not use CMake to build boost and it's not officially supported (https://github.com/boostorg/cmake?tab=readme-ov-file#boost-cmake-support-infrastructure). So you should always check bjam files first and consider them as source of truth when building, because any difference between those files may lead you to a build trap. 

I see in the official bjam file, it's exposed `BOOST_PROCESS_DYN_LINK`, but not `BOOST_PROCESS_SOURCE`: https://github.com/boostorg/process/blob/boost-1.86.0/build/Jamfile#L55

Have you tried defining only `BOOST_PROCESS_DYN_LINK` in cpp_info instead? It should work, as https://github.com/boostorg/process/blob/boost-1.86.0/include/boost/process/v2/detail/config.hpp#L157 shows it and plus an `BOOST_SYMBOL_IMPORT` as we want to consume the DLL.

robomics:
@uilianries, yes defining `BOOST_PROCESS_DYN_LINK`or `BOOST_PROCESS_STATIC_LINK` is one of the attempts I made (see https://github.com/conan-io/conan-center-index/pull/25388/commits/6b83ca6738e118616db74c9533affbcb4f4d1152), but those defines alone didn't seem to be enough: [ci output](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/25388/8-windows-visual_studio/boost/1.86.0//summary.json).

uilianries:
@robomics Thank you for checking it previously. I'm looking into it right, it seems be a bug, but with a suggested fix: https://github.com/boostorg/process/pull/409

I'll test locally on my Windows machine first, in case working, I'll send my changes to this PR.

uilianries:
I would prefer disabling Boost.Process for 1.86.0 when building on Windows + Shared instead of applying that patch. However, the recipe uses [no-copy-source](https://docs.conan.io/2/reference/conanfile/attributes.html#no-copy-source) and will not allow us customizing the option value `without_process`, based on `shared` option.

The current error is reported already here: https://github.com/boostorg/process/issues/408
And, there is a hotfix: https://github.com/boostorg/process/pull/409

Nekto89:
I'm not sure whether this should be done in this PR or in separate one.

Boost process allows choosing between std::filesystem and boost::filesystem 
https://github.com/boostorg/process/blob/9561ebad1c1170ed7a489fc1ced8425f34ce84bc/build/Jamfile#L20C1-L20C70

There is already option in boost recipe "filesystem_use_std_fs" that switches usage to std::filesystem for some boost headers (boost::dll). From name of the option I would assume that it should also affect boost process.

uilianries:
> I'm not sure whether this should be done in this PR or in separate one.
> 
> Boost process allows choosing between std::filesystem and boost::filesystem [boostorg/process@`9561eba`/build/Jamfile#L20C1-L20C70](https://github.com/boostorg/process/blob/9561ebad1c1170ed7a489fc1ced8425f34ce84bc/build/Jamfile#L20C1-L20C70)
> 
> There is already option in boost recipe "filesystem_use_std_fs" that switches usage to std::filesystem for some boost headers (boost::dll). From name of the option I would assume that it should also affect boost process.

@Nekto89 A separated one, please. You could open an issue asking that feature in case you need it. I would not add in case not really required, as boost has several options already.

uilianries:
A second option would be disabling Boost Process on Windows, but only for 1.86.0. Who is needing process, would need to enable it manually:

    conan install -r conancenter --requires=boost/1.86.0 -o "&:without_process=False"
    
And we would need only this:

```
diff --git a/recipes/boost/all/conanfile.py b/recipes/boost/all/conanfile.py
index 5f7121e1b..f3052453d 100644
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -589,6 +589,13 @@ class BoostConan(ConanFile):
                 elif not self._has_cppstd_14_supported:
                     disable_graph()
 
+            # TODO: Revisit on Boost 1.87.0
+            # It's not possible to disable process only when having shared parsed already.
+            # https://github.com/boostorg/process/issues/408
+            # https://github.com/boostorg/process/pull/409
+            if Version(self.version) == "1.86.0" and is_msvc(self):
+                setattr(self.options, "without_process", True)
+
     @property
     def _configure_options(self):
         return self._dependencies["configure_options"]
@@ -758,6 +765,10 @@ class BoostConan(ConanFile):
             raise ConanInvalidConfiguration("Boost.Cobalt requires a C++20 capable compiler. "
                                             "Please, set compiler.cppstd and use a newer compiler version, or disable from building.")
 
+        # TODO: Revisit on Boost 1.87.0. Remove in case Process is fixed.
+        if Version(self.version) == "1.86.0" and is_msvc(self) and self.options.get_safe("shared") and self.options.get_safe("without_process", None) == False:
+            raise ConanInvalidConfiguration(f"{self.ref} Boost.Process will fail to be consumed as shared library on MSVC. See https://github.com/boostorg/process/issues/408.")
+
     def _with_dependency(self, dependency):
         """
         Return true when dependency is required according to the dependencies-x.y.z.yml file
```

In terms of CI and maintenance looks much simpler, but for users will be worst as need to enable an option that's working already when building as static library.

uilianries:
After talking to @AbrilRBS it looks more correct sacrificing Boost.Process on 1.86.0, as we have no sure from the author about the proposed PR. And we strict avoid patch projects, only as last resource, and Boost is even more controlled as it affects more users. 

I believe it will be fixed for 1.87.0, but meanwhile, you can use `-o "&:without_process=False"` to enable Boost.Process.

robomics:
Thanks for looking more in to this @uilianries!
It's a bit unfortunate that it won't be possible to build boost::process:v2 v1.86.0 as a shared lib on Windows, but I understand your concerns.
I think your solution for v1.86.0 is ok, as boost::process::v2 is anyway available in earlier releases (even though as a header-only library).

uilianries (approved):
LGTM. @robomics and @Nekto89 Thank you for digging into this hotfix. Unfortunately we could not have the best result, Windows shared support, but I hope will be working on the next release.

AbrilRBS (approved):

PR #25392, opened 2024-09-24
packio: migrate to Conan v2, drop old versions

valgur:
### Summary
Changes to recipe:  **packo/[*]**

#### Details
Continues from #16876 by @SpaceIm.

I dropped even more versions - everything up to v2.1.0, which was released in late 2020. This allowed the recipe to be cleaned up significantly by dropping the exceptions needed by the old versions.

---


recipes/packio/all/conanfile.py
```diff
@@ -27,81 +33,80 @@ class PackioConan(ConanFile):
...
 
-        if self.options.get_safe("standalone_asio"):
-            self.requires("asio/1.18.1")
+    def requirements(self):
+        if self.options.msgpack:
+            self.requires("msgpack-cxx/6.1.1")
+        if self.options.nlohmann_json:
+            self.requires("nlohmann_json/3.11.3")
+        if self._use_boost_json or not self.options.standalone_asio:
+            self.requires("boost/1.83.0")
```
AbrilRBS, line 79: Are newer versions not compatible?

AbrilRBS (commented):
Thanks, I think removing older versions is a good call. I have one minor question before approving, otherwise looks good

recipes/packio/all/conanfile.py
valgur: It avoids a conflict with `boost/1.83.0` in `msgpack-cxx`. It built fine with Boost 1.85.0 otherwise.

PR #25394, opened 2024-09-24
theora: add package_type, simplify test_package, fix component name

valgur:
### Summary
Changes to recipe:  **theora/1.1.1**

#### Details

---


recipes/theora/all/conanfile.py
```diff
@@ -73,9 +72,9 @@ def package(self):
     def package_info(self):
         self.cpp_info.set_property("pkg_config_name", "theora_full_package_do_not_use")    # to avoid conflicts with theora component
 
-        self.cpp_info.components["theora"].set_property("pkg_config_name", "theora")
```
AbrilRBS, line 76: This might be a noobie question, but what's the insight behind this fix? Is it related to the global name in consumer requires clauses, in which using `theora::theora` would bring unexpected requirements? Thanks!

valgur: That's exactly the reason, yeah. There's no way to distinguish between the whole-package "component" and the subcomponent otherwise. Pango, Cairo and some other recipes use a similar naming scheme for the core component.

Issue #25391, opened 2024-09-24
[package] glu/system and opengl/system don't work with mingw@Linux


madebr:
### Description

When using a mingw toolchain on Linux, you don't want to install system libglvnd or libglu packages.
The mingw toolchain provides everything you need.

Applying the patch at the bottom makes the package build and usable.

### Package and Environment Details

* Package Name/Version: **opengl/system** and **glu/system**
* Operating System+version: **Linux (all)**
* Compiler+version: **GCC 13**

### Conan profile

```
Host profile:
[settings]
arch=x86
build_type=Release
compiler=gcc
compiler.exception=seh
compiler.libcxx=libstdc++11
compiler.threads=win32
compiler.version=13
os=Windows
[buildenv]
*:AR=i686-w64-mingw32-ar
*:AS=i686-w64-mingw32-as
*:CC=i686-w64-mingw32-gcc
*:CXX=i686-w64-mingw32-g++
*:CPP=i686-w64-mingw32-cpp
*:LD=i686-w64-mingw32-ld
*:NM=i686-w64-mingw32-nm
*:OBJCOPY=i686-w64-mingw32-objcopy
*:OBJDUMP=i686-w64-mingw32-objdump
*:RANLIB=i686-w64-mingw32-ranlib
*:RC=i686-w64-mingw32-windres
*:STRIP=i686-w64-mingw32-strip
```

### Steps to reproduce

conan install opengl/system -pr:h mingw32 -pr:b default

### Logs

These patches

```patch
--- a/recipes/glu/all/conanfile.py
+++ b/recipes/glu/all/conanfile.py
@@ -29,6 +29,8 @@ class SysConfigGLUConan(ConanFile):
         self.info.clear()
 
     def system_requirements(self):
+        if self.settings.os == "Windows":
+            return
         dnf = package_manager.Dnf(self)
         dnf.install(["mesa-libGLU-devel"], update=True, check=True)
 
--- a/recipes/opengl/all/conanfile.py
+++ b/recipes/opengl/all/conanfile.py
@@ -23,6 +23,9 @@ class SysConfigOpenGLConan(ConanFile):
         self.info.clear()
 
     def system_requirements(self):
+        if self.settings.os == "Windows":
+            return
+
         dnf = package_manager.Dnf(self)
         dnf.install_substitutes(["libglvnd-devel"], ["mesa-libGL-devel"], update=True, check=True)
 
```

SpaceIm:
[...]

PR #25399, opened 2024-09-24
opengl: avoid unecessary call to package manager for irrelevant host OS

SpaceIm:
### Summary
Changes to recipe:  **opengl/all**

#### Motivation
Currently, when you cross-build from Linux or FreeBSD to another OS than those 2 ones, this recipe installs an OpenGL lib unrelated to host OS, which is useless. Indeed, there is no early check, just attempt to run a bunch of system package managers. So it tries to call build machine system package manager and... succeeds but it's just build machine lib.

Relates to https://github.com/conan-io/conan-center-index/issues/25391

#### Details
Prevent to call system package manager if not Linux, FreeBSD or SunOS (all these package managers calls in system_requirements() are related to these OS).

Also support SunOS in the meantime by trying installation with PkgUtil (package seems to be mesalibs, see https://www.opencsw.org/packages/CSWmesa/).

---

### DIFF
--- a/recipes/opengl/all/conanfile.py
+++ b/recipes/opengl/all/conanfile.py
@@ -23,6 +23,9 @@ class SysConfigOpenGLConan(ConanFile):
         self.info.clear()
 
     def system_requirements(self):
+        if self.settings.os not in ["Linux", "FreeBSD", "SunOS"]:
+            return
+
         dnf = package_manager.Dnf(self)
         dnf.install_substitutes(["libglvnd-devel"], ["mesa-libGL-devel"], update=True, check=True)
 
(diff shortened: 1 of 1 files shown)


valgur (approved):

recipes/opengl/all/conanfile.py
```diff
@@ -36,12 +39,15 @@ def system_requirements(self):
...
 
         zypper = package_manager.Zypper(self)
-        zypper.install_substitutes(["Mesa-libGL-devel", "glproto-devel"], 
+        zypper.install_substitutes(["Mesa-libGL-devel", "glproto-devel"],
                                    ["Mesa-libGL-devel", "xorgproto-devel"], update=True, check=True)
 
         pkg = package_manager.Pkg(self)
         pkg.install(["libglvnd"], update=True, check=True)
 
+        pkg_util = package_manager.PkgUtil(self)
```
jcar87, line 48: I may be missing something in the PR description but I can't find any details provided for this change?

SpaceIm: It's the last sentence of PR description, about SunOs

jcar87 (approved):

recipes/opengl/all/conanfile.py
jcar87: thanks! I knew i was missing something

franramirez688 (approved):

PR #25405, opened 2024-09-25
qcustomplot: add package_type, transitive_headers=True

valgur:
### Summary
Changes to recipe:  **qcustomplot/[*]**

#### Details
test_package fails to link with `transitive_libs=False`.

The OpenGL direct dependency should not be necessary since the code only uses qtOpenGL and not OpenGL itself.

---


recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,17 +40,14 @@ def config_options(self):
...
-        # FIXME: we shouldn't have to force shared in qt, but config file
-        #        generated by conan in qt static is likely broken, or maybe env vars.
-        self.options["qt"].shared = True
+        self.options["qt"].gui = True
+        self.options["qt"].widgets = True
 
     def requirements(self):
         if Version(self.version) >= "2.0.0":
-            self.requires("qt/6.4.1")
+            self.requires("qt/[>=6.4 <7]", transitive_headers=True, transitive_libs=True)
```
AbrilRBS, line 48: So this never worked in Conan 2 then? Nice catch if so!

recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,17 +40,14 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        # FIXME: we shouldn't have to force shared in qt, but config file
-        #        generated by conan in qt static is likely broken, or maybe env vars.
-        self.options["qt"].shared = True
+        self.options["qt"].gui = True
+        self.options["qt"].widgets = True
```
AbrilRBS, line 44: This will either generate missing binaries in the CI, or if it's a hard requirement, also be ignored if the user sets something else thru the `!important` options introduced in Conan 2. this should be checked in the validate method if so :)

AbrilRBS (commented):
Some comments, thanks for the fix!

recipes/qcustomplot/all/conanfile.py
valgur: There's already a validation check for them in place and they are enabled by default anyway in the qt recipe.

uilianries: qcustomplot should not count on it, each package should independent from each other. In case it requires a specific option, the same should be validated as well. You need to consider users can customize their profiles and have different results than those from CCI. Please, update validate() with that validation.

recipes/qcustomplot/all/conanfile.py
uilianries: Why transitive_libs BTW?

valgur: test_package without transitive_libs=True:
```
undefined reference to symbol '_ZN7QWidget11setGeometryERK5QRect@@Qt_6'
qt00e75f421b97e/p/lib/libQt6Widgets.so.6: error adding symbols: DSO missing from command line
```

valgur:
I don't have any good ideas on how to approach the `qt.qpa.plugin: Could not find the Qt platform plugin "cocoa" in ""` error right now.

recipes/qcustomplot/all/conanfile.py
uilianries: Perfect, thank you!

Issue #24577, opened 2024-07-10
A case for improved OpenMP support on CCI


valgur:
# Motivation

OpenMP (Open Multi-Processing) is a widely adopted API that provides parallel programming in C, C++, and Fortran, optionally with multi-platform shared-memory support. It is a critical building block for many numerical and scientific libraries, but is used in general-purpose software as well for easy parallelization support.

OpenMP consists of a set of `#pragma omp ...` compiler directives that get translated into OpenMP API calls, which are then handled by the runtime library, which is often (but not always) provided with the compiler toolchain and linked dynamically. The translation of OpenMP directives needs to be enabled with a suitable compiler flag, typically `-fopenmp` or `/openmp`. Passing the same flag to a linker instructs it to link against the OpenMP runtime library found in the compiler toolchain.

Based on the info in recipes alone, there are currently about 50 recipes on CCI currently or upcoming in PRs. Some more notable ones include:

- casadi: a symbolic framework for automatic differentiation and numeric optimization.
- ceres-solver: large-scale, non-linear optimization library.
- colmap: a photogrammetry software for 3D reconstruction from images.
- ensmallen: a header-only C++ library for mathematical optimization.
- fftw: a library for computing the discrete Fourier transform (DFT).
- flann: a library for fast approximate nearest neighbors search in high-dimensional spaces.
- imagemagick: a cross-platform software suite for displaying, converting, and editing raster images.
- g2o: a framework for optimizing graph-based nonlinear error functions.
- lightgbm: a gradient boosting framework that uses tree-based learning algorithms.
- metis: a library for partitioning graphs, finite element meshes, and producing fill-reducing orderings.
- opencv: an open-source computer vision and machine learning software library.
- openmvg: an open-source multiple view geometry library for 3D reconstruction.
- pcl: The Point Cloud Library for 2D/3D image and point cloud processing.
- stdgpu: a library for efficient GPU computing with standard C++.
- suitesparse: a suite of sparse matrix algorithms. The main open-source sparse matrix library.
- vlfeat: an open-source library for computer vision applications.
- xgboost: an optimized gradient boosting library designed for performance and speed.
- xtensor: a C++ library for multi-dimensional arrays with NumPy-like syntax.

Although some of these libraries support alternative parallelization mechanisms, such as plain pthreads or TBB, several of these libraries don't and are entirely (e.g. SuiteSparse) or close to unusable without parallelization via OpenMP enabled.

# Problems

[...]

PR #25419, opened 2024-09-26
cryptopp: Add use_openmp option

smessmer:
This option is present in Crypto++'s CMakeList.txt but was not exposed in the conanfile.py. Adding exposure for it.

### Summary
Changes to recipe:  **cryptopp/\***

#### Motivation
This is an option that allows enabling OpenMP for the cryptographic algorithms in Crypto++. It will speed up some and slow down other algorithms.

#### Details
Default is `False` since it can have a negative performance impact on some algorithms. But it will speed up others, so it should be up to the caller.

---

### DIFF
--- a/recipes/cryptopp/all/conanfile.py
+++ b/recipes/cryptopp/all/conanfile.py
@@ -189,6 +189,12 @@ class CryptoPPConan(ConanFile):
         elif self.settings.os == "Windows":
             self.cpp_info.components["libcryptopp"].system_libs = ["bcrypt", "ws2_32"]
 
+        if not self.options.shared and self.options.use_openmp:
+            if self.settings.compiler in ("gcc", "clang"):
+                openmp_flag = ["-fopenmp"]
+                self.cpp_info.components["libcryptopp"].cxxflags = openmp_flag
+                self.cpp_info.components["libcryptopp"].exelinkflags = openmp_flag
+
         # TODO: to remove in conan v2 once cmake_find_package* & pkg_config generators removed
         self.cpp_info.names["pkg_config"] = "libcryptopp"
         self.cpp_info.components["libcryptopp"].names["cmake_find_package"] = legacy_cmake_target
--- a/recipes/cryptopp/all/conanfile.py
+++ b/recipes/cryptopp/all/conanfile.py
@@ -26,10 +26,12 @@ class CryptoPPConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "use_openmp": [True, False],
     }
     default_options = {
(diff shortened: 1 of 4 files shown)


smessmer:
No idea why ci fails. Asked for access in the access task because i thought that would allow me to read the CI logs, but even though I saw a task saying I supposedly have access now, I still can't read the CI log. Seems it's a pretty complicated process to try and contribute to conan. Not sure how to proceed.

valgur:
Thanks! You will also need to list OpenMP in the `package_info()` linker flags for static builds to work for consumers.
See the code snippet in #24577 for an example, but skip `cpp_info.cflags` and `cpp_info.cxxflags`, since cryptopp does not appear to be using OpenMP in its public headers.

recipes/cryptopp/all/conanfile.py
```diff
@@ -179,6 +189,12 @@ def package_info(self):
         elif self.settings.os == "Windows":
             self.cpp_info.components["libcryptopp"].system_libs = ["bcrypt", "ws2_32"]
 
+        if not self.options.shared and self.options.use_openmp:
+            if self.settings.compiler in ("gcc", "clang"):
+                openmp_flag = ["-fopenmp"]
+                self.cpp_info.components["libcryptopp"].cxxflags = openmp_flag
+                self.cpp_info.components["libcryptopp"].exelinkflags = openmp_flag
```
valgur, line 196: @jcar87 Why not include MSVC and apple-clang support as well? Why `exelinkflags` but not `sharedlinkflags`?

`cxxflags` should be left unset as none of the public headers use `#pragma omp` or `#include <omp.h>` in their sources (it's not marked as such in the list of OpenMP consumers here at least: https://github.com/conan-io/conan-center-index/issues/24577#issuecomment-2241199222). And if it were necessary, then the `if not self.options.shared` part should be omitted.

jcar87: indeed it is `sharedlinkflags`, thanks.


MSVC is not needed here because `/openmp` is a compiler flag, not a linker one:
```
LINK : warning LNK4044: unrecognized option '/openmp'; ignored 
```
MSVC knows to link the runtime when one of the objects being linked require it (because they were compiled with `/openmp`) - fairly certain msvc embeds some directive in the obj that then gets processed by the linker - so no need to handle anything for msvc in this case.

OpenMP support in apple-clang is not documented by Apple, likely not officially supported at all - and the runtime is externally provided. We're just ensuring this works in the most likely used scenario - while the issue at large remains to be addressed.

jcar87 (approved):

franramirez688 (approved):

PR #25423, opened 2024-09-26
[qcustomplot] Use version range for Qt

uilianries:
### Summary
Changes to recipe:  **qcustomplot/2.1.1**

#### Motivation

The maintained version of Qt-5 is 5.15.14, resulting in missing binary packages when building this package. Using version ranges for Qt solves the current error.

---


AbrilRBS (dismissed):

valgur:
@uilianries Did you see #25405?

uilianries:
@valgur Nope, sorry. We are in middle of a maintenance, so I'm not checking PRs. I triggered your PR now to be built, if it passes at first build, I'll close mine, then we merge yours. Thank you!

AbrilRBS (approved):

jcar87 (dismissed):
check CI results for missing binaries

valgur:
Totally cosmetic and optional, but I'd suggest something like this for the ~30 recipes using Qt:
https://github.com/valgur/conan-center-index/blob/dev/recipes/qcustomplot/all/conanfile.py#L70-L85
Slightly cleaner, IMO. Although, this should really be fixed in the Qt recipe in the first place.

uilianries:
The current pre-checks results have the follow output: https://github.com/conan-io/conan-center-index/pull/25423/checks?check_run_id=36662334371

- Some invalid configuration because Qt6 requires C++17, so those configuration using C++14 are excluded. It's fine.
- Missing binaries can be divided in follow sections:
  - Cross-built package on Mac. It's expected because Qt6 Conan recipe is not prepared for cross-building (yet)
  - Qt5 is missing in the CI, but can be built locally, the qcustomplot 1.x uses Qt5 only. I built some configuration locally, here are my logs:
    - Mac armv8 all static: 
    - Mac armv8 all shared:
    - Linux x86_64 all static: TODO
    - Linux x86_64 all shared: TODO
- Linux builds are suspended when related to Qt6 due glibc version in the CI. Still, I can build it locally, here are my logs:
  - Linux x86_64 all static: TODO
  - Linux x86_64 all shared: TODO

jcar87:
> Qt5 is missing in the CI, but can be built locally, the qcustomplot 1.x uses Qt5 only. I built some configuration locally, here are my logs

Updated binaries should now be present
macOS x86_64 can happen on CI https://github.com/conan-io/conan-center-index/pull/26529 is merged first

recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,26 +40,26 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        # FIXME: we shouldn't have to force shared in qt, but config file
-        #        generated by conan in qt static is likely broken, or maybe env vars.
-        self.options["qt"].shared = True
```
franramirez688, line 44: Is this solved then?

recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,26 +40,26 @@ def config_options(self):
...
         if self.options.with_opengl and self.settings.os == "Windows":
             self.requires("opengl/system")
 
     def validate(self):
-        if self.info.settings.compiler.cppstd:
-            min_cppstd = "11" if Version(self.dependencies["qt"].ref.version) < "6.0.0" else "17"
-            check_min_cppstd(self, min_cppstd)
-        if not (self.dependencies["qt"].options.gui and self.dependencies["qt"].options.widgets):
-            raise ConanInvalidConfiguration(f"{self.ref} requires qt gui and widgets")
+        min_cppstd = "11" if Version(self.dependencies["qt"].ref.version) < "6" else "17"
```
franramirez688:
```suggestion
        min_cppstd = "11" if Version(self.version) < "2.0.0" else "17"
```

recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,26 +40,26 @@ def config_options(self):
...
+        check_min_cppstd(self, min_cppstd)
```
franramirez688, line 60: Let's increase the `required_conan_version`

recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,26 +40,26 @@ def config_options(self):
...
+
+    def build_requirements(self):
+        self.tool_requires("qt/<host_version>")
+        self.tool_requires("cmake/[>=3.27 <4]") # to be able to use CMAKE_AUTOMOC_EXECUTABLE
 
     def requirements(self):
         if Version(self.version) >= "2.0.0":
-            self.requires("qt/6.4.1")
+            # INFO: Public header qcustomplot.h includes QObject
+            self.requires("qt/[>=6.5 <7]", transitive_headers=True, options={"widgets": True, "gui": True})
```
franramirez688: Are we sure about this range? The library changelog says this:

```
#### Version 2.1.1 released on 06.11.22 ####

Added features:
  - Qt6.4 Compatibility
...

#### Version 2.1.0 released on 29.03.21 ####

Added features:
  - Compatibility up to Qt 6.0
...

#### Version 2.0.1 released on 25.06.18 ####

Other:
  ...
  - Qt 5.11 compatibility
```

franramirez688 (commented):
Looking great @uilianries 

A few comments to be discussed before moving forward 😁

recipes/qcustomplot/all/conanfile.py
franramirez688: The suggestion could change depending on this comment https://github.com/conan-io/conan-center-index/pull/25423#discussion_r2309361890

recipes/qcustomplot/all/conanfile.py
uilianries: The idea is enforcing Qt6, because Qt5 is [reaching its end-of-support soon.](https://www.qt.io/blog/qt-5.15-extended-support-for-subscription-license-holders); Still, users can use `[replace_requires]`.

What's your suggestion?

recipes/qcustomplot/all/conanfile.py
uilianries: Yes. That limitation dates back to 2021: https://github.com/conan-io/conan-center-index/pull/6029#issuecomment-869411143, specifically when involving Conan 1.x only.

recipes/qcustomplot/all/conanfile.py
uilianries: Done: [c917181](https://github.com/conan-io/conan-center-index/pull/25423/commits/c9171810d6fd76d1123230c6514f2746a060522d)

recipes/qcustomplot/all/conanfile.py
jcar87: while the libraries support qt5 and we still have it in Conan Center, we should still cover it in any version range.

if the `>= "2.0.0":` never supported qt5, I wouldn't bother (unless it's trivial for us to test)
in general when going from a fixed version to a version range, if the lower bound is published in the Conan 2 repository, please keep it, that is , it should probably be `[>=6.4.1 <7]`

uilianries:
```suggestion
            self.requires("qt/[>=6.4.1 <7]", transitive_headers=True, options={"widgets": True, "gui": True})
```

uilianries: Done: [676c398](https://github.com/conan-io/conan-center-index/pull/25423/commits/676c398cb27109f88f5b72aef87c77729d1ba7a7)

recipes/qcustomplot/all/conanfile.py
```diff
@@ -69,11 +70,14 @@ def source(self):
...
-        tc.variables["QT_VERSION"] = self.dependencies["qt"].ref.version
-        tc.variables["QCUSTOMPLOT_USE_OPENGL"] = self.options.with_opengl
+        tc.cache_variables["QCUSTOMPLOT_SRC_DIR"] = self.source_folder.replace("\\", "/")
+        tc.cache_variables["QCUSTOMPLOT_VERSION"] = str(self.version)
+        tc.cache_variables["QCUSTOMPLOT_VERSION_MAJOR"] = str(Version(self.version).major)
+        tc.cache_variables["QT_VERSION"] = str(self.dependencies["qt"].ref.version)
+        tc.cache_variables["QCUSTOMPLOT_USE_OPENGL"] = self.options.with_opengl
+        qt_tools_rootdir = self.conf.get("user.qt:tools_directory", self.dependencies["qt"].cpp_info.bindirs[0])
+        tc.cache_variables["CMAKE_AUTOMOC_EXECUTABLE"] = os.path.join(qt_tools_rootdir, "moc.exe" if self.settings_build.os == "Windows" else "moc")
+        tc.cache_variables["CMAKE_AUTORCC_EXECUTABLE"] = os.path.join(qt_tools_rootdir, "rcc.exe" if self.settings_build.os == "Windows" else "rcc")
```
franramirez688: I saw that `Qt6` recipe is already defining in its `package_info` this conf:

```python
        self.conf_info.define("user.qt:tools_directory", os.path.join(self.package_folder, "bin" if self.settings.os == "Windows" else "libexec"))
```
So, as this is something that was not defined before this PR, and also there are no issues about requiring it, could we limit this usage for `Qt6` to avoid extra complexity? 

```python
            if Version(self.version) >= "2.0.0":  # Qt6.x only
                # user.qt:tools_directory is defined by qt6.x recipe
                qt_tools_rootdir = self.conf.get("user.qt:tools_directory")
                suffix = ".exe" if self.settings_build.os == "Windows" else ""
                for tool in ["moc", "rcc"]:    
                    tc.cache_variables[f"CMAKE_AUTO{tool.upper()}_EXECUTABLE"] = os.path.join(qt_tools_rootdir, f"{tool}{suffix}")
```
Notice that we can use it directly as we're consuming Qt as a `tool_requires`.

recipes/qcustomplot/all/conanfile.py
```diff
@@ -39,26 +40,26 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
-        # FIXME: we shouldn't have to force shared in qt, but config file
-        #        generated by conan in qt static is likely broken, or maybe env vars.
-        self.options["qt"].shared = True
+
+    def build_requirements(self):
+        self.tool_requires("qt/<host_version>")
+        self.tool_requires("cmake/[>=3.27 <4]") # to be able to use CMAKE_AUTOMOC_EXECUTABLE
```
franramirez688: Let's not put the upper limit:

```suggestion
        self.tool_requires("cmake/[>=3.27]") # to be able to use CMAKE_AUTOMOC_EXECUTABLE
```

uilianries:
> should we stop publishing revisions for the older versions? 

I agree, as no other recipe in CCI is consuming it.

recipes/qcustomplot/all/conanfile.py
uilianries: Done, commit: 2189109c8

PR #25434, opened 2024-09-26
krb5: set pkg_config_name, add cross-compilation support

valgur:
### Summary
Changes to recipe:  **krb5/[*]**

#### Motivation
Setting `cpp_info.names["pkg_config"]` is not compatible with Conan v2 and fails to be found in `libpq`, for example.

#### Details

Also moved `libverto::libverto` requires to the root component (`mit-krb5`) based on `objdump -p | grep NEEDED` info and an associated linker error when using krb5 as a libpq dependency.

---


xakod:
Thank you for support krb5 recipe!

PR #25437, opened 2024-09-27
Add embree 4.3.3

EstebanDugueperoux2:
### Summary
Changes to recipe:  **embree/4.3.3**

#### Motivation
New attempt to have https://github.com/conan-io/conan-center-index/pull/18147/ working


uilianries (commented):
@EstebanDugueperoux2 Thank you for your PR. I would like to ask you to reduce the number of option to minimal, even excluding all them if you really do not really use them. 

I have been mitigating several bugs in CCI bugs due untested options, so we started to ask contributors to not only validate new options, but share a full of using it to be sure the recipe will work always.

As your PR has several options, it would demand several hours of validation, building combination of options and checking in the upstream what should be the result. 

Please, keep the recipe as simple as possible. In case really needing an option now or in the future, no problem, it can be integrated, but should be validated. As alternative, you can use the configuration `tools.cmake.cmaketoolchain:extra_variables` to inject more CMake definitions that are not listed in the recipe.

EstebanDugueperoux2:
> @EstebanDugueperoux2 Thank you for your PR. I would like to ask you to reduce the number of option to minimal, even excluding all them if you really do not really use them.
> 
> I have been mitigating several bugs in CCI bugs due untested options, so we started to ask contributors to not only validate new options, but share a full of using it to be sure the recipe will work always.
> 
> As your PR has several options, it would demand several hours of validation, building combination of options and checking in the upstream what should be the result.
> 
> Please, keep the recipe as simple as possible. In case really needing an option now or in the future, no problem, it can be integrated, but should be validated. As alternative, you can use the configuration `tools.cmake.cmaketoolchain:extra_variables` to inject more CMake definitions that are not listed in the recipe.

@uilianries thanks for your feedback.
I have enabled all options and removed ones already enabled by default in embree source.

Regards.

uilianries:
>  I have enabled all options and removed ones already enabled by default in embree source.

Thank you for reducing and checking in the upstream's CMakelists.txt

I still see several options available, which means, we need a full build covering all those new options in both values True/False. As the CI will not provide it due its coverage to only shared and fPIC, I will ask you to build locally and report it, the full build logs.

That's why I insist in not having custom options, unless you really need and use it directly. There is a cost of review and maintenance for each new option like I commented before. 

Regards.

recipes/embree/all/conanfile.py
```diff
@@ -0,0 +1,164 @@
...
+        tc.variables["EMBREE_ISA_NEON2X"] = self._has_neon
+        tc.variables["EMBREE_ISA_SSE2"] = self._has_sse_avx
+        tc.variables["EMBREE_ISA_SSE42"] = self._has_sse_avx
+        # For Emscripten disable TBB and all ISAs. It will compile only for SSE
+        if self.settings.os == "Emscripten":
+            tc.variables["EMBREE_ISA_AVX"] = self._has_sse_avx
+            tc.variables["EMBREE_ISA_AVX2"] = self._has_sse_avx
+            tc.variables["EMBREE_ISA_AVX512"] = self._has_sse_avx and not is_msvc(self)
+            if is_msvc(self):
+                tc.variables["USE_STATIC_RUNTIME"] = is_msvc_static_runtime(conanfile)
```
jcar87, line 120:
```suggestion
                tc.variables["USE_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
```

(see linter :) )

recipes/embree/all/conanfile.py
```diff
@@ -0,0 +1,164 @@
...
+        if self.settings.compiler == "clang" and compiler_version < "4":
+            raise ConanInvalidConfiguration("Clang < 4 is not supported")
+
+        check_min_vs(self, 191)
+
+        if self.settings.os == "Linux" and self.settings.compiler == "clang" and self.settings.compiler.libcxx == "libc++":
+            raise ConanInvalidConfiguration(f"{self.ref} cannot be built with clang libc++, use libstdc++ instead")
+
+        if self.settings.compiler == "apple-clang" and not self.options.shared and compiler_version >= "9.0":
+            raise ConanInvalidConfiguration(f"{self.ref} static with apple-clang >=9 and multiple ISA (simd) is not supported")
```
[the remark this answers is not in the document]
EstebanDugueperoux2: I switch this PR to draf to comment the validate method to analysis this linkage error.

EstebanDugueperoux2:
Hi @jcar87 @uilianries , it seems to have on issue on the CI, the "Pre-Checks" step is running since  14 hours.

jcar87:
> Hi @jcar87 @uilianries , it seems to have on issue on the CI, the "Pre-Checks" step is running since 14 hours.

just checked, there's an issue reporting the error, but it's a clear error
`def validate(self):` has been commented out, and the `raise ConanInvalidConfiguration` is now inside the requirements method, when it shouldn't be - these needs to be addressed

recipes/embree/all/conanfile.py
```diff
@@ -0,0 +1,163 @@
...
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.settings.os != "Emscripten":
+            self.requires("onetbb/2021.12.0")
+
+    # def validate(self):
```
jcar87, line 72:
```suggestion
    def validate(self):
```

recipes/embree/all/conanfile.py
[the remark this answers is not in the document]
EstebanDugueperoux2: I don't see linkage error.
Could you give me more details?

recipes/embree/all/conanfile.py
```diff
@@ -0,0 +1,153 @@
...
+        tc.variables["EMBREE_ISA_NEON2X"] = self._has_neon
+        tc.variables["EMBREE_ISA_SSE2"] = self._has_sse_avx
+        tc.variables["EMBREE_ISA_SSE42"] = self._has_sse_avx
+        # For Emscripten disable TBB and all ISAs. It will compile only for SSE
+        if self.settings.os == "Emscripten":
+            tc.variables["EMBREE_ISA_AVX"] = self._has_sse_avx
+            tc.variables["EMBREE_ISA_AVX2"] = self._has_sse_avx
+            tc.variables["EMBREE_ISA_AVX512"] = self._has_sse_avx and not is_msvc(self)
+            if is_msvc(self):
+                tc.variables["USE_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
```
uilianries, line 109:
```suggestion
        if is_msvc(self):
            tc.variables["USE_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
```
Reduce indentation to cover method scope.

uilianries (approved):
LGTM. Thank you for packaging Embree 4.x!

AbrilRBS (approved):

uilianries (approved):
LGTM.

- Embree 4 has several options and is hard to track all configurations
- TBB if configured here: https://github.com/RenderKit/embree/blob/v4.3.3/CMakeLists.txt#L255
- SSE and AVX option may depend the arch: https://github.com/RenderKit/embree/blob/v4.3.3/CMakeLists.txt#L407
- Neon requires Mac: https://github.com/RenderKit/embree/blob/v4.3.3/CMakeLists.txt#L399
- The extra libraries generated when static, they are marked private in CMake, but exposed and consumed in examples: https://github.com/RenderKit/embree/blob/v4.3.3/kernels/CMakeLists.txt#L350

PR #25444, opened 2024-09-27
gnutls: Add 3.8.7, and backport macos compilation fixes to 3.8.2 to 3.7.8

AbrilRBS:
Had to use this library and found some issues, took the time to also backport a fix


uilianries (approved):
LGTM. Patch are the same as described.

PR #25447, opened 2024-09-27
Add au recipe

ericriff:
### Summary
Changes to recipe:  **au/[0.3.5]**

#### Motivation
Add recipe for https://github.com/aurora-opensource/au

This is an incomplete recipe, it only exposes the main `Au::au` target and not the `io` component (which is used for the `<<` operator).

I'm not sure why it is needed since it works anyway? Their CMake is a bit weird.

---

### DIFF
--- a/recipes/au/all/patches/0001-v0.3.5-disable-gtest.patch
+++ b/recipes/au/all/patches/0001-v0.3.5-disable-gtest.patch
@@ -1,0 +1,58 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index b18407f..dbd35af 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -37,23 +37,23 @@ set(AU_EXPORT_SET_NAME AuHeaders)
+ 
+ enable_testing()
+ 
+-# Bring in GoogleTest so we can build and run the tests.
+-include(FetchContent)
+-FetchContent_Declare(
+-   googletest
+-   GIT_REPOSITORY https://github.com/google/googletest.git
+-   GIT_TAG 58d77fa8070e8cec2dc1ed015d66b454c8d78850 # Release 1.12.1
+-   FIND_PACKAGE_ARGS
+-      1.12.1
+-      NAMES GTest
+-)
+-
+-# https://google.github.io/googletest/quickstart-cmake.html
+-# For Windows: Prevent overriding the parent project's compiler/linker settings
+-set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
+-
+-FetchContent_MakeAvailable(googletest)
+-include(GoogleTest)
++## Bring in GoogleTest so we can build and run the tests.
++#include(FetchContent)
++#FetchContent_Declare(
++#   googletest
++#   GIT_REPOSITORY https://github.com/google/googletest.git
++#   GIT_TAG 58d77fa8070e8cec2dc1ed015d66b454c8d78850 # Release 1.12.1
++#   FIND_PACKAGE_ARGS
++#      1.12.1
++#      NAMES GTest
++#)
++#
++## https://google.github.io/googletest/quickstart-cmake.html
++## For Windows: Prevent overriding the parent project's compiler/linker settings
++#set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
++#
++#FetchContent_MakeAvailable(googletest)
++#include(GoogleTest)
+ 
+ add_subdirectory(au)
+ 
+diff --git a/cmake/HeaderOnlyLibrary.cmake b/cmake/HeaderOnlyLibrary.cmake
+index 28daf00..fbf0fab 100644
+--- a/cmake/HeaderOnlyLibrary.cmake
++++ b/cmake/HeaderOnlyLibrary.cmake
+@@ -75,7 +75,7 @@ function(header_only_library)
+   )
+ 
+   # Add the test, if requested.
+-  if (DEFINED ARG_GTEST_SRCS)
++  if (OFF)
+     add_executable("${ARG_NAME}_test")
+     target_sources("${ARG_NAME}_test" PRIVATE ${ARG_GTEST_SRCS})
+     target_link_libraries(
--- a/recipes/au/all/test_package/test_package.cpp
+++ b/recipes/au/all/test_package/test_package.cpp
@@ -7,9 +7,9 @@
 using namespace au;
 
 int main(void) {
-    constexpr auto lenght = meters(100.0);
+    constexpr auto length = meters(100.0);
 
-    std::cout << lenght << " == " << lenght.in(kilo(meters)) <<" km" << std::endl;
+    std::cout << length << " == " << length.in(kilo(meters)) <<" km" << std::endl;
 
     return EXIT_SUCCESS;
 }
--- a/recipes/au/all/test_package/test_package.cpp
+++ b/recipes/au/all/test_package/test_package.cpp
@@ -1,0 +1,15 @@
+#include <iostream>
+
+#include "au/au.hh"
+#include "au/io.hh"
+#include "au/units/meters.hh"
+
+using namespace au;
+
+int main(void) {
+    constexpr auto lenght = meters(100.0);
+
+    std::cout << lenght << " == " << lenght.in(kilo(meters)) <<" km" << std::endl;
+
+    return EXIT_SUCCESS;
+}
--- a/recipes/au/all/conandata.yml
+++ b/recipes/au/all/conandata.yml
@@ -1,0 +1,9 @@
+sources:
+  "0.3.5":
+    url: "https://github.com/aurora-opensource/au/archive/refs/tags/0.3.5.tar.gz"
+    sha256: "7ec826dc42968dc1633de56e4f9d06e70de73e820d2ac4788e8453343a622c9b"
+patches:
+  "0.3.5":
+    - patch_file: "patches/0001-v0.3.5-disable-gtest.patch"
+      patch_description: "Remove GTest dependency and disable tests build"
+      patch_type: "conan"
--- a/recipes/au/all/conanfile.py
+++ b/recipes/au/all/conanfile.py
@@ -39,6 +39,9 @@ class PackageConan(ConanFile):
     def layout(self):
         cmake_layout(self, src_folder="src")
 
+    def package_id(self):
+        self.info.clear()
+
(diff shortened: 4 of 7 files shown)


recipes/au/all/patches/0001-v0.3.5-disable-gtest.patch
```diff
@@ -0,0 +1,58 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
ericriff, line 1: Proposed a solution upstram, but their CMake is a bit odd so some feedback there is welcome ;)
https://github.com/aurora-opensource/au/pull/297

chiphogg:
> I'm not sure why it is needed since it works anyway? Their CMake is a bit weird.

Hi!  I'm commenting here so that I can stay in the loop on this PR.  (And also --- thanks so much for adding Au to conan!  Hugely appreciated.)

A couple of questions:

1. You're not sure why _what_ is needed, since _what else_ works anyway?  I wasn't able to figure this out from reading the rest of the PR summary.
2. What about our CMake is a bit weird?

For question (2), note that I'm not at all offended or anything --- this was my first time using or writing CMake, so I definitely _expected_ it to look weird, compared to what experienced people would write!  Instead, I'm asking because this seems like an awesome learning opportunity for me.  I did buy the CMake book and try to follow best practices using the latest CMake version, but I'm also coming from a bazel-first mindset, so I'm sure I have many blind spots.

Thanks again :slightly_smiling_face:

ericriff:
>You're not sure why what is needed, since what else works anyway? I wasn't able to figure this out from reading the rest of the PR summary.

If you look at the test package, I'm including the `io` header and `std::cout`-ing something constructed with `meters()`, without "linking" in `Au::io`. Yet it works anyways.
Just in case you're not familiar with conan, the usual workflow here is to remove all the library-provided CMake files from the packages and use the conan-generated ones. In this case, I'm removing the `AuConfig.cmake` and friends, so the targets Au provided normally are ignored. This recipe on its current form only provides the `Au::au` target.

> What about our CMake is a bit weird?
For question (2), note that I'm not at all offended or anything --- this was my first time using or writing CMake, so I definitely expected it to look weird, compared to what experienced people would write! Instead, I'm asking because this seems like an awesome learning opportunity for me. I did buy the CMake book and try to follow best practices using the latest CMake version, but I'm also coming from a bazel-first mindset, so I'm sure I have many blind spots.

Let me start with an apology, I did not meant to be criticize your code. English is not my first language and I may have said something that came out disrespectful, that was not my intention **at all**.
What I find _odd_ is that
1. GTest gets "leaked" to the consumer of `Au`. It should 100% be an internal thing, the consumer of `Au` shouldn't have to have it installed to be able to use this lib. This helps minimize the footprint of the library and simplifies packaging.
2. The CMake macro that declrares your headers only targets automatically defines tests, which makes the decoupling of the packages library and testing a bit harder. A common practice is to do something like this
```cmake
add_library(myCoolLibrary)

target_sources(..)
target_link_libraries(...)

if(BUILD_TESTING)
   add_subdirectiory(tests)
endif()
```

Then you completely ignore the tests with a single flag.

chiphogg:
No apology necessary.  :slightly_smiling_face:  The only reason I mentioned "I'm not offended" was because of what _I_ was saying in that message: I thought I could easily be mistaken for getting defensive, since tone doesn't translate well on the internet.

I think I'm starting to see the outlines of the changes we'll want to end up making.  Probably something like the following:

- Consolidate the library targets into two: one "main" one, and one for `testing` which depends on GTest
- Rework the header-only library utility to separate out test creation, since library targets won't be 1:1 with test targets in CMake anymore
- Create appropriate CMake options, which should make it easy to both (a) avoid building tests, and (b) avoid bringing in the GTest dependency (which are separate issues)

I'll need to go back and remember all the different use cases I tested when we first added CMake support (system install, FetchContent, ...), and make sure all those still work.  And we'll need to test the new use cases you're providing.  But I think we can make this work.

I'm assuming that once Au is in shape, we wouldn't need the patch file that's currently in this present PR --- does that sound right to you?

AbrilRBS (commented):
Hi @ericriff thanks a lot for taking the time to create the recipe for Au, we really appreciate it (And more so me personally, I've wanted to add this for a bit now (see https://github.com/aurora-opensource/au/issues/257) but hadn't found the time to do so)

This looks good to me as a first approach, but I'll need to circle back to this after rechecking @chiphogg comment in https://github.com/aurora-opensource/au/pull/297#issuecomment-2379924401 :)

ericriff:
>I'm assuming that once Au is in shape, we wouldn't need the patch file that's currently in this present PR --- does that sound right to you?

Exactly. Once Au doesn't publicly depend on GTest or any other internal tool, this patch will go away.

chiphogg:
> Just in case you're not familiar with conan, the usual workflow here is to remove all the library-provided CMake files from the packages and use the conan-generated ones. In this case, I'm removing the `AuConfig.cmake` and friends, so the targets Au normally provides are ignored. This recipe on its current form only provides the `Au::au` target.

This part was really intriguing and surprising to me.  So if a library has their own `CMakeLists.txt` files, and then they want to add conan support, conan will ignore those files and generate its own?  Is there a good reference where I can read about and understand this?

Also, I'm working on the Au-side changes, but it's a pretty busy and exciting time at work right now, so I don't have as much bandwidth right now as I often do at other times.  I'll do my best to keep the ball moving forward, though.

ericriff:
> > Just in case you're not familiar with conan, the usual workflow here is to remove all the library-provided CMake files from the packages and use the conan-generated ones. In this case, I'm removing the `AuConfig.cmake` and friends, so the targets Au normally provides are ignored. This recipe on its current form only provides the `Au::au` target.
> 
> This part was really intriguing and surprising to me. So if a library has their own `CMakeLists.txt` files, and then they want to add conan support, conan will ignore those files and generate its own? Is there a good reference where I can read about and understand this?

Not exactly. The CMake files conan replaces are the ones used to find the libraries once installed. In this case, `AuConfig.cmake` and `AuVersion.cmake`.
It also removes the find Modules if they're provided (in this case it would be something like `FindAu.cmake` and the package config files (`Au.pc`).

AbrilRBS:
Note that what @ericriff is saying is refering to the Conan Center Index context. Users adding their own support for Conan in their libraries are free to use their own config files without issue :)

recipes/au/all/patches/0001-v0.3.5-disable-gtest.patch
uilianries: Thank you for posting it to the upstream first!

recipes/au/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,15 @@
+#include <iostream>
+
+#include "au/au.hh"
+#include "au/io.hh"
+#include "au/units/meters.hh"
+
+using namespace au;
+
+int main(void) {
+    constexpr auto lenght = meters(100.0);
```
chiphogg, line 10: Nit: typo (`length`).

chiphogg (commented):
I had a general question w.r.t. strategy for this PR.

Right now, this PR patches the stock 0.3.5 version of Au.  Simultaneously, we're working on updating Au's CMake upstream, to be more friendly to conan (and vcpkg, I think).  But those changes obviously won't affect 0.3.5, which is set in stone at this point.

Does that imply that (1) we should plan to land this present PR, in close to its present form, _regardless_ of what's going on in the Au repo?  Does it mean (2) Au should try to get a 0.3.6 release out soon, as the "first conan release"?  Or does it mean that (3) we should use a specific (near-future) commit from HEAD as the release here?

I'm open to suggestion here.  I assume (3) is a no-go.  I'm guessing we want to do (1); but as for (2), I'd be open to consider cutting a release sooner than I had been planning.

AbrilRBS:
@chiphogg 

> Does that imply that (1) we should plan to land this present PR, in close to its present form, regardless of what's going on in the Au repo? Does it mean (2) Au should try to get a 0.3.6 release out soon, as the "first conan release"? Or does it mean that (3) we should use a specific (near-future) commit from HEAD as the release here?

We can go for a combination of both. We can release 0.3.5 as-is, and you can take the necessary time for 0.3.6. we'll then update the recipe without issue to handle the new version, so you can always release at your own cadance. How does that sound?

chiphogg:
> @chiphogg
> 
> > Does that imply that (1) we should plan to land this present PR, in close to its present form, regardless of what's going on in the Au repo? Does it mean (2) Au should try to get a 0.3.6 release out soon, as the "first conan release"? Or does it mean that (3) we should use a specific (near-future) commit from HEAD as the release here?
> 
> We can go for a combination of both. We can release 0.3.5 as-is, and you can take the necessary time for 0.3.6. we'll then update the recipe without issue to handle the new version, so you can always release at your own cadance. How does that sound?

Makes sense.  Let's do that!

recipes/au/all/test_package/test_package.cpp
ericriff: Done!

AbrilRBS:
Some background on the holdup for this PR until I get a chance to discuss it with the team:

`au::testing` seems to be a target that requires normally on `gtest`, but this is a bit problematic in Conan, as this could create potential conflicts with users using `gtest` for testing as a `test_requires`, which would be suprising. 

@chiphogg can you give me some more insight into this target? What's is intended usage? I see that it declares some matchers, but it would be nice to get a bit more insight

chiphogg:
Yep, that's right: `au::testing` is a library target that depends on `gtest`, despite not being a test.  The point is to define some testing helpers to make it easier to write tests with Au quantities.  For example, the `IsNear(...)` matcher is an effective workaround for a very longstanding defect in the `EXPECT_NEAR` macro from `gtest`.

`au::testing` is a "leaf" target: nothing in the library depends on it.  For conan, we could simply omit it.  The only downside would be that users who do use `gtest` wouldn't have access to it.  If we could figure out a way for them to "opt in", that'd be great, but the main thing is to get the core library onto conan IMO.

Also, the current version on `main` has been reworked so that `Au::au` and `Au::testing` are the only two public facing library targets, FWIW.

AbrilRBS:
So just to clarify, is this target always used as part of gtest testing?

chiphogg:
> So just to clarify, is this target always used as part of gtest testing?

Yes.  People who use gtest would find this target helpful in writing their tests.  But people who use other testing libraries wouldn't get any use out of it.

AbrilRBS (approved):

PR #25457, opened 2024-09-29
Add qoi

Julianiolo:
### Summary
Changes to recipe:  **qoi/1.0.0**

#### Details
I'm unsure on what to use as the version. The library does not seem to have an official versioning. The specification however is 1.0 ([see](https://qoiformat.org/qoi-specification.pdf)).

---

### DIFF
--- a/recipes/qoi/all/test_package/test_package.c
+++ b/recipes/qoi/all/test_package/test_package.c
@@ -1,50 +1,10 @@
 #include <stdio.h>
-#include <stdlib.h>
-#include <stdint.h>
 
 #define QOI_IMPLEMENTATION
 #include "qoi.h"
 
 int main(void) {
-    // 2x2 rgba sample image
-    const uint8_t sample_img[] = {
-        0x00, 0x00, 0x00, 0xff,
-        0xff, 0x00, 0x00, 0xff,
-        0x00, 0xff, 0x00, 0xff,
-        0x00, 0x00, 0xff, 0xff,
-    };
-
-    qoi_desc desc;
-    desc.width = 2;
-    desc.height = 2;
-    desc.channels = 4;
-    desc.colorspace = QOI_SRGB;
-
-    int out_len = 0;
-    void* encoded_img = qoi_encode(sample_img, &desc, &out_len);
-    if(encoded_img == NULL) {
-        printf("Error encoding image\n");
-        return EXIT_FAILURE;
-    }
-
-    qoi_desc desc2;
-    void* decoded_img = qoi_decode(encoded_img, out_len, &desc2, 0);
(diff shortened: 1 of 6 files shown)


toge (approved):

AbrilRBS:
> The library does not seem to have an official versioning. The specification however is 1.0 ([see](https://qoiformat.org/qoi-specification.pdf)).

@Julianiolo do you think we could ask upstream to generate a tag? Maybe they just have not thought about the usefulness of doing so for package managers

Julianiolo:
@AbrilRBS I don't know if the intend on having versions, but we could probably ask them?

Julianiolo:
@AbrilRBS I mean I kinda don't think they are planning on having any changes, as the specification stands, and the code is pretty minimal. The Readme and tests are updating more often. So I think either just calling it 1.0, or going with the commit hash might be the best idea, as there probably won't be any other versions soon.

But you can surely also ask them, I just don't know how active the owner is.

jcar87:
@ team - no need to wait for a tag.

The upstream readme mentions existing packages in well-known repositories
https://github.com/phoboslab/qoi/?tab=readme-ov-file#packages

We need a version scheme that monotonically increases
* https://aur.archlinux.org/pkgbase/qoi-git
* https://packages.debian.org/bookworm/source/qoi

`0.x.cci.date` as previously discussed should be as good as any, and is more or less the approach by Debian

Julianiolo:
@jcar87 Thanks for the answer!

How do you mean `0.x.cci.date` exactly? I assume `x` and `date` will be replaced by something?

recipes/qoi/all/test_package/test_package.c
```diff
@@ -0,0 +1,50 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+
+#define QOI_IMPLEMENTATION
+#include "qoi.h"
+
+int main(void) {
```
jcar87, line 8: I would simplify the implementation by a lot - we don't intend to test functionality at all, just integration that the library can be included.

`#include "qoi.h"` -> if this works it tends to be enough

we may want to refer to a symbol or something defined, for example we can print or access the contents of `qoi_padding` - that's enough for the test.

Julianiolo: @jcar87 yeah, I might have done a little overboard on that one... 😅

jcar87 (approved):
Thanks @Julianiolo !!

toge (approved):

recipes/qoi/all/test_package/test_package.c
jcar87: Thanks so much for addressing it! :)

PR #25458, opened 2024-09-29
libpq: add v17.0 using the new Meson build system

valgur:
### Summary
Changes to recipe:  **libpq/17.0**

#### Motivation
Switching to the new Meson-based build system to hopefully get rid of the maintenance overhead of the large number of complex patches and additional installation steps required in the Autotools-based recipe.

See the original commit in the repo for more context: https://github.com/postgres/postgres/commit/e6927270cd18d535b77cbe79c55c6584351524be

#### Details
Added nearly all supported external dependencies as options. Some of them are only used by the server executable, which is out of scope for the recipe, but I decided to keep them since the building of executables cannot be disabled, so might as well keep the server and other executables in the package as an option.

These PRs are required for optional dependencies:
- #25434
- #25455

---


recipes/libpq/meson/conanfile.py
```diff
@@ -0,0 +1,378 @@
...
+
+    def build(self):
+        apply_conandata_patches(self)
+        meson = Meson(self)
+        meson.configure()
+        meson.build()
+
+    def _fix_static_library_names(self):
+        # Meson outputs static libraries with a .a extension on Windows,
+        # which is not compatible with Conan and CMake.
```
jwillikers, line 230: Fyi, newer CMake versions, 3.29 and later, support `.a` file extensions for static libraries on Windows when building with MSVC. See the release notes [here](https://cmake.org/cmake/help/latest/release/3.29.html#other-changes).

It would be nice to see support for using the `.a` suffixes eventually. I think Meson made the right decision to use `.a` instead of `.lib` on Windows for static libraries to avoid confusion with import libraries which use the same `.lib` extension.

jwillikers (approved):
+1 for Meson

PR #25467, opened 2024-09-30
OpenEXR: Add version 3.3.0

irieger:
### Summary
Changes to recipe:  **openexr/3.3.0**

#### Motivation
Update the version of openexr which uses a new internal framework underneath for all APIs that is more performance optimized than before.

#### Details
Haven't fully analyzed the full change set ("504 changed files with 35,088 additions and 40,445 deletions") but the build process hasn't changed much. The most prominent change is that with 3.3 C++ 17 is required.

https://github.com/AcademySoftwareFoundation/openexr/compare/v3.2.4...v3.3.0

---


recipes/openexr/3.x/conanfile.py
```diff
@@ -59,6 +70,10 @@ def requirements(self):
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):
             check_min_cppstd(self, self._min_cppstd)
+        
+        minimum_version = self._minimum_compiler_version[str(self.settings.compiler)]
```
AbrilRBS, line 74:
```suggestion
        minimum_version = self._minimum_compiler_version.get(str(self.settings.compiler))
```

AbrilRBS (approved):

perseoGI (approved):

jcar87:
Hi @irieger  - is C++17 strictly required? other than the changes in their CI (which does indeed use C++17), I have not seen anything else that seems to require c++17 - I'm able to build the most recent version we have with C++11 on macOS. Interested in this because openexr is a dependency of opencv, and would like it to require a newer version, and it would be the only dependency pushing the entire dependency graph to be c++17 on Windows (where c++14 is still the default for MSVC). 

Our CI is passing with C++14 on Windows https://github.com/conan-io/conan-center-index/pull/28009 as well

Any info is appreciated, thanks!

irieger:
> Hi @irieger - is C++17 strictly required? other than the changes in their CI (which does indeed use C++17), I have not seen anything else that seems to require c++17 - I'm able to build the most recent version we have with C++11 on macOS. Interested in this because openexr is a dependency of opencv, and would like it to require a newer version, and it would be the only dependency pushing the entire dependency graph to be c++17 on Windows (where c++14 is still the default for MSVC).
> 
> Our CI is passing with C++14 on Windows #28009 as well
> 
> Any info is appreciated, thanks!

Hey,

I haven't tested this myself to be honest. It was - as far as I remember - discussed in the project Slack that folks have been happy now that the restriction of C++14 was lifted and new stuff could be used and the project itself moved all to C++17 so I set it as the min requirement.

As many tools used and maintained mainly in the VFX industry, the project is working based on the VFX reference platform and switched its reference with 3.3.0 to CY 2024. Which is the fourth year of the platform supporting 2017 so I guess they decided with this minor version which still was quite major in internal changes that they move on. But I mean that doesn't mean they necesarily alread use C++17 features.

PR #25478, opened 2024-10-01
NASM: use NMake when building with clang/clang-cl

dutow:
### Summary
Changes to recipe:  **nasm/* **

#### Motivation
Currently build fails with clang on Windows

#### Details
This commit copies the clang detection logic from another recipe and uses Windows settings with NMake correctly when using it.

---

### DIFF
--- a/recipes/nasm/all/conanfile.py
+++ b/recipes/nasm/all/conanfile.py
@@ -21,18 +21,22 @@ class NASMConan(ConanFile):
 
     settings = "os", "arch", "compiler", "build_type"
 
+    def _is_clang_cl(self):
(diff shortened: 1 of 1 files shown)


jcar87:
Hi @dutow - thanks for your PR.

From what I can see, the nmake scripts are hardcoded to use `cl.exe` (msvc)  

What is the profile that you are using? I'd recommend just building nasm with msvc, in the sense that it is an executable and the compiler setting is being removed from the package ID anyway - but happy to look into building it with clang if this is a hard requirement.

jcar87:
Closing due to inactivity - however please feel free to reopen if we have more clarity around cl.exe/clang-cl.exe usage and if the suggestion above does not work- thanks!

PR #25500, opened 2024-10-03
cpython: avoid invalid configurations

valgur:
### Summary
Changes to recipe:  **cpython/[*]**

#### Motivation
The cpython recipe is currently marked as invalid on
- macOS x86_64 due to tk and ncurses not supporting cross-building.
  `tk/8.6.10: Invalid: The tk conan recipe does not currently support Macos cross-builds. A contribution to add this functionality would be welcome.`
  `ncurses/6.4: Invalid: Cross building to/from arm is (currently) not supported`
- cpython/*:shared=False on MSVC for v3.10 and greater due to build errors.
  `cpython/3.12.7: Invalid ID: Static msvc build disabled (>=3.10) due to "AttributeError: module 'sys' has no attribute 'winver'"`

Both of these are avoidable.

This affects the build of these configurations for `nanobind`, for example: https://github.com/conan-io/conan-center-index/pull/20297#issuecomment-2390819786

/cc @Ahajha

#### Details

---

### DIFF
--- a/recipes/cpython/all/conanfile.py
+++ b/recipes/cpython/all/conanfile.py
@@ -66,7 +66,7 @@ class CPythonConan(ConanFile):
 
     @property
     def _supports_modules(self):
-        return not is_msvc(self) or self.options.shared
+        return not is_msvc(self) or self.options.get_safe("shared")
 
     @property
     def _version_suffix(self):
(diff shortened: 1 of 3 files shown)


Ahajha:
I like the goal here. There's another invalid config, relating to `mpdecimal` on Windows/shared. The `cxx` option is enabled by default, but is incompatible with the `shared` option on Windows, so the Windows/all shared build is invalid. So probably want to also just disable the `cxx` option, at least on Windows.

Fixing the issues that come up from these might not be trivial, but I can take a look if I have some spare time.

recipes/cpython/all/conanfile.py
```diff
@@ -98,6 +103,12 @@ def configure(self):
             self.options.rm_safe("with_sqlite3")
             self.options.rm_safe("with_tkinter")
             self.options.rm_safe("with_lzma")
+        if is_msvc and Version(self.version) >= "3.10":
+            # Static CPython on Windows is only loosely supported, see https://github.com/python/cpython/issues/110234
+            # 3.10 fails during the test, 3.11 fails during the build (missing symbol that seems to be DLL specific: PyWin_DLLhModule)
+            # Disabling static MSVC builds (>=3.10) due to "AttributeError: module 'sys' has no attribute 'winver'"
+            self.package_type = "static-library"
```
Ahajha, line 110: Shouldn't this be `shared-library`?

valgur: Absolutely. Thanks!

Ahajha:
@AbrilRBS To fix the Mac cross-compile errors, we would need an additional build-time dependency on `cpython` itself. Can CCI support this?

Ahajha:
@valgur Seems that cpython doesn't support Mac cross compiles, but it does support universal builds (see https://github.com/python/cpython/issues/90905). That might go into a rabbithole of needing to update dependencies to also support universal builds.

PR #25532, opened 2024-10-06
opensubdiv: Enable opengl by default and metal for MacOS

EstebanDugueperoux2:
### Summary

#### Details

---


recipes/opensubdiv/all/conanfile.py
```diff
@@ -33,14 +33,14 @@ class OpenSubdivConan(ConanFile):
     default_options = {
         "shared": False,
         "fPIC": True,
-        "with_tbb": False,
-        "with_opengl": False,
-        "with_omp": False,
+        "with_tbb": True,
```
AbrilRBS, line 36: I need a bit more insight into this. Are there the default values in upstream? Or is it just that these values are more convinient to any user for the recipe, and specifically your releated PR? Thanks!

AbrilRBS (commented):
Sorry I didn't ping you that this was relaunched, but I see you already saw it! I have one question regarding the changes to the default values :)

recipes/opensubdiv/all/conanfile.py
EstebanDugueperoux2: Hi @AbrilRBS , in https://github.com/PixarAnimationStudios/OpenSubdiv/blob/release/CMakeLists.txt#L323 we can see that by default all options are enabled with a double negation.

EstebanDugueperoux2: This is a prerequisite of https://github.com/conan-io/conan-center-index/pull/24506 on which there is yet many works, notably plugins managment in package_info() for which I have send you a private message (https://cpplang.slack.com/archives/D071ARHMDL0/p1727905406255289).

EstebanDugueperoux2:
Hi @jcar87 @AbrilRBS , any news about this PR?

EstebanDugueperoux2:
Hi @jcar87 ,

I have updated PR description about the motivation of this change.

Regards.

jcar87 (commented):
Thanks for your contribution.
We try to be conservative when it comes to changing the default options of recipes, and we also are very wary of approving PRs where there are changes that are not explained. 

We have a few concerns when reviewing this - this PR says "enable opengl by default and metal for macOS" but 
- this enables opengl by default on Linux, and disables the option altogether on other OSs,  however I can't see this reflected in the upstream scripts. OpenGL is searched for in all platforms (https://github.com/PixarAnimationStudios/OpenSubdiv/blob/7d0ab5530feef693ac0a920585b5c663b80773b3/CMakeLists.txt#L358-L361) and the only platform when it may be ignored is IOS (https://github.dev/PixarAnimationStudios/OpenSubdiv/blob/release/CMakeLists.txt) - so just trying to understand why this PR would effectively limit it to Linux only 
- There is no mention in the PR description or otherwise as to why `with_tbb` and `with_omp` are being enabled - it may be that there are valid reasons, however - we ask contributors that unless it's strictly related to the goal of the PR (enabling opengl/metal) then please do not make additional changes, especially without an explanation - we have a very difficult task of reviewing a large number of pull requests - the quickest way to get your PR merged is to make the smallest possible PR that satisfies your objective. We have had issues in the past with "unrelated" changes causing unnecessary breakages and disruption.
- I'm a bit confused as to how this is a pre-requisite of https://github.com/conan-io/conan-center-index/pull/24506 - if downstream consumers in Conan Center require builds with non default options - it should be possible to handle that in a way that CI will not end up with a failure - maybe it won't run the build, but this won't hold up a review from us - we will likely just test it locally and attach the build logs as we have done in the past
- OpenUSD does not require this recipe to have OpenMP or TBB enabled at all - can't find any references to uses of the Opensubdiv TBBEvaluator or OMPEvaluator (this would be headers  `<opensubdiv/osd/ompEvaluator.h>` and `<opensubdiv/osd/tbbEvaluator.h>`

EstebanDugueperoux2:
Hi @jcar87,

thanks for the review, I have removed useless changes, omp and tbb options enablements were for optimisation.
About openGL option, you can see in https://github.com/conan-io/conan-center-index/pull/24506 a failing build:

https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-24506/3/package_build_logs/build_log_openusd_25_02_a3be653c5e9066140da394c450506c76_7b82da7ea4cc5f6a1ee7b904b93e072be6f83f52.txt

were openusd build failed because it can find opensubdiv/osd/glslPatchShaderSource.h which seems available only with opensubdiv:*:with_opengl=True.

This is the main reason of this PR.

Regards.

jcar87:
Thanks for the clarifications and the motivation, this helps a lot!!
Will review this alongside the team, cc @uilianries

perseoGI (approved):
Greetings @EstebanDugueperoux2 
First I want to say thank you for your contribution.

The original intention of this PR (fixing CI compilation on https://github.com/conan-io/conan-center-index/pull/24506) is legitim but we, as maintainers, have other methods to fix the missing binaries as @jcar87 said. 

Nevertheless, this PR is fixing an OpenGL error on `OS != Macos` due to missing requirements.
I've tested this recipe locally in different operating systems and now it works well with OpenGL option enabled.

recipes/opensubdiv/all/conanfile.py
```diff
@@ -86,7 +86,14 @@ def requirements(self):
...
                 self.requires("onetbb/2020.3.3", transitive_headers=True)
             else:
-                self.requires("onetbb/2021.10.0", transitive_headers=True)
+                self.requires("onetbb/2021.12.0", transitive_headers=True)
+        if self.options.get_safe("with_opengl"):
+            self.requires("opengl/system")
+            self.requires("glfw/3.4")
+        if self.options.get_safe("with_metal"):
+            self.requires("metal-cpp/14.2")
+        if self.options.get_safe("with_dx"):
```
perseoGI, line 95: On the other hand @EstebanDugueperoux2,
Why is this requirement needed?
I've tested locally on a Windows machine and see no difference on build logs when compiling with `with_dx=True` on both `master` and this PR branch.

Also, if you enable `examples` (line 140), and `with_dx`, you will have a compilation error:

```
d3d11ShaderCache.obj : error LNK2001: unresolved external symbol D3DCompile [C:\Users\perseo\.conan2\p\b\opensec2a33492099e\b\build\examples\dxViewer\dxViewer 
.vcxproj]
```

Do you have any idea?
Thank you!

EstebanDugueperoux2: Hi @perseoGI ,
thanks for you test, I have no windows machine on my side.
Indeed DX11SDK seems used only by examples and as this last are disabled by default, I'll rollback my changes about with_dx option.

perseoGI: Greetings @EstebanDugueperoux2,
Thank you for your rollback.
The easiest way to get a PR reviewed and merged is having all the changes in the PR under control and with a justification & description in the details box.
If you consider those DX were needed, it may be easier to open another PR after this one gets merged with a more descriptive title and information.

Thank you in advance!
I'm approving again the PR 🐸

perseoGI (approved):

PR #25543, opened 2024-10-07
ffmpeg: disable autodetected deps in ./configure, add libsndio support, use non-system libvdpau

valgur:
### Summary
Changes to recipe:  **ffmpeg/[*]**

#### Motivation
Static build of ffmpeg failed for me locally due to libsndio autodetection during ./configure picking up a non-Conan system version of the library.

#### Details
- Disabled all dependency autodetections with `--disable-autodetect` ./configure flag to avoid cases like in the future. The flag is available all the way back to the oldest v4.4.4 on CCI.
- Added support for libsndio. Disabled it by default.
- Replaced `vdpau/system` with `libvdpau/1.5`, which was added a few months ago: #22236.
- Added a version range for pkg-config.

---


valgur:
Duplicate of #26559.

PR #25552, opened 2024-10-07
turtle: add new version 2.0.0

toge:
### Summary
Changes to recipe:  **turtle/2.0.0**

#### Motivation
turtle/2.0.0 requires C++14.

#### Details
https://github.com/mat007/turtle/compare/v1.3.2...v2.0.0

---

### DIFF
--- a/recipes/turtle/all/conandata.yml
+++ b/recipes/turtle/all/conandata.yml
@@ -1,7 +1,10 @@
 sources:
-  "1.3.1":
-    url: "https://github.com/mat007/turtle/archive/v1.3.1.tar.gz"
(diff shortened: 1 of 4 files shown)


AbrilRBS (commented):
The error comes from `apple-clang` having a default cppstd of `gnu98`, so we would need to force it to 14.

My recommendation is to use upstream's CMakeLists's to handle the install, and in a `generate()` method, add the workaround lines of 
```py
tc = CMakeToolchain(self)
if not self.settings.compiler.cppstd:
```
for versions >= 2.0

Wdyt @toge? :)

ErniGH (approved):
Thanks

AbrilRBS:
Oh @toge I didn't catch it's a header-only recipe, good catch updating the test_package!

AbrilRBS (approved):

Issue #25565, opened 2024-10-08
[package] libarchive/3.7.4 (and other versions too): Calling convention conflict on Windows x86 with zlib.


Adnn:
### Description

We are facing a `__cdecl` vs `__stdcall` issue in one of our projects on Windows when targeting x86. This projects requires both zlib/1.2.13 and libarchive/3.7.4.

Basically, when linking the final application, we get linker errors such as:

> archive.lib(archive_write_set_format_7zip.obj) : error LNK2019: unresolved external symbol _crc32@12 referenced in func
> tion __7z_close [our.vcxproj]
> archive.lib(archive_write_set_format_zip.obj) : error LNK2001: unresolved external symbol _crc32@12 [our.vcxproj]
>     Hint on symbols that are defined and could potentially match:
>       _crc32

To confirm the problem `dumpbin /SYMBOLS zlib.lib` list defined symbols such as `crc32`, whereas `dumpbin /SYMBOLS archive.lib` shows UNDEF symbols such as `_crc32@12`.

It appears that the recipe for zlib on Conancenter does not define ZLIB_WINAPI macro, resulting in `__cdecl` convention for the library. Yet, digging into libarchive recipe [we found patches](https://github.com/conan-io/conan-center-index/blob/08d6b641f8a6b9c95379986c846c447b95a2b786/recipes/libarchive/all/patches/0001-3.7.4-zlib-winapi.patch) that override a test intended to deduce if zlib was compiled with ZLIB_WINAPI (see [zlibCompileFlags() in the manual](https://zlib.net/manual.html)), and always define the macro instead, creating an inconsistency in this situation.

### Package and Environment Details

* Package Name/Version: **libarchive/3.7.4**
* Operating System+version: **Windows 10**
* Compiler+version: **msvc**
* Conan version: **conan 2.5.0**

### Conan profile

[settings]
arch=x86
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=static
compiler.runtime_type=Release
compiler.version=193
os=Windows
[options]
libarchive/*:with_lzma=True
[tool_requires]
*: cmake/3.27.7

### Steps to reproduce

Please see PR #25567.

### Logs

<details><summary>Click to expand log</summary>

```
archive.lib(archive_write_set_format_7zip.obj) : error LNK2019: unresolved external symbol _crc32@12 referenced in func
tion __7z_close [our.vcxproj]
[...]

PR #25567, opened 2024-10-08
Extend test_package code to trigger linkage against upstreams.

Adnn:
### Summary
Changes to recipe:  **libarchive/[all]**

#### Motivation
This PR extends the test package to reproduce issue #25565 

#### Details
The test_package code calls `archive_read_support_filter_all()` which triggers linkage against upstreams used to support compression formats.
This should trigger the linker error described in #25565 when building on Windows `arch=x86`.

---

### DIFF
--- a/recipes/libarchive/all/test_package/test_package.c
+++ b/recipes/libarchive/all/test_package/test_package.c
@@ -4,5 +4,11 @@
 int main(int argc, char *argv[]) {
     printf("libarchive version: %s\n", archive_version_string());
     printf("libarchive details: %s\n", archive_version_details());
+
+    // Further test correct linkage against upstreams
+    struct archive * a = archive_read_new();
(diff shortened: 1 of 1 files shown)


Adnn:
We are happy to also work on actually fixing the issue.

What we currently do locally is that we plainly disable patch `0001-zlib-winapi`, and things are fine for all our supported environments. Yet before simply adding a commit disabling the patch to this PR, we might need to hear the rationale behind it.

jcar87:
Hi @Adnn  - is this PR still needed given https://github.com/conan-io/conan-center-index/issues/25565? (we tend to keep test package C/C++ code as simple as can be), thanks!

jcar87:
Closing due to inactivity - but please let us know if there is anything to do on our end! Thanks

PR #25568, opened 2024-10-08
msgpack-cxx: change cmake config file name

uyha:
Upstream README is wrong, a PR is already created there to fix it.

### Summary

#### Details

---


AbrilRBS (commented):
Thanks @uyha! I'll need to double-check with the team what the best approach to avoid breaking potential users of the old buggy name would be (Probably a `cmake_target_alias` property could be set!), thanks for your patience :)

AbrilRBS (approved):
@uyha sorry for the delay, this was trickier than expected to get right.

The final approach will be to keep the old buggy name for old published versions, and only from a future new version onwards to fix the name, as not to break current consumers with the new revision.

Thanks a lot for taking the time to create the PR in the first place, we appreciate it :)

AbrilRBS:
@uyha do you plan on opening a PR upstream to fix their readme?

ErniGH (approved):
Thanks! 😄

recipes/msgpack-cxx/all/conanfile.py
```diff
@@ -24,15 +24,16 @@ class MsgpackCXXConan(ConanFile):
     }
     no_copy_source = True
 
-    def configure_options(self):
-        # No boost was added since 4.1.0
+    def config_options(self):
```
AbrilRBS, line 27: The option was not being removed because the method was not named correctly!

AbrilRBS (approved):

ErniGH (approved):

uyha:
> @uyha do you plan on opening a PR upstream to fix their readme?

I did, https://github.com/msgpack/msgpack-c/pull/1142

uyha:
> @uyha sorry for the delay, this was trickier than expected to get right.
> 
> The final approach will be to keep the old buggy name for old published versions, and only from a future new version onwards to fix the name, as not to break current consumers with the new revision.
> 
> Thanks a lot for taking the time to create the PR in the first place, we appreciate it :)

But isn't introducing a mechanism for aliasing better? it introduces an opporunity to fix buggy situations like this.

PR #25569, opened 2024-10-08
rabbitmq-c: fix a failing try_compile()

valgur:
### Summary
Changes to recipe:  **librabbitmq-c/[*]**

#### Motivation
Required for https://github.com/alanxz/rabbitmq-c/blob/v0.14.0/CMakeLists.txt#L132-L133
Otherwise the test fails with "CheckSymbolExists.c:(.text+0x1f): undefined reference to `ENGINE_new'".

#### Details

---

### DIFF
--- a/recipes/rabbitmq-c/all/conanfile.py
+++ b/recipes/rabbitmq-c/all/conanfile.py
@@ -67,6 +67,9 @@ class RabbitmqcConan(ConanFile):
         tc.variables["ENABLE_SSL_SUPPORT"] = self.options.ssl
         tc.variables["BUILD_API_DOCS"] = False
         tc.variables["RUN_SYSTEM_TESTS"] = False
+        # Required for https://github.com/alanxz/rabbitmq-c/blob/v0.14.0/CMakeLists.txt#L132-L133
+        # Otherwise fails with "CheckSymbolExists.c:(.text+0x1f): undefined reference to `ENGINE_new'"
+        tc.variables["CMAKE_TRY_COMPILE_CONFIGURATION"] = str(self.settings.build_type)
         tc.generate()
 
         if self.options.ssl:


AbrilRBS (approved):

jcar87:
This codepath https://github.com/alanxz/rabbitmq-c/blob/v0.14.0/CMakeLists.txt#L132-L133 isn't tested during a CI build because openssl support is set of off - could we attach evidence of what happens without this fix, and what the logs look like with the fix?

valgur:
See the comment in the recipe, please.

jcar87:
The reason we ask for evidence of a successful and a failing log is because it happens quite often that a seemingly trivial PR claiming to fix something doesn't actually fix it. It would massively speed up the review process for us if we could look at a log, rather than have to build this ourselves locally which is what we would do. 

Here's a good example of a before and after: https://github.com/conan-io/conan-center-index/pull/25781

valgur:
I'll wait for the new CMakeDeps instead.

jcar87:
I'm well aware of the lack of adverse side effects. It's less about the potential side effects, and more about ensuring that when with that option is enabled, there isn't a problem further down the line. This happens in a lot of cases, we get contributors (and it happens to us as well)  - opening PRs with trivial changes that look very likely to solve a problem and then when we actually test it, we still can't build a package from the recipe.
Sometimes, we discover that while it does fix a particular issue, because it happens when we set an option to a non-default value  - perhaps the build with that option fails somewhere else, so there are other fixes. So it makes it easier to get a PR: "fix building with  ssl=True", here's the logs before and after. - rather than having to guess whether this was already an issue and it was wrong in the default configuration, or whether the build succeeds with `ssl=True` after this change.

Here's a recent example: https://github.com/conan-io/conan-center-index/pull/25850 - this PR fixes an actual issue, and links to the documentation to support it and I would approve it based on that alone, and yet, the recipe still doesn't build on Android.

PR #25570, opened 2024-10-08
glog: fix a try_compile() instead of disabling it

valgur:
### Summary
Changes to recipe:  **glog/[*]**

#### Motivation
Fix the try_compile() with a gflags dependency instead of disabling it. This one: https://github.com/google/glog/blob/v0.7.1/cmake/DetermineGflagsNamespace.cmake#L37

There's another instance of try_compile() against Libunwind::Libunwind, but that works thanks to that target being defined by the project itself instead of CMakeDeps. Still a bit fragile, though.

#### Maintainer changes
- Stop publishing new revisions for older versions
- Remove test_v1_package
- Refactor `_compilers_minimum_version` to use `check_min_cppstd()` instead
- Remove/refactor conditionals for versions no longer maintained in the recipe
- remove patches that are no longer used

---


jcar87 (approved):
Thanks!

PR #25574, opened 2024-10-09
watcher-c: add recipe

toge:
### Summary
Changes to recipe:  **watcher-c/0.12.1**

#### Motivation
watcher/0.12.0 starts to provide c binding.

#### Details

---

### DIFF
--- a/recipes/watcher-c/all/conandata.yml
+++ b/recipes/watcher-c/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "0.12.1":
+    url: "https://github.com/e-dant/watcher/archive/release/0.12.1.tar.gz"
+    sha256: "aab03c11d729b998f16a97fd6f9c6e0579466f4f0d208650117159818a11cf20"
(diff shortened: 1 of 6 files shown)


AbrilRBS:
Hi @toge, I see that we already provide a C++ version of this recipe - why did you decide to create a different PR?

jcar87:
> Hi @toge, I see that we already provide a C++ version of this recipe - why did you decide to create a different recipe?

the C++ version is a header only library, this one is actually built - if both were header only l think keeping the existing recipe would be okay.

We do have other recipes that have C++ and C libraries both in the same recipe, or in separate recipes.

Either way, I would suggest holding off on this until this is requested by users. 

Closing this, as looking at the repository, the C bindings are quite new and not included in any distro I can see, the version is still 0.x, and it's unclear exactly what is the primary build system (the Readme doesnt mention Meson, the CMake scripts do seem to support the C bindings). When a library supports multiple build systems, its really difficult to maintain full parity across them, so we need to be conservative. Meson soname/filename conventions have caused us enough headaches as it is.

We will of course reconsider if/when this is requested by users - thanks for your contribution!

PR #25588, opened 2024-10-10
Add rand_seed parameter to openssl 3.x.x

mefff:
### Summary
Changes to recipe:  openssl/3.x.x

#### Motivation
I need to customize this flag for my openssl package.

#### Details
Add option flag to customize the `--with-rand-seed` openssl configuration flag. More info on the flag on the openssl documentation: https://github.com/openssl/openssl/blob/master/INSTALL.md#seeding-the-random-generator

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -89,12 +89,14 @@ class OpenSSLConan(ConanFile):
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
         "tls_security_level": [None, 0, 1, 2, 3, 4, 5],
+        "rand_seed": ["none", "getrandom", "devrandom", "os", "egd", "rdcpu"],
     }
     default_options = {key: False for key in options.keys()}
     default_options["fPIC"] = True
     default_options["no_md2"] = True
     default_options["openssldir"] = None
     default_options["tls_security_level"] = None
+    default_options["rand_seed"] = "os"
 
     @property
     def _is_clang_cl(self):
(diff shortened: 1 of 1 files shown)


mefff:
hey @AbrilRBS it seems like the CLA bot it's stuck, I already signed it, is there any way to retrigger the job? Also not sure about the `continuous-integration/jenkins/pr-merge` can't see the details, not allowed to.

uilianries:
@mefff Thank you for your PR! 😄

uilianries:
@mefff Thank you for your PR! 😄 

Since Conan 1.17.0 it's possible to pass extra configure arguments without touching the Conan recipe: https://docs.conan.io/2/reference/config_files/global_conf.html

> tools.gnu:extra_configure_args: List of extra arguments to pass to configure when using AutotoolsToolchain and GnuToolchain

For your case, you can do:

```
conan install -r conancenter --requires=openssl/3.14.0 --build=openssl/3.14.0 -c tools.gnu:extra_configure_args='["--with-rand-seed=os"]'
```

The same can be passed to your Conan profile as well:

```ini
include(default)

[conf]
openssl/* :tools.gnu:extra_configure_args=["--with-rand-seed=os"]
```

Still, this configuration will not affect the package ID of the built package for OpenSSL, instead, you need to use the configuration `tools.info.package_id:confs` to make it count:

```
-c tools.info.package_id:confs='["tools.gnu:extra_configure_args"]'
```

Then your package ID will be unique for each extra configure argument.

----

As you may have noted, the OpenSSL recipe in ConanCenterIndex has +60 options, which results in an impossible scenario to be tested using all combinations. When adding rand seed option, the CI will not validate using other values, but only the default, so the reviewer and the PR author should make sure about those results when using a different value to avoid breaking users. Please, feel free to share your thoughts about using the configuration.

PR #25591, opened 2024-10-11
Add exprtk/latest

Ahajha:
Draft - the mirror will be updated soon. Will update when that happens.

~~The mirror at https://github.com/ArashPartow/ is no longer being updated. As far as I'm aware, the code is only available @HEAD at https://www.partow.net/downloads/. Not sure if there's a better way to deal with the single-version-available-at-a-time situation.~~

### DIFF
--- a/recipes/exprtk/all/conandata.yml
+++ b/recipes/exprtk/all/conandata.yml
@@ -1,4 +1,6 @@
 sources:
+  "latest":
+    url: "https://www.partow.net/downloads/exprtk.zip"
   "0.0.2":
     url: "https://github.com/ArashPartow/exprtk/archive/0.0.2.tar.gz"
     sha256: "7e8de4a0bfc9855c1316d8b8bc422061aef9a307c2f42d2e66298980463195c1"
--- a/recipes/exprtk/config.yml
+++ b/recipes/exprtk/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "latest":
+    folder: "all"
   "0.0.2":
     folder: "all"
   "0.0.1":


HazyMrf:
Thank you so much, @Ahajha! Currently our project uses "patched" exprtk for clang-19 compliance. Waiting till this PR will be merged to switch to official Conan package as soon as possible!

recipes/exprtk/all/conandata.yml
```diff
@@ -1,4 +1,6 @@
 sources:
+  "latest":
+    url: "https://www.partow.net/downloads/exprtk.zip"
```
jcar87, line 3: Unfortunately - if a library doesn't provide sources that are sha/content stable, it is not eligible for inclusion in Conan Center Index - as it does not allow for content validation/verification. This is particularly a problem when the code comes some a self-hosted website, rather than a well known source (e.g. github, etc).

Please do reach out to the author in case this is something they are willing to address - otherwise, I'd imagine the only alternative accept is a github mirror that republishes `0.0.2` in a way that the sha is stable, and the contents are verifiably the same as https://github.com/ArashPartow/exprtk/archive/0.0.2.tar.gz.

(We do have it as a task to address the only other case of this in Conan Center, which is the msys2 recipe)

Ahajha: I've been in contact with the author, they will be making a new version at some point. When they do I will update the recipe.

ArashPartow:
@Ahajha when you have a moment can you please close this PR.

Ahajha:
@ArashPartow I'm leaving it in draft for now while I wait - it won't get merged as is. But I can fully close it if you would like.

PR #25593, opened 2024-10-11
xerces-c: fix handling of nsl library

jcar87:
### Summary
Changes to recipe: xerces-c all

#### Motivation
nsl is not a system library that is installed by default or found in all distros - so it shouldn't be listed as a system dependency without further safeguards.

#### Details
However, nsl appears to be completely unused even when found by the build system (there are no evaluations of the `HAVE_LIBNSL` macro, and no uses of any headers in the `rpcsvc/*` folder. 
On systms where gcc is configured to pass `--as-needed` to the linker, the dependency is never encoded even if libnsl was found.

### DIFF
--- a/recipes/xerces-c/all/conanfile.py
+++ b/recipes/xerces-c/all/conanfile.py
@@ -120,6 +120,10 @@ class XercesCConan(ConanFile):
         tc = CMakeToolchain(self)
         # Because upstream overrides BUILD_SHARED_LIBS as a CACHE variable
         tc.cache_variables["BUILD_SHARED_LIBS"] = "ON" if self.options.shared else "OFF"
+
+        # Prevent linking against unused found library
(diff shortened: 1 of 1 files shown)


jcar87:
> On systms where gcc is configured to pass --as-needed to the linker, the dependency is never encoded even if libnsl was found.

We ensured this was the case for our images in https://github.com/conan-io/conan-docker-tools/pull/584, however I suspect some of the Conan 1.x docker images are still legacy and not configured with this - otherwise the hook warning would *not* have been triggered.

I can confirmed that the Conan 2.x packages we published *DO NOT* encode a dependency on libnsl for the shared variant - so that's another indicator that it is not needed.

also the upstream .pc file does not mention this:
https://github.com/apache/xerces-c/blob/53c16411466bf90c62617831fe92ed0f41e70882/xerces-c.pc.in#L10

AbrilRBS (approved):

perseoGI (approved):

PR #25594, opened 2024-10-11
Fix: exceptions during build of bitserializer

jrosiek:
### Summary
Changes to recipe:  **bitserializer/[>=0.50]**

#### Motivation
The build fails when two variants of the package are build on the same machine (with different `with_rapidaml` option value).

#### Details
The same source location is used to build different variants of package. In effect, the same file is patched more than once, what causes an exception. 

---

### DIFF
--- a/recipes/bitserializer/all/conanfile.py
+++ b/recipes/bitserializer/all/conanfile.py
@@ -134,8 +134,11 @@ class BitserializerConan(ConanFile):
         if Version(self.version) >= "0.50":
             # Remove 'ryml' subdirectory from #include
             replace_in_file(
-                self, os.path.join(self.source_folder, "include", "bitserializer", "rapidyaml_archive.h"),
-                "#include <ryml/", "#include <",
+                self, 
+                os.path.join(self.source_folder, "include", "bitserializer", "rapidyaml_archive.h"),
+                "#include <ryml/", 
+                "#include <",
+                strict=False
             )
 
     def build(self):


jcar87:
Hi @jrosiek, thanks for your contribution.

Could you provide more details as to which exceptions you get and how to reproduce the errors on our end? Thanks :)

jrosiek:
Here you go:
```
conan remove 'bitserializer/*'
conan install --requires=bitserializer/0.70 -o '&:with_csv=True' -o '&:with_rapidyaml=True' --build=missing
conan install --requires=bitserializer/0.70 -o '&:with_csv=True' -o '&:with_rapidyaml=False' --build=missing
```

The `with_csv` is required for the recipe to ensure a new package_id is generated for each `install` call; otherwise it is treated as header-only and the second call would do nothing.

AbrilRBS:
@jrosiek thanks a lot for your PR. Could you try checking if the current changes also solve your issue? Thanks!

jrosiek:
Yes, the change fixes the issue on my side. Thanks.

AbrilRBS (approved):

jcar87:
I'm just double checking the very first commit of this recipe that contained the `no_copy_source`: https://github.com/conan-io/conan-center-index/pull/1599/commits/d4f8bc28fcec2c0239140424b8d93d15d4053797 

and no context is provided to justify it  - it is useful in some cases, but don't think it is necessary here

jrosiek:
Any chance for a second review?

jcar87 (approved):

jrosiek:
Thank you

PR #25596, opened 2024-10-11
libarchive: remove outdated patch

jcar87:
Fix https://github.com/conan-io/conan-center-index/issues/25565

This patch should no longer be needed as per https://github.com/jcar87/conan-center-index/blob/2b89e705a66320ac5185fe9e0e866fb5ea78218e/recipes/libarchive/all/conanfile.py#L165


AbrilRBS (approved):

ErniGH (approved):
LGTM

Issue #25585, opened 2024-10-10
[request] open62541/1.4.6


ErikTempelaarVO:
### Package Name/Version

open62541/1.4.6

### Changelog

https://github.com/open62541/open62541/releases/tag/v1.4.6

### Context about the new update

The above-mentioned version is newly released by the upstream project and not yet available as a recipe.
Please add this version.

Latest version available as recipe is 1.3.9 and dates back to november '23.
In the meantime, the project has released quite a few updates, in branches for 1.3.x and 1.4.x (and even in 1.2.x)

Could be nice to have (some of) these available in conan, this allows using these updated versions in the open62541pp that is in the works here: #23291

toge:
@ErikTempelaarVO 
I have a question about this issue.
I have created a PR about this issue.

I don't know the functionality of open62541, so I just created a PR for a successful build with 1.4.6.
I would appreciate if you could comment on the content of the PR.

PR #25598, opened 2024-10-11
open62541: add version 1.4.6

toge:
### Summary
Changes to recipe:  **open62541/1.4.6**

#### Motivation
Try to close https://github.com/conan-io/conan-center-index/issues/25585.

#### Details
https://github.com/open62541/open62541/compare/v1.3.9...v1.4.6

---


AbrilRBS:
@toge you might be missing bcrypt system_lib for the new version under Windows for the current compilation issue?

toge:
@AbrilRBS 
Thank you for your advice.
I solved this issue to include "iphlpapi.h".
I will try to create PR for upstream.

AbrilRBS (approved):

ErikTempelaarVO:
Hey, @toge, thanks for updating this recipe.
Unfortunately, it doesn't work for me on MinGW Windows.

I believe the new patch is the culprit, because locally I can compile open62541 v1.4.6 just fine.

```
#if defined(_WIN32)
#  include <winsock2.h>
#  include <iphlpapi.h>
#endif
```

When I apply the patch the build breaks. Maybe this patch is too generic?
I guess it should only apply to builds with MSVC, and not any other _WIN32?

Any thoughts, @toge @AbrilRBS ?

ErikTempelaarVO:
According to this [issue](https://github.com/open62541/open62541/issues/6423)
one should #include windows.h (that will #include winbase.h and that contains the declaration)

I manually changed the existing 0004-include-iphlpapi.patch to:
```
diff --git a/src/ua_util.c b/src/ua_util.c
index 13eb0fb..d832f0a 100644
--- a/src/ua_util.c
+++ b/src/ua_util.c
@@ -11,6 +11,10 @@
  * unit for the generated code from UA_INLINABLE definitions. */
 #define UA_INLINABLE_IMPL 1
 
+#if defined(_WIN32)
+//#  include <winsock2.h>
+#  include <windows.h>
+#endif
 #include <open62541/types_generated_handling.h>
 #include <open62541/util.h>
 

```

Exported the library into my local cache and was able to successfully build in MinGW via conan

PR #25623, opened 2024-10-14
lief: add version 0.15.1

toge:
### Summary
Changes to recipe:  **leif/0.15.1**

#### Motivation
There are lots of improvements since 0.13.1.

#### Details
https://github.com/lief-project/LIEF/compare/0.13.1...0.15.1

Changelog
https://lief.re/doc/stable/changelog.html#july-23th-2024

---

### DIFF
--- a/recipes/lief/all/conanfile.py
+++ b/recipes/lief/all/conanfile.py
@@ -65,7 +65,7 @@ class LiefConan(ConanFile):
                 "msvc": "191",
             },
             "17": {
-                "gcc": "8",
+                "gcc": "8" if Version(self.version) < "0.15.1" else "10",
                 "clang": "7",
                 "apple-clang": "12",
                 "Visual Studio": "16",
--- a/recipes/lief/all/conanfile.py
+++ b/recipes/lief/all/conanfile.py
@@ -50,6 +50,8 @@ class LiefConan(ConanFile):
 
     @property
     def _min_cppstd(self):
+        if Version(self.version) >= "0.15.1":
+            return "17"
         return "14" if self.options.with_frozen else "11"
 
     @property
@@ -61,6 +63,13 @@ class LiefConan(ConanFile):
                 "apple-clang": "10",
                 "Visual Studio": "15",
                 "msvc": "191",
+            },
+            "17": {
+                "gcc": "8",
+                "clang": "7",
+                "apple-clang": "12",
+                "Visual Studio": "16",
+                "msvc": "192",
             },
         }.get(self._min_cppstd, {})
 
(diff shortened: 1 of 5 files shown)


AbrilRBS:
Any insight into this build failure @toge?

recipes/lief/all/conanfile.py
```diff
@@ -79,11 +69,16 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("mbedtls/3.2.1")
+        # TODO: Check if we actually need transitive_headers
+        # LIEF.lib(x509.obj) external symbol mbedtls_snprintf referenced in function lief_mbedtls_x509_dn_gets
+        self.requires("mbedtls/3.2.1", transitive_headers=True, transitive_libs=True)
```
jcar87, line 74: @AbrilRBS - this is not a case of transitive_libs (not even transitive_headers), as the library is already added in the linker statement by conan when creating a static library.

```
1>  LIEF.lib(x509.obj) : error LNK2019: unresolved external symbol mbedtls_snprintf referenced in function "int __cdecl `anonymous namespace'::lief_mbedtls_x509_dn_gets(char *,unsigned __int64,struct mbedtls_asn1_named_data const *)" (?lief_mbedtls_x509_dn_gets@?A0x25ca5665@@YAHPEAD_KPEBUmbedtls_asn1_named_data@@@Z)
1>  conan-center-index\recipes\lief\all\test_package\build\msvc-194-x86_64-17-release\Release\test_package.exe : fatal error LNK1120: 1 unresolved externals
1>Done building target "Link" in project "test_package.vcxproj" -- FAILED.
```

The implementation of lief, is calling into `mbedtls_snprintf` which _should_ be in `mbedx509.lib` - which is linked in the test package that is failing.

The problem is the logic inside `platform.h` in mbedtls - that function is defined to be one thing or the other, depending on the macro deductions in that file - so this impacts how mbedtls is built, and what value the macros resolve to when mbedtls is consumed - and presumably they are different at those two points, that's why the "wrong" symbol is requested

the culprit is in the mbedtls recipe:
https://github.com/luzpaz/conan-center-index/blob/6ed97106162ee022c6f0db0db4478bd18bca5227/recipes/mbedtls/all/conanfile.py#L90-L94

Which is puzzling as that should not be needed if the header does the autodetuction by itself, and it is not mirroring anything that the upstream CMake does (the generated cmake files also dont propagate any compiler definitions). Perhaps you have more insights since you were involved in approving that PR - I may be missing something.

jcar87: (addressed in https://github.com/conan-io/conan-center-index/pull/26299)

jcar87:
(superseded by https://github.com/conan-io/conan-center-index/pull/26300 - preserving the commits in this PR)

PR #25641, opened 2024-10-17
Fix compilation error on non-msvc compilers without gnu extensions enabled

jrosiek:
### Summary
Changes to recipe:  **pcapplusplus/[>=23.09]**

#### Motivation
The package fails to compile on non-msvc compiler when `cppstd` is not set to `gnuXX`.
The error looks like this:
```
pcapplusplus/23.09: Invalid: The cppstd GNU extension is required 
```

Example profile:
```
[settings]
arch=x86_64
os=Linux
compiler=gcc
compiler.version=11
compiler.cppstd=20
compiler.libcxx=libstdc++11
build_type=Debug
```

#### Details
This PR simply removes the requirement for gnu extensions for any compiler.

---

### DIFF
--- a/recipes/pcapplusplus/cmake/conanfile.py
+++ b/recipes/pcapplusplus/cmake/conanfile.py
@@ -55,7 +55,7 @@ class PcapplusplusConan(ConanFile):
             raise ConanInvalidConfiguration(f"{self.ref} does not support Windows shared builds for now")
         if self.settings.compiler.cppstd:
             # popen()/pclose() usage
-            check_min_cppstd(self, self._min_cppstd, gnu_extensions=not is_msvc(self))
+            check_min_cppstd(self, self._min_cppstd)
         if self.settings.os not in ("FreeBSD", "Linux", "Macos", "Windows"):
             raise ConanInvalidConfiguration(f"{self.settings.os} is not supported")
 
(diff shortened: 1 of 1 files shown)


AbrilRBS (commented):
Thanks a lot for taking the time to create a PR addresing your issue, we appreciate it! :)

This was originally added becase upstream requires gnu extensions activated, and it's only working because this line in upstream https://github.com/seladb/PcapPlusPlus/blob/master/CMakeLists.txt#L123 is forcing the gnu extensions back on, so the recipe simply translated that requirement there.

For me it would be a bit suprising to get the extensions if I've told Conan not to use them, but I can also see why it would be a pain to have to set it individually for recipes like this. Let me talk with the team about it, see what the best approach would be

jrosiek:
Keep in mind that this flag in CMake does not force/require the extensions to be present: https://cmake.org/cmake/help/latest/prop_tgt/CXX_EXTENSIONS.html#prop_tgt:CXX_EXTENSIONS

> For some compilers, this results in adding a flag such as -std=gnu++11 instead of -std=c++11 to the compile line. This property is ON by default. 

So it's semantics is different than the conan's check.

AbrilRBS (approved):

jcar87 (approved):

AbrilRBS:
With
```
replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
                        "set(CMAKE_CXX_EXTENSIONS ON)",
                        "set(CMAKE_CXX_EXTENSIONS OFF)")
```
and `conan create . --version=24.09 -b=missing -s="&:compiler.cppstd=17"`, this works, approving, thanks!

jrosiek:
@AbrilRBS I'm not sure why you need patching. Does it fail for you on some compiler with my changes only?

PR #25646, opened 2024-10-18
minio-cpp: add recipe

toge:
### Summary
Changes to recipe:  **minio-cpp/0.3.0**

#### Motivation
The client library for minio (S3 compatible object storage).

#### Details
https://github.com/minio/minio-cpp/

---


recipes/minio-cpp/all/conanfile.py
```diff
@@ -0,0 +1,101 @@
...
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("curlpp/0.8.1.cci.20240530", transitive_headers=True)
+        self.requires("inih/58")
+        self.requires("nlohmann_json/3.11.3", transitive_headers=True)
+        self.requires("openssl/3.3.2")
+        self.requires("pugixml/1.14")
+        self.requires("zlib/1.3.1")
```
AbrilRBS, line 59: Openssl and zlib should use version ranges!

recipes/minio-cpp/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,10 @@
+#include <miniocpp/client.h>
+
+int main(int argc, char* argv[]) {
+  minio::s3::BaseUrl base_url("play.min.io");
+
+  minio::creds::StaticProvider provider(
+      "Q3AM3UQ867SPQQA43P2F", "zuf+tfteSlswRu7BJ86wekitnifILbZam1KYY3TG");
```
AbrilRBS, line 7: This reads a bit too much like IDs and connections, I'd simplify the test package a lot, a simple call would be ok, thanks!

AbrilRBS (commented):
Thanks @toge! I have a few minor requests, otherwise looks good!

recipes/minio-cpp/all/conanfile.py
toge: Thank you for your review.
This is the point I forgot to correct from my private recipe.
Fixed.

recipes/minio-cpp/all/test_package/test_package.cpp
toge: Simplified.

AbrilRBS (dismissed):
Thanks!

recipes/minio-cpp/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,9 @@
+#include <miniocpp/client.h>
+
+int main(int argc, char* argv[]) {
+  minio::s3::BaseUrl base_url("example.com");
+
+  minio::creds::StaticProvider provider("access_key", "secret_key");
+
+  minio::s3::Client client(base_url, &provider);
```
AbrilRBS, line 8: This only stores the attributes, connection is done thru method calls, 👍

uilianries (approved):
LGTM. I changed the CMakeDeps to customize the cmake targets, instead of patching the source.

AbrilRBS (approved):
Thanks!

PR #25647, opened 2024-10-18
sqlite3: fix compiling for MSVC with non UTF-8 code page

X1aomu:
### Summary
Changes to recipe:  **sqlite3**

#### Motivation
The upstream source files're encoded in UTF-8, but some non-ASCII characters may be not be decoded in the user's current code page (e.g. CP936). Thus we need to instruct the MSVC to use UTF-8.

#### Details
A failed compilation look like this. 

```
sqlite3/3.46.1: RUN: cmake --build "D:\cache\conan\p\b\sqlit063ff4ed98214\b\build" --config Release
MSBuild version 17.11.9+a69bbaaf5 for .NET Framework

  1>Checking Build System
  Building Custom Rule D:/cache/conan/p/b/sqlit063ff4ed98214/b/CMakeLists.txt
  sqlite3.c
  sqlite3.vcxproj -> D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\Release\sqlite3.lib
  Building Custom Rule D:/cache/conan/p/b/sqlit063ff4ed98214/b/CMakeLists.txt
  shell.c
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27256,1): warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlit
e3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27709,35): error C2001: newline in constant [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27710,5): error C2143: syntax error: missing ';' before 'const' [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27724,16): error C2065: 'zBom': undeclared identifier [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27724,26): warning C4047: '=': 'int' differs in levels of indirection from 'const char *' [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27791,18): error C2065: 'zBom': undeclared identifier [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27791,20): error C2065: 'zBom': undeclared identifier [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27791,20): warning C4047: 'function': 'const char *' differs in levels of indirection from 'int' [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27791,20): warning C4024: 'oPutsUtf8': different types for formal and actual parameter 1 [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27804,18): error C2065: 'zBom': undeclared identifier [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27804,20): error C2065: 'zBom': undeclared identifier [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27804,20): warning C4047: 'function': 'const char *' differs in levels of indirection from 'int' [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]
D:\cache\conan\p\b\sqlit063ff4ed98214\b\src\shell.c(27804,20): warning C4024: 'oPutsUtf8': different types for formal and actual parameter 1 [D:\cache\conan\p\b\sqlit063ff4ed98214\b\build\sqlite3-bin.vcxproj]

sqlite3/3.46.1: ERROR:
Package 'c258ebf002cfa8d3ee0d78ed530bc27f279541b4' build failed
sqlite3/3.46.1: WARN: Build folder D:\cache\conan\p\b\sqlit063ff4ed98214\b\build
ERROR: sqlite3/3.46.1: Error in build() method, line 160
        cmake.build()
```

---

### DIFF
--- a/recipes/sqlite3/all/CMakeLists.txt
+++ b/recipes/sqlite3/all/CMakeLists.txt
@@ -165,12 +165,6 @@ if(ENABLE_FTS5 OR ENABLE_MATH_FUNCTIONS)
     endif()
 endif()
 
-# The upstream source files're encoded in UTF-8,
-# but some non-ASCII characters may be not be decoded
-# in the user's current code page (e.g. CP936).
-# Thus we need to instruct the MSVC to use UTF-8.
-target_compile_options(${PROJECT_NAME} PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")
-
 include(GNUInstallDirs)
 install(TARGETS ${PROJECT_NAME}
         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
@@ -192,5 +186,10 @@ if(SQLITE3_BUILD_EXECUTABLE)
     if(NOT HAVE_SYSTEM)
         target_compile_definitions(sqlite3-bin PRIVATE SQLITE_NOHAVE_SYSTEM)
     endif()
+    # The upstream source files're encoded in UTF-8,
+    # but some non-ASCII characters may be not be decoded
+    # in the user's current code page (e.g. CP936).
+    # Thus we need to instruct the MSVC to use UTF-8.
+    target_compile_options(sqlite3-bin PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")
     install(TARGETS sqlite3-bin DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif()
--- a/recipes/sqlite3/all/CMakeLists.txt
+++ b/recipes/sqlite3/all/CMakeLists.txt
@@ -169,9 +169,7 @@ endif()
 # but some non-ASCII characters may be not be decoded
 # in the user's current code page (e.g. CP936).
 # Thus we need to instruct the MSVC to use UTF-8.
-if(MSVC)
-    add_compile_options("/utf-8")
-endif()
+target_compile_options(${PROJECT_NAME} PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")
 
 include(GNUInstallDirs)
 install(TARGETS ${PROJECT_NAME}
--- a/recipes/sqlite3/all/CMakeLists.txt
+++ b/recipes/sqlite3/all/CMakeLists.txt
@@ -190,6 +190,8 @@ if(SQLITE3_BUILD_EXECUTABLE)
     # but some non-ASCII characters may be not be decoded
     # in the user's current code page (e.g. CP936).
     # Thus we need to instruct the MSVC to use UTF-8.
-    target_compile_options(sqlite3-bin PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+    if(SQLITE3_VERSION VERSION_LESS "3.47.0")
+        target_compile_options(sqlite3-bin PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")
+    endif()
     install(TARGETS sqlite3-bin DESTINATION ${CMAKE_INSTALL_BINDIR})
 endif()
--- a/recipes/sqlite3/all/CMakeLists.txt
+++ b/recipes/sqlite3/all/CMakeLists.txt
@@ -165,6 +165,14 @@ if(ENABLE_FTS5 OR ENABLE_MATH_FUNCTIONS)
     endif()
 endif()
 
+# The upstream source files're encoded in UTF-8,
+# but some non-ASCII characters may be not be decoded
+# in the user's current code page (e.g. CP936).
+# Thus we need to instruct the MSVC to use UTF-8.
+if(MSVC)
+    add_compile_options("/utf-8")
+endif()
+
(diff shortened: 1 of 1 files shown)


X1aomu:
Hi, @AbrilRBS 

ping friendly :)

It seems some new checks was added recently, but their status remain at _Waiting for status to be reported_. Is there any problem?

recipes/sqlite3/all/CMakeLists.txt
```diff
@@ -165,6 +165,14 @@ if(ENABLE_FTS5 OR ENABLE_MATH_FUNCTIONS)
     endif()
 endif()
 
+# The upstream source files're encoded in UTF-8,
+# but some non-ASCII characters may be not be decoded
+# in the user's current code page (e.g. CP936).
+# Thus we need to instruct the MSVC to use UTF-8.
+if(MSVC)
+    add_compile_options("/utf-8")
+endif()
```
uilianries, line 174:
```suggestion
target_compile_options(${PROJECT_NAME} PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")
```

Prefer using targets always to not spread the option globally.

X1aomu: Updated, thanks

X1aomu: Oops, just found that `target_compile_options(${PROJECT_NAME} PRIVATE "$<$<C_COMPILER_ID:MSVC>:/utf-8>")` will not work for the initial problem.

The problematic file `shell.c` belongs to target `sqlite3-bin` not `${PROJECT_NAME}`, so I need to apply the flag to the correct target.

That problem is a bit hard to find. I initially tested your suggestion on the newest version `3.47.0` and everything was fine.
However, the version `3.47.0` does not have the charset issue, so that it's ok without the `/utf-8` flag. The older version e.g. `3.46.1` still suffer, which is the latest one until I make this PR...

So here is a new problem: should we restrict this tweak to specific versions?

uilianries: Good catch, thank you for informing! 

Yes, only the affected version and be applied to sqlite3-bin via target.

X1aomu: Fixed

jcar87:
Thanks @X1aomu for opening this PR and reporting this issue.

If this issue is already fixed in 3.47 and newer - I would avise to simply use those instad of backporting the fix to other versions. We generally don't backport fixes if there's already released versions where the issue is already fixed.

If there's something in your dependency graphs bringing older versions of sqlite3 affected by this bug, I would advise:

* If you are able, use a version override at the top level conanfile.py: https://docs.conan.io/2.11/reference/conanfile/methods/requirements.html#override
* Otherwise, consider using a `[replace_requires]` in a profile ([Docs](https://docs.conan.io/2.11/reference/config_files/profiles.html#replace-requires) here) - for example:

```
sqlite3/*: sqlite/3.47.2
```

Also note that passing `/utf-8` to the compiler has other side effects in the generated code, since it implies `/execution-charset:utf-8` - thus may not be safe patch to apply as it changes behaviour. I suspect one would need: `/source-charset:utf-8` instead - given the description of the problem.

You can pass this without modifying the recipe, e.g. in your profile that uses msvc:
```
[settings]
compiler=msvc

[conf]
sqlite3/*:tools.build:cflags=["/source-charset:utf-8"]
```

If there are any dependencies/recipes in Conan Center still resolving to an earlier version let us know so that we can prioritise them. Thank you!

X1aomu:
That makes sense. Thanks for mention.

PR #25648, opened 2024-10-18
bx: add msvc version 194

danniesim:
### Summary
Changes to recipe:  **bgfx/cci.20230216**

#### Motivation
Only one library can be changed in the same PR
See: #25606

#### Details
See: #25606


AbrilRBS (approved):
Thanks for taking the time to split this into multiple PRs. This is an unfortunate limitation of the current CI system, which we're planning to replace super soon, sorry we were not fast enought to avoid the pain this time around :)

jcar87 (approved):

PR #25654, opened 2024-10-18
openssl: remove soname version on android

hypengw:
### Summary
Changes to recipe:  **openssl/3.x.x**

#### Motivation

https://github.com/openssl/openssl/issues/3902

#### Details

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -465,6 +465,8 @@ class OpenSSLConan(ConanFile):
                 shared_target = 'shared_target  => "gnu-shared",'
             if self.options.get_safe("fPIC", True):
                 shared_cflag = 'shared_cflag => "-fPIC",'
+        if self.settings.os == 'Android':
+            shared_extension = r'shared_extension => ".so",'
 
         if self.settings.os in ["iOS", "tvOS", "watchOS"] and self.conf.get("tools.apple:enable_bitcode", check_type=bool):
             cflags.append("-fembed-bitcode")


jcar87:
Hi @hypengw  - thanks for opening this PR.

Unfortunately this is not something we could merge without causing issues: changing the soname for the same version between recipe revisions would "break" any pre-existing shared library packages of anything that depends on OpenSSL (for Android).


* We've been unable to pinpoint up-to-date documentation on the support of soname'd .so files on Android -  most information I can find about this issue dates 2011-2016
* The issue referenced: https://github.com/openssl/openssl/issues/3902 was closed, but never lead to any changes in OpenSSL - and they preserved this behaviour for version 3 as well
* Looking at this comment in CMake's Android platform configuration: https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/Platform/Android.cmake?ref_type=heads#L20-24 - it would appear that newer versions of Android _do_ support correct soname linking/loading. 
* I can see in the Android 23 changelog https://android.googlesource.com/platform/bionic/+/refs/tags/ndk-r23-beta4/android-changes-for-ndk-developers.md that support for this was added/improved

If there's more information you could provide us to reproduce this issue, including which version of the Android API level are being targeted and how we can reproduce this, we'll happily look into it so that we can figure out a less potentially disruptive solution.

maikelvdh:
@jcar87 I would like to point out that the respective change is actually for the generated configuration matching with what is done upstream in: https://github.com/openssl/openssl/commit/4f10a996e5123c20315912149f586c481960e0de. Given that this configuration file isn't used, but generated in the Conan recipe it does make sense to apply the change suggested in this PR.

jcar87:
Thanks @maikelvdh - that is useful information.

The following still applies:

> changing the soname for the same version between recipe revisions would "break" any pre-existing shared library packages of anything that depends on OpenSSL 

Merging the PR would cause breakages since anything that depends on shared OpenSSL for Android, in a way where Conan would not re-trigger a rebuild.

The case still stands that in more recent versions of Android the current setup should work. We'll look into what we can do it if a reproducible bug is reported, thanks!

maikelvdh:
> The case still stands that in more recent versions of Android the current setup should work. We'll look into what we can do it if a reproducible bug is reported, thanks!

The problem still exists in Android context and the problem is rather related to the packaging of the native libraries within the context of Android Gradle Plugin when producing a library or application which is not supporting versioned shared libraries as such.

jcar87:
> > The case still stands that in more recent versions of Android the current setup should work. We'll look into what we can do it if a reproducible bug is reported, thanks!
> 
> The problem still exists in Android context and the problem is rather related to the packaging of the native libraries within the context of Android Gradle Plugin when producing a library or application which is not supporting versioned shared libraries as such.

if you are able to create a minimal reproducible example for us to look at, please open a new issue and we'll get on it, thanks! :)

PR #25672, opened 2024-10-19
Draft: OpenImageIO: Add 3.0.0.0 release

irieger:
### Summary
Changes to recipe:  **openimageio/3.0.0.0**

**NOTE:** Right now this is a draft using the beta-tag to prepare the MR to be updated once the release is ready.

#### Motivation
OpenImageIO 3.0.0.0 brings a larger update of the infrastructure, new file formats (especially Jpeg XL which now becomes popular) etc.

#### Details
Larger changes in the build system introducing the need to adapt a few things, especially:
- libjxl/JpegXS support only being available in 3.0.0.0 and newer
- boost no longer needed beginning with 3.0.0.0
- C++17 now the required standard for Version >= 3

---
- [x ] Tested locally with at least one configuration using a recent version of Conan

### DIFF
--- a/recipes/openimageio/all/conanfile.py
+++ b/recipes/openimageio/all/conanfile.py
@@ -27,42 +27,44 @@ class OpenImageIOConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
-        "with_libjpeg": ["libjpeg", "libjpeg-turbo"],
-        "with_libpng": [True, False],
-        "with_freetype": [True, False],
-        "with_hdf5": [True, False],
-        "with_opencolorio": [True, False],
-        "with_opencv": [True, False],
-        "with_tbb": [True, False],
         "with_dicom": [True, False],
         "with_ffmpeg": [True, False],
+        "with_freetype": [True, False],
         "with_giflib": [True, False],
+        "with_hdf5": [True, False],
+        "with_jxl": [True, False],
         "with_libheif": [True, False],
-        "with_raw": [True, False],
+        "with_libjpeg": ["libjpeg", "libjpeg-turbo"],
+        "with_libpng": [True, False],
+        "with_libwebp": [True, False],
+        "with_opencolorio": [True, False],
+        "with_opencv": [True, False],
         "with_openjpeg": [True, False],
         "with_openvdb": [True, False],
         "with_ptex": [True, False],
-        "with_libwebp": [True, False],
+        "with_raw": [True, False],
+        "with_tbb": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
-        "with_libjpeg": "libjpeg",
-        "with_libpng": True,
-        "with_freetype": True,
-        "with_hdf5": True,
-        "with_opencolorio": True,
-        "with_opencv": False,
-        "with_tbb": False,
         "with_dicom": False,  # Heavy dependency, disabled by default
         "with_ffmpeg": True,
+        "with_freetype": True,
         "with_giflib": True,
+        "with_hdf5": True,
+        "with_jxl": True,
+        "with_libjpeg": "libjpeg",
         "with_libheif": True,
-        "with_raw": False,  # libraw is available under CDDL-1.0 or LGPL-2.1, for this reason it is disabled by default
+        "with_libpng": True,
+        "with_libwebp": True,
+        "with_opencolorio": True,
         "with_openjpeg": True,
         "with_openvdb": False,  # FIXME: broken on M1
+        "with_opencv": False,
         "with_ptex": True,
-        "with_libwebp": True,
+        "with_raw": False,  # libraw is available under CDDL-1.0 or LGPL-2.1, for this reason it is disabled by default
+        "with_tbb": False,
     }
 
     def export_sources(self):
@@ -75,6 +77,8 @@ class OpenImageIOConan(ConanFile):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+        if Version(self.version) < "3.0.0.0":
+            self.options.with_jxl = False  #rm_safe("with_jxl")
 
     def requirements(self):
         # Required libraries
@@ -97,6 +101,8 @@ class OpenImageIOConan(ConanFile):
             self.requires("fmt/9.1.0", transitive_headers=True)
 
         # Optional libraries
+        if self.options.with_jxl:
+            self.requires("libjxl/0.10.3")
         if self.options.with_libpng:
             self.requires("libpng/1.6.42")
         if self.options.with_freetype:
@@ -128,7 +134,6 @@ class OpenImageIOConan(ConanFile):
             self.requires("ptex/2.4.2")
         if self.options.with_libwebp:
             self.requires("libwebp/1.3.2")
-        # Todo add JXL for OpenImageIO 3.0.0.0
 
         # TODO: R3DSDK dependency
         # TODO: Nuke dependency
@@ -178,6 +183,8 @@ class OpenImageIOConan(ConanFile):
         tc.variables["USE_JPEG"] = True
         # OIIO CMake files are patched to check USE_* flags to require or not use dependencies
         tc.variables["USE_JPEGTURBO"] = (self.options.with_libjpeg == "libjpeg-turbo")
+        if Version(self.version) >= '3.0.0.0':
+            tc.variables["USE_JXS"] = self.options.with_jxl
         tc.variables["USE_LIBHEIF"] = self.options.with_libheif
         tc.variables["USE_LIBPNG"] = self.options.with_libpng
         tc.variables["USE_LIBRAW"] = self.options.with_raw
@@ -193,7 +200,6 @@ class OpenImageIOConan(ConanFile):
         tc.variables["USE_TBB"] = self.options.with_tbb
 
         # Unsupported options
-        tc.variables["USE_JXL"] = False  # Todo, is available as a package
         tc.variables["USE_NUKE"] = False
         tc.variables["USE_R3DSDK"] = False
 
@@ -265,7 +271,6 @@ class OpenImageIOConan(ConanFile):
             "openexr::openexr",
         ]
 
-
         if Version(self.version) < "3.0.0.0":
             open_image_io_util.requires += [
                 "boost::filesystem",
@@ -280,6 +285,8 @@ class OpenImageIOConan(ConanFile):
                 "boost::regex",
             ]
 
+        if self.options.with_jxl:
+            open_image_io.requires.append("libjxl::libjxl")
         if self.options.with_libjpeg == "libjpeg":
             open_image_io.requires.append("libjpeg::libjpeg")
         elif self.options.with_libjpeg == "libjpeg-turbo":
--- a/recipes/openimageio/all/conanfile.py
+++ b/recipes/openimageio/all/conanfile.py
@@ -32,9 +32,9 @@ class OpenImageIOConan(ConanFile):
         "with_freetype": [True, False],
         "with_giflib": [True, False],
         "with_hdf5": [True, False],
-        "with_jxl": [True, False],
         "with_libheif": [True, False],
         "with_libjpeg": ["libjpeg", "libjpeg-turbo"],
+        "with_libjxl": [True, False],
         "with_libpng": [True, False],
         "with_libwebp": [True, False],
         "with_opencolorio": [True, False],
@@ -53,9 +53,9 @@ class OpenImageIOConan(ConanFile):
         "with_freetype": True,
         "with_giflib": True,
         "with_hdf5": True,
-        "with_jxl": True,
-        "with_libjpeg": "libjpeg",
         "with_libheif": True,
+        "with_libjpeg": "libjpeg",
+        "with_libjxl": False,  # TODO: Currently produces link failues
         "with_libpng": True,
         "with_libwebp": True,
         "with_opencolorio": True,
@@ -78,7 +78,7 @@ class OpenImageIOConan(ConanFile):
         if self.options.shared:
             self.options.rm_safe("fPIC")
         if Version(self.version) < "3.0.0.0":
-            self.options.with_jxl = False  #rm_safe("with_jxl")
+            self.options.with_libjxl = False  #rm_safe("with_libjxl")
 
     def requirements(self):
         # Required libraries
@@ -101,7 +101,7 @@ class OpenImageIOConan(ConanFile):
             self.requires("fmt/9.1.0", transitive_headers=True)
 
         # Optional libraries
-        if self.options.with_jxl:
+        if self.options.with_libjxl:
             self.requires("libjxl/0.10.3")
         if self.options.with_libpng:
             self.requires("libpng/1.6.42")
@@ -134,6 +134,10 @@ class OpenImageIOConan(ConanFile):
             self.requires("ptex/2.4.2")
         if self.options.with_libwebp:
             self.requires("libwebp/1.3.2")
+
+        # TODO: Temporary, to be removed
+        # Works with build=missing now, awaiting https://github.com/conan-io/conan-center-index/pull/25660
+        self.requires('lcms/2.16', override=True)
 
         # TODO: R3DSDK dependency
         # TODO: Nuke dependency
@@ -183,9 +187,9 @@ class OpenImageIOConan(ConanFile):
         tc.variables["USE_JPEG"] = True
         # OIIO CMake files are patched to check USE_* flags to require or not use dependencies
         tc.variables["USE_JPEGTURBO"] = (self.options.with_libjpeg == "libjpeg-turbo")
-        if Version(self.version) >= '3.0.0.0':
-            tc.variables["USE_JXS"] = self.options.with_jxl
         tc.variables["USE_LIBHEIF"] = self.options.with_libheif
+        if Version(self.version) >= '3.0.0.0':
+            tc.variables["USE_LIBJXL"] = self.options.with_libjxl
         tc.variables["USE_LIBPNG"] = self.options.with_libpng
         tc.variables["USE_LIBRAW"] = self.options.with_raw
         tc.variables["USE_LIBWEBP"] = self.options.with_libwebp
@@ -285,7 +289,7 @@ class OpenImageIOConan(ConanFile):
                 "boost::regex",
             ]
 
-        if self.options.with_jxl:
+        if self.options.with_libjxl:
             open_image_io.requires.append("libjxl::libjxl")
         if self.options.with_libjpeg == "libjpeg":
             open_image_io.requires.append("libjpeg::libjpeg")
--- a/recipes/openimageio/all/conanfile.py
+++ b/recipes/openimageio/all/conanfile.py
@@ -137,7 +137,7 @@ class OpenImageIOConan(ConanFile):
 
         # TODO: Temporary, to be removed
         # Works with build=missing now, awaiting https://github.com/conan-io/conan-center-index/pull/25660
-        self.requires('lcms/2.16', override=True)
+        #self.requires('lcms/2.16', override=True)
 
         # TODO: R3DSDK dependency
         # TODO: Nuke dependency
--- a/recipes/openimageio/all/conanfile.py
+++ b/recipes/openimageio/all/conanfile.py
@@ -79,7 +79,8 @@ class OpenImageIOConan(ConanFile):
     def requirements(self):
         # Required libraries
         self.requires("zlib/[>=1.2.11 <2]")
-        self.requires("boost/1.84.0")
+        if Version(self.version) < "3.0.0.0":
+            self.requires("boost/1.84.0")
         self.requires("libtiff/4.6.0")
         self.requires("imath/3.1.9", transitive_headers=True)
         self.requires("openexr/3.2.3")
@@ -127,12 +128,17 @@ class OpenImageIOConan(ConanFile):
             self.requires("ptex/2.4.2")
         if self.options.with_libwebp:
             self.requires("libwebp/1.3.2")
+        # Todo add JXL for OpenImageIO 3.0.0.0
+
         # TODO: R3DSDK dependency
         # TODO: Nuke dependency
 
     def validate(self):
         if self.settings.compiler.cppstd:
-            check_min_cppstd(self, 14)
+            if Version(self.version) >= "3.0.0.0":
+                check_min_cppstd(self, 17)
+            else:
+                check_min_cppstd(self, 14)
         if is_msvc(self) and is_msvc_static_runtime(self) and self.options.shared:
             raise ConanInvalidConfiguration(
                 "Building shared library with static runtime is not supported!"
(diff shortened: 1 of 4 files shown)


AbrilRBS:
@irieger are these the only changes you expect to make? I think that maybe we want to split the new major into its own recipe if the changes are going to end up making the old recipe more confusing, thanks!

AbrilRBS:
Also, I've converted this to a draft to avoid merging it by mistake, but otherwise the CI should work just the same :)

irieger:
> @irieger are these the only changes you expect to make? I think that maybe we want to split the new major into its own recipe if the changes are going to end up making the old recipe more confusing, thanks!

I'm open to split it up if you think that it would be best to split? I don't currently see much more changes than we currently have, maybe the need for a few more being done in the patching to get the linking fixed, there seems to be a private/public linking issue of some third party libs not translating to the final linking step which I'll look at in the coming days.

irieger:
> Also, I've converted this to a draft to avoid merging it by mistake, but otherwise the CI should work just the same :)

How does this work? As I'm normally used to gitlab at work as well as for my personal stuff, I'm not yet firm with all the Github tools. For gitlab just prefixing the title with Draft is enough to prevent the merge already. According to goolge there should be a setting below the Reviewers selection (I assume on the right side of the description), but I don't see a link to change the state there.

AbrilRBS:
> , but I don't see a link to change the state there.

Oh, that's because I already set is as a draft. Marking it as such simply tells ours bot not to automatically merge it, and disables the merge button in the UI

recipes/openimageio/all/conanfile.py
```diff
@@ -96,6 +101,8 @@ def requirements(self):
             self.requires("fmt/9.1.0", transitive_headers=True)
 
         # Optional libraries
+        if self.options.with_libjxl:
+            self.requires("libjxl/0.10.3")
         if self.options.with_libpng:
             self.requires("libpng/1.6.42")
```
AbrilRBS, line 107: We can set this to a version range and close https://github.com/conan-io/conan-center-index/pull/23122 in the process

irieger: Just added the version range as suggested.

irieger:
@AbrilRBS Build locally has just worked after the lcms 2.16 was merged. Now in the CCI we have "Error running command `conan graph info --requires..." doesn't really give more information so might that be related to the lcms thingy?

irieger:
Btw. in the OpenImageIO Slack channel, the maintainer announced that on monday a release candidate should be tagged and next friday there should be the release. I plan to get everything ready based on the RC next week and then should just need to swap the source URL & checksum hopefully.

recipes/openimageio/all/conanfile.py
```diff
@@ -75,11 +77,14 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+        if Version(self.version) < "3.0.0.0":
+            self.options.with_libjxl = False  #rm_safe("with_libjxl")
```
AbrilRBS, line 81: Option values can not be changed after the `config_options` method, `configure()` is too late

- If this is a default chaning with the versions, move this check to `config_options`
- If the recipe no longer has this option, remove it as in the comment
- If it supports the option, but for some reason True is failing to compile right now, adding a check in the validate() methiod would be preferred :)

irieger: Hey @AbrilRBS, thanks for following up. I had recently tried to rework the whole PR a bit but haven't got it to fully link. Just merged the state frem here and into my local state. I now followed your earlier suggestion and split it by creating a 3.x folder for the new stuff. Wasn't sure if I should rename the existing stuff probably?
(openexr has a 2.x and 3.x folder)

(Still pushed to have it up here, but it will likely fail to link. Up for suggestions as I've only had minimal time but also didn't got why that could now fail.

recipes/openimageio/all/conanfile.py
irieger: Ups, somehow I seem to have accidentially lost the recent change here. Will fix that quickly

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_dicom": [True, False],
+        "with_ffmpeg": [True, False],
+        "with_freetype": [True, False],
+        "with_giflib": [True, False],
+        "with_hdf5": [True, False],
+        "with_libheif": [True, False],
+        "with_libjpeg": ["libjpeg", "libjpeg-turbo"],
```
AbrilRBS, line 36: no mozjpeg?

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+        "shared": False,
+        "fPIC": True,
+        "with_dicom": False,  # Heavy dependency, disabled by default
+        "with_ffmpeg": True,
+        "with_freetype": True,
+        "with_giflib": True,
+        "with_hdf5": True,
+        "with_libheif": True,
+        "with_libjpeg": "libjpeg",
+        "with_libjxl": True,  # TODO: Currently produces link failues
```
AbrilRBS, line 57: Is this when set to False? We might want to add it to the validate function

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @staticmethod
+    def _conan_comp(name):
+        return f"openimageio_{name.lower()}"
+
+    def _add_component(self, name):
+        component = self.cpp_info.components[self._conan_comp(name)]
+        component.set_property("cmake_target_name", f"OpenImageIO::{name}")
+        component.names["cmake_find_package"] = name
+        component.names["cmake_find_package_multi"] = name
```
AbrilRBS, line 225: Can be removed for Conan 2

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+        return component
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "OpenImageIO")
+        self.cpp_info.set_property("pkg_config_name", "OpenImageIO")
+
+        self.cpp_info.names["cmake_find_package"] = "OpenImageIO"
+        self.cpp_info.names["cmake_find_package_multi"] = "OpenImageIO"
```
AbrilRBS, line 233: Same here

recipes/openimageio/3.x/patches/3.0.0.3-cmake-targets.patch
```diff
@@ -0,0 +1,560 @@
...
+index f8670a83f..27c279166 100644
+--- src/cmake/externalpackages.cmake
++++ src/cmake/externalpackages.cmake
+@@ -66,23 +66,22 @@ checked_find_package (OpenEXR REQUIRED
+ # library.
+ include_directories(BEFORE ${IMATH_INCLUDES} ${OPENEXR_INCLUDES})
+ set (OPENIMAGEIO_IMATH_TARGETS Imath::Imath)
+-set (OPENIMAGEIO_OPENEXR_TARGETS OpenEXR::OpenEXR)
+-set (OPENIMAGEIO_IMATH_DEPENDENCY_VISIBILITY "PRIVATE" CACHE STRING
++set (OPENIMAGEIO_OPENEXR_TARGETS openexr::openexr)
```
AbrilRBS, line 46: This can be set in the `CMakeDeps` generator

`cd.set_property("openexr", "cmake_target_name", "OpenEXR::OpenEXR")`
so that we can save some patch diffs

Sane with `OPENIMAGEIO_IMATH_DEPENDENCY_VISIBILITY`, I think it'd be better to set it in the `CMakeToolchain` generator

recipes/openimageio/3.x/patches/3.0.0.3-cmake-targets.patch
```diff
@@ -0,0 +1,560 @@
...
+diff --git src/iconvert/CMakeLists.txt src/iconvert/CMakeLists.txt
+index 735b88def..8f5bb3b54 100644
+--- src/iconvert/CMakeLists.txt
++++ src/iconvert/CMakeLists.txt
+@@ -2,4 +2,4 @@
+ # SPDX-License-Identifier: Apache-2.0
+ # https://github.com/AcademySoftwareFoundation/OpenImageIO
+ 
+-fancy_add_executable (LINK_LIBRARIES OpenImageIO)
++fancy_add_executable (LINK_LIBRARIES OpenImageIO fmt::fmt)
```
AbrilRBS, line 363: WHy is fmt not handled aready? Is it because upstream has it vendorized?

recipes/openimageio/3.x/conanfile.py
AbrilRBS: I see that upstream acually handles only turbo, nevermind!

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+        "with_freetype": True,
+        "with_giflib": True,
+        "with_hdf5": True,
+        "with_libheif": True,
+        "with_libjpeg": "libjpeg",
+        "with_libjxl": True,  # TODO: Currently produces link failues
+        "with_libpng": True,
+        "with_libwebp": True,
+        "with_openjpeg": True,
+        "with_openvdb": False,  # FIXME: broken on M1
```
AbrilRBS, line 61: Is OpenImage's build broken, or is the breakage in openvdb?

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+    def requirements(self):
+        # Required libraries
+        self.requires("zlib/[>=1.2.11 <2]")
+        self.requires("libtiff/4.7.0")
+        self.requires("imath/3.1.9", transitive_headers=True)
+        self.requires("openexr/3.2.3")
+        if self.options.with_libjpeg == "libjpeg":
+            self.requires("libjpeg/9e")
+        elif self.options.with_libjpeg == "libjpeg-turbo":
+            self.requires("libjpeg-turbo/3.0.4")
```
AbrilRBS, line 88:
```suggestion
            self.requires("libjpeg-turbo/[>=3.0.3 <4]")
```
(Or .4 if it's actually necessary!)

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+            self.requires("ptex/2.4.2")
+        if self.options.with_libwebp:
+            self.requires("libwebp/1.3.2")
+
+        # TODO: R3DSDK dependency
+        # TODO: Nuke dependency
+
+    def build_requirements(self):
+        # A minimum cmake version is now required that is reasonably new
+        self.build_requires("cmake/[>=3.18.2]")
```
AbrilRBS, line 133:
```suggestion
        self.build_requires("cmake/[>=3.18.2 <4]")
```

_Juust in case_

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+
+    def validate(self):
+        if self.settings.compiler.cppstd:
+            check_min_cppstd(self, 17)
```
AbrilRBS, line 137:
```suggestion
        check_min_cppstd(self, 17)
```
It's meant to always be set in Conan 2

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,303 @@
...
+        # Unsupported options
+        tc.variables["USE_NUKE"] = False
+        tc.variables["USE_R3DSDK"] = False
+
+        tc.generate()
+        cd = CMakeDeps(self)
+        cd.generate()
+
+    def build(self):
+        apply_conandata_patches(self)
```
AbrilRBS, line 201: This should now be done in the `source()` method after the `get()` call

AbrilRBS (commented):
Thanks! Some comments on the PR, but it looks great :)

recipes/openimageio/3.x/conanfile.py
irieger: Comment was a leftover I missed from an intermediate state. Thanks. Removing.

recipes/openimageio/3.x/patches/3.0.0.3-cmake-targets.patch
irieger: Upstream does offer to vendorize, but shouldn't vendorize with the settings if I didn't miss something. Somehow it did lead to some include errors before I patched around. Maybe I should have another look and try to see why it is working outside of conan.

recipes/openimageio/3.x/conanfile.py
irieger: Good question. Didn't test, just copied from the 2.x (all) recipe. Will retry if it still fails.

recipes/openimageio/3.x/conanfile.py
irieger: Ok, changed.

recipes/openimageio/3.x/conanfile.py
irieger: It looks like it works in the meantime, so changed it to true. Or maybe we better leave it at false to keep old behaviour but remove the comment only, what do you think?

recipes/openimageio/3.x/patches/3.0.0.3-cmake-targets.patch
irieger: Tried to change, somehow now it fails to link again. Not sure why. So I pushed the changes to see if it is a local problem or if the CI also runs into it.

recipes/openimageio/3.x/conanfile.py
AbrilRBS: I think new major versions are ok with default changes of options, specially if that means following upstream decision to turn it on if available.

irieger:
Iirc there was some chatter that with the new CI infrastructure there could be alternative configurations of packages being built that aren't using defaults? For macOS I can only build with `--build=missing -o:a 'minizip-ng/*:with_zlib=True' -o:a 'minizip-ng/*:with_libcomp=False'` due to:
```
opencolorio/2.4.0: Invalid: opencolorio/2.4.0 requires minizip-ng with with_zlib = True. On Apple platforms with_libcomp = False is also needed to enable the with_zlib option.
```

Due to this, currently neither opencolorio nor openimageio will be build for macOS in the CI. Would be nice if it could be done?

recipes/openimageio/3.x/patches/3.0.0.3-cmake-targets.patch
irieger: Fails locally, and the CI doesn't build for macOS, so maybe there is a problem. There is a load of the messages in the following style, happening for all the integrated tools when they are trying to link. Just a small excerpt:
```
[  0%] Built target CopyFiles
[ 13%] Built target OpenImageIO_Util
[ 88%] Built target OpenImageIO
[ 89%] Linking CXX executable ../../bin/iconvert
ld: warning: ignoring duplicate libraries: '-lc++', '/Users/ingmar/.conan2/p/b/opene614278b8c5e70/p/lib/libOpenEXRUtil-3_2.a'
  "Iex_3_3::IoExc::IoExc(char const*)", referenced from:
      OpenImageIO_v3_0::OpenEXRInput::valid_file(OpenImageIO_v3_0::Filesystem::IOProxy*) const in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO_v3_0::ImageSpec&, OpenImageIO_v3_0::ImageSpec const&) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInput::seek_subimage(int, int) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInputStream::seekg(unsigned long long) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInputStream::read(char*, int) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO_v3_0::ImageSpec const&, OpenImageIO_v3_0::ImageOutput::OpenMode) in libOpenImageIO.a[85](exroutput.cpp.o)
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, int, OpenImageIO_v3_0::ImageSpec const*) in libOpenImageIO.a[85](exroutput.cpp.o)
      ...
  "Iex_3_3::IoExc::~IoExc()", referenced from:
      OpenImageIO_v3_0::OpenEXRInput::valid_file(OpenImageIO_v3_0::Filesystem::IOProxy*) const in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO_v3_0::ImageSpec&, OpenImageIO_v3_0::ImageSpec const&) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInput::seek_subimage(int, int) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInputStream::seekg(unsigned long long) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXRInputStream::read(char*, int) in libOpenImageIO.a[84](exrinput.cpp.o)
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO_v3_0::ImageSpec const&, OpenImageIO_v3_0::ImageOutput::OpenMode) in libOpenImageIO.a[85](exroutput.cpp.o)
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, int, OpenImageIO_v3_0::ImageSpec const*) in libOpenImageIO.a[85](exroutput.cpp.o)
      ...
  "Imf_3_3::OutputFile::copyPixels(Imf_3_3::InputPart&)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::copy_image(OpenImageIO_v3_0::ImageInput*) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputFile::writePixels(int)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::write_scanlines(int, int, int, OpenImageIO_v3_0::TypeDesc, void const*, long long, long long) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputFile::setFrameBuffer(Imf_3_3::FrameBuffer const&)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::write_scanlines(int, int, int, OpenImageIO_v3_0::TypeDesc, void const*, long long, long long) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputFile::OutputFile(Imf_3_3::OStream&, Imf_3_3::Header const&, int)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO_v3_0::ImageSpec const&, OpenImageIO_v3_0::ImageOutput::OpenMode) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputPart::copyPixels(Imf_3_3::InputPart&)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::copy_image(OpenImageIO_v3_0::ImageInput*) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputPart::writePixels(int)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::write_scanlines(int, int, int, OpenImageIO_v3_0::TypeDesc, void const*, long long, long long) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputPart::setFrameBuffer(Imf_3_3::FrameBuffer const&)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::write_scanlines(int, int, int, OpenImageIO_v3_0::TypeDesc, void const*, long long, long long) in libOpenImageIO.a[85](exroutput.cpp.o)
  "Imf_3_3::OutputPart::OutputPart(Imf_3_3::MultiPartOutputFile&, int)", referenced from:
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, OpenImageIO_v3_0::ImageSpec const&, OpenImageIO_v3_0::ImageOutput::OpenMode) in libOpenImageIO.a[85](exroutput.cpp.o)
      OpenImageIO_v3_0::OpenEXROutput::open(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&, int, OpenImageIO_v3_0::ImageSpec const*) in libOpenImageIO.a[85](exroutput.cpp.o)

```

AbrilRBS: Will check tomorrow, thanks for the report!

recipes/openimageio/3.x/test_package/conanfile.py
```diff
@@ -0,0 +1,25 @@
...
+    def layout(self):
+        cmake_layout(self)
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def test(self):
+        if not can_run(self):
```
perseoGI, line 23:
```suggestion
        if can_run(self):
```

This bug is also present in `all` folder.

topheg:
Hello, I tried your recipe for OIIO 3, and have a few remarks / questions

Though libpng is added as a dependency, PNG support is not compiled, I always get : 
> "CMake Warning at src/png.imageio/CMakeLists.txt:11 (message):
>   libpng not found, so PNG support will not work

I also see that the build system of OIIO downloads and builds fmt itself, even though it is provided as a dependency.
> -- Building package fmt  locally
> --         Building local fmt 10.2.1 from https://github.com/fmtlib/fmt
> --         Downloading local https://github.com/fmtlib/fmt

And one last thing, I see libsquish is declared as a dependency, is it still used in OIIO 3 ? I cannot find any mention of it in the docs.

Anyway, thanks for the great work !

irieger:
> Hello, I tried your recipe for OIIO 3, and have a few remarks / questions
> 
> Though libpng is added as a dependency, PNG support is not compiled, I always get :
> 
> > "CMake Warning at src/png.imageio/CMakeLists.txt:11 (message):
> > libpng not found, so PNG support will not work
> 
> I also see that the build system of OIIO downloads and builds fmt itself, even though it is provided as a dependency.
> 
> > -- Building package fmt  locally
> > --         Building local fmt 10.2.1 from https://github.com/fmtlib/fmt
> > --         Downloading local https://github.com/fmtlib/fmt
> 
> And one last thing, I see libsquish is declared as a dependency, is it still used in OIIO 3 ? I cannot find any mention of it in the docs.
> 
> Anyway, thanks for the great work !

Thx for the input. You are right on both points. Missed the png in all the verbosity of the build. And libsquish was also overlooked. Just fixed both. Now looking into getting the pipeline work again.

recipes/openimageio/3.x/patches/3.0.0.3-cmake-targets.patch
irieger: You wanted to have a look back some time. Anything happened or any chance you could look at it.

Just did a bit more work on it as it currently still strangely builds fmt internally instead of using the dependency etc. but I'm on that. The linking however still fails. Seems like the problem I locally prevented via `mold` as the standard linker will not find dependencies of linked so's themselves if not specifically declared in the list of libs to link. See #23421

irieger:
Note for reviewer: The build issue for shared builds looks like what was discussed in https://github.com/conan-io/conan-center-index/issues/23421

irieger:
Also just created #26364 to solve the graph conflict introduced by a libtiff recipe update a few days ago.

irieger:
Note: I have this successfully build on Windows/Linux and macOS (arm, I don't run this pipeline for x64 macOS anymore).

irieger:
@AbrilRBS @perseoGI Hey, would be nice to get some input here how we could move forward. OIIO just released the next version but I'll spare myself from keeping this PR up to date as long as I'm waiting for feedback but would update once we have some concept to move forward.

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,313 @@
...
+        "with_opencv": [True, False],
+        "with_openjpeg": [True, False],
+        "with_openvdb": [True, False],
+        "with_ptex": [True, False],
+        "with_raw": [True, False],
+        "with_tbb": [True, False],
+
+        # To be replaced with some proper check or fix of the build?
+        # https://github.com/conan-io/conan-center-index/issues/23421
+        "cci_hack": [True, False],
```
irieger, line 49: This is just to test out if with this I can prevent the CCI build (ok, just realized I could have done the same with just hardcoding the Linux path to not support shared but I'll wait for generally feedback how to move forward before the next round of changes anyway).

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,313 @@
...
+        if is_msvc(self) and is_msvc_static_runtime(self) and self.options.shared:
+            raise ConanInvalidConfiguration(
+                "Building shared library with static runtime is not supported!"
+            )
+
+        # https://github.com/conan-io/conan-center-index/issues/23421
+        if self.options.cci_hack and self.settings.os == "Linux" and self.options.shared:
+            raise ConanInvalidConfiguration(
+                "Bug in Linux linking!"
+            )
```
irieger, line 152: I'd really appreciate if we could find a way forward to fix this issue. https://github.com/conan-io/conan-center-index/issues/23421

irieger (commented):
Awaiting some more feedback as this is blocked to be properly cleaned up.

crgnam:
I would like to check in with this because the current available release is 2.5.18.0 which is very out of date (and notably requires boost which can cause certain issues).

irieger:
> I would like to check in with this because the current available release is 2.5.18.0 which is very out of date (and notably requires boost which can cause certain issues).

Best chance is to just use it (more or less) as is. I use it in its current state I think with only my hack to test some CI issue removed since months. But basically I gave up hoping for a review I guess. Still will likely update it for myself so feel free to use it.

irieger:
This recipe again shows that it is really hard for a CCI recipe that provides something more complex with the hardcoded version dependencies in each package. You always tend to run into something where your dependencies share a dependency but use different version.

P.S. I happily use stuff in my local artifactory/build pipeline where I override several dependencies to use the newest in all repos as well as for some packages enable/disable options.

recipes/openimageio/3.x/conanfile.py
```diff
@@ -0,0 +1,324 @@
...
+        elif self.options.with_libjpeg == "libjpeg-turbo":
+            self.requires("libjpeg-turbo/[>=3.0.3 <4]")
+        self.requires("pugixml/1.15")
+        self.requires("tsl-robin-map/1.4.0")
+        self.requires("fmt/11.2.0", transitive_headers=True)
+        self.requires("opencolorio/2.4.2")
+
+        self.requires("lcms/2.17", override=True)
+
+        # Optional libraries
```
irieger, line 107: This will get a few bumps:
* hdf for example doesn't publish 1.14.5 anymore etc.
* General check what the current versions are for stuff not allowed to use ranges.

Also would be nice to get feedback where ranges are accepted. I think libjpeg is now also changes to ranges in many recipes.

irieger:
Currently updating for OpenImageIO 3.1.6.2 (3.1 is the newest version/fall 2025 release branch, with OIIO having yearly release cycle for main changes). I just added 2.5 related recipes with two I needed to change to get it to build locally as otherwise conan couldn't resolve version conflicts of the different dependencies.

Will push the changes today or tomorrow for 3.1.6.2 itself also removing some temporary code I added in the process where I found a solution now.

irieger:
Made it to build - with the temporary override for lcms and two other PRs build before (#28556 and #28554) - locally.

Getting the solver happy is really hard with this recipe as many of the dependencies have shared other dependencies so having everything match up feels hard.

irieger:
Small update:
I just successfully updated to OpenImageIO 3.1.6.2 based on this recipe (and the related PRs) on my local CCI mirror (building in my local CI from my own branch [just a few not yet merged recipes integrated on top of CCI master] with profiles I tuned for my use case). Successfully builds with Release, RelWithDebInfo and Debug for my target platforms, which are Linux (x64 and arm64) as well as macOS (arm64). (And updated in my consumer app and the few things I tested worked.)

jcar87:
Hi @irieger - thanks so much for your efforts and apologies for the delay in getting to this - a delay of almost 1 year in getting this should not have happened at all. 

Whilst reviewing this very PR - we eventually determined that the existing recipe for 2.x did not require too many changes to support 3.x: we have worked on this here: https://github.com/conan-io/conan-center-index/pull/28618 (should already be merged and published).  Please let us know if this works as expected, and if it doesn't, let us know and we will prioritise the efforts to rectify it.

irieger:
Hey @jcar87 

Thanks for integrating 3.1.6.2. Its nice if it can be largely reduced. There is a few small things that I encountered now, trying to integrated/build it in my local CCI clone and then locally debugging and working on a small extension and found a larger problem for which I'm starting a discussion:
* Add optional openjph dependency introduced with 3.0 (or 3.1):  #28626
* Issue to discuss the underlying problem of dependency management that I encountered: #28627

I'll close this PR however as it is definitely outdated now.

PR #25690, opened 2024-10-22
tree-sitter-sql: add recipe

toge:
### Summary
Changes to recipe:  **tree-sitter-sql/***

#### Motivation
I want to use tree-sitter more. (SQL, CQL, CMake, Cypher...)

#### Details
There are several tree-sitter sql parsers.
So I decided to prefix them for identification.

---


AbrilRBS:
Thanks @toge!
As we everything tree-sitter releated, I'm asking @perseoGI to review it, as he's a big fan of it :)

AbrilRBS (commented):
If/when this gets merged, we should also send a PR upstream updating their readme to be mention Conan alongside the rest of them! :)

recipes/derekstride-tree-sitter-sql/all/test_package/conanfile.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
+
+    def layout(self):
+        cmake_layout(self)
+
+    def build_requirements(self):
+        if not self.conf.get("tools.gnu:pkg_config", check_type=str):
+            self.tool_requires("pkgconf/2.2.0")
```
AbrilRBS, line 20: Remember that pkgconf can now be a version range :)

recipes/derekstride-tree-sitter-sql/all/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+cmake_minimum_required(VERSION 3.15)
+project(tree-sitter-sql LANGUAGES C)
+
+find_package(tree-sitter REQUIRED CONFIG)
+include(GenerateExportHeader)
+
+file(WRITE api.h [[
```
AbrilRBS, line 7: I want to understand only 1 thing about this. Is this file then not provided by upstream either? Or what is its pourpose here?

AbrilRBS (commented):
Minor review comments, looks good to me otherwise, but will let Perseo have the last word :)

recipes/derekstride-tree-sitter-sql/all/CMakeLists.txt
toge: @AbrilRBS 
I just follow the recipe of `tree-sitter-c`.
https://github.com/conan-io/conan-center-index/blob/master/recipes/tree-sitter-c/all/CMakeLists.txt

I think all recipes of tree-sitter in cci should have same style as much as possible.

toge: @AbrilRBS 
I just follow the recipe of `tree-sitter-c`.
https://github.com/conan-io/conan-center-index/blob/master/recipes/tree-sitter-c/all/CMakeLists.txt

I think all recipes of tree-sitter in cci should have same style as much as possible.

recipes/derekstride-tree-sitter-sql/all/test_package/conanfile.py
toge: Fixed.

recipes/derekstride-tree-sitter-sql/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import copy, get
+import os
+
+
+required_conan_version = ">=1.53.0"
+
+
+class DerekstrideTreeSitterSqlConan(ConanFile):
+    name = "derekstride-tree-sitter-sql"
```
perseoGI, line 11: Everything looks good, but I have a question here. Why are we adding the authors name inside the recipe name?
Wouldn't be more convenient to call it `tree-sitter-sql` as it is called in cargo and pip https://github.com/DerekStride/tree-sitter-sql#using-cargo?
Also to be more homogeneous with existing tree-sitter recipes.

There are other projects, TJ one https://github.com/tjdevries/tree-sitter-sql and https://github.com/m-novikov/tree-sitter-sql
First seems to be abandoned and the other haven't been updated in a while. 
So I will suggest to rename to `tree-sitter-sql` and in the future we may consider adding other providers.
What do you think? It may be a personal preference.

uilianries: I agree in renaming it to `tree-sitter-sql`. The project seems be consolidated over other package managers and most active between their homonyms.

toge: @perseoGI @uilianries 
Thank you for your advise!
I will rename this recipe.

I had used to use m-novikov's `tree-sitter-sql` in previous, which was the most advanced implementation at the time.
Since m-novikov's tree-sitter-sql is not that old, I wondered if I should give such a generic name as `tree-sitter-sql` only to derekstride's.

toge: fixed.

Please review it again.

uilianries: Thank you @toge san! :smile:

recipes/tree-sitter-sql/all/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
+cmake_minimum_required(VERSION 3.15)
+project(tree-sitter-sql LANGUAGES C)
+
+find_package(tree-sitter REQUIRED CONFIG)
+include(GenerateExportHeader)
+
+file(WRITE api.h [[
```
uilianries, line 7: I don't think it's needed to create this file, the project offers a very similar one:

`./bindings/c/tree-sitter-sql.h:const TSLanguage *tree_sitter_sql(void);`

toge: @uilianries 
Thank you for your review.
I know your pointed file.

The main difference is api.h includes generated `tree_sitter_sql_export.h` as following:
```c
#ifndef TREE_SITTER_C_EXPORT_H
#define TREE_SITTER_C_EXPORT_H

#ifdef TREE_SITTER_C_STATIC_DEFINE
#  define TREE_SITTER_C_EXPORT
#  define TREE_SITTER_C_NO_EXPORT
#else
#  ifndef TREE_SITTER_C_EXPORT
#    ifdef tree_sitter_sql_EXPORTS
        /* We are building this library */
#      define TREE_SITTER_C_EXPORT 
#    else
        /* We are using this library */
#      define TREE_SITTER_C_EXPORT 
#    endif
#  endif

#  ifndef TREE_SITTER_C_NO_EXPORT
#    define TREE_SITTER_C_NO_EXPORT 
#  endif
#endif

#ifndef TREE_SITTER_C_DEPRECATED
#  define TREE_SITTER_C_DEPRECATED __attribute__ ((__deprecated__))
#endif

#ifndef TREE_SITTER_C_DEPRECATED_EXPORT
#  define TREE_SITTER_C_DEPRECATED_EXPORT TREE_SITTER_C_EXPORT TREE_SITTER_C_DEPRECATED
#endif

#ifndef TREE_SITTER_C_DEPRECATED_NO_EXPORT
#  define TREE_SITTER_C_DEPRECATED_NO_EXPORT TREE_SITTER_C_NO_EXPORT TREE_SITTER_C_DEPRECATED
#endif

/* NOLINTNEXTLINE(readability-avoid-unconditional-preprocessor-if) */
#if 0 /* DEFINE_NO_DEPRECATED */
#  ifndef TREE_SITTER_C_NO_DEPRECATED
#    define TREE_SITTER_C_NO_DEPRECATED
#  endif
#endif

#endif /* TREE_SITTER_C_EXPORT_H */
```

It helps to build windows shared build.

As I wrote the thread of this PR, I just followed tree-sitter-c style.
I want CCI's community to define the standard style of tree-sitter's syntax recipe.

If it is defined, I'm going to create several recipes(C++, CQL, CMake,...) about tree-sitter following that standard style later.

uilianries: @toge Sorry, I didn't read completely that header file. Indeed you did a great job investigating that project :smile: 

I would not recommend creating that custom header, even if it's case of sacrificing shared library support on Windows, because it will result not only in an unique new header, but we will become the maintainers. The project itself is a bit limited in terms of building, as you noted, so don't worry trying to add extra support.

AbrilRBS (approved):

AbrilRBS (approved):

uilianries (approved):
LGTM. It looks aligned to the upstream in terms of build steps.

PR #25694, opened 2024-10-22
Qt6: honor "disabled_feature" for some QtTools (assistant, linguist)

SvenRoederer:
### Summary

#### Motivation
Improve the usefulness of conan-option "disable_feature"

#### Details
The current recipe provides the package-option "disable_feature", which is passed down to Qt build-script. This option provides some detailed control of subcomponents to be build.
In our project we only need QtLingusit from QtTools, but not the QtAssistant, QtDesigner and the other small tools. Esp. dropping the Assistant and Designer from our build safes some resources. But specifying them makes conan packaging fail, as it will not find the binaries. 

build is tested on Windows with `qt:disabled_features = assistant designer distancefieldgenerator pixeltool qtattributionsscanner qdbus qtdiag qtplugininfo`

---


jcar87:
Now solved via https://github.com/conan-io/conan-center-index/pull/29176

Thanks for providing a way to reproduce the error. Please note that we had to make additional changes as the fixes proposed in this PR were not fully solving them, but it was a good start. We have left out any changes related to the schannel plugin - I dont think they were needed or attached to the correct target, however if there are any problems please open an new issue and we'll look into it, thanks and apologies for the delay!

Issue #25685, opened 2024-10-22
[package] pybind11/0.2.13: CMakeDeps incorrectly links to pybind11_all_do_not_use


ZbigniewRA:
### Description

When CMakeDeps is used for `pybind11` it generates the following in `pybind11_json-Target-release.cmake` (tested on Windows):
```
set_property(TARGET pybind11_json_DEPS_TARGET
             APPEND PROPERTY INTERFACE_LINK_LIBRARIES
             $<$<CONFIG:Release>:${pybind11_json_FRAMEWORKS_FOUND_RELEASE}>
             $<$<CONFIG:Release>:${pybind11_json_SYSTEM_LIBS_RELEASE}>
             $<$<CONFIG:Release>:pybind11_all_do_not_use;nlohmann_json::nlohmann_json>)
```

As the name suggests `pybind11_all_do_not_use` is not meant to be linked to directly. Instead `pybind11::pybind11` and/or other components defined by `pybind11` should be linked.

I tried to fix the recipe by adding this to `package_info()`:
```
self.cpp_info.requires = ["pybind11::pybind11", "nlohmann_json::nlohmann_json"]
```
But it has no effect.
Martin Valgur kindly suggested using `pybind11::pybind11_` could work, but the result is even more surprising:
```
             $<$<CONFIG:Release>:pybind11::pybind11;nlohmann_json::nlohmann_json;pybind11_all_do_not_use>)
```

This problem is reproducible in the `test_package` included with `pybind11` recipe.  
Please note that `test_package` builds and works, but `pybind11_all_do_not_use` includes all components contained in `pybind11`, and is undesirable. It includes components both for module development and python embedding, and drags in assortment of components for very specific use cases, that are not meant to be the default. Hence the `do_not_use` name.

### Package and Environment Details

* Package Name/Version: **pybind11/0.2.13**
* Operating System+version: **Windows 11**
* Compiler+version: **Visual Studio 2022**
* Docker image: **N/A**
* Conan version: **conan 2.8.1**
* Python version: **Python 3.12.6**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows

Build profile:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows

[...]

PR #25703, opened 2024-10-23
pybind11_json: avoid linking pybind11_all_do_not_use

ZbigniewRA:
### Summary
Changes to recipe:  **pybind11_json/0.2.x**

#### Motivation
`CMakeDeps` incorrectly links to `pybind11_all_do_not_use`. See: https://github.com/conan-io/conan-center-index/issues/25685  
This is technically a breaking change, since `test_package` requires now linking to `pybind11::embed`. But this is how it should be, as the client project should choose whether `pybind11::embed` or `pybind11::module` should be linked.

#### Details
- Linking `pybind11_json` doesn't link to `pybind11_all_do_not_use` any more.
- Added version 0.2.14.

---

### DIFF
--- a/recipes/pybind11_json/all/conandata.yml
+++ b/recipes/pybind11_json/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.2.14":
+    url: "https://github.com/pybind/pybind11_json/archive/0.2.14.tar.gz"
+    sha256: "bc4ad7e308add59886a961c21f3ba431e43fe7faa2ef5bd9925c66d042d28cde"
   "0.2.13":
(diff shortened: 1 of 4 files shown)


ZbigniewRA:
Some help with fixing this PR would be nice. See: https://github.com/conan-io/conan-center-index/issues/25685

Ahajha:
The missing binary for cpython is due to requesting just `cpython/*:shared`. The Conan 2 pipeline only builds that binary with all of _its_ dependencies as shared too. Kinda annoying. So either you need to set a bunch of dependencies as shared, or the binary needs to be manually generated. They would need to be generated for all platforms. You also might run into invalid configurations in the test package, I suppose we'll see.

PR #25710, opened 2024-10-24
Add `.conanrunner` & `.conanrc` to gitignored ignored, alongside `conan_home`

AbrilRBS:
`.conanrunner` is a folder created by the runner feature when using it to create packages from docker, and it's helpful if it's already ignored by git

Alongside it `conan_home` is my (maybe in the docs too?) convention for the value of `CONAN_HOME` env var or the `conan_home` variable in `.conanrc` (Files which for now will also be ignored), which is an useful feature when trying to isolate runner builds, for which creating a separate conan_home for the build is usually the best approach

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ conanbuildinfo.txt
 conaninfo.txt
 graph_info.json
 build/
-conan_home/
 .conanrunner/
 .conanrc
 
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@ conanbuildinfo.txt
 conaninfo.txt
 graph_info.json
 build/
+conan_home/
+.conanrunner/
 
 # CMake
 CMakeUserPresets.json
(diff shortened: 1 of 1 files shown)


uilianries (approved):

.gitignore
```diff
@@ -8,6 +8,9 @@ conanbuildinfo.txt
 conaninfo.txt
 graph_info.json
 build/
+conan_home/
```
jcar87, line 11: Don't think this is a convention - mine is different altogether. If this isn't recommended or mentioned in the Conan client docs, I would remove this from .gitignore.

Remember you can always configure a global gitignore in your profile to achieve the same effect

jcar87 (changes requested):
see comment

.gitignore
AbrilRBS, line 11:
```suggestion
```

jcar87 (approved):

uilianries (approved):

PR #25717, opened 2024-10-25
[glfw] do not link explicitly to opengl

planetmarshall:
### Summary
Changes to recipe:  **glfw**

#### Motivation

Projects using GLFW cannot be cross compiled to targets that depend on OpenGLES rather than OpenGL, as the recipe introduces an explicit dependency on `libGL`

#### Details
Remove the explicit dependency on `opengl/system`. From [Building applications](https://github.com/glfw/glfw/blob/master/docs/build.md)

Additionally, GLFW always loads wayland and xkbcommon dynamically via `dlopen` or equivalent - so disable static linkage.

> _Note that the glfw target does not depend on OpenGL, as GLFW loads any OpenGL, OpenGL ES or Vulkan libraries it needs at runtime. If your application calls OpenGL directly, instead of using a modern extension loader library, use the OpenGL CMake package._

---

### DIFF
--- a/recipes/glfw/all/conanfile.py
+++ b/recipes/glfw/all/conanfile.py
@@ -61,6 +61,8 @@ class GlfwConan(ConanFile):
 
         if self.options.get_safe("with_wayland"):
             self.options["xkbcommon"].with_wayland = True
+            self.options["wayland"].shared = True
+            self.options["xkbcommon"].shared = True
 
     def layout(self):
         cmake_layout(self, src_folder="src")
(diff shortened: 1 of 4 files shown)


jcar87:
Thanks @planetmarshall  - for providing links to the documentation as well.
Indeed it would be correct to not "link" the opengl target - as as you mention it is loaded at runtime (via `dlopen` or equivalent) - however, based on what we've seen, opengl _is_ a transitive (implicit) dependency of glfw, and removing the dependency on opengl may actually break consumers that were not explicitly linking it on the consumer side. The glfw documentation mentions that this would be necessary in the case where the consumer is making direct explicit calls to the OpenGL APIs.

In the end we've gone for somethign we consider lower risk: use the `libs=False` requirement trait - this still causes the dependency, ensures the headers are available for the consumer, but inhibits Conan from expressing it at "link time" - so the glfw library should *not* link against libgl.so (or any, for that matter), and instead do the dlopen correctly - please let us know if this solves your issue.

PR #25745, opened 2024-10-28
new package: dpp 10.0.34

braindigitalis:
### Summary
Adds new recipe  **dpp/10.0.34**

#### Motivation
This library does not exist in Conan yet.

#### Details
Adds the initial build for dpp 10.0.34. I am the lead developer of the library in the recipe.
Installs from a zip of the release.

---

### DIFF
--- a/recipes/dpp/all/conandata.yml
+++ b/recipes/dpp/all/conandata.yml
@@ -2,3 +2,4 @@ sources:
   "10.0.34":
     url:
     - "https://github.com/brainboxdotcc/DPP/archive/refs/tags/v10.0.34.zip"
+    sha256: "944dcaa19638ff88feceae2f543662675eb8aceb59458ad369edc28819e05054"
--- a/recipes/dpp/all/conandata.yml
+++ b/recipes/dpp/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "10.0.34":
+    url:
+    - "https://github.com/brainboxdotcc/DPP/archive/refs/tags/v10.0.34.zip"
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -2,8 +2,7 @@ import os
 from conan import ConanFile
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
-from conan.tools.scm import Git
-from conan.tools.files import download, unzip, get, copy
+from conan.tools.files import get, copy
 
 required_conan_version = ">=1.54"
 
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -63,6 +63,7 @@ class DPPConan(ConanFile):
         cmake.build()
 
     def package(self):
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
         cmake = CMake(self)
         cmake.install()
 
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -64,6 +64,7 @@ class DPPConan(ConanFile):
         tc.cache_variables["BUILD_VOICE_SUPPORT"] = True
         tc.cache_variables["DPP_BUILD_TEST"] = False
         tc.cache_variables["BUILD_SHARED_LIBS"] = True
+        tc.cache_variables["AVX_TYPE"] = "AVX0"
         tc.generate()
 
     def build(self):
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -84,4 +84,7 @@ class DPPConan(ConanFile):
         self.cpp_info.libs = ["dpp"]
         self.cpp_info.set_property("cmake_file_name", "dpp")
         self.cpp_info.set_property("cmake_target_name", "dpp::dpp")
-
+        # On windows only, the headers and libs go into dpp-10.0 subdirectories.
+        if self.settings.os == "Windows":
+            self.cpp_info.includedirs = ["include/dpp-10.0"]
+            self.cpp_info.libdirs = ["lib/dpp-10.0"]
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -36,9 +36,6 @@ class DPPConan(ConanFile):
     def layout(self):
         cmake_layout(self)
 
-    def export(self):
-        git = Git(self, self.recipe_folder)
-        git.coordinates_to_conandata()
 
     def build_requirements(self):
         self.tool_requires("cmake/[>=3.16 <4]")
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -3,7 +3,7 @@ from conan import ConanFile
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
 from conan.tools.scm import Git
-from conan.tools.files import download, unzip
+from conan.tools.files import download, unzip, get
 
 required_conan_version = ">=1.65"
 
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -4,6 +4,7 @@ from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
 from conan.tools.files import get, copy
+from conan.tools.scm import Version
 
 required_conan_version = ">=1.54"
 
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -29,8 +29,8 @@ class DPPConan(ConanFile):
 
     def requirements(self):
         self.requires("nlohmann_json/3.11.2")
-        self.requires("openssl/3.1.2")
-        self.requires("zlib/1.3")
+        self.requires("openssl/[>=1.1 <4]")
+        self.requires("zlib/[>=1.2.11 <2]")
         self.requires("opus/1.4")
 
     def layout(self):
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -9,7 +9,6 @@ required_conan_version = ">=1.65"
 
 class DPPConan(ConanFile):
     name = "dpp"
-    version = "10.0.34"
     license = "Apache-2.0"
     package_type = "shared-library"
     url = "https://github.com/conan-io/conan-center-index"
@@ -45,20 +44,12 @@ class DPPConan(ConanFile):
         self.tool_requires("cmake/[>=3.16 <4]")
 
     def source(self):
-        # This environment variable should only be set by D++ library developers to ensure that conan builds succeed
-        # without having to wait for release. It will check out the development branch where conanfile.py is being
-        # developed and tested. If you are NOT sure what this does, DO NOT SET IT. You won't get the D++ release
-        # you expect!
-        if 'DPP_CONAN_TESTING' in os.environ:
-            git = Git(self)
-            git.clone(url="https://github.com/brainboxdotcc/DPP.git", target=".")
-            git.checkout(commit="conan-the-librarian")
-        else:
-            zip_name = "DPP.zip"
-            download(self, f"https://github.com/brainboxdotcc/DPP/archive/refs/tags/v{self.version}.zip", zip_name)
-            unzip(self, zip_name, '.', False, None, True)
-            os.unlink(zip_name)
-
+        #zip_name = "DPP.zip"
+        #download(self, f"https://github.com/brainboxdotcc/DPP/archive/refs/tags/v{self.version}.zip", zip_name)
+        #unzip(self, zip_name, '.', False, None, True)
+        #os.unlink(zip_name)
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+  
     def generate(self):
         deps = CMakeDeps(self)
         deps.generate()
@@ -82,3 +73,8 @@ class DPPConan(ConanFile):
         self.cpp_info.libs = ["dpp"]
         self.cpp_info.includedirs = ["include/dpp-10.0"]
         self.cpp_info.libdirs = ["lib/dpp-10.0"]
+    def package_info(self):
+        self.cpp_info.libs = ["package_lib"]
+        self.cpp_info.set_property("cmake_file_name", "dpp")
+        self.cpp_info.set_property("cmake_target_name", "dpp::dpp")
+
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -61,6 +61,8 @@ class DPPConan(ConanFile):
         deps = CMakeDeps(self)
         deps.generate()
         tc = CMakeToolchain(self)
+        tc.cache_variables["DPP_NO_VCPKG"] = True
+        tc.cache_variables["DPP_USE_EXTERNAL_JSON"] = True
         tc.cache_variables["CONAN_EXPORTED"] = True
         tc.cache_variables["BUILD_VOICE_SUPPORT"] = True
         tc.cache_variables["DPP_BUILD_TEST"] = False
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -5,7 +5,7 @@ from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
 from conan.tools.scm import Git
 from conan.tools.files import download, unzip
 
-required_conan_version = ">=2.0"
+required_conan_version = ">=1.65"
 
 class DPPConan(ConanFile):
     name = "dpp"
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -47,6 +47,9 @@ class DPPConan(ConanFile):
         git = Git(self, self.recipe_folder)
         git.coordinates_to_conandata()
 
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.16 <4]")
+
     def source(self):
         # This environment variable should only be set by D++ library developers to ensure that conan builds succeed
         # without having to wait for release. It will check out the development branch where conanfile.py is being
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -5,7 +5,7 @@ from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
 from conan.tools.scm import Git
 from conan.tools.files import download, unzip
 
-required_conan_version = ">=1.65"
+required_conan_version = ">=2.0"
 
 class DPPConan(ConanFile):
     name = "dpp"
@@ -15,8 +15,8 @@ class DPPConan(ConanFile):
     description = "D++ is a lightweight and efficient library for Discord"
     topics = ("discord")
     settings = "os", "compiler", "build_type", "arch"
-    options = {"shared": [True, False], "fPIC": [True, False]}
-    default_options = {"shared": True, "fPIC": True}
+    options = {"shared": [True, False]}
+    default_options = {"shared": True}
 
     @property
     def _min_cppstd(self):
@@ -31,14 +31,6 @@ class DPPConan(ConanFile):
         self.requires("openssl/3.1.2")
         self.requires("zlib/1.3")
         self.requires("opus/1.4")
-
-    def config_options(self):
-        if self.settings.os == "Windows":
-            self.options.rm_safe("fPIC")
-
-    def configure(self):
-        if self.options.shared:
-            self.options.rm_safe("fPIC")
 
     def layout(self):
         cmake_layout(self)
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -11,6 +11,7 @@ class DPPConan(ConanFile):
     name = "dpp"
     version = "10.0.34"
     license = "Apache-2.0"
+    package_type = "shared-library"
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/brainboxdotcc/DPP"
     description = "D++ is a lightweight and efficient library for Discord"
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -5,13 +5,14 @@ from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
 from conan.tools.scm import Git
 from conan.tools.files import download, unzip
 
-required_conan_version = ">=2.0"
+required_conan_version = ">=1.65"
 
 class DPPConan(ConanFile):
     name = "dpp"
     version = "10.0.34"
     license = "Apache-2.0"
-    url = "https://github.com/brainboxdotcc/DPP"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/brainboxdotcc/DPP"
     description = "D++ is a lightweight and efficient library for Discord"
     topics = ("discord")
     settings = "os", "compiler", "build_type", "arch"
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -42,7 +42,7 @@ class DPPConan(ConanFile):
             )
 
     def requirements(self):
-        self.requires("nlohmann_json/3.11.2")
+        self.requires("nlohmann_json/3.11.2", transitive_libs=True, transitive_headers=True)
         self.requires("openssl/[>=1.1 <4]")
         self.requires("zlib/[>=1.2.11 <2]")
         self.requires("opus/1.4")
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -1,5 +1,6 @@
 import os
 from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
 from conan.tools.files import get, copy
@@ -20,9 +21,24 @@ class DPPConan(ConanFile):
     def _min_cppstd(self):
         return 17
 
+    @property
+    def _compilers_minimum_version(self):
+        return {
+            "apple-clang": "14",
+            "clang": "10",
+            "gcc": "8",
+            "msvc": "191",
+            "Visual Studio": "16",
+        }
+
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):
             check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
 
     def requirements(self):
         self.requires("nlohmann_json/3.11.2")
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -1,0 +1,87 @@
+import os
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.scm import Git
+from conan.tools.files import download, unzip
+
+required_conan_version = ">=2.0"
+
+class DPPConan(ConanFile):
+    name = "dpp"
+    version = "10.0.34"
+    license = "Apache-2.0"
+    url = "https://github.com/brainboxdotcc/DPP"
+    description = "D++ is a lightweight and efficient library for Discord"
+    topics = ("discord")
+    settings = "os", "compiler", "build_type", "arch"
+    options = {"shared": [True, False], "fPIC": [True, False]}
+    default_options = {"shared": True, "fPIC": True}
+
+    @property
+    def _min_cppstd(self):
+        return 17
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+
+    def requirements(self):
+        self.requires("nlohmann_json/3.11.2")
+        self.requires("openssl/3.1.2")
+        self.requires("zlib/1.3")
+        self.requires("opus/1.4")
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            self.options.rm_safe("fPIC")
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+
+    def layout(self):
+        cmake_layout(self)
+
+    def export(self):
+        git = Git(self, self.recipe_folder)
+        git.coordinates_to_conandata()
+
+    def source(self):
+        # This environment variable should only be set by D++ library developers to ensure that conan builds succeed
+        # without having to wait for release. It will check out the development branch where conanfile.py is being
+        # developed and tested. If you are NOT sure what this does, DO NOT SET IT. You won't get the D++ release
+        # you expect!
+        if 'DPP_CONAN_TESTING' in os.environ:
+            git = Git(self)
+            git.clone(url="https://github.com/brainboxdotcc/DPP.git", target=".")
+            git.checkout(commit="conan-the-librarian")
+            zip_name = "DPP.zip"
+        else:
+            download(self, f"https://github.com/brainboxdotcc/DPP/archive/refs/tags/v{self.version}.zip", zip_name)
+            unzip(self, zip_name, '.', False, None, True)
+            os.unlink(zip_name)
+
+    def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.cache_variables["CONAN_EXPORTED"] = True
+        tc.cache_variables["BUILD_VOICE_SUPPORT"] = True
+        tc.cache_variables["DPP_BUILD_TEST"] = False
+        tc.cache_variables["BUILD_SHARED_LIBS"] = True
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["dpp"]
+        self.cpp_info.includedirs = ["include/dpp-10.0"]
+        self.cpp_info.libdirs = ["lib/dpp-10.0"]
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -3,7 +3,7 @@ from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
-from conan.tools.files import get, copy
+from conan.tools.files import get, copy, rmdir
 from conan.tools.scm import Version
 
 required_conan_version = ">=1.54"
@@ -77,6 +77,8 @@ class DPPConan(ConanFile):
         copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
         cmake = CMake(self)
         cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
 
     def package_info(self):
         self.cpp_info.libs = ["dpp"]
--- a/recipes/dpp/all/conanfile.py
+++ b/recipes/dpp/all/conanfile.py
@@ -56,8 +56,8 @@ class DPPConan(ConanFile):
             git = Git(self)
             git.clone(url="https://github.com/brainboxdotcc/DPP.git", target=".")
             git.checkout(commit="conan-the-librarian")
-            zip_name = "DPP.zip"
         else:
+            zip_name = "DPP.zip"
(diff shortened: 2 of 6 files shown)


braindigitalis:
> ### Conan v1 pipeline ❌
> 
> Warning
> 
> Conan Center will stop receiving updates for Conan 1.x packages soon - please see [announcement](https://github.com/conan-io/conan-center-index/discussions/25461).
> 
> Changes not allowed in build 1:
> 
> ```
> [recipes/dpp/all/conanfile.py, recipes/dpp/config.yml]
> ```
> 
> The folder `test_package` is required for every recipe. Please, add it for `dpp/all/test_package`.
> ### Conan v2 pipeline ❌
> 
> 
> The v2 pipeline failed. Please, review the errors and note this is required for pull requests to be merged. In case this recipe is still not ported to Conan 2.x, please, ping `@conan-io/barbarians` on the PR and we will help you.
> 
> Changes not allowed in build 1:
> 
> ```
> [recipes/dpp/all/conanfile.py, recipes/dpp/config.yml]
> ```
> 
> The folder `test_package` is required for every recipe. Please, add it for `dpp/all/test_package`.

how exactly is this to be tested with a test package? it is a discord library and requires a token to test it?

AbrilRBS:
Hi @braindigitalis thanks a lot for taking the time to add this to CCI, we appreciate it!

A few points about the PR:

I recommend you use the package template available in https://github.com/conan-io/conan-center-index/tree/master/docs/package_templates/cmake_package as the basis for your package, which will improve the recipe and simplify the review process

> how exactly is this to be tested with a test package? it is a discord library and requires a token to test it?

There's no need to actually run the library/tests - usually ensuring that a call to any of the functions implemented in a `.cpp` file is enough for the test_package, ideally something that interacts with the dependencies, but we don't want to actually make connections to any server and the like. Usually this translates into calling a version function, or instantiating a client but not connecting it, or something along those lines. We can help find the right think to do once the recipe has been transformed to the package template linked above

(Also note that currently your PR is not set to allow us to directly commit on your branch, which limits how useful we can be while reviewing, I also recommend enabling it here, thanks!)

Again, thanks for the effort in opening the PR, happy to help where needed :)

braindigitalis:
... why are there no logs?
i can't exactly diagnose and fix if there are no logs from the CI.
DPP builds successfully on all of the platforms listed above, outside of conan. they are all officially supported.

Also the reason you cannot contribute to the PR is because it is from within an organisation, this is a known issue in github. Someone reported it as an issue in 2021 and github being github, still havent fixed it.

uilianries:
@braindigitalis Hello and thank you for your PR! Please, take a look in the build log available on the comment above: https://github.com/conan-io/conan-center-index/pull/25745#issuecomment-2443622585

As Abril commented, it's highly recommended using the templates to avoid extra changes needed. You actual recipe misses conandata.yml and requires Conan 2.x only (Conan 1.x is still mandatory in CCI).

braindigitalis:
everything I've read says you're getting rid of Conan 1.x, surely we aren't expected to still support it too? I'm confused now as I've spent a long time getting this ready using only Conan 2 stuff.

braindigitalis:
your template appears to be full of irrelevant stuff that is not needed by this library. why do you have stuff in the template to toggle fPIC, this isn't usually something we let the end user decide on, as it simply breaks builds...? a lot of stuff in there for detection of compiler etc we do in dpp's makefile, seems a lot of it is to work around broken cmakelists on projects that are not properly portable... is this the intent of the template?

also this cannot be worked around.

```
CMake 3.16 or higher is required.  You are running version 3.15.7
```

we require cmake 3.16 (around since 2019!), what's with the ancient version?

uilianries:
> everything I've read says you're getting rid of Conan 1.x, surely we aren't expected to still support it too? I'm confused now as I've spent a long time getting this ready using only Conan 2 stuff.

Most of Conan 2.x features are compatible to Conan 1.x so all recipes in CCI are running for both major versions. Indeed there are new features for Conan 2.x that really nice and will not be ported to Conan 1.x. Specifically speaking about your current recipe only `coordinates_to_conandata` is a Conan 2.x feature and not available for Conan 1.x

The Conan 1.x support in ConanCenterIndex should be ended by the next week, so you can wait until the next week to re-visit this PR, or just update your recipe to be both Conan 1.x and Conan 2.x compatible.

Still, we use [conandata.yml](https://docs.conan.io/2/tutorial/creating_packages/handle_sources_in_packages.html#using-the-conandata-yml-file) and not Git commands directly in the recipe. The main advantage is being able to parse, validate and conandata; and re-use the same file for multiple versions without touching the recipe. The template shows how to use conandata in ConanCenterIndex: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L98

Using conandata.yml will make your recipe compatible with both Conan 1.x and 2.x btw

> your template appears to be full of irrelevant stuff that is not needed by this library. why do you have stuff in the template to toggle fPIC, this isn't usually something we let the end user decide on, as it simply breaks builds...? a lot of stuff in there for detection of compiler etc we do in dpp's makefile, seems a lot of it is to work around broken cmakelists on projects that are not properly portable... is this the intent of the template?

The template covers a general cmake project, most compatible with almost all project that are using cmake in ConanCenterIndex. The fPIC and shared option are configurable yes, and for most of projects it's transparent as CMake itself is capable to manage it.

> CMake 3.16 or higher is required.  You are running version 3.15.7

Please, read the design decision from 2020: https://github.com/conan-io/tribe/blob/main/design/004-tools-cmake.md

In case you need >=3.15, which is totally fine, you can use tool_requirements, as indicate in the template: https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L95

braindigitalis:
how do i get the test errors please? They were messaged into this thread last time, this time i'm not seeing them?

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.scm import Git
+from conan.tools.files import download, unzip
+
+required_conan_version = ">=1.65"
+
+class DPPConan(ConanFile):
+    name = "dpp"
+    license = "Apache-2.0"
+    package_type = "shared-library"
```
AbrilRBS, line 13: If this is alwas a shared library, then we should not have a `shared` option either (Or the other way around! Set this to `library` to allow both static and shared variants)

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/brainboxdotcc/DPP"
+    description = "D++ is a lightweight and efficient library for Discord"
+    topics = ("discord")
+    settings = "os", "compiler", "build_type", "arch"
+    options = {"shared": [True, False]}
+    default_options = {"shared": True}
```
AbrilRBS, line 20: Note that if as per the comment above both static and shared are allowed, CCI recipes usually have shared: False as their default

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        return 17
+
+    def validate(self):
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+
+    def requirements(self):
+        self.requires("nlohmann_json/3.11.2")
+        self.requires("openssl/3.1.2")
+        self.requires("zlib/1.3")
```
AbrilRBS, line 33:
```suggestion
        self.requires("openssl/[>=1.1 <4]")
        self.requires("zlib/[>=1.2.11 <2]")
```

Some libraries can be made version ranges as per https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges :)

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+  
+    def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.cache_variables["CONAN_EXPORTED"] = True
+        tc.cache_variables["BUILD_VOICE_SUPPORT"] = True
+        tc.cache_variables["DPP_BUILD_TEST"] = False
+        tc.cache_variables["BUILD_SHARED_LIBS"] = True
```
AbrilRBS, line 60: This can be dropped if we keep the shared option, Conan already sets this

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["dpp"]
+        self.cpp_info.includedirs = ["include/dpp-10.0"]
+        self.cpp_info.libdirs = ["lib/dpp-10.0"]
+    def package_info(self):
+        self.cpp_info.libs = ["package_lib"]
+        self.cpp_info.set_property("cmake_file_name", "dpp")
+        self.cpp_info.set_property("cmake_target_name", "dpp::dpp")
```
AbrilRBS, line 79:
```suggestion
```

typo!

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        self.requires("openssl/3.1.2")
+        self.requires("zlib/1.3")
+        self.requires("opus/1.4")
+
+    def layout(self):
+        cmake_layout(self)
+
+    def export(self):
+        git = Git(self, self.recipe_folder)
+        git.coordinates_to_conandata()
```
AbrilRBS, line 41:
```suggestion
```

This method currently has no effect in CCI's case and should not be used, all the sources come from the `source()` method already

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+            check_min_cppstd(self, self._min_cppstd)
+
+    def requirements(self):
+        self.requires("nlohmann_json/3.11.2")
+        self.requires("openssl/3.1.2")
+        self.requires("zlib/1.3")
+        self.requires("opus/1.4")
+
+    def layout(self):
+        cmake_layout(self)
```
AbrilRBS, line 37:
```suggestion
        cmake_layout(self, src_folder="src")
```

so that calling `conan source . --version=10.0.34` does not pollute the current working directory mostly

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
+import os
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.scm import Git
+from conan.tools.files import download, unzip, get
+
+required_conan_version = ">=1.65"
```
AbrilRBS, line 8:
```suggestion
required_conan_version = ">=1.54"
```

Nothing currently using a newer Conan version

recipes/dpp/all/conanfile.py
braindigitalis: is the intent here to remove the whole method?

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        tc.cache_variables["BUILD_SHARED_LIBS"] = True
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
```
AbrilRBS, line 69:
```suggestion
        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
        cmake = CMake(self)
```

(Plus the needed import in Line 6) to keep the license around in the final package

recipes/dpp/all/conanfile.py
braindigitalis: shared is better supported by our library than static, which is why we set the default to static: true. in fact, it is preferred that people don't build the library static unless they really know what they are doing as it will then bring in static openssl which is a security risk.

recipes/dpp/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "10.0.34":
+    url:
+    - "https://github.com/brainboxdotcc/DPP/archive/refs/tags/v10.0.34.zip"
```
AbrilRBS, line 4:
```suggestion
    - "https://github.com/brainboxdotcc/DPP/archive/refs/tags/v10.0.34.zip"
    sha256: "944dcaa19638ff88feceae2f543662675eb8aceb59458ad369edc28819e05054"
```

We need to keep a checksum to ensure no changes to the sources go unnoticed

recipes/dpp/all/conanfile.py
braindigitalis: BUILD_SHARED_LIBS does much more in our build system so it should be kept and set to a value based on the self.default_options.shared

recipes/dpp/all/conanfile.py
AbrilRBS: I see. Then I'd suggest not allowing a static build in this case if that's ok with you (So removing the shared option from the recipe and keeping the package_type as-is)

braindigitalis:
```
[HOOK - conan-center.py] pre_export(): WARN: [FPIC OPTION (KB-H006)] This recipe does not include an 'fPIC' option. Make sure you are using the right casing
```

We absolutely do NOT want end users changing this! It breaks our build! Why is it mandatory to have? We are on vcpkg, aur, gentoo, freebsd ports, and a ton of other places and none of them require this. are we ok to ignore its setting?

recipes/dpp/all/conanfile.py
braindigitalis: sounds like a plan, i'll do that

recipes/dpp/all/conanfile.py
braindigitalis: including the define of the method name?

AbrilRBS:
> We absolutely do NOT want end users changing this! It breaks our build! Why is it mandatory to have? We are on vcpkg, aur, gentoo, freebsd ports, and a ton of other places and none of them require this. are we ok to ignore its setting?

No worries @braindigitalis, this is just a warning to ensure we don't miss missnamings when the shared option is available, as that needs a bit of extra care on our side

The recipe now looks great, let's see if this compiles ok, thanks for quickly adressing the PR reviews!

braindigitalis:
> > We absolutely do NOT want end users changing this! It breaks our build! Why is it mandatory to have? We are on vcpkg, aur, gentoo, freebsd ports, and a ton of other places and none of them require this. are we ok to ignore its setting?
> 
> No worries @braindigitalis, this is just a warning to ensure we don't miss missnamings when the shared option is available, as that needs a bit of extra care on our side
> 
> The recipe now looks great, let's see if this compiles ok, thanks for quickly adressing the PR reviews!

Thanks! It was failing to build the test earlier, not sure why. I couldnt get the error log to find out. Hopefully all the other changes we have made have fixed that.

braindigitalis:
can you please shed some light on what these failures are about @AbrilRBS ? 

The 1.6 failure seems strange, as std::filesystem is something we detect internally and have a find script for. Conan should not be intercepting this, and it seems it is fine in 2.0 but not 1.6.

The 2.0 failure is indeed in the test program, but this test program works fine if i build and install dpp to my system outside conan, a dppConfig.cmake is installed during the install process:
```
-- Installing: /usr/local/lib/cmake/dpp/dpp-config.cmake
-- Installing: /usr/local/lib/cmake/dpp/dpp-config-version.cmake
-- Installing: /usr/local/lib/cmake/dpp/dpp.cmake
-- Installing: /usr/local/lib/cmake/dpp/dpp-debug.cmake
```
This being the case why would conan fail? does it do some voodoo here?

braindigitalis:
wait a minute.... why is it trying to use gcc 5?!

```
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=5
```

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def package(self):
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["dpp"]
+        self.cpp_info.includedirs = ["include/dpp-10.0"]
+        self.cpp_info.libdirs = ["lib/dpp-10.0"]
```
AbrilRBS, line 66:
```suggestion
```

This is causing the test_package issues. Conan already sets this to `lib` and `include` by default, and locally I'm seeing

```
-- Installing: /root/.conan2/p/b/dppcae3bf40fb4b5/p/lib/libdpp.so.10.0.34
-- Set non-toolchain portion of runtime path of "/root/.conan2/p/b/dppcae3bf40fb4b5/p/lib/libdpp.so.10.0.34" to ""
-- Installing: /root/.conan2/p/b/dppcae3bf40fb4b5/p/lib/libdpp.so
-- Installing: /root/.conan2/p/b/dppcae3bf40fb4b5/p/include
-- Installing: /root/.conan2/p/b/dppcae3bf40fb4b5/p/include/dpp
-- Installing: /root/.conan2/p/b/dppcae3bf40fb4b5/p/include/dpp/sslclient.h
-- Installing: /root/.conan2/p/b/dppcae3bf40fb4b5/p/include/dpp/user.h
```
so the default installation _paths_ are a-ok

AbrilRBS (commented):
Fix for the test_package issue.

Also, when compiling locally for mac, I'm seeing:

```
[ 42%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dave/version.cpp.o
[ 43%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/discordclient.cpp.o
[ 43%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/discordevents.cpp.o
[ 44%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/discordvoiceclient.cpp.o
[ 44%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dispatcher.cpp.o
[ 45%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dns.cpp.o
[ 45%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dtemplate.cpp.o
[ 46%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/emoji.cpp.o
[ 46%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/entitlement.cpp.o
[ 47%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/etf.cpp.o
[ 47%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_create.cpp.o
/Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:250:92: warning: adding 'uint64_t' (aka 'unsigned long long') to a string does not append to the string [-Wstring-plus-int]
  250 |                 creator.log(dpp::ll_warning, "Attempted to verify credential for unrecognized user ID: " + uid);
      |                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
/Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:250:92: note: use array indexing to silence this warning
  250 |                 creator.log(dpp::ll_warning, "Attempted to verify credential for unrecognized user ID: " + uid);
      |                                                                                                          ^    
      |                                              &                                                           [    ]
/Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:643:51: warning: adding 'uint64_t' (aka 'unsigned long long') to a string does not append to the string [-Wstring-plus-int]
  643 |                 throw std::invalid_argument("Unknown user ID: " + user_id);
      |                                             ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
/Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:643:51: note: use array indexing to silence this warning
  643 |                 throw std::invalid_argument("Unknown user ID: " + user_id);
      |                                                                 ^        
      |                                             &                   [        ]
[ 48%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_delete.cpp.o
In file included from /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/discordvoiceclient.cpp:32:
In file included from /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/voice/enabled/enabled.h:42:
In file included from /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/library/../include/dpp/isa_detection.h:24:
/Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/library/../include/dpp/isa/neon.h:59:75: error: extraneous ')' before ';'
   59 |                         neon_float increment_vector = vmulq_f32(vdupq_n_f32(increment), floats));
      |                                                                                                ^
[ 48%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_execute.cpp.o
[ 49%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_update.cpp.o
[ 49%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/channel_create.cpp.o
[ 50%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/channel_delete.cpp.o
1 error generated.
make[2]: *** [library/CMakeFiles/dpp.dir/__/src/dpp/discordvoiceclient.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
2 warnings generated.
make[1]: *** [library/CMakeFiles/dpp.dir/all] Error 2
make: *** [all] Error 2

dpp/10.0.34: ERROR: 
Package 'af8638d5202ad1fe911bcc8efbce95e7cc7fc7cc' build failed
dpp/10.0.34: WARN: Build folder /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/build/Release
ERROR: dpp/10.0.34: Error in build() method, line 56
	cmake.build()
	ConanException: Error 2 while executing
```

Running with
```
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=16
os=Macos

```

Is this a known issue? Thanks!

AbrilRBS:
> wait a minute.... why is it trying to use gcc 5?!

@braindigitalis Conan Center Index builds for a variety of configurations, and it's up to the recipes to state which ones they do not support. 

See https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/cmake_package/all/conanfile.py#L47 for how this is done elsewhere, and fill the versions as needed. This will also help users building from sources when building for older configuations to have a clear error message

braindigitalis:
> Fix for the test_package issue.
> 
> Also, when compiling locally for mac, I'm seeing:
> 
> ```
> [ 42%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dave/version.cpp.o
> [ 43%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/discordclient.cpp.o
> [ 43%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/discordevents.cpp.o
> [ 44%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/discordvoiceclient.cpp.o
> [ 44%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dispatcher.cpp.o
> [ 45%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dns.cpp.o
> [ 45%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/dtemplate.cpp.o
> [ 46%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/emoji.cpp.o
> [ 46%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/entitlement.cpp.o
> [ 47%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/etf.cpp.o
> [ 47%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_create.cpp.o
> /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:250:92: warning: adding 'uint64_t' (aka 'unsigned long long') to a string does not append to the string [-Wstring-plus-int]
>   250 |                 creator.log(dpp::ll_warning, "Attempted to verify credential for unrecognized user ID: " + uid);
>       |                                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
> /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:250:92: note: use array indexing to silence this warning
>   250 |                 creator.log(dpp::ll_warning, "Attempted to verify credential for unrecognized user ID: " + uid);
>       |                                                                                                          ^    
>       |                                              &                                                           [    ]
> /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:643:51: warning: adding 'uint64_t' (aka 'unsigned long long') to a string does not append to the string [-Wstring-plus-int]
>   643 |                 throw std::invalid_argument("Unknown user ID: " + user_id);
>       |                                             ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
> /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/dave/session.cpp:643:51: note: use array indexing to silence this warning
>   643 |                 throw std::invalid_argument("Unknown user ID: " + user_id);
>       |                                                                 ^        
>       |                                             &                   [        ]
> [ 48%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_delete.cpp.o
> In file included from /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/discordvoiceclient.cpp:32:
> In file included from /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/src/dpp/voice/enabled/enabled.h:42:
> In file included from /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/library/../include/dpp/isa_detection.h:24:
> /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/src/library/../include/dpp/isa/neon.h:59:75: error: extraneous ')' before ';'
>    59 |                         neon_float increment_vector = vmulq_f32(vdupq_n_f32(increment), floats));
>       |                                                                                                ^
> [ 48%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_execute.cpp.o
> [ 49%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/automod_rule_update.cpp.o
> [ 49%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/channel_create.cpp.o
> [ 50%] Building CXX object library/CMakeFiles/dpp.dir/__/src/dpp/events/channel_delete.cpp.o
> 1 error generated.
> make[2]: *** [library/CMakeFiles/dpp.dir/__/src/dpp/discordvoiceclient.cpp.o] Error 1
> make[2]: *** Waiting for unfinished jobs....
> 2 warnings generated.
> make[1]: *** [library/CMakeFiles/dpp.dir/all] Error 2
> make: *** [all] Error 2
> 
> dpp/10.0.34: ERROR: 
> Package 'af8638d5202ad1fe911bcc8efbce95e7cc7fc7cc' build failed
> dpp/10.0.34: WARN: Build folder /Users/abril/coding/conan-center-index/recipes/dpp/all/conan_home/p/b/dpp55d84446a36e9/b/build/Release
> ERROR: dpp/10.0.34: Error in build() method, line 56
> 	cmake.build()
> 	ConanException: Error 2 while executing
> ```
> 
> Running with
> 
> ```
> arch=armv8
> build_type=Release
> compiler=apple-clang
> compiler.cppstd=gnu17
> compiler.libcxx=libc++
> compiler.version=16
> os=Macos
> ```
> 
> Is this a known issue? Thanks!

thanks, this is a known issue, but i forgot to add a workaround. We just need this:

```cmake
tc.cache_variables["AVX_TYPE"] = "AVX0"
```
to force DPP to not attempt to enable AVX/NEON for now, until this can be patched properly upstream. It isn't something i'd want to try and hack in downstream with a patch in conan, and it doesnt cause any issue to disable NEON except a slight loss of performance when mixing received audio which isnt a feature many users use anyway.

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
```
AbrilRBS, line 79:
```suggestion
        cmake.install()
        rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
```

plus the import for rmdir above. Conan generates the config files needed for other Conan dependencies to use them

braindigitalis:
what can i do with this, should i just rm the file from the output after running the build process?

```cmake
dpp/10.0.34: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
dpp/10.0.34: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
[HOOK - conan-center.py] post_package(): ERROR: [PC-FILES (KB-H020)] The conan-center repository doesn't allow the packages to contain `pc` files. The packages have to be located using generators and the declared `cpp_info` information (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H020-PC-FILES) 
[HOOK - conan-center.py] post_package(): ERROR: [PC-FILES (KB-H020)] Found files: ./lib/pkgconfig/dpp.pc (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H020-PC-FILES) 
[HOOK - conan-center.py] post_package(): ERROR: [CMAKE-MODULES-CONFIG-FILES (KB-H016)] The conan-center repository doesn't allow the packages to contain CMake find modules or config files. The packages have to be located using generators and the declared `cpp_info` information (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H016-CMAKE-MODULES-CONFIG-FILES) 
[HOOK - conan-center.py] post_package(): ERROR: [CMAKE-MODULES-CONFIG-FILES (KB-H016)] Found files: ./lib/cmake/dpp/dpp-config.cmake (https://github.com/conan-io/conan-center-index/blob/master/docs/error_knowledge_base.md#KB-H016-CMAKE-MODULES-CONFIG-FILES) 
ERROR: [HOOK - conan-center.py] post_package(): Some checks failed running the hook, check the output
```

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,85 @@
+import os
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
+from conan.tools.files import get, copy
```
AbrilRBS, line 6:
```suggestion
from conan.tools.files import get, copy, rmdir
```

braindigitalis: thanks, i'll merge these, first i want to see how far the conan 2 build gets

recipes/dpp/all/conanfile.py
braindigitalis: Turns out this is needed but only when building for windows. On windows its placed in dpp-10.0 for the lib and includes.

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+  
+    def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
+        tc.cache_variables["CONAN_EXPORTED"] = True
+        tc.cache_variables["BUILD_VOICE_SUPPORT"] = True
+        tc.cache_variables["DPP_BUILD_TEST"] = False
+        tc.cache_variables["BUILD_SHARED_LIBS"] = True
+        tc.cache_variables["AVX_TYPE"] = "AVX0"
```
uilianries, line 68:
```suggestion
        tc.cache_variables["AVX_TYPE"] = False
```
The default for AVX_TYPE is disabled in the upstream, why should be customized here?

https://github.com/brainboxdotcc/DPP/blob/v10.0.34/CMakeLists.txt#L33C8-L33C16

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.16 <4]")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+  
+    def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
+        tc = CMakeToolchain(self)
```
uilianries, line 63:
```suggestion
        tc = CMakeToolchain(self)
        tc.cache_variables["DPP_NO_VCPKG"] = True
        tc.cache_variables["DPP_USE_EXTERNAL_JSON"] = True
```

- Avoid running vcpkg in case installed in the user's system: https://github.com/brainboxdotcc/DPP/blob/v10.0.34/CMakeLists.txt#L26
- Enforce using nlohmann::json provided by Conan: https://github.com/brainboxdotcc/DPP/blob/v10.0.34/CMakeLists.txt#L31

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,90 @@
...
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+
+    def package_info(self):
+        self.cpp_info.libs = ["dpp"]
+        self.cpp_info.set_property("cmake_file_name", "dpp")
+        self.cpp_info.set_property("cmake_target_name", "dpp::dpp")
+        # On windows only, the headers and libs go into dpp-10.0 subdirectories.
+        if self.settings.os == "Windows":
+            self.cpp_info.includedirs = ["include/dpp-10.0"]
+            self.cpp_info.libdirs = ["lib/dpp-10.0"]
```
uilianries, line 90:
```suggestion
            self.cpp_info.libdirs = ["lib/dpp-10.0"]
        elif self.settings.os in ["Linux", "FreeBSD"]:
            self.cpp_info.system_libs = ["pthread"]
        self.cpp_info.defines = ["DPP_USE_EXTERNAL_JSON"]
```

- The project requires threads, but Conan will not magically solve it when consuming the library, need to explicitly pass as system library: https://github.com/brainboxdotcc/DPP/blob/v10.0.34/library/CMakeLists.txt#L75
- The definition DPP_USE_EXTERNAL_JSON is exposed in the public header json_fwd.h as well. Need to forward the definition in order to consume nlohmann::json from Conan.

uilianries (commented):
Almost good to go.

recipes/dpp/all/conanfile.py
AbrilRBS: Seems to be the easiest way to fix https://github.com/conan-io/conan-center-index/pull/25745#issuecomment-2445378823

uilianries: Thank you!

recipes/dpp/all/conanfile.py
braindigitalis: dpp automatically adds threads in its own cmakelists, we do compile on freebsd without any hacks or patches needed, is the extra stuff for that really necessary here?

braindigitalis: also what about the other bsd variants such as openbsd?

braindigitalis:
I've bought through your recommendations @uilianries however I left the pr with ci running last night waiting for it to complete and this AM it still hasn't finished, how long should the CI take to run, was this just an anomaly? Thanks!

braindigitalis:
this now fails because it does not publicly bring in nlohmann json. any ideas?

recipes/dpp/all/conanfile.py
```diff
@@ -0,0 +1,95 @@
...
+        if self.settings.compiler.get_safe("cppstd"):
+            check_min_cppstd(self, self._min_cppstd)
+        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
+        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
+            raise ConanInvalidConfiguration(
+                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
+            )
+
+    def requirements(self):
+        self.requires("nlohmann_json/3.11.2")
```
uilianries, line 45:
```suggestion
        self.requires("nlohmann_json/3.11.2", transitive_libs=True, transitive_headers=True)
```

nlohmann_json is exposed in public headers. Conan 2.x makes all dependencies as private by default. Need to tell Conan to expose nlohmann_json headers and make any definition as public to the downstream

recipes/dpp/all/conanfile.py
uilianries: > really necessary here?

No. But because this project is shared library only, it carries the metadata pointing system libraries, including pthread. It would be needed in case building as static library.

uilianries (approved):
LGTM. The recipe looks good to be merged now. Thank you for your PR! 

To understand the review process better, please, read https://github.com/conan-io/conan-center-index/blob/master/docs/review_process.md#getting-your-pull-request-reviewed

In summary, need to pass by the CI and have +2 approves.

AbrilRBS (approved):

uilianries:
@braindigitalis Your PR has been merged and is now available in Conan Center:

```
$ conan search -r conancenter dpp 
Found 1 pkg/version recipes matching dpp in conancenter
conancenter
  dpp
    dpp/10.0.34
```

The https://conan.io/center updates the data base using a scheduled job, so the package should be listed there after some hours. Regards.

PR #25771, opened 2024-10-31
add glad version 2.x

planetmarshall:
### Summary
Changes to recipe:  **glad/2.x**

#### Motivation

Add [glad](https://github.com/Dav1dde/glad) v2.

#### Details

Adds glad v2.

Moves Glad v 0.x to subfolder as recipes are significantly different.

---


recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,138 @@
...
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/Dav1dde/glad"
+    license = "MIT"
+    settings = "os", "compiler", "build_type", "arch"
+
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "no_loader": [True, False],
+        "spec": ["gl", "egl", "glx", "wgl"], # Not relevant for 2.x. A spec will be included unless its version is None
```
planetmarshall, line 23: I've left this in because it's in the `0.x` recipe - but specs aren't mutually exclusive - in fact a GLES spec won't work without also including egl.

Would be good if we can just remove this option?

jay-tux:
@planetmarshall Thanks for packaging this already! However, will this one have support for disabling the debug mode? The current (0.1.36) links gl* functions to glad_debug_gl* instead of glad_gl*. It would be nice if I could disable this behavior (as I have a custom error handler that's a little more useful)

planetmarshall:
> @planetmarshall Thanks for packaging this already! However, will this one have support for disabling the debug mode? The current (0.1.36) links gl* functions to glad_debug_gl* instead of glad_gl*. It would be nice if I could disable this behavior (as I have a custom error handler that's a little more useful)

I've disabled the debugging layer by default (as it's normally opt-in anyway).

It can be enabled by specifying the option `-o glad/*:debug=True`

rmatab:
@ErniGH Could you please review this? :pray:

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+        "debug": False,
+        "multicontext": False,
+        "gl_profile": "compatibility",
+        "gl_version": "3.3",
+        "gles1_version": "None",
+        "gles2_version": "None",
+        "glsc2_version": "None",
+        "egl_version": "None",
+        "glx_version": "None",
+        "wgl_version": "None"
```
ErniGH, line 57:
```suggestion
        "gles1_version": None,
        "gles2_version": None,
        "glsc2_version": None,
        "egl_version": None,
        "glx_version": None,
        "wgl_version": None
```

This was a limitation from conan1

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+                       "4.3", "4.4", "4.5", "4.6"],
+        "gles1_version": ["None", "1.0"],
+        "gles2_version": ["None", "2.0", "3.0", "3.1", "3.2"],
+        "glsc2_version": ["None", "2.0"],
+        # if specification is egl
+        "egl_version": ["None", "1.0", "1.1", "1.2", "1.3", "1.4", "1.5"],
+        # if specification is glx
+        "glx_version": ["None", "1.0", "1.1", "1.2", "1.3", "1.4"],
+        # if specification is wgl
+        "wgl_version": ["None", "1.0"]
```
ErniGH, line 39: Change also all this `"None"` to `None`

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+        "wgl_version": "None"
+    }
+
+    def export_sources(self):
+        copy(self, "CMakeLists.txt", self.recipe_folder, self.export_sources_folder)
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
+            del self.options.wgl_version
```
ErniGH, line 66: Why is this option not compatible with Windows?

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+    
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables.update({
+                "Python_EXECUTABLE": sys.executable,
```
ErniGH, line 88: This line is fragile If we need `Python` in the recipe, we can use `tools.cmake.cmaketoolchain:extra_variables` in your profile pointing to your own python.

ErniGH (commented):
Apologies for the delay. Unfortunately, this PR slipped through the cracks, and due to its complexity, it took me longer than expected to review.

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,147 @@
...
+    license = "MIT"
+    settings = "os", "compiler", "build_type", "arch"
+
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "no_loader": [True, False],
+        "spec": ["gl", "egl", "glx", "wgl"], # Not relevant for 2.x. A spec will be included unless its version is None
+        "extensions": ["ANY"], # A comma separated list of extensions, if missing all extensions are included
+        "debug": [True, False], # Enable the additional GLAD Debugging layer
```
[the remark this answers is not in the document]
planetmarshall: That is how I'd originally implemented it - but see comment by @jay-tux  [above](https://github.com/conan-io/conan-center-index/pull/25771#issuecomment-2508976747). The debug option is an additional diagnostic layer so it's a bit more than just an unoptimized build with symbols which is what most Debug builds are.

I'll have a look through the index and see if there's anything similar. The GLAD debug build is described in detail [here](https://github.com/Dav1dde/glad/wiki/C#debugging)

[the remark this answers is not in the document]
jay-tux: @ErniGH some additional context regarding GLAD's debug "mode" - OpenGL by default returns status code (in the same vein as traditional C functions set `errno`). The debug mode adds a wrapper around each call, which checks said return value and prints something on errors.

But I could see where the naming might be confusing. A good middle ground could be the following:
- We rename the **option** to something like `debug_layer` 
- In `Debug` mode, the `debug_layer` is enabled by default (but can be disabled)
- In `Release` mode, the `debug_layer` is disabled by default (but can be enabled)

I feel like this might be a solution that offers flexibility, doesn't confuse, and results in predictable/expected behavior.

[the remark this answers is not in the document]
planetmarshall: @ErniGH could we do that in `config_options`? If we did

```
def config_options(self):
  self.options.debug_wrappers = self.settings.build_type == "Debug"
```

can the user still override that setting downstream?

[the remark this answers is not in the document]
ErniGH: I understand now. Yes, I think renaming the option will be better for future reviews. Thanks for the explanation 😄

recipes/glad/2.x/conanfile.py
planetmarshall: Good spot - think this was an error on my part. The condition should be inverted.

recipes/glad/2.x/conanfile.py
planetmarshall: I think I might have done this just for the CCI build - currently CMake `FindPython` finds quite an old version of Python on macos.

recipes/glad/2.x/test_package/test_package.c
```diff
@@ -0,0 +1,24 @@
...
+int main() {
+    GLFWwindow* window;
+
+    #ifdef __APPLE__
+    if (!glfwInit()) {
+        return -1;
+    }
+    #endif
+
+    window = glfwCreateWindow(1, 1, "Hello World", NULL, NULL);
```
jcar87: Please do not initialise UI, or interact with the filesystem or network in any test package, thanks!

ErniGH (dismissed):
LGTM Thanks for your effort

ErniGH:
Sorry, in my previous review, I asked to change the value options value from 'None' to None, thinking the issue was fixed in Conan2. But after testing it, I realized it still needs to be treated as a string to disable the 'gl_version' option.

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+            "glsc2": self.options.glsc2_version,
+            "egl": self.options.egl_version,
+            "glx": self.options.glx_version,
+        }
+
+        wgl_version = self.options.get_safe("wgl_version")
+        if wgl_version:
+            spec_api["wgl"] = wgl_version
+
+        return ";".join(api(spec_api))
```
franramirez688: I think that it's easier to read:

```suggestion
def _get_api(self):
    spec_api = {
        "gl": self.options.gl_version,
        "gles1": self.options.gles1_version,
        "gles2": self.options.gles2_version,
        "glsc2": self.options.glsc2_version,
        "egl": self.options.egl_version,
        "glx": self.options.glx_version,
        "wgl": self.options.get_safe("wgl_version", "None")
    }
    return ";".join(
        f"{name}:{self.options.gl_profile}={version}" if name == "gl" else f"{name}={version}"
        for name, version in spec_api.items()
        if version != "None"
    )
```

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/Dav1dde/glad"
+    license = "MIT"
+    settings = "os", "compiler", "build_type", "arch"
+
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "no_loader": [True, False],
+        "extensions": [None, "ANY"],
```
franramirez688: We already have in the older versions:

```python
"extensions": ["ANY"],

# ...

"extensions": "",
```
Is it not better that way?

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,158 @@
...
+            spec_api["wgl"] = wgl_version
+
+        return ";".join(api(spec_api))
+
+    def package(self):
+        CMake(self).install()
+        copy(self, "LICENSE", self.source_folder, os.path.join(self.package_folder, "licenses"))
+
+    def package_info(self):
+        self.cpp_info.libs = collect_libs(self)
```
franramirez688: Are the library names created differently from this?
```suggestion
        self.cpp_info.libs = ["glad"]
```

franramirez688 (commented):
Thanks for this PR @planetmarshall 

Looking great! Several comments to be discussed 🙏

ErniGH:
All the changes from the review have been included. It seems that `collect_libs(self)` always generates `["glad"]`. I’ve tested it with different options and on different systems, and it always produces the same result.

I’ve also tested all the different options locally. Here are the logs:
[glad2-debug_layer=True.log](https://github.com/user-attachments/files/22208371/glad2-debug_layer.True.log)
[glad2-egl_version.log](https://github.com/user-attachments/files/22208373/glad2-egl_version.log)
[glad2-gl_profile=core.log](https://github.com/user-attachments/files/22208374/glad2-gl_profile.core.log)
[glad2-gles1_version.log](https://github.com/user-attachments/files/22208375/glad2-gles1_version.log)
[glad2-gles2_version.log](https://github.com/user-attachments/files/22208376/glad2-gles2_version.log)
[glad2-glsc2_version.log](https://github.com/user-attachments/files/22208377/glad2-glsc2_version.log)
[glad2-glx_version.log](https://github.com/user-attachments/files/22208378/glad2-glx_version.log)
[glad2-multicontext=True.log](https://github.com/user-attachments/files/22208379/glad2-multicontext.True.log)
[glad2-no_loader=True.log](https://github.com/user-attachments/files/22208380/glad2-no_loader.True.log)
[glad2-wgl_version(windows).log](https://github.com/user-attachments/files/22208381/glad2-wgl_version.windows.log)

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,149 @@
+import os
+import sys
+from pathlib import Path
+
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import collect_libs, copy, get
```
franramirez688:
```suggestion
from conan.tools.files import copy, get
```

recipes/glad/2.x/conanfile.py
```diff
@@ -0,0 +1,149 @@
...
+from conan.errors import ConanInvalidConfiguration
+
+class GladConan(ConanFile):
+    name = "glad"
+    description = "Multi-Language GL/GLES/EGL/GLX/WGL Loader-Generator based on the official specs."
+    topics = ("opengl",)
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/Dav1dde/glad"
+    license = "MIT"
+    settings = "os", "compiler", "build_type", "arch"
```
franramirez688, line 17:
```suggestion
    settings = "os", "compiler", "build_type", "arch"
    package_type = "library"
```

franramirez688 (approved):

Issue #22989, opened 2024-03-05
[package] poco/1.13.1: Cannot link with -static-libgcc/-static-libstdc++


stygianstein:
### Description

Simple test application created using Poco, linking does not respect the "-static-libgcc, -static-libstdc++" flags. CMake links using "Poco::Poco" to link across all libraries, but results in libgcc and libstdc++ being dynamically linked.
Disabling Poco::DataMySQL works as expected, so this is either caused by Poco's handling of libmysqlclient or by libmysqlclient itself, the version included by Poco is 8.1.0

### Package and Environment Details

* Package Name/Version: **poco/1.13.1**
* Operating System+version: **Linux Debian 12**
* Compiler+version: **GCC 13**
* Docker image: **gcc:13.2**
* Conan version: **conan 2.0.17**
* Python version: **Python 3.11.2**

### Conan profile

[settings]
os=Linux
arch=x86_64
build_type=Release
compiler=gcc
compiler.version=13
compiler.cppstd=17
compiler.libcxx=libstdc++11

### Steps to reproduce

conan install --requires poco/1.13.1 -pr:h x64-gcc-release -pr:b x64-gcc-release -b "*"
[files.zip](https://github.com/conan-io/conan-center-index/files/14498620/files.zip)

### Logs

<details><summary>Click to expand log</summary>

```
ldd out/build/x64-gcc-release/PocoApp
        linux-vdso.so.1 (0x00007fffb30d5000)
        libstdc++.so.6 => /usr/local/lib64/libstdc++.so.6 (0x00007f649e976000)
        libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f649e965000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f649e886000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f649e6a3000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f649ebdd000)
        libgcc_s.so.1 => /usr/local/lib64/libgcc_s.so.1 (0x00007f649e67e000)
```

</details>

SpaceIm:
It's a side effect of https://github.com/conan-io/conan-center-index/blob/c4308b493f37390333773d651c80b015b714ae69/recipes/libmysqlclient/all/conanfile.py#L275-L278

It can't be fixed in recipe, it's conan issue. See https://github.com/conan-io/conan/issues/15646

Issue #22620, opened 2024-02-01
[package] flann/1.9.2: static runtime is being overriden in package_info


Jimbopython:
### Description

When building the flann library via conan, it does not respect setting a static runtime, when building it as a static library. 

In the generated cmake finder module, the libc++ is mentioned as a dependency:

```
cat FindFlann.cmake | grep c++
set(Flann_SYSTEM_LIBS m c++)
set(Flann_flann_cpp_s_SYSTEM_LIBS c++)
```

This might come from the conanfile of flann:
```
if not self.options.shared:
    libcxx = stdcpp_library(self)
    if libcxx:
        self.cpp_info.components["flann_cpp"].system_libs.append(libcxx)
```
It may be changed to something like:
```
if not self.options.shared and self.options.compiler.get_safe("runtime") == "shared":
    libcxx = stdcpp_library(self)
    if libcxx:
        self.cpp_info.components["flann_cpp"].system_libs.append(libcxx)
```

### Package and Environment Details

* Package Name/Version: **flann/1.9.2**
* Operating System+version: **Arch Linux**
* Compiler+version: **Clang 16**
* Conan version: **conan 1.62.0**
* Python version: **Python 3.11.6**

### Conan profile

[settings]
os=Linux
arch=x86_64
compiler=clang
compiler.version=16
compiler.libcxx=libc++
compiler.runtime=static
compiler.cppstd=20

[env]
CC=/usr/bin/clang
CXX=/usr/bin/clang++

### Steps to reproduce

conan install flann/1.9.2@ --install-folder build --build -pr:b clang_test -pr:h clang_test -s:h build_type=Release -g cmake_find_package -o shared=False

### Logs

<details><summary>Click to expand log</summary>

```
Conan 1 is on a deprecation path, please consider migrating to Conan 2
Configuration (profile_host):
[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.cppstd=20
compiler.libcxx=libc++
compiler.runtime=static
compiler.version=16
os=Linux
[options]
shared=False
[build_requires]
[env]
CC=/usr/bin/clang
CXX=/usr/bin/clang++
Configuration (profile_build):
[settings]
arch=x86_64
compiler=clang
compiler.cppstd=20
compiler.libcxx=libc++
compiler.runtime=static
compiler.version=16
os=Linux
[options]
[build_requires]
[env]
CC=/usr/bin/clang
CXX=/usr/bin/clang++
[...]

PR #16236, opened 2023-02-22
snappy: do not explicitly link the C++ standard library

thejohnfreeman:
Explicitly linking the standard library makes it impossible (without manually editing Conan-generated modules) for downstream builders to statically link the standard library when using GCC. The GCC option `-lstdc++` overrides the option `-static-libstdc++`, even if `-static-libstdc++` appears after.

Specify library name and version:  **lib/1.0**

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #25779, opened 2024-10-31
libvpx: fix redundant -lc++

canmor:
### Summary
Changes to recipe:  **libvpx**

- fix redundant -lc++

#### Motivation
Consuming libvpx on macOS keep receiving warnings like:

```
ld: warning: ignoring duplicate libraries: '-lc++'
```

#### Details
I guess C++ runtime library would be specified by conan in `-stdlib` flag (such as -stdlib=libc++ for Clang or -stdlib=libstdc++11 for GCC), adding the standard library again with self.cpp_info.system_libs.append(libcxx) result in warnings:

```
ld: warning: ignoring duplicate libraries: '-lc++'
```

Which is complained from clang on macOS, generated verbose link command is like:

```shell
/<path-to-toochain>/c++ -stdlib=libc++ -g -arch arm64 -isysroot /<path-to-xcode-sdk>/MacOSX15.1.sdk -Wl,-search_paths_first -Wl,-headerpad_max_install_names CMakeFiles/example.dir/src/example.cpp.o -o example   -L/<path-to-conan>/p/b/hello31cb6e20b6eb3/p/lib  -L/<path-to-conan>/p/b/libvp0331582145844/p/lib  -Wl,-rpath,/<path-to-conan>/p/b/hello31cb6e20b6eb3/p/lib -Wl,-rpath,/<path-to-conan>/p/b/libvp0331582145844/p/lib /<path-to-conan>/p/b/hello31cb6e20b6eb3/p/lib/libhello.a /<path-to-conan>/p/b/libvp0331582145844/p/lib/libvpx.a -lc++
```

---

### DIFF
--- a/recipes/libvpx/all/conanfile.py
+++ b/recipes/libvpx/all/conanfile.py
@@ -266,9 +266,5 @@ class LibVPXConan(ConanFile):
     def package_info(self):
         self.cpp_info.set_property("pkg_config_name", "vpx")
         self.cpp_info.libs = [self._lib_name]
-        if not self.options.get_safe("shared"):
-            libcxx = stdcpp_library(self)
-            if libcxx:
-                self.cpp_info.system_libs.append(libcxx)
         if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.system_libs.extend(["m", "pthread"])


jcar87:
Hi @canmor, thank you for opening this PR.

These changes would actually re-introduce the problem that this is trying to fix: when a *C* only consumer tries to link against a static library that requires the C++ runtime, it would break.

You would see that warning if you have a C++ consumer - I can see in your log that you are using `/<path-to-toochain>/c++` - the c++ compiler implicitly instructs the linker to link agianst libc++ - that's why it is redundant in your case from a C++ consumer perspective. But a C consumer will behave differently, especially on Linux, and thus removing that check would actually break the recipe in those circumstances.

canmor:
@jcar87 Appreciate for you fast response and valuable input, I guess we can figure out a solution for C++ consumer without affecting C consumer, how about checking'compiler.libcxx' to make this behavior C exclusive?

I didn't get my device on hand currently, I can give it a try later.

Plus, do you have any idea to improve the 'test_package' for preventing a regression like this?

canmor:
Sadly this is not the way `package_info()` works, it won't evaluate during consuming. I can’t find a way to determine whether the consumer is pure C or C++.

And, after some searches on GitHub, I found many PR/issue out there related to kind of this issue, some of them fall into the troubles of linking with the static C++ runtime library (since dependent recipe which declares libc++ like this leads to dynamic C++ runtime):

https://github.com/conan-io/conan-center-index/pull/16236
https://github.com/conan-io/conan-center-index/issues/22989
https://github.com/conan-io/conan-center-index/issues/22620

I have not idea how to fix this in recipe, I'm going to close this PR, but I will be glad to hear any further advice or news about this.

Thanks for all your help.

jcar87:
Thanks @canmor  - we are currently working on improvements on the `package_info()` model to be able to record the "linking language" - such that, at least CMake consumers can do the right thing :

https://cmake.org/cmake/help/v3.31/prop_tgt/LINKER_LANGUAGE.html

So I believe once this is rolled out and implemented in the recipe, for CMake consumers the problems will be addressed: no redundant linking of the C++ runtime, and ability for consumers to decide whether to use the static runtime.

This is however, still a bigger issue that affects not only Conan. I suspect upstream pkg-config files (.pc) for "libraries with a pure C interface that make calls to C++ functionality" already may mark the C++ runtime in the `libs.private` field of the .pc fields - and would exhibit the same behaviour on the consumer side

PR #25786, opened 2024-11-01
autoconf: Do not clear "os" from package info

cubanpit:
Avoids compatibility issues between Windows and Linux.

### Summary
Changes to recipe:  **autoconf**

#### Motivation
Not including `os` information in the package means that packages built on Linux and Windows mix up, with same package ID and different package revisions. The most recent one will then be pulled from the remote and used, but they might actually not be fully compatible between OSes. In our case the Windows-built version did not have the executable flag set on the scripts in `bin/`, this caused issues when using the package in a Linux environment.
The symptoms were similar to https://github.com/conan-io/conan/issues/14345, so we initially thought the problem happened in decompressing the package files.

#### Details
Just keep the `os` information, it causes minimal duplication and it makes sure the package works in all environments.

---

### DIFF
--- a/recipes/autoconf/all/conanfile.py
+++ b/recipes/autoconf/all/conanfile.py
@@ -38,7 +38,9 @@ class AutoconfConan(ConanFile):
         self.requires("m4/1.4.19") # Needed at runtime by downstream clients as well
 
     def package_id(self):
-        self.info.clear()
+        del self.info.settings.arch
+        del self.info.settings.compiler
+        del self.info.settings.build_type
 
     def build_requirements(self):
         self.tool_requires("m4/1.4.19")
--- a/recipes/autoconf/all/conanfile.py
+++ b/recipes/autoconf/all/conanfile.py
@@ -41,6 +41,7 @@ class AutoconfConan(ConanFile):
         del self.info.settings.arch
         del self.info.settings.compiler
         del self.info.settings.build_type
+        self.info.requires.clear()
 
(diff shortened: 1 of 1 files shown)


uilianries (approved):
Good catch!

recipes/autoconf/all/conanfile.py
```diff
@@ -38,7 +38,9 @@ def requirements(self):
         self.requires("m4/1.4.19") # Needed at runtime by downstream clients as well
 
     def package_id(self):
-        self.info.clear()
+        del self.info.settings.arch
```
memsharded, line 41: The ``self.info.clear()`` also clears:
- Possible variability from ``options``, not an issue in this case
- Possible variability from ``dependencies``. In this case, it could be a problem as ``self.requires("m4/1.4.19")`` is a requirement, and changing that version will create another ``package_id``. Most like the ``self.info.requires.clear()`` is necessary too

cubanpit: I understand and I added this change to the branch.
I would imagine that a new version of m4 can change the output, but then the package revision would change as well.
A change in the version of m4 would also mean a change in the recipe revision though, so a new package would be generated anyway, does this actually avoid duplication?

cubanpit:
I think this has been ready for merging for a long time now. Anyone willing to review and merge?

memsharded:
Thanks for the ping @cubanpit, merged!

PR #25795, opened 2024-11-02
xxhash: Add header_only option

### LINKED ISSUE #25778: [package] xxhash/0.8.2: missing options
### Description

The recipe is missing options for `XXH_STATIC_LINKING_ONLY` and `XXH_IMPLEMENTATION` (header only).

* Package Name/Version: **xxhash/0.8.2**

### Conan profile

N/A

### Steps to reproduce

N/A

### Logs

N/A

vasama:
### Summary
Changes to recipe:  **xxhash/0.8.2**

#### Motivation
* `XXH_STATIC_LINKING_ONLY` enables static allocation of hash states. This is important for performance reasons.
* The header-only mode allows inlining the hashing functions in the caller. This is important for performance reasons.

Closes #25778.

#### Details
* Add `header_only` option (default: false), define `XXH_INLINE_ALL` in consumers in header-only mode.
* Define `XXH_STATIC_LINKING_ONLY` in consumers when the package is not a shared library.
* Improve the test package to test linking in header-only mode.

---

### DIFF
--- a/recipes/xxhash/all/conanfile.py
+++ b/recipes/xxhash/all/conanfile.py
@@ -1,4 +1,5 @@
 from conan import ConanFile
+from conan.errors import ConanException
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
 import os
@@ -17,11 +18,13 @@ class XxHashConan(ConanFile):
     settings = "os", "arch", "compiler", "build_type"
     options = {
         "shared": [True, False],
+        "header_only": [True, False],
         "fPIC": [True, False],
         "utility": [True, False],
     }
     default_options = {
         "shared": False,
+        "header_only": False,
         "fPIC": True,
         "utility": True,
     }
@@ -34,8 +37,14 @@ class XxHashConan(ConanFile):
             del self.options.fPIC
 
     def configure(self):
-        if self.options.shared:
+        if self.options.shared and self.options.header_only:
+            raise ConanException("Options 'shared' and 'header_only' cannot both be true at once.")
+        if self.options.shared or self.options.header_only:
             self.options.rm_safe("fPIC")
+        if self.options.shared:
+            self.options.rm_safe("header_only")
+        elif self.options.header_only:
+            self.options.rm_safe("shared")
         self.settings.rm_safe("compiler.cppstd")
         self.settings.rm_safe("compiler.libcxx")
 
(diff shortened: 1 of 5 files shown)


uilianries (changes requested):
Thank you for your suggestion.

I understand your point of view about performance benefit, but it's a feature/build to be discussed with the upstream first.

Conan tries to keep as closer as possible to the upstream implementation, and there is no such header-only option supported right now in the upstream. 

The [XXH_STATIC_LINKING_ONLY](https://github.com/Cyan4973/xxHash/blob/v0.8.2/xxhash.c#L40) is always defined as well.

The [Makefile](https://github.com/Cyan4973/xxHash/blob/v0.8.2/Makefile) build both static and shared library, but does not offer a header-only option.

The unofficial [CMakeLists.txt](https://github.com/Cyan4973/xxHash/blob/v0.8.2/cmake_unofficial/CMakeLists.txt) has support for both static and shared library, but not for header-only.

Plus, checking other package managers, I did not find any of them offering such approach, so I would prefer not integrating this change to the current recipe as we will be maintaining a non-official build.

Please, discuss with the upstream, trying to integrate a header-only option there first.

vasama:
> The [XXH_STATIC_LINKING_ONLY](https://github.com/Cyan4973/xxHash/blob/v0.8.2/xxhash.c#L40) is always defined as well.

It needs to be defined for the consumers as well. When this macro is not defined, the internal types are not defined in the header. The implementation defines it because it needs access to those types. Consumers need access to those types in order to allocate them statically. As the naming suggest, this macro should only be defined when linking statically, which is what this PR does.

> Please, discuss with the upstream, trying to integrate a header-only option there first.

Header-only usage is something explicitly supported by upstream. See [here](https://github.com/Cyan4973/xxHash/blob/dev/xxhash.h#L284). It's presumably not part of the build scripts because header-only usage doesn't actually require any build steps. So to be clear, before accepting this, upstream needs a CMake option that builds nothing and installs just the header?

vasama:
> So to be clear, before accepting this, upstream needs a CMake option that builds nothing and installs just the header?

@uilianries are you able to clarify this?

uilianries:
> Header-only usage is something explicitly supported by upstream. See [here](https://github.com/Cyan4973/xxHash/blob/dev/xxhash.h?rgh-link-date=2024-11-06T10%3A03%3A33Z#L284). It's presumably not part of the build scripts because header-only usage doesn't actually require any build steps.

Currently, the documentation is quite limited, with only a single line in the README and a blog post that references an "obscure build flag." This lack of clarity makes it challenging for users to understand and utilize the feature effectively. Additionally, I’ve noticed that no other package manager exposes a similar feature, which raises questions about its necessity and usability within ConanCenterIndex ecosystem.

While we won’t be able to accept this feature at this time, I encourage you to keep it in your fork. You can continue to develop and use it there. If you encounter issues with storing pre-built packages, consider using the [local-recipes-index-remote](https://blog.conan.io/2024/04/23/Introducing-local-recipes-index-remote.html) feature to fetch your own GitHub repository when consuming your recipe, so you can consume your own Conan recipe and keep the header-only feature as well.

Thank you again for your feature suggestion! I appreciate the effort you put into developing this new feature. However, due my presented concerns, I'll be closing this PR at this time. Regards.

Issue #25797, opened 2024-11-02
[package] box2d/3.0.0: error: loop not vectorized


skhaz:
### Description

I am having a problem only when compiling box2d with `emsdk`

```
/home/runner/.conan2/p/b/box2d5931abb0d273d/b/src/src/contact_solver.c:865:6: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
...
Several similar erros
```

How can I override the `conanfile.txt` to bypass this warning?

### Package and Environment Details

* Package Name/Version: **zlib/1.2.8**
* Operating System+version: **Linux Ubuntu 18.04**
* Compiler+version: **GCC 8**
* Docker image: **conanio/gcc8**
* Conan version: **conan 1.18.0**
* Python version: **Python 3.7.4**

### Conan profile

include(default)

[settings]
arch=wasm
os=Emscripten

[tool_requires]
*: emsdk/3.1.44

### Steps to reproduce

It is complex to reproduce, see the logs here https://github.com/carimbolabs/carimbo/actions/runs/11643393796/job/32424090261

### Logs

<details><summary>Click to expand log</summary>

```
conan install . --output-folder=build --build=missing --profile=webassembly --settings compiler.cppstd=17 --settings build_type=Release

======== Input profiles ========
Profile host:
[settings]
arch=wasm
build_type=Release
compiler=gcc
compiler.cppstd=17
compiler.libcxx=libstdc++11
compiler.version=11
os=Emscripten
[tool_requires]
*: emsdk/3.1.44

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux


======== Computing dependency graph ========
box2d/3.0.0: Not found in local cache, looking in remotes...
box2d/3.0.0: Checking remote: conancenter
box2d/3.0.0: Downloaded recipe revision 9c01bcc59d7d7c8d19715e2adb0a510b
simde/0.8.2: Not found in local cache, looking in remotes...
simde/0.8.2: Checking remote: conancenter
simde/0.8.2: Downloaded recipe revision 5e1edfd5cba92f25d79bf6ef4616b972
pkgconf/2.0.3: Not found in local cache, looking in remotes...
pkgconf/2.0.3: Checking remote: conancenter
[...]

PR #25815, opened 2024-11-03
box2d: Do not force warnings as errors by default

skhaz:
### Summary
Changes to recipe:  **box2d/3.0.0**

#### Motivation
This (hopefully) closes https://github.com/conan-io/conan-center-index/issues/25797

#### Details
Box2D compiles on Linux, Windows, and macOS, but not for WebAssembly using emsdk. This PR attempts to address the compilation error I’m encountering and should not impact other users.

---

### DIFF
--- a/recipes/box2d/all/conanfile.py
+++ b/recipes/box2d/all/conanfile.py
@@ -78,7 +78,7 @@ class Box2dConan(ConanFile):
     def generate(self):
         tc = CMakeToolchain(self)
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
-	tc.cache_variables["CMAKE_COMPILE_WARNING_AS_ERROR"] = False
+        tc.cache_variables["CMAKE_COMPILE_WARNING_AS_ERROR"] = False
         if Version(self.version) < "3.0.0":
             tc.variables["BOX2D_BUILD_TESTBED"] = False
             tc.variables["BOX2D_BUILD_UNIT_TESTS"] = False
--- a/recipes/box2d/all/conanfile.py
+++ b/recipes/box2d/all/conanfile.py
@@ -78,6 +78,7 @@ class Box2dConan(ConanFile):
     def generate(self):
         tc = CMakeToolchain(self)
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
+        tc.variables["CMAKE_CXX_FLAGS"] = "-O3 -ftree-vectorize -fopt-info-vec -Wno-pass-failed"
         if Version(self.version) < "3.0.0":
             tc.variables["BOX2D_BUILD_TESTBED"] = False
             tc.variables["BOX2D_BUILD_UNIT_TESTS"] = False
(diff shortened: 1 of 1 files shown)


skhaz:
Please help @conan-io/barbarians

skhaz:
Please help @memsharded

recipes/box2d/all/conanfile.py
```diff
@@ -78,6 +78,7 @@ def source(self):
     def generate(self):
         tc = CMakeToolchain(self)
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
+        tc.variables["CMAKE_CXX_FLAGS"] = "-O3 -ftree-vectorize -fopt-info-vec -Wno-pass-failed"
```
jcar87, line 81: Hi @skhaz - thanks for your contribution.

Doing it this way does not take into account the compiler - they are passed unconditionally to all compilers regardless of whether they understand those particular flags. Additionally, `-O3` would trample and override the user's build_type setting - since it is being applied unconditionally.

This looks like a bug in the upstream recipe itself - I would first ensure that it is, at the very least, reported upstream to the maintainers: https://github.com/erincatto/box2d

In this instance, however I would possibly advise simply setting 
```
tc.cache_variables["CMAKE_COMPILE_WARNING_AS_ERROR"] = False
```

since this is failing on emscriptem because this setting is too strict

jcar87 (changes requested):
see comment

recipes/box2d/all/conanfile.py
skhaz: Thank you, I have updated my pull request and raised an issue on box2d repository.

skhaz:
Thank you @AbrilRBS. I will keep an eye and when they provide a fix I will let you know.

jcar87:
Thanks @skhaz ! Running this on CI - please let us know if with that setting, it at least lets you proceed :D

skhaz:
Is there a way to test this branch locally, I mean, instead of conan pulling the packages specs from the main repository/internet, fetch from my branch?

jcar87 (approved):

lindequist:
First of all, sorry if I commenting in the wrong place. I was also trying to use box2d/3.0.0 with emsdk/3.1.50, and I also got errors a la:
```
.conan2/p/b/box2d6487a51d42777/b/src/src/contact_solver.c:810:6: error: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Werror,-Wpass-failed=transform-warning]
  810 | void b2WarmStartContactsTask( int startIndex, int endIndex, b2StepContext* context, int colorIndex )
```
I tried with the changes merged into master here, but it didn't fix the issue for me. Then, just for testing, I checked out conan-center-index and added this line instead in `recipes/box2d/all/conanfile.py`:
```
tc.cache_variables["CMAKE_C_FLAGS"] = "-Wno-error=pass-failed"
```
After exporting using `conan export . --version=3.0.0`, I was able to build the box2d dependency, and the warning that was treated as an error before is now just a warning.

skhaz:
I simply give up on box2d, I am using chipmunk.

skhaz:
https://github.com/erincatto/box2d/issues/832#issuecomment-2466603695

PR #25827, opened 2024-11-04
Add history.h as a symlink to readline.h to align with libreadline

rbu9fe:
### Summary
Changes to recipe:  **editline/3.1**

#### Motivation
`editline` is a readline-compatible library, whereas `libreadline` exports its API in two header files: `readline.h` and `history.h`. Some consumers support using any of these readline-compatible libraries, which only works when they expose the same functions as well as headers. Since `editline` only provides a single header `readline.h` the Conan package should provide a symbolic link `history.h` linking to `readline.h` so that consumers find the expected header.

Note that the [`editline` Debian package](https://packages.debian.org/de/bookworm/libedit-dev) does the same thing [here](https://salsa.debian.org/debian/libedit/-/blob/master/debian/libedit-dev.links?ref_type=heads).

#### Details
Add a symbolic link history.h linking to (and next to) readline.h.

---
- [x ] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)

### DIFF
--- a/recipes/editline/all/conanfile.py
+++ b/recipes/editline/all/conanfile.py
@@ -3,7 +3,7 @@ from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import fix_apple_shared_install_name
 from conan.tools.build import cross_building
 from conan.tools.env import VirtualRunEnv
-from conan.tools.files import copy, get, rm, rmdir
+from conan.tools.files import copy, get, rm, rmdir, chdir
 from conan.tools.gnu import Autotools, AutotoolsDeps, AutotoolsToolchain
 from conan.tools.layout import basic_layout
 import os
@@ -83,6 +83,10 @@ class EditlineConan(ConanFile):
         rmdir(self, os.path.join(self.package_folder, "share"))
         rm(self, "*.la", os.path.join(self.package_folder, "lib"))
         fix_apple_shared_install_name(self)
+
+        # Create symlink history.h as consumers may expect to find some functions therein instead of editline.h
+        with chdir(self, os.path.join(self.package_folder, "include", "editline")):
+            os.symlink("readline.h", "history.h")
 
     def package_info(self):
         self.cpp_info.set_property("pkg_config_name", "libedit")


uilianries:
@rbu9fe Hello! :wave: 

Thank you for your suggestion of customizing this Conan recipe :smile: 

>  Some consumers support using any of these readline-compatible libraries, which only works when they expose the same functions as well as headers. 

Do you have a real example of a project only consuming history.h and are blocked without this PR?

-----

I did a quick search and I only found Debian and Ubuntu using this symlink:

- Debian, you mentioned the link in your description above
- Ubuntu https://git.launchpad.net/ubuntu/+source/libedit/tree/debian/libedit-dev.links?h=ubuntu/jammy

Other Linux distros are only building without such trick:

- Arch Linux (BTW I use Arch): https://gitlab.archlinux.org/archlinux/packaging/packages/editline/-/blob/main/PKGBUILD?ref_type=heads
- Fedora: https://src.fedoraproject.org/rpms/editline/blob/rawhide/f/editline.spec
- OpenSuse: https://build.opensuse.org/projects/openSUSE:Factory/packages/libedit/files/libedit.spec?expand=1
- Gentoo: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-libs/editline/editline-1.17.1-r1.ebuild
- FreeBSD (not linux): https://freebsd.pkgs.org/14/freebsd-amd64/editline-1.17.1.pkg.html#files

And other multiproposal package managers are not creating that symlink:

- Homebrew: https://github.com/Homebrew/homebrew-core/blob/e2422d347823dad357062c503f5c63a72532d1cd/Formula/lib/libedit.rb
- Conda Forge: https://github.com/conda-forge/libedit-feedstock/blob/main/recipe/build.sh

----

I understand your concern about aligning that header according to libreadline, but I'm afraid this is a particular case that may risk breaking general users, than solving some real issue. Regards!

rbu9fe:
I'm using [lldpd ](https://github.com/lldpd/lldpd) together with editline, which doesn't work when editline doesn't provide history.h.
How can adding a symbolic link risk breaking other users? Since it's an additive change only I thought it's not causing side effects.

uilianries:
@rbu9fe Hello again!

>  How can adding a symbolic link risk breaking other users? Since it's an additive change only I thought it's not causing side effects.

That change is adding a header that was not expected before in the package and is not part of the default build of libedit. As ConanCenterIndex tries to keep as closer as possible to the upstream project and there is no large evidence of that pattern, but only Debian using it, I would ask you to open an issue to the upstream first to discuss a proper solution that affects all consumers. 

Still, for your specific case, I would recommend building lldpd with the compiler defines `HAVE_LIBREADLINE` and `HAVE_EDITLINE_READLINE_H` but **not** with `HAVE_READLINE_HISTORY`: 

https://github.com/lldpd/lldpd/blob/9034c9332cca0c8b1a20e1287f0e5fed81f7eb2a/src/client/client.h#L56

So you will be able to build without errors. 

In case you really want to proceed with that change, you have a few options:

- Use a [Artifactory](https://docs.conan.io/2/tutorial/conan_repositories/setting_up_conan_remotes/artifactory/artifactory_ce_cpp.html) instance with a copy your changed recipe and built package.

- Use [Local Recipes Index](https://docs.conan.io/2/tutorial/conan_repositories/setup_local_recipes_index.html) to point your fork as the remote when building editline.

Regards.

jcar87:
Closing this due to inactivity and given the evidence provided by @uilianries 

for lldpd in particular, the logic that termines whether that header is attempted for inclusion is here:
https://github.com/lldpd/lldpd/blob/9034c9332cca0c8b1a20e1287f0e5fed81f7eb2a/m4/ax_lib_readline.m4#L113

I'm a bit surprised - it does appear to be able to handle the case where there's no additional header required to make it work with editline. May want to reach out to the lldpd maintainers first about this, thanks!

PR #25828, opened 2024-11-04
[efsw] Honor fPIC and require Conan 2.x features

uilianries:
Changes to recipe:  **efsw/1.4.0**

#### Motivation

The PR #25448 introduced the new recipe for efsw. However, it missed my review. 

#### Details

- The upstream enforces fPIC always, overriding the recipe option: https://github.com/SpartanJ/efsw/blob/1.4.0/CMakeLists.txt#L5
- The upstream generates a second static library that's packaged, but not listed in cpp_info: 
  - https://github.com/SpartanJ/efsw/blob/1.4.0/CMakeLists.txt#L27
  - https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-25448/11/package_build_logs/build_log_efsw_1_4_0_6542073374ca622afe046e6c97ae7645_13be611585c95453f1cbbd053cea04b3e64470ca.success.txt
- Explicit is no longer required as the require requires Conan 2.x. Using implements feature.

---

### DIFF
--- a/recipes/efsw/all/conanfile.py
+++ b/recipes/efsw/all/conanfile.py
@@ -1,12 +1,12 @@
 from conan import ConanFile
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
-from conan.tools.files import copy, get, rm, rmdir
+from conan.tools.files import copy, get, rm, rmdir, replace_in_file
 from conan.tools.microsoft import is_msvc, is_msvc_static_runtime
 import os
 
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=2.1"
 
 
 class EfswConan(ConanFile):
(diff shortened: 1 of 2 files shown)


Julianiolo:
@uilianries thanks for the thorough review! :)

I'm curious on what you mean exactly by "honor fpic", since you seem to be patching out the hard coded command to enable fpic?

uilianries:
> I'm curious on what you mean exactly by "honor fpic", since you seem to be patching out the hard coded command to enable fpic?

The recipe offers fPIC option, which is managed already by the Conan toolchain, in this CMakeToolchain. It injects `-fpic` already via CMake. So, in case you pass `-o &:fPIC=False` the package ID will point that the binaries was built without fPIC, but actually was the opposite, because the upstream forces fPIC always, by using set(...) in its Cmakelists.txt.

You can check it in the official Conan docs: 
https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#generated-files

The same applies for BUILD_SHARED_LIBS when using CMakeToolchain.

uilianries:
> Also, does it now only build a shared library? (never a static lib?)

No, now it will generate OR libefsw.a OR libefsw.so. Please, check the build log that I linked in the description, from your PR. It generated a twin static library named as `libefsw-static.a`, plus, the libefsw.a, both were installed in the package folder, but only one was listed to be linked. 

They are twins, you can check the upstream's CMakeLists.txt, the idea was generating both static and shared at same build, but there is this is small bug when building static.

Julianiolo:
Ah ok, my bad. I interpreted it as honoring efsw's decision to always be fpic, but now it makes sense.

danimtb (approved):

PR #25832, opened 2024-11-04
exprtk: add version 0.0.3

toge:
### Summary
Changes to recipe:  **exprtk/0.0.3**

#### Motivation
There are several improvements in 0.0.3.
There is the official license file since 0.0.3.

#### Details
https://github.com/ArashPartow/exprtk/compare/0.0.2...0.0.3

---

### DIFF
--- a/recipes/exprtk/config.yml
+++ b/recipes/exprtk/config.yml
@@ -1,5 +1,5 @@
 versions:
-  "0.0.3:
+  "0.0.3":
     folder: "all"
   "0.0.2":
     folder: "all"
--- a/recipes/exprtk/config.yml
+++ b/recipes/exprtk/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "0.0.3:
+    folder: "all"
   "0.0.2":
     folder: "all"
   "0.0.1":
--- a/recipes/exprtk/all/conandata.yml
+++ b/recipes/exprtk/all/conandata.yml
@@ -1,8 +1,10 @@
 sources:
+  "0.0.3":
+    url: "https://github.com/ArashPartow/exprtk/archive/0.0.3.tar.gz"
+    sha256: "6c412af7862acdfea74c092e71d33cfc32e8aa361fe535affc86665bf21acef6"
   "0.0.2":
     url: "https://github.com/ArashPartow/exprtk/archive/0.0.2.tar.gz"
     sha256: "7e8de4a0bfc9855c1316d8b8bc422061aef9a307c2f42d2e66298980463195c1"
-
   "0.0.1":
-    url: "https://github.com/ArashPartow/exprtk/archive/refs/tags/0.0.1.tar.gz"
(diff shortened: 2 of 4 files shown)


jcar87:
The error during validation is:
```
 "Error parsing YAML: while parsing a block mapping\n  in \"<unicode string>\", line 1, column 1:\n    versions:\n    ^\nexpected <block end>, but found '<scalar>'\n  in \"<unicode string>\", line 3, column 14:\n        folder: \"all\"\n                 ^"
§§
```

which is caused by a missing double quote in the yaml file (also reported by the linter, but slightly differently)

recipes/exprtk/config.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  "0.0.3:
```
jcar87, line 2:
```suggestion
  "0.0.3":
```

AbrilRBS (approved):
Thanks!

danimtb (approved):

ArashPartow:
@toge  Thanks for the update, however can you please explain why you felt the need to change the example?

ArashPartow:
@jcar87 wrt: "Feel free to analyse the code, inquire the PR author or voice any other concerns in the PR where this happened: https://github.com/conan-io/conan-center-index/pull/25832"

I already did ( https://github.com/conan-io/conan-center-index/pull/25832#issuecomment-2455437938 ) yet have not heard back from the PR creator.

How should we proceed?  the short of it would be to remove this particular tag from Conan.

valgur:
@ArashPartow Please consider using an `-rc` or `-alpha` suffix when tagging non-final releases to minimize any chance of confusion im the future. It's preferable if the tags can be considered immutable.

ArashPartow:
@valgur that's a good idea, but how about an even better one: We don't add every single random tag from each project into Conan, regardless of suffixes or prefixes. and we instead do what projects like ossfuzz and Conda etc do and make any such PRs pending until users (not just conan admins etc) that have been formally registered to review and ok the PR and not gamify the creation of PRs in such important infra like Conan.  Just an idea.

valgur:
> @valgur that's a good idea, but how about an even better one: We don't add every single random tag from each project into Conan, regardless of suffixes or prefixes. and we instead do what projects like ossfuzz and Conda etc do and make any such PRs pending until users (not just conan admins etc) that have been formally registered to review and ok the PR and not gamify the creation of PRs in such important infra like Conan.  Just an idea.  
> 
> 
> 
> 

Ok, I'm a random user of VTK that wanted to use the latest, presumably the most bug-free, version of ExprTk. Not hypothetically. How am I supposed to know that your latest release is a stable one or not? SemVer is not that complicated and exists as a common convention for a good reason.

ArashPartow:
@valgur  wrt VTK/Paraview I think @spyridon97 is handling that side of things.  https://gitlab.kitware.com/paraview/paraview/-/issues/22554

PR #25836, opened 2024-11-04
Revert "exprtk: add version 0.0.3 (#25832)"

ArashPartow:
Revert the 0.0.3 update until verification by external testers has completed as the tag commit  may change.

fyi: @toge  @jcar87  @AbrilRBS

PR #25840, opened 2024-11-05
plutovg: add version 0.0.9

toge:
### Summary
Changes to recipe:  **plutovg/0.0.9**

#### Motivation
There is a new feature in 0.0.9.

#### Details
https://github.com/sammycage/plutovg/compare/v0.0.8...v0.0.9

---

### DIFF
--- a/recipes/plutovg/all/conandata.yml
+++ b/recipes/plutovg/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.0.9":
+    url: "https://github.com/sammycage/plutovg/archive/refs/tags/v0.0.9.tar.gz"
+    sha256: "462a07ef38ecb2c3ed4404a675a7ed9b65002b5c9065b8b7fd6e4b808eef0fbd"
   "0.0.8":
     url: "https://github.com/sammycage/plutovg/archive/refs/tags/v0.0.8.tar.gz"
     sha256: "090dd5d38e04e0900bf6843b2b38ce7108cab815c1b5496c934af65c51965618"
--- a/recipes/plutovg/all/conanfile.py
+++ b/recipes/plutovg/all/conanfile.py
@@ -14,9 +14,9 @@ class PlutoVGConan(ConanFile):
     name = "plutovg"
     description = "Tiny 2D vector graphics library in C"
     license = "MIT",
-    topics = ("vector", "graphics", )
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/sammycage/plutovg"
+    topics = ("vector", "graphics", )
     package_type = "library"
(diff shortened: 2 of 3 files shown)


jcar87:
Thanks @toge 

We have asked before about what process you follow to determine there is a new version - as it has happened quite a few times that we end up publishing a version tracking a tag that ends up changing because it's not an actual release. I'll ask again, in the interest of the community, so that we can take appropriate measure to avoid this.

I will close this PR - on the basis that this is a `0.0.9` version that was released ~2 hours ago - I think it might be more beneficial for less-established libraries and pre-releases to react on demand from users, so that we can have a better idea on how to prioritise this.

Also keep in mind that we've had issues in the past - the first 24-48 hours after a release are the timeframe in which the maintainers are most likely to "retag" or "republish" the same version. Please abstain for opening PRs to update versions that have existed for such a short period of time, thanks!

toge:
@jcar87 
I understood.
I am going to follow community policy.

I had thought that the strength of package managers - with multiple versions to choose - is that user could use the latest release as soon as it was released, but I had not considered the negative impact that this had on the community and the review process.

When upgrading, I used to check for changes and only update if there was a modification that would change the execution, but now I will try to create a PR after some time has passed.

ArashPartow:
@toge:  

"that the strength of package managers - with multiple versions to choose"  

This is an absolutely false and flawed assumption. One of the key strengths of a robust package management system is that it only makes available the best versions/instance of the packages it maintains.

This should NOT be a race to to see who can make a PR for a new commit to a repo being maintained by Conan and get it merged into conan. It's fine to wait some time, a day even a week etc just to see how things pan out - no need to literally PR to Conan **47 minutes** after a commit has landed.



@jcar87 : Seeing as though you're attending PRs can you **please** look into  https://github.com/conan-io/conan-center-index/pull/25836

ArashPartow:
@jcar87  -  _**" I'll make sure the version is removed before going live."**_ thank-you that sounds like the best solution. 

That being said, should I close the revert PR or how should I proceed from here - as other needless changes were made to the package such as the test code was needlessly modified in the package (gives off  **xz controversy/fiasco** vibes)

jcar87:
Close the PR.  We'll align the test package so that it does nothing - test packages in conan center should be as minimal as possible - we tend to be lenient to avoid en masse refactores. Reality is that neither version of the test package is ideal -  Conan Center has never been a testing service - will make sure it conforms to the minimum we require.

>  as other needless changes were made to the package such as the test code was needlessly modified in the package (gives off xz controversy/fiasco vibes)

Feel free to analyse the code, inquire the PR author or voice any other concerns in the PR where this happened: https://github.com/conan-io/conan-center-index/pull/25832

PR #25848, opened 2024-11-05
Add support for building Qt Advanced Docking System version 4.3.1

DonOregano:
### Summary
Changes to recipe:  **qt-advanced-docking-system/4.3.1**

#### Motivation
Add support for building new version of Qt Advanced Docking System. I have not touched the old recipe, since it appears to be Conan 1 only. Maybe that should be removed.

#### Details
Add a new recipe for 4.x.x versions of the package, and move the previous "all" to a 3.x.x directory. Maybe the 3.x.x should be removed and 4.x.x renamed to all, but that decision should be made by someone who knows more about "the maintaining old packages strategy of Conan".

Also added a test_package that just tries to build against the package. The test does not run anything.

The one "weirdness" of this package is that I want it to be optional whether the Qt used comes from Conan or from another installation of Qt, from outside of Conan. I could not find an example of how to do this, so I added an option "qt_from_conan" that controls this. Please let me know if there is a better way to do this!

---


DonOregano:
There were a couple of previous attempts at this that I did not notice when I started creating this PR. But they were both over a year old...

jcar87:
we have merged the update to conan 2 and the new version in https://github.com/conan-io/conan-center-index/pull/26496 - please let us know if there are issues, thanks!

PR #22980, opened 2024-03-05
fix gmp/6.3.0:ctest:command not found issue

crstzh:
For gmp/6.3.0, it might show the below error if CMake is not present in the run environment.

```
======== Testing the package: Executing test ========
gmp/6.3.0 (test package): Running test()
gmp/6.3.0 (test package): RUN: ctest --output-on-failure -C Release
/bin/sh: line 1: ctest: command not found
```
we can bypass this by using the `conanbuild` environment on non-Windows platforms.

Specify library name and version:  **gmp/6.3.0**
- [X ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [X ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #25859, opened 2024-11-06
gmp: dont use ctest in test package

jcar87:
### Summary
Changes to recipe:  gmp/all (test package only)

#### Motivation
See PR https://github.com/conan-io/conan-center-index/pull/22980 

### Details
* Don't use ctest - in particular, to work around the inability to call `self.dependencies` in the test() method
* Simplify the test:
   - Conan 2 handles the fpic flag for consumers as well, so it's unclear what the test fpic test was trying to work around, perhaps a Conan 1 limitation
   - There was also a logic error in the test, which leaves `fPIC` enabled for the C++ test, regardless of whether it was previously false or not - best to leave this handled by CMakeToolchain
   - remove Conan 1x. test package
   - simplify implementation of C test to minimum possible


uilianries (approved):
LGTM. Much simpler!

PR #25867, opened 2024-11-07
tree-sitter-cql: add recipe

toge:
### Summary
Changes to recipe:  **tree-sitter-cql/***

#### Motivation
CQL(Cassandra Query Language) parser on tree-sitter.

#### Details
https://github.com/shotover/tree-sitter-cql/tree/main/src

---

### DIFF
--- a/recipes/tree-sitter-cql/all/CMakeLists.txt
+++ b/recipes/tree-sitter-cql/all/CMakeLists.txt
@@ -4,7 +4,7 @@ project(tree-sitter-cql LANGUAGES C)
 find_package(tree-sitter REQUIRED CONFIG)
 
 add_library(${PROJECT_NAME}
-    src/parser.c
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/src/parser.c"
 )
 target_link_libraries(${PROJECT_NAME}
     PUBLIC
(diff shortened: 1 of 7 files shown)


perseoGI (approved):
LGTM now! Thank you Toge 🐸

toge:
@perseoGI 
Thank you for your improvements!

uilianries:
I did few changes to this PR:

- The upstream supports both static and shared on Windows, the configuration is over the ifdef TREE_SITTER_HIDE_SYMBOLS: https://github.com/shotover/tree-sitter-cql/blob/v0.2.0/src/parser.c#L22349
- Simplified the test package to consume only CMakeDeps, I see no reason for testing using extra pkgconfig in this case
- Updated the recipe to be 2.x compatible only, using implements


NOTE:

The upstream has a Makefile already, I would be more inclined to use that, but the same is not prepared for Windows and keeping a build for msvc (CMake) and another with Make for Linux and Macos is even worst, so adopting an extra cmakefile is less bad I hope.

This project receives only few updates and last time was 6 months ago, plus, I only found Gentoo packaging it, so there is no much say about adding more complexity when packaging.

toge:
@uilianries 
Thanks a lot for improvements!

perseoGI (approved):

Issue #25605, opened 2024-10-12
[request] SDL3


skhaz:
### Package Name/Version

SDL/3.1.3

### Webpage

https://www.libsdl.org

### Source code

https://github.com/libsdl-org/SDL/releases/tag/preview-3.1.3

### Description of the library/tool

Could we please have SDL3?

zby-c:
+1, please

AbrilRBS:
Initial work is https://github.com/conan-io/conan-center-index/pull/25873, contributions welcomed!

skhaz:
I can help to test on Windows, Linux, macOS and WebAssembly (emsdk), for sure! Thank you.

PR #25873, opened 2024-11-07
sdl3: Add new major version support

AbrilRBS:
A new recipe with a different approach was needed, to have better maintainability for future versions. Most of the options/behaviour has been carried over, but if you need any specific option we're missing, please do let us know!

Closes https://github.com/conan-io/conan-center-index/issues/25605
Closes https://github.com/conan-io/conan-center-index/pull/26464


recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def package(self):
+        copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install()
+
+    def package_info(self):
+        self.cpp_info.libs = ["SDL3"]
+        self.cpp_info.set_property("cmake_file_name", "SDL3")
+        self.cpp_info.set_property("cmake_target_name", "SDL3::SDL3")
```
madebr, line 183: In addition, create a `SDL3::SDL3-static` when creating a static library and
`SDL3::SDL3-shared` when building a shared library.

Also create `SDL3::Headers` for the headers.

recipes/sdl/config.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  "3.1.6-pre":
+    folder: 3.x
```
elvisdukaj, line 3:
```suggestion
  "3.2.0":
    folder: 3.x
```

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,193 @@
...
+        # - If they have deps, its default value is False
+        # - Otherwise, they are True
+        # See CMakeLists.txt#L220, but we #L1167 seems to indicate otherwise, so set them all to True
+        for subsystem, dependencies in _subsystems:
+            # Don't use is with get_safe, it does not return None, but a _PackageOption with a None value
+            # Let the overridden equality check if the option is None
+            if self.options.get_safe(subsystem) == None:
+                default_subsystem_value = True
+                if subsystem == "hidapi" and self.settings.os == "visionOS":
+                    # CMakeLists.txt#L343
```
elvisdukaj, line 75:
```suggestion
                    # CMakeLists.txt#L378
```

On the 3.2.0 this is line 378

elvisdukaj (commented):
When commenting the "hidapi" on Windows I was able to build this package for Windows and msvc shared/static and debug/release

skhaz:
Excuse my ignorance, but how can I help with testing? I’m very interested in having SDL3 in the emsdk.

elvisdukaj:
> Excuse my ignorance, but how can I help with testing? I’m very interested in having SDL3 in the emsdk.

@skhaz, if you have the project cloned and this PR than you need to run the following command. If you don't you need the [GitHub CLI](https://cli.github.com) tools, autentificate and than `gh pr --checkout 25873` or something similar/

```
cd recipes/sdl
conan create 3.x --version <version> --profile <profile_emsdk> 
```

This PR is actually using a pre-release version, see the config.yml file. I suggested to move on the 3.2.0 instead.

elvisdukaj:
I am trying to build this recipe for macOS. Let's see how far I can go.

@AbrilRBS should I fork from you and proposing the changes there, suggest changes here, or what is your preferred way? Let me know :)

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,227 @@
...
+    settings = "os", "arch", "compiler", "build_type"
+
+    options = {
+        **{
+            "shared": [True, False],
+            "fPIC": [True, False],
+        },
+        **{subsystem: [None, True, False] for subsystem, _ in _subsystems},
+        **{
+            "alsa": [None, True, False],
```
madebr, line 47: This is missing options for x11, Wayland and probably others.
(Check the summary at the end of configuration)

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,227 @@
...
+        # Todo: Remove after recipe develop is finished
+        tc.cache_variables["SDL_TEST_LIBRARY"] = True
+        tc.cache_variables["SDL_TESTS"] = True
+        tc.cache_variables["SDL_EXAMPLES"] = True
+        tc.cache_variables["CMAKE_TRY_COMPILE_CONFIGURATION"] = str(self.settings.build_type)
+        if self._needs_libusb:
+            tc.cache_variables["SDL_HIDAPI_LIBUSB"] = True
+            tc.cache_variables["SDL_HIDAPI_LIBUSB_SHARED"] = self.dependencies["libusb"].options.get_safe("shared", False)
+        if self.options.get("pulseaudio"):
+            tc.cache_variables["SDL_PULSEAUDIO"] = True
```
madebr, line 178: Is there a reason you're not assigning `bool(self.options.get("pulseaudio"))` h

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,227 @@
...
+
+    def package_info(self):
+        sdl_lib_name = "SDL3"
+        if (is_msvc(self) or self._is_clang_cl) and not self.options.shared:
+            sdl_lib_name = f"{sdl_lib_name}-static"
+        self.cpp_info.libs = [sdl_lib_name]
+        self.cpp_info.set_property("cmake_file_name", "SDL3")
+        self.cpp_info.set_property("cmake_target_name", "SDL3::SDL3")
+        if self.options.shared:
+            self.cpp_info.set_property("cmake_target_aliases", ["SDL3::SDL3-static"])
```
madebr, line 215: This logic is inverted.

recipes/sdl/3.x/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "3.2.2":
```
madebr, line 2: We're at 3.2.4 already

madebr (changes requested):

recipes/sdl/3.x/conandata.yml
AbrilRBS: Thanks for the heads-up!

recipes/sdl/3.x/conanfile.py
AbrilRBS: Yes, we have yet to implement the rest of the options, we're testing the best approach with these 2 for now :)

recipes/sdl/3.x/conanfile.py
AbrilRBS: We still need the conditional for the variable below, as accessing a non existing dependencie errors out, so at that point I think this way is cleaner for both :)

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,224 @@
...
+        **{subsystem: None for subsystem, _ in _subsystems},
+        **{
+            "alsa": None,
+            "pulseaudio": None,
+        }
+    }
+
+    def config_options(self):
+        if self.settings.os == "Windows":
+            del self.options.fPIC
```
madebr, line 66: You can remove alsa and pulseaudio here on Android and Apple.

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,224 @@
...
+        tc.cache_variables["SDL_STATIC"] = not self.options.shared
+        # Todo: Remove after recipe develop is finished
+        tc.cache_variables["SDL_TEST_LIBRARY"] = True
+        tc.cache_variables["SDL_TESTS"] = True
+        tc.cache_variables["SDL_EXAMPLES"] = True
+        tc.cache_variables["CMAKE_TRY_COMPILE_CONFIGURATION"] = str(self.settings.build_type)
+        if self._needs_libusb:
+            tc.cache_variables["SDL_HIDAPI_LIBUSB"] = True
+            tc.cache_variables["SDL_HIDAPI_LIBUSB_SHARED"] = self.dependencies["libusb"].options.get_safe("shared", False)
+        if self.options.get("pulseaudio"):
```
madebr, line 177: I think you need to use `get_safe` here.

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,253 @@
...
+            "sndio": False,
+            ## Video
+            "opengl": True,  # TODO: Off by default in apple_os
+            "opengles": True,
+            ## Other
+            "libudev": True,
+        }
+    }
+
+    def config_options(self):
```
madebr, line 76: I'd also throw a ConanException here if the os is Android because without Java classes, the compiled C library is useless.

SDL's CMake script compiles the java classes, so Android can be supported if you provide a jdk and a Android SDK (it needs android.jar).

madebr: No problem. Documenting this, and raising an exception is always useful for the next person.

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,450 @@
...
+            tc.cache_variables["SDL_X11_XVM"] = self.options.get_safe("xvm")
+
+        with_wayland = self.options.get_safe("wayland", False)
+        tc.cache_variables["SDL_WAYLAND"] = with_wayland
+        if with_wayland:
+            tc.cache_variables["SDL_WAYLAND_SHARED"] = self.dependencies["wayland"].options.shared
+        if not with_x11 and not with_wayland:
+            # Disable windowing support:
+            # https://github.com/libsdl-org/SDL/blob/main/docs/README-cmake.md#cmake-fails-to-build-without-x11-or-wayland-support
+            tc.cache_variables["SDL_UNIX_CONSOLE_BUILD=ON"] = True
```
valgur, line 313:
```suggestion
            tc.cache_variables["SDL_UNIX_CONSOLE_BUILD"] = True
```

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,450 @@
...
+            "libudev": [True, False],
+            "dbus": [True, False],
+            "libusb": [True, False],
+        }
+    }
+
+    default_options = {
+        **{
+            "shared": False,
+            "fPIC": False,
```
valgur, line 73:
```suggestion
            "fPIC": True,
```
Probably wasn't intentional?

valgur:
@AbrilRBS @perseoGI
I opened a PR for complete X11 support: https://github.com/AbrilRBS/conan-center-index/pull/13

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,470 @@
...
+            if self.options.xcursor:
+                self.cpp_info.requires.append("xorg::xcursor")
+            if self.options.xinput:
+                self.cpp_info.requires.append("xorg::xi")
+            if self.options.xfixes:
+                self.cpp_info.requires.append("xorg::xfixes")
+            if self.options.xrandr:
+                self.cpp_info.requires.append("xorg::xrandr")
+            if self.options.xscrnsaver:
+                self.cpp_info.requires.append("xorg::xscrnsaver")
```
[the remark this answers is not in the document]
valgur: Good point.
It sounds like these dynamically-loaded dependencies should be using `self.requires(..., libs=False, run=True)` or something similar to exclude them from `self.cpp_info.requires`. Which does not work as expected, though, afaik, based on some prior PRs.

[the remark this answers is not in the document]
madebr: Indeed! I did not know about the conan 2 `libs=xxx run=yyy` feature.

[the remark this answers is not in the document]
madebr: Also, building a shared SDL3 library needs this. When linking to a shared SDL3 library, you don't have to link to any dependency.

recipes/sdl/3.x/conanfile.py
AbrilRBS: We have been able to compile Android locally and run the test_package in an emulator, so my current idea is to let users build for Android if they have already set things up properly, but don't expect the recipe to fully provide it. Does that sound good to you?

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,466 @@
...
+        sdl_lib_name = "SDL3"
+        if (is_msvc(self) or self._is_clang_cl) and not self.options.shared:
+            sdl_lib_name = f"{sdl_lib_name}-static"
+        self.cpp_info.libs = [sdl_lib_name]
+        self.cpp_info.set_property("cmake_file_name", "SDL3")
+        self.cpp_info.set_property("cmake_target_name", "SDL3::SDL3")
+        self.cpp_info.set_property("cmake_target_aliases", [
+            "SDL3::SDL3-shared" if self.options.shared else "SDL3::SDL3-static",
+            # Modelling the headers as a target is not quite feasible, so use an alias instead
+            "SDL3::Headers"
```
valgur, line 354: Could be done with a component that sets `includedirs` and no `libs` or `requires`, no?

recipes/sdl/3.x/conanfile.py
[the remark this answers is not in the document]
AbrilRBS: Thanks for the clarification. It's also my understanding that there's no way to have Conan transparently handle this case for now, and that creating the unused component might be the only way, but that also feels iffy.

I'd thus like to understand if the overlinking here causes too much damage to consumers

recipes/sdl/3.x/conanfile.py
```diff
@@ -0,0 +1,193 @@
...
+# Subsystems, CMakeLists.txt#L234
+_subsystems = [
+    ("audio", []),
+    ("video", []),
+    ("gpu", ["video"]),
+    ("render", ["video"]),
+    ("camera", ["video"]),
+    ("joystick", []),
+    ("haptic", ["joystick"]),
+    ("hidapi", []),
```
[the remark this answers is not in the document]
AbrilRBS: We have fixed the libusb issues here, which can't be build as shared for now until the new CMakeConfigDeps is generally available in the Conan client, thanks for testing on your end :)

recipes/sdl/3.x/conanfile.py
AbrilRBS: Yes, that's another approach and I might just go for it after we check that the rest of the recipe looks ok - having only the global component would be better but alas

AbrilRBS:
We think that the recipe is now mostly there, all our local tests on different configurations seem ok - we additionally tested on Android (@perseoGI will post the logs shortly)

If anyone can take a double look and test the generated package on their end, we'd appreciate it, thanks!

The known issues/limitations for now are:
- Overlinking when shared
- Linking with `SDL3::Headers` will link everything, it's currently an alias for the global library

recipes/sdl/3.x/conanfile.py
madebr: No, the test_package is a executable that runs in cli mode only. 
Android gui executables must be built as a shared library, and have the java classes.

I think you can add android support by addind a jdk and android-ndk to the (build) requirements.
If cmake finds those, it'll create a SDL3-3.x.y.jar archive.

madebr: FYI, I added support for building the SDL tests (not the examples) as a single-arch-debug-apk to CMake (if it finds all dependencies, only tested on Linux, and is only for internal SDL use). This was done to avoid Gradle.

recipes/sdl/3.x/conanfile.py
madebr: This will cause an issue if you ever add sdl2-compat to cci

perseoGI:
Android compilation logs:

```sh
$ conan profile show -pr android-packed 

[settings]
arch=armv8
build_type=Release
compiler=clang
compiler.cppstd=17
compiler.libcxx=c++_static
compiler.version=17
os=Android
os.api_level=26
[tool_requires]
*: cmake/3.29.7, android-ndk/r27c
[conf]
tools.build:verbosity=verbose
tools.compilation:verbosity=verbose
tools.system.package_manager:sudo=true

```

```sh
$ conan create recipes/sdl/3.x/ --version 3.2.4 --build='&' -pr:h android-packed
```

[sdl-3.2.4-android-compilation.txt](https://github.com/user-attachments/files/19013228/sdl-3.2.4-android-compilation.txt)

And a quick test running on an Android emulator:

```
$ emulator -avd Pixel_8_API_VanillaIceCream -netdelay none -netspeed full
<other terminal>
$ adb push recipes/sdl/3.x/test_package/build/clang-17-armv8-17-release/test_package /data/local/tmp 
$ adb shell file /data/local/tmp/test_package
/data/local/tmp/test_package: ELF shared object, 64-bit LSB arm64, dynamic (/system/bin/linker64), for Android 26, built by NDK r27c (12479018), BuildID=0ce060d23c7564c0d2c3840e3ce4db826110d874, not stripped
$ adb shell
$ [adb] /data/local/tmp/test_package
SDL version 3002004
```

recipes/sdl/3.x/conanfile.py
AbrilRBS: > I think you can add android support by adding a jdk and android-ndk to the (build) requirements.

This is something that we expect users to add using their profiles instead of having recipes explicitly depend on those

madebr: That's reasonable.
I expect the sdl3 conan recipe to fail in build or package when the java classes are not built.

AbrilRBS: The issue is that this might be a bit hard to detect the case where the user has properly configured it on their side and when they have not

madebr: You can detect by checking for a few `ANDROID_*` environment variables, and the presence of `javac` and `jar.
But that will break easily.
The pragmatic approach is to fail the build when the jar file is not created.
Don't overcomplicate things.

perseoGI:
Greetings @madebr,

I'm attaching a short screen record where the execution of different SDL examples can be seen:

https://github.com/user-attachments/assets/36d60bd2-d952-433a-bd10-23bd7c9cdf7b

This is running on a MacOS machine. 
Testing examples o Android will require creating a native application and importing SDL as dependency. 
I have not bandwidth enough to create such examplenright now.

madebr:
That's great! All I'm asking is to throw an error when building an android package (with a few comments why), so the next dev knows what to do.

recipes/sdl/3.x/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "3.2.4":
```
madebr, line 2: 3.2.6 :)

recipes/sdl/3.x/conanfile.py
AbrilRBS: Looks like I was wrong, adding it as a proper component was not quite as involved as I first thought!

AbrilRBS:
> That's great! All I'm asking is to throw an error when building an android package (with a few comments why), so the next dev knows what to do.

@madebr we have guarded Android builds behind a conf, users that have properly set up their env can set it to build this for Android as needed

perseoGI (approved):
🐸

madebr (approved):

jcar87:
Thanks everyone for your involvement in this!

Issue #23456, opened 2024-04-10
[package] opentelemetry-cpp/1.12.0: ERROR: opentelemetry-cpp/1.12.0: Required package 'prometheus-cpp' not in component 'requires'


shaqtsui:
### Description

When install opentelemetry-cpp with options: `opentelemetry-cpp/*:with_prometheus=True`
Got: ERROR: opentelemetry-cpp/1.12.0: Required package 'prometheus-cpp' not in component 'requires'

### Package and Environment Details

* Package Name/Version: **opentelemetry-cpp/1.12.0**
* Operating System+version: **Windows 10**
* Compiler+version: **msvc 2022**
* Conan version: **conan 2.2.2**
* Python version: **Python 3.12.1**

### Conan profile

profile buid(profile_build.txt):
```
[settings]
arch=x86_64
build_type=Release
os=Windows
compiler=msvc
compiler.version=193
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
```
profile host(profile_host.txt):
```
[settings]
arch=x86_64
build_type=Release
os=Windows
compiler=msvc
compiler.version=193
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
```

conanfile.txt
```
[requires]
opentelemetry-cpp/1.12.0
[options]
opentelemetry-cpp/*:with_prometheus=True
```

### Steps to reproduce

`conan install . --profile:build=profile_build.txt --profile:host=profile_host.txt --build=missing`

### Logs

<details><summary>Click to expand log</summary>

```
PS D:\Projects\optlex\build\.conan\opentelemetry-cpp\1.12.0> conan install . --profile:build=profile_build.txt --profile:host=profile_host.txt --build=missing

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows


======== Computing dependency graph ========
Graph root
    conanfile.txt: D:\Projects\optlex\build\.conan\opentelemetry-cpp\1.12.0\conanfile.txt
Requirements
[...]

PR #25879, opened 2024-11-07
opentelemetry-cpp: Cleanup and fixes for new versions

AbrilRBS:
Started as a fix for https://github.com/conan-io/conan-center-index/issues/23456, but that has been fixed for a while now, and took the chance to improve the recipe a bit, besides what other concurrent PRs are doing now


recipes/opentelemetry-cpp/all/conanfile.py
```diff
@@ -172,11 +167,10 @@ def validate(self):
                 f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support."
             )
 
-        if self.settings.os != "Linux" and self.options.shared:
+        if Version(self.version) == "1.8.3" and self.settings.os != "Linux" and self.options.shared:
```
AbrilRBS, line 170: This is the main change, where the shared build issue seems not to reproduce outside this version, and all 3 others compile fine for me locally on macos

AbrilRBS:
🤔 now crosscompilation is an issue. Maybe the checks were right in the first place!

PR #25889, opened 2024-11-09
mpdecimal: Fix lib name on MSVC/shared, fix Mac cross-builds

### LINKED ISSUE #23994: [package] mpdecimal/2.5.1: CI fail when compiling on MacOS
### Description

Failed execution in CI: https://github.com/conan-io/conan-center-index/pull/23989

Failed when running test in CI in MacOS when building as Shared.

Ahajha:
When using the MSBuildDeps generator with mpdecimal(msvc/shared), it generates the wrong lib name and thus fails to build. I assume CMake does some more advanced logic and so this was never caught.

The correct name is (for example) `libmpdec-2.5.1.dll.lib`, previously this was just `libmpdec-2.5.1.lib`.

This change is necessary for supporting shared mpdecimal on Windows in CPython.

Also fixes #23994 (this is fixed by both of my other open mpdecimal PRs, since it's necessary to build. First one to be merged will fix it)

I also have two other mpdecimal PRs out: #25896 #25903 (superset of this one)


jcar87:
> I also added some extra cleanup commits, none of these are necessary and I'm happy to revert any of them if it makes it easier to review the PR 

yes please - this makes it a lot easier to review, especially with the size of the backlog - thanks!

Ahajha:
@jcar87 I reverted the extra changes, diff is much smaller now :) I'll make a followup PR after the important 3 are merged to just do all the cleanup.

recipes/mpdecimal/2.5.x/conanfile.py
```diff
@@ -105,9 +105,10 @@ def generate(self):
             tc.generate()
 
             deps = AutotoolsDeps(self)
-            if is_apple_os(self) and self.settings.arch == "armv8":
-                deps.environment.append("LDFLAGS", ["-arch arm64"])
-                deps.environment.append("LDXXFLAGS", ["-arch arm64"])
+            if is_apple_os(self):
```
jcar87, line 108: I appreciate this was already in the recipe, but I don't fully understand why it was needed.

Conan already adds `-arch xxx` when Cross building, and it should not be needed when *not* cross building in the sense that the compiler will already target the "native" architecture.
Pretty sure this can be removed, unless there's a CI log to look at?

Ahajha: https://github.com/conan-io/conan-center-index/issues/23994

Points to a log on May 14th

Build: https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23989/3-macos-clang/mpdecimal/2.5.1//9f69305d358977042aa4b289c18f45fcf5827ece-build.txt
Test: https://c3i.jfrog.io/c3i/misc-v2/logs/pr/23989/3-macos-clang/mpdecimal/2.5.1//9f69305d358977042aa4b289c18f45fcf5827ece-test.txt

The build log actually shows the exact invocation:
```
g++ -pthread -I. -I../libmpdec -Wall -Wextra -std=c++11 -pedantic -DNDEBUG -stdlib=libc++ -std=c++17 -m64 -O3 -isysroot /Applications/conan/xcode/13.0/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk -arch x86_64 -fPIC -c decimal.cc -o .objs/decimal.o
g++ -pthread -L../libmpdec -dynamiclib -install_name @rpath/libmpdec++.3.dylib -undefined dynamic_lookup -compatibility_version 2.5 -current_version 2.5.1 -o libmpdec++.2.5.1.dylib .objs/decimal.o -lmpdec -lm
ld: warning: ignoring file ../libmpdec/libmpdec.dylib, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
ld: warning: ignoring file .objs/decimal.o, building for macOS-arm64 but attempting to link with file built for unknown-x86_64
```

Ahajha: So maybe fully removing the section would fix it?

Ahajha: Actually, I might know why. I don't think `LDXXFLAGS` is a "standard" variable (correct me if I'm wrong). So maybe instead of this, if `LDFLAGS` is properly set, we could add `LDFLAGS` as well as `LDXXFLAGS` when linking.

Ahajha: I'll do some experiments locally and see.

Ahajha: I can also fix it by doing the following:
```
if cross_building(self):
    deps.environment.append("LDXXFLAGS", ["$LDFLAGS"])
```
That might be a more general cross-building fix too.

jcar87: Yes! it's exactly that - Conan is setting up the right variables, but they're not used by the build script - that's why it only affects the C++ library but the C one is fine.

```
            tc = AutotoolsToolchain(self)
            tc.configure_args.append("--enable-cxx" if self.options.cxx else "--disable-cxx")
            tc_env = tc.environment()
            tc_env.append("LDXXFLAGS", ["$LDFLAGS"])
            tc.generate(tc_env)
```

Something like the above will work - no need to make it conditional, as Conan also sets up other variables that may be useful (e.g the sysroot when needed).

Ahajha: Alternatively, I can add this in the build step (or add the equivalent as a patch)
```
replace_in_file(self, os.path.join(self.build_folder, "libmpdec++", "Makefile"), "$(LDXXFLAGS)", "$(LDXXFLAGS) $(LDFLAGS)")
```

@jcar87 thoughts on possible alternatives?

Ahajha: Never mind, just saw your comment. I like that version, I'll update it in a minute.

recipes/mpdecimal/2.5.x/conanfile.py
```diff
@@ -219,7 +215,10 @@ def package(self):
     def package_info(self):
         lib_pre_suf = ("", "")
         if is_msvc(self):
-            lib_pre_suf = ("lib", f"-{self.version}")
+            if self.options.shared:
+                lib_pre_suf = ("lib", f"-{self.version}.dll")
```
jcar87, line 219: I can see CI is not building windows + shared - so would need some logs that show that these are the files created `.dll.lib` on Windows n-

otherwise, we may as well consider - if a shared libmpdec++ is *never* possible on Windows - we could just make it the default on Windows so that it's not set to be built?

Ahajha: In #25903 I actually partially remove the restriction, it is in fact possible to mix these options as of 2.5.1. I'll see what I can find for logs.

Ahajha: In this log you can see the generated lib names:
https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-25889/6/package_build_logs/build_log_mpdecimal_2_5_1_254318b465fb6c8ce4fdfbadfde99a36_056ff9a90cff2a089a33d9c99cd4b6d10281a6b1.success.txt

You can see in the link step:
```
/out:libmpdec-2.5.1.dll /implib:libmpdec-2.5.1.dll.lib
```
and in the package step:
```
mpdecimal/2.5.1: package(): Packaged 1 '.dll' file: libmpdec-2.5.1.dll
mpdecimal/2.5.1: package(): Packaged 1 '.h' file: mpdecimal.h
mpdecimal/2.5.1: package(): Packaged 1 '.lib' file: libmpdec-2.5.1.dll.lib
```

jcar87: awesome!

jcar87: PR LGTM once we restore the default value of cxx :D

jcar87 (approved):
LGTM, thanks!

memsharded (approved):

PR #25892, opened 2024-11-09
cmake: add version 3.31.0

toge:
### Summary
Changes to recipe:  **cmake/3.31.0**

#### Motivation
There are lots of new features in 3.31.0.
https://github.com/Kitware/CMake/compare/v3.30.5...v3.31.0

#### Details
https://cmake.org/cmake/help/latest/release/3.31.html

---


jcar87 (changes requested):
please retain the 3.x.x folder and do not make any formatting modifications to the recipe (this is in general, for all PRs). In this case doing so would result in unnecessarily generating binaries for all versions, when nothing has affectively changed.

Thank you!

toge:
@jcar87 
Is 3.x.x still using?
I can't find the way to use it.

jcar87:
3.x.x was left for users of Conan Center Index that do not rely on the remote, and would like to manually export the recipe and build binaries from it - perhaps from a local fork, or for a repository they control - so the recipe is still maintained (albeit only on contributors help alone) even if it not checked by CI.

A comment may be handy - but at the same time, and we continue to ask all contributors - please refrain from doing maintenance on top of other changes. If the goal of the PR is to add version 3.31.0 - that's a single, straight forward change in only two files - same goes for removing the test_v1_package, refrain from doing this as well - we will take care of it (perhaps in a single commit down the line). By all means it's not harmful, but the quickest way to get a review from us is by sticking to the goal of the PR.

toge:
@jcar87 
Now I understand.
I will try to revert delete test_v1_package.

One more question.
Should I separate PRs add new version and removing older versions?

jcar87 (approved):
Verified checksums against: https://github.com/Kitware/CMake/releases/download/v3.31.0/cmake-3.31.0-SHA-256.txt

uilianries (approved):

PR #25895, opened 2024-11-09
fswatch: add package

winternet:
### Summary
Add [fswatch](https://github.com/emcrisostomo/fswatch) package.

#### Motivation
Make fswatch available to the community.

#### Details
While the latest release version is 1.17.1, there is a documented [known issue](https://github.com/emcrisostomo/fswatch/issues/291#issuecomment-1230541663) which leads to the version 1.17.1-36 

---

### DIFF
--- a/recipes/fswatch/all/CMakeLists.txt
+++ b/recipes/fswatch/all/CMakeLists.txt
@@ -1,0 +1,5 @@
+cmake_minimum_required(VERSION 3.5)
+project(cmake_wrapper)
+
+message(STATUS "Conan CMake Wrapper")
+add_subdirectory("src")
--- a/recipes/fswatch/all/CMakeLists.txt
+++ b/recipes/fswatch/all/CMakeLists.txt
@@ -1,5 +1,0 @@
-cmake_minimum_required(VERSION 3.5)
-project(cmake_wrapper)
-
-message(STATUS "Conan CMake Wrapper")
-add_subdirectory("src")
--- a/recipes/fswatch/all/conandata.yml
+++ b/recipes/fswatch/all/conandata.yml
@@ -1,7 +1,4 @@
 sources:
-  "1.17.1":
-    url: "https://github.com/emcrisostomo/fswatch/archive/refs/tags/1.17.1.zip"
-    sha256: "2a2be51807a7138cf50b327a9de29b96892dd357ab3142a5b4e36cb380ee4470"
-  "1.17.1-36":
-    url: "https://github.com/emcrisostomo/fswatch/archive/refs/tags/1.17.1-36-gba411e0.zip"
-    sha256: "520220c952208747206197cfdd403114c05f8d5612ccc0be9ded5e9a8261fe63"
+  "1.17.1.cci.20220902":
+    url: "https://github.com/emcrisostomo/fswatch/archive/ba411e0.zip"
+    sha256: "278476f0f2178bf59dac494c07d6a9b9d181eac8d14f37d17eaf8c6d04a3b4d9"
--- a/recipes/fswatch/all/conandata.yml
+++ b/recipes/fswatch/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
   "1.17.1.cci.20220902":
-    url: "https://github.com/emcrisostomo/fswatch/archive/ba411e0.zip"
+    url: "https://github.com/emcrisostomo/fswatch/archive/ba411e0d0fabcd5cbf0881f1380482e2f5ab9f47.zip"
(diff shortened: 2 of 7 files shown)


recipes/fswatch/all/conandata.yml
```diff
@@ -0,0 +1,7 @@
+sources:
+  "1.17.1":
+    url: "https://github.com/emcrisostomo/fswatch/archive/refs/tags/1.17.1.zip"
+    sha256: "2a2be51807a7138cf50b327a9de29b96892dd357ab3142a5b4e36cb380ee4470"
+  "1.17.1-36":
```
AbrilRBS, line 5: Note that this version, as per semver semantics, would come _before_ `1.17.1`, that is, `1.17.1-36 < 1.17.1`, which I don't think is the intention here!

We should clarify if upstream is expecting to release a new version soon with this fix, else we'll need to roll our own schema for this

recipes/fswatch/all/CMakeLists.txt
```diff
@@ -0,0 +1,5 @@
+cmake_minimum_required(VERSION 3.5)
+project(cmake_wrapper)
+
+message(STATUS "Conan CMake Wrapper")
+add_subdirectory("src")
```
AbrilRBS, line 5: Why is this needed? This is doing nothing, and directly using upstream's CMakeLists would work

AbrilRBS (commented):
Thanks a lot for taking the time to create the PR, we appreciate it!

I have some comments, specially regarding the chosen prerelease version scheme, but otherwise the recipe looks fine :)

recipes/fswatch/all/CMakeLists.txt
winternet: Thanks! I removed the wrapper

recipes/fswatch/all/conandata.yml
winternet: Nice catch, you're absolutely right. Since the latest commit was two years ago and the repository is more or less in hibernation mode, I don't think there will be a tag or release soon. Is there the possibility to refer to a specific [commit](https://github.com/emcrisostomo/fswatch/commit/ba411e0d0fabcd5cbf0881f1380482e2f5ab9f47) without the need to assign it a version?

winternet:
> Thanks a lot for taking the time to create the PR, we appreciate it!
> 
> I have some comments, specially regarding the chosen prerelease version scheme, but otherwise the recipe looks fine :)

Thanks for your review - very appreciated. I have some questions regarding the failing builds:
- Linux: Builds fine on my system but on CI the test would not build (`undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'`) which suggests that pthreads is missing - do you have a hint what could be missing?
- Windows/macOS: Is it mandatory to fix the builds or would it be ok to provide the package only for Linux?

recipes/fswatch/all/conandata.yml
AbrilRBS: Yes! We can set this version to `1.17.1.cci.YYYYMMDD` for the date of the latest commit, which is our way of indicating that the version comes from a specific commit and is not tagged

winternet: Thanks, I adapted the `1.17.1-36` version according to the proposed scheme and removed the (faulty) 1.17.1 version.

AbrilRBS (approved):

uilianries:
@winternet I did few changes your PR, fixing some previous errors.

- The pthread is indeed always required. There is a package with a port of pthread for Windows, but I think is too much, let's skip msvc instead
- The project has hardcoded configuration for Mac, which fails when cross-building from Mac M1 (CI environment) to Intel. I would not suggest trying to fix it, if the upstream does not support it.


> Windows/macOS: Is it mandatory to fix the builds or would it be ok to provide the package only for Linux?

In case a contributor has no time or access to a platform, it should not be a blocker, we can merge the partial recipe and add messages and invalid configuration asking to future support.

> Linux: Builds fine on my system but on CI the test would not build (undefined reference to symbol 'pthread_create@@GLIBC_2.2.5') which suggests that pthreads is missing - do you have a hint what could be missing?

I use Linux as well (BTW I use Arch), in case you are having trouble to reproduce the same result, I would recommend using the same docker image as used by the CI:

    docker pull conanio/gcc11-ubuntu16.04:2.9.2
    cd conan-center-index/recipes/fswatch/
    docker run --rm -ti -v $PWD:/home/conan/project --platform=linux/amd64 conanio/gcc11-ubuntu16.04:2.9.2
    cd /home/conan/project/
    conan create all --version=1.17.1.cci.20220902

uilianries (approved):

PR #25896, opened 2024-11-09
mpdecimal: Add version 4.0.0

Ahajha:
CPython 3.13 uses this version


recipes/mpdecimal/2.5.x/conanfile.py
```diff
@@ -205,7 +206,7 @@ def package(self):
             mpdecppdir = self.build_path / "libmpdec++"
             copy(self, "mpdecimal.h", src=mpdecdir, dst=pkg_dir / "include")
             if self.options.cxx:
-                copy(self, "decimal.hh", src=mpdecppdir, dst=pkg_dir / "include")
+                copy(self, "decimal.hh", src=self.source_path / "libmpdec++", dst=pkg_dir / "include")
```
Ahajha, line 209: This section is weird. For some reason, in the cache folder, most of the the files are symlinks in 4.0.0, but actual files in 2.5.1. So we need to copy from the source folder instead. But this only affects the `libmpdec++` folder. In fact, making this change to the `libmpdec` folder causes the C header to not get packaged. I don't understand any of why this is the case, but it works like this.

Also to note: I'm just being consistent here with the usage of `*_path` variables, I know they're deprecated.

recipes/mpdecimal/2.5.x/patches/4.0.0-0001-msvc-fixes.patch
```diff
@@ -0,0 +1,97 @@
...
+@@ -27,9 +27,9 @@ WARN = /W4 -Wno-undefined-inline
+ WARN = /W4
+ !endif
+ 
+-MPD_CXXFLAGS = $(WARN) /nologo $(OPT)
+-MPD_CXXFLAGS_SHARED = /DBUILD_LIBMPDECXX $(WARN) /nologo $(OPT_SHARED) $(PGOFLAGS)
+-MPD_BIN_CXXFLAGS_SHARED = $(WARN) /nologo $(OPT_SHARED) $(PGOFLAGS)
++MPD_CXXFLAGS = $(WARN) /nologo $(OPT) $(CONAN_CXXFLAGS) $(CONAN_CFLAGS)
++MPD_CXXFLAGS_SHARED = /DBUILD_LIBMPDECXX $(WARN) /nologo $(OPT_SHARED) $(PGOFLAGS) $(CONAN_LDFLAGS)
++MPD_BIN_CXXFLAGS_SHARED = $(WARN) /nologo $(OPT_SHARED) $(PGOFLAGS) $(CONAN_LDFLAGS)
```
Ahajha, line 30: Not sure if this section is still necessary, is this 1.x legacy? If so, I can remove from previous versions too.

recipes/mpdecimal/config.yml
```diff
@@ -1,4 +1,6 @@
 versions:
+  "4.0.0":
+    folder: "2.5.x"
```
perseoGI, line 3: Looking good! But it seems a little bit weird to have the `4.x.x` versions inside the `2.5.x` folder...
Maybe if changes aren't that big, we could consider renaming the `2.5.x` folder to `all`

perseoGI: Do you dare to do it?!

Ahajha: Ah, yes I can :)

Ahajha: Done!

recipes/mpdecimal/2.5.x/patches/4.0.0-0001-msvc-fixes.patch
Ahajha: I'm removing these sections in my followup PR

recipes/mpdecimal/2.5.x/conanfile.py
Ahajha: I haven't tested yet, but I'm hoping the weirdness will go away in the followup cleanup PR.

Ahajha:
(I deleted my comment - bad rebase, in case anyone following along at home is confused :P)

What usually happens is that I develop on any of 4 different machines, and when I forget to pull on the one I'm using lots of bad things happen. At least I was able to salvage it this time :)

Ahajha:
@perseoGI Can you look at this PR again when you get a chance? Thanks!

AbrilRBS (approved):

perseoGI (approved):

Ahajha:
@AbrilRBS Can this be merged?

Ahajha:
Thanks @AbrilRBS!

PR #25900, opened 2024-11-09
libraw: Bump lcms dependency to 2.16

irieger:
### Summary
Changes to recipe:  **libraw/all**

#### Motivation
I'm currently preparing OpenColorIO 3.0.0 (#25672) which has libraw as an optional dependency but also uses opencolorio and libjxl, both depending on lcms. (libjxl depends on 2.16, opencolorio on 2.14 (see #25899). This MR would ensure that all dependencies use the same version.

#### Details
Only bumping the lcms dependency for compatiblity with the other openimageio requirements.

---

### DIFF
--- a/recipes/libraw/all/conanfile.py
+++ b/recipes/libraw/all/conanfile.py
@@ -62,7 +62,7 @@ class LibRawConan(ConanFile):
         elif self.options.with_jpeg == "mozjpeg":
             self.requires("mozjpeg/4.1.3")
         if self.options.with_lcms:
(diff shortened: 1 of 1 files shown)


jcar87:
Thanks @irieger  - I'm assuming this is solving a conflict, correct? (or at least a conflict that would arise in OpenColorIO 3.0.0)

irieger:
Thx, that is fine. Will now try if I get openimageio/3.0.0 finally to fully build

PR #25903, opened 2024-11-10
mpdecimal: Fix cxx+shared, remove unnecessary automake dependency

### LINKED ISSUE #25803: [package] mpdecimal/2.5.1: Extra dependency on autoconf and msys2 for MSVC builds
### Description

The package sets dependence on "Autoconf" and "Msys2" for MSVC, although only NMAKE is used during build. 

The function that determines the dependence at the current moment:
```py
    def build_requirements(self):
        if is_msvc(self):
            self.tool_requires("automake/1.16.5")
        else:
            # required to support windows as a build machine
            if self._settings_build.os == "Windows":
                self.win_bash = True
                if not self.conf.get("tools.microsoft.bash:path", check_type=str):
```

But can be perfectly simplified to:
```py
    def build_requirements(self):
        # required to support windows as a build machine
        if not is_msvc(self) and self._settings_build.os == "Windows":
            self.win_bash = True
            if not self.conf.get("tools.microsoft.bash:path", check_type=str):
```

With this configuration, everything builds without any problem and extra large dependencies.

* Package Name/Version: **mpdecimal/2.5.1**
* Operating System+version: **Windows 11**
* Compiler+version: **Visual Studio 2022**
* Conan version: **conan 1.65.0**
* Python version: **Python 3.9.13**

### Conan profile

[settings]
os=Windows
os_build=Windows
arch=x86_64
arch_build=x86_64
compiler=Visual Studio
compiler.runtime=MDd
compiler.version=17
compiler.toolset=v143
build_type=Debug
[options]
[build_requires]
[env]

### Steps to reproduce

conan info -b * mpdecimal/2.5.1@

### Logs

```
mpdecimal/2.5.1: Forced build from source
m4/1.4.19: Forced build from source
autoconf/2.72: Forced build from source
automake/1.16.5: Forced build from source
msys2/cci.latest: Forced build from source
msys2/cci.latest, m4/1.4.19, autoconf/2.72, automake/1.16.5, mpdecimal/2.5.1
```

</details>

Ahajha:
This PR has 2 small but important fixes:

- Remove the invalid configuration when both `mpdecimal/*:shared=True` and `mpdecimal/*:cxx=True`. This is a huge deal for the CPython recipe, since this will allow the Windows/all shared build to be valid again, and build in CI.
- Fixes #25803

I also have two other mpdecimal PRs out: #25889 (subset of this one) #25896


jcar87:
> I'm happy to revert any of these additional changes if it makes reviewing easier.

yes please!

Ahajha:
@jcar87 I removed most of the unnecessary stuff - I kept the part about removing the automake dependency. This PR is a superset of #25889, so feel free to skip that one or merge that one first.

recipes/mpdecimal/2.5.x/conanfile.py
```diff
@@ -145,13 +133,16 @@ def _build_msvc(self):
...
             copy(self, f"libmpdec-{self.version}.dll.lib", libmpdec_folder, dist_folder)
         else:
             copy(self, f"libmpdec-{self.version}.lib", libmpdec_folder, dist_folder)
         if self.options.cxx:
+            if self.options.shared:
+                copy(self, f"libmpdec++-{self.version}.dll", libmpdecpp_folder, dist_folder)
+                copy(self, f"libmpdec++-{self.version}.dll.lib", libmpdecpp_folder, dist_folder)
+            else:
+                copy(self, f"libmpdec++-{self.version}.lib", libmpdecpp_folder, dist_folder)
             copy(self, "decimal.hh", libmpdecpp_folder, dist_folder)
```
Ahajha, line 145: This packaging step as a whole is a mess, I'm going to clean it up in a future PR.

Currently, it copies from the source folder (mpdecimal does an in-source build) to somewhere in the build folder. It _then_ gets copied to the package folder. But the `.exp` files don't get copied the second time, so I omitted those. (I'm not entirely certain if they're generated at all).

This whole build process mimics some of mpdecimal's documentation, but that is specifically or packaging "their way", so we don't need to do that.

jcar87: Yes,  I noticed this :D 

Recommendation for `in-source` builds would be:

```
    def layout(self):
        self.folders.build = "src"
```

then in the `package()` method just copy the library artifacts from `self.build_folder` _directly_ - without that intermediate step. 
The .exp files are indeed generated

jcar87: Yup that's preferable, thanks!

recipes/mpdecimal/2.5.x/conanfile.py
```diff
@@ -57,20 +53,16 @@ def validate(self):
...
-        if is_msvc(self):
-            self.tool_requires("automake/1.16.5")
-        else:
-            # required to support windows as a build machine
-            if self._settings_build.os == "Windows":
-                self.win_bash = True
-                if not self.conf.get("tools.microsoft.bash:path", check_type=str):
-                    self.tool_requires("msys2/cci.latest")
+        if not is_msvc(self) and self.settings_build.os == "Windows":
+            self.win_bash = True
```
jcar87, line 63: given that our CI only covers win+msvc, I suppose this is an untested branch now.
Is this to support mingw?

Chances are that we need to detect clang and check if it it needs to fall outside of this conditional as well

Ahajha: I believe I tested this branch with Mingw and it worked. I can play around with it, but regardless of if it works or not this should still be a strict improvement, because the only change is just removing the automake dependency, which definitely was not used.

Ahajha: I'm actually not very familiar with what `msys2` is used for in general. Is it just whenever we need access to a unix-style shell in a Windows build? So if we're calling commands like `ls`, `sed`, etc. (things that aren't available in Windows)?

Ahajha: I tested mingw and clang. The former works, the latter I don't know how to get working, but I think that's beyond the scope of this PR. I'm certain that it didn't work before either.

Ahajha: @jcar87 Can you take another look at this please?

jcar87: Thanks for testing this out in mingw!!

jcar87 (approved):
LGTM!

czoido (approved):

PR #25937, opened 2024-11-13
Drop all non-migrated deprecated recipes

valgur:
Just a suggestion. These recipes have not been migrated, likely won't be, and are not usable with the new Conan 2-only repo.

Removed recipes:
- at-spi2-atk
- at-spi2-core/all
- b2/standard
- co
- cpp-taskflow
- diligentgraphics-spirv-headers
- diligentgraphics-spirv-tools
- diligentgraphics-vulkan-headers
- flatc
- google-cloud-cpp/all
- jsonformoderncpp
- msgpack
- mysql-connector-c
- tbb


AbrilRBS:
I think removing them is counterproductive, see the recent gcloud migration. Thanks for trying to clean up, but I think keeping them is better :)

valgur:
@AbrilRBS Ok, but what about recipes that 100% will not be updated or usable, like jsonformoderncpp?

PR #25939, opened 2024-11-14
Use lowercase advapi32 to avoid linker error on linux with mingw

pkoshevoy:
### Summary
Changes to recipe:  **boost/1.80.0**

#### Motivation
linker error occurs when cross-compiling a project on linux with mingw and conan boost dependency:
/usr/bin/x86_64-w64-mingw32-ld: cannot find -lAdvapi32: No such file or directory

#### Details
File names are generally case sensitive on linux.
Boost recipe lists Advapi32 as one of the system_libs, but the file on disk is actually all lower case:
/usr/x86_64-w64-mingw32/lib/libadvapi32.a

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -2046,7 +2046,7 @@ class BoostConan(ConanFile):
 
             if not self.options.get_safe("without_process"):
                 if self.settings.os == "Windows":
-                    self.cpp_info.components["process"].system_libs.extend(["ntdll", "shell32", "Advapi32", "user32"])
+                    self.cpp_info.components["process"].system_libs.extend(["ntdll", "shell32", "advapi32", "user32"])
                 if self._shared:
                     self.cpp_info.components["process"].defines.append("BOOST_PROCESS_DYN_LINK")
 


jcar87:
Hi @pkoshevoy - thanks for this PR!

This looks good - is this happening when trying to build/consume boost from a Linux OS, with a mingw toolchain to target Windows?
Just double checking to try and reproduce this locally

pkoshevoy:
> Hi @pkoshevoy - thanks for this PR!
> 
> This looks good - is this happening when trying to build/consume boost from a Linux OS, with a mingw toolchain to target Windows? Just double checking to try and reproduce this locally

Yes, this was breaking a gitlab ci/cd pipeline that was building a windows binary via a native mingw toolchain on Ubuntu 24.04.  My workaround in absence of a recipe fix was to symlink libadvapi32.a to libAdvapi32.a within the docker image, but it would be much cleaner to just fix the recipe.

jcar87:
> just FYI: microsoft documentation has "Advapi32.lib"

The file as provided by the Windows SDK is actually called `AdvAPI32.lib` - but It's a moot point on Windows since the filesystem is expected to be case insensitive.

All other references to `advapi32` as a system library in Conan Center are actually all lower case, with the Boost recipe being the one exception - so I think the proposed one will work with the linkers in both platforms, and there's precedence of it working for other recipes.

PR #25947, opened 2024-11-14
OpenCV/4.x: align JPEG codec option with project upstream defaults

### LINKED ISSUE #25946: [package] opencv/4.10.0: use `libjpeg-turbo` instead of `libjpeg` per default
### Description

Since OpenCV version 3.4.2, `libjpeg-turbo` is used as the default library for JPEG support instead of `libjpeg`. The support for `libjpeg` has been deprecated (see https://github.com/opencv/opencv/tree/4.x/3rdparty#readme).

In PR #24814 the default values in the conanfile were set such that they match the defaults from OpenCV. However, the `"with_jpeg"` option was left unchanged.

The default option should be set to `libjpeg-turbo`.

* Package Name/Version: **opencv/4.10.0**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

### Steps to reproduce

conan install --requires=opencv/4.10.0

### Logs

```
[...]
Requirements
[...]
    libiconv/1.17#73fefc1b696e069df90fd1d18aa63edd - Downloaded (conancenter)
    libjpeg/9e#68269859e4325ddc3f995d1fd3fc9187 - Cache
    libmp3lame/3.100#44b12d19316eb2b223d98d3e75dae438 - Downloaded (conancenter)
[...]
```

</details>

danteu:
### Summary
Changes to recipe:  **OpenCV/4.x:**

#### Motivation
`libjpeg-turbo` is used as the default JPEG codec library since OpenCV 3.4.2. This commit aligns the default set in the conanfile.
https://github.com/opencv/opencv/tree/4.x/3rdparty#readme

#### Details
Set the default option for `"with_jpeg"` to `"libjpeg-turbo"` for OpenCV/4.x.

Closes #25946

---

### DIFF
--- a/recipes/opencv/4.x/conanfile.py
+++ b/recipes/opencv/4.x/conanfile.py
@@ -188,7 +188,7 @@ class OpenCVConan(ConanFile):
         "with_wayland": True,
         # imgcodecs module options
         "with_avif": False,
-        "with_jpeg": "libjpeg",
+        "with_jpeg": "libjpeg-turbo",
         "with_png": True,
         "with_tiff": True,
         "with_jpeg2000": "openjpeg",


jcar87:
Hi @danteu - thanks for opening this PR and bringing this to our attention.

Unfortunately I don't think this is a change that we can currently accept. libjpeg is a dependency of a lot of things in Conan Center, so we need to ensure that the default libjpeg variant use is coherent across all recipes with their default options.

The default options of a library are a guidance for us, but not a hard requirement, as our priority is prioritising, wherever we can, graph coherence - otherwise enabling this would cause certain subgraphs to require libjpeg-turbo and conflict with libjpeg, or bring about other graph conflicts with different versions of libjpeg-turbo.

What we want to do, and we will look into this next year, is to make `libjpeg-turbo` the default implementation across all of Conan Center.

At the moment, if you wish to swap the implementation without altering recipes (or recipe defaults), I would consider using the `replace_requires` conf in a profile:

https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires

```
[replace_requires]
libjpeg/*: libjpeg-turbo/3.0.4
```

This should cause the entire graph to choose libjpeg-turbo to satisfy the requirement wherever libjpeg is required. Please let us know if this works, otherwise keep an eye out for when we do switch the default across the entire repo!

danteu:
Hi @jcar87, thank you for your quick and helpful reply; I appreciate it very much!

We were wondering why `libjpeg` is still the default for this recipe and were suspecting an oversight - turns out there is a good reason. Is there, by chance, a Github issue where the migration to `libjpeg-turbo` can be tracked?

Thanks again for your response!

PR #25960, opened 2024-11-15
aws-sdk-cpp: fix an ignored validate_build()

valgur:
`validate_build()` is defined twice, causing the first one to be ignored.


ericLemanissier:
That would have been caught by pylint `recipes/aws-sdk-cpp/all/conanfile.py:557:4: E0102: method already defined line 553 (function-redefined)`. Is this check part of the current linters ? Where can we find the sources of the linter by the way ?

jcar87 (approved):

jcar87:
Thanks!

jcar87:
>  Is this check part of the current linters ? Where can we find the sources of the linter by the way ?

the check is part of the new linter. It will be made available at a later date, TBD.

franramirez688 (approved):

PR #25961, opened 2024-11-15
Use `sqlite3/[>=3.45.0 <4]` for compatibility with Qt recipes

valgur:
Would be nice to get rid of a major source of version conflicts on CCI.

Both Qt recipes already use the mentioned version range, so I assume it's considered safe?
https://github.com/conan-io/conan-center-index/blob/a7af27e0b555eadb9011747d5078294f6815ab59/recipes/qt/6.x.x/conanfile.py#L394

### DIFF
--- a/docs/adding_packages/dependencies.md
+++ b/docs/adding_packages/dependencies.md
@@ -74,6 +74,7 @@ version range only when a requirement for a newer version is needed.
 * meson: `[>=1.2.3 <2]`
 * pkgconf: `[>=2.2 <3]`
 * xz_utils: `[>=5.4.5 <6]`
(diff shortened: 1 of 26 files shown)


jcar87:
this PR is fine, but closing this as running this on CI would result in a very large amount of revisions and packages to be published.

Will be aiming for this in a more staged manner - at the moment the goal is to reduce the amount of versions simultaneously maintained in the sqlite3 recipe.

The version range in the Qt recipe should help _prevent_ version conflicts depending on what the Conan solver sees first - I can't find any reports of users having version conflicts with Qt / sqlite3 - but if there are, let us know sot hat we can address those.

PR #25964, opened 2024-11-16
Flac: don't build programs for Apple non-macOS

kambala-decapitator:
### Summary
Changes to recipe:  **flac/all**

#### Motivation

Currently Flac can't be built for iOS, disabling this option fixes that

#### Details

When building for iOS, Flac fails at cmake configuration step with an error:

```
CMake Error at src/flac/CMakeLists.txt:23 (install):
  install TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
  target "flacapp".


CMake Error at src/metaflac/CMakeLists.txt:16 (install):
  install TARGETS given no BUNDLE DESTINATION for MACOSX_BUNDLE executable
  target "metaflac".
```

The PR disables building supplemental programs for Apple non-macOS.

---

### DIFF
--- a/recipes/flac/all/conanfile.py
+++ b/recipes/flac/all/conanfile.py
@@ -20,10 +20,12 @@ class FlacConan(ConanFile):
     options = {
         "shared": [True, False],
         "fPIC": [True, False],
+        "with_programs": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
+        "with_programs": True,
     }
 
     def export_sources(self):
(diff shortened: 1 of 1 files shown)


kambala-decapitator:
@jcar87 thank you for the feedback!

> Either disable building the programs when on iOS/tvOS/etc (is_apple_os_(self) and self.settings.os != "Macos" ) 

went with this approach

> It's an interesting thing, CMake itself sets all add_executable to be bundles by default when not on macOS 

it makes complete sense because Apple non-macOS can't really run plain executables (unless you use jailbroken device)

jcar87 (approved):

uilianries (approved):
LGTM. Situation clarified.

PR #25966, opened 2024-11-17
frugen: new recipe for frugen and libfru v2.0

alatarum:
### Summary
New recipe:  **frugen/2.0**

#### Motivation
Frugen is a small but useful tool to work with IPMI FRU data widely used in telecom industry to encode hardware inventory information.

#### Details
This adds a new recipe. The recipe mostly focus on `libfru` which is part of frugen and can be used in other projects to encode/decode FRU information but also deploys `frugen` tool itself. 
Includes patch to support macOS: https://codeberg.org/IPMITool/frugen/pulls/8

---
**tested with Conan version 2.7.1**

### DIFF
--- a/recipes/frugen/all/conanfile.py
+++ b/recipes/frugen/all/conanfile.py
@@ -5,6 +5,8 @@ from conan.tools.files import get, rmdir, apply_conandata_patches, export_conand
 from conan.tools.microsoft import is_msvc
 import os
 
+required_conan_version = ">=2.1"
+
 class FrugenConan(ConanFile):
     name = "frugen"
     description = "IPMI FRU Information generator / editor tool and library"
@@ -28,7 +30,7 @@ class FrugenConan(ConanFile):
 
     def validate(self):
         if is_msvc(self):
-            raise ConanInvalidConfiguration("Unsupported compiler. This package currently does not support MSVC")
+            raise ConanInvalidConfiguration("MSVC not supported by the library")
 
     def export_sources(self):
         export_conandata_patches(self)
@@ -58,6 +60,8 @@ class FrugenConan(ConanFile):
             tc.cache_variables["BINARY_STATIC"] = not self.options.shared
         tc.cache_variables["BINARY_32BIT"] = False
         tc.cache_variables["DEBUG_OUTPUT"] = False
+        # Dont let CMake find doxygen and try to build documentation
+        tc.cache_variables["CMAKE_DISABLE_FIND_PACKAGE_Doxygen"] = True
         tc.generate()
         deps = CMakeDeps(self)
         deps.generate()
--- a/recipes/frugen/all/conanfile.py
+++ b/recipes/frugen/all/conanfile.py
@@ -1,0 +1,79 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import get, rmdir, apply_conandata_patches, export_conandata_patches
+from conan.tools.microsoft import is_msvc
+import os
+
+class FrugenConan(ConanFile):
+    name = "frugen"
+    description = "IPMI FRU Information generator / editor tool and library"
+    license = ("Apache-2.0", "GPL-2.0-or-later")
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://codeberg.org/IPMITool/frugen"
+    topics = ("hardware", "ipmi", "fru")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+        "with_json": [True, False]
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_json": True
+    }
+    implements = ["auto_shared_fpic"]
+
+    def validate(self):
+        if is_msvc(self):
+            raise ConanInvalidConfiguration("Unsupported compiler. This package currently does not support MSVC")
+
+    def export_sources(self):
+        export_conandata_patches(self)
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        if self.options.with_json:
+            self.requires("json-c/0.18")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIB"] = self.options.shared
+        tc.cache_variables["ENABLE_JSON"] = self.options.with_json
+        if self.settings.os != "Macos":
+            tc.cache_variables["BINARY_STATIC"] = not self.options.shared
+        tc.cache_variables["BINARY_32BIT"] = False
+        tc.cache_variables["DEBUG_OUTPUT"] = False
+        tc.generate()
+        deps = CMakeDeps(self)
+        deps.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
+
+    def package_info(self):
+        if self.options.shared:
+            self.cpp_info.components["fru-shared"].libs = ["fru"]
+        else:
+            self.cpp_info.components["fru-static"].libs = ["fru"]
--- a/recipes/frugen/all/patches/0003-let-conan-choose-optimization-flags.patch
+++ b/recipes/frugen/all/patches/0003-let-conan-choose-optimization-flags.patch
@@ -1,0 +1,15 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 763c0b3..0ce32df 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -39,8 +39,8 @@ option(ENABLE_JSON "enable JSON support" ON)
+ option(JSON_STATIC "link json-c library statically" OFF)
+ option(DEBUG_OUTPUT "show extra debug output" OFF)
+ 
+-set(CMAKE_C_FLAGS_RELEASE "-Os -Wall -Werror -Wfatal-errors")
+-set(CMAKE_C_FLAGS_DEBUG "-g3 -O0 -Wall -Werror -Wfatal-errors")
++# set(CMAKE_C_FLAGS_RELEASE "-Os -Wall -Werror -Wfatal-errors")
++# set(CMAKE_C_FLAGS_DEBUG "-g3 -O0 -Wall -Werror -Wfatal-errors")
+ if(MSVC)
+ 	# warning level 4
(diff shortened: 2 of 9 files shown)


recipes/frugen/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
+from conan import ConanFile
```
jcar87, line 1:
```suggestion
from conan import ConanFile
from conan.errors import ConanInvalidConfiguration
```

recipes/frugen/all/conanfile.py
```diff
@@ -0,0 +1,79 @@
...
+        apply_conandata_patches(self)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_SHARED_LIB"] = self.options.shared
+        tc.variables["ENABLE_JSON"] = self.options.with_json
+        tc.variables["BINARY_STATIC"] = not self.options.shared
+        tc.variables["BINARY_32BIT"] = False
+        tc.variables["JSON_STATIC"] = False
+        tc.variables["DEBUG_OUTPUT"] = False
```
jcar87, line 60:
```suggestion
        tc.cache_variables["BUILD_SHARED_LIB"] = self.options.shared
        tc.cache_variables["ENABLE_JSON"] = self.options.with_json
        tc.cache_variables["BINARY_STATIC"] = not self.options.shared
        tc.cache_variables["BINARY_32BIT"] = False
        tc.cache_variables["JSON_STATIC"] = False
        tc.cache_variables["DEBUG_OUTPUT"] = False
```

recipes/frugen/all/test_package/conanfile.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+
+class TestPackageConan(ConanFile):
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False]
+    }
+    default_options = {
+        "shared": False
+    }
```
jcar87, line 13:
```suggestion
```

recipes/frugen/all/test_package/conanfile.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def layout(self):
+        cmake_layout(self)
+
+    def requirements(self):
+        self.requires(self.tested_reference_str)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_SHARED_LIB"] = self.options.shared
```
jcar87, line 23:
```suggestion
        tc.variables["BUILD_SHARED_LIB"] = self.dependencies['frugen'].options.shared
```

recipes/frugen/all/conanfile.py
alatarum: done

alatarum: Now it passes build, thank you!
But I still don't understand, why it did not work with windows runner, will try again.

recipes/frugen/all/conanfile.py
alatarum: done

recipes/frugen/all/test_package/conanfile.py
alatarum: done

recipes/frugen/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        "with_json": [True, False]
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_json": True
+    }
+    implements = ["auto_shared_fpic"]
+
+    exports_sources = "CMakeLists.txt", "*.c", "*.h"
```
AbrilRBS, line 28:
```suggestion
```

The recipe is already brining the sources from the release

recipes/frugen/all/conanfile.py
```diff
@@ -0,0 +1,80 @@
...
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.cache_variables["BUILD_SHARED_LIB"] = self.options.shared
+        tc.cache_variables["ENABLE_JSON"] = self.options.with_json
+        tc.cache_variables["BINARY_STATIC"] = not self.options.shared
+        tc.cache_variables["BINARY_32BIT"] = False
+        tc.cache_variables["JSON_STATIC"] = False
```
AbrilRBS, line 60: This is a bit confusing, I'd remove this line, as the patch is basically overriding the upstream behaviour and making this flag useless
```suggestion
```

AbrilRBS (commented):
Thanks! A few minor extra changes, otherwise looks great :)

recipes/frugen/all/conanfile.py
alatarum: good point, will fix

recipes/frugen/all/conanfile.py
alatarum: Actually, I just have no clear understanding, what it suppose to do )

recipes/frugen/all/conanfile.py
```diff
@@ -0,0 +1,77 @@
...
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+        "with_json": True
+    }
+    implements = ["auto_shared_fpic"]
+
+    def validate(self):
+        if self.settings.os in ["Windows", "Macos"]:
+            raise ConanInvalidConfiguration("Unsupported os. This package currently does not support Macos and Windows")
```
jcar87, line 30: I think windows _is_ supported (as per the documentation)  just not with MSVC + Windows SDK -  please try to avoid invalidating platforms just to get CI to pass - it gives us an opportunity to look at what's wrong and advise

alatarum: I bet no one tested latest release under Windows, even with MinGW =). But you right, it should work. What can I refer as an example on how to do this better?

jcar87: Thanks @alatarum !! What I mean is that - invalidating is "probably" fine - but as reviewers we need to see why it is being invalidated. Sometimes it is 100% true that indeed it is an "unsupported OS" - this can be evidenced by going to the documentation, or similar - but other times, it fails for other reasons (could be something related to our CI infrastructure, or something else) - and more often than not we end up with a PR a few days/weeks/months down the line removing a "too strict" invalidation :/ 

What we need to see is the supporting evidence behind the "unsupported os" - so that we can advise to tailor what needs to be invalidated, for example
- "doesn't work on windows with msvc because it requires sys/time.h"
- "doesnt work on macOS because it includes the wrong header and there is little evidence that it was ever used or tested on macOS"

That leaves a "paper trail" of sorts, for us to see, and for everyone else in the future when they need to reason why something was done :) 

And hopefully to avoid misunderstanding - we don't expect contributors to test or try to fix platforms where CI is failing  - I appreciate your efforts on this front on macOS! Sometimes patches are trivial - but a lot of times they aren't, and they lead us to believe that certain variants of libraries would _only_ exist in Conan, and that's not our goal - quite the opposite in fact, as that would also mean that we would we providing something that was never fully tested _outside_ of the build process of Conan Center, which is quite a risk!

jcar87 (approved):
Thanks @alatarum!!

Approving patch to support macoS https://codeberg.org/IPMITool/frugen/pulls/8 - as you have mentioned this will be supported in the next release - we're assuming based on the commit history that you are either a maintainer or involved in that project - thanks!

recipes/frugen/all/patches/0003-let-conan-choose-optimization-flags.patch
```diff
@@ -0,0 +1,15 @@
...
+index 763c0b3..0ce32df 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -39,8 +39,8 @@ option(ENABLE_JSON "enable JSON support" ON)
+ option(JSON_STATIC "link json-c library statically" OFF)
+ option(DEBUG_OUTPUT "show extra debug output" OFF)
+ 
+-set(CMAKE_C_FLAGS_RELEASE "-Os -Wall -Werror -Wfatal-errors")
+-set(CMAKE_C_FLAGS_DEBUG "-g3 -O0 -Wall -Werror -Wfatal-errors")
++# set(CMAKE_C_FLAGS_RELEASE "-Os -Wall -Werror -Wfatal-errors")
```
AbrilRBS, line 11: @alatarum I removed this as Conan usually injects these flags based on the build_type provided in the profile, so that users can better select their desired optimization. Is -Os in this case critical?

alatarum: Thanks for testing and improving!
I almost sure that -Os is just a personal preference of the author and not actually needed. I will look is we should remove this lines in upstream also.

BTW is it possible to ask conan to provide PkgConfig files for json-c when this recipe pulls it as a dependency? Then first patch not needed.

alatarum:
> Thanks @alatarum!!
> 
> Approving patch to support macoS https://codeberg.org/IPMITool/frugen/pulls/8 - as you have mentioned this will be supported in the next release - we're assuming based on the commit history that you are either a maintainer or involved in that project - thanks!

Yes, not a maintainer, but contributing there.

recipes/frugen/all/patches/0003-let-conan-choose-optimization-flags.patch
AbrilRBS: using `PkgConfigDeps` class would allow that to work, but trying locally, then I get some include errors for other files, which confuses me a bit:

You can try locally with:

```diff
diff --git a/recipes/frugen/all/conandata.yml b/recipes/frugen/all/conandata.yml
index cc18fdc79..1474da704 100644
--- a/recipes/frugen/all/conandata.yml
+++ b/recipes/frugen/all/conandata.yml
@@ -4,9 +4,6 @@ sources:
     sha256: "cb5c59cf3593b162b9faedf1e5ae3b1f7e32fb732a61ada6baf7de69632bff77"
 patches:
   "2.0":
-    - patch_file: "patches/0001-cmake-use-find_package-to-locate-json-c-library.patch"
-      patch_description: "Upstream prefer to use pkg_check_modules instead of find_package to keep capatibility with old versions"
-      patch_type: "conan"
     - patch_file: "patches/0002-libfru-fix-build-on-MacOS.patch"
       patch_description: "Fixes build on MacOS"
       patch_type: "portability"
diff --git a/recipes/frugen/all/conanfile.py b/recipes/frugen/all/conanfile.py
index 17b6eeec9..b2becc771 100644
--- a/recipes/frugen/all/conanfile.py
+++ b/recipes/frugen/all/conanfile.py
@@ -1,6 +1,7 @@
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.gnu import PkgConfigDeps
 from conan.tools.files import get, rmdir, apply_conandata_patches, export_conandata_patches
 from conan.tools.microsoft import is_msvc
 import os
@@ -65,6 +66,8 @@ class FrugenConan(ConanFile):
         tc.generate()
         deps = CMakeDeps(self)
         deps.generate()
+        pkgdeps = PkgConfigDeps(self)
+        pkgdeps.generate()
 
     def build(self):
         cmake = CMake(self)
```

alatarum: Hmm, I'm getting linking error:
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -ljson-c: No such file or directory

But let's keep current implementation

AbrilRBS: We can go for a follow-up PR once we find how to properly use it there, but for now this looks good!

AbrilRBS (approved):

PR #25990, opened 2024-11-21
multiple recipes: align dependency on libdeflate

kambala-decapitator:
### Summary
Changes to recipe:  **libtiff/all**

#### Motivation

libdeflate < v1.20 can't be built for iOS with apple-clang 15, upstream added a workaround only in v1.20, see https://github.com/ebiggers/libdeflate/issues/354

#### Details

Now using the latest available version of libdeflate

---

### DIFF
--- a/recipes/gdal/pre_3.5.0/conanfile.py
+++ b/recipes/gdal/pre_3.5.0/conanfile.py
@@ -208,7 +208,7 @@ class GdalConan(ConanFile):
         if self.options.get_safe("with_zlib", True):
             self.requires("zlib/[>=1.2.11 <2]")
         if self.options.with_libdeflate:
-            self.requires("libdeflate/1.19")
+            self.requires("libdeflate/1.22")
         if self.options.with_libiconv:
             self.requires("libiconv/1.17")
         if self.options.get_safe("with_zstd"):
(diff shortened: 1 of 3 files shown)


AbrilRBS:
Hi @kambala-decapitator thanks a lot for this. I've noticed that this would create a few conflicts downstream, so here is the current plan:

- Bring all the libdefltate dependencies in-line in [`7bc254a` (#25990)](https://github.com/conan-io/conan-center-index/pull/25990/commits/7bc254a4f45f1b46d876616bc11492e8837540a7)
- This would create missing binaries if merged, so we're halting the PR until we're ready to launch the necessary jobs to fill those gaps

In the meantime, if you or anyone else needs this dependency to be used graph-wide, please take a look at the `[replace_requires]` section of the profiles https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires, with something like

```
[replace_requires]
libdeflate/*: libdeflate/1.22
```

kambala-decapitator:
@AbrilRBS thanks a lot!

Do I understand correctly that I don't have anything else to do in this PR and now it's entirely on the CCI team?

AbrilRBS:
Exactly @kambala-decapitator, sorry I didnt make it clear enough at first!

kambala-decapitator:
@AbrilRBS anything blocking this?

jcar87 (approved):

jcar87:
Have done this across multiple PRs - but with this PR the version should be aligned - some binaries were not built by CI due to: binaries missing for _older_ versions of gdal (assuming those older versions are not used by current conan2 compatible recipes), and where hdf5 is required for x86_64 on macOS (because it won't crossbuild yet)

PR #25991, opened 2024-11-21
sdl_image: add v2.8.2

kambala-decapitator:
### Summary
Changes to recipe:  **sdl_image/2.8.2**

#### Motivation

The latest available 2.x

#### Details

Tested building for macOS and iOS using apple-clang 15.

CI had an issue when building Linux, so I added a patch with workaround and also opened a PR in upstream: https://github.com/libsdl-org/SDL_image/pull/479

---


recipes/sdl_image/all/conanfile.py
```diff
@@ -169,7 +170,11 @@ def generate(self):
         cd = CMakeDeps(self)
         cd.generate()
 
+    def _patch_sources(self):
+        apply_conandata_patches(self)
+
     def build(self):
+        self._patch_sources()
```
AbrilRBS, line 177: Nowadays we prefer this to to be done in the source method

kambala-decapitator: fixed, thanks

czoido (approved):

PR #25998, opened 2024-11-21
re2c: add version 4.0.2

toge:
### Summary
Changes to recipe:  **re2c/4.0.2**

#### Motivation
Major version up.

#### Details
https://re2c.org/releases/release_notes.html#release-4-0
https://github.com/skvadrik/re2c/compare/3.1...4.0.2

---


memsharded (commented):
Thanks for the contribution!

I was surprised to see an autotools build while checking the ``replace_in_file()`` in the diff, when the upstream project supports CMake and it seems pretty straightforward to use, so I am trying to update the recipe with it.

memsharded:
Ok, it breaks for version 1.3 from Dec 14, 2019, because that one didn't support CMake, will check with the team.

jcar87 (approved):

jcar87:
@memsharded  - keep in mind that given the recent issues we've experienced with any build system switch (where a recipe goes from being built with one build system to another), we are being *very* conservative. Approving this considering this is an executable (rather than a library), and the build system is CMake - but otherwise I'd advise that for any new version added, the build system of the recipe is retained unless there are very strong reasons or advantages (to the users) that justify it.

PR #26008, opened 2024-11-22
[dcmtk] Backport fixes for CVEs and data corruption bug

cschreib-ibex:
### Summary
Changes to recipe:  **dcmtk/3.6.7** and **dcmtk/3.6.8**

#### Motivation
This PR proposes to backport two sets of fixes from upstream:

 - Upstream ticket: https://support.dcmtk.org/redmine/issues/1114
   Fix commit: https://github.com/DCMTK/dcmtk/commit/c78e434c0c5f9d932874f0b17a8b4ce305ca01f5
   This fixes CVE-2024-34508 and CVE-2024-34509 for versions 3.6.7 and 3.6.8. This will be fixed upstream starting from 3.6.9 (not yet released).

 - Upstream ticket: https://support.dcmtk.org/redmine/issues/1143
   "Fix" commit: https://github.com/DCMTK/dcmtk/commit/8ccfd5a07024e50b160da0231524da535c745b79
   The issue is caused by a buggy feature, which was introduced in 3.6.8 and enabled by default. The commit above (to be released in 3.6.9) disables the feature by default, but doesn't fix the actual bug (that is still being investigated).

#### Details
The CVEs being patched cause a SEGFAULT (denial of service) when a malformed request is sent to a C-Store SCP. I was able to reproduce this on versions 3.6.7 and 3.6.8 on Linux platforms, but not on Windows. The fix commit was made in March this year, but no new release has been issued since, so all available versions on the Conan center are affected.

The other issue is caused by buggy code in the character set conversion code of the new `oficonv` implementation. This "pass-through" feature makes conversion a no-op when converting between identical source and target character sets (e.g. source is UTF-8 and target is also UTF-8). This was introduced in 3.6.8, when `oficonv` was added and enabled as default conversion implementation. Unfortunately this pass-through feature is buggy, and causes data corruption when the data to be converted is more than 1024 characters long. The fix commit (from January this year, to be released in 3.6.9) disables the pass-through feature by default and adds a toggle to re-enable it if needed. This was not done to fix the actual bug; the feature was actually unwanted, since it bypassed important data consistency checks. It is expected that a complete fix will follow, but I don't expect the pass-through feature to be enabled by default again.

None of these patches will need applying to the next version 3.6.9. An alternative solution is to wait for 3.6.9 to be released, and then remove 3.6.7 and 3.6.8 from the Conan center. I don't know when that next version will be released though; 3.6.8 is now almost a year old.

---


AbrilRBS (dismissed):
Thanks a lot for prividing context and traceability links to the diffs, we appreciate it

This looks good on my end - as this is a security fix I've asked more than 1 teammember to also review this, thanks a lot for your patience :)

franramirez688 (approved):

PR #23090, opened 2024-03-14
rocksdb/6.29.5: add patch to ignore thirdparty.inc

thejohnfreeman:
**rocksdb/6.29.5**

RocksDB has [special instructions](https://github.com/facebook/rocksdb/blob/v6.29.5/CMakeLists.txt#L11) for compiling on Windows that asks builders to edit the special file [`thirdparty.inc`](https://github.com/facebook/rocksdb/blob/v6.29.5/thirdparty.inc) to set variables pointing to all of the optional dependencies they want to enable. The main `CMakeLists.txt` follows the best practice of calling `find_package()` and linking to imported targets, but only for non-MSVC generators.

Thus, the latest RocksDB recipe in Conan Center builds with MSVC, but does not correctly link to any optional dependencies that are enabled. `thirdparty.inc` is included with nonsense paths and the build fails when it cannot find third-party headers.

The old RocksDB recipe in Conan Center used a CMake subproject model. It had its own [`CMakeLists.txt`](https://github.com/conan-io/conan-center-index/blob/1a1b7fc7959cdf8fc97cce745fe844e31194db29/recipes/rocksdb/all/CMakeLists.txt) that first called `conan_basic_setup()` before calling `add_subdirectory()` for the project. I don't know how, but it worked.

This patch removes the branch on MSVC that includes `thirdparty.inc` instead of calling `find_package()`, but only for version 6.29.5 (the one I need for now).

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #26009, opened 2024-11-22
rocksdb: patching thirdparty.inc. Modernized.

franramirez688:
ORIGINAL PR: https://github.com/conan-io/conan-center-index/pull/23090 ( I just added a commit to modernize the recipe but can not push directly to the original branch. Keeping the original author's commits anyway)

### Summary
Changes to recipe:  **rocksdb/x.x.x**

* Modernized recipe.
* Ignoring the `thirdparty.inc` file through more patches for all the versions.


AbrilRBS (approved):

thejohnfreeman:
Thank you for working on this! I see you changed `required_conan_version` to ">=2.0". Was that necessary or just recommended? My team is still largely using Conan 1.x. I _think_ we are ready to switch to 2.x, but would love to leave the 1.x option open if possible. For now, we keep a recipe for RocksDB, with the patch, in our own project, but I would love to get rid of that in favor of the version in Conan Center.

RazielXYZ:
@thejohnfreeman Conan Center is no longer accepting packages/recipe updates/anything for Conan 1.x, so even if the 1.x stuff is left in the recipe, it won't propagate to the 1.x CC

RazielXYZ (approved):

RazielXYZ:
@thejohnfreeman @franramirez688 if I'm reading rocksdb's CMakeLists right, setting ROCKSDB_SKIP_THIRDPARTY makes it so none of the third party libs are handled or included/linked on MSVC, only on the else() branch if the if(MSVC) check. The MSVC branch only handles libs through thirdparty.inc. So I think the patch is actually necessary for versions that have ROCKSDB_SKIP_THIRDPARTY as well.

RazielXYZ (changes requested):
Patch still needed even with ROCKSDB_SKIP_THIRDPARTY

thejohnfreeman:
That's correct. If `ROCKSDB_SKIP_THIRDPARTY` is passed, then _no_ third-party dependencies are searched when compiling with MSVC. We don't want that. We just want them searched with `find_package()`, with the additional handling that non-MSVC compilers enjoy.

franramirez688:
@thejohnfreeman @RazielXYZ Thanks a lot for your comments! I did not realize about that, my bad 🙏 
I'll be working on the new patch 😉 I'll keep you posted

AbrilRBS (approved):

ErniGH (approved):
LGTM

PR #26023, opened 2024-11-24
mbedtls Fix Shared build MSVC

luadebug:
### Summary
Changes to recipe:  **mbedtls**

#### Motivation
Gain Windows OS opportunity to build shared lib & link over it. Fix shared library DLL symbol links for MSVC.

#### Details
Fixed export of symbols for MSVC DLL build.

---


jcar87:
Thanks @luadebug  - unfortunately this is not something we can accept at the moment.

The `embedtls` library itself does *not* support being built as shared library on Windows: https://github.com/Mbed-TLS/mbedtls/issues/470, and using the CMake functionality to export all symbols also is known to cause issues: https://github.com/Mbed-TLS/mbedtls/issues/470#issuecomment-1046758271 - while this feature is used in other recipes, we no longer advocate its use and recipes should just mark the shared configuration on Windows as invalid if there is no existing support for shared library usage.

While it may be possible to fix this properly - it's best to submit those fixes to the upstream maintainers of mbedtls rather than Conan Center. Libraries in Conan Center as provided "as is".

PR #26028, opened 2024-11-25
simple_enum: add new version 0.8.11

toge:
### Summary
Changes to recipe:  **simple_enum/0.8.11**

#### Motivation
There are several improvements since 0.8.0.

#### Details
https://github.com/arturbac/simple_enum/compare/v0.8.0...v0.8.11

---


recipes/simple_enum/all/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "0.8.4":
+    url: "https://github.com/arturbac/simple_enum/archive/refs/tags/v0.8.4.tar.gz"
```
AbrilRBS, line 3: As far as I can see, the new version requires C++23 [arturbac/simple_enum@`v0.8.0...v0.8.4`#diff-1e7de1ae2d](https://github.com/arturbac/simple_enum/compare/v0.8.0...v0.8.4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR10)


Also, I wonder how the glaze requirement is being handled in this case, it seems like it depends on it unconditionally?

AbrilRBS (commented):
Thanks! Some questions about the new version and our dependency handling :)

recipes/simple_enum/all/conandata.yml
AbrilRBS: My reading from https://github.com/arturbac/simple_enum/blob/v0.8.4/CMakeLists.txt#L33 is that glaze would be needed if we were using the CMake installation (Plus a defines too?) Can you double check? Thanks!

toge: @AbrilRBS 
Thanks!
I didn't read CMakeLists.txt so much.
While I can use simple_enum without glaze, I will follow official CMakeLists.txt.

AbrilRBS: No worries. Should we ask upstream about it, see if they support this no-glaze-drpendency case?

AbrilRBS (dismissed):
As discussed in https://github.com/conan-io/conan-center-index/pull/26028#pullrequestreview-2458146730, the glaze requirement seems to be required

recipes/simple_enum/all/conanfile.py
```diff
@@ -72,3 +55,5 @@ def package(self):
     def package_info(self):
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
+
+        self.cpp_info.defines.append("SIMPLE_ENUM_GLZ_3_1_x")
```
jcar87, line 59: I may be missing something here, where does this come from?
I can't find references in https://github.dev/arturbac/simple_enum/tree/v0.8.11

recipes/simple_enum/all/conanfile.py
```diff
@@ -21,35 +20,19 @@ class SimpleEnumConan(ConanFile):
...
-
-    def export_sources(self):
-        export_conandata_patches(self)
+        return 23 if Version(self.version) >= "0.8.4" else 17
 
     def layout(self):
         basic_layout(self, src_folder="src")
 
+    def requirements(self):
+        self.requires("glaze/4.0.1")
```
jcar87, line 29: the public headers of simple_enum dont make any references to glaze whatsoever - in the codebase it seems like its used only in examples and tests, but not the actual header library. it's documented as being opt-in by the end user, to include both, so I'm not sure if this is required here

jcar87 (commented):
some outstanding questions

jcar87:
Closing as stale as there were review comments pending to be addressed. Please reopen or open a new PR if this is still relevant, thanks!

PR #26037, opened 2024-11-26
openvino/2024.5.0 update

culhatsker:
### Summary
Changes to recipe:  **openvino/2024.5.0**

#### Motivation
New version

#### Details
Added JAX frontend as option

---
- [x] Tested locally with at least one configuration using a recent version of Conan (v2.6.0)


recipes/openvino/all/conanfile.py
```diff
@@ -62,7 +63,8 @@ class OpenvinoConan(ConanFile):
         "enable_tf_frontend": True,
         "enable_tf_lite_frontend": True,
         "enable_paddle_frontend": True,
-        "enable_pytorch_frontend": True
+        "enable_pytorch_frontend": True,
+        "enable_jax_frontend": True
```
uilianries, line 67: Please, add condition to remove enable_jax_frontend in case the built version is < 2024.3.0:

```python
def config_options(self):
    if Version(self.version) < "2024.3.0":
        del self.options.enable_jax_frontend
```

recipes/openvino/all/conanfile.py
```diff
@@ -226,6 +228,7 @@ def generate(self):
         toolchain.cache_variables["ENABLE_OV_TF_LITE_FRONTEND"] = self.options.enable_tf_lite_frontend
         toolchain.cache_variables["ENABLE_OV_ONNX_FRONTEND"] = self.options.enable_onnx_frontend
         toolchain.cache_variables["ENABLE_OV_PYTORCH_FRONTEND"] = self.options.enable_pytorch_frontend
+        toolchain.cache_variables["ENABLE_OV_JAX_FRONTEND"] = self.options.enable_jax_frontend
```
uilianries, line 231:
```suggestion
        if Version(self.version) >=  "2024.3.0":
            toolchain.cache_variables["ENABLE_OV_JAX_FRONTEND"] = self.options.enable_jax_frontend
```

This options does not exist before 2024.3.0: https://github.com/openvinotoolkit/openvino/compare/2024.2.0...2024.3.0#diff-dd9d470bfd7d5c3c875bd6f1c182ac653ef5fb312ae28c24b8ea39c5074e7c8e

recipes/openvino/all/conanfile.py
```diff
@@ -365,6 +368,8 @@ def package_info(self):
                 openvino_runtime.requires.append("protobuf::libprotobuf")
             if self.options.enable_pytorch_frontend:
                 openvino_runtime.libs.append("openvino_pytorch_frontend")
+            if self.options.enable_jax_frontend:
```
uilianries, line 371:
```suggestion
            if Version(self.version) >= "2024.3.0" and self.options.enable_jax_frontend:
```

uilianries (commented):
The option ENABLE_OV_JAX_FRONTEND is only exposed since 2024.3.0

uilianries:
I would suggest by removing old versions from config.yml and conandata.yml as well. OpenVino takes several hours to be built and the number of version is increasing. Old versions will be always available in Conan Center, only will stop receiving updates from the recipe.

uilianries:
@culhatsker It sound to me too. I just would like to know if there is any possible backport for older versions, like bringing features or bugfixes to old version already in CCI. Otherwise, I would support the idea of keeping the latest version only for now.

culhatsker:
> @culhatsker It sound to me too. I just would like to know if there is any possible backport for older versions, like bringing features or bugfixes to old version already in CCI. Otherwise, I would support the idea of keeping the latest version only for now.

I'll remove some older versions, but keep latest 2023.3.0 as it may be updated eventually.

uilianries:
@culhatsker Hello, thank you not only for checking those versions, but also for reducing the number!

The current error is related to the new option `openvino_jax_frontend`: https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-26037/2/package_build_logs/build_log_openvino_2024_5_0_2b72f2e10b5e263fe92c46034643984c_2e47e0b5ec010123a40a342396d465b3c1715aed.success.txt

I can found:

```
[ 58%] Linking CXX shared library /home/conan/workspace/cci_prod_PR-26037/conan-home/p/openvdce7b68dd69e5/s/src/bin/intel64/Release/libopenvino_jax_frontend.so
[ 58%] Building CXX object src/frontends/pytorch/src/CMakeFiles/openvino_pytorch_frontend.dir/op/arange.cpp.o
[ 58%] Built target openvino_jax_frontend
```

But the same is not listed when executed when running Cmake install. My guess is it's missing from install CMakeLists.txt, in OpenVino. As workaround, you could use `copy(self, "*openvino_jax_frontend.*", src=os.path.join(self.build_folder, "bin/intel64/Release/"), dst=os.path.join(self.package_folder, "lib"))`

Of course, would need to use a dynamic path, as the path is related to the build type and arch.

culhatsker:
@uilianries turns out this component behaves tricky, unlike other components and also makes sense only for our Python API, so I decided I will postpone implementing it for an undefined period of time.

uilianries:
@culhatsker Thank you for clarifying!

uilianries (approved):
LGTM.

culhatsker:
@uilianries may I ask you to merge it?

uilianries:
@culhatsker This PR still requires 1 extra review from the Conan team. It's marked already to be reviewed.

AbrilRBS (approved):

PR #26049, opened 2024-11-27
SDL: fix Android shared library and expose HIDAPI setting

kambala-decapitator:
### Summary
Changes to recipe:  **sdl/all**

#### Motivation

- fixes building Android shared library
- exposes SDL_HIDAPI setting

#### Details

Even though SDL is a C library, on Android it starts requiring C++ only when you build the HIDAPI subsystem, see here: https://github.com/libsdl-org/SDL/blob/bbc9c75618fbc60583123e03f31f4e856f2654bf/cmake/sdlchecks.cmake#L1105-L1106

Without the fix cmake picks the default C++ library from NDK toolchain which is static. This is an issue when your profile requests shared C++ library and could be a source of strange behavior later, like we faced in our project: https://github.com/vcmi/vcmi/issues/4973

---

### DIFF
--- a/recipes/sdl/all/conanfile.py
+++ b/recipes/sdl/all/conanfile.py
@@ -191,6 +191,7 @@ class SDLConan(ConanFile):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        self._patch_sources()
 
     def _patch_sources(self):
         apply_conandata_patches(self)
@@ -335,7 +336,6 @@ class SDLConan(ConanFile):
         tc.generate()
 
     def build(self):
-        self._patch_sources()
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
--- a/recipes/sdl/all/conanfile.py
+++ b/recipes/sdl/all/conanfile.py
@@ -139,8 +139,10 @@ class SDLConan(ConanFile):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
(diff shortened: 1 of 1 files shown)


recipes/sdl/all/conanfile.py
```diff
@@ -77,6 +78,7 @@ class SDLConan(ConanFile):
         "opengles": True,
         "vulkan": True,
         "libunwind": True,
+        "hidapi": True,
```
perseoGI, line 81: Do we need this `hidapi` option to be enabled by default. Adding new options by default may cause undesired behaviors on other users.
Which are the side effects of this option?
If your project needs it, wouldn't be safer to add it as a "False defaulted" option and enabled it on your side?

Thank you!

kambala-decapitator: it's on by default in SDL, so no change for existing users, see here: https://github.com/libsdl-org/SDL/blob/a5da9be5e64671c29595fe12132bbf105f71da05/CMakeLists.txt#L487C16-L487C22

perseoGI: Okey, I see!
LGTM then

perseoGI (approved):

AbrilRBS (approved):
Thank you @kambala-decapitator!

One extra thing somewhat related. In https://github.com/conan-io/conan-center-index/pull/25873 I'm trying to get SDL3 version working. If you ever port to the new version, having feddback there would be awesome, thanks!

Issue #26062, opened 2024-11-28
[package] libssh2/1.11.0: Requirement on cmake 3.30.5


DevAGCO:
### Description

With the latest version 1.11.1 a new tool depedency has appeared (to cmake 3.30.5).

It looks like previous versions of libssh2 (tested with 1.11.0 and 1.10.0) have suddenly also inherited from that new requirement. Breaking existing builds that are not able to build cmake (e.g. armv7 platform). 

### Package and Environment Details

* Package Name/Version: libssh2/1.11.0
* Operating System+version: Raspbian armv7
* Compiler+version: GCC 8.3.0
* Python version: Python 3.7.3

### Conan profile

[settings]
arch=armv7
build_type=Release
compiler=gcc
compiler.cppstd=gnu14
compiler.libcxx=libstdc++11
compiler.version=8
os=Linux

### Steps to reproduce

Simply trying to build the package

### Logs

<details><summary>Click to expand log</summary>

```
...
Build requirements
    cmake/3.30.5#540068cc37f1d21b42b729c82d6bcb29:7a99b07c654f17f335428a58b7f18cae9b625984 - Invalid
ERROR: There are invalid packages:
cmake/3.30.5: Invalid: CMake binaries are only provided for x86_64 and armv8 architectures
```

</details>

cf https://github.com/conan-io/conan-center-index/pull/25639

jcar87:
Hi @DevAGCO  - thanks for reporting this.

As a workaround, if you have a recent-enough CMake version installed on your system, could you try adding the following to your profile?

```
[platform_tool_requires]
cmake/[>=3.15 <4]
```

this should cause Conan to skip the cmake dependency altogether and assume it is installed system wide

AbrilRBS:
Hi @DevAGCO thanks a lot for your report. You're right that the dependency for CMake in this case was incorrectly added. I'll be rolling back this change alongside some improvements to the recipe, but for now @jcar87's suggestion might help solve the issues in the meantime :)

DevAGCO:
@AbrilRBS Great, thanks! :)

@jcar87 I tried this and other similar attempts to overrule the requirement coming from the libssh2 recipe. But did not manage to make it work (yet)...

DevAGCO:
Thanks for your care @jcar87.
`cmake version 3.24.4` (built from source)
`Conan version 2.6.0`
Raspberry Pi 4 Model B Rev 1.1
`uname -m` : `armv7l`
[...]

PR #26069, opened 2024-11-29
libssh2: Fix cmake dependency for old versions

AbrilRBS:
### Summary
Changes to recipe:  **libssh2/[*]**

#### Motivation

Closes https://github.com/conan-io/conan-center-index/issues/26062

#### Details

The CMake dependency was incorrecty added on a previous PR for old versions

### DIFF
--- a/recipes/libssh2/all/conanfile.py
+++ b/recipes/libssh2/all/conanfile.py
@@ -87,7 +87,7 @@ class Libssh2Conan(ConanFile):
         self.settings.rm_safe("compiler.cppstd")
 
     def requirements(self):
-        if self.options.with_zlib or self.options.crypto_backend == "openssl":
+        if self.options.with_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
         if self.options.crypto_backend == "openssl":
             self.requires("openssl/[>=1.1 <4]")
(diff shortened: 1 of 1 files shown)


recipes/libssh2/all/conanfile.py
```diff
@@ -97,9 +97,6 @@ def requirements(self):
             else:
                 self.requires("mbedtls/2.28.4")
 
-    def build_requirements(self):
-        self.tool_requires("cmake/[>=3.20 <4]")
```
AbrilRBS, line 101: As per @jcar87, this is in fact needed for 1.11 onwards for the upstream CMakelists to work with Conan-generated targets

recipes/libssh2/all/conanfile.py
```diff
@@ -87,7 +87,7 @@ def configure(self):
         self.settings.rm_safe("compiler.cppstd")
 
     def requirements(self):
-        if self.options.with_zlib:
+        if self.options.with_zlib or self.options.crypto_backend == "openssl":
             self.requires("zlib/[>=1.2.11 <2]")
```
jcar87, line 91: zlib is already a transitive requirement of openssl3 - not sure this should be explicit on openssl alone

the find_package call was probably added to support static builds (it needs to be aware of linking transitive static dependencies), but also to support generating the correct .pc file for static dependencies - this is already handled by Conan and the .pc files are removed - so it would be unneeded/redundant
https://github.com/libssh2/libssh2/commit/67ac735ad09e48bb5c92e765c72a6564d5873c27

jcar87 (approved):

PR #26082, opened 2024-12-01
caches: add version 0.4.0

JoelLefkowitz:
### Summary
Changes to recipe:  **caches/0.4.0**

#### Motivation
Adds a new package [caches](https://github.com/joelLefkowitz/caches)

---


perseoGI (changes requested):
Hi @JoelLefkowitz, thank you for your contribution! We see that you are an active open source contributor 🐸 

I see this conanfile is a replica of the one in your project (upstream) and looks good.
But in Conan Center Index we have some standards and general rules. So some modifications have to be done in this recipe for us to merge into CCI.

We suggest using our templates, in this case, as it is a `header only lib`, this one https://github.com/conan-io/conan-center-index/blob/master/docs/package_templates/header_only/all/conanfile.py
`version` attribute can't be declared as CCI uses `conandata` and `config` yaml files to manage versioning.

Also, one thing very important to understand is that Conan Center Index CI (C3I) has not the purpose of running upstream tests (that responsibility is delegated to upstream), but checking if the package has been correctly "packaged".
As a general rule, `test_package` should be as simple as possible, and should aim to check if header files exist in the package, and if there are any linkage problems. But it should not use dependencies as `gtest`. As I said, that is an upstream task.
So, having `gtest` as a dependency in a CCI recipe is unnecessary.

Moreover, following the template, header only recipes shouldn't use `generate` method unless necessary. 

Sorry for all the comments! We must keep uniformity and follow our standards in CCI!
If you need any help I can guide you or even commit to your PR.

Thank you and happy coding 🐸

JoelLefkowitz:
Thanks so much @perseoGI, I'll update the conanfile and cut down the test file 🙂

JoelLefkowitz:
@perseoGI could you please review these changes again? I have removed gtest and the large test spec from the test_package. I have also switched to the header-only library conanfile template. One difference I have introduced is the line `generators = "SConsDeps"` in the conanfile. This is because I am using Scons to resolve the include paths to run the project's linters.

perseoGI (approved):

AbrilRBS (approved):

PR #26092, opened 2024-12-03
Remove mpdecimal 2.4.2

Ahajha:
Why:
- 2.4.2's build process is significantly different enough that it has remained in its own folder. It contains a few specific patches to "catch it up to date" with the later versions. For example, shared libraries on anything other than MSVC use .so file extensions, it lacks mingw support, and the cxx option doesn't exist in that version.
- 2.4.2 was released nearly a decade ago (around 2016), there have been 3 new versions since then (2.5.0/1 and 4.0.0).
- 2.4.2 is unused in CCI. The other versions are only used in CPython, where 2.4.2 was used up until 3.7, which is EOL and thus was removed.

If there is ever a desire to restore Python 2 support, perhaps this could be revisited, but currently I don't see a point in keeping it.


AbrilRBS (approved):
I agree, thanks a lot. I'm approving as this actually does not need to run on our CI and won't publish new revisions, but will need to work around current limitations of the CI system

valgur (approved):

perseoGI (approved):

ErniGH (approved):
LGTM Thanks for the cleanup! :smile:

PR #26095, opened 2024-12-03
rusty-cpp: new recipe

seekstar:
### Summary
Changes to recipe:  **rusty-cpp**

#### Motivation

This is a library I find handy to use, and others may find it useful, too.

It is currently header-only. If I make it non-header-only in the future, should I modify the existing `conanfile.py`, or should I create a new folder and write a new `conanfile.py` there?

### DIFF
--- a/recipes/rusty-cpp/all/conanfile.py
+++ b/recipes/rusty-cpp/all/conanfile.py
@@ -1,4 +1,5 @@
 from conan import ConanFile
+from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeToolchain
 from conan.tools.files import copy, get
 from conan.tools.layout import basic_layout
@@ -24,6 +25,9 @@ class RustyCppConan(ConanFile):
 
     def layout(self):
         basic_layout(self, src_folder="src")
+    
+    def validate(self):
+        check_min_cppstd(self, 17)
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
--- a/recipes/rusty-cpp/all/conanfile.py
+++ b/recipes/rusty-cpp/all/conanfile.py
@@ -1,6 +1,6 @@
 from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeToolchain
-from conan.tools.files import copy, export_conandata_patches, get
+from conan.tools.files import copy, get
 from conan.tools.layout import basic_layout
 import os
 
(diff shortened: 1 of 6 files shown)


recipes/rusty-cpp/all/conanfile.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+required_conan_version = ">=2.0"
+
+
+class RustyCppConan(ConanFile):
+    name = "rusty-cpp"
+    description = "Write C++ code like Rust!"
+    # dual licensed under the Apache License v2.0 and the MIT License
+    license = "Apache-2.0", "MIT"
```
valgur, line 15:
```suggestion
    license = "Apache-2.0 OR MIT"
```

AbrilRBS (approved):
This looks good on my end, thanks a lot @seekstar for taking the time to submit this to CCI, we appreciate it.

Some notes:

> It is currently header-only. If I make it non-header-only in the future, should I modify the existing conanfile.py, or should I create a new folder and write a new conanfile.py there?

For this, the best approach would be to add a `header_only` option for the current recipe, (alongside the necessary compilation options) that toggles the header only behaviour as needed. See for example [`fmt`](https://github.com/conan-io/conan-center-index/blob/master/recipes/fmt/all/conanfile.py) that does something similar, but we can always help for your specific case when/if the times comes :)

- I've added a `package_id` method - header only libraries are the same for every platform, so this ensures only 1 package is needed everywhere
- I've added a min cppstd check for C++17 based on this line from [CMakeLists.txt](https://github.com/seekstar/rusty-cpp/blob/main/CMakeLists.txt#L5)

Otherwise, the recipe looks great!

recipes/rusty-cpp/all/conanfile.py
```diff
@@ -0,0 +1,53 @@
...
+import os
+
+
+required_conan_version = ">=2.0"
+
+
+class RustyCppConan(ConanFile):
+    name = "rusty-cpp"
+    description = "Write C++ code like Rust!"
+    license = "Apache-2.0 OR MIT"
```
AbrilRBS, line 15:
```suggestion
    license = "Apache-2.0", "MIT"
```

let's wait until Conan 2.x has better modelling of licenses to make this OR decision on behalf of the consumer, safer to have the list approach for now unless upstream clearly states its SPDX identifier

AbrilRBS: Tracked as https://github.com/conan-io/conan/issues/17446

AbrilRBS (approved):

czoido (approved):

seekstar:
Thank you very much. I've applied changes from this PR to the conanfile in the upstream repo.

The readme hasn't mentioned `conan` yet. I tried `conan install --requires="rusty-cpp/[*]` in a project using `rusty-cpp` but the command simply creates many shell scripts like `conanbuildenv-release-x86_64.sh`. I also checked [the document](https://docs.conan.io/2/reference/commands/install.html) but still don't know what's the correct usage of this command.

PR #26096, opened 2024-12-03
libiec61850: new recipe

inPhraZ:
### Summary
Changes to recipe:  **libiec61850/1.5.3**

#### Motivation
Adds libiec61850: an open-source library for the IEC 61850 protocols.

https://github.com/mz-automation/libiec61850

#### Details

---


recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_TESTS"] = False
+        tc.variables["BUILD_SHARED_LIBS"] = False
+        tc.variables["BUILD_STATIC_LIBS"] = True
```
AbrilRBS, line 40:
```suggestion
        tc.variables["BUILD_SHARED_LIBS"] = self.options.shared
        tc.variables["BUILD_STATIC_LIBS"] = not self.options.shared
```

But Conan is already injecting `BUILD_SHARED_LIBS`

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,65 @@
...
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        copy(self, "COPYING", self.source_folder, join(self.package_folder, "licenses"))
+        copy(self, "*.h", self.source_folder, join(self.package_folder, "include"))
+        copy(self, "*.lib", self.source_folder, join(self.package_folder, "lib"))
+        copy(self, "*.so", self.source_folder, join(self.package_folder, "lib"))
+        copy(self, "*.a", self.source_folder, join(self.package_folder, "lib"))
```
AbrilRBS, line 53: Is the install from Conan not doing this already?

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,65 @@
...
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    languages = ["C"]
```
AbrilRBS, line 27:
```suggestion
    languages = ["C"]
    implements = ["auto_shared_fpic"]
```

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,65 @@
...
+        cmake_layout(self)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_TESTS"] = False
+        tc.variables["BUILD_SHARED_LIBS"] = False
```
AbrilRBS, line 39: We probably need to add
```
tc.cache_variables["FIND_PACKAGE_DISABLE_Doxygen"] = True
```

to ensure Doxygen is not picked from the system, as upstream tries to use it

recipes/libiec61850/all/test_package/test_package.cpp
```diff
@@ -0,0 +1,33 @@
+#include <iostream>
+#include <format>
+#include <cstdlib>
+
+#include "libiec61850/iec61850_client.h"
+
+int main(int argc, char **argv)
+{
+    std::string hostname("localhost");
```
AbrilRBS, line 9: Let's avoid hitting the network layer for the test_package. Calling a simple function should be enough

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,65 @@
...
+    languages = ["C"]
+
+    def layout(self):
+        cmake_layout(self)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
```
AbrilRBS, line 36: Upstream CMakeLists.txt seems to be vendorizing a few dependencies - this would be an issue when using the recipe within a graph that shares them from CCI, we'll need to check if those can be made to come from Conan

recipes/libiec61850/all/conanfile.py
inPhraZ: Yes, you're right.
These lines are unnecessary.

recipes/libiec61850/all/conanfile.py
inPhraZ: Thanks for the feedback!
Hmm, since the search for some optional dependencies is explicitly hardcoded into the CMake files, we don't have much flexibility to modify it right now to fetch them from CCI. Given this, the library will be built without TLS and log service support by default.

danimtb:
Hi @inPhraZ!

I have contributed some changes to your PR. The library had some issues packaging the correct configuration for shared/static (the library builds both by default). There were some issues with the cmake install as well mixing both configurations, so I replaced them with the copy calls.

I tested on windows both shared and static and they seem to work. Let's see what the CI says. Thank you!

inPhraZ:
Thanks @danimtb
I really appreciate your help!

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,71 @@
...
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_TESTS"] = False
+        tc.variables["WITH_MBEDTLS"] = False
```
czoido, line 40: Checking the upstream CMake I suspect this will have no effect, I think this are calculated if you have downloaded the sources to a folder and also there's another `WITH_MBEDTLS3`

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,71 @@
...
+        copy(self, "COPYING", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        if self.settings.os == "Windows":
+            cmake.install(component="Development")  # Install header files
+            # Copy files manually no avoid messing with cmake install different targets mixing shared/static
+            copy(self, "*.lib", self.build_folder, os.path.join(self.package_folder, "lib"), keep_path=False)
+            copy(self, "*.dll", self.build_folder, os.path.join(self.package_folder, "bin"), keep_path=False)
+        else:
+            cmake.install()
+            rmdir(self, os.path.join(self.package_folder, "share"))
```
czoido, line 63: I think we can do this without distinguishing if windows or not
```suggestion
        cmake.install(component="Development")  # Install header files
        # Copy files manually because upstream CMakeLists tries to install both shared and static at once
        copy(self, "*.a", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
        copy(self, "*.so", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
        copy(self, "*.dylib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
        copy(self, "*.lib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
        copy(self, "*.dll", src=self.build_folder, dst=os.path.join(self.package_folder, "bin"), keep_path=False)
```

recipes/libiec61850/all/conanfile.py
danimtb:
```suggestion
```

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,66 @@
...
+        targets = ["iec61850-shared", "hal-shared"] if self.options.get_safe("shared") else ["iec61850", "hal"]
+        cmake.build(target=targets)
+
+    def package(self):
+        copy(self, "COPYING", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        cmake = CMake(self)
+        cmake.install(component="Development")  # Install header files
+        # Copy files manually because upstream CMakeLists tries to install both shared and static at once
+        copy(self, "*.a", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+        copy(self, "*.so", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
```
czoido, line 55:
```suggestion
        copy(self, "*.so*", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
```

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,66 @@
...
+        tc.variables["BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_TESTS"] = False
+        tc.cache_variables["FIND_PACKAGE_DISABLE_Doxygen"] = True
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        targets = ["iec61850-shared", "hal-shared"] if self.options.get_safe("shared") else ["iec61850", "hal"]
+        cmake.build(target=targets)
```
czoido, line 47:
```suggestion
        target = "iec61850-shared" if self.options.get_safe("shared") else "iec61850"
        cmake.build(target=target)
```

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,68 @@
+import os
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
+from conan.tools.files import get, copy, rmdir
```
danimtb, line 4:
```suggestion
from conan.tools.files import get, copy
```

danimtb (dismissed):

czoido (approved):

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,68 @@
...
+        # Copy files manually because upstream CMakeLists tries to install both shared and static at once
+        copy(self, "*.so*", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+        copy(self, "*.dylib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+        copy(self, "*.dll", src=self.build_folder, dst=os.path.join(self.package_folder, "bin"), keep_path=False)
+        copy(self, "*.a", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+        copy(self, "*hal.lib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+        copy(self, "*iec61850.lib", src=self.build_folder, dst=os.path.join(self.package_folder, "lib"), keep_path=False)
+
+    def package_info(self):
+        self.cpp_info.components["libiec61850"].libs = ["iec61850"]
```
jcar87, line 62: Seems a bit odd to have a _single_ component - typically these would go in the "root", and you'd expect components to exist when there is more than one

In this case, I'd say there _is_ more than one: `iec61850` and `hal` - I'd recommend just defining the two components, and make `iec61850` depend on the `hal` component, which reflects what upstream does: 

https://github.com/mz-automation/libiec61850/blob/3a1480e5948ca53b4dcc5fa15b05598fa3322bb9/src/CMakeLists.txt#L342-L345

https://github.com/mz-automation/libiec61850/blob/3a1480e5948ca53b4dcc5fa15b05598fa3322bb9/src/CMakeLists.txt#L355-L358

(`target_link_libraries` without a modifier are [implicitly public](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents) - but it's worth checking the generated exported targets. 

where the upstream library is able to generate cmake exported targets, we need to make sure we remain as close as possible to the target names they use - so I'd imagine this is one of those cases where the shared and static libraries have different names - worth checking this, as well as the contents of the `.pc` file and ensure we are compatible from a consumer point of view

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,68 @@
...
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_TESTS"] = False
+        tc.cache_variables["FIND_PACKAGE_DISABLE_Doxygen"] = True
+        tc.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        target = "iec61850-shared" if self.options.get_safe("shared") else "iec61850"
```
jcar87, line 46: as a side comment - in some cases where the build is "fast"  (and Im not sure whether that's the case here or not) - I personally don't think it's a big deal if the project builds both - we only need to ensure only the variant matching the value of the `shared` option is packaged.

recipes/libiec61850/all/conanfile.py
```diff
@@ -0,0 +1,68 @@
...
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_EXAMPLES"] = False
+        tc.variables["BUILD_TESTS"] = False
```
jcar87, line 39: rule of thumb: if the expectation for a user is to pass them via `-D` in the command line (as is this case) these should always be `cache_variables` - this is also covered in our docs:
https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#variables

`tc.variables` should be reserved for things CMake expects in a toolchain file.

jcar87 (changes requested):
cc @danimtb  @czoido

recipes/libiec61850/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "1.5.3":
+    url: "https://github.com/mz-automation/libiec61850/archive/refs/tags/v1.5.3.tar.gz"
```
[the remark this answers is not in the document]
inPhraZ: > any reason why we wouldnt start with the most recent?

Yes, the latest release version (1.6.0) has an issue where one of the required header files is missing from the installation path, causing the test_package to fail. Although this issue has been addressed and fixed in this [PR](https://github.com/mz-automation/libiec61850/pull/531#issue-2612334252), it has not yet been included in an official release.

czoido (dismissed):

danimtb (dismissed):

jcar87 (approved):

PR #26101, opened 2024-12-04
libdwarf: add version 0.11.1

jeremy-rifkin:
### Summary
Changes to recipe:  **libdwarf/0.11.1**

Adding newly released version

---

### DIFF
--- a/recipes/libdwarf/all/conandata.yml
+++ b/recipes/libdwarf/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.11.1":
+    url: "https://github.com/davea42/libdwarf-code/archive/refs/tags/v0.11.1.tar.gz"
+    sha256: "dfd24db87c354b3ba6fa72d29a481014f81768fea15ddcde96a8a938b1557c17"
   "0.11.0":
     url: "https://github.com/davea42/libdwarf-code/archive/refs/tags/v0.11.0.tar.gz"
     sha256: "5135af38dc202206538a3dcdc90675aa8b320e132156849c6855dea5ea0c0729"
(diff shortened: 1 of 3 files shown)


danimtb:
hi @jeremy-rifkin and thanks for the PR!

We have verified that the contents of this patch are the same as the pre-existing patches, just adjusted the lines in the diff to account for the new version. We are trying to minimize the number of patches in conan-center and try to report the issues to the library repo and maintainers upstream. In this case the patch was there previously (introduced in previous PRs https://github.com/conan-io/conan-center-index/pull/22629, https://github.com/conan-io/conan-center-index/pull/25177 ...).

I have tested this on Windows msvc 193 and the patch is needed for the library to build, so it should be required.

danimtb:
After some tests with the contents of the patch, they have revealed that the patch is only required for Windows and that the CMake-related ones are not necessary. We have seen that this has happened since 0.9.2 version, so we will amend this issue in an upcoming PR after merging this one.

As stated before, the goal is to reduce the influence of patches as much as possible in conan center.

@jeremy-rifkin since you were updating the recipe and adding some of those patches in the past, let us know if there is something we could be missing. Thank you! 😄

danimtb (approved):

jeremy-rifkin:
Hi thanks for looking into this! If the recipe works without some patches that’s fantastic, I just copied over everything that appeared to still be relevant 😄

franramirez688 (approved):
As @danimtb said above, we could summarize this as (since version 0.9.2):

* UNIX patches are not needed
* Windows needs only the header patches because of this https://github.com/davea42/libdwarf-code/commit/6ffd41d39ba8e5db8651a35ac4f975baf786de4c

Let's open another PR to simplify those patches. I also recommend removing the 0.5.0, 0.7.0, and 0.9.0 in that future PR.

PR #26102, opened 2024-12-04
cspice: Add Linux armv8 support for N0067

sassy-asjp:
### Summary
Changes to recipe:  **cspice/0067**

#### Motivation
The current cspice recipe rejects building on Linux armv8 as it cannot find the correct source package to download, however it is possible to build cspice for Linux armv8 even if it is not officially released on the platform by NASA, and is even packaged for Linux armv8 by conda-forge.

#### Details
The patches to build for Apple Silicon macOS (essentially removal of `-m64`) are the same for Linux armv8 and appear to what conda-forge uses for their Linux armv8 cspice package.

https://naif.jpl.nasa.gov/pipermail/spice_discussion/2021-April/000466.html

This adds the Linux armv8 to the sources data via the Apple Silicon macOS tarball from NASA.

This was tested using a Linux virtual machine running on an M1 Pro Mac.

---

### DIFF
--- a/recipes/cspice/all/conandata.yml
+++ b/recipes/cspice/all/conandata.yml
@@ -17,8 +17,8 @@ sources:
           url: "https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_N0067/C/PC_Linux_GCC_64bit/packages/cspice.tar.Z"
           sha256: "60a95b51a6472f1afe7e40d77ebdee43c12bb5b8823676ccc74692ddfede06ce"
         "armv8":
-          url: "https://naif.jpl.nasa.gov/pub/naif/misc/toolkit_N0067/C/MacM1_OSX_clang_64bit/packages/cspice.tar.Z"
-          sha256: "0deae048443e11ca4d093cac651d9785d4f2594631a183d85a3d58949f4d0aa9"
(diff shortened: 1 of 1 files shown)


sassy-asjp:
@ErniGH That was the impression I got from the linked discussion, which lead me to trying to macOS/M1 distribution on Linux, which worked. However, digging deeper into it, it seems like conda-forge downloads the Linux x86 64 sources then applies a patch to remove the `-m64` flag.

Would it be preferable if I rewrote this in a similar way? Would you have a recommendation of how to apply a patch to only one platform?

ErniGH:
Hello, I've been giving it another thought and reviewing the issue. We can directly use the same Linux package for x86_64 on ARMv8. We don't need to patch anything; by using the CMakeLists (https://github.com/conan-io/conan-center-index/blob/master/recipes/cspice/all/CMakeLists.txt) and the CMake toolchain, Conan can infer the necessary flags for your profile and build without any problems.
I tested it with a Linux image, which works perfectly for me. You can try it! 😄. 
I pushed the changes directly to your branch.

sassy-asjp:
@ErniGH It works for me too! Thanks for the help!

czoido (approved):

PR #26104, opened 2024-12-04
glaze: add version 4.3.1

toge:
### Summary
Changes to recipe:  **glaze/4.3.1**

#### Motivation
There are several improvements and new features since 4.0.1.

#### Details
https://github.com/stephenberry/glaze/releases/tag/v4.3.1
https://github.com/stephenberry/glaze/compare/v4.0.1...v4.3.1

---

### DIFF
--- a/recipes/glaze/all/conandata.yml
+++ b/recipes/glaze/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "4.2.3":
-    url: "https://github.com/stephenberry/glaze/archive/v4.2.3.tar.gz"
-    sha256: "cfe439b211dcca62649e92d3e8b404477b02df07522bde843e98ce25ff5bf824"
+  "4.2.4":
+    url: "https://github.com/stephenberry/glaze/archive/v4.2.4.tar.gz"
+    sha256: "dd409d60b60cb65c7a71ade6331a649710d0eb3d56c66c3bfd461e837359cc2a"
   "4.0.1":
     url: "https://github.com/stephenberry/glaze/archive/v4.0.1.tar.gz"
     sha256: "0026aca33201ee6d3a820fb5926f36ba8c838bfd3120e2e179b0eee62b5bd231"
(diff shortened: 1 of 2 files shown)


RazielXYZ (approved):

RazielXYZ (approved):

toge:
@leha-bot 
Thanks!
I updated it.

RazielXYZ:
There have been many new features and performance improvements since the version that is currently on CC (4.0.1), so it would be great if this could get in at some point soon!

uilianries (approved):
LGTM. Arigato @toge !! :green_heart: 

I just reviewed the diff of both versions available in CCI: https://github.com/stephenberry/glaze/compare/v4.0.1...v4.3.1

- Lots of changes regarding documentation and cpp code
- No direct build changes, only version bump in CMakeLists.txt
- Few changes in tests, but we are safe as we do not build them
- No license change

@pwqbot Thank you for pinging!

jcar87:
> @czoido @uilianries @AbrilRBS Could you please help review this PR? It’s very straightforward but has been pending for over a month.

Thanks @pwqbot  - please be mindful that we have quite a large backlog, and while the PR looks straightforward, each version added goes through a review process:

- We make sure that the new version has actually been released upstream - you'd be surprised how many times we've been caught out by contributors adding not-yet-released version because there is a work-in-progress branch or tag - we end up building from the wrong commit - causing confusions to users that end up using pre-releases, and disruption once the actual version is released (as there can be breaking changes)
- Ensure that the released version has been out at least for 48 hours - haste is never on our side, it happens quite frequently that library maintainers re-tag the same version with different contents because of some last minute bug - this is unfortunately but happens more that we would like
- Check if the license has not changed (minor in the sense that it has no major technical implications, but important to a significant part of our user base)
- We check the diff to see if there are any significant changes in the build scripts that should be mirrored in the recipe file - as it is possible for the PR to "pass" CI but have issues, for example, issues we've seen in the past:
   - build scripts requiring new versions of cmake
   - there are new optional dependencies that are automatically searched for by the build system - this can cause issues for users that have these installed from the system
   - there are new components libraries that need to be expressed in package_info
   - there are compiler definitions or options that need to be transitively propagated to consumers
  
Not exhaustive - but we do this for every PR that adds a new version to ensure that our users will not face issues or disruption - Conan Center has a large user base and disruption can affect a lot of users. Thanks for understanding and apologies for the delay.

PR #26109, opened 2024-12-04
mingw-builds: Add 14.2.0

AbrilRBS:
### Summary
Changes to recipe:  **mingw-builds/14.2.0**

#### Motivation

I've been using it this last few days locally and works as expected, and is been useful to debug some gcc 14 related issues (See https://github.com/conan-io/conan-center-index/issues/26034)

Some extra cleanup can be made, but that would regenerate revisions for the old versions, no need to do that unless another valid change would be needed

### DIFF
--- a/recipes/mingw-builds/all/conandata.yml
+++ b/recipes/mingw-builds/all/conandata.yml
@@ -1,4 +1,20 @@
 sources:
+  "14.2.0":
+    mcf:
+      seh:
+        ucrt:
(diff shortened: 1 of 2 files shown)


jcar87:
Thanks! 

Confirmed this works well with the following host profile:

```
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.exception=seh
compiler.libcxx=libstdc++
compiler.threads=posix
compiler.version=14
os=Windows
[tool_requires]
*: mingw-builds/[*]
```

PR #26110, opened 2024-12-05
openvino: fix yaml

AMDmi3:
### Summary
Changes to recipe:  **openvino**

Remove stray `patch:` line which breaks the schema. 

---

### DIFF
--- a/recipes/openvino/all/conandata.yml
+++ b/recipes/openvino/all/conandata.yml
@@ -31,4 +31,3 @@ sources:
     "onednn_gpu":
       url: "https://github.com/oneapi-src/oneDNN/archive/cb77937ffcf5e83b5d1cf2940c94e8b508d8f7b4.tar.gz"
       sha256: "2ca304c033786aa5c3ec1ec6f8fc3936ae5c6874d5964b586311da11bec2ec4a"
-patches:


jcar87:
Thanks @AMDmi3 - apologies for this, this was detected by our CI but not correctly reported before merging - I'll make sure this is addressed.

Remember we can only merge this after the CLA is signed, thanks :)

AMDmi3:
I'm not signing any CLAs, sorry, but if it's not fixed I'll have to remove ConanCenter from Repology.

jcar87:
Given that this is a trivial change we can take care of this, closing in favour of https://github.com/conan-io/conan-center-index/pull/26111.

Please note that unless I'm mistaken, this is not an error: the yaml passes yamllint validation and is correctly processed by Conan - it is equivalent of having `patches: null`, which afaik is valid yaml. I'm unsure of the Repology integration, but it may be worth accounting for things like this.

on the other hand, it does happen that our linter has the ability to pick this up, but it is a warning and not necessarily an error.

AMDmi3:
> Please note that unless I'm mistaken, this is not an error

This is a valid yaml, yes, but at least Repology doesn't expect that what supposed to be a list becomes null suddenly. Is there a schema definition for the format?

PR #26119, opened 2024-12-05
Fix/openmpi

memsharded:
### Summary
Changes to recipe:  **openmpi**

- Using ``libevent`` dependency from Conan package, using the same approach as ``hwloc``. This avoids problems when libevent is installed in the system, the default policy will use it, but the consumers of such libevent are completely unaware
- Correctly model the ``opal`` component, verified upstream that it generates an ``opal.cp`` Pkgconfig file, that ``orte`` depends on ``opal``. This allows being used with the new CMakeDeps generator

### DIFF
--- a/recipes/openmpi/all/conanfile.py
+++ b/recipes/openmpi/all/conanfile.py
@@ -62,6 +62,7 @@ class OpenMPIConan(ConanFile):
         # transitive_headers=True is not needed for any dependencies.
         self.requires("hwloc/2.10.0")
         self.requires("zlib/[>=1.2.11 <2]")
+        self.requires("libevent/2.1.12")
         if self.settings.os == "Linux":
             self.requires("libnl/3.8.0")
         if self.options.get_safe("with_verbs"):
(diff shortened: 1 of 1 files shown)


valgur:
I would appreciate a reference to the `opal.pc` being generated in the upstream project, if possible. I based the components and .pc files in the recipe on the output of `make install`, so I'm wondering what I missed.

In the related issue you also mentioned
> Unfortunately, I have tried it with the OpenMPI recipe, and it has some bad practice there that is blocking its usage by the new CMakeDeps generator. We will try to address the issue in ConanCenter OpenMPI recipe to unblock this.

What was the issue there? I would like to avoid it in the future.

memsharded:
> Unfortunately, I have tried it with the OpenMPI recipe, and it has some bad practice there that is blocking its usage by the new CMakeDeps generator. We will try to address the issue in ConanCenter OpenMPI recipe to unblock this.

I meant "bad practice" for the new CMakeDeps generator, from the CPS point of view. Having multiple ``self.cpp_info..libs = ["lib1", "lib2"]`` will be very discouraged for the new ``CMakeDeps`` generator, it will be temporarily managed, but later deprecated, and when defined in ``self.cpp_info.components["mycomp"].libs = ["lib1", "lib2"]`` will not be allowed. Please do not rush to change this in ConanCenter, as this depends on how the ``CMakeDeps`` and CPS evolve, maybe there is a case for link groups. In this case it seems the ``opal`` component is the right model to align with pkgconfig ``opal.pc`` upstream.

PR #26133, opened 2024-12-08
simdjson: add version 3.11.2

toge:
### Summary
Changes to recipe:  **simdjson/3.11.2**

#### Motivation
There are lots of improvements since 3.10.1.

#### Details
https://github.com/simdjson/simdjson/compare/v3.10.1...v3.11.2

---

### DIFF
--- a/recipes/simdjson/all/conandata.yml
+++ b/recipes/simdjson/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "3.11.0":
+    url: "https://github.com/simdjson/simdjson/archive/v3.11.0.tar.gz"
+    sha256: "f3469e776ca704cfda47f0a43331690c882f82c9c0c6f185452387c1e222a63e"
   "3.10.1":
     url: "https://github.com/simdjson/simdjson/archive/v3.10.1.tar.gz"
     sha256: "1e8f881cb2c0f626c56cd3665832f1e97b9d4ffc648ad9e1067c134862bba060"
--- a/recipes/simdjson/all/conandata.yml
+++ b/recipes/simdjson/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "3.11.0":
-    url: "https://github.com/simdjson/simdjson/archive/v3.11.0.tar.gz"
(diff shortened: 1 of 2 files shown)


ErniGH:
Hi @toge, thanks for your PR! It looks good overall. The issue I see is that in upstream, the CMakeLists uses something like `include(cmake/CPM.cmake)` and in `dependencies/CMakeLists` it uses `CPMAddPackage` with external dependencies, which could pose a problem in the future.
[simdjson/simdjson@`v3.10.1...v3.11.2`#diff-1e7de1ae2d](https://github.com/simdjson/simdjson/compare/v3.10.1...v3.11.2#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR290)
[simdjson/simdjson@`v3.10.1...v3.11.2`#diff-d52c26e7a1](https://github.com/simdjson/simdjson/compare/v3.10.1...v3.11.2#diff-d52c26e7a1e6f0148faf32d9742cb36be9918aab8e0ae1ffa4304104090c759cR56)

Given this, I was reviewing the build logs, and it doesn’t seem like these dependencies are being used in any way. I tried toggling the `SIMDJSON_ALLOW_DOWNLOADS`, `SIMDJSON_CXXOPTS`, and `SIMDJSON_GOOGLE_BENCHMARKS` variables to see if I could change the behavior, but nothing happened. In fact, I was getting a warning indicating they weren’t being used.

@uilianries finally solved the mystery—thank you very much! These dependencies are only used when the `SIMDJSON_DEVELOPER_MODE` variable is set. After some testing and modifying the recipe, I confirmed that setting this variable to `ON` triggers the download of all those external dependencies.
https://github.com/simdjson/simdjson/blob/d4bf0cc7ec12c30c68cb7ed443895df546390283/CMakeLists.txt#L265

I’m writing this so that in the future, we’re cautious about the possibility of a small change upstream causing Conan users to unknowingly download external packages.

Thank you so much for your effort!

ErniGH (approved):
Thanks!

toge:
@ErniGH 
Thank you for notice!
I will check carefully about it when glaze is updated.

jcar87 (approved):

PR #26138, opened 2024-12-09
depot_tools: Added latest revision

klausholstjacobsen:
### Summary
depot_tools:cci.20241206

#### Motivation
Want to use newest depot_tools

---

### DIFF
--- a/recipes/depot_tools/all/test_package/conanfile.py
+++ b/recipes/depot_tools/all/test_package/conanfile.py
@@ -14,5 +14,4 @@ class TestPackageConan(ConanFile):
         self.tool_requires(self.tested_reference_str)
 
     def test(self):
-        # cit: Chrome Infrastructure CLI
-        self.run("cit --help")
+        self.run("gclient --version")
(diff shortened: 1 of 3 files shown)


ErniGH (dismissed):
Thanks for your effort 😄

recipes/depot_tools/all/test_package/conanfile.py
```diff
@@ -14,5 +14,4 @@ def build_requirements(self):
         self.tool_requires(self.tested_reference_str)
 
     def test(self):
-        # cit: Chrome Infrastructure CLI
-        self.run("cit --help")
+        self.run("gclient --version")
```
jcar87, line 17: has this changed between versions? this PR only checks the published version -  does this test_package still work with the older versions?

ErniGH: You are right this test with the `20200407` version doesn't work on Macos

uilianries:
I made a change to the test package: the depot tools is a collection of tools, and most of them are wrapped by a python script. The same package carries a wrapper for python3 (vpython3), and when running it for the first time, including any of those tools, it downloads several dependencies, resulting in test package not only needed network, but delaying its execution as well. 

As the recipe does not build anything, but only extracts and copies files to the package, I guess checking if it's installed will be enough.

uilianries (dismissed):

uilianries (approved):
Removed extra version prefix to avoid version mismatch when checking.

PR #26143, opened 2024-12-09
Fix bare `self.version` comparisions in CCI

AbrilRBS:
Follow up to https://github.com/conan-io/conan-center-index/pull/26042, only fixing the comparision issues

* Fix version comparisons between two strings (`self.version` and a string) - by surrounding `self.version` with `Version()` - string comparisons are lexicographically sorted and may not sort properly.  Note that for the recipes impacted in this PR this was *not* an issue, this is only to prevent issues in the future as versions are added/removed
* Fix typo in pcapplusplus patch to fix cross-compilation


perseoGI (approved):
LGTM!

danimtb (approved):

PR #26146, opened 2024-12-09
Add libe57format/3.2.0 release

EstebanDugueperoux2:
### Summary
Changes to recipe:  **libe57format/3.2.0**

#### Motivation
Add 3.2.0 release

### DIFF
--- a/recipes/libe57format/all/conandata.yml
+++ b/recipes/libe57format/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "3.2.0":
+    url: "https://github.com/asmaloney/libE57Format/archive/refs/tags/v3.2.0.tar.gz"
+    sha256: "afe4de960ba486760702127d8f3dc537c9b5921af7cfb2a75177fe2f80956e89"
   "3.1.1":
     url: "https://github.com/asmaloney/libE57Format/archive/refs/tags/v3.1.1.tar.gz"
     sha256: "949e73db3cb90ed7d286c49d12c6925813ead8d92ff9b84e0fba17fa015194d0"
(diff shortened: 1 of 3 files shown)


jcar87 (changes requested):
Thanks @EstebanDugueperoux2  - please only make the changes required by the PR objetive (adding 3.2.0) - we've had issues with hasty removals of old logic leading to unintended bugs. For the time being, the team will try to take on these clean ups _after_ the PR has been reviewed and if it doesn't cause delays in merging a useful feature.

EstebanDugueperoux2:
> Thanks @EstebanDugueperoux2 - please only make the changes required by the PR objetive (adding 3.2.0) - we've had issues with hasty removals of old logic leading to unintended bugs. For the time being, the team will try to take on these clean ups _after_ the PR has been reviewed and if it doesn't cause delays in merging a useful feature.

Hi @jcar87, the issue is that only adding new release make pipeline failed.
And it seems patch done inside the recipe is no more usefull.

jcar87:
> Hi @jcar87, the issue is that only adding new release make pipeline failed.

mmm, we can have a look, what was the CI failure in this case?

> And it seems patch done inside the recipe is no more usefull.

The removal of the `_patch_sources()` method is in effect for all already published versions, so it is introducing regressions for pre-existing versions w.rt. `POSITION_INDEPENDENT_CODE`, and introducing a regression for 3.1.1 by removing the warnings as errors - if this PR were merged, it could cause issues to users of 3.1.1 - 

> And it seems patch done inside the recipe is no more usefull.
We keep asking contributors to avoid these kind of changes because we very often see that they are more invasive and have side effects that are not intended, as it seems to be the case here

EstebanDugueperoux2:
Hi @jcar87 I have updated recipe to relax replace_in_file with strict=False because "set_warning_as_error()" doesn't exists anymore in recent release.
It was the cause of previous pipeline failures.

jcar87 (approved):
Thanks for amending the PR!

Issue #17846, opened 2023-06-06
[package] openssl/3.1.1: Cross-compilation 'Windows:x86_64' to 'Windows:armv8' (ARM64) with zlib fails using MSVC


zenden2k:
### Description

Cross-compilation 'Windows:x86_64' to 'Windows:armv8' (ARM64) with default options fails using MSVC compiler. 

```
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol deflate referenced in function bio_zlib_write
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol deflateEnd referenced in function bio_zlib_free
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol inflate referenced in function bio_zlib_read
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol inflateEnd referenced in function bio_zlib_free
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol deflateInit_ referenced in function bio_zlib_write
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol inflateInit_ referenced in function bio_zlib_read
libcrypto.lib(libcrypto-lib-c_zlib.obj) : error LNK2019: unresolved external symbol zError referenced in function bio_zlib_read
providers\legacy.dll : fatal error LNK1120: 7 unresolved externals
```

If I add the **openssl:no_zlib=True** option then compilation succeeds. 

### Package and Environment Details

* Package Name/Version: **openssl/3.1.1**
* Operating System+version: **Windows 11 22H2**
* Compiler+version: **Visual Studio 2022 (C++ compiler 19.36.32532 for ARM64)**
* Conan version: **conan 1.57.0**
* Python version: **Python 3.11.3**

### Conan profile

Host profile (vs2022_arm64_release):
```
[settings]
os=Windows
os_build=Windows
arch=armv8
arch_build=x86_64
compiler=Visual Studio
compiler.version=17
compiler.runtime=MT
build_type=Release
[options]
[build_requires]
[env]
````

Build profile (vs2022_x64):
```
[settings]
os=Windows
os_build=Windows
arch=x86_64
arch_build=x86_64
compiler=Visual Studio
compiler.version=17
compiler.runtime=MT
build_type=Release
[options]
[...]

PR #26151, opened 2024-12-10
openssl: ability use jom and fix Windows arm64

jcar87:
DRAFT

Note: building this natively on Windows arm64 will require `--settings:build "jom/*:arch=x86_64`  - may want to tweak it behind a conf

The VC-WIN64-ARM is for "Windows OneCore" as per https://openssl-library.org/policies/platforms/index.html, and is not the one intended when targetting desktop Windows arm64. 
Fix by using `"VC-WIN64-CLANGASM-ARM"` instead - this uses Visual C++ as the compiler, and `clang-cl` as the assembler, because `nasm` does not target arm64, more details here: https://www.linaro.org/blog/openssl-for-windows-on-arm/

Other changes:
- Keep the most recent patch releases for recent minors
- Remove Conan 1.x lines that are no-ops in Conan 2.x
- Remove test_v1_package

Close https://github.com/conan-io/conan-center-index/issues/22147
Close https://github.com/conan-io/conan-center-index/issues/17846


recipes/openssl/3.x.x/conanfile.py
```diff
@@ -282,7 +279,7 @@ def _targets(self):
             "Windows-x86-Visual Studio": "VC-WIN32",
             "Windows-x86_64-Visual Studio": "VC-WIN64A",
             "Windows-armv7-Visual Studio": "VC-WIN32-ARM",
-            "Windows-armv8-Visual Studio": "VC-WIN64-ARM",
```
jcar87, line 285: The `VC-WIN64-ARM` is for "Windows OneCore" as per https://openssl-library.org/policies/platforms/index.html, and is not the one intended when targetting desktop Windows arm64.

recipes/openssl/3.x.x/conanfile.py
```diff
@@ -470,6 +467,11 @@ def _create_targets(self, cflags, cxxflags, defines, ldflags):
             cflags.append("-fembed-bitcode")
             cxxflags.append("-fembed-bitcode")
 
+        if is_msvc(self) or self._is_clang_cl:
```
memsharded, line 470: AutotoolsToolchain currently does:
```
    def _msvc_extra_flags(self):
        if is_msvc(self._conanfile) and check_min_vs(self._conanfile, "180",
                                                     raise_invalid=False):
            return ["-FS"]
        return []
```
So not covering ClangCL, maybe it needs to be added there?

jcar87: oh this block needs to be removed 100% 
`-FS` is already passed by `AutotoolsToolchain` - so this is, at best, redundant

the problem is that ... is still fails sporadically with the concurrent access error, despite the fact that -FS is passed. To be fair jom was just for faster local testing, but it would be good if this was working well

valgur: These can be dropped, but it's only available since `2.1.0`, so you might want to update `required_conan_version` accordingly.

valgur:
Why not make building with jom the default? It really is painfully slow without it.

I did have to include a fallback to `nmake` since building with `jom` occasionally failed due to some kind of concurrency issues:
https://github.com/valgur/conan-center-index/blob/468bd176475728c8e1d477fa6b8408bcce8f8107/recipes/openssl/3.x.x/conanfile.py#L533-L540

I would also suggest the following patch, which reverts a workaround from the Makefiles that was only included for Borland compilers, which are not even (officially) supported by Conan. As the comment says, I found a 25% speedup when building inside a Windows VM.
```python
# Remove unnecessary indirection via `cmd /C`, which is needed only for the bcc32 compiler.
# Speeds up the build by 25% or so.
# https://github.com/openssl/openssl/issues/9931#issuecomment-1476649801
replace_in_file(self, os.path.join(self.source_folder, "Configurations", "windows-makefile.tmpl"),
                r'cmd /C "$makedepcmd $cflags $defs $srcs > $dep 2>&1"',
                r'$makedepcmd $cflags $defs $srcs > $dep 2>&1')
```

jcar87:
> I did have to include a fallback to `nmake` since building with `jom` occasionally failed due to some kind of concurrency issues:
> [https://github.com/valgur/conan-center-](https://github.com/valgur/conan-center-index/blob/468bd176475728c8e1d477fa6b8408bcce8f8107/recipes/openssl/3.x.x/conanfile.py#L533-L540)

If using jom fails occasionally in a way that we do not understand - I'd say that's pretty much self-explanatory as to why we're *not* making it it default. 

As for `cmd /c`, I'd wait until the OpenSSL maintainers make a decision. I'd suggest opening a PR against OpenSSL to remove that or restrict it to that compiler. In your case, you observe a 25% speedup, but this comment https://github.com/openssl/openssl/issues/9931#issuecomment-1476113592 here suggested no noticeable speedup. So I'd say some investigation is required.

AbrilRBS (approved):

memsharded (approved):

Issue #26152, opened 2024-12-10
[opentdf-client] upstream project has been removed


valgur:
The upstream project for the `opentdf-client` [recipe](https://github.com/conan-io/conan-center-index/tree/master/recipes/opentdf-client/all) has been removed together with its source archives: https://github.com/opentdf/client-cpp
The new GitHub project appears to have switched over to Go and no longer even offers a C++ SDK: https://github.com/opentdf

Should the recipe be removed? Or kept alive with CCI source mirrors?

Note: I have no personal interest in this recipe, just reporting a potential issue.

/cc @patmantru since you have been most involved with the recipe in the past.

AbrilRBS:
Hi @valgur thanks a lot for the heads-up. This is in fact a bit unfortunate, removing the sources seems a bit counterproductive, but alas!

I've created the linked PR above to at least let current users compile still, but we'll need to later have a conversation with the team as to the future of the recipe in the git repository

PR #26155, opened 2024-12-10
Use CCI's backup source for deleted openrdf-client sources

AbrilRBS:
### Summary
Changes to recipe:  **opentdf-client/[*]**

#### Motivation

Seems like the upstream project has switched away from C++ with little announcement, and in the process, the remote has disappeared alongside its sources.

We'll need to check if keeping the recipe in the CCI git repository makes sense then (Remember that removing it from this repo would in no case remove it from the `conancenter` Conan remote, those binaries will always be present)

#### Details

Closes https://github.com/conan-io/conan-center-index/issues/26152

As can be seen from the sha256 not changing, the sources are 1:1 to their original counterparts

### DIFF
--- a/recipes/opentdf-client/all/conandata.yml
+++ b/recipes/opentdf-client/all/conandata.yml
@@ -1,21 +1,21 @@
 sources:
   "1.5.6":
-    url: "https://github.com/opentdf/client-cpp/archive/1.5.6.tar.gz"
+    url: "https://c3i.jfrog.io/artifactory/conan-center-backup-sources/3f46d9a175ce67fc2814faf34a42bc6ff9519d9a80ae7dbdd8d34317231a404a"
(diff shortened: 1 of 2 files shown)


AbrilRBS:
Need to check how to rename the get file so it understands to unzip it as a `tar.gz`

jcar87:
Left a question here: https://github.com/orgs/opentdf/discussions/1819 that remains without a response.

Don't think this is worth the hassle, unless users of this library have issues. Arguably, there's also some work on their end given that they are now using a library that appears unpublished - if I was in this situation I would cease using this library.

Pending a confirmation, I think this could be a case for "deprecated" - appreciate that some external users may have scripts that attempt to export/build all recipes, but that's outside of our scope - hopefully they can come up with a mechanism to mark known expected failures.

PR #26160, opened 2024-12-11
ginkgo: add version 1.9.0

MarcelKoch:
### Summary
Changes to recipe:  **ginkgo/1.9.0**

#### Motivation/Details
This PR adds version 1.9.0 of the Ginkgo library.

---

### DIFF
--- a/recipes/ginkgo/all/conanfile.py
+++ b/recipes/ginkgo/all/conanfile.py
@@ -33,28 +33,43 @@ class GinkgoConan(ConanFile):
         "fPIC": [True, False],
         "openmp": [True, False],
         "cuda": [True, False],
+        "half": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": False,
         "openmp": False,
         "cuda": False,
+        "half": False,
     }
 
     @property
     def _min_cppstd(self):
-        return "14"
+        if Version(self.version) >= "1.9.0":
+            return "17"
+        else:
+            return "14"
 
     @property
     def _minimum_compilers_version(self):
-        return {
-            "Visual Studio": "16",
-            "msvc": "193",
-            "gcc": "5.4",
-            "clang": "3.9",
-            "apple-clang": "10.0",
-            "intel": "18",
-        }
+        if Version(self.version) >= "1.9.0":
+            return {
+                "Visual Studio": "16",
+                "msvc": "193",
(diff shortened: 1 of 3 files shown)


MarcelKoch:
cc @upsj

recipes/ginkgo/all/conanfile.py
```diff
@@ -102,7 +117,10 @@ def loose_lt_semver(v1, v2):
 
     def build_requirements(self):
         if Version(self.version) >= "1.7.0":
-            self.tool_requires("cmake/[>=3.16 <4]")
+            if self.options.cuda:
+                self.tool_requires("cmake/[>=3.18 <4]")
```
jcar87, line 121: Any details as to why this is needed are appreciated, thanks!

upsj: See the [README](https://github.com/ginkgo-project/ginkgo?tab=readme-ov-file#linux-and-mac-os), the CUDA build requires a more recent CMake version.

recipes/ginkgo/all/conanfile.py
```diff
@@ -33,28 +33,43 @@ class GinkgoConan(ConanFile):
...
         "openmp": [True, False],
         "cuda": [True, False],
+        "half": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": False,
         "openmp": False,
         "cuda": False,
+        "half": False,
```
jcar87, line 43: Upsteram seems to be setting this to `True` by default, except for Windows
https://github.com/ginkgo-project/ginkgo/blob/d7e1450b6ba9ee90dbaa839f4b4b5a5ad59e28cc/CMakeLists.txt#L46-L51

Would it not make sense to do the same in the recipe?  (for cases like this, we can leave the value in `default_options` undefined, and conditionally set a value in the `config_options()` method - as long as it's only conditional on settings)

MarcelKoch: I think the conan package was intended to disable all optional features by default and only enable those which are explicitly requested by the user. To leave it consistent, I also disabled the half precision feature by default.

jcar87:
We have added version 1.10.0 at [27709](https://github.com/conan-io/conan-center-index/pull/27709) - continued from this PR, thanks for your contribution!

PR #26162, opened 2024-12-11
libudev: add a non-system recipe

### LINKED ISSUE #26161: libudev libs: Host dependency and cross compilation
### What is your question?

Hi !, my question is regarding building libudev libs for linking to host - for cross compilation purposes.
Existing libudev only seem to install/check 'system' udev libs, but doesn't pull source code and build.  As a result even cross compilation of 'libudev' itself fails (in test_package) as it doesn't find right '.h' files.

So, for cross compilation case, wherein linker requires corresponding udev libs, what is the suggested option?

valgur:
### Summary
Changes to recipe:  **libudev/255.15**

#### Motivation
`libudev/system` is used in several recipes on CCI, but like other `system` packages, it is not usable in a cross-compilation context.

- Resolves #26161

#### Details
The `libudev` library is a part of systemd, but neither `libsystemd` or `libudev` depend on each other. I considered making it a component of `libsystemd`, but keeping them independent of each other seemed preferable, since they are not really used simultaneously in any recipes and it allows for a lighter `libudev` dependency. The only downside is slight duplication in the recipe.

The recipe cross-compiles successfully with `aarch64-linux-gnu-gcc-14`.

---

### DIFF
--- a/recipes/libudev/all/conandata.yml
+++ b/recipes/libudev/all/conandata.yml
@@ -1,9 +1,9 @@
 sources:
-  "255.13":
-    url: "https://github.com/systemd/systemd-stable/archive/refs/tags/v255.13.tar.gz"
-    sha256: "b24bbfc18e27339f02bdef12d2be95b747c766e7eb9248554e53ad8979c01f2c"
+  "255.15":
+    url: "https://github.com/systemd/systemd-stable/archive/refs/tags/v255.15.tar.gz"
+    sha256: "f9c1754d65b6ab12c658e832292b3d3dd26722d016f9d065f85fe6f9747e51d2"
 patches:
-  "255.13":
+  "255.15":
     - patch_file: "patches/255.13/0001-fix-format-overflow.patch"
       patch_type: "bugfix"
       patch_description: "Fix a format overflow fatal compiler warning"
--- a/recipes/libudev/all/conandata.yml
+++ b/recipes/libudev/all/conandata.yml
@@ -1,0 +1,9 @@
+sources:
+  "255.13":
+    url: "https://github.com/systemd/systemd-stable/archive/refs/tags/v255.13.tar.gz"
+    sha256: "b24bbfc18e27339f02bdef12d2be95b747c766e7eb9248554e53ad8979c01f2c"
+patches:
+  "255.13":
+    - patch_file: "patches/255.13/0001-fix-format-overflow.patch"
+      patch_type: "bugfix"
+      patch_description: "Fix a format overflow fatal compiler warning"
--- a/recipes/libudev/all/conanfile.py
+++ b/recipes/libudev/all/conanfile.py
@@ -1,54 +1,149 @@
+import os
+import tarfile
+
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
-from conan.tools.system import package_manager
-from conan.tools.gnu import PkgConfig
+from conan.tools.env import VirtualBuildEnv
+from conan.tools.files import download, move_folder_contents, copy, export_conandata_patches, apply_conandata_patches
+from conan.tools.gnu import PkgConfigDeps
+from conan.tools.layout import basic_layout
+from conan.tools.meson import MesonToolchain, Meson
+from conan.tools.scm import Version
 
 required_conan_version = ">=1.50.0"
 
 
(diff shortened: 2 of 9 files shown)


vithalsm:
@valgur (in continuation of other thread on cross compiling libudev from x86_64 to x86_64). 
We tried with applying the diffs from this PR.  However, libudev build is failing as it can't find a file 'sys/capability.h'.  However, libcap is cross built and available for host.  
I don't see INC path getting included (explicity), wondering whether that could be the issue.

If you can help us with quick look please.  May be by cross building for x86.
============================ Build logs (error part) ===============
Command line: `/usr/x86-10.3-glibc-2.34/bin/x86_64-myplat_buildroot-linux-gnu-gcc /home/aos-dev/.conan2/p/b/libud3200b2f38cb14/b/build-relwithdebinfo/meson-private/tmp3ule2c14/testfile.c -o /home/aos-dev/.conan2/p/b/libud3200b2f38cb14/b/build-relwithdebinfo/meson-private/tmp3ule2c14/output.obj -c --sysroot=/usr/myplat-sysroot -D_FILE_OFFSET_BITS=64 -O0 -std=gnu11` -> 0
Message: gperf len type is size_t
Running compile:
Working directory:  /home/aos-dev/.conan2/p/b/libud3200b2f38cb14/b/build-relwithdebinfo/meson-private/tmp8pheyca2
Code:
 
        #ifdef __has_include
         #if !__has_include("sys/capability.h")
          #error "Header 'sys/capability.h' could not be found"
         #endif
        #else
         #include <sys/capability.h>
        #endif
-----------
Command line: `/usr/x86-10.3-glibc-2.34/bin/x86_64-myplat_buildroot-linux-gnu-gcc /home/aos-dev/.conan2/p/b/libud3200b2f38cb14/b/build-relwithdebinfo/meson-private/tmp8pheyca2/testfile.c -E -P --sysroot=/usr/myplat-sysroot -D_FILE_OFFSET_BITS=64 -P -O0 -std=gnu11` -> 1
stderr:
/home/aos-dev/.conan2/p/b/libud3200b2f38cb14/b/build-relwithdebinfo/meson-private/tmp8pheyca2/testfile.c:4:12: error: #error "Header 'sys/capability.h' could not be found"
    4 |           #error "Header 'sys/capability.h' could not be found"
      |            ^~~~~
-----------
Has header "sys/capability.h" : NO
 
../src/meson.build:729:8: ERROR: Problem encountered: POSIX caps headers not found

valgur:
@vithalsm Should be fixed now. Give it another try.

vithalsm:
> @vithalsm Should be fixed now. Give it another try.

@valgur Yes, its fixed and could build libudev successfully.
But, now, dpdk is failing as it can't find '-ludev'.  Again, I don't see INC and LIB paths present in the command line.  Something similar to libudev not finding libcap.  Did you also observe and/or fixed same?

vithalsm:
> > @vithalsm Should be fixed now. Give it another try.
> 
> @valgur Yes, its fixed and could build libudev successfully. Thanks again. But, now, dpdk is failing as it can't find '-ludev'. Again, I don't see INC and LIB paths present in the command line. Something similar to libudev not finding libcap. Did you also observe and/or fixed same?

@valgur may be if you can attempt to cross compile your 'dpdk' with x86_64 build and target, would be helpful.  We are getting into cyclical issues. :)

valgur:
@vithalsm i think the `with_libibverbs=False` itself might have been what fixed it for you, somehow. `libcap` is only used in dpdk libraries that link against `libmlx4` and `libmlx5` from rdma-core.

If you want to keep using dpdk with rdma-core, I suggest adding
```
[replace_requires]
libudev/system: libudev/255.15
```
to your profile to override the version everywhere.

vithalsm:
> @vithalsm i think the `with_libibverbs=False` itself might have been what fixed it for you, somehow. `libcap` is only used in dpdk libraries that link against `libmlx4` and `libmlx5` from rdma-core.
> 
> If you want to keep using dpdk with rdma-core, I suggest adding
> 
> ```
> [replace_requires]
> libudev/system: libudev/255.15
> ```
> 
> to your profile to override the version everywhere.

@valgur thank you for this suggestion, was looking for exactly like this.  It helped.

Issue #14440, opened 2022-11-26
[package] minizip/1.2.13: Windows MinGW Clang fails to compile


LargoUsagi:
### Description

This compiles using GCC and Clang under Arch Linux, and GCC using mingw on windows, but I had the preference of using clang through mingw.

### Package and Environment Details

* Package Name/Version: **minizip/1.2.13**
* Operating System+version: **Windows 10**
* Compiler+version: **Clang 15**
* Conan version: **conan 1.54.0**
* Python version: **Python 3.10.8**

### Conan profile

[settings]
os=Windows
os_build=Windows
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=clang
compiler.version=15
compiler.libcxx=libstdc++11
[options]
[build_requires]
[env]
CC=C:/msys64/mingw64/bin/clang
CXX=C:/msys64/mingw64/bin/clang++

### Steps to reproduce

conan install . --build=missing --profile=gcc -if=cmake-build-release -s build_type=Release
conan install . --build=missing --profile=gcc -if=cmake-build-debug -s build_type=Debug

```
[requires]
glfw/3.3.8
glm/cci.20220420
assimp/5.2.2

[generators]
cmake
```

### Logs

<details><summary>Click to expand log</summary>

```
minizip/1.2.13: Copying sources to build folder
minizip/1.2.13: Building your package in C:\Users\LargoUsagi\.conan\data\minizip\1.2.13\_\_\build\3822ec74d7dc1a129150f4f5e1c981542581f77c
minizip/1.2.13: Generator txt created conanbuildinfo.txt
minizip/1.2.13: Calling generate()
minizip/1.2.13: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
minizip/1.2.13: Preset 'debug' added to CMakePresets.json. Invoke it manually using 'cmake --preset debug'
minizip/1.2.13: If your CMake version is not compatible with CMakePresets (<3.19) call cmake like: 'cmake <path> -G "MinGW Makefiles" -DCMAKE_TOOLCHAIN_FILE=C:\Users\LargoUsagi\.conan\data\minizip\1.2.13\_\_\build\3822ec74d7dc1a129150f4f5e1c981542581f77c\build\gener
[...]

PR #26165, opened 2024-12-11
zlib: reflect correct library name on Windows mingw

jcar87:
### Summary
Changes to recipe:  zlib

#### Motivation
Close https://github.com/conan-io/conan-center-index/issues/14440

#### Details
* Fix an issue where if zlib is built with clang on msys2/clang64 - the library in cpp_info does not match the generated files
* Note that this is a legacy bug as a result of https://github.com/conan-io/conan-center-index/pull/10771, perhaps to work around a Conan limitation at the time. Nowadays we wouldn't accept a recipe that modifies the generated library files, and cpp_info should merely reflect the generated files - but we also can't undo that now, because renaming libraries across recipe revisions would cause widespread issues.

* Note that as a result of the aforementioned PR, windows + msys2/clang + zlib with shared=True will not build - no attempt is made in this PR to fix this.

tested on Windows to work with:
* `msvc` (shared and static)
* `clang-cl` from an msvc installation (shared and static)
* `gcc14` mingw (shared and static)
* `clang19` msys2 (static only, for shared lld does not accept `--version-script`)


jcar87:
cc @mhthies

memsharded (approved):

mhthies:
Thanks for fixing and merging this so quickly! :)

PR #26182, opened 2024-12-13
Security / Vulnerability Scanner tool addition / SBOM : trivy and syft

roalter:
### Summary
Changes to recipe:  **trivy/0.58.0** and **syft/1.18.0**

#### Motivation
SBOM and Vulnerability scanner are two tools that can improove the security of your projects. As those tools are often used during the integration process, it becomes handy when a tool could be included by tool_requires directly (as cmake can be).

#### Details
Those two recipes will ass syft and trivy to the bin path, downloading the binary GO package. No recompiliing

---

### DIFF
--- a/recipes/syft/all/conandata.yml
+++ b/recipes/syft/all/conandata.yml
@@ -1,0 +1,10 @@
+sources:
+  "1.18.0":
+    Linux:
+      x86_64:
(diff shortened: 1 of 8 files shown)


jcar87:
Hi @roalter  - thanks for your pull requests.
Please split these into two separate ones, so that we can evalaute each separately.

If you could additional information on how this is used by C++ developers - that would be great - as we want to ensure recipes we add to Conan Center are motivated by community interest. Thanks!

PR #26190, opened 2024-12-15
package_templates: remove a misplaced config.yml

valgur:
Remove a `cmake_package/all/test_package/config.yml` accidentally added in #25849.

@uilianries


uilianries (approved):
Good catch! Thank you!

PR #17545, opened 2023-05-12
mppp: update fmt, fix compilation error on with_fmt=True

toge:
Specify library name and version:  **mppp/0.27**

When I created mppp recipe with `with_fmt=True`, there is a compilation error following:
> Could not find a configuration file for package "fmt" that is compatible

It may be caused by `find_package(fmt)` with old version(6.2).
```cmake
set(_MPPP_MIN_FMT_VERSION "6.2")
if(MPPP_WITH_FMT)
    find_package(fmt ${_MPPP_MIN_FMT_VERSION} REQUIRED CONFIG)
```

Then I created the patch to remove version number from `find_package`.

---

- [x] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [x] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [x] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #26191, opened 2024-12-15
mppp: add version 2.0.0

toge:
### Summary
Changes to recipe:  **mppp/2.0.0**

#### Motivation
Major version up.
In 2.0.0, `with_arb` is deleted and `with_flint` is added.

#### Details
https://github.com/bluescarni/mppp/releases/tag/v2.0.0
https://github.com/bluescarni/mppp/compare/v1.0.3...v2.0.0

### Conan maintainer changes:
* Stop publishing new revisions for older versions, keeping only the most recent 1.x and 2.x
* Remove patches for the dropped versions
* Use `deps.set_property("fmt", "cmake_config_version_compat", "AnyNewerVersion")` instead of a patch to work around error caused by version requirement in `find_package` where the Conan-generated package config does not handle this correctly

---


ErniGH (approved):
LGTM Thanks!

recipes/mppp/all/patches/2.0.0-0002-remove-fmt-version.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 3a5400a..7907741 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -406,7 +406,7 @@ endif()
+ # to print 128-bit integers.
+ set(_MPPP_MIN_FMT_VERSION "6.2")
+ if(MPPP_WITH_FMT)
+-    find_package(fmt ${_MPPP_MIN_FMT_VERSION} REQUIRED CONFIG)
```
jcar87, line 9: is this patch is still needed? the minimum version listed there is 6.2 - if CMake fails because the version is newer than this, you may have some luck with the following, without the need of a patch:

```
deps.set_property("fmt",  "cmake_config_version_compat", "AnyNewerVersion")
```

toge: @jcar87 
Thank you for your review.
Because there are compilation errors without patches, I add cmake_config_version_compat definition.
It is more easy way. 
Thanks!

ErniGH:
Hello @toge and @jcar87, I've been reviewing this PR. I tried compiling it without the latest fix, and it works just as well. I also tried compiling version `0.27` without using the patch that fixes the `fmt` version, which was why this patch was originally added, and it also worked without any issues.

(0.27 with the fmt patch) https://github.com/conan-io/conan-center-index/pull/17545

Digging deeper, I found that the bug was described in https://github.com/conan-io/conan/issues/13809 and later resolved in https://github.com/conan-io/conan/pull/14176.

It seems that this patch might not be necessary since June 2023.
If that's the case, we could do a good cleanup of this recipe!

jcar87:
> It seems that this patch might not be necessary since June 2023. If that's the case, we could do a good cleanup of this recipe!

Not quite!  It may be that the patch is unnecessary, however the fix in https://github.com/conan-io/conan/pull/14176/files only kicks in if the upstream recipe (`fmt` in this case), sets the `cmake_config_version_compat` property in its package_info - and it currently doesn't - so even current versions of Conan could still faill with the issue here: https://github.com/conan-io/conan-center-index/pull/17545 - because the new default `SameMajorVersion` is equivalent to the behaviour prior to that fix

ErniGH:
@jcar87  You're right, the fix is still necessary, but we can remove the patches and do a bit of cleanup.

jcar87 (approved):

PR #26192, opened 2024-12-15
odpi: new recipe

inPhraZ:
### Summary
New recipe:  **odpi/5.4.1**

#### Motivation
Oracle Database Programming Interface for C (ODPI-C) is an open source library of C code that simplifies access to Oracle Database for applications written in C or C++

Upstream repository: https://github.com/oracle/odpi

---

### DIFF
--- a/recipes/odpi/all/conanfile.py
+++ b/recipes/odpi/all/conanfile.py
@@ -1,0 +1,75 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.files import copy, chdir, get, rm, rmdir
+from conan.tools.gnu import Autotools, AutotoolsDeps, AutotoolsToolchain, PkgConfigDeps
+from conan.tools.layout import basic_layout
+import os
+
+
+required_conan_version = ">=2.0.9"
+
+class ODPIConan(ConanFile):
+    name = "odpi"
+    description = "Oracle Database Programming Interface for Drivers and Applications"
+    license = ("UPL-1.0", "Apache-2.0")
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/oracle/odpi"
+    topics = ("oracle", "database", "oci")
+    package_type = "library"
+    settings = "os", "arch", "compiler", "build_type"
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    languages = ["C"]
+    implements = ["auto_shared_fpic"]
+
+    def configure(self):
+        if self.options.shared:
+            self.options.rm_safe("fPIC")
+        self.settings.rm_safe("compiler.cppstd")
+        self.settings.rm_safe("compiler.libcxx")
+
+    def layout(self):
+        basic_layout(self)
+
+    def validate(self):
+        if self.settings.os != "Linux":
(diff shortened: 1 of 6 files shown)


ErniGH:
Short summary:
* Shared-library "A sample Makefile is provided if you wish to build ODPI-C as a shared library" https://github.com/oracle/odpi/blob/d97327db237d2efd10905f4339be63ca20f12e38/doc/src/user_guide/installation.rst#L14
* Add the missing layout folder
* Validate is no longer required
* New build_requiremets for Windows support
* PkgConfigDeps and AutotoolsDeps are not required because the recipe doesn't have any dependencies
* Windows uses a special Makefile.win32 
* Copy libs dlls and headers on Windows
* Remove self.cpp_info.set_property("pkg_config_name", "odpi") we can take the recipe name and upstream don't define the pkg_config_name
* Small change on the test to get some console output
* Use fix_apple_shared_install_name
* ~New patch from: https://github.com/oracle/odpi/issues/196~ - Fixed on version 5.6.0

ErniGH (approved):

uilianries (approved):
LGTM.

The reported issue resulted in a merged patch, adding flexibility when building odpi project. It fixes the Linux and Mac builds by giving Conan the chance to control the compiler flags.

The upstream also gives support for Windows and, it's documented how to do that.

recipes/odpi/all/conanfile.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
+
+    def generate(self):
+        tc = AutotoolsToolchain(self)
+        if self.settings.os == "Linux":
+            tc.extra_cflags = ["-fPIC"]
+            tc.extra_ldflags = ["-shared"]
```
jcar87, line 41: both of these are already default here:
https://github.com/oracle/odpi/blob/main/Makefile#L57-L61

perhaps I'm missing something, why is it needed here?

uilianries: The change from that patch uses `?=`, which is overridden by the CFLAGS and LDFLAGS produced by AutotoolsToolchain. That generated Conan toolchain will contain:

```
export CFLAGS="$CFLAGS -m64 -O3"
export LDFLAGS="$LDFLAGS -m64"
```

Because I'm building Release + x86_64.

Still, the package type is `shared-library`, because the upstream only generates shared and the documentation only mentions shared library (even the libraries names are hardcoded to shared). So the AutoToolchain does not pass `-fPIC`, because it is expected to have it listed. In case using regular package type `library` and building shared=True, we will have:

```
export CFLAGS="$CFLAGS -m64 -fPIC -O3"
export LDFLAGS="$LDFLAGS -m64"
```

Now the `-fPIC` is there, but we still miss `-shared`. 

-----

Another approach would be using append (`+=`) instead of replacing defaults, but it will result in conflict with Debug/Release flags and optimization flags too, as the original always use `-g -O2`

uilianries: Added commit [023e757](https://github.com/conan-io/conan-center-index/pull/26192/commits/023e757c5bdf017c743de0e5e691aafb280e7dc5) with a few comments in the recipe. Plus, I bumped to the release 5.6.0, where they have adopted the patch for Makefile, so we no longer need to keep it here.

ErniGH (approved):

uilianries (approved):
LGTM. 

@inPhraZ Thank you for your PR 

And a big thanks to the Oracle team for the quick answer, fix and release! 👏

PR #26204, opened 2024-12-16
libnetfilter_log: add new recipe (#26202)

### LINKED ISSUE #26202: [request] libnetfilter_log/1.0.2
### Package Name/Version

libnetfilter_log/1.0.2

### Webpage

https://www.netfilter.org/projects/libnetfilter_log/index.html

### Source code

https://www.netfilter.org/projects/libnetfilter_log/files/libnetfilter_log-1.0.2.tar.bz2

libnetfilter_log is a userspace library providing interface to packets that have been logged by the kernel packet filter. This library has been previously known as libnfnetlink_log.

nikitasinys:
### Summary
Changes to recipe:  **libnetfilter_log/1.0.2**, fixes #26202

#### Motivation
To handle nftable logs in userspace application (this library provides interface to packets that have been logged by the kernel packet filter)

#### Details
Question about making libmnl and libnfnetlink as transitive dependencies:
other similar libraries (libnetfilter_queue, libnetfilter_conntrack) have next requirements:
```
        self.requires("libmnl/1.0.4", transitive_headers=True, transitive_libs=True)
        self.requires("libnfnetlink/1.0.2", transitive_headers=True) 
```
but I didn't found in headers any includes of these dependencies

### Maintainers changes

* Removed libnetfilter_conntrack as direct dependency: It's an optional dependency: https://git.netfilter.org/libnetfilter_log/tree/configure.ac?h=libnetfilter_log-1.0.2#n53. And Linux distros like [Debian](https://packages.debian.org/bookworm/libnetfilter-log-dev) and [Arch](https://archlinux.org/packages/extra/x86_64/libnetfilter_log/) don't use it by default. Plus, the [official website](https://netfilter.org/projects/libnetfilter_log/index.html) does not mention it as a dependency. In case really needing libnetfilter_conntrack as a dependency, please, comment about.

* Increased Conan 2.18 as minimal version: So we can use self.cpp_info.set_property("pkg_config_name", "none") and generate the .pc files according to the upstream

* Simplified the test package to avoid opening any socket or touching the networking.

---


nikitasinys:
Hi @uilianries,
Can you please look at this PR? It was hanging in pre-checks for month (the same as #26200) and now already for weeks failed with message "Conan Center maintainers need to approve this CI run".

recipes/libnetfilter_log/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+    package_type = "library"
+    languages = ["C"]
+    implements = ["auto_shared_fpic"]
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires("libmnl/1.0.4")
+        self.requires("libnetfilter_conntrack/1.0.9")
```
AbrilRBS: I see that autotools is looking for this package when configuring, but then it's not present in either of the upstream .pc files even though it seems to be used at some points in the code. Do you have any insight into it? Thanks!

recipes/libnetfilter_log/all/conanfile.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def package(self):
+        copy(self, "COPYING", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        autotools = Autotools(self)
+        autotools.install()
+
+        rm(self, "*.la", os.path.join(self.package_folder, "lib"))
+        rmdir(self, os.path.join(self.package_folder, "share"))
+        rmdir(self, os.path.join(self.package_folder, "etc"))
+        #rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
```
AbrilRBS: To re-add before merging

franramirez688 (approved):

Issue #26197, opened 2024-12-16
[package] libtiff/4.7.0: liblzma isn't used on Linux


Nekto89:
### Description

libtiff package seems to be ignoring liblzma on Linux. Maybe it's caused by case-sensitivity of file system and CMake tries to look for improper name. On Windows I don't see this error.
```
--   LZMA2 support:                      Requested:True Availability:0 Support:FALSE
```
Same error can be found in latest PR with libtiff
https://c3i.jfrog.io/c3i/misc-v2/logs/pr/25314/4-linux-gcc/libtiff/4.7.0//8141500ef0fe16b0b5a7ae18bd1ea1ba6d45de2c-build.txt

### Package and Environment Details

* Package Name/Version: **libtiff/4.7.0**
* Operating System+version: **Linux Ubuntu 16.04**
* Compiler+version: **GCC 11**
* Docker image: **conanio/gcc11-ubuntu16.04**
* Conan version: **conan 2.1.0**
* Python version: **Python 3.7.13**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

### Steps to reproduce

docker run --rm -it -v C:\dev\conan\tmp:/home/conan/logs conanio/gcc11-ubuntu16.04
conan install --requires=libtiff/4.7.0 --build=libtiff/4.7.0 > /home/conan/logs/output.log 2>&1

### Logs

<details><summary>Click to expand log</summary>

```

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=11
os=Linux


======== Computing dependency graph ========
libtiff/4.7.0: Not found in local cache, looking in remotes...
libtiff/4.7.0: Checking remote: conancenter
libtiff/4.7.0: Downloaded recipe revision 0c57e6ba6b5ab09736fdace745bbdbd3
zlib/1.3.1: Not found in local cache, looking in remotes...
zlib/1.3.1: Checking remote: conancenter
zlib/1.3.1: Downloaded recipe revision f52e03ae3d251dec704634230cd806a2
[...]

PR #26209, opened 2024-12-16
libtiff >= 4.5.1: fix case sensitiity issue for discovery of jbig, lzma and zstd

SpaceIm:
### Summary
Changes to recipe:  **libtiff >= 4.5.1**

#### Motivation
lzma may not be enabled during CMake configuration due to case sensitivity issue (in the CMake config file name and `NAME_FOUND` variable) coming from the logic of the recipe since libtiff 4.5.1. jbig and zstd may suffer from the same issue.

#### Details

Since 4.5.1, this recipe renames target name of several dependencies (and config file name for few of them) instead of using a more intrusive patch. But it's partially broken, several dependencies are discovered upstream with a name in find_package() whose casing is slightly different from official one, so it has to be changed as well (well technically find_package() also try to find`<lowercaseName>-config.cmake` when `nameConfig.cmake` is not found, but I prefer to be conservative here, it's quite easy to make a mistake).
In the specific case of lzma, we have to be careful since upstream calls `find_package(liblzma)`, therefore we set `cmake_file_name` of `xz_utils` to `liblzma`, but afterwards upstream relies on a `LIBLZMA_FOUND` variable expected to be set to TRUE by lzma config file, but CMakeDeps defines `liblzma_FOUND` (`<cmake_file_name>_FOUND`) in generated config file, therefore a patch is needed.

See:
https://gitlab.com/libtiff/libtiff/-/blob/v4.5.1/cmake/JBIGCodec.cmake#L30
https://gitlab.com/libtiff/libtiff/-/blob/v4.5.1/cmake/LZMACodec.cmake#L29-32
https://gitlab.com/libtiff/libtiff/-/blob/v4.5.1/cmake/ZSTDCodec.cmake#L31

closes https://github.com/conan-io/conan-center-index/issues/26197

---


SpaceIm:
Ok, I see that some patches of 4.5.0 related to checks of dependencies and still needed in 4.5.1+ have not been propagated in offended PR https://github.com/conan-io/conan-center-index/pull/17946 (see some issues of features discovery inhttps://c3i.jfrog.io/c3i/misc-v2/logs/pr/17946/6-linux-gcc/libtiff/4.5.1//c8e091d122e92e6b94a7b875fb0b03e5b7d485fc-build.txt, like `-- Looking for jbg_newlen - not found` )

jcar87:
> cmake_file_name of xz_utils to liblzma, but afterwards upstream relies on a LIBLZMA_FOUND variable expected to be set to TRUE by lzma config file, but CMakeDeps defines liblzma_FOUND (<cmake_file_name>_FOUND) in generated config file, therefore a patch is needed.

The file `conan-official-xz_utils-variables.cmake` which is set in the xz-utils recipe https://github.com/conan-io/conan-center-index/blob/9a66422e07df06d2c502501de6e00b8b1213b563/recipes/xz_utils/all/conanfile.py#L183 - is already defining this variable, so regardless of the name of the generated files by Conan, it will be set - it's not necessarily bad to re-enforce it, but the issue on its own should be solved by `deps.set_property("xz_utils", "cmake_file_name", "liblzma")`

jcar87 (approved):

PR #26236, opened 2024-12-20
added extra-cmake-modules 6.8.0

boussaffawalid:
### Summary
Changes to recipe:  **extra-cmake-modules/6.8.0**

#### Motivation

* Added new version: 6.8.0
  * Fixed `find_package` line for Qt. Added patch.
* Stopped publishing new revisions for older versions.

#### Details

Conan's qt package does not define config file for each qt component, also <component>_FOUND are not defined.

For example: 
* `find_package(Qt6Core)` does not work with conan, and therefore I had to change it to `find_package(Qt6 COMPONENTS Core)`
* `Qt5Core_FOUND` is not defined by conan, and I had to change it to `if (TARGET Qt5::Core)`

---

### DIFF
--- a/recipes/extra-cmake-modules/all/conandata.yml
+++ b/recipes/extra-cmake-modules/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "6.8.0":
+    url: "https://download.kde.org/stable/frameworks/6.8/extra-cmake-modules-6.8.0.tar.xz"
+    sha256: "ff8a0bf72285bec1768e3acd8f7c665a26d55a1527e96d73e35789dc9f0e3472" 
   "6.2.0":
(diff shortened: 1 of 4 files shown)


jcar87:
Thanks @boussaffawalid !

can't find a motivation or description for the `patches/0001-fix-find_package-qt.patch` patch - please ensure there's sufficient information in the PR description so that we can review this, thanks!

boussaffawalid:
@jcar87 conan's qt package does not define config file for each qt component, also <component>_FOUND are not defined.

For example: 
* `find_package(Qt6Core)` does not work with conan, and therefore I had to change it to `find_package(Qt6 COMPONENTS Core)`
* `Qt5Core_FOUND` is not defined by conan, and I had to change it to `if (TARGET Qt5::Core)`

PR #26245, opened 2024-12-22
opusfile: fix building for Android with http=True

kambala-decapitator:
### Summary
Changes to recipe:  **opusfile/all**

#### Motivation

Fixes building opusfile with `http=True` for Android

#### Details

Build fails due to using C99 features but compiling in C89 mode. I've opened issue in upstream https://github.com/xiph/opusfile/issues/52 as I'm not sure what's the right way to proceed here (if I can blindly replace 89 with 99). In the recipe I replace 89 with 99 only for Android (because again I'm not sure if there will be any side effects on other platforms).

---

### DIFF
--- a/recipes/opusfile/all/conanfile.py
+++ b/recipes/opusfile/all/conanfile.py
@@ -135,6 +135,9 @@ class OpusFileConan(ConanFile):
             msbuild.platform = "Win32" if self.settings.arch == "x86" else msbuild.platform
             msbuild.build(os.path.join(sln_folder, "opusfile.sln"), targets=["opusfile"])
         else:
+            if self.settings.os == "Android":
+                replace_in_file(self, os.path.join(self.source_folder, "configure.ac"), "c89", "c99")
+
             autotools = Autotools(self)
             autotools.autoreconf()
             autotools.configure()
--- a/recipes/opusfile/all/conanfile.py
+++ b/recipes/opusfile/all/conanfile.py
@@ -8,7 +8,7 @@ from conan.tools.layout import basic_layout
 from conan.tools.microsoft import is_msvc, MSBuild, MSBuildDeps, MSBuildToolchain
 import os
 
-required_conan_version = ">=1.54.0"
+required_conan_version = ">=2.4.0"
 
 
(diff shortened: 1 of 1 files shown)


uilianries (dismissed):
@kambala-decapitator Hello! :wave: 

Thank you for pushing PR and alerting the upstream about the case. Indeed I could reproduce your scenario: [opusfile-0.12-android-error.log](https://github.com/user-attachments/files/20103691/opusfile-0.12-android-error.log)

And when applying your patch, it works normally for Android: [opusfile-0.12-android-patched.log](https://github.com/user-attachments/files/20103696/opusfile-0.12-android-patched.log)

I did a small change in your PR to accept `settings.compiler.cstd` in case defined in the profile. 

I found another related issue in the upstream: https://github.com/xiph/opusfile/issues/44

And as you pointed, the [C99](https://en.cppreference.com/w/c/99) is needed due function inline feature.

recipes/opusfile/all/conanfile.py
```diff
@@ -135,6 +135,10 @@ def build(self):
             msbuild.platform = "Win32" if self.settings.arch == "x86" else msbuild.platform
             msbuild.build(os.path.join(sln_folder, "opusfile.sln"), targets=["opusfile"])
         else:
+            if self.settings.os == "Android":
+                cstd = self.settings.get_safe("compiler.cstd", default="99")
+                replace_in_file(self, os.path.join(self.source_folder, "configure.ac"), "c89", f"c{cstd}")
+
```
jcar87:
```suggestion
            if self.settings.os == "Android":
                # See https://github.com/conan-io/conan-center-index/pull/26245#pullrequestreview-2825164395
                cstd = self.settings.get_safe("compiler.cstd", default="99")
                replace_in_file(self, os.path.join(self.source_folder, "configure.ac"), "c89", f"c{cstd}")

```

jcar87:
Thanks @uilianries  for providing the build logs - good for traceability.

the issues are in the Linux headers (the android Linux headers) - not the library itself.

This library appears unmaintained in terms of releases and maintainers acknowledging issues - and under those circumstances, I would discourage use

PR #26256, opened 2024-12-24
Updated cuda-api-wrappers with versions 0.7.1 and 0.8.0

eyalroz:
### Summary
Changes to recipe:  **cuda-api-wrappers**

#### Motivation
* [previous PR](https://github.com/conan-io/conan-center-index/pull/25029) for one the last version was ignored for months
* New version of the library have been released
* Mistype in the CMakeLists.txt for the test package

#### Details
* Updated the recipe with versions [0.7.1](https://github.com/eyalroz/cuda-api-wrappers/releases/v0.7.1) and [0.8.0](https://github.com/eyalroz/cuda-api-wrappers/releases/v0.7.1) 
* fixed a mistype in the test package CMakeLists.txt.

---

### DIFF
--- a/recipes/cuda-api-wrappers/all/conanfile.py
+++ b/recipes/cuda-api-wrappers/all/conanfile.py
@@ -44,6 +44,6 @@ class CudaApiWrappersConan(ConanFile):
     def package_info(self):
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
-
+        self.cpp_info.set_property("cmake_target_name", "cuda-api-wrappers::runtime-and-driver")
         if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.system_libs = ["pthread"]
--- a/recipes/cuda-api-wrappers/all/conanfile.py
+++ b/recipes/cuda-api-wrappers/all/conanfile.py
@@ -3,6 +3,7 @@ import os
 from conan import ConanFile
 from conan.tools.files import get, copy
 from conan.tools.layout import basic_layout
+from conan.tools.scm import Version
 
 required_conan_version = ">=1.52.0"
 
@@ -45,5 +46,8 @@ class CudaApiWrappersConan(ConanFile):
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
         self.cpp_info.set_property("cmake_target_name", "cuda-api-wrappers::runtime-and-driver")
+        if Version(self.version) < "0.7.0":
+            # For previously published versions the target name was different, maintain compatibility
+            self.cpp_info.set_property("cmake_target_aliases", ["cuda-api-wrappers::cuda-api-wrappers"])
         if self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.system_libs = ["pthread"]
--- a/recipes/cuda-api-wrappers/all/test_package/CMakeLists.txt
+++ b/recipes/cuda-api-wrappers/all/test_package/CMakeLists.txt
@@ -4,7 +4,7 @@ project(test_package LANGUAGES CXX)
 find_package(cuda-api-wrappers REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
-target_link_libraries(${PROJECT_NAME} PRIVATE cuda-api-wrappers::cuda-api-wrappers)
+target_link_libraries(${PROJECT_NAME} PRIVATE cuda-api-wrappers::runtime-and-driver)
 target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_11)
 target_compile_definitions(${PROJECT_NAME} PRIVATE
     CUDA_API_WRAPPERS_VERSION=\"${cuda-api-wrappers_VERSION_STRING}\")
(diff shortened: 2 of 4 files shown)


eyalroz:
@valgur  , @franramirez688 : Four months have passed, and I've released a new version. Closed the last PR, opened this one. Please don't let me wait four more months...

recipes/cuda-api-wrappers/all/test_package/CMakeLists.txt
```diff
@@ -1,10 +1,10 @@
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.25)
```
jcar87, line 1: thanks @eyalroz  for your PR.

it's a bit unusual for a test package (and a test package alone) to require a higher version of CMake - because the `find_package(cuda-api-wrappers REQUIRED CONFIG)` call is resolved with files that COnan generates targeting compatible with at least CMake 3.15 

if you could provide more details as to why this is needed (e.g. what error this is fixing, with logs), that would be great, thanks!

eyalroz: Well, first of all, the (non-test) package does requires v3.25; it's just  package doesn't get build - it's a header-only library. The need is for decent and relatively comprehensive CUDA support, and as you may be aware, that's still very much a work in progress at Kitware. Some changes in locating / identifying the "NVIDIA Toolkit Extensions" library were made in v3.25, which make it actually _work_ on Windows.

eyalroz: > So while this project https://github.com/eyalroz/cuda-api-wrappers/blob/master/CMakeLists.txt would require a recent version of CMake, it's not really part of the package process.

I'm the opposite of an expert on conan matters, but still... CMake 3.25 is necessary to properly locate and register the relevant dependencies in the exported `.cmake` files; does that not make it part of the packaging process?

> As for the test package in _this_ repository: the purpose is to have a very simple check of verifying that `find_package(cuda-api-wrappers)` works, for a header-only library, that the compiler can locate a header from the package

But that's not how the library works. `find_package(cuda-api-wrappers)` locates configuration files; and these files require CMake 3.25 . Are you suggesting I write a test package which does a lower-level file search? I suppose I could do that, but it wouldn't really test much of anything.

jcar87: What I mean is that the the test package does not require CMake 3.25, and it should work with whatever version was there before (just for the purposes of the test package).

>  in the exported .cmake files; does that not make it part of the packaging process?
Conan Center does not make use of packaged .cmake files - instead, the relevant information is included in `package_info()` - and then Conan generates the files for the consumer. Those files, as generated by Conan, target CMake 3.15  (and usually, when they are generated by CMake,t hey are also compatible with other versions).

As I mentioned before, the recipe for `cuda-api-wrappers`, not only is it not generating or packaging the exported CMake files - it is not even invoking CMake at all 
https://github.com/conan-io/conan-center-index/blob/b9c01ff294197ca0120d734546d838811324e36a/recipes/cuda-api-wrappers/all/conanfile.py

> But that's not how the library works. find_package(cuda-api-wrappers) locates configuration files; and these files require CMake 3.25 .

The configuration files that the test package are locating, are generated by Conan and they are compatible with CMake 3.15 - 

>  Are you suggesting I write a test package which does a lower-level file search?
Just suggesting that for the sake of the test_package inside Conan Center in this PR - these cmake changes are not necessary - I understand that they are necessary for actual users of the library, but that is not the in the scope of the test package.

eyalroz: > the test package does not require CMake 3.25, and it should work with whatever version was there before 

Ok, removed that requirement and it does indeed work. :-)

Sorry for being a little slow to understand.

jcar87 (approved):

recipes/cuda-api-wrappers/all/conanfile.py
```diff
@@ -44,6 +45,9 @@ def package(self):
     def package_info(self):
         self.cpp_info.bindirs = []
         self.cpp_info.libdirs = []
-
+        self.cpp_info.set_property("cmake_target_name", "cuda-api-wrappers::runtime-and-driver")
+        if Version(self.version) < "0.7.0":
+            # For previously published versions the target name was different, maintain compatibility
+            self.cpp_info.set_property("cmake_target_aliases", ["cuda-api-wrappers::cuda-api-wrappers"])
```
jcar87, line 51: For older versions and just a precaution I'm also adding the previously used target name - to avoid disruption if existing users update the recipe revision.

users wishing to move to a newer version will have to use the correct target name

recipes/cuda-api-wrappers/all/test_package/CMakeLists.txt
jcar87: > Sorry for being a little slow to understand.

Oh no need to apologise! Our goal here is to ensure that the reason for changes proposed in PRs is understood and visible to anyone who reviews the PR now or in the future - as it helps with ongoing maintenance too :)

PR #26260, opened 2024-12-24
minizip: fix building for Android API level < 24

kambala-decapitator:
### Summary
Changes to recipe:  **minizip/all**

#### Motivation

Fixes building for Android API level < 24 again.

#### Details

An improvement over #15532. Not sure how/why it worked with Conan v1 previously, but now another adjustment is needed (maybe something changed in NDK, was testing with r27c now).

I've also opened PR in upstream: https://github.com/madler/zlib/pull/1025, but Conan patch includes only one change from it as otherwise it'd require more patches (that line has changed between versions), the other one is already handled by the injected `IOAPI_NO_64` macro.

Also moved applying patches to `source()`.

---

### DIFF
--- a/recipes/minizip/all/conandata.yml
+++ b/recipes/minizip/all/conandata.yml
@@ -14,11 +14,23 @@ sources:
 patches:
   "1.3.1":
     - patch_file: "patches/minizip.patch"
+    - patch_file: "patches/android-pre-24.diff"
+      patch_description: "Fix building for Android against API level < 24"
+      patch_type: "portability"
   "1.2.13":
     - patch_file: "patches/minizip.patch"
+    - patch_file: "patches/android-pre-24.diff"
+      patch_description: "Fix building for Android against API level < 24"
+      patch_type: "portability"
   "1.2.12":
     - patch_file: "patches/minizip.patch"
     - patch_file: "patches/ioapi.patch"
+    - patch_file: "patches/android-pre-24.diff"
+      patch_description: "Fix building for Android against API level < 24"
+      patch_type: "portability"
   "1.2.11":
     - patch_file: "patches/minizip.patch"
     - patch_file: "patches/miniunz.patch"
+    - patch_file: "patches/android-pre-24.diff"
+      patch_description: "Fix building for Android against API level < 24"
+      patch_type: "portability"
--- a/recipes/minizip/all/conanfile.py
+++ b/recipes/minizip/all/conanfile.py
@@ -52,6 +52,7 @@ class MinizipConan(ConanFile):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
 
     def generate(self):
         tc = CMakeToolchain(self)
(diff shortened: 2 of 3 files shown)


jcar87:
Thanks! 

I can see multiple PRs that aim to target older Android API levels - what's the rationale for this? we're unlikely to support this unless there's a strong reason. As far as I can see:
* API level 24 was introduced in Android 7.0 released in the second half of 2016 - https://developer.android.com/tools/releases/platforms#7.0. It's unlikely that any Android devine (at least, smartphone) released before that is still supported by the manufacturer
* The google play store does not accept anything lower than API level 34 for new apps
* The current version of Android (Android 15) will not allow installing apks that target APIs 24 or older

We're trying to understand what the motivation is - as this imposes a higher maintenance burden on us, which has an impact on the entire community - and while we are appreciative of teams still target very old versions

kambala-decapitator:
Thank you for the comment, @jcar87 

> I can see multiple PRs that aim to target older Android API levels - what's the rationale for this?

our project uses Conan to manage dependencies and supports Android 21+. Currently I'm in the process of migrating to v2, so I'm fixing anything I find in the recipes that prevents me from building packages.

> It's unlikely that any Android devine (at least, smartphone) released before that is still supported by the manufacturer

it's true. But we're trying to make the project accessible to the widest audience possible.

> The google play store does not accept anything lower than API level 34 for new apps

sorry, seems you're confusing SDK version (max version of APIs that can be used, SDK that you build your code againstl Google docs call it _target API level_) and API level (minimum OS version that can run your code). In Apple world it's similar: you build against a modern SDK (say, macOS 15), but your app runs on _deployment target_ (minimum OS version) you specify (currently the earliest possible with the latest Xcode is macOS 10.13).

Btw Conan doesn't model SDK version.

> The current version of Android (Android 15) will not allow installing apks that target APIs 24 or older

I guess you're talking about [this](https://developer.android.com/about/versions/15/behavior-changes-all#minimum-target-api-level). But it's also about _target API level_, i.e. SDK version that you build against, just like I explained above.

Until now the only restriction on the minimum API level (besides what NDK states) was Google disabling its Google Services for old devices, e.g. recently they've dropped the services for API level < 21. And if you don't use them, you can keep lower API level if you need it, e.g. 19 or 16.

> this imposes a higher maintenance burden on us

what kind of maintenance do you mean exactly? AFAIK CCI doesn't build Android packages (just like Apple non-macOS), so its support is entirely on the community.

jcar87:
> what kind of maintenance do you mean exactly? AFAIK CCI doesn't build Android packages (just like Apple non-macOS), so its support is entirely on the community.

By resources I mean:
* CI resources: any PR to change recipes will still be validated on the main platforms (Windows, Linux, Mac) - regardless of whether the changes affect those platforms or not
* Reviewer resources: we still need to validate these changes work
* Ongoing maintenance: it means the Conan team bears the support if anything doesn't work

kambala-decapitator:
@jcar87 thank you for the response!

I can't insist on merging PRs about old Android versions of course, but since they require only a few lines to change, in my view it's affordable. Fortunately, most of the changes can be done on profile level as well, so it's still solvable even without having the changes in the recipes.

If in the end you're going to reject this PR, you could also close #26240 and require dropping the same adjustment in #26245.

jcar87:
Thanks @kambala-decapitator - I think will proceed closing the PRs - not sure it's worth supporting something when all signals point that there's very few users, vendors no long support it and library maintainers don't care to support it either - if there's evidence to the contrary we can always revisit. Happy to hear these things are solvable at the profile level - it's exactly for cases like this

PR #25305, opened 2024-09-18
sfml: Test suport for SFML 3 rc1

AbrilRBS:
### Summary
Changes to recipe:  **sfml/3**

#### Motivation

They recently released the first RC for 3.0, and wanted to get ahead of the curve with it. This probably won't get merged until the official 3.0 release, but it can't hurt to do the work now 

#### Details

The `2.X` changes are just the moved files from the old `all`, I screwed the moving and Git thinks they are new files. The `all` recipe for version 3 is better reviewed without a diff, as it's a new recipe

Things to note:

* This has been succesfully tested in Macos, Linux, Windows and Android - In Macos I have also run most of the examples present in the SFML repo:
  - We found an issue here when running `graphic` dependant tests when `-o="*:shared=False"`, help is appreciated
  - For example, here is the output for the `island` example:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[SFOpenGLView enableKeyRepeat]: unrecognized selector sent to instance 0x142854be0'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000018158a2ec __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x000000018106e158 objc_exception_throw + 60
	2   CoreFoundation                      0x000000018163c56c -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x00000001814f3f3c ___forwarding___ + 1580
	4   CoreFoundation                      0x00000001814f3850 _CF_forwarding_prep_0 + 96
	5   island                              0x000000010412a378 -[SFOpenGLView initWithFrame:fullscreen:highDpi:] + 84
	6   island                              0x000000010412b338 -[SFWindowController setupWindowWithMode:andStyle:] + 180
	7   island                              0x000000010412af30 -[SFWindowController initWithMode:andStyle:andState:] + 344
	8   island                              0x000000010412c980 _ZN2sf4priv15WindowImplCocoaC2ENS_9VideoModeERKNS_6StringEjNS_5StateERKNS_15ContextSettingsE + 108
	9   island                              0x00000001041208b4 _ZN2sf4priv10WindowImpl6createENS_9VideoModeERKNS_6StringEjNS_5StateERKNS_15ContextSettingsE + 568
	10  island                              0x000000010411f0f8 _ZN2sf6Window6createENS_9VideoModeERKNS_6StringEjNS_5StateERKNS_15ContextSettingsE + 68
[...]

PR #26276, opened 2024-12-31
sfml: add version 3.0.0

OMGtechy:
changes:
- cmake required version 3.1 -> 3.24
- added SFML 3.0.0 source zip + sha256
- updated test_package.cpp to work with both SFML 3.0.0

The cmake minimum version was changed to 3.24 because:
- Linux SFML 3.0.0 requires 3.22
- Windows SFML 3.0.0 requires 3.24

It was deemed simpler to just have a single version required for the sake of the test package.

More info: https://github.com/SFML/SFML/blob/7f1162dfea4969bc17417563ac55d93b72e84c1e/CMakeLists.txt#L1-L9

### Summary

#### Details

---

### DIFF
--- a/recipes/sfml/all/conanfile.py
+++ b/recipes/sfml/all/conanfile.py
@@ -91,9 +91,11 @@ class SfmlConan(ConanFile):
         tc.variables["SFML_BUILD_AUDIO"] = self.options.audio
         tc.variables["SFML_INSTALL_PKGCONFIG_FILES"] = False
         tc.variables["SFML_GENERATE_PDB"] = False
-        tc.variables["SFML_USE_SYSTEM_DEPS"] = True
+        tc.variables["SFML_USE_SYSTEM_DEPS"] = Version(self.version) < "3.0.0"
         tc.variables["WARNINGS_AS_ERRORS"] = False
-        if Version(self.version) >= "2.6.0":
+        if Version(self.version) >= "3.0.0":
+            tc.variables["CMAKE_CXX_STANDARD"] = 17
+        elif Version(self.version) >= "2.6.0":
             tc.variables["CMAKE_CXX_STANDARD"] = 11
         if is_msvc(self):
             tc.variables["SFML_USE_STATIC_STD_LIBS"] = is_msvc_static_runtime(self)
--- a/recipes/sfml/all/test_package/CMakeLists.txt
+++ b/recipes/sfml/all/test_package/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.24)
 project(test_package LANGUAGES CXX)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
--- a/recipes/sfml/all/conandata.yml
+++ b/recipes/sfml/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "3.0.0":
+    url: "https://www.sfml-dev.org/files/SFML-3.0.0-sources.zip"
+    sha256: 8cc41db46b59f07c44ecf21c74a0f956d37735dec9d90ff4522856cb162ba642
   "2.6.2":
     url: "https://www.sfml-dev.org/files/SFML-2.6.2-sources.zip"
     sha256: "19d6dbd9c901c74441d9888c13cb1399f614fe8993d59062a72cfbceb00fed04"
--- a/recipes/sfml/all/test_package/test_package.cpp
+++ b/recipes/sfml/all/test_package/test_package.cpp
@@ -1,3 +1,5 @@
+#include <iostream>
+
 #include <SFML/System.hpp>
 
 #ifdef SFML_WITH_WINDOW
@@ -16,10 +18,16 @@
 int main()
 {
     sf::Clock clock;
-    clock.getElapsedTime().asSeconds();
+    std::cout << "Seconds elapsed: "
+              << clock.getElapsedTime().asSeconds()
+              << std::endl;
 
 #ifdef SFML_WITH_WINDOW
+ #if SFML_VERSION_MAJOR >= 3
+    sf::VideoMode videoMode({720, 480});
+ #else
     sf::VideoMode videoMode(720, 480);
+ #endif
 #endif
 
 #ifdef SFML_WITH_GRAPHICS
@@ -29,13 +37,14 @@ int main()
 
 #ifdef SFML_WITH_NETWORK
     sf::TcpListener listener;
-    listener.isBlocking();
+    std::cout << "TCP blocking: "
+              << listener.isBlocking()
+              << std::endl;
 #endif
 
 #ifdef SFML_WITH_AUDIO
     sf::SoundBuffer buffer;
-    sf::Sound sound;
-    sound.setBuffer(buffer);
+    sf::Sound sound(buffer);
 #endif
 
     return 0;
--- a/recipes/sfml/config.yml
+++ b/recipes/sfml/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "3.0.0":
+    folder: all
   "2.6.2":
     folder: all
   "2.6.1":


OMGtechy:
I saw some other attempts that looked abandoned or were more complex in implementation:
- https://github.com/conan-io/conan-center-index/pull/26266
- https://github.com/conan-io/conan-center-index/pull/25305

The hope that this will be simpler, avoid any patches and Just Work :tm:

OMGtechy:
RE multiple being developed at the same time (see https://github.com/conan-io/conan-center-index/pull/26266) - happy to discuss the merits of each / take the judgement of maintainers or whatever.

They each appear to have their own merits. Perhaps they could be combined?

Pros of this one:
- doesn't have a separate 3.0.0 directory
- doesn't use any patches
- less code duplication (due to lack of 3.0.0 directory)
- still checks for sfml-main

Pros of alternative:
- doesn't change default for SFML_USE_SYSTEM_DEPS

[ feel free to add anything I missed here - I am almost certainly bias as author ]

Sil3ntStorm:
The sole reason mine exists was that I require SFML 3 for a project and it wasn't available on conan, nor did I see a PR for it. So I adjusted and added SFML 3 and figured I might as well upstream it for the benefits of others after some preliminary testing.

I didn't knowingly remove anything that didn't produce errors for me (or later anything that created linter warnings in CCI). So if I removed sfml-main then it caused errors and didn't appear to be required.
I also didn't look into the USE_SYSTEM_DEPS variable much, but it sounds like something CCI likes to do to ensure things are consumed from conan rather than something else. Which is why I kept it and went the patch approach, which is only a minor patch.

recipes/sfml/all/conanfile.py
```diff
@@ -51,6 +52,10 @@ def configure(self):
     def layout(self):
         cmake_layout(self, src_folder="src")
 
+    def build_requirements(self):
+        if Version(self.version) >= "3.0.0":
+            self.tool_requires("cmake/[>=3.24]")
```
ChrisThrasher, line 57: 3.24 is only required on Windows. All other OSes only require 3.22.

recipes/sfml/all/conanfile.py
```diff
@@ -69,7 +74,18 @@ def requirements(self):
             if Version(self.version) >= "2.6.0":
                 self.requires("minimp3/cci.20211201")
 
+    @property
+    def _min_cppstd(self):
+        if Version(self.version) >= "3.0.0":
+            return 17
+        elif Version(self.version) >= "2.6.0":
+            return 11
```
ChrisThrasher, line 82: Version 2.6 and prior requires C++03, not C++11.

recipes/sfml/all/conanfile.py
```diff
@@ -91,10 +107,12 @@ def generate(self):
         tc.variables["SFML_BUILD_AUDIO"] = self.options.audio
         tc.variables["SFML_INSTALL_PKGCONFIG_FILES"] = False
         tc.variables["SFML_GENERATE_PDB"] = False
-        tc.variables["SFML_USE_SYSTEM_DEPS"] = True
+        tc.variables["SFML_USE_SYSTEM_DEPS"] = Version(self.version) < "3.0.0"
```
ChrisThrasher, line 110: Why change this? A package manager probably wants to use its own dependencies instead of the dependencies bundled with the library.

recipes/sfml/all/conanfile.py
```diff
@@ -91,10 +107,12 @@ def generate(self):
...
         tc.variables["WARNINGS_AS_ERRORS"] = False
-        if Version(self.version) >= "2.6.0":
-            tc.variables["CMAKE_CXX_STANDARD"] = 11
+
+        if self.settings.compiler.cppstd is not None:
+            tc.cache_variables["CMAKE_CXX_STANDARD"] = self._min_cppstd
```
ChrisThrasher, line 114: There is no need to set this. The CMake script already uses `target_compile_features` to set the proper C++ standard version.

recipes/sfml/all/test_package/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.24)
```
ChrisThrasher, line 1: Building SFML requires 3.22 (or 3.24 on Windows as mentioned above). Using SFML doesn't require anything particularly new. 3.1 is probably too old but something like 3.10 or 3.16 is probably sufficient. 3.16 or 3.22 are reasonable choices since it's rare you find an OS still in use today that ships with anything older than that.

ChrisThrasher (changes requested):

recipes/sfml/all/conanfile.py
OMGtechy: Have I misread?

The  code does what you say, with the exception of being inclusive of 2.6.0. That said - the prior version did the same for 2.6.0.

recipes/sfml/all/conanfile.py
OMGtechy: It was a pain getting things to work without. One ends up having to patch how things are searched for.

Apparently, in more recent versions of conan, it's better at building them. This resolved all of the package finding issues. I imagine this will be a more consistent build as a result, but that's just conjecture on my part.

recipes/sfml/all/conanfile.py
OMGtechy: It does not build for MSVC without (which is what lead to adding this)

recipes/sfml/all/test_package/CMakeLists.txt
OMGtechy: This is just the test package rather than what's shipped. But for the main package, you're right, it should help those using --build=always/missing.

recipes/sfml/all/conanfile.py
OMGtechy: Fixed in a6b8e39528c02a2aa8f3397f7be3ba7814d9159c

recipes/sfml/all/test_package/CMakeLists.txt
OMGtechy: Package part fixed in a6b8e39528c02a2aa8f3397f7be3ba7814d9159c

recipes/sfml/all/conanfile.py
```diff
@@ -206,8 +225,13 @@ def objc():
...
         }
+
         if self.settings.os in ["Windows", "Android", "iOS"]:
-            sfml_main_suffix = "-d" if self.settings.build_type == "Debug" else ""
+            if Version(self.version) < "3.0.0":
+                sfml_main_suffix = "-d" if self.settings.build_type == "Debug" else ""
+            else:
+                sfml_main_suffix = "-s"
+
             sfmlmain_libs = [f"sfml-main{sfml_main_suffix}"]
```
JackCarterSmith, line 235: "main" component seem to be always produce a static "-s" in output, I think you should force the suffix for this one.

recipes/sfml/all/conanfile.py
```diff
@@ -69,7 +75,18 @@ def requirements(self):
             if Version(self.version) >= "2.6.0":
                 self.requires("minimp3/cci.20211201")
```
JackCarterSmith, line 77: openal-soft deps have been dropped and replaced by "miniaudio/0.11.21" since v3.
"ogg/1.3.5" for the audio component and "glad/0.1.36" for window component are missing too.

JackCarterSmith (changes requested):

recipes/sfml/all/conanfile.py
eXpl0it3r: FYI: Will be reverted with SFML 3.0.1 

https://github.com/SFML/SFML/pull/3431

recipes/sfml/all/conanfile.py
eXpl0it3r: Glad shouldn't really be listed, as it's an implementation detail of sorts. You can't just use your own glad headers with SFML, as we've made them work for SFML while allowing users to provide their own glad headers separately without breaking SFML.

recipes/sfml/all/conanfile.py
JackCarterSmith: In this case we should remove the "-s" from suffix.

recipes/sfml/all/conanfile.py
JackCarterSmith: Thanks for the clarification. I don't use glad in my applications, I preferred to include the dependency in case.

JackCarterSmith:
Any news about a "possible" 3.x release?

recipes/sfml/all/test_package/conanfile.py
```diff
@@ -15,8 +16,13 @@ def layout(self):
     def requirements(self):
         self.requires(self.tested_reference_str)
 
+    def build_requirements(self):
+        if Version(self.tested_reference_str) >= "3.0.0":
+            self.tool_requires("cmake/[>=3.24]")
```
AnatoliiShablov, line 21: Will this continue to work with sfml 2? There will be no dependency on cmake 3.24, but in test package 3.24 is minimum required version

JackCarterSmith: I suppose this affect only the v3 of the package, but the v2 seem to don't use cmake-conan at least.

OMGtechy:
Closing this - feel free to fork and continue - I just don't think I'll be getting back to it any time soon.

PR #26280, opened 2024-12-31
libnfs: add version 6.0.2

toge:
### Summary
Changes to recipe:  **libnfs/6.0.2**

#### Motivation
There are several improvements since 5.0.3.

#### Details
https://github.com/sahlberg/libnfs/compare/libnfs-5.0.3...libnfs-6.0.2

---

### DIFF
--- a/recipes/libnfs/all/conanfile.py
+++ b/recipes/libnfs/all/conanfile.py
@@ -76,7 +76,7 @@ class LibnfsConan(ConanFile):
         self.cpp_info.set_property("cmake_file_name", "libnfs")
         self.cpp_info.set_property("cmake_target_name", "libnfs::libnfs")
         self.cpp_info.set_property("pkg_config_name", "libnfs")
-        if self.settings.os == "Windows" and self.options.shared:
+        if self.settings.os == "Windows":
             self.cpp_info.libs = ["libnfs"]
         else:
             self.cpp_info.libs = ["nfs"]
--- a/recipes/libnfs/all/conanfile.py
+++ b/recipes/libnfs/all/conanfile.py
@@ -76,7 +76,7 @@ class LibnfsConan(ConanFile):
         self.cpp_info.set_property("cmake_file_name", "libnfs")
         self.cpp_info.set_property("cmake_target_name", "libnfs::libnfs")
         self.cpp_info.set_property("pkg_config_name", "libnfs")
-        if self.settings.os == "Windows":
+        if Version(self.version) >= "6.0.2" and self.settings.os == "Windows":
             self.cpp_info.libs = ["libnfs"]
         else:
             self.cpp_info.libs = ["nfs"]
--- a/recipes/libnfs/all/conanfile.py
+++ b/recipes/libnfs/all/conanfile.py
@@ -1,6 +1,6 @@
 from conan import ConanFile
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
-from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir, rename
 from conan.tools.scm import Version
 import os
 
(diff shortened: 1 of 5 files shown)


recipes/libnfs/all/conanfile.py
```diff
@@ -59,6 +69,8 @@ def package(self):
         copy(self, "LICENCE*.txt", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
         cmake = CMake(self)
         cmake.install()
+        if Version(self.version) >= "6.0.2" and self.settings.os == "Windows" and self.options.shared:
+            rename(self, os.path.join(self.package_folder, "bin", "nfs.dll"), os.path.join(self.package_folder, "bin", "libnfs.dll"))
```
jcar87, line 73: This is odd - on Windows, consumers will link against the import library (`.lib`), which embeds the name of the expected `.dll` file -  does renaming the DLL not break at runtime because the correct DLL cannot be found?

toge: @jcar87 
Agreed.
However, in 6.0.2, when I build a shared library on Windows, the test_package fails to run due to lack of dll.
I will look into the reason and report back to upstream if necessary.

jcar87: Hi @toge ! Thanks so much for looking into it - any updates? anything we can do?

toge: @jcar87 
Upstream updated on this matter.
https://github.com/sahlberg/libnfs/commit/87c2b2a3dc605d4c7da932b2f88b002df722642a

I created the patch file about it instead of renaming dll file name.
Could you review it again?

recipes/libnfs/all/conanfile.py
```diff
@@ -66,6 +76,9 @@ def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "libnfs")
         self.cpp_info.set_property("cmake_target_name", "libnfs::libnfs")
         self.cpp_info.set_property("pkg_config_name", "libnfs")
-        self.cpp_info.libs = ["nfs"]
+        if Version(self.version) >= "6.0.2" and self.settings.os == "Windows":
+            self.cpp_info.libs = ["libnfs"]
+        else:
+            self.cpp_info.libs = ["nfs"]
```
uilianries, line 82:
```suggestion
         self.cpp_info.libs = ["nfs"]
```

Hello! It works on Windows, using version 2.0.6 with the patch, I just tested locally running Conan 2.14:

- [libnfs-6.0.2-windows-static.log](https://github.com/user-attachments/files/19485313/libnfs-6.0.2-windows-static.log)
- [libnfs-6.0.2-windows-shared.log](https://github.com/user-attachments/files/19485323/libnfs-6.0.2-windows-shared.log)

uilianries: @toge Thank you for informing. I'll re-check, because it really worked for me.

jcar87: I think @uilianries  - you may have meant building _without_ the patch?

jcar87:
I've had a chance to look at this issue.

Without any patching, the scripts for `6.0.2` generate `nfs.lib` (import library) and `nfs.dll` (DLL). 
However, binaries linked with `nfs.lib` will encode a dependency on `libnfs.dll` (with the prefix), which is not in the package and thus will fail at runtime. This is presumably caused by this file: https://github.com/sahlberg/libnfs/blob/18c5c73ee88bb7dc8da0d55dc95164bb77e49dc6/lib/libnfs-win32.def#L1 
the `LIBRARY` statement is the name of the DLL that will be encoded in the consumers.


This is a bug in the upstream CMake scripts - so, `6.0.2` + windows + shared + cmake  is broken upstream. 

Now, it is very clear from the upstream sources that the `.dll` file should be named `libnfs.dll` - there are visual studio solutions and command line scripts (and the version.rc file) that point to this being the case.

This deviates from CMake's default behaviour. I suspect https://github.com/sahlberg/libnfs/commit/87c2b2a3dc605d4c7da932b2f88b002df722642a this actually fixes it, although I'd have probably tried to use the CMake target properties only on Windows (e.g. [`RUNTIME_OUTPUT_NAME`](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_NAME.html) and similar)

I'd say the original renaming of the DLL also worked, however, it's difficult for us to review a PR without knowing what the problem is, and our position is always to reflect upstream - rather than to alter reality via the recipe. if upstream is broken, at the very least, it should be reported or clarified - the name of a DLL is quite important because it gets "embedded" in downstream consumers, so we can't make an arbitrary decision without clarifying it first.

jcar87 (dismissed):

jcar87 (approved):

PR #26284, opened 2025-01-01
libultrahdr: Add new library

irieger:
### Summary
Changes to recipe:  **libultrahdr/1.4.0**

#### Motivation
libultrahdr is a new dependency of OpenImageIO (optional one).

#### Details
Adding this new library to extend support of image formats build with OpenImageIO. Definitely needs some more eyes on the patch & original CMakeLists as this is quite ugly and overrides quite a bit of things that I think regularly come from the toolchain.

Especially the following topics stand out, but maybe I missed something else:
* Deactivation of a number of `set` calls setting some options that - to my understanding - should be set by the Toolchain
* It contains a long list of arch-flags that I think should maybe be commented out? (See comment to have it not cluter this overview)
* Not sure if there is any implication from this, but I was forced to change `cmake_policy(SET CMP0091 OLD)` to `cmake_policy(SET CMP0091 NEW)` as otherwise I got a CMake error from the CI.

---


irieger:
```
  if(ARCH STREQUAL "i386")
    add_compile_options(-m32)
    add_compile_options(-march=i386)
    add_compile_options(-mtune=generic)
  elseif(ARCH STREQUAL "amd64")
    add_compile_options(-m64)
    add_compile_options(-march=x86-64)
    add_compile_options(-mtune=generic)
  elseif(ARCH STREQUAL "arm")
    add_compile_options(-march=armv7-a)
    add_compile_options(-marm)
    if(NOT ANDROID_ABI)
      add_compile_options(-mfloat-abi=hard)
    endif()
    add_compile_options(-mfpu=neon-vfpv3)
    add_compile_options(-fno-lax-vector-conversions)
  elseif(ARCH STREQUAL "aarch64")
    add_compile_options(-march=armv8-a)
    add_compile_options(-fno-lax-vector-conversions)
  elseif(ARCH STREQUAL "riscv64")
    add_compile_options(-march=rv64gc)
    add_compile_options(-mabi=lp64d)
  elseif(ARCH STREQUAL "riscv32")
    add_compile_options(-march=rv32gc)
    add_compile_options(-mabi=ilp32d)
  elseif(ARCH STREQUAL "loong64")
    add_compile_options(-march=loongarch64)
    add_compile_options(-mabi=lp64d)
  endif()
```
Would it make sense to comment this out? Shouldn't proper flags be provided by the toolchain? What would happen if one already has some `-march`/`-mtune` flags in the profile? (I for example build my stuff with for arch `x86-64-v3`)

ErniGH (dismissed):
Sorry for the delay. Thanks for the PR the patch seems necessary. I’ve tested the 'with_jpeg' options locally, and it works. 

Thanks!

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
```diff
@@ -0,0 +1,163 @@
...
+ 
+ # side effects
+-if(CMAKE_CROSSCOMPILING AND UHDR_ENABLE_INSTALL)
+-  set(UHDR_ENABLE_INSTALL FALSE) # disable install and uninstall targets during cross compilation.
+-  message(STATUS "Install and uninstall targets - Disabled")
+-endif()
++# if(CMAKE_CROSSCOMPILING AND UHDR_ENABLE_INSTALL)
++#   set(UHDR_ENABLE_INSTALL FALSE) # disable install and uninstall targets during cross compilation.
++#   message(STATUS "Install and uninstall targets - Disabled")
++# endif()
```
jcar87: what is the motivation behind this patch? I may be missing something

If the install targets work fine when cross-building, please try to simplify the patch, preferring something like below


```
-if(CMAKE_CROSSCOMPILING AND UHDR_ENABLE_INSTALL)
+if(0)
```

instead of the whole block

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
```diff
@@ -0,0 +1,163 @@
...
+-    message(FATAL_ERROR "Could NOT find JPEG (missing: JPEG_LIBRARIES JPEG_INCLUDE_DIRS),\
+-                         retry after installing JPEG library at sysroot or try 'cmake -DUHDR_BUILD_DEPS=1'")
+-  else()
+-    message(STATUS "Found JPEG: ${JPEG_LIBRARIES} (found version \"${JPEG_VERSION}\")")
+-  endif()
++if(CONAN_USE_JPEG)
++  find_package(JPEG REQUIRED)
+ endif()
+-
+-if(NOT JPEG_FOUND)
```
jcar87: this conditional is already "false" because due to `JPEG REQUIRED`, if it is not found the execution is stopped - so no need to patch this whole block out

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
```diff
@@ -0,0 +1,163 @@
...
+-  find_package(JPEG QUIET)
+-  if(NOT JPEG_FOUND)
+-    message(FATAL_ERROR "Could NOT find JPEG (missing: JPEG_LIBRARIES JPEG_INCLUDE_DIRS),\
+-                         retry after installing JPEG library at sysroot or try 'cmake -DUHDR_BUILD_DEPS=1'")
+-  else()
+-    message(STATUS "Found JPEG: ${JPEG_LIBRARIES} (found version \"${JPEG_VERSION}\")")
+-  endif()
++if(CONAN_USE_JPEG)
++  find_package(JPEG REQUIRED)
+ endif()
```
jcar87: instead of patching this block, consider handling this in the recipe, by making JPEG required - saves us the patch

```
tc.cache_variables["CMAKE_REQUIRE_FIND_PACKAGE_JPEG"] = True
```

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
```diff
@@ -0,0 +1,163 @@
...
+-        INSTALL_COMMAND ""
+-    )
+-  endif()
++if(CONAN_USE_JPEGTURBO)
++  find_package(libjpeg-turbo REQUIRED)
++  add_library(JPEG::JPEG ALIAS libjpeg-turbo::libjpeg-turbo)
++endif()
++if(CONAN_USE_MOZJPEG)
++  find_package(mozjpeg REQUIRED)
++  add_library(JPEG::JPEG ALIAS mozjpeg::mozjpeg)
```
jcar87: please handle this in the recipe in the `generate()` method, something along the lines of the following might work without having to patch:



```
deps = CMakeDeps(self)

if self.options.jpeg:
    deps.set_property(self.options.jpeg, "cmake_file_name", "JPEG")
    deps.set_property(self.options.jpeg, "cmake_target_name", "JPEG::JPEG")
deps.generate()
```

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
```diff
@@ -0,0 +1,163 @@
...
++target_link_libraries(${UHDR_TARGET_NAME} PRIVATE JPEG::JPEG)
+ set_target_properties(${UHDR_TARGET_NAME}
+                       PROPERTIES PUBLIC_HEADER ultrahdr_api.h)
+ if(BUILD_SHARED_LIBS)
+@@ -767,7 +723,7 @@ if(BUILD_SHARED_LIBS)
+   if(${CMAKE_SYSTEM_NAME} MATCHES "Android")
+     target_link_libraries(${UHDR_TARGET_NAME_STATIC} PRIVATE ${log-lib})
+   endif()
+-  target_link_libraries(${UHDR_TARGET_NAME_STATIC} PRIVATE ${JPEG_LIBRARIES})
++  target_link_libraries(${UHDR_TARGET_NAME_STATIC} PRIVATE JPEG::JPEG)
```
jcar87: `JPEG_LIBRARIES` is already handled by the Conan CMake generators and has a value of `JPEG::JPEG` - so I think this is unnecessary/ redundant, unless I'm missing something?

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
```diff
@@ -0,0 +1,163 @@
...
+   combine_static_libs(${UHDR_CORE_LIB_NAME} ${UHDR_TARGET_NAME_STATIC})
+   if(NOT MSVC)
+     set_target_properties(${UHDR_TARGET_NAME_STATIC}
+@@ -789,7 +745,8 @@ if(UHDR_BUILD_JAVA)
+ endif()
+ 
+ if(UHDR_ENABLE_INSTALL)
+-  if(NOT(MSVC OR XCODE))
++  if(TRUE)
++  #if(NOT(MSVC OR XCODE))
```
jcar87:
```suggestion
-  if(NOT(MSVC OR XCODE))
+  if(TRUE)
```

This is fine, we can already see the `-` line 

However, we probably need a little context justifying this?

recipes/libultrahdr/all/conanfile.py
```diff
@@ -0,0 +1,102 @@
...
+        tc = CMakeToolchain(self)
+
+        # Force-disable fallback to internal dependency builder if no deps found
+        tc.variables["UHDR_BUILD_DEPS"] = False
+        if self.options.with_jpeg == "libjpeg":
+            tc.variables["CONAN_USE_JPEG"] = True
+        elif self.options.with_jpeg == "libjpeg-turbo":
+            tc.variables["CONAN_USE_JPEGTURBO"] = True
+        elif self.options.with_jpeg == "mozjpeg":
+            tc.variables["CONAN_USE_MOZJPEG"] = True
```
jcar87: Please prefer `cache_variables` instead of `variables` (see [docs](https://docs.conan.io/2/reference/tools/cmake/cmaketoolchain.html#variables))

* `variables` are defined in the toolchain file and evaluated during the `project()` call - these are reserved for variables that the CMake documentation intends to be defined in a toolchain file
* `cache_variables` have a value passed with `-D` directly during the invocation, so they have a value since the beginning - and they match the way users would configure the project

also please set `UHDR_BUILD_EXAMPLES` to False

recipes/libultrahdr/all/conanfile.py
```diff
@@ -0,0 +1,102 @@
...
+        elif self.options.with_jpeg == "libjpeg-turbo":
+            self.requires("libjpeg-turbo/3.0.0")
+        elif self.options.with_jpeg == "mozjpeg":
+            self.requires("mozjpeg/4.1.3")
+
+    def validate(self):
+        check_min_cppstd(self, 17)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
```
jcar87, line 62:
```suggestion
    def source(self):
        get(self, **self.conan_data["sources"][self.version], strip_root=True)
        apply_conandata_patches(self)
```

(no need for a `_patch_sources()` method either, unless it has more complex logic)

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
czoido: That's a good idea! If adding the `CMAKE_REQUIRE_FIND_PACKAGE_JPEG` variable that was added in CMake 3.22 we should also add a tool_requires starting in that CMake version: `self.tool_requires("cmake/[>=3.22 <4]")`

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
irieger: But the required will only happen if the library is JPEG, below are the alternative jpeg providers mozjpeg and jpeg-turbo! I thought it would be much cleaner to just have the three blocks that are relevant. Could otherwise leave it in for transparency but also used `if(0)`?

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
irieger: As in the other comment, there is 3 alternative JPEG provider libraries supported, so it only requires the one based on what variable is set. So it still would require to manually ensure the find package for the right package. I personally think it is more readable as is but I'm open to changes, but they must honor this choise between options.

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
irieger: Patching this out was needed if I "cross-compile". For example here a build for macOS x64 on my arm64 mac:
```
conan create all --version=1.3.0 --build=missing -pr:h macos_13.3_x64_release -pr:b macos_13.3_arm64_release
```
This creates an emtpy package (license and conanmanifest/info, no includes or library) when I leave this block in, with it commented out/using `if(0)`. Without this block, I have the library in the package.

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
irieger: Can't fully follow how this would work. Can you explain more or give a link to a more detailed example? I tried to adapt it (btw. I did remove the False option from with_jpeg as the library really needs a provider for JPEG.

```
        if self.options.with_jpeg:
            deps.set_property(self.options.with_jpeg, "cmake_file_name", "JPEG")
            deps.set_property(self.options.with_jpeg, "cmake_target_name", "JPEG::JPEG")
```

(without the differentiation in the CMake:
```
...
libultrahdr/1.3.0: CMakeDeps necessary find_package() and targets for your CMakeLists.txt
    find_package(JPEG)
   
 target_link_libraries(... JPEG::JPEG)
...
-- Configuring done
CMake Error at CMakeLists.txt:562 (target_link_libraries):
  Target "core" links to:

    JPEG::JPEG

    * An ALIAS target is missing.



CMake Error at CMakeLists.txt:703 (target_link_libraries):
  Target "uhdr" links to:

    JPEG::JPEG

    * An ALIAS target is missing.



-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
irieger: This was as with the alias targets for the alternative libs, I either have JPEG::JPEG or create it as an alias to the alternative provider.

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
irieger: I mean we want cmake.install to work also for MSVC (and Xcode) builds probably? Otherwise the install step isn't executed, see comment on the top.

recipes/libultrahdr/all/conanfile.py
irieger: Ok, removed that manually while also removing the _patch_sources method.

recipes/libultrahdr/all/conanfile.py
irieger: Changed.

irieger:
Besides the other changes, I just saw today that also in January there was a release of 1.4.0. I switched to that as it contains some bugfixes and improvemenets: https://github.com/google/libultrahdr/releases/tag/v1.4.0

Should I probably remove 1.3.0 then as no one would be using it currently and so it might just be overhead I guess?

ErniGH:
Thanks for the version update. Yes, it will be better to keep just the last version

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
jcar87: The suggested code in https://github.com/conan-io/conan-center-index/pull/26284#discussion_r2164715365 - would cause whichever jpeg implementation was chosen, to "pose" as `JPEG` and `JPEG::JPEG` - so the is no need for special handling in the cmakelists

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
jcar87: What you are describing is what 
```
  deps.set_property(self.options.jpeg, "cmake_target_name", "JPEG::JPEG")
```
does, but without a patch

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
jcar87: makes sense, thanks!

recipes/libultrahdr/all/conanfile.py
jcar87: thanks!

recipes/libultrahdr/all/conanfile.py
jcar87: thanks!

recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch
jcar87: strange!
I've pushed a commit with the changes I'm suggesting, here: 
https://github.com/conan-io/conan-center-index/commit/4ed99bf6a3354a1baeb8c0cf804242cdc15e3fe3

and it appears to work fine - could you try running the recipe creation with the changes from that commit?

it's basically simplifying the patch, while relying on these cmakedeps `set_property` - note how the proposed changes allow to remove most of the patches

```
       deps.set_property(self.options.with_jpeg, "cmake_file_name", "JPEG")
            deps.set_property(self.options.with_jpeg, "cmake_target_name", "JPEG::JPEG")
            ```
 this will cause CMakeDeps to generate JPEG-config.cmake , with a target called `JPEG::JPEG`, rather than whatever the package_info() says for the specific library
 esentially, `libjpeg-turbo` and `mozjpeg` will pose as JPEG only when building this project (as these calls are private to the build folder of this recipe) - obviously for consumer the info in package_info() is still needed and correct

File: recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch (hunk 2 above)
Comment (irieger) [reply recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch]: Fixed as you suggested below so this should be obsolete.

File: recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch (hunk 3 above)
Comment (irieger) [reply recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch]: Ok, now I got how it works. Should hopefully be as you suggested. Please have another look and see if that was how you meant. Largely reduced the patch! Thanks, this review is a good learning experience.

File: recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch (hunk 4 above)
Comment (irieger) [reply recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch]: Oh, thanks. I just saw this when I was already having done another pass after having slept over it, so it is not exactly as you commit, but I think it reduced the patch even more the way I cleaned it up after implementing your suggestion.

File: recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch (hunk 5 above)
Comment (irieger) [reply recipes/libultrahdr/all/patches/1.3.0_cmake-portability.patch]: Thanks, works in the new state

**Comment by perseoGI** [comment]
Greetings!

Following @ErniGH suggestion, I've tested this new recipe locally with the following Android profile:

```
[settings]
arch=armv8
build_type=Release
compiler=clang
compiler.cppstd=17
compiler.libcxx=c++_static
compiler.version=19
os=Android
os.api_level=28

[tool_requires]
ninja/[*]
android-ndk/r28b

[conf]
```
And it works and runs correctly on an Android Emulator:

```
emu64a:/data/local/tmp $ ./test_package
Creating new decoder 0xb4000075bf879290
```

jcar87 (approved):
thanks everyone!

PR #26309, opened 2025-01-06
ensure sqlite3 resolves to 3.44.2 or newer

jcar87:
* multiple recipes: ensure that references to mysqlite3 can resolve to version 3.44.2 or newer
* version ranges: for the lower bound, use the last well known version to work, rather than arbitrarily raising or lowering it
* mfast: amend test_package to align with current practices: depend on a single package ID, simplify test to minimal include/link, ensure expected additional cmake functionality is seen by cmake, remove test_v1_package
* librasterlite: use version ranges to avoid version conflicts, remove test_v1_package

Other recipes that depend on sqlite3 (not in this PR) should already reference 3.44.2 or a version range.
This will allow us to stop maintaining recipe revisions for versions older than 3.44.2 - making PRs against sqlite3 easier to go through CI.

Note: some recipes in this PR are not currently ported to Conan 2 - but ensuring that this is taken into account in any PR that targets that goal.

Note: not aiming to amend this for all recipes that depend on sqlite 3- that would take too long in CI - will otherwise fix anything based on reported version conflicts.


valgur:
Related:
- #25961

PR #26313, opened 2025-01-06
ccache: bump `zstd` and `hiredis` deps

gegles:
### Summary
Changes to recipe:  **ccache/4.10.2**

#### Details

---

### DIFF
--- a/recipes/ccache/all/conanfile.py
+++ b/recipes/ccache/all/conanfile.py
@@ -46,7 +46,7 @@ class CcacheConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("zstd/1.5.6")
+        self.requires("zstd/[>=1.5 <1.6]")
         if self.options.redis_storage_backend:
             self.requires("hiredis/1.2.0")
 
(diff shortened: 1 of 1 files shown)


recipes/ccache/all/conanfile.py
```diff
@@ -46,9 +46,9 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("zstd/1.5.5")
+        self.requires("zstd/1.5.6")
```
jcar87, line 49:
```suggestion
        self.requires("zstd/[>=1.5 <1.6]")
```

jcar87 (approved):

jcar87:
Hi @gegles  - LGTM! Have amended the PR to use a version range for zstd.

However please note that we no longer consider PRs that bump dependencies without a reason. Reasons are typically "solves a version conflict" (in which case please indicate which) or "requires new version" or "new version has bug or security fixes that impact this library) - in this case indicate which. These helps us prioritise and keep versions conflicts at a minimum.

Approving this PR because we already use a version range for zstd elsewhere, and the version of hiredis in this PR matches the version referenced by other recipes in the current revisions of the git repository

PR #26316, opened 2025-01-06
reflect-cpp: Added new version 0.17.0

liuzicheng1987:
### Summary
Changes to recipe:  **reflect-cpp/0.17.0**

#### Motivation
reflect-cpp 0.17.0 was recently released. This updates the Conan package.

#### Details
- reflect-cpp 0.17.0 added support for Cap'n Proto, so I added this as an option. 
- reflect-cpp 0.17.0 switched to jsoncons as the package for CBOR, so I updated the dependencies accordingly, while maintaining backward compatibility.

---

### DIFF
--- a/recipes/reflect-cpp/all/conandata.yml
+++ b/recipes/reflect-cpp/all/conandata.yml
@@ -2,3 +2,6 @@ sources:
   "0.16.0":
     url: "https://github.com/getml/reflect-cpp/archive/refs/tags/v0.16.0.tar.gz"
     sha256: "a84d94dbd353d788926d6e54507b44c046863f7bc4ecb35afe0338374a68a77d"
+  "0.17.0":
+    url: "https://github.com/getml/reflect-cpp/archive/refs/tags/v0.17.0.tar.gz"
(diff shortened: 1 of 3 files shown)


ErniGH:
Hello @liuzicheng1987 thanks for your PR, I have made some small changes to your PR. On one hand, the new option `with_capnproto` must be removed from the old version since that option didn’t exist before. On the other hand, since we have this case where sometimes we have the option and sometimes we don’t, we use `get_safe`. 

https://github.com/getml/reflect-cpp/compare/v0.16.0...v0.17.0

ErniGH:
We no longer publish releases less than 48 hours after they are created, so we’re going to wait a little bit before merging. Thank you so much for your patience! 😄

liuzicheng1987:
@ErniGH , thanks for the changes. I was thinking about this problem, and didn't really know how to handle it, but now I do. Will be useful for the next update.

As far as the 48 hours are concerned, that's totally fine.

ErniGH (approved):
LGTM Thanks!

perseoGI (approved):

PR #26330, opened 2025-01-08
GStreamer: update all recipes, add support for all available plugins, add gst-orc recipe

valgur:
### Summary
Changes to recipe:  **gstreamer/1.24.10**

#### Motivation
GStreamer is one of the cornerstones of modern multimedia handling, so it deserves a Conan v2 compatible, up-to-date and a more feature-complete recipe.

#### Details
- Adds `gst-orc` recipe, which is used by GStreamer to enable SIMD support in its libraries and plugins.
- Adds additional support for all plugins that have the necessary dependencies available from Conan or as standard system libs.
- Adds `with_introspection` support for all recipes.

The recipes have been successfully tested on Linux, Windows and macOS. Linux and Windows built without issues as both static and shared with all optional deps enabled. macOS had some issues with some minor deps like `json-glib` and `soundtouch`, but otherwise worked correctly as well.

I have taken care to match the `cpp_info.requires` and `cpp_info.system_libs` to the .pc file contents installed by the libraries and to the dependencies applied in the Meson build system of the project.

The `gst-plugins-bad` recipe is larger than the others combined, so it might be a good idea to split it off into a separate PR.

---


valgur:
I also have a fully-featured `gst-plugins-rs` recipe available (https://github.com/valgur/conan-center-index/blob/dev/recipes/gst-plugins-rs/all/conanfile.py), but it will require [Rust support](https://github.com/valgur/conan-center-index/blob/dev/recipes/rust/all/conanfile.py), obviously.

maxsharabayko:
Does this add `gstmpegts-1.0`?
Can some dependencies be made optional? E.g. disable `gst-plugins-bad` if a project does not need them.

PR #26361, opened 2025-01-11
zxing: include path

fnadeau:
### Summary
Changes to recipe:  **zxing/2.3.0**

#### Motivation
Have application build with `gcc $(pkg-config --cflags --libs zxing)` compatible with conan package zxing.

#### Details
This essentially revert 5593391809f7b45eaff9b031e140674dbad9da77. Not having the proper include paths when using conan packages forces the use of patches for application that relies on zxing, ex: [gstreamer pull request](https://github.com/conan-io/conan-center-index/pull/26330/files#diff-afd5ee22e5752bcdd9578fc3894f750772aeec60782bef64f7818660299d511dR66)

[Official pkgConfig for ZXing add ZXing/ to include path](https://github.com/zxing-cpp/zxing-cpp/blob/master/core/zxing.pc.in).

This makes conan package behave the same as official package.

bonus: add gcc >=13 support for old package.

---


planetmarshall:
Might be worth bumping the required C++ standard as [zxing-cpp requires C++20](https://github.com/zxing-cpp/zxing-cpp/blob/7be9048e8176fba4f6ffa3dcf528282c6b0151a9/CMakeLists.txt#L74) from version 2.3.0

fnadeau:
> Might be worth bumping the required C++ standard as [zxing-cpp requires C++20](https://github.com/zxing-cpp/zxing-cpp/blob/7be9048e8176fba4f6ffa3dcf528282c6b0151a9/CMakeLists.txt#L74) from version 2.3.0

C++20 has been marked has minimum for version  >= 2.3.0.

jcar87:
> bonus: add gcc >=13 support for old package.

There are already newer versions of this library - if this is already fixed in those versions and users can update to those versions without much hassle, please remove this change from the PR - we consider backports in cases where an issue is fixed upstream but not yet released, or when there is a very well known used "older" version where it is understood users cannot simply update to. In general and to make our life as reviewers easier, we ask that PRs start with the minimal require to satisfy the objective (assuming that is fixing the include path here)

fnadeau:
@jcar87 backport has been reverted.

uilianries (commented):
@fnadeau Thank you for providing this hotfix.

About moving to C++20 as required, I see C++17 support is still working for the version 2.3.0:

- The changelog points C++20 as default standard, but did not mention required: https://github.com/zxing-cpp/zxing-cpp/releases/tag/v2.3.0
- The upstream's CMakeLists.txt only configures to C++20 in case CMAKE_CXX_STANDARD is not set: https://github.com/zxing-cpp/zxing-cpp/blob/v2.3.0/CMakeLists.txt#L73
- The README file still lists C++17: https://github.com/zxing-cpp/zxing-cpp?tab=readme-ov-file#features
- Locally, I can build zxing 2.3.0 with C+17: [zxing-cpp-2.3.0-linux-static.log](https://github.com/user-attachments/files/19471101/zxing-cpp-2.3.0-linux-static.log)

It's true the upstream lists C++20 compiler in the build instruction, but also comments C++17 is still supported (with limitations): https://github.com/zxing-cpp/zxing-cpp/tree/v2.3.0?tab=readme-ov-file#build-instructions

Based on this, I would not recommend enforcing C++20. It's expected who is consuming the project has information about its capabilities. You could add a `self.output.warning("Using C++17 lacks support for DataBarLimited and multi-symbol and position independent detection for DataMatrix")`

fnadeau:
@jcar87 Excellent point. I'll see what answer I can get about those include path in both gstreamer and zxing project.

uilianries:
@fnadeau Thank you for opening a discussion upstream! 😄 

Linking here your related discussion and PR:

- https://github.com/zxing-cpp/zxing-cpp/discussions/921
- https://github.com/zxing-cpp/zxing-cpp/pull/924

PR #26362, opened 2025-01-12
xkbcommon: allow wayland version range

planetmarshall:
### Summary
Changes to recipe:  **xkbcommon**

#### Motivation

Allow a version range for the wayland dependency to avoid common upstream version conflicts.

#### Details

---


jcar87:
Hi @planetmarshall - the entire repository seems to be alined around wayland 1.22.0 - in which scenarios is this possibly causing conflicts?

planetmarshall:
I had some issues using Qt because the wayland dependency (1.22) did not match the system wayland (1.23) - which was the one actually loaded by system OpenGL libraries (nvidia in this case).

I believe the actual solution in this case was to use shared wayland and link to the system library, so happy to close.

PR #26367, opened 2025-01-13
libsersi: Deprecate opendis6 in favour of libsersi

AbrilRBS:
### Summary
Changes to recipe:  **libsersi/[*]**/**opendis/[*]**

#### Motivation

Upstream renamed, with the new CI system this is now actually not an issue

#### Details

See https://github.com/conan-io/conan-center-index/pull/24487 for details

(Also https://github.com/conan-io/conan-center-index/pull/24601 for the old CI system workaround, this PR combines both)

The idea is that after the current opendis6 revision is published, we can stop publishing new revisions for it by removing from the Git repository (But note that we won't remove the already-published binaries for Opendis6 at any point)

### DIFF
--- a/recipes/libsersi/all/conandata.yml
+++ b/recipes/libsersi/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
   "0.1.0":
-    url: "https://github.com/crhowell3/libseri/archive/refs/tags/0.1.0.tar.gz"
+    url: "https://github.com/crhowell3/libsersi/archive/refs/tags/0.1.0.tar.gz"
     sha256: "b3ce8c73a5b69b963e3c5090e2e237fe142f9aff69a1b6b477b52e2a6a55abb4"
--- a/recipes/opendis6/all/conandata.yml
+++ b/recipes/opendis6/all/conandata.yml
@@ -1,4 +1,4 @@
 sources:
   "0.1.0":
-    url: "https://github.com/crhowell3/libseri/archive/refs/tags/0.1.0.tar.gz"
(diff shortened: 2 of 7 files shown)


recipes/libsersi/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "0.1.0":
+    url: "https://github.com/crhowell3/libsersi/archive/refs/tags/0.1.0.tar.gz"
```
jcar87, line 3:
```suggestion
    url: "https://github.com/crhowell3/libsersi/archive/refs/tags/v0.1.0.tar.gz"
```

recipes/opendis6/all/conandata.yml
```diff
@@ -1,4 +1,4 @@
 sources:
   "0.1.0":
-    url: "https://github.com/crhowell3/opendis6/archive/refs/tags/0.1.0.tar.gz"
-    sha256: "7acfd6ecdcea03c75f93834c4e8ad532aee03eb0fdd2736c9095e2d548214125"
+    url: "https://github.com/crhowell3/libsersi/archive/refs/tags/v0.1.0.tar.gz"
```
jcar87, line 3: This should remain as it was before, since the contents _have_ changed - this new URL uses a different project name, with different library names - even if it did build, the package info would be incorrect. While it is unfortunate that the original sources arent there, the deprecation message should take care of it
Merging this PR should result in exporting libsersi and all its packages, and just the opendis6 recipe with the attribute, but without packages.

jcar87 (approved):

PR #26370, opened 2025-01-13
maintenance: reduce maintained versions for hictk, openexr, uwebsockets

jcar87:
Stop publishing new revisions for versions that are too old or where there are newer versions that are presumed to be non-breaking (this PR should not result in publishing new revisions)
(follow up PRs coming to align versions/dependencies where applicable)


ErniGH (approved):
LGTM nice cleanup

PR #26374, opened 2025-01-13
simdjson: maintain fewer versions

jcar87:
simdjson: reduce versions concurrently maintained - update version used in fastgltf


ErniGH (approved):
LGTM

PR #26393, opened 2025-01-15
Add OptionsPricingModels/1.0.0 recipe with test_package

AldridgeDylan:
OptionsPricingModels/1.0.0

**Description**
Option Pricing Models implemented in C++ using Black-Scholes, Binomial, and Monte Carlo methods.

**Changelog**
- Initial recipe for version 1.0.0
- Includes test_package for validation

### DIFF
--- a/recipes/optionspricingmodels/all/conandata.yml
+++ b/recipes/optionspricingmodels/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "1.0.0":
+    url: "https://github.com/AldridgeDylan/Option-Pricing-Models/archive/refs/tags/V1.0.0.tar.gz"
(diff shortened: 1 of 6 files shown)


jcar87:
Apologies for how long its taken us to get to this PR.

We have decided not to include it in Conan Center at present time. When considering recipes to include, we check things such as the README, whether it is still maintained, presence in other repositories, evidence of use, etc (non-exhaustive list). Today it appears the repository has not received any updates in a year, and there's no evidence of use of any kind - so our concern in accepting this would be that we would be adding a library with no current prospects of being used, which would add to our maintenance tasks. From looking at the readme, it is not immediately evident how the library would be used - however that may just be my lack of knowledge in the area. 

We'll happily provide guidance to host a conanfile.py in the upstream repository, as well as instructions, and will consider this for inclusion in the future if the situation changes- thanks for your contribution!

PR #26400, opened 2025-01-16
b2: simplify, add cross-compilation support, unvendor nlohmann_json

valgur:
### Summary
Changes to recipe:  **b2/[*]**

#### Motivation
The `b2` recipe currently does a lot of things except passing the correct CXX compiler to the build, so while it generally builds thanks to a default C++ compiler or being present on the system, it does not respect the compiler info from the profile and does not pass the default C++ build flags from Conan to the build. As a result, cross-compilation fails as well.

The `toolset` and `use_cxx_env` are something that should really be handled by the build profile. Adjusting the toolset param for the build script adds basically just the standard profile-specific build flags already handled by Conan and can be ignored, I would say: https://github.com/bfgroup/b2/blob/5.2.1/src/engine/build.sh#L326-L422
For Windows builds there are some more specific build flags and the toolset param can't be skipped either: https://github.com/bfgroup/b2/blob/5.2.1/src/engine/config_toolset.bat#L30-L257

With that in mind, I propose deprecating these two options.

#### Details
I also unvendored the `nlohmann_json` dependency, which was added in recent versions. It can use a version range since the dependency won't be visible in the graph.

The package step requires a build-context version of b2 to be present.

C++11 is required, so I added a validate_build check.

I also noticed that the install command complains about not being able to build grammar info due to bison not being present, so I added it as an optional feature to the recipe.

Overall I would say that the core logic of generate/build steps are much simpler as a result. The only remaining complexity stems from trying to manage the CXX and CXXFLAGS values correctly, and the unvendoring of nlohmann_json.

#### Build logs
- aarch64-linux-gnu-gcc-14: [b2_gcc_cross.log](https://github.com/user-attachments/files/18436874/b2_gcc_cross.log)
- MSVC: [b2_msvc.log](https://github.com/user-attachments/files/18436875/b2_msvc.log)
- MinGW:  [b2_mingw.log](https://github.com/user-attachments/files/18436876/b2_mingw.log)

---


valgur:
@uilianries You might find this one relevant since you have been working on Boost lately.

valgur:
The CXX and CXXFLAGS handling could probably be simplified quite a bit by getting them from `GnuToolchain`, but it's not officially supported yet, afaik, and would raise the minimum required Conan version to 2.4, which is not good for a pretty critical recipe.

recipes/b2/portable/conanfile.py
```diff
@@ -20,189 +23,182 @@ class B2Conan(ConanFile):
...
+        if toolset:
+            match = re.match(r"v(\d+)(\d)$", toolset)
+            if match:
+                return f"{match.group(1)}.{match.group(2)}"
+        return ""
 
-    def _is_macos_intel_or_arm(self, settings):
-        return settings.os == "Macos" and settings.arch in ["x86_64", "armv8"]
+    @property
+    def _toolset(self):
```
valgur, line 129: The toolset and toolset version logic is copied directly from the Boost recipe.

PR #26408, opened 2025-01-17
add redis plus plus coroutines

aligirayhanozbay:
### Summary
Changes to recipe: redis-plus-plus

#### Motivation
Enables coroutines support for redis-plus-plus which requires `-DREDIS_PLUS_PLUS_BUILD_CORO=ON`

#### Details
Adds a coroutines option to the conanfile. When turned on, it passes the `REDIS_PLUS_PLUS_BUILD_CORO` cache variable with the value `ON`.

---

### DIFF
--- a/recipes/redis-plus-plus/all/conanfile.py
+++ b/recipes/redis-plus-plus/all/conanfile.py
@@ -23,12 +23,14 @@ class RedisPlusPlusConan(ConanFile):
         "fPIC": [True, False],
         "with_tls": [True, False],
         "build_async": [True, False],
+        "coroutines": [True, False]
     }
     default_options = {
         "shared": False,
         "fPIC": True,
         "with_tls": False,
         "build_async": False,
+        "coroutines": False
     }
 
     @property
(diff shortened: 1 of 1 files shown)


uilianries:
@aligirayhanozbay Hello :wave: Thank you for your suggestion of adding a new option for coroutines in redis-plus-plus :smile: 

Checking the upstream, that project has even more options, like REDIS_PLUS_PLUS_ASYNC_FUTURE, which could use boost as a dependency. 

Still, in ConanCenterIndex we try to be conservative when adding a new option, as a new option affects the package ID, resulting in a new update for any user and package consuming it. Sometimes, we need to rebuild packages internally to update the dependency graph in Conan Center. 

The CI will not test this new option as well, so we will need to check it locally, building a matrix, and pushing the build log. As your new option has a relation to the async option, it requires an extra combination of builds to check. The result is an exponential number of builds for each new option. 

On the other hand, you can use Conan config to extend CMakeToolchain definition, build the package without changing the recipe, much cheaper and simpler than changing the recipe. For your specific case, you can follow:

`conan install -r conancenter --requires=redis-plus-plus/1.3.13 -o "&:build_async=True" -c tools.cmake.cmaketoolchain:extra_variables='{"REDIS_PLUS_PLUS_BUILD_CORO": "ON"}' --build=missing`

To keep this configuration persistent, you can pass this configuration to your Conan profile as well:

```
# ~/.conan2/profiles/redis-plus-plus
...
[conf]
redis-plus-plus/*: tools.cmake.cmaketoolchain:extra_variables={"REDIS_PLUS_PLUS_BUILD_CORO": "ON"}
```

This profile will activate this configuration only for the `redis-plus-plus` package. 

The configuration will work for your build, but will not affect your package ID. In case you want to have a distinct package ID for this configuration, you can use Conan configuration `tools.info.package_id:confs`, which can be added to your profile as well:

`tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]`

I hope this solution fits in your current scenario. Regards!

PR #26411, opened 2025-01-17
lielab: add version 0.4.2

msparapa:
### Summary
Changes to recipe:  lielab/0.4.2

#### Motivation
Adding lielab version 0.4.2

#### Details
Various new features and bug fixes in lielab v0.4.2.

---

### DIFF
--- a/recipes/lielab/all/conandata.yml
+++ b/recipes/lielab/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "0.4.2":
+    url: "https://github.com/sandialabs/Lielab/archive/refs/tags/v0.4.2.tar.gz"
+    sha256: "F423E53A5F56FAFAE52144848A2FA0D1BC9BC30FA67811EF3CDF5C9A31D50A30"
   "0.4.1":
     url: "https://github.com/sandialabs/Lielab/archive/refs/tags/v0.4.1.tar.gz"
     sha256: "862BEC2114B9E554F77ECB2F7B3D5CEABF7374A9A0E7F80BCB94A868F3D9F51C"


khalid80 (approved):

davidsanfal (dismissed):

msparapa:
I see the title was changed from version 0.4.2 to 0.5.1. My intent was to submit the recipe for 0.5.X at a later date via a different PR because I changed the library from header only to static lib as of 0.5.0. Meaning the recipe would need slight restructuring. But I can also submit it however you would prefer.

davidsanfal:
Hi @msparapa ! First of all, sorry for taking so long to review your pull request.

I changed the title because I saw that you had already published version 0.5.1 and I was rewriting the recipe to that new version.
If you review the PR , you'll see that we've changed to static-library and added the information in package_info.
If you see anything that isn't correct, don't doubt to change it.

br and thanks for your contribution!

msparapa:
Hey not a problem. I'm in no rush so the delay is no big deal. Thanks for updating the recipe too, this helps.

I do have a question about the older recipes though. What happens to v0.4.0 and v0.4.1 already on conan center? Nothing important depends on them so if they vanish, that's OK. But that doesn't really seem to be in the spirit of package managers in general. My plan was to have folders /0.4.X and /0.5.X to separate out the header only from static library, but I was only guessing based on how other packages were treated when there are major changes. I don't really know how the back end of conan center works.

franramirez688:
> I do have a question about the older recipes though. What happens to v0.4.0 and v0.4.1 already on conan center? Nothing important depends on them so if they vanish, that's OK. But that doesn't really seem to be in the spirit of package managers in general. My plan was to have folders /0.4.X and /0.5.X to separate out the header only from static library, but I was only guessing based on how other packages were treated when there are major changes. I don't really know how the back end of conan center works.

Those versions will still be available. Removing versions from the conandata.yml means that we are not going to publish more revisions for those versions, but the already existing ones will be there 😁
Doing that helps us keep the recipes as simple as possible without splitting them into several folders, keeping old patches, tons of if-else checks for versions, etc. Thanks a lot for asking.

davidsanfal (approved):

recipes/lielab/all/conanfile.py
```diff
@@ -15,69 +14,59 @@ class LielabConan(ConanFile):
...
-    @property
-    def _compilers_minimum_version(self):
-        return {
-            "gcc": "11",
-            "clang": "12",
-            "apple-clang": "13.1",
-            "Visual Studio": "17",
-            "msvc": "193",
-        }
+        self.requires("eigen/3.4.0", transitive_headers=True)
```
franramirez688:
```suggestion
        self.requires("eigen/[>=5.0.0 <6]", transitive_headers=True)
```

davidsanfal (approved):

franramirez688:
Thanks a lot @msparapa for contributing, and again, apologies for the long delay in reviewing this 🙏 
Any new future PRs to improve or add new versions are more than welcome.

msparapa:
Thank you both for your work on the recipe and explanations!

Issue #26265, opened 2024-12-25
[package] avahi/0.8: shadows standard headers


nine:
### Description

With integration of pull request #20741 all avahi headers were exposed without `avahi-*` prefix. This causes a collision with standard headers. E.g. `<avahi-common/malloc.h>` collides with `<malloc.h>` shipped with glibc on GNU/Linux. 

Affected headers:
1. `<avahi-common/error.h>`
2. `<avahi-common/malloc.h>`

f.y.i @jwillikers, @AbrilRBS

### Package and Environment Details

* Package Name/Version: **avahi/0.8**
* Operating System+version: **Linux Debian 12.8**
* Compiler+version: **GCC 12.2.0**
* Conan version: **conan 2.11.0**
* Python version: **Python 3.11.2**

### Conan profile

Host profile:
```
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
```

Build profile:
```
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu17
compiler.libcxx=libstdc++11
compiler.version=12
os=Linux
```

### Steps to reproduce

```
git clone https://github.com/nine/conan-center-index.git
git checkout 26265_fix_shadowed_headers
cd conan-center-index
conan create recipes/avahi/all --version=0.8
```

### Logs

<details><summary>Click to expand log</summary>

```
avahi/0.8 (test package): Running CMake.build()
avahi/0.8 (test package): RUN: cmake --build "/home/nine/src/github.com/nine/conan-center-index/recipes/avahi/all/test_package/build/gcc-12-x86_64-gnu17-release" -- -j8
[ 25%] Building C object CMakeFiles/test_package.dir/test_package.c.o
[ 50%] Building C object CMakeFiles/client-browse-services.dir/client-browse-services.c.o
/home/nine/src/github.com/nine/conan-center-index/recipes/avahi/all/test_package/client-browse-services.c: In function ‘main’:
/home/nine/src/github.com/nine/conan-center-index/recipes/avahi/all/test_package/client-browse-services.c:183:5: error: implicit declaration of function ‘malloc_trim’ [-Werror=implicit-function-declaration]
  183 |     malloc_trim(0);
      |     ^~~~~~~~~~~
[ 75%] Linking C executable test_package
cc1: some warnings being treated as errors
[...]

PR #20741, opened 2023-10-23
[avahi] Expose both root include folder and module name

uilianries:
Specify library name and version:  **avahi/0.8**

The avahi package also provides `mdnsresponder`,  and when doing it, the consumer does not expect avahi, but the true include header, like `#include <dns.h>`. 

This reverts part of the commit https://github.com/conan-io/conan-center-index/pull/20706/commits/88c85c9b2c5cc25e2a1574fcc937695bdc7cadc3, but keep both root and module name as include folder.

Required by https://github.com/conan-io/conan-center-index/pull/20287

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #26413, opened 2025-01-18
avahi: fix shadowed headers (#26265)

nine:
### Summary
Changes to recipe:  **avahi/0.8**

#### Motivation
This PR fixes the shadowing of standard headers as described in #26265.

#### Details
This PR reverts the change introduced in #20741 for headers in `<avahi-common/*.h>`.

---

### DIFF
--- a/recipes/avahi/all/conanfile.py
+++ b/recipes/avahi/all/conanfile.py
@@ -113,7 +113,8 @@ class AvahiConan(ConanFile):
             self.cpp_info.components[lib].names["cmake_find_package"] = lib
             self.cpp_info.components[lib].names["cmake_find_package_multi"] = lib
             self.cpp_info.components[lib].libs = [avahi_lib]
-            self.cpp_info.components[lib].includedirs = ["include", os.path.join("include", avahi_lib)]
+            if lib != "common":
+                self.cpp_info.components[lib].includedirs = ["include", os.path.join("include", avahi_lib)]
         self.cpp_info.components["compat-libdns_sd"].libs = ["dns_sd"]
 
         self.cpp_info.components["client"].requires = ["common", "dbus::dbus"]
--- a/recipes/avahi/all/test_package/test_package.c
+++ b/recipes/avahi/all/test_package/test_package.c
@@ -1,8 +1,10 @@
+#include <malloc.h>
 #include <stdlib.h>
 
 #include "dns_sd.h"
-#include "avahi-core/log.h"
-
+#include <avahi-core/log.h>
+#include <avahi-common/malloc.h>
+#include <avahi-common/error.h>
 
 int main(void) {
     DNSServiceRef sdRef;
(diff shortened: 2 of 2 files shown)


recipes/avahi/all/conanfile.py
```diff
@@ -113,7 +113,8 @@ def package_info(self):
             self.cpp_info.components[lib].names["cmake_find_package"] = lib
             self.cpp_info.components[lib].names["cmake_find_package_multi"] = lib
             self.cpp_info.components[lib].libs = [avahi_lib]
-            self.cpp_info.components[lib].includedirs = ["include", os.path.join("include", avahi_lib)]
+            if lib != "common":
```
uilianries, line 116: It could be applied to other components too, but this recipe was created 4 years ago and may break users, so let's keep only your change for now.

recipes/avahi/all/test_package/CMakeLists.txt
```diff
@@ -5,3 +5,7 @@ find_package(Avahi CONFIG REQUIRED)
 
 add_executable(${PROJECT_NAME} test_package.c)
 target_link_libraries(${PROJECT_NAME} PRIVATE avahi::compat-libdns_sd avahi::core)
+
```
uilianries, line 8: Please, revert your test package changes. The idea of test package is validating the package itself, not running full examples, we just want to check for artifacts and linkage status, so we keep as simple as possible always.

I see you are trying to validate this explored bug, so I would suggest only adding:

```
#include <avahi-common/malloc.h>
#include <avahi-common/error.h>
```

to `all/test_package/test_package.cpp`

uilianries (commented):
@nine Thank you for your PR! Indeed you found a recipe bug :bug: 

The upstream does not include subfolders, but uses avahi-xxx as namespace: https://github.com/avahi/avahi/blob/master/examples/client-browse-services.c#L33

Please, take a look at my review, so we can progress with this PR. Cheers!

recipes/avahi/all/test_package/CMakeLists.txt
nine: OK, I've updated the package accordingly that both `<malloc.h>` and `<avahi-common/malloc.h>` are included.

nine:
@uilianries please re-review.

uilianries (approved):
LGTM. This PR fixes a recipe bug that has been integrated since its creation date (4 years ago). The Avahi project does not include its folders to include dirs, instead use them as a namespace, their examples shows it: https://github.com/avahi/avahi/blob/master/examples/client-browse-services.c?rgh-link-date=2025-01-20T15%3A58%3A37Z#L33

The current recipe conflicts with both malloc.h from stdlib and Avahi. This PR removes the inclusion of that folder.

The same could be applied to all subfolders, but I'm afraid it may break users, as this property has been present since the recipe's first day in Conan Center. 

Thank you for proving this fix!

nine:
@uilianries can I help to get a second review?

nine:
Maintainers can you please help with the second review? 

@franramirez688, @ErniGH, @AbrilRBS

uilianries (approved):

uilianries (approved):

franramirez688 (approved):

PR #26423, opened 2025-01-20
Add frugally-deep version 0.16.1

Dobiasd:


jcar87:
Hi @Dobiasd - thanks for your contribution! When adding new versions please limit the PR to adding the most recent version, unless there is a strong reason why a user may choose to be on a specific older version - we're happy to make this change in the PR as well

Dobiasd:
Ah, ok, got it, thanks! I just removed version 0.16.1 again. :heavy_check_mark:

czoido (approved):

ErniGH (approved):
LGTM Thanks!

PR #26447, opened 2025-01-22
ffmpeg: obey pkg_config configuration

jcar87:
### Summary
Changes to recipe:  ffmpeg/all

#### Motivation
If the user specifies which pkg-config to use, propagate that to the ffmpeg scripts.

#### Details

Unlike other tools (autotools, cmake), the `./configure` script in ffmpeg is *not* strictly derived from autotools, so it makes choices of its own.
`pkg-config` is taken from https://github.com/FFmpeg/FFmpeg/blob/master/configure#L404, which resolves to either
- `pkg-config` (https://github.com/FFmpeg/FFmpeg/blob/master/configure#L4125)
- `[cross-prefix]-pkg-config` (https://github.com/FFmpeg/FFmpeg/blob/6ecc96f4d08d74b0590ab03f39f93f386910c4c0/configure#L4700C1-L4700C58)- but only when a `--cross-prefix` is provided

In either case, if the user specifies the Conan configuration `tools.gnu:pkg_config` or `PKG_CONFIG` - it will be completely ignored. This is particularly problematic when crossbuilding and we need a pkg-config tailored to handle system dependencies, e.g
- we are expecting to use a prefixed executable, e.g. `aarch64-linux-gnu-pkgconf`, but we either did not pass `cross-prefix`, *OR* we are using `pkgconfg` (newer debians use pkgconf)
- we are using a wrapper as documented in https://autotools.info/pkgconfig/cross-compiling.html?mtm_campaign=homepage&amp;mtm_content=redirects  

In both cases, when cross building and dealing with system dependencies, it is important that *both* of the following are true:

- pkg-config only locates libraries for the host platform (the "target", what we are building _for_), see for [example](https://github.com/jcar87/conan-crossbuild-with-sysroot/blob/main/support_files/pkg-config#L7-L9)
- pkg-config **must** prune "system" paths (locations where the compiler and linker already look), so that they don't end up having higher precedence than Conan ones. When doing a native build, `pkg-config` already does this by default, but if we want to reuse the native executable, we need to tailor this to the sysroot in question, see [example](https://github.com/jcar87/conan-crossbuild-with-sysroot/blob/main/support_files/pkg-config#L14-L16) 

The ffmpeg configure behaviour might explain why there was already logic in the recipe, chances are this was failing on systems where `pkg-config` as an executable is not found on the path

---

### DIFF
--- a/recipes/ffmpeg/all/conanfile.py
+++ b/recipes/ffmpeg/all/conanfile.py
@@ -373,7 +373,8 @@ class FFMpegConan(ConanFile):
                 self.tool_requires("nasm/2.16.01")
             else:
                 self.tool_requires("yasm/1.3.0")
-        if not self.conf.get("tools.gnu:pkg_config", check_type=str):
+        if self.settings.os != "Linux" and not self.conf.get("tools.gnu:pkg_config", check_type=str):
+            # See https://github.com/conan-io/conan-center-index/pull/26447#discussion_r1926682155
             self.tool_requires("pkgconf/[>=2.1 <3]")
         if self._settings_build.os == "Windows":
             self.win_bash = True


recipes/ffmpeg/all/conanfile.py
```diff
@@ -642,11 +642,12 @@ def opt_append_disable_if_set(args, what, v):
         ranlib = buildenv_vars.get("RANLIB")
         if ranlib:
             args.append(f"--ranlib={unix_path(self, ranlib)}")
-        # for some reason pkgconf from conan can't find .pc files on Linux in the context of ffmpeg configure...
-        if self._settings_build.os != "Linux":
-            pkg_config = self.conf.get("tools.gnu:pkg_config", default=buildenv_vars.get("PKG_CONFIG"), check_type=str)
-            if pkg_config:
-                args.append(f"--pkg-config={unix_path(self, pkg_config)}")
+        pkg_config = self.conf.get("tools.gnu:pkg_config", default=buildenv_vars.get("PKG_CONFIG"), check_type=str)
```
jcar87, line 645: @memsharded  - I'm not entirely sure about  buildenv_vars.get("PKG_CONFIG") - I suppose that covers when a user has the variable defined in the `[buildenv]` section, but not if a user has it in the actual environment (without having it in the profile)

recipes/ffmpeg/all/conanfile.py
```diff
@@ -642,11 +642,12 @@ def opt_append_disable_if_set(args, what, v):
         ranlib = buildenv_vars.get("RANLIB")
         if ranlib:
             args.append(f"--ranlib={unix_path(self, ranlib)}")
-        # for some reason pkgconf from conan can't find .pc files on Linux in the context of ffmpeg configure...
```
jcar87, line 645: "some reason"
the `pkgconf` from the Conan package is not able to locate `.pc` files for system dependencies, e.g. in `/usr/lib/aarch64-linux-gnu/libfoo.pc`, because it is hardcoded to look at `/lib/pkgconfig` and `/share/pkgconfig` - on Ubuntu there's no system .pc files in those directories. 

it works when `pkgconf` is used to locate conan-generated .pc files, and on platforms where PKG_CONFIG_PATH is defined in the shell environment (like msys2) - but won't work on Linux. 

We should always recommend that on Linux (and any other similar OS that already has a pkg-config), that the system pkg-config is used, and let build systems fall back to locating a prefixed pkg-config when crossbuilding - and this should be the default behaviour without requiring additional configuration

memsharded (approved):

recipes/ffmpeg/all/conanfile.py
valgur: Also, besides xorg, none of the packages really need to be `/system`, really:
- `vaapi/system`: https://github.com/valgur/conan-center-index/blob/dev/recipes/libva/all/conanfile.py
- `vdpau/system`: https://github.com/valgur/conan-center-index/blob/dev/recipes/vdpau/all/conanfile.py

PR #26457, opened 2025-01-23
Open62541 qnx support

klausholstjacobsen:
### Summary
Changes to recipe:  **open62541/1.4.8**

#### Motivation
Added qnx support and added 1.4.8 to conandata

#### Details
Upstream PR for master: https://github.com/open62541/open62541/pull/6967
Upstream PR for 1.4 branch: https://github.com/open62541/open62541/pull/6966

---


recipes/open62541/all/conanfile.py
```diff
@@ -363,8 +366,9 @@ def generate(self):
 
     def _patch_sources(self):
         apply_conandata_patches(self)
-        if Version(self.version) >= "1.3.1":
-            os.unlink(os.path.join(self.source_folder, "tools", "cmake", "FindPython3.cmake"))
+        py_path = os.path.join(self.source_folder, "tools", "cmake", "FindPython3.cmake")
+        if os.path.exists(py_path):
+            os.unlink(py_path)
```
valgur, line 371:
```suggestion
        rm(self, "FindPython3.cmake", os.path.join(self.source_folder, "tools", "cmake"))
```

recipes/open62541/all/conandata.yml
```diff
@@ -18,6 +21,19 @@ sources:
     url: "https://github.com/open62541/open62541/archive/v1.0.6.tar.gz"
     sha256: "299940025c14929533064abe0044d5805ea50d52b32d05ad9bc0e6996569c2a6"
 patches:
+  "1.4.8":
```
valgur, line 24: - Are the previous patches in v1.4.6 no longer needed?
- I would mark the patch type as "portability".
- Add `patch_source: ...` with the PR URL if it has been submitted upstream.

jcar87 (changes requested):
hi @klausholstjacobsen - thanks for your contribution. Looks like some patches from earlier versions are missing and they are causing CI failures - could you take a look? thanks so much! :)

recipes/open62541/all/conandata.yml
klausholstjacobsen: Adding the necessary patches and changed patch type!

jcar87 (changes requested):
* See note about `UA_PARSING`
* Please only add the most recent version (`1.4.11.1`), unless there are strong reasons why a user on one of the earlier versions would be unable to update directly to the latest.

recipes/open62541/all/conanfile.py
```diff
@@ -352,6 +357,7 @@ def generate(self):
             tc.variables["UA_MSVC_FORCE_STATIC_CRT"] = True
 
         tc.variables["UA_COMPILE_AS_CXX"] = self.options.cpp_compatible
+        tc.variables["UA_PARSING"] = self.options.parsing
```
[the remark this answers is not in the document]
klausholstjacobsen: That's an incredible catch right there! Thank you!
It's called UA_ENABLE_PARSING not UA_PARSING.

Look here: https://www.open62541.org/doc/master/building.html

Fixed!

klausholstjacobsen:
Removed versions older than 1.4.x and 1.3.x to comply with good practice of only keeping two latest minor versions

uilianries:
@klausholstjacobsen Hello! Thank you for reducing the version list to be built. I see the version 1.4.8 is still listed, including the latest 1.4.11.1. Do you really need 1.4.8? Otherwise, I would vote for keeping 1.4.11.1.

recipes/open62541/all/conanfile.py
```diff
@@ -109,6 +109,7 @@ class Open62541Conan(ConanFile):
         "cpp_compatible": [True, False],
         # UA_ENABLE_STATUSCODE_DESCRIPTIONS=readable_statuscodes
         "readable_statuscodes": [True, False],
+        "parsing": [True, False],
```
uilianries, line 112: Hello, why exactly `parsing` option is needed now? Is it really critical for this new version? It's `False` by default, so it seems be only a possibility of customization, not a must feature. 

When adding an option, it requires an extra effort building and validating it in different scenarios (e.g. Windows, Linux Mac), only to make sure we are not breaking users when enabling and the option will really work with the build. This extra effort not only cost resources in terms of review and CI, but may not worth most of cases, as not all users are interested or do not know about those new options, as there is no open issue asking or a chorus for that.

Still, Conan 2.x provides a feature that make possible to inject any CMake definition when building a package, without changing the recipe. Much cheaper and convenient when needing to customize a package without touch any Conan recipe in ConanCenterIndex:

https://docs.conan.io/2/reference/config_files/global_conf.html#introduction-to-configuration

> tools.cmake.cmaketoolchain:extra_variables: Dictionary with variables to be injected in CMakeToolchain (potential override of CMakeToolchain defined variables)

As it's a dictionary, we should use it as follow: 

    conan create recipes/open62541/all --version=1.4.11.1 -c 'tools.cmake.cmaketoolchain:extra_variables={"UA_ENABLE_PARSING": "ON"}'

And it can be added to your Conan profile as well, so you do not need to pass via command line always:

```
[conf]
tools.cmake.cmaketoolchain:extra_variables={"UA_ENABLE_PARSING": "ON"}
```

Still, this configuration will **not** affect the package ID, so the new package should override the existent one in your Conan cache.
But you can mark to generate a new package ID, based on this configuration value too, so this new package will differ from others in your Conan cache and will not collide:

> tools.info.package_id:confs: List of existing configuration to be part of the package ID

So you will need to include:

    conan create recipes/open62541/all --version=1.4.11.1 -c 'tools.cmake.cmaketoolchain:extra_variables={"UA_ENABLE_PARSING": "ON"}' -c 'tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]'

And you can add to your Conan profile as well. 

This approach is much more flexible and is valid to any recipe that's is built with Conan 2.x. Please, consider using it first.

klausholstjacobsen: Personally I think that a recipe should expose all options available in the component it build.
I do not like the notion of configuring some options through the recipe options and some through some other means of injecting options directly into the component build.

Secondly tools.cmake.cmaketoolchain expose stuff that should be completely internal to the recipe, namely that it builds the component with cmake. Comsumers should not have knowledge about how a recipe builds a component.

klausholstjacobsen:
> @klausholstjacobsen Hello! Thank you for reducing the version list to be built. I see the version 1.4.8 is still listed, including the latest 1.4.11.1. Do you really need 1.4.8? Otherwise, I would vote for keeping 1.4.11.1.

To comply with official guidelines we should only keep 1.4.11.1 and 1.3.9, correct?

jcar87:
> > @klausholstjacobsen Hello! Thank you for reducing the version list to be built. I see the version 1.4.8 is still listed, including the latest 1.4.11.1. Do you really need 1.4.8? Otherwise, I would vote for keeping 1.4.11.1.
> 
> To comply with official guidelines we should only keep 1.4.11.1 and 1.3.9, correct?

I think there is some confusion here, perhaps we can clarify our guidelines.
The wording here: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/sources_and_patches.md#removing-old-versions - applies when removing actively maintained versions for the recipe.

When _adding_ versions to a recipe, we ask that only a single new version is added (hopefully the latest), and not any older version, unless there's a very strong reason for it - typical reasons include: the most recent version requires more invasive changes to the recipe, the older version is an LTS, or is widely used and is not easy for consumers to migrate to the most recent. Unless that's the case here, please _only_ add 1.4.11.1

recipes/open62541/all/conanfile.py
```diff
@@ -269,6 +271,7 @@ def _get_multithreading_option(self):
 
     def generate(self):
         tc = CMakeToolchain(self)
+        tc.extra_cflags.extend(["-Wno-unused-variable"])
```
jcar87, line 274: If this is only causing warnings (and not errors) - please remove this line from the recipe. If this is an issue, please report this to the open62541 maintainers.
Some users _do_ want to be able to see the warnings produced by third party open source libraries, whereas others may want to hide them to minimise the logs - the latter should be handled in the profile using the appropriate conf.

klausholstjacobsen: Agree, Done!

klausholstjacobsen:
@jcar87, yes I find the docs for adding/removing version a bit confusing! That being said, then I agree 1.4.11.1 should be the only one to add! Fixed!

PR #26459, opened 2025-01-23
Libjwt

JuanBC9:
### Summary
Added recipe: **libjwt/1.18.3**

#### Motivation
Added new recipe for libjwt library

## New recipe

### DIFF
--- a/recipes/libjwt/all/conanfile.py
+++ b/recipes/libjwt/all/conanfile.py
@@ -26,6 +26,7 @@ class libjwtRecipe(ConanFile):
 
     def requirements(self):
         self.requires("openssl/3.3.2")
+        self.requires("jansson/2.14")
 
     def system_requirements(self):
         apt = package_manager.Apt(self)


AbrilRBS:
Hi @JuanBC9 thanks a lot for taking the time to implement this. Is there a specific reason to stay on v1? I have implemented the necessary changes in https://github.com/zaleos/conan-center-index/pull/1 to move this PR to v3, please review it :)

IvanRibakov:
Hi @AbrilRBS , this PR along with https://github.com/conan-io/conan-center-index/pull/26449 is an effort to ultimately bring [libstirshaken](https://github.com/signalwire/libstirshaken) to Conan. If my memory serves well, libstirshaken was not compatible with `libjwt` above v1 at the moment. Could we include in the recipe latest versions of v1 and v3? Could those co-exist in a single recipe?

recipes/libjwt/all/conanfile.py
```diff
@@ -0,0 +1,82 @@
...
+        dnf.install(["jansson-devel"], update=True, check=True)
+
+        zypper = package_manager.Zypper(self)
+        zypper.install(["libjansson-devel"], update=True, check=True)
+
+        pacman = package_manager.PacMan(self)
+        pacman.install(["jansson"], update=True, check=True)
+
+        pkg = package_manager.Pkg(self)
+        pkg.install(["jansson"], update=True, check=True)
```
IvanRibakov, line 48: @AbrilRBS when creating this PR we had a doubt about how best to handle this dependency. We saw that it is already in Conan but we couldn't find a better way to satisfy current `libjwt` build process than by installing a system requirement. I saw that you managed to avoid a system dependency in your PR but it's not clear to me if the same can be attempted with v1.

AbrilRBS:
Hi @IvanRibakov sorry for the delay answering this

> Could we include in the recipe latest versions of v1 and v3? Could those co-exist in a single recipe?

Yes, recipes can support multiple versions by branching where behaviours in their build systems differ, and having the different versions be declared in the config.yml and conandata.yml files

recipes/libjwt/all/conanfile.py
AbrilRBS: Yes, the same approach should work for both versions, as Conan is already managing the necessary dependencies in the respective recipes - if any of the required configurations fail, we can take a look at that specific case, but usually just using normal Conan recipes for dependencies is much more desired

jcar87:
Thanks @JuanBC9 
We are continuing in this PR: #27596  - as there were some fixes needed to handle jenssen and other small things.

jcar87:
https://github.com/conan-io/conan-center-index/pull/27596 has now been merged, apologies for the delay. Thank you for your contribution!

Issue #26379, opened 2025-01-14
[question]  Could we add the arm64 building support for msys2? 


codereba:
### What is your question?

MSYS2 already provides some support for arm64 now, contains clang arm64 and cross build tool from x64 to arm64, please refer to:
    https://www.msys2.org/wiki/arm64/
    https://packages.msys2.org/groups/mingw-w64-cross-toolchain

Currently msys2 conanfile.py just raise an exception:
    raise ConanInvalidConfiguration("Only Windows x64 and arm64 supported")

Could we add the arm64 building support for msys2? let user to choose how to use msys2 for building.

jcar87:
`msys2` in Conan Center is used to provide bash and autotools for recipes that only support being built in that environment, and *NOT* the compiler (gcc, clang) - that is still msvc
As far as I know, these tools are still x86-64, as per the link you have provided:

<img width="906" alt="Screenshot 2025-01-14 at 09 56 49" src="https://github.com/user-attachments/assets/6023a5cd-562a-40ce-9d5e-c47ffcfbc489" />

This is easy to reflect with Conan,  you can do do so:

- via command line `--setings:build "msys2/*:arch=x86_64`
- modify your build profile to contain, in the `[settings]` section:
```
msys2/*:arch=x86_64
```
 (note that you will need to pass two different profiles when invoking `conan install` or `conan create`)

This should suffice to build libraries to target armv8 on Windows for libraries that require msys2, for example:

```
conan install --require=libx264/cci.20240224 --build=missing --settings:build="msys2/*:arch=x86_64"
```

This should successfully build the arm64 version of libx264 on Windows:

```
======== Input profiles ========
Profile host:
[settings]
arch=armv8
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows
[conf]


Profile build:
[settings]
arch=armv8
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows
msys2/*:arch=x86_64
[conf]
```

Let us know if this works for you!

codereba:
Hi @jcar87 , 

Thanks for your response, 

I have a question, if the msys2 files are x64 arch then they can run on windows arm by the arm64 emulation, why msys2 recipe deny the armv8 as the target arch?
[...]

PR #26466, opened 2025-01-26
msys2:change the error message to let user know how to work on windows arm64.

codereba:
fixes https://github.com/conan-io/conan-center-index/issues/26379

### Summary
Changes to recipe:  **msys2/cci.latest**

#### Motivation
msys2 now doesn't fully support native arm64, but its default installed x86_64 executable files can work on x86 emulation on windows 11 arm, please refer to:
https://learn.microsoft.com/en-us/windows/arm/apps-on-arm-x86-emulation

And can use the mingw-w64-cross-gcc package fro arm64 cross build from x64, please refer to:
https://github.com/Windows-on-ARM-Experiments/msys2-woarm64-build

so, we can set the arch of build profile to x64 and the arch of host profile to arm64 for msys2:
```
Profile host:
[settings]
arch=armv8
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows
[conf]

Profile build:
[settings]
arch=armv8
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows
msys2/*:arch=x86_64
[conf]
```
or add the --settings:build command line argument: 
`conan install --require=libxxx/xxx --build=missing --settings:build="msys2/*:arch=x86_64"`

That will try to run msys2 in x86 emulation mode, work together with other arm64 native tool; and if any recipe needs arm cross build, it can install mingw-w64-cross-gcc package and utilize it manually.

#### Details
The error message can be refined to tip user how to correctly run x64 files of msys2, if user doesn't have this background knowledge, they will consume many time to research, and may not find the simple way to make the build successfully.
I think msys2 will support native arm64 in future, and this rare issue only happened between windows with x86 emulation and no arm64 tool, it may be not important enough to write in the document.

---

### DIFF
--- a/recipes/msys2/all/conanfile.py
+++ b/recipes/msys2/all/conanfile.py
@@ -67,6 +67,8 @@ class MSYS2Conan(ConanFile):
     def validate(self):
         if self.settings.os != "Windows":
             raise ConanInvalidConfiguration("Only Windows supported")
+        if self.settings.arch == "armv8":
+            raise ConanInvalidConfiguration("msys2 does not support arm64 natively now, you may use the setting of build: \"msys2/*:arch=x86_64\" to let msys2 run on x86 emulation.")
         if self.settings.arch != "x86_64":
             raise ConanInvalidConfiguration("Only Windows x64 supported")
 


jcar87:
Thanks @codereba  - thanks in favour of https://github.com/conan-io/conan-center-index/pull/27882, which will make it fallback on existing binaries without having to do anything extra on the users end.

PR #26477, opened 2025-01-28
llama-cpp: Add new b4570 version

AbrilRBS:
### Summary
Changes to recipe:  **llama-cpp/b4570**

#### Motivation

We wanted to use it in a future blogpost

#### Details

Lots of changes for this new version, so the package_info is split in two: the old one, and the new one. Maybe it would be better to remove the old versions from the conandata and just keep this new version, but thought that the current approach is a bit better until we can decide

The ARM msvc validation check is due to this:

`ggml/src/ggml-cpu/CMakeLists.txt` has this when adding ggml-cpu:

```
if (MSVC AND NOT CMAKE_C_COMPILER_ID STREQUAL "Clang")
            message(FATAL_ERROR "MSVC is not supported for ARM, use clang")
        else()
```

As the cpu backend is always enabled in the recipe, I just validate this case out

### DIFF
--- a/recipes/llama-cpp/all/conandata.yml
+++ b/recipes/llama-cpp/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "b4570":
+    url: "https://github.com/ggerganov/llama.cpp/archive/refs/tags/b4570.tar.gz"
+    sha256: "35bfe07807fd0cf30710023765b9a7ab6c1003f27ef907ce9cea2c5464411430"
   "b4079":
     url: "https://github.com/ggerganov/llama.cpp/archive/refs/tags/b4079.tar.gz"
     sha256: "79093413dcdbd30f83b800aeb958c87369fdfdaf4e5603b094185898ff404a32"
--- a/recipes/llama-cpp/all/conandata.yml
+++ b/recipes/llama-cpp/all/conandata.yml
@@ -8,9 +8,6 @@ sources:
   "b3542":
     url: "https://github.com/ggerganov/llama.cpp/archive/refs/tags/b3542.tar.gz"
     sha256: "6f8b23d930400fce5708d2c85022ef33f1083af8f6ac395abefadacee0942e78"
-  "b3040":
-    url: "https://github.com/ggerganov/llama.cpp/archive/refs/tags/b3040.tar.gz"
-    sha256: "020e040139660eb40113503bb1057d5387677d249b990e008e04821532f7cd62"
 patches:
   "b4570":
     - patch_file: "patches/b4570-001-curl-patch-targets.patch"
@@ -18,5 +15,3 @@ patches:
     - patch_file: "patches/b4079-001-curl-patch-targets.patch"
   "b3542":
     - patch_file: "patches/b3542-001-curl-patch-targets.patch"
-  "b3040":
-    - patch_file: "patches/b3040-001-curl-patch-targets.patch"
--- a/recipes/llama-cpp/all/conandata.yml
+++ b/recipes/llama-cpp/all/conandata.yml
@@ -12,6 +12,8 @@ sources:
     url: "https://github.com/ggerganov/llama.cpp/archive/refs/tags/b3040.tar.gz"
     sha256: "020e040139660eb40113503bb1057d5387677d249b990e008e04821532f7cd62"
 patches:
+  "b4570":
+    - patch_file: "patches/b4570-001-curl-patch-targets.patch"
   "b4079":
     - patch_file: "patches/b4079-001-curl-patch-targets.patch"
   "b3542":
--- a/recipes/llama-cpp/all/conanfile.py
+++ b/recipes/llama-cpp/all/conanfile.py
@@ -51,7 +51,7 @@ class LlamaCppConan(ConanFile):
         check_min_cppstd(self, 17 if self._is_new_llama else 11)
 
     def validate_build(self):
-        if Version(self.version) >= "b4570" and self.settings.compiler == "msvc" and "arm" in self.settings.arch:
+        if self._is_new_llama and self.settings.compiler == "msvc" and "arm" in self.settings.arch:
             raise ConanInvalidConfiguration("llama-cpp does not support ARM architecture on msvc, it recommends to use clang instead")
 
     def layout(self):
--- a/recipes/llama-cpp/all/conanfile.py
+++ b/recipes/llama-cpp/all/conanfile.py
@@ -68,6 +68,7 @@ class LlamaCppConan(ConanFile):
         tc.variables["LLAMA_CURL"] = self.options.get_safe("with_curl")
         if cross_building(self):
             tc.variables["LLAMA_NATIVE"] = False
+            tc.variables["GGML_NATIVE_DEFAULT"] = False
 
         tc.variables["GGML_BUILD_TESTS"] = False
         #tc.variables["GGML_BUILD_EXAMPLES"] = self.options.get_safe("with_examples")
@@ -108,6 +109,7 @@ class LlamaCppConan(ConanFile):
         self.cpp_info.components["common"].requires = ["llama"]
         if self.options.with_curl:
             self.cpp_info.components["common"].requires.append("libcurl::libcurl")
+            self.cpp_info.components["common"].defines.append("LLAMA_USE_CURL")
         if is_apple_os(self):
             self.cpp_info.components["common"].frameworks.extend(["Foundation", "Accelerate", "Metal"])
         elif self.settings.os in ("Linux", "FreeBSD"):
--- a/recipes/llama-cpp/all/conanfile.py
+++ b/recipes/llama-cpp/all/conanfile.py
@@ -192,6 +192,3 @@ class LlamaCppConan(ConanFile):
                     self.cpp_info.components["ggml-blas"].frameworks.append("Accelerate")
                 if "metal" in backends:
                     self.cpp_info.components["ggml-metal"].frameworks.extend(["Metal", "MetalKit", "Foundation", "CoreFoundation"])
-            if "cuda" in backends:
-                # TODO: Add CUDA information
-                pass
--- a/recipes/llama-cpp/all/conanfile.py
+++ b/recipes/llama-cpp/all/conanfile.py
@@ -9,7 +9,7 @@ from conan.tools.files import copy, get, rmdir, apply_conandata_patches, export_
 from conan.tools.scm import Version
 
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=2.0.9"
 
 
 class LlamaCppConan(ConanFile):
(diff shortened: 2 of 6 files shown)


recipes/llama-cpp/all/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "b4570":
```
czoido, line 2: We are missing this version in the `config.yml`, It appear in the linter as warning, but it advises not to handle that in this PR. I think that's an error in this case and we should change the linter advise.

recipes/llama-cpp/all/conanfile.py
```diff
@@ -83,14 +64,18 @@ def generate(self):
...
         tc.variables["LLAMA_CURL"] = self.options.get_safe("with_curl")
-        tc.variables["BUILD_SHARED_LIBS"] = bool(self.options.shared)
-        tc.variables["GGML_CUDA"] = self.options.get_safe("with_cuda")
-        if hasattr(self, "settings_build") and cross_building(self):
+        if cross_building(self):
             tc.variables["LLAMA_NATIVE"] = False
+            tc.variables["GGML_NATIVE_DEFAULT"] = False
+
+        tc.variables["GGML_BUILD_TESTS"] = False
+        #tc.variables["GGML_BUILD_EXAMPLES"] = self.options.get_safe("with_examples")
```
czoido, line 77: Is this a leftover?

recipes/llama-cpp/all/conanfile.py
```diff
@@ -114,23 +99,80 @@ def package(self):
...
+            self.cpp_info.components["common"].includedirs = [os.path.join("include", "common")]
+            self.cpp_info.components["common"].libs = ["common"]
+            self.cpp_info.components["common"].requires = ["llama"]
+            if self.options.with_curl:
+                self.cpp_info.components["common"].requires.append("libcurl::libcurl")
+                self.cpp_info.components["common"].defines.append("LLAMA_USE_CURL")
+            if is_apple_os(self):
+                self.cpp_info.components["common"].frameworks.extend(["Foundation", "Accelerate", "Metal"])
+            elif self.settings.os in ("Linux", "FreeBSD"):
+                self.cpp_info.components["common"].system_libs.extend(["dl", "m", "pthread", "gomp"])
```
czoido, line 121: I think this code is common for all versions?

recipes/llama-cpp/all/conandata.yml
AbrilRBS: Good catch, my bad!

recipes/llama-cpp/all/conanfile.py
```diff
@@ -114,23 +99,80 @@ def package(self):
...
+
+    def package_info(self):
+        # New structure in llama
+        if Version(self.version) >= "b4570":
+            self.cpp_info.components["common"].includedirs = [os.path.join("include", "common")]
+            self.cpp_info.components["common"].libs = ["common"]
+            self.cpp_info.components["common"].requires = ["llama"]
+            if self.options.with_curl:
+                self.cpp_info.components["common"].requires.append("libcurl::libcurl")
+                self.cpp_info.components["common"].defines.append("LLAMA_USE_CURL")
```
czoido, line 117: I think we don't need to define this in the package_info?

recipes/llama-cpp/all/conanfile.py
```diff
@@ -114,23 +99,80 @@ def package(self):
...
+        return results
+
+    def package_info(self):
+        # New structure in llama
+        if Version(self.version) >= "b4570":
+            self.cpp_info.components["common"].includedirs = [os.path.join("include", "common")]
+            self.cpp_info.components["common"].libs = ["common"]
+            self.cpp_info.components["common"].requires = ["llama"]
+            if self.options.with_curl:
+                self.cpp_info.components["common"].requires.append("libcurl::libcurl")
```
czoido, line 116: shouldn't this be defined for both versions?

recipes/llama-cpp/all/conanfile.py
```diff
@@ -114,23 +99,80 @@ def package(self):
...
+            results.append("blas")
+            results.append("metal")
+        return results
+
+    def package_info(self):
+        # New structure in llama
+        if Version(self.version) >= "b4570":
+            self.cpp_info.components["common"].includedirs = [os.path.join("include", "common")]
+            self.cpp_info.components["common"].libs = ["common"]
+            self.cpp_info.components["common"].requires = ["llama"]
```
czoido, line 114: Is this true? I'm not sure that llama requires common, this requires was not defined for previous versions

recipes/llama-cpp/all/conanfile.py
```diff
@@ -114,23 +99,80 @@ def package(self):
...
-            self.cpp_info.components["common"].system_libs.extend(["dl", "m", "pthread", "gomp"])
+            results.append("blas")
+            results.append("metal")
+        return results
+
+    def package_info(self):
+        # New structure in llama
+        if Version(self.version) >= "b4570":
+            self.cpp_info.components["common"].includedirs = [os.path.join("include", "common")]
+            self.cpp_info.components["common"].libs = ["common"]
```
czoido, line 113: We can move this out from the if Version and use for all versions

recipes/llama-cpp/all/conanfile.py
```diff
@@ -114,23 +99,80 @@ def package(self):
...
+                self.cpp_info.components["ggml"].defines.append(f"GGML_USE_{backend.upper()}")
+                self.cpp_info.components["ggml"].requires.append(f"ggml-{backend}")
+
+            if is_apple_os(self):
+                self.cpp_info.components["ggml-blas"].frameworks.append("Accelerate")
+
+            if self.options.with_cuda and not self.options.shared:
+                self.cpp_info.builddirs.append(os.path.join("lib", "cmake"))
+                module_path = os.path.join("lib", "cmake", "llama-cpp-cuda-static.cmake")
+                self.cpp_info.set_property("cmake_build_modules", [module_path])
```
czoido, line 158: this can also be shared between both versions

recipes/llama-cpp/all/conanfile.py
AbrilRBS: More of a reminder that this currently breaks the build for examples for some reason, wanted to re-check today, but will remove it if I make no progress

recipes/llama-cpp/all/conanfile.py
AbrilRBS: `target_link_libraries     (${TARGET} PRIVATE ${LLAMA_COMMON_EXTRA_LIBS} PUBLIC llama Threads::Threads)` with `TARGET` being `common` here too

recipes/llama-cpp/all/conanfile.py
AbrilRBS: Upstream's CMakeLists contains `target_compile_definitions(${TARGET} PUBLIC LLAMA_USE_CURL)` with `TARGET` being `common`

czoido: And is this new from this version or should we add it to the old versions too?

recipes/llama-cpp/all/conanfile.py
AbrilRBS: No, because if no requires is explicitly defined, all components get all the library requirements. I can be more explicity to avoid overlinking in old versions

recipes/llama-cpp/all/conanfile.py
AbrilRBS: Seems like it was missing for the old versions too, as this same line is present in the previous version too

recipes/llama-cpp/all/conanfile.py
```diff
@@ -99,30 +105,40 @@ def package(self):
...
-            self.cpp_info.components["common"].libs = ["common"]
-            self.cpp_info.components["common"].requires = ["llama"]
-            if self.options.with_curl:
-                self.cpp_info.components["common"].requires.append("libcurl::libcurl")
-                self.cpp_info.components["common"].defines.append("LLAMA_USE_CURL")
-            if is_apple_os(self):
-                self.cpp_info.components["common"].frameworks.extend(["Foundation", "Accelerate", "Metal"])
-            elif self.settings.os in ("Linux", "FreeBSD"):
-                self.cpp_info.components["common"].system_libs.extend(["dl", "m", "pthread", "gomp"])
+        self.cpp_info.components["llama"].set_property("cmake_target_aliases", ["common"])
```
czoido, line 118:
```suggestion
```

czoido, line 118:
```suggestion
```

czoido, line 118:
```suggestion
```

recipes/llama-cpp/all/conandata.yml
```diff
@@ -1,17 +1,17 @@
...
-    url: "https://github.com/ggerganov/llama.cpp/archive/refs/tags/b3040.tar.gz"
-    sha256: "020e040139660eb40113503bb1057d5387677d249b990e008e04821532f7cd62"
 patches:
+  "b4570":
+    - patch_file: "patches/b4570-001-curl-patch-targets.patch"
   "b4079":
     - patch_file: "patches/b4079-001-curl-patch-targets.patch"
   "b3542":
     - patch_file: "patches/b3542-001-curl-patch-targets.patch"
-  "b3040":
```
czoido, line 16: This version had a slightly different logic, because did not have the ggml component so it the recipe was a bit cleaner removing this, and it's also quiet outdated

recipes/llama-cpp/all/conanfile.py
```diff
@@ -37,36 +37,35 @@ class LlamaCppConan(ConanFile):
...
-        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
-        if minimum_version and Version(self.settings.get_safe("compiler.version")) < minimum_version:
-            raise ConanInvalidConfiguration(
-                f"{self.ref} requires {str(self.settings.compiler)}>={minimum_version}."
-            )
+        check_min_cppstd(self, 17 if self._is_new_llama else 11)
+
+    def validate_build(self):
+        if self._is_new_llama and self.settings.compiler == "msvc" and "arm" in self.settings.arch:
+            raise ConanInvalidConfiguration("llama-cpp does not support ARM architecture on msvc, it recommends to use clang instead")
```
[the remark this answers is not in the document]
AbrilRBS: Good catch, sorry about that:

`ggml/src/ggml-cpu/CMakeLists.txt` has this when adding ggml-cpu:

```
if (MSVC AND NOT CMAKE_C_COMPILER_ID STREQUAL "Clang")
            message(FATAL_ERROR "MSVC is not supported for ARM, use clang")
        else()
```

As the cpu backend is always enabled in the recipe, I just validate this case out - we can add an optin for the cpu backend if we think it's a better approach too

recipes/llama-cpp/all/test_package/CMakeLists.txt
```diff
@@ -5,5 +5,5 @@ project(test_package CXX)
 find_package(llama-cpp REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
-target_link_libraries(${PROJECT_NAME} PRIVATE llama-cpp::llama llama-cpp::common)
+target_link_libraries(${PROJECT_NAME} PRIVATE llama common)
```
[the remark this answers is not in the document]
AbrilRBS: Yes! The generated targets outside conan are `llama` etc, and the only namespaced ones are `ggml::<backend>`s. Discussing it with @czoido we saw that as we already don't follow the file_name, we can probably keep these namespaced to avoid collisions, even if it means deviating from upstream

recipes/llama-cpp/all/conanfile.py
```diff
@@ -37,36 +37,35 @@ class LlamaCppConan(ConanFile):
...
 
     @property
-    def _compilers_minimum_version(self):
-        return {
-            "gcc": "8"
-        }
+    def _cuda_build_module(self):
+        cuda_target = "ggml-cuda" if self._is_new_llama else "ggml"
+        return textwrap.dedent(f"""\
+            find_dependency(CUDAToolkit REQUIRED)
```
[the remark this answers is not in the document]
czoido: Sorry, I think this may be a bit confusing as you are pointing out. I'm going to add a comment to clarify that as we don't have a CUDA recipe (yet) this is needed to propagate that info for consumers that enable the cuda option in llama-cpp

jcar87 (approved):

PR #26479, opened 2025-01-29
[breakpad] Fix compilation with gcc-13

artem-ogre:
### Summary
Changes to recipe:  **breakpad/cci.20210521**

#### Motivation
Breakpad would not compile with gcc-13 on Ubuntu 24.02 Noble Numbat

#### 1. Fix missing include
See for reference: https://gcc.gnu.org/gcc-13/porting_to.html

> Some C++ Standard Library headers have been changed to no longer include other headers that were being used internally by the library. As such, C++ programs that used standard library components without including the right headers will no longer compile.
> 
> The following headers are used less widely in libstdc++ and may need to be included explicitly when compiling with GCC 13:
> - <string> (for std::string, std::to_string, std::stoi etc.)
> - <system_error> (for std::error_code, std::error_category, std::system_error).
> - <cstdint> (for std::int8_t, std::int32_t etc.)
> - <cstdio> (for std::printf, std::fopen etc.)
> - <cstdlib> (for std::strtol, std::malloc etc.)

In particular header `<cstdint>` now needs to be included directly to compile with gcc-13

#### 2. Fix `-Wnonnull` error
`fclose` is called on a null file (see the if-check above it). The patch removes the call to `fclose`

### DIFF
--- a/recipes/breakpad/all/patches/0004-Fix-gcc13-compilation.patch
+++ b/recipes/breakpad/all/patches/0004-Fix-gcc13-compilation.patch
@@ -8,13 +8,3 @@
  #include <assert.h>
  #include <sys/types.h>
  
---- a/configure
-+++ b/configure
-@@ -7476,7 +7476,6 @@
- 
- done
- 
--as_fn_append WARN_CXXFLAGS " -Werror"
- ac_ext=c
- ac_cpp='$CPP $CPPFLAGS'
- ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
--- a/recipes/breakpad/all/patches/0004-Fix-gcc13-compilation.patch
+++ b/recipes/breakpad/all/patches/0004-Fix-gcc13-compilation.patch
@@ -1,0 +1,20 @@
+--- a/src/client/linux/handler/minidump_descriptor.h
++++ b/src/client/linux/handler/minidump_descriptor.h
+@@ -30,6 +30,7 @@
+ #ifndef CLIENT_LINUX_HANDLER_MINIDUMP_DESCRIPTOR_H_
+ #define CLIENT_LINUX_HANDLER_MINIDUMP_DESCRIPTOR_H_
+ 
++#include <cstdint>
+ #include <assert.h>
+ #include <sys/types.h>
+ 
+--- a/configure
++++ b/configure
+@@ -7476,7 +7476,6 @@
+ 
+ done
+ 
+-as_fn_append WARN_CXXFLAGS " -Werror"
+ ac_ext=c
+ ac_cpp='$CPP $CPPFLAGS'
+ ac_compile='$CC -c $CFLAGS $CPPFLAGS conftest.$ac_ext >&5'
--- a/recipes/breakpad/all/conandata.yml
+++ b/recipes/breakpad/all/conandata.yml
@@ -7,3 +7,4 @@ patches:
     - patch_file: "patches/0001-Use_conans_lss.patch"
     - patch_file: "patches/0002-Remove-hardcoded-fpic.patch"
     - patch_file: "patches/0003-Fix-gcc11-compilation.patch"
+    - patch_file: "patches/0004-Fix-gcc13-compilation.patch"
--- a/recipes/breakpad/all/conandata.yml
+++ b/recipes/breakpad/all/conandata.yml
@@ -8,3 +8,4 @@ patches:
     - patch_file: "patches/0002-Remove-hardcoded-fpic.patch"
     - patch_file: "patches/0003-Fix-gcc11-compilation.patch"
     - patch_file: "patches/0004-Fix-gcc13-compilation.patch"
+    - patch_file: "patches/0005-Apply-upstream-fixes.patch"
--- a/recipes/breakpad/all/patches/0005-Apply-upstream-fixes.patch
+++ b/recipes/breakpad/all/patches/0005-Apply-upstream-fixes.patch
@@ -1,0 +1,24 @@
+--- a/src/processor/exploitability_linux.cc
++++ b/src/processor/exploitability_linux.cc
+@@ -516,18 +516,17 @@
+            raw_bytes_tmpfile);
+   FILE* objdump_fp = popen(cmd, "r");
+   if (!objdump_fp) {
+-    fclose(objdump_fp);
+     unlink(raw_bytes_tmpfile);
+     BPLOG(ERROR) << "Failed to call objdump.";
+     return false;
+   }
+-  if (fread(objdump_output_buffer, 1, buffer_len, objdump_fp) <= 0) {
+-    fclose(objdump_fp);
++  if (fread(objdump_output_buffer, 1, buffer_len, objdump_fp) != buffer_len) {
++    pclose(objdump_fp);
+     unlink(raw_bytes_tmpfile);
+     BPLOG(ERROR) << "Failed to read objdump output.";
+     return false;
+   }
+-  fclose(objdump_fp);
++  pclose(objdump_fp);
+   unlink(raw_bytes_tmpfile);
+   return true;
+ }


recipes/breakpad/all/patches/0004-Fix-gcc13-compilation.patch
```diff
@@ -0,0 +1,20 @@
...
+ #include <assert.h>
+ #include <sys/types.h>
+ 
+--- a/src/processor/exploitability_linux.cc
++++ b/src/processor/exploitability_linux.cc
+@@ -516,7 +516,6 @@
+            raw_bytes_tmpfile);
+   FILE* objdump_fp = popen(cmd, "r");
+   if (!objdump_fp) {
+-    fclose(objdump_fp);
```
jcar87, line 17: Thanks @artem-ogre  -
Has this issue been reported upstream?

artem-ogre: Sure, Luis, thanks for looking into it!
Here's the output that I get if I don't remove `fclose`:
<pre>
<b>/home/artem/.conan2/p/b/breakfa9cf26fba332/b/src/src/processor/exploitability_linux.cc:</b> In static member function ‘<b>static bool google_breakpad::ExploitabilityLinux::</b><font color="#8AE234"><b>DisassembleBytes</b></font>(const std::string&amp;, const uint8_t*, unsigned int, char*)’:
<b>/home/artem/.conan2/p/b/breakfa9cf26fba332/b/src/src/processor/exploitability_linux.cc:519:11:</b> <font color="#EF2929"><b>error: </b></font>argument 1 null where non-null expected [<font color="#EF2929"><b>-Werror=nonnull</b></font>]
  519 |     <font color="#EF2929"><b>fclose(objdump_fp)</b></font>;
      |     <font color="#EF2929"><b>~~~~~~^~~~~~~~~~~~</b></font>
In file included from <b>/usr/include/c++/13/cstdio:42</b>,
                 from <b>/usr/include/c++/13/ext/string_conversions.h:45</b>,
                 from <b>/usr/include/c++/13/bits/basic_string.h:4109</b>,
                 from <b>/usr/include/c++/13/string:54</b>,
                 from <b>/usr/include/c++/13/bits/locale_classes.h:40</b>,
                 from <b>/usr/include/c++/13/bits/ios_base.h:41</b>,
                 from <b>/usr/include/c++/13/ios:44</b>,
                 from <b>/usr/include/c++/13/ostream:40</b>,
                 from <b>/usr/include/c++/13/iostream:41</b>,
                 from <b>/home/artem/.conan2/p/b/breakfa9cf26fba332/b/src/src/google_breakpad/processor/minidump.h:88</b>,
                 from <b>/home/artem/.conan2/p/b/breakfa9cf26fba332/b/src/src/google_breakpad/processor/exploitability.h:44</b>,
                 from <b>/home/artem/.conan2/p/b/breakfa9cf26fba332/b/src/src/processor/exploitability_linux.h:41</b>,
                 from <b>/home/artem/.conan2/p/b/breakfa9cf26fba332/b/src/src/processor/exploitability_linux.cc:37</b>:
<b>/usr/include/stdio.h:184:12:</b> <font color="#34E2E2"><b>note: </b></font>in a call to function ‘<b>int</b><font color="#8AE234"><b> fclose</b></font>(FILE*)’ declared ‘<b>nonnull</b>’
  184 | extern int <font color="#34E2E2"><b>fclose</b></font> (FILE *__stream) __nonnull ((1));
      |            <font color="#34E2E2"><b>^~~~~~</b></font>
</pre>

artem-ogre: I also see from the output that `-Werror` is enabled, maybe alternative would be to patch the configuration not to set it.

artem-ogre: I modified the fix to disable `-Werror` that's also much better for the future proofing. Please let me know what you think.

uilianries:
@artem-ogre Hello and thank you for providing this suggested patch.

Still, the current master branch of Breakpad has no hotfix integrated: https://chromium.googlesource.com/breakpad/breakpad/+/refs/heads/main/src/client/linux/handler/minidump_descriptor.h

I also did not find any occurrence of this patch or bug mentioned in upstream's bug list: https://issues.chromium.org/issues?q=componentid:1629899%2B%20cstdint

Could you please confirm the upstream is aware of this bug somehow?

I can mention Gentoo has a similar patch, but they didn't change the -Werror: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/breakpad/files/breakpad-2022.07.12-gcc13.patch

artem-ogre:
@uilianries 
The `fclose` fix was for this file `src/processor/exploitability_linux.cc`.

The master (https://chromium.googlesource.com/breakpad/breakpad/+/refs/heads/main/src/processor/exploitability_linux.cc) does not call `fclose`. Looks like it was heavily refactored since the snapshot used in the recipe.
This should explain the points you raised. Thanks for the thorough research by the way!

artem-ogre:
As a side note: You have a lot of experience supporting packaging libraries. You probably have the same opinion that treating warnings as errors is not a great idea when providing a library source to the users (perfectly fine for the development workflow). It is rather unconventional that breakpad does this.

artem-ogre:
@uilianries Here, I found `fclose` fix in master: https://chromium.googlesource.com/breakpad/breakpad/+/30c7f3cfc11cdbf93a12efbe9d46c66d9785879e%5E%21/

uilianries:
@artem-ogre Thank you for pointing out those sources. I can reproduce your reported error by using GCC13 on Linux: [breakpad-20210521-linux-gcc13.log](https://github.com/user-attachments/files/18605203/breakpad-20210521-linux-gcc13.log)

When applying a patch ONLY to include cstdint is passes for me, no further errors:

[breakpad-20210521-linux-gcc13-patched.log](https://github.com/user-attachments/files/18605267/breakpad-20210521-linux-gcc13-patched.log)

I would request to simplify your patch to only include cstdint. I understand your concern about, but we have several projects using Werror that don't cause problems for us.

uilianries:
@artem-ogre out of curiosity, the Breakpad project offers the newer tag 2023.06.01. Meanwhile, the version maintained in CCI is based on 20210521. Do you need a newer package version?

artem-ogre:
Yes, if it builds for your configuration, it does not mean it has to build for every g++ 13 configuration.
It could be that our compilers are not exactly the same versions, or it could be standard library are not the same.
For example your compilation command has `-std=gnu++20` and mine does not.
Feel free to diff them yourself:
[artem-command.txt](https://github.com/user-attachments/files/18605456/artem-command.txt)
[uilian-command.txt](https://github.com/user-attachments/files/18605455/uilian-command.txt)

I would appreciate if you could indicate a way forward so that it also builds for my configuration :)
If you are not willing to accept my patch please indicate how else can this be fixed.

artem-ogre:
> @artem-ogre out of curiosity, the Breakpad project offers the newer tag 2023.06.01. Meanwhile, the version maintained in CCI is based on 20210521. Do you need a newer package version?

Right now my main concern is to make the existing configuration work with Ubuntu 24.04 🙂
But if there is a more recent version in the conan-center-index I would be happy to look into upgrading the dependency. We only use a fraction of breakpad's functionality.

artem-ogre:
Maybe if I share my profile and compiler info to help you reproduce the problem?

I'm on Ubuntu 24.04 Noble Numbat.

<pre><font color="#FCE94F"><b>└─$ </b></font>g++ --version
g++ (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
Copyright (C) 2023 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
</pre>

My conan profile looks something like this:
```
[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.libcxx=libstdc++11
compiler.version=13
build_type=Release

[conf]
tools.system.package_manager:mode = install
tools.system.package_manager:sudo = True
```

artem-ogre:
Would it be fine applying a patch similar to this one: https://chromium.googlesource.com/breakpad/breakpad/+/30c7f3cfc11cdbf93a12efbe9d46c66d9785879e%5E%21/
and leaving `-Werror` alone?

uilianries:
> @uilianries Yes, it builds for your configuration, but it does not build for mine. It could be that our compilers are not exactly the same versions, or it could be standard library are not the same. For example your compilation command has `-std=gnu++20` and mine does not. Feel free to diff them yourself: [artem-command.txt](https://github.com/user-attachments/files/18605456/artem-command.txt) [uilian-command.txt](https://github.com/user-attachments/files/18605455/uilian-command.txt)
> 
> I would appreciate if you could indicate a way forward so that it also builds for my configuration :) If you are not willing to accept my patch please indicate how else can this be fixed.

Indeed there is something more peculiar. I started a new Docker container with Ubuntu 24.04 and GCC13 installed. Now I can reproduce your case regarding the fclose:

[breakpad-20210521-ubuntu24.04-gcc13.log](https://github.com/user-attachments/files/18605611/breakpad-20210521-ubuntu24.04-gcc13.log)

uilianries:
> Would it be fine applying a patch similar to this one: [chromium.googlesource.com/breakpad/breakpad/+/30c7f3cfc11cdbf93a12efbe9d46c66d9785879e%5E%21](https://chromium.googlesource.com/breakpad/breakpad/+/30c7f3cfc11cdbf93a12efbe9d46c66d9785879e%5E%21/) and leaving `-Werror` alone?

Yes, please. I just tested using that patch it really works:

[breakpad-patch-fclose.log](https://github.com/user-attachments/files/18605713/breakpad-patch-fclose.log)

Please, simplify your current patch to include cstdint. Then add a second patch, only addressed to this official one.

Thank you for keeping pushing this PR.

uilianries (approved):
LGTM. Both patches are okay, they are integrated in the upstream or by other package managers.

- Patch for fclose prevents -Wnonnull: https://chromium.googlesource.com/breakpad/breakpad/+/30c7f3cfc11cdbf93a12efbe9d46c66d9785879e%5E%21/
- Patch for cstdint is integrated in Gentoo: https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/breakpad/files/breakpad-2022.07.12-gcc13.patch 

Thank you for your PR!

AbrilRBS (approved):

artem-ogre:
@uilianries 
> I understand your concern about, but we have several projects using Werror that don't cause problems for us.
Just wanted to express my opinion on this: treating warnings as errors for a library provided as source is like having a ticking code-rot bomb. It limits the users in how they can build the library. Just one new warning introduced in a more recent version of the compiler and the bomb goes off :)

I'm rather confident that this breakpad version will break with more recent gcc versions. It happened in the past and will happen in the future.

uilianries:
> Just wanted to express my opinion on this: treating warnings as errors for a library provided as source is like having a ticking code-rot bomb. It limits the users in how they can build the library. Just one new warning introduced in a more recent version of the compiler and the bomb goes off :)
>
> I'm rather confident that this breakpad version will break with more recent gcc versions. It happened in the past and will happen in the future.

@artem-ogre Thank you for expressing your concern, indeed it may break in the future. As you commented about using a newer version of Breakpad, it could be included in a patch for -Werror a next PR, to avoid more delays in this one. In case it's a blocker for you right now, please, feel free to share your build log again.

artem-ogre:
Can this be merged? Anything else I can do?

artem-ogre:
Looks like the pre-checks are stuck?

PR #26481, opened 2025-01-29
[libe57format] Fix compilation with gcc-13

artem-ogre:
See: https://gcc.gnu.org/gcc-13/porting_to.html

### Summary
Changes to recipe:  **libe57format/[2.3.0]**

#### Motivation
Does not compile with gcc-13

#### Details
See: https://gcc.gnu.org/gcc-13/porting_to.html
`<cstdint>` now needs to be included directly.

### DIFF
--- a/recipes/libe57format/all/conandata.yml
+++ b/recipes/libe57format/all/conandata.yml
@@ -1,3 +1,6 @@
+patches:
+  "2.3.0":
+  - patch_file: patches/0001-Fix-gcc13-compilation.patch
 sources:
   "3.2.0":
     url: "https://github.com/asmaloney/libE57Format/archive/refs/tags/v3.2.0.tar.gz"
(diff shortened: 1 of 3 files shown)


artem-ogre:
Rebased on top of master.

jcar87:
> Rebased on top of master.

Thanks @artem-ogre  - bear in mind that rebasing on top of master is only necessary when the recipe has been updated in master - if that is not the case, it tends to be a little counter productive, as it causes CI to be relaunched. If a PR is merged to master that modifies a recipe, all other PRs that touch the same recipe are "invalidated" - it's only in those cases that rebasing is necessary so that the new recipe revision can be computed by CI.

artem-ogre:
@jcar87 
The recipe was updated on master :)

artem-ogre:
@uilianries @AbrilRBS 
Can you please approve this PR? It is doing exactly the same as https://github.com/conan-io/conan-center-index/pull/26479 so hopefully it's easy for you to do.

artem-ogre:
Can this be merged? Anything else I can do?

Issue #26480, opened 2025-01-29
[package] poco/*: dependency "https://github.com/conan-io/conan-center-index/blob/master/recipes/sqlite3/all/conandata.yml" not available in conan-center-index


schwaerz:
### Description

I wanted to rebuild poco/1.12.4 from source. It was complaining about a missing dependency: sqlite3/3.45.0.

I checked the conanfile for poco, there is a hard-coded reference to that dependency if the sqlite support should be enabled: https://github.com/conan-io/conan-center-index/blob/master/recipes/poco/all/conanfile.py#L161

In sqlite conandata.yml, however, it is not available: https://github.com/conan-io/conan-center-index/blob/master/recipes/sqlite3/all/conandata.yml

Supposedly, this will be an easy fix. I'll give it a shot.

### Package and Environment Details

* Package Name/Version: **poco/1.12.4**
* Operating System+version: **Linux Ubuntu 22.04**
* Compiler+version: **GCC 8**
* Docker image: **ubuntu20.04 + gcc8 (custom)**
* Conan version: **conan 2.11.0**
* Python version: **Python 3.8.x**

### Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=17
compiler.libcxx=libstdc++11
compiler.version=8
os=Linux
os.distro=Ubuntu
os.distro.version=20.04
[options]
enable_data_postgresql=False
[buildenv]
CC=/usr/bin/gcc-8
CXX=/usr/bin/g++-8

### Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=17
compiler.libcxx=libstdc++11
compiler.version=8
os=Linux
os.distro=Ubuntu
os.distro.version=20.04
[buildenv]
CC=/usr/bin/gcc-8
CXX=/usr/bin/g++-8

### Steps to reproduce

conan install -r poco/1.12.4 --build=missing --profile:all=ubuntu20.0_gcc8_release

### Logs

<details><summary>Click to expand log</summary>

```
2025-01-29 11:29:36,334 INFO ---- build poco/1.12.4#b14e8a5d84932bf65a28c1191437e294 ----
WARN: legacy: Unscoped option definition is ambiguous.
Use '&:enable_data_postgresql=False' to refer to the current package.
Use '*:enable_data_postgresql=False' or other pattern if the intent was to apply to dependencies
======== Input profiles ========
Profile host:
[settings]
arch=x86_64
[...]

PR #26482, opened 2025-01-29
Fix sqlite3 dependencies

schwaerz:
3.45.0 -> 3.45.3: 26480

### Summary
Changes to recipes:  
**apr-util/all**
**dlib/all**
**drogon/all**
**oatpp-sqlite/all**
**poco/all**
**sqlite_orm/all**
**sqlitecpp/all**

#### Motivation
[Link to the bug: ](https://github.com/conan-io/conan-center-index/issues/26480)

#### Details
Just replacing all instances of where the no-longer-existing `sqlite3/3.45.0` was a requirement by the existing one `sqlite3/3.45.3`

---

### DIFF
--- a/recipes/apr-util/all/conanfile.py
+++ b/recipes/apr-util/all/conanfile.py
@@ -88,7 +88,7 @@ class AprUtilConan(ConanFile):
         if self.options.with_mysql:
             self.requires("libmysqlclient/8.1.0")
         if self.options.with_sqlite3:
-            self.requires("sqlite3/3.45.0")
+            self.requires("sqlite3/3.45.3")
         if self.options.with_expat:
             self.requires("expat/[>=2.6.2 <3]")
         if self.options.with_postgresql:
--- a/recipes/dlib/all/conanfile.py
+++ b/recipes/dlib/all/conanfile.py
@@ -96,7 +96,7 @@ class DlibConan(ConanFile):
         if self.options.get_safe("with_webp"):
             self.requires("libwebp/1.3.2")
         if self.options.with_sqlite3:
-            self.requires("sqlite3/3.45.0")
+            self.requires("sqlite3/3.45.3")
         if self.options.with_openblas:
             self.requires("openblas/0.3.26")
 
--- a/recipes/drogon/all/conanfile.py
+++ b/recipes/drogon/all/conanfile.py
@@ -130,7 +130,7 @@ class DrogonConan(ConanFile):
         if self.options.get_safe("with_mysql"):
             self.requires("libmysqlclient/8.1.0")
         if self.options.get_safe("with_sqlite"):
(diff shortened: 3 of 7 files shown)


schwaerz:
Not sure whether I have to do anything so you can proceed with reviewing this PR?

valgur:
1. Please use `sqlite3/[>=3.45.0 <4]` for consistency with the majority of sqlite dependencies on ConanCenter.
2. It might not get merged: https://github.com/conan-io/conan-center-index/pull/25961#issuecomment-2572982030

schwaerz:
@valgur If I understand the comment in #25961 correctly, I should rather go for changing one recipe at a time? As we're only facing this issue with poco right now, I'd be totally fine with that. Just though: Why not fix the other recipes, too ;)

But probably this is already a chunk small enough for a merge?

jcar87:
Hi @schwaerz  - each recipe is currently on 5 configurations, 10 if the recipe has a `shared` option - if the recipe has say, 5 versions in the `config.yml` thats would be 50 packages per recipe. Some recipes have far more than 5 currently maintained versions - so we want to avoid any PR causing a significant resource utilisation (that could also take several hours). So while CI supports its - the guidance is the same as it has always been in Conan Center: 1 recipe per PR.  Although this one is a lot smaller than the one recently closed.

The issue https://github.com/conan-io/conan-center-index/issues/26480 is not fully clear - the Conan Center remote does have `sqlite3/3.45.0` - even when recipes are no longer in the most recent git revision, they remain in the Conan Center remote.

Either way - yes, please refactor the PR to the minimal that solves your problem - it's a lot easier for us to review and quicker to run on CI.

ericLemanissier:
The problem is that some consumers do not rely on binaries or recipes provided by conan-center (which seems to be the case reported by https://github.com/conan-io/conan-center-index/issues/26480). This can be mandatory in cases where consumer need to host everything themselves, and not rely on a third party remote. In such cases, we can simply maintain a private clone of conan-center-index, and pull from upstream when needed. As you can imagine, these use-cases gain a lot in maintenance if conan-center-index is in a consistent state most of the time, referencing only recipes present in conan-center-index.

schwaerz:
@jcar87 Thanks for the reply. We have the same or similar use-case as described by @ericLemanissier: We do not have direct access to the Conan center internet server from CI. Also, some binaries for certain configurations are (or where?) (understandably, effort-wise) missing.

What I'll do: Minimize this PR to the absolute minimum as suggested from your end. Hoping that you'll be able to approve it then.

Thanks.

schwaerz:
@jcar87 Sorry it took me some time to finalize this (testing on our end). But I was able to reduce the changes to a bare minimum. Please check.

franramirez688 (approved):

AbrilRBS (approved):

PR #26500, opened 2025-01-31
libffi: fix conditionals when building on Windows

root-kidik:
bug

conan profile
```
[settings]
os=Windows
arch=x86_64
compiler=clang
compiler.version=18
compiler.libcxx=libc++
build_type=Debug
compiler.cppstd=20

[conf]
```

conanfile.txt
```
[requires]
qt/6.7.3

[generators]
CMakeDeps
CMakeToolchain

[layout]
cmake_layout
```

error
```
ERROR: libffi/3.4.4: Error in generate() method, line 130
        ar_wrapper = unix_path(self, self.dependencies.build["automake"].conf_info.get("user.automake:lib-wrapper"))
        KeyError: "'automake' not found in the dependency set"
```

### DIFF
--- a/recipes/libffi/all/conanfile.py
+++ b/recipes/libffi/all/conanfile.py
@@ -62,7 +62,7 @@ class LibffiConan(ConanFile):
             self.win_bash = True
             if not self.conf.get("tools.microsoft.bash:path", default=False, check_type=str):
                 self.tool_requires("msys2/cci.latest")
-        if is_msvc(self):
+        if is_msvc(self) or self.settings.compiler == "clang":
             self.tool_requires("automake/1.16.5")
 
     def source(self):
--- a/recipes/libffi/all/conanfile.py
+++ b/recipes/libffi/all/conanfile.py
@@ -62,8 +62,7 @@ class LibffiConan(ConanFile):
             self.win_bash = True
             if not self.conf.get("tools.microsoft.bash:path", default=False, check_type=str):
                 self.tool_requires("msys2/cci.latest")
-        if is_msvc(self) or self.settings.compiler == "clang":
-            self.tool_requires("automake/1.16.5")
+        self.tool_requires("automake/1.16.5")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
--- a/recipes/libffi/all/conanfile.py
+++ b/recipes/libffi/all/conanfile.py
@@ -11,7 +11,7 @@ import glob
 import os
 import shutil
 
-required_conan_version = ">=1.57.0"
+required_conan_version = ">=2.0"
 
 
 class LibffiConan(ConanFile):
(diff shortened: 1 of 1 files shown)


root-kidik:
@amoeba @cynix

root-kidik:
@memsharded

recipes/libffi/all/conanfile.py
```diff
@@ -62,7 +62,7 @@ def build_requirements(self):
             self.win_bash = True
             if not self.conf.get("tools.microsoft.bash:path", default=False, check_type=str):
                 self.tool_requires("msys2/cci.latest")
-        if is_msvc(self):
+        if is_msvc(self) or self.settings.compiler == "clang":
```
cynix, line 65: Since this recipe always uses autotools regardless of OS/compiler, I think we should just get rid of this condition and always require automake.

cynix: However, if that's undesirable and you want to only require this to satisfy the reference below, then this condition should be consistent with line 94:

https://github.com/root-kidik/conan-center-index/blob/4687515fa9ab64d2bd669bedef7e8882c984aa05/recipes/libffi/all/conanfile.py#L94

root-kidik: @cynix I took your advice and added that an automake is always required

jcar87: Apologies for the confusion here - but guarding automake for msvc only is intentional - we do expect those tools to be available on other systems (and this should be restored to what it was before)

cynix: In that case it feels like this should be a requirement on Windows rather than MSVC?

FWIW, I’ve had to fork at least 1 recipe (cyrus-sasl) for our internal use, to add the automake/1.16.5 requirement, because the highest version of automake available in our environment (CentOS 7) was insufficient for that package. 

It seems CCI provides recipes for many modern build tools. Isn’t it better for packages to always declare their tool requirements? If the user wishes to use the tools from their environment rather than a package, they then have the choice of defining those via `platform_tool_requires`.

root-kidik: @cynix I need to change something in my PR?

root-kidik: @cynix @jcar87

jcar87: Thanks @root-kidik  - I've had a closer look and there was indeed a mismatch between the conditionals. 
Based on the issue you are reporting, it seems like you are using clang inside an msys2 environment or similar, correct? In that case - the build should use autotools from that environment and not require the additional dependency. 
If you are using clang from Visual Studio, your profile is missing the `compiler.runtime` setting (either `dyanmic` or `static`)

root-kidik: yep, clang inside an msys2

root-kidik: @cynix @jcar87 - can we merge it?

jcar87 (approved):

recipes/libffi/all/conanfile.py
cynix: Sorry to hijack the thread...

@jcar87 any guidance regarding my comment above? If the convention is to _not_ `tool_require(automake/...)` in the recipe for POSIX environments, how do I inject this dependency myself (via command line or profile or something) when building packages in an environment that does not have a recent enough automake version?

PR #26508, opened 2025-02-02
SFML add support for new v3

JackCarterSmith:
### Summary
Added new recipe version:  **sfml/3.0.0**

#### Motivation
Release of SFML version 3.0.0

#### Details
- Change in CMake targets name between v2 and v3.
- Added missing glad requirement.
- Updated test_package to v3.
- Fix naming in libraries (-s and -d suffix).
- Added cppstd requirement of 17 (as required by SFMLv3).
- Replaced openal by miniaudio.
- Conan requirement changed to 2.0.0.
- CMake requirement changed to 3.23.
- I think it's cleaner to separate v2 scripts from v3 scripts, as v2 shouldn't be evolving from now on.
(better not to have several branches between versions 2 and 3 in the same file)

---
- [ ] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240):
  - There is an another one (#26276) still active but it lacks fixes for glad, component suffixes, openal/miniaudio replacement and I prefer to propose a version that isolates v3 from v2 rather than having them coexist in the same branch.
- [X] Tested locally with at least one configuration using a recent version of Conan: tested and retro-tested (2.6.2) on both Windows 22H2 and Linux mint 6.8.0-52 SMP PREEMPT_DYNAMIC)


fgallegosalido:
There are at least 3 other pull requests for SFML 3:

- https://github.com/conan-io/conan-center-index/pull/25305
- https://github.com/conan-io/conan-center-index/pull/26266
- https://github.com/conan-io/conan-center-index/pull/26276

We should decide which one is going to be merged and close the other 3. Pinning @ChrisThrasher to give his opinion on this.

ChrisThrasher:
I'm afraid I'm not equipped to review all of these PRs. Someone who knows Conan better is most qualified to determine that. What I can do however is answer specific questions about how to build the library. Translating that into the right Conan code is not something I can do.

fgallegosalido:
Fair enough. Pinning @AbrilRBS too, as she's the author of the first draft and seems more qualified to review all these prs.

Sil3ntStorm:
> There is an another one (https://github.com/conan-io/conan-center-index/pull/26276) still active but it lacks fixes for glad, component suffixes, openal/miniaudio replacement and I prefer to propose a version that isolates v3 from v2 rather than having them coexist in the same branch.

What do you mean? Mine also separates the 3.x into a new recipe same as yours. Only difference seems to be that I didn't rename the old one to 2.x but instead added a 3.x one, as I had seen done in other packages, and it produces less noise in the diff.
What do you mean wrt glad? It appears to be building just fine and I had no problems using it.

JackCarterSmith:
> What do you mean? Mine also separates the 3.x into a new recipe same as yours. Only difference seems to be that I didn't rename the old one to 2.x but instead added a 3.x one, as I had seen done in other packages, and it produces less noise in the diff. What do you mean wrt glad? It appears to be building just fine and I had no problems using it.

Yes you separated v2 and v3, but you dropped the "main" component of sfml and missed the miniaudio and ogg deps (as normally required by sfml v3). The main component seem to be always generated in static. It's a dummy component only usefull on android and win32 application, compatibility purpose.
Admittedly, moving file create more noise, but the tracking of files from v2 to v3 is clearer.

I know that sfml has added miniaudio (including GLAD) directly to their sources. I think it's better to declare the dependencies explicitly (and in the event to use a different version of these deps, we can change them in the conan settings).
stb and minimp3 are already explicitly declared, it's just a continuation.

Perhaps @ChrisThrasher could tell us more about which parts/components of sfml can be dropped without any problems.

NL;DR in my opinion:
- #26266 drop components and new libs deps,
- #26276 lacks isolation between versions and new libs deps issue too.

ChrisThrasher:
> Perhaps @ChrisThrasher could tell us more about which parts/components of sfml can be dropped without any problems.

I don't think any of SFML's 6 modules (system, window, graphics, audio, network, main) can be removed, if that's what you're asking about.

Sil3ntStorm:
> Yes you separated v2 and v3, but you dropped the "main" component of sfml and missed the miniaudio and ogg deps (as normally required by sfml v3). The main component seem to be always generated in static. It's a dummy component only usefull on android and win32 application, compatibility purpose. Admittedly, moving file create more noise, but the tracking of files from v2 to v3 is clearer.

If I removed something then it caused issues on Windows and was not required to build and use sfml on Windows. 
I am using my version on Windows just fine.

> I know that sfml has added miniaudio (including GLAD) directly to their sources. I think it's better to declare the dependencies explicitly (and in the event to use a different version of these deps, we can change them in the conan settings). stb and minimp3 are already explicitly declared, it's just a continuation.

If SFML includes it and uses the internal copy of it regardless, then what good is it to declare it as a dependency when it's not actually required as a dependency / being used as a dependency?

JackCarterSmith:
> If I removed something then it caused issues on Windows and was not required to build and use sfml on Windows. I am using my version on Windows just fine.

Yeaaaah... We can't just fix a bug/issue by removing the associated warning.
I've noticed that the ‘main’ component is always build as static lib regardless of the shared setting status. Like in the PR #26276, we should add a "-s" in the filename (and -d when debug).
Removing this component could have broken the android build with SFML.

> If SFML includes it and uses the internal copy of it regardless, then what good is it to declare it as a dependency when it's not actually required as a dependency / being used as a dependency?

The first thing I would say is to indicate the dependency on the user of the conan recipe. The number of hours I waste looking for ‘hidden’ lib dependencies...
And to continue, I'm using a lib that's more up-to-date or older, but sfml is using its own hard-coded lib without knowing it. To me, that sounds like trouble.
I'm not sure now, I'd have to check that the internal libs are **in fact** overridden by conan.

(The sfml PRs must end, maybe I will abandon this PR and test/enhance @OMGtechy's one?)
I admit that getting rid of patches is a ‘sexy’ idea but I haven't been able to check them all. Keeping versions 2 and 3 isolated, in the way conan manages recipe files, doesn't seem like a good idea in hindsight, even if it makes the file more complex.
Maybe we can start from the PR 26276, check about the internal libs-dependencies and forget the isolation between version.

PR #26510, opened 2025-02-03
ninja: remove build_type option in package_id() property

tkhyn:
Ninja being an application package which is intended to be required only when added to `tool_requires`, there is no need to follow the profile's build_type.

### Summary
Changes to recipe:  **ninja/\***

#### Motivation
conan-installing ninja as tool-requires with a debug profile builds ninja in debug mode. It should not happen and the release version of ninja should be downloaded (or built) instead

#### Details
This PR excludes the `build_type` setting from the `package_id()` method (as discussed in conan-io/conan#17667)

---

### DIFF
--- a/recipes/ninja/all/conanfile.py
+++ b/recipes/ninja/all/conanfile.py
@@ -21,6 +21,7 @@ class NinjaConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def package_id(self):
+        del self.info.settings.build_type
         del self.info.settings.compiler
 
     def source(self):


jcar87:
> conan-installing ninja as tool-requires with a debug profile builds ninja in debug mode. It should not happen and the release version of ninja should be downloaded (or built) instead

In order to achieve this, when doing something like this:
When doing
```
def build_requirements(self):
     self.tool_requires("ninja/[*]")
 ```


> Ninja being an application package which is intended to be required only when added to tool_requires, there is no need to follow the profile's build_type.

We would expect the most typical scenario to be that the build_profile is in Release  - precisely to avoid a case where tools like Ninja are brought without optimisations - what does your Conan invocation look like, that results in a debug version of ninja being pulled?

tkhyn:
Hi and thanks for the feedback,

`conan install --tool-requires=ninja/[*]@` will trigger a failure to download binaries anywhere where the default profile has `build_type` set to something else than 'Release'. This is a possibility when running conan manually or through an automated tool (for example https://github.com/conan-io/cmake-conan).

I agree that this is not a very common situation but it can happen, and I think that if the aim is to make sure that tools like ninja are always shipped in release then this should be 'enforced' at the tool's recipe level.

The reason I stumbled upon this is that in my own conan 2 wrapper (https://github.com/tkhyn/cmake-conanfile) where I do not want to rely on a user's default profile for the build profile, and when not cross-compiling I copy the host profile across, which brings the `build_type` with it. I have a simple workaround (forcing the `build_type` to Release in the build profile), but I thought that to ensure that ninja was shipped in release mode it was better to do that within the recipe, as it is hinged at in the linked discussion in conan-io/conan#17667.

jcar87:
> Hi and thanks for the feedback,
> 
> `conan install --tool-requires=ninja/[*]@` will trigger a failure to download binaries anywhere where the default profile has `build_type` set to something else than 'Release'. This is a possibility when running conan manually or through an automated tool (for example https://github.com/conan-io/cmake-conan).

As far as I'm aware - `cmake-conan` for Conan 2 does *not* do that https://github.com/conan-io/cmake-conan/blob/develop2/conan_provider.cmake#L666-L667 - it very explicitly relies on the default, autodetected profile regardless.

> 
> I agree that this is not a very common situation but it can happen, and I think that if the aim is to make sure that tools like ninja are always shipped in release then this should be 'enforced' at the tool's recipe level.

I think there is a misconception here - removing the `self.info.settings.build_type` from the `package_id` causes "both" debug and release builds to map to the same package ID. It does not guarantee that when the package was created that it was always built in Release. If you are bringing the binary from Conan Center, sure, that is the case. But if you are not, and your profile has `build_type=Debug` and you happen to do `--build=missing` - then you'd end up with a Debug build of ninja.


> The reason I stumbled upon this is that in my own conan 2 wrapper (https://github.com/tkhyn/cmake-conanfile) where I do not want to rely on a user's default profile for the build profile, and when not cross-compiling I copy the host profile across, which brings the `build_type` with it. I have a simple workaround (forcing the `build_type` to Release in the build profile), but I thought that to ensure that ninja was shipped in release mode it was better to do that within the recipe, as it is hinged at in the linked discussion in [conan-io/conan#17667](https://github.com/conan-io/conan/issues/17667).

@memsharded  may be able to provide better context here, but this is something that was done back when Conan didn't support two profiles (build and host), but now that Conan 2 unconditionally assumes two profiles, the recommendation _is_ to have the build profile set to Release.

I think what you are describing as a workaround is actually perfectly valid: you are telling Conan that any dependency in the build context (that is, executable tools required during the build) - must have been built with Release.

tkhyn:
> As far as I'm aware - `cmake-conan` for Conan 2 does _not_ do that https://github.com/conan-io/cmake-conan/blob/develop2/conan_provider.cmake#L666-L667 - it very explicitly relies on the default, autodetected profile regardless.

Hum, no, the build profile is set to the default, so if the default profile is Debug.

> I think there is a misconception here ...

Not really, I should have been clearer sorry, but my underlying assumption is that in the context of conan-center (as we are on the conan-center-index repo), app packages are built in Release.

> But if you are not, and your profile has build_type=Debug and you happen to do --build=missing - then you'd end up with a Debug build of ninja.

I am not sure I follow. At the moment, even when using conan-center, ``--build-missing`` is building ninja in Debug mode if the build profile's `build_type` is set to Debug and that is what I think is not desirable, and is fixed by deleting `settings.build_type` to generate the package ID.


In the end I'm perfectly happy with forcing Release in build profiles as far as my CMake module is concerned but I thought this PR could help resolve an undesirable behaviour. Happy to have it closed if you think it's not really a valid edge case.

jcar87:
> Hum, no, the build profile is set to the default, so if the default profile is Debug it will want a Debug build of ninja.

Typically, the default autodetected profile by Conan will always be release - `cmake-conan` does not use the settings of the current build.

> --build-missing is building ninja in Debug mode if the build profile's build_type is set to Debug 

This is where I'm confused as to why this is considered undesirable -  if the build profile's build_type is set to Debug, it is because the user has made a conscious decisions to make it that way - Conan's own defaults certain don't set it to Debug, so it must be a user choice 

>  and is fixed by deleting settings.build_type to generate the package ID.

If the binary does not already exist - it will be built with Debug in this instance. So the behaviour you are describing of "bring me the Release binary when I'm in Debug" only works if the binary was already explicitly created beforehand - not all of our users retrieve binaries from Conan Center, even if they use the recipes, so we cannot make an assumption that they have created the package with the Release setting.

tkhyn:
> not all of our users retrieve binaries from Conan Center, even if they use the recipes, so we cannot make an assumption that they have created the package with the Release setting.

Again, I reason in the context of conan-center. If people upload Debug binaries of conan-center recipes in their own conan repositories, I would suggest it's their problem!

That said, your reasoning makes a lot of sense, if the build profile explicitly sets `build_type=Debug` then it should probably indeed get Debug binaries.

The reasons I initiated this PR are:

- @memsharded's comment [here](https://github.com/conan-io/conan/issues/17667#issuecomment-2624102597)
- the fact I ran into what felt like an undesirable behaviour myself
- the fact that many app package recipes in conan-center-index do indeed delete `self.info.settings.build_type` in `package_id()`

However, there are also as many app package recipes in conan center that do not delete `build_type`, and in fact the documentation (which I could probably have searched for earlier in the discussion or even before submitting the change) indeed [recommends not deleting `build_type`](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/conanfile_attributes.md#settings) - looks like you wrote this!

Sorry for the inconvenience and thanks for the discussion.

PR #26520, opened 2025-02-03
OpenColorIO: Bump version of OpenEXR to a supported one

irieger:
### Summary
Changes to recipe:  **opencolorio/?**

#### Motivation
Bumping the dependency to OpenEXR 3.3.2 (latest one) as the current one is no longer supported and thus doesn't support libdeflate ranges. However, to be compatible with libtiff when both are required by OpenImageIO, they require a version of OpenEXR with libdeflate support.

#### Details
I have problems to solve the dependency tree of #25672 due to libdeflate so this change should solve the problem. Also I guess it is beneficial to have a supported version of a dependency in the requires.

---

### DIFF
--- a/recipes/opencolorio/all/conanfile.py
+++ b/recipes/opencolorio/all/conanfile.py
@@ -51,7 +51,7 @@ class OpenColorIOConan(ConanFile):
         if Version(self.version) < "2.2.0":
             self.requires("openexr/2.5.7")
         else:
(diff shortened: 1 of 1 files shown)


AbrilRBS (approved):
openexr is one of those libraries that is super split between many versions in recipes, so it'd be hard to unify them all to 1 :/

Issue #20257, opened 2023-10-02
[package] qt/6.x: Qt 6 can not be cross-compiled


petermbauer:
### Description

Since there is a check implemented in `validate()` and setting the `QT_HOST_PATH` CMake variable is not implemented, it is not possible to cross-compile Qt, e.g. for x86_64 on an Apple Silicon machine.

### Package and Environment Details

* Package Name/Version: **qt/6.5.2**
* Operating System+version: **MacOS**
* Compiler+version: **XCode 14**

### Conan profile

[settings]
os=Macos
os_build=Macos
arch=x86_64
arch_build=armv8
compiler=apple-clang
compiler.version=14
compiler.libcxx=libc++
build_type=Release
[options]
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install --build missing qt/6.5.2

### Logs

not relevant

jwillikers:
Aww yes, won't the Qt recipe effectively have to add itself as a `tool_requires` when cross-compiling? The Wayland recipe works like this.

petermbauer:
> Aww yes, won't the Qt recipe effectively have to add itself as a `tool_requires` when cross-compiling? The Wayland recipe works like this.

This would already be in place: https://github.com/conan-io/conan-center-index/blob/master/recipes/qt/6.x.x/conanfile.py#L396-L397
But at least two more changes are needed:
- set `QT_HOST_PATH` via CMake to the installation folder of the `tool_requires` package so the tools like moc are taken from this "Host Qt"
- depending on which kind of cross-building is done, set `QT_FORCE_BUILD_TOOLS=ON` so the tools like moc are also built. Not sure if this can be determined automatically or if this should be implemented as an option.

jwillikers:
Nice, I didn't know that we already added it as a `tool_requires`. That `QT_HOST_PATH` setting is of course needed. With your changes, you'll want to set `QT_HOST_PATH` using the same condition for the `tool_requires`, i.e. `if cross_compiling`.

Since Qt is available as a `tool_requires`, why would `QT_FORCE_BUILD_TOOLS` be necessary?

petermbauer:
> Since Qt is available as a `tool_requires`, why would `QT_FORCE_BUILD_TOOLS` be necessary?

[...]

PR #26529, opened 2025-02-04
qt6: add ability to crossbuild

jcar87:
### Summary
Changes to recipe:  qt/6.x.x

#### Motivation
Provide ability to cross-build recipe

#### Details
* Preserve `Qt6HostInfoConfig.cmake` in the package folder - this is needed by Qt itself when crossbuilding (it does not define targets, just some meta-information about the build)
* Define `QT_HOST_PATH` to point to the self-dependency in the build context 
* In lieu of `Qt6CoreTools` - simply pre-include the target executable information that is already in the dependency in the build context - that will define things like `Qt6::moc` and will use those instead
* Include executables part of Qtbase tools that were not defined in the targets file - these are needed when crossbuilding qt
* Delete logic around `android_sdk` (is not an option)
* Minimally adapt the test_package to get it to work when cross-building (it's subject to further changes in the future)

### Tested on
* macOS arm64 -> x86_64 (should be on CI too)
* Linux aarch64 -> x86_64

Android may require additional work, as it requires an SDK setup (and not just the NDK). Not invalidating this, as it may be possible to inject the necessary variables via `tools.cmake.cmaketoolchain:extra_variables` to get a successful build

Close https://github.com/conan-io/conan-center-index/issues/20257
Close https://github.com/conan-io/conan-center-index/issues/8120

---

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -264,9 +264,6 @@ class QtConan(ConanFile):
             self.output.debug(f"qt6 option: {option}")
 
     def validate_build(self):
-        if cross_building(self):
-            raise ConanInvalidConfiguration("cross compiling qt 6 is not yet supported. Contributions are welcome")
-
         check_min_cppstd(self, 17)
 
     def validate(self):
@@ -624,7 +621,12 @@ class QtConan(ConanFile):
         if self.options.cross_compile:
             tc.variables["QT_QMAKE_DEVICE_OPTIONS"] = f"CROSS_COMPILE={self.options.cross_compile}"
         if cross_building(self):
-            tc.variables["QT_HOST_PATH"] = self.dependencies.direct_build["qt"].package_folder
+            # Mainly to locate Qt6HostInfoConfig.cmake
+            tc.cache_variables["QT_HOST_PATH"] = self.dependencies.direct_build["qt"].package_folder
+            # Stand-in for Qt6CoreTools - which is loaded for the executable targets
+            tc.cache_variables["CMAKE_PROJECT_Qt_INCLUDE"] = os.path.join(self.dependencies.direct_build["qt"].package_folder, self._cmake_executables_file)
+            # Ensure tools for host are always built
+            tc.cache_variables["QT_FORCE_BUILD_TOOLS"] = True
 
         tc.variables["FEATURE_pkg_config"] = "ON"
         if self.settings.compiler == "gcc" and self.settings.build_type == "Debug" and not self.options.shared:
@@ -663,8 +665,6 @@ class QtConan(ConanFile):
                     self.info.settings.compiler.runtime = "MT/MTd"
             elif self.info.settings.compiler == "msvc":
                 self.info.settings.compiler.runtime_type = "Release/Debug"
-        if self.info.settings.os == "Android":
-            del self.info.options.android_sdk
 
     def source(self):
         destination = self.source_folder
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):

franramirez688 (approved):

jcar87:
[qt-crossbuild-log.txt](https://github.com/user-attachments/files/18674429/qt-crossbuild-log.txt)

Log for Linux x86_64 -> armv8 build

valgur:
Also - how did you handle the `xorg/system` dependency for cross-compilation? I assume you provided it somehow?

jcar87:
> Did you also manage to cross-compile a recipe consuming the cross-compiled Qt? I think it might have required some additional steps to get it working.

We are able to crossbuild `qt-advanced-docking-system` with the changes from this PR

> Edit: nevermind, I now saw that you cross-compiled for x86_64, which sidesteps this issue.

We've done both, cross armv8->x86_64 and x86_64 -> armv8 - in both cases we are using Desktop OpenGL and x11, as provided by Ubuntu. This docker image can be used as an example: https://github.com/jcar87/cci-crossbuild-example - I reckon something like this would work:

```
# build Qt for the local architecture, incl. all dependencies
conan create recipes/qt/6.x.x --version=6.7.3 --build=missing -c tools.system.package_manager:mode=install

# build Qt for the target architecture, incl. all dependencies,
# --build=missing --build="missing:qt/*" is needed to avoid re-building the build context qt
conan create recipes/qt/6.x.x --version=6.7.3 --build=missing --build="missing:qt/*" -pr arm64  -c tools.system.package_manager:mode=install
```

For system dependencies when crossbuilding, we roughly see the following cases:
- When working on a multiarch debian, it is possible to rely on Conan to install the packages for the "foreign" architecture, provided dpkg and the apt sources are correctly configured to support this (the dockefile linked above is an example of how to do this for a recent ubuntu). When using the Debian/Ubuntu provided cross-compiler, the sysroot is implicitly `/`  (does not require an additional flag), and all "system" paths are correctly configured for the target architecture - however it's important to use a variant of pkg-config [that also is ](https://github.com/jcar87/cci-crossbuild-example/blob/main/arm64_profile#L12)
- when working with a sysroot (`--sysroot` is passed to the compiler, e.g. with `tools.build:sysroot`) - this will depend on what the actual target platform is and what functionality the user requires. If the user requires xorg / desktop opengl  - this should be pre-installed in the sysroot. We would expect this to be the case when targeting something like Raspberry Pi or Jetson, that the `.pc` files are also in the sysroot. This will require a properly configured `pkg-config`, and the `tools.system.package_manager:mode` in the host context to bypass the build machine (`report` or `report-installed` should do it )
- if an implicit sysroot (e.g. from an SDK), or an explicit sysroot (`--sysroot`) already provide EGL/x11 or any other library, but does not provide `.pc` files,  this is not much different than "system_libs": everything is already in the default search path and no additional flags are required other than the library names - however mileage may vary here if tools expect a .pc file and can't fall back on defaults. When they can fall back, the `[platform_requires]` here may be useful to tell Conan that `*/system` requirements already installed and the logic in the recipes can be bypassed - but that feature is still experimental.  But again, I'd caution users to not pay for what they don't use - and consider that recipes in Conan Center index likely enable far more optional dependencies than they actually require.

Cyriuz:
Thanks for this! I'm looking into using this but it causes a lot of rebuilding of Qt and I was wondering if the tool_requires should inherit the settings from the parent to avoid this? Usually you would already have the native arch Qt in the cache and, for example, I use the translation tools, so I need them enabled too for the tool require.

For example I was thinking something like this:
```
if cross_building(self):
    self.tool_requires(RecipeReference(self.name, self.version, self.user, self.channel).repr_notime(), options=self.options)
```

recipes/qt/6.x.x/conanfile.py
```diff
@@ -624,7 +621,12 @@ def generate(self):
         if self.options.cross_compile:
             tc.variables["QT_QMAKE_DEVICE_OPTIONS"] = f"CROSS_COMPILE={self.options.cross_compile}"
         if cross_building(self):
-            tc.variables["QT_HOST_PATH"] = self.dependencies.direct_build["qt"].package_folder
+            # Mainly to locate Qt6HostInfoConfig.cmake
+            tc.cache_variables["QT_HOST_PATH"] = self.dependencies.direct_build["qt"].package_folder
+            # Stand-in for Qt6CoreTools - which is loaded for the executable targets
+            tc.cache_variables["CMAKE_PROJECT_Qt_INCLUDE"] = os.path.join(self.dependencies.direct_build["qt"].package_folder, self._cmake_executables_file)
+            # Ensure tools for host are always built
```
Cyriuz, line 628: What's the reason for this? It makes the build not work on iOS since the tools doesn't build for that platform.

PR #26531, opened 2025-02-05
botan: add version 3.8.1

reneme:
### Summary
Updates recipe:  **botan/3.7.0**

#### Motivation

New version released: [changelog](https://github.com/randombit/botan/blob/master/news.rst#version-370-2025-02-04)

---


reneme:
Mhh, MSVC in your CI seems to encounter an ICE. 😨

```
botan_all.cpp(3069): note: see reference to alias template instantiation 'std::invoke_result_t<FnT,BlockT...>' being compiled
INTERNAL COMPILER ERROR in 'C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.36.32532\bin\HostX64\x64\cl.EXE'
    Please choose the Technical Support command on the Visual C++
    Help menu, or open the Technical Support help file for more information
NMAKE : fatal error U1077: 'cl /DBOTAN_DLL=__declspec(dllexport)  /Zc:preprocessor /std:c++20 /EHs /GR /D_WIN32_WINNT=0x0600 /MD /bigobj   /O2 /Oi /Zc:throwingNew /W4 /wd4251 /wd4275 /wd5072 -DBOTAN_IS_BEING_BUILT  /I build/include/public /I build/include/internal  /external:W0 /external:I build/include/external /nologo /c botan_all.cpp /Fobuild/obj/lib/botan_all.obj' : return code '0xc0000005'
Stop.
```

Judging from the failing template initialization (`std::invoke_result_t<FnT,BlockT...>`) the code in question was likely introduced with this pull request: https://github.com/randombit/botan/pull/3883.

A suitable workaround could be to disable the relevant botan module for affected MSVC versions by adding `--disable-modules=bitvector`. This will transitively disable the newly introduced implementation of the "Classic McEliece" for such builds, but everything else should work as expected.

Feedback would be welcome here. :)

RazielXYZ:
3.7.1 was also released a few weeks ago.

Local latest msvc does not seem to suffer from this ICE with bitvector. I don't think there's any good way for conan to figure out the msvc build with enough accuracy at configuration time, considering it only takes the toolchain version (193 in this case), and forcing bitvector to disabled for msvc in general just because of the CI being behind doesn't seem like a good solution.

If I recall correctly there are ways to skip certain configs on CI, we might have to do that unless there's a better alternative.

Phatcat:
Please bump to 3.8.1 now it's out.

RazielXYZ:
Looks like the same ICE as before. At this point MSVC 14.36 (which is what CC seems to be trying to use) is very old so I'm not sure what the approach would be. I would hope there are plans to use a newer one, but if not, skipping CI for it would still be the best option I think

recipes/botan/all/conanfile.py
```diff
@@ -312,6 +312,10 @@ def _configure_cmd(self):
         botan_extra_cxx_flags = []
         build_flags = []
 
+        if Version(self.version) >= '3.7' and is_msvc(self) and Version(self.settings.compiler.version) < '194':
+            # https://github.com/conan-io/conan-center-index/pull/26531#issuecomment-2636504872
+            build_flags.append('--disable-modules=bitvector')
```
reneme, line 317: Note that the bitvector is mostly a helper structure for the post-quantum algorithm "Classic McEliece". By disabling bitvector you also disable this algorithm.

```suggestion
        if Version(self.version) >= '3.7' and is_msvc(self) and Version(self.settings.compiler.version) < '194':
            # https://github.com/conan-io/conan-center-index/pull/26531#issuecomment-2636504872
            # Note that this disables support for Classic McEliece
            build_flags.append('--disable-modules=bitvector')
```

reneme (commented):
Thanks for looking into a workaround for this. Much appreciated.

uilianries:
> Thanks for looking into a workaround for this. Much appreciated.

@reneme Thank you for suggesting. We did a few tests on Windows to check how to avoid that ICE: We found that when using MSVC >=19.39 that does not happen and we did not need the workaround as well. For MSVC 19.30, the workaround of bitvector works, but fails in another point. 

Conan has the update version for MSVC as a setting, but it is not configured by default, so we preferred to keep 193 with a workaround instead of only filtering by "< 19.39". As the CI is running 19.36, it works when disabling bitvector.

uilianries:
@reneme After talking one more time about the current error with the team, occurs that will be better only skip that configuration from the CI, instead of the workaround. The fact it will private users that are running 19.39 of building this recipe.

We will be updating the CI configuration to skip that specific scenario. Sorry for the confusion.

recipes/botan/all/conanfile.py
```diff
@@ -312,6 +312,11 @@ def _configure_cmd(self):
         botan_extra_cxx_flags = []
         build_flags = []
 
+        if Version(self.version) >= '3.7' and is_msvc(self) and Version(self.settings.compiler.version) < '194':
+            # https://github.com/conan-io/conan-center-index/pull/26531#issuecomment-2636504872
+            # Note that this disables support for Classic McEliece
+            build_flags.append('--disable-modules=bitvector')
+
```
uilianries, line 319:
```suggestion
```

RazielXYZ:
@uilianries If there's any way for us to help validate, let us know. I can validate/test latest MSVC here as well as multiple linux configurations.

uilianries:
All looking good, I just updated the logs: https://github.com/conan-io/conan-center-index/pull/26531#issuecomment-2895068491

uilianries (approved):
LGTM. All documented logs are passing when using newer compiler version with C++20. 

Unfortunately we will not have a pre-built package in Conan Center, but the recipe should be available. 

Thank you @reneme @RazielXYZ for all your support!

ErniGH (approved):
LGTM

PR #26533, opened 2025-02-05
sdl_image: Remove support for versions not present in config

AbrilRBS:
Follow-up to https://github.com/conan-io/conan-center-index/pull/26527, I split it into two to make the bugfix faster to merge, but these also add missing features and leave the recipe in a nice place

Note that I've been able to compile with every option set true, except for libavif which fails to compile


AbrilRBS:
Update: The newer libavif version does compile for me, I guess it'd be a matter of syncing all of the 4 usages in CCI to the newer version at some point

jcar87 (approved):

PR #26549, opened 2025-02-06
Fixed an issue where the wrong ZLIB library could be specified during OpenSSL configuration.

### LINKED ISSUE #18011: openssl/3.1.0 openssl/3.1.1 shared on windows tries to call LoadLibrary with zdll.lib
### Description

openssl/3.1.0 openssl/3.1.1 depending on shared version of zlib/1.2.13 on windows tries to call LoadLibrary with zdll.lib as the parameter resulting in this error during initialization.

libcurl_test.exe - Bad Image
---------------------------
C:\conan_testing\.conan\data\zlib\1.2.13\_\_\package\e9a552ebe8f994398de9ceee972f0ad207df0658\lib\zdll.lib is either not designed to run on Windows or it contains an error. Try installing the program again using the original installation media or contact your system administrator or the software vendor for support. Error status 0xc000012f. 

Making this change to the recipe, seems to fix the issue:

diff --git a/recipes/openssl/3.x.x/conanfile.py b/recipes/openssl/3.x.x/conanfile.py
index 59da5e3bc..940300c30 100644
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -391,10 +391,9 @@ class OpenSSLConan(ConanFile):
             else:
                 args.append("zlib")

-            args.extend([
-                '--with-zlib-include="%s"' % include_path,
-                '--with-zlib-lib="%s"' % lib_path
-            ])
+            args.append('--with-zlib-include="%s"' % include_path)
+            if self._settings_build.os != "Windows" or not self.dependencies["zlib"].options.shared:
+                args.append('--with-zlib-lib="%s"' % lib_path)

         for option_name in self.default_options.keys():
             if self.options.get_safe(option_name, False) and option_name not in ("shared", "fPIC", "openssldir", "capieng_dialog", "enable_capieng", "zlib", "no_fips", "no_md2"):

* Package Name/Version: **openssl/3.1.1, openssl/3.1.0**
* Compiler+version: **Microsoft Visual Studio 2022**
* Conan version: **conan 1.59.0**
* Python version: **Python 3.11.2**

### Conan profile

[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=17
os=Windows
os_build=Windows
[options]
libcurl:shared=True
openssl:shared=True
zlib:shared=True
[conf]
[build_requires]
[env]

### Steps to reproduce

extract the attached libcurl_test.zip, from the extracted directory run:
conan install -pr libcurl_test_rel -r conancenter conanfile.py --build=missing
open libcurl_test.sln with Visual Studio and build the release version of the project
run libcurl_test.exe notice the popup error

[libcurl_test.zip](https://github.com/conan-io/conan-center-index/files/11820267/libcurl_test.zip)

### Logs

<details><summary>log of the conan install command</summary>

```
PS C:\misc\c++\libcurl_test\libcurl_test> conan install -pr libcurl_test_rel -r conancenter conanfile.py --build=missing
Configuration:
[settings]
arch=x86_64
arch_build=x86_64
build_type=Release
compiler=Visual Studio
compiler.runtime=MD
compiler.version=17
os=Windows
os_build=Windows
[options]
libcurl:shared=True
openssl:shared=True
zlib:shared=True
[build_requires]
[env]

WARN: libcurl/7.88.1: requirement openssl/[>=1.1 <4] overridden by your conanfile to openssl/3.1.1
conanfile.py: Installing package
Requirements
    libcurl/7.88.1 from 'artifactory' - Cache
    openssl/3.1.1 from 'artifactory' - Cache
    zlib/1.2.13 from 'artifactory' - Cache
Packages
    libcurl/7.88.1:932f591ffc9643ea53d8b373a1b03b18d5aba33a - Cache
    openssl/3.1.1:e5e306e07768a1034b6ccda478904c1f0f1081a8 - Cache
    zlib/1.2.13:e9a552ebe8f994398de9ceee972f0ad207df0658 - Cache

Installing (downloading, building) binaries...
zlib/1.2.13: Already installed!
openssl/3.1.1: Already installed!
libcurl/7.88.1: Already installed!
conanfile.py: Generator visual_studio_multi created conanbuildinfo_multi.props
conanfile.py: Generator visual_studio_multi created conanbuildinfo_release_x64_v143.props
conanfile.py: Generator txt created conanbuildinfo.txt
conanfile.py: Aggregating env generators
conanfile.py: Generated conaninfo.txt
conanfile.py: Generated graphinfo
conanfile.py: ==== imports ====
conanfile.py: cwd=C:\misc\c++\libcurl_test\libcurl_test
conanfile.py: arch=x86_64
conanfile.py: build_type=Release [abc]
conanfile.py: package_folder=None
conanfile.py: Copying dll files to x64\Release
conanfile.py imports(): Copied 4 '.dll' files: libcurl.dll, zlib1.dll, libssl-3-x64.dll, libcrypto-3-x64.dll
```

</details>

tylerbrawl:
### Summary
Changes to recipe:  **openssl/3.x.x**

#### Motivation
This PR was created pursuant to the resolution of #18011, which is reproducible in an application developed by my organization. Essentially, the official Conan recipe for OpenSSL could result in a misconfiguration where the library attempts to load a static library (`zdll.lib`) as a DLL.

#### Details
In #18011, it was demonstrated that the official Conan recipe for OpenSSL could result in a build misconfiguration, leading to a "Bad Image" message box appearing at runtime. This was ultimately the result of attempting to load a static library as a DLL. I confirmed this by placing a breakpoint on the following Win32 APIs:

  * `LoadLibraryA()`
  * `LoadLibraryW()`

These are the standard Win32 APIs for injecting a DLL into the address space of the current process at runtime. (For more information, refer to the MSDN at
https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya.) Checking the call stack in a debugger just before the message box would appear revealed that an attempt was being made from within a module by the name of `libcrypto-3-x64.dll` to load a static library (`zdll.lib`) as a DLL. Such an operation is inherently malformed, which is why the "Bad Image" message box would appear.

As it turns out, the issue was caused by an error in the `conanfile.py` for OpenSSL. When ZLIB was built as a shared library, the `OpenSSLConan::_configure_args()` function would historically append the following command line arguments to the list used to configure OpenSSL:

  * `zlib-dynamic`
  * `--with-zlib-lib=[Path to Static Import Library]`

Here, `[Path to Static Import Library]` represents the path to the static import library which is generated alongside ZLIB's DLL in a shared build of the project. When using the official Conan recipe for ZLIB, this file is called `zdll.lib`, which - not coincidentally - is the same library as the one which `libcrypto-3-x64.dll` was erroneously attempting to load as a DLL. Still, it is worth asking why this behavior occurred in the first place. After all, even when a shared library is built, the linker may still require access to the corresponding static import library `zdll.lib`.

Well, the issue appears to have been caused by a misunderstanding of the `--with-zlib-lib` command line argument. To be more precise, it is important to understand how the `zlib-dynamic` command line argument affects how it is interpreted. Inspecting the `Configure` file located at the root of the OpenSSL source code repository reveals the following crucial information contained in a comment:

```
[no-]zlib     [don't] compile support for zlib compression.
zlib-dynamic  Like "zlib", but the zlib library is expected to be a shared
              library and will be loaded in run-time by the OpenSSL library.
```

The implication of the above is that the `--with-zlib-lib` command line argument is expected to be used to specify the ZLIB DLL (typically `zlib1.dll`) when the `zlib-dynamic` command line argument is also specified, rather than the corresponding static import library (`zdll.lib`).

This commit addresses that issue by modifying the `OpenSSLConan::_configure_args()` function so that the correct library is specified on Windows for the `--with-zlib-lib` command line argument. As a result, `libcrypto-3-x64.dll` should now attempt to load the correct DLL for ZLIB at runtime. At a minimum, this prevents the aforementioned "Bad Image" message box from appearing. One might also speculate what other subtle OpenSSL bugs may have been fixed by this change.

NOTE: The changes in this commit do _not_ modify the library specified via the `--with-zlib-lib` command line argument on Linux. A comment within `OpenSSLConan::_configure_args()` suggests that the "right file" will be specified, but it might be worth verifying this claim in response to the findings described above.

This commit fixes #18011.

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -382,44 +382,27 @@ class OpenSSLConan(ConanFile):
             include_path = self._adjust_path(zlib_cpp_info.includedirs[0])
             is_shared_zlib = self.dependencies["zlib"].options.shared
 
+
+            # the --with-zlib-lib flag takes a different value depending on platform and if ZLIB is shared
+            # From https://github.com/openssl/openssl/blob/openssl-3.4.1/INSTALL.md#with-zlib-lib
+            # On Unix: the directory where the zlib library is (for -L flag)
+            # On Windows with static zlib: the path to the static library to link (assumed)
+            # On Windows with shared zlib: the leaf name of the dll (its loaded with LoadLibrary)
             if self._use_nmake:
-                # The libraries defined in zlib_cpp_info.libs are *static* libraries. On Windows, we expect this list to be
-                # populated with one of the following:
-                #
-                #   * The static library zlib.lib if ZLIB was built as a static library.
-                #   * The import library zdll.lib if ZLIB was built as a shared library.
-                #
-                # (NOTE: These library names are inconsistent with those used by ZLIB's original CMakeLists.txt file. We
-                # assume that our patched CMakeLists.txt is being used, instead.)
-                #
-                # There is an important distinction to make here. Specifically, when the "zlib-dynamic" command line argument
-                # is specified during the configuration of OpenSSL (as would be the case when is_shared_zlib == True), the
-                # library specified via the "--with-zlib-lib" command line argument is expected to be a *DLL*, rather than a
-                # static library.
-                #
-                # This implies that we cannot always rely on the library specified as the first entry in zlib_cpp_info.libs,
-                # since said library will actually be the import library in the event that is_shared_zlib == True. In that case,
-                # we ought to instead specify the generated DLL zlib1.dll.
-
-                if is_shared_zlib:
-                    # NOTE: We specify the generated DLL as a relative path, rather than an absolute path. Since the DLL is loaded
-                    # at runtime by OpenSSL, we cannot possibly predict what the "best" relative path would be.
-                    zlib_path = "zlib1.dll"
-                else:
-                    zlib_parent_path = zlib_cpp_info.libdirs[0]
-                    zlib_path = f"{zlib_parent_path}/zlib.lib"
-
-                lib_path = self._adjust_path(zlib_path)
+                # notes: consider where this should be "if on windows"
+                #        zlib1 is assumed to be the name of the zlib1.dll for all windows configurations
+                lib_path = self._adjust_path(os.path.join(zlib_cpp_info.libdirs[0], f"{zlib_cpp_info.libs[0]}.lib"))
+                zlib_lib_flag = "zlib1" if is_shared_zlib else lib_path
             else:
                 # Just path, GNU like compilers will find the right file
-                lib_path = self._adjust_path(zlib_cpp_info.libdirs[0])
+                zlib_lib_flag = self._adjust_path(zlib_cpp_info.libdirs[0])
 
             zlib_configure_arg = "zlib-dynamic" if is_shared_zlib else "zlib"
             args.append(zlib_configure_arg)
 
             args.extend([
                 f'--with-zlib-include="{include_path}"',
-                f'--with-zlib-lib="{lib_path}"',
+                f'--with-zlib-lib="{zlib_lib_flag}"',
             ])
 
         for option_name in self.default_options.keys():
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -380,16 +380,42 @@ class OpenSSLConan(ConanFile):
         if not self.options.no_zlib:
             zlib_cpp_info = self.dependencies["zlib"].cpp_info.aggregated_components()
             include_path = self._adjust_path(zlib_cpp_info.includedirs[0])
+            is_shared_zlib = self.dependencies["zlib"].options.shared
+
             if self._use_nmake:
-                lib_path = self._adjust_path(os.path.join(zlib_cpp_info.libdirs[0], f"{zlib_cpp_info.libs[0]}.lib"))
+                # The libraries defined in zlib_cpp_info.libs are *static* libraries. On Windows, we expect this list to be
+                # populated with one of the following:
+                #
+                #   * The static library zlib.lib if ZLIB was built as a static library.
+                #   * The import library zdll.lib if ZLIB was built as a shared library.
+                #
+                # (NOTE: These library names are inconsistent with those used by ZLIB's original CMakeLists.txt file. We
+                # assume that our patched CMakeLists.txt is being used, instead.)
+                #
+                # There is an important distinction to make here. Specifically, when the "zlib-dynamic" command line argument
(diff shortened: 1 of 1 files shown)


recipes/openssl/3.x.x/conanfile.py
```diff
@@ -391,16 +391,42 @@ def _configure_args(self):
...
+                # since said library will actually be the import library in the event that is_shared_zlib == True. In that case,
+                # we ought to instead specify the generated DLL zlib1.dll.
+
+                if is_shared_zlib:
+                    # NOTE: We specify the generated DLL as a relative path, rather than an absolute path. Since the DLL is loaded
+                    # at runtime by OpenSSL, we cannot possibly predict what the "best" relative path would be.
+                    zlib_path = "zlib1.dll"
+                else:
+                    zlib_parent_path = zlib_cpp_info.libdirs[0]
+                    zlib_path = f"{zlib_parent_path}/zlib.lib"
```
tylerbrawl, line 421: The use of the hard-coded string `"zlib.lib"` here could alternatively be replaced with `f"{zlib_cpp_info.libs[0]}.lib"`, but it seems a little odd to do so when we need to use a hard-coded string for `"zlib1.dll"`, anyways. (At least, I am unaware of an equivalent to `cpp_info.libs` for DLLs. Let me know if I am missing something here!)

Did you have a preference between the two? Is there a precedent for this type of issue?

jcar87: hi @tylerbrawl - thanks so much for the thorough explanation ! You are exactly right.

We have a feature coming where the zlib recipe can specify the location/filename of the DLL (in addition to that of the import library or link library). I think the attribute is already in the most recent conan versions, and could be reflected in the zlib recipe right away. You are also right - it is unfortunate by the zlib recipe does deviate from the canonical filenames in some scenarios.

jcar87:
Thanks @tylerbrawl - we're still reviewing this.

A bit puzzled as the `test_package` for OpenSSL should technically be catching this issue when `*:shared=True` (both openssl and zlib are shared) - which is the case in our CI. Will continue investigating (The issue is indeed confirmed and the fix is mostly correct!)

jcar87 (approved):

jcar87:
Thanks @tylerbrawl - have confirmed the bug with the repro steps from the linked issue, and have touched up the proposed changes to avoid more nesting (as well as linking as the relevant openssl build documentation). Thanks so much for the detailed PR and apologies about the delay

Issue #26551, opened 2025-02-07
[package] qt/6.7.3: host libb2 may be used


fnadeau:
### Description

If host has libb2 installed, build will go through but example will fail to link due to missing system lib in link command.

Note: shared build works but makes them less shareable since libb2 is not declared as dependency.

### Package and Environment Details

* Package Name/Version: **qt/6.7.3**
* Operating System+version: **Arch**
* Compiler+version: **GCC 14.2.1 20240910**
* Docker image: **archlinux/base-devel**
* Conan version: **conan 2.11.0**
* Python version: **Python 3.13.1**

### Conan profile

Host profile:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux

Build profile:
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux

### Steps to reproduce

docker run -it --rm archlinux/base-devel bash
pacman -Syu python-virtualenv libb2 git cmake ninja libglvnd xkeyboard-config
cd
python -m venv conan
source conan/bin/activate
pip install conan
git clone https://github.com/conan-io/conan-center-index.git
conan profile detect
conan create conan-center-index/recipes/qt/6.x.x/conanfile.py --version 6.7.3 --build=missing -c tools.system.package_manager:mode=install
conan create conan-center-index/recipes/qt/6.x.x/conanfile.py --version 6.7.3 --build=missing -c tools.system.package_manager:mode=install -o='qt/*:shared=True'

### Logs

<details><summary>Fail when shared=False</summary>

```

qt/6.7.3 (test package): Running CMake.build()
qt/6.7.3 (test package): RUN: cmake --build "/root/conan-center-index/recipes/qt/6.x.x/test_package/build/gcc-14-x86_64-gnu17-release" -- -j8
[  0%] Built target test_package_autogen_timestamp_deps
[ 14%] Automatic MOC for target test_package
[ 14%] Built target test_package_autogen
[ 28%] Automatic RCC for example.qrc
[ 57%] Building CXX object CMakeFiles/test_package.dir/test_package_autogen/mocs_compilation.cpp.o
[ 71%] Building CXX object CMakeFiles/test_package.dir/test_package_autogen/EWIEGA46WW/qrc_example.cpp.o
[...]

PR #26552, opened 2025-02-07
qt6: Add option for libb2

fnadeau:
### Summary
Changes to recipe:  **qt/6.7.3**

#### Motivation
If libb2 is present on host and lib is build as static, link will fail. See issue #26551

#### Details
Pass libb2=no to Qt make rather than relying on auto detection.

Would require more patch to be able to use libb2 from conan. 
- In qtbase/src/corelib/CMakeLists.txt, Qt uses Libb2::Libb2 rather than libb2::libb2 in conan
- Need to remove qtbase/cmake/FindLibb2.cmake since conan provide it own find

But I wasn't able to make it work. Qt will fail at install because libb2 is not declared as dependency (from the Qt build, not conan).

Anyhow, with this patch, at least system libb2 is not used.
---

### DIFF
--- a/recipes/qt/6.x.x/conanfile.py
+++ b/recipes/qt/6.x.x/conanfile.py
@@ -50,6 +50,7 @@ class QtConan(ConanFile):
         "with_fontconfig": [True, False],
         "with_icu": [True, False],
         "with_harfbuzz": [True, False],
+        "with_libb2": [True, False],
         "with_libjpeg": ["libjpeg", "libjpeg-turbo", False],
         "with_libpng": [True, False],
         "with_sqlite3": [True, False],
@@ -95,6 +96,7 @@ class QtConan(ConanFile):
         "with_fontconfig": True,
         "with_icu": True,
         "with_harfbuzz": True,
+        "with_libb2": False,
         "with_libjpeg": False,
         "with_libpng": True,
         "with_sqlite3": True,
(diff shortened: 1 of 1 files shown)


jcar87:
> In qtbase/src/corelib/CMakeLists.txt, Qt uses Libb2::Libb2 rather than libb2::libb2 in conan
> Need to remove qtbase/cmake/FindLibb2.cmake since conan provide it own find

Conan/the recipes can tweak the name of the cmake target exposed to the qt build system, and there's always a fair chance that FindLibb2.cmake would work correectly (depending on whether it has dependencies itself).

While this PR is under review, there's a chance that adding the following in your profile could achieve the same result without having to modify the recipe - by skipping Qt's checks that otherwise result in the system variant being pulled
```
qt/6*:tools.cmake.cmaketoolchain:extra_variables={'CMAKE_DISABLE_FIND_PACKAGE_Libb2':'ON'}
````

valgur:
Yeah, something like this should work better: https://github.com/valgur/conan-center-index/commit/c3c52b704ecdb2bc528f45046b7a592c1442e03f

fnadeau:
Closing:
- building with ```-c:a qt/6*:tools.cmake.cmaketoolchain:extra_variables="{'CMAKE_DISABLE_FIND_PACKAGE_Libb2':'ON'}"``` does the trick
- proposed solution by @valgur has it fully covered rather than just a pach  (pun intented)

PR #26567, opened 2025-02-10
box2d/3.0.0: fix build issue with clang-19 on MacOS

elvisdukaj:
### Summary
Changes to recipe:  **box2d/3.0.0**

#### Motivation

box2d fails to build on MacOS with clang-19, because box2d builds traits compiler warning as error and clang-19 has stricter rules than the installed apple-clang 16.

I've created an [issue](https://github.com/erincatto/box2d/issues/882) on the upstream project. 

#### Details

clang-19 fails the build with the following error, for more details, see the issue above.

```
contact_solver.c:824:28: error: AVX vector return of type 'simde__m256' (vector of 8 'simde_float32' values) without 'avx' enabled changes the ABI [-Werror,-Wpsabi]
  824 |                 b2FloatW tangentY = sub( simde_mm256_setzero_ps(), c->normal.X );
      |                                          ^
```

The fix is to disable the warnign as errors.

---

### DIFF
--- a/recipes/box2d/all/patches/3.0.0-0003-disable-warning-as-errors.patch
+++ b/recipes/box2d/all/patches/3.0.0-0003-disable-warning-as-errors.patch
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d6590d7..365f95e 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -46,7 +46,7 @@ endif()
+ 
+ # C++17 needed for imgui
+ set(CMAKE_CXX_STANDARD 17)
+-set(CMAKE_COMPILE_WARNING_AS_ERROR ON)
++set(CMAKE_COMPILE_WARNING_AS_ERROR OFF)
+ 
+ # set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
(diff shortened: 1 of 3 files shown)


jcar87 (approved):
Thanks!

recipes/box2d/all/patches/3.0.0-0003-disable-warning-as-errors.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d6590d7..365f95e 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -46,7 +46,7 @@ endif()
+ 
+ # C++17 needed for imgui
+ set(CMAKE_CXX_STANDARD 17)
+-set(CMAKE_COMPILE_WARNING_AS_ERROR ON)
++set(CMAKE_COMPILE_WARNING_AS_ERROR OFF)
```
jcar87, line 10: I would probably suggest _not_ giving it a value, but rather commenting it out - this way users can enable warnings as errors externally (for those users who require it) - rather than have cmake disable it

elvisdukaj: Done!

jcar87: Awesome, thanks!!

jcar87 (approved):

PR #26572, opened 2025-02-10
glaze: add version 4.4.3

toge:
### Summary
Changes to recipe:  **glaze/4.4.2**

#### Motivation
- Add TRACE Option
- AVX2 is used by default, also disable AVX2 flag is added.

#### Details
https://github.com/stephenberry/glaze/compare/v4.3.1...v4.4.2

---

### DIFF
--- a/recipes/glaze/all/conandata.yml
+++ b/recipes/glaze/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "4.4.2":
-    url: "https://github.com/stephenberry/glaze/archive/v4.4.2.tar.gz"
-    sha256: "5f9c8efe35491f90755ef7a9d392ddd2ac395fd2e005e3ca61b5daf54ebfc9de"
+  "4.4.3":
+    url: "https://github.com/stephenberry/glaze/archive/v4.4.3.tar.gz"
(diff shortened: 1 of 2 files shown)


RazielXYZ:
@toge 4.4.2 is out as well, might as well change this to that

toge:
@RazielXYZ 
Thanks!
I updated it.
For 48 hour rule, I will update to 4.4.3 the day after tomorrow.

RazielXYZ:
Of course 4.4.3 would release right after we bump this up, as is tradition 😆 

Anyway, I would very much appreciate if we can get one of the 4.4.x versions on CC ASAP, as I am having a very nasty bug in 4.3.1 that has been fixed in 4.4.2 (probably earlier too, but that's what I tested)

toge:
@RazielXYZ 
Only what we can do is waiting for.

jcar87:
Thanks @toge!

> AVX2 is used by default, also disable AVX2 flag is added.
Screenshot 2025-02-28 at 09 24 11

Thanks for pointing this out - don't think we have to reflect anything in the recipe, as it is a header-only. To double check I built and created the targets from upstream, and there are no flags propagated to the consumers - the rest is handled directly in the headers with preprocessor macros

jcar87 (approved):

PR #26582, opened 2025-02-12
swig: added version 4.3.0

tomas-krupa:
### Summary
Changes to recipe:  **swig/4.3.0**

#### Motivation
update to the new version of the tool

#### Details
just straight update of version

---

### DIFF
--- a/recipes/swig/all/conandata.yml
+++ b/recipes/swig/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "4.3.0":
+    url: "https://github.com/swig/swig/archive/refs/tags/v4.3.0.tar.gz"
+    sha256: "f2136da1137a20dfcec795fe0d17ca1a2465d28e3b307f122526629b6b2f2294"
   "4.2.1":
     url: "https://github.com/swig/swig/archive/refs/tags/v4.2.1.tar.gz"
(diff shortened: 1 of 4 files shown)


uilianries (commented):
@tomas-krupa Thank you for your PR! 

The changes between the versions are looking good, it seems be the upstream improved msys support too: https://github.com/swig/swig/compare/v4.2.1...v4.3.0#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810

The patch all/patches/0001-4.3.0-swig-linux-library-path.patch could be removed. It was added in the very first revision in CCI, but the upstream has own opinion of not using it: https://github.com/swig/swig/pull/2509

Plus, the environment variable `SWIG_LIB` can be used instead, to point where the library is installed. The recipe has `self.buildenv_info.define_path("SWIG_LIB"` configured already, so for who is consuming this package, it should not be a problem.

valgur:
Please also consider https://github.com/conan-io/conan-center-index/pull/25132 if you're already  working on SWIG.

uilianries (approved):
LGTM.

The patches are the very same as before in the CCI.

I'll open a separated issue to discuss about patching problem.

Thank you for your PR!

Issue #26512, opened 2025-02-02
[bug] Version resolution seems broken: Version range is ignored


irieger:
### Describe the bug

Environment: Linux (Arch), gcc 14.2.1, conan 2.12.1. Profile (was auto-detected, clean home via `CONAN_HOME` to start on a clean slate):
```
[settings]
arch=x86_64
build_type=Release
compiler=gcc
compiler.cppstd=gnu20
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux
```

Build command (ffmpeg I often deactivate as it increases the build time massively, doesn't matter for this problem):
```
conan create 3.x --version=3.0.3.1 --build=missing -o:a '&:with_ffmpeg=False' --update
```
In the `recipes/openimageio` folder of my branch for OpenImageIO 3.0: https://github.com/conan-io/conan-center-index/pull/25672

I just also tried it with locally setting `self.requires("libdeflate/1.22", override=True)` into the requires which allows to solve the graph.

**The problem**
```
ERROR: Version conflict: Conflict between libdeflate/1.19 and libdeflate/1.22 in the graph.
Conflict originates from openexr/3.2.3

Run 'conan graph info ... --format=html > graph.html' and open 'graph.html' to inspect the conflict graphically.
```

The graph shows openexr depending on `libdeflate/1.19` still. Which I don't understand, as the recipe in the current master is using a version range (`self.requires("libdeflate/[>=1.19 <=1.22]")` since my MR https://github.com/conan-io/conan-center-index/pull/26364). Also with `--build='*'` or `--build=openexr` it still fails to resolve. Am I missing something or is there maybe a bug in dependency graph resolution?

### How to reproduce it

Run build command or corresponding graph commands in the `recipes/openimageio` folder of my branch for OpenImageIO 3.0: https://github.com/conan-io/conan-center-index/pull/2567:
```
conan create 3.x --version=3.0.3.1 --build=missing --update
```

ericLemanissier:
openexr/3.2.3 is not maintained any more in [CCI](https://github.com/conan-io/conan-center-index/blob/master/recipes/openexr/3.x/conandata.yml), you should try with openexr/3.3.2 or openexr/3.2.4

valgur:
@memsharded Can you comment a bit on whether having two requires in the graph with
- `libdeflate/[>=1.19 <=1.22]`
- `libdeflate/1.19`
causing a version conflict is the intended/expected behavior or not.

[...]

PR #26593, opened 2025-02-13
vulkan-memory-allocator: relax vulkan-headers version requirement

amitiunin:
### Summary
Changes to recipe:  **vulkan-memory-allocator/***

#### Motivation
`vulkan-headers` version is too restrictive as VMA states to support Vulkan 1.0...1.4

#### Details
Modified `vulkan-headers` requirement to be at least 1.0

Tested on Windows/MSVC with vulkan-headers already in cache (uses an existing version) and using `conanio/gcc11-ubuntu16.04` image (takes the latest `vulkan-headers`)

---

### DIFF
--- a/recipes/vulkan-memory-allocator/all/conanfile.py
+++ b/recipes/vulkan-memory-allocator/all/conanfile.py
@@ -29,7 +29,7 @@ class VulkanMemoryAllocatorConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("vulkan-headers/1.3.243.0")
+        self.requires("vulkan-headers/[>=1.0]")
 
     def package_id(self):
         self.info.clear()


recipes/vulkan-memory-allocator/all/conanfile.py
```diff
@@ -29,7 +29,7 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("vulkan-headers/1.3.243.0")
+        self.requires("vulkan-headers/[>=1.0]")
```
valgur, line 32: This is more likely to create conflicts in Conan graph resolution than avoid them, unfortunately. Can't recommend.

https://github.com/conan-io/conan-center-index/issues/26512#issuecomment-2631405093

jcar87: @amitiunin - your understanding is correct.

amitiunin:
#27299 contains the same changes, I'm closing this one

PR #26602, opened 2025-02-14
Add onetbb 2021.13.0 and add recipe flags to build into apple frameworks

geertbleyen:
### Summary
Changes to recipe:  onetbb

#### Motivation
in OneTBB 2021.13.0 a new flag has been added to allow tbb to be build into Apple .frameworks instead of dylibs. iOS doesn't allow for bare dylibs to be loaded unless they are system libaries.

#### Details
- add source url for version 2021.13.0
- add an 'Apple OS only' recipe option to enable oneTBB to build into frameworks

I'm aware there is also a PR for 2022.0.0, but that one doesn't add the build option 'build_apple_frameworks' and makes some changes to the test_package, so scope isn't the same.
Since that PR doesn't make changes to the recipe itself, merging should be easy.

---

### DIFF
--- a/recipes/onetbb/all/conandata.yml
+++ b/recipes/onetbb/all/conandata.yml
@@ -2,6 +2,9 @@ sources:
   "2022.0.0":
     url: "https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2022.0.0.tar.gz"
     sha256: "e8e89c9c345415b17b30a2db3095ba9d47647611662073f7fbf54ad48b7f3c2a"
+  "2021.13.0":
+    url: "https://github.com/oneapi-src/oneTBB/archive/refs/tags/v2021.13.0.tar.gz"
+    sha256: "3ad5dd08954b39d113dc5b3f8a8dc6dc1fd5250032b7c491eb07aed5c94133e1"
   "2021.12.0":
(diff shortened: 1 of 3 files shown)


uilianries (approved):
LGTM.

@geertbleyen Thank you for bringing this feature to the OneTBB Conan recipe. I see you implemented it in OneTBB as well 👏 

I just tested locally and is looking good, the framework files are generated and packaged. The .dylib files are still packaged, they are not needed in case having the bundled framework. Still, it does not harm, so it's fine to keep it as well.

Here is my local build log, using the option `build_apple_framework=True`: 
- [onetbb-2021.13.0-apple-framework.log](https://github.com/user-attachments/files/19007380/onetbb-2021.13.0-apple-framework.log)

uilianries:
@geertbleyen Do you have any restrictions that make you using 2021.13.0, or could you use 2022.0.0 instead?

jcar87:
> @geertbleyen Do you have any restrictions that make you using 2021.13.0, or could you use 2022.0.0 instead?
> 
> EDIT: Added both versions.

Given that it adds a minor of the previous major - may not be the worst - however, tbb tends to be very backwards compatible and I would advise using the latest where possible.

PR #26640, opened 2025-02-18
msgpack23: add version 2.1

rwindegger:
### Summary
Changes to recipe:  **msgpack23/2.0**

#### Motivation
Add a new library for MessagePack based on the C++23 standard.

#### Details
Add a new library that supports serialization and deserialization of the MessagePack standard.

---

### DIFF
--- a/recipes/msgpack23/all/conandata.yml
+++ b/recipes/msgpack23/all/conandata.yml
@@ -1,0 +1,7 @@
+sources:
+  "1.0":
+    url:
+    - "https://github.com/rwindegger/msgpack23/archive/refs/tags/v1.0.tar.gz"
+    sha256: "0E3296B3162DE73CA866E973B7E447BB8C3233A854A5F94F1AE0CAA8B0E60D4C"
+requirements:
+  - "gtest/1.15.0"
--- a/recipes/msgpack23/all/conandata.yml
+++ b/recipes/msgpack23/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "1.0":
+  "2.0":
     url:
-    - "https://github.com/rwindegger/msgpack23/archive/refs/tags/v1.0.tar.gz"
-    sha256: "0E3296B3162DE73CA866E973B7E447BB8C3233A854A5F94F1AE0CAA8B0E60D4C"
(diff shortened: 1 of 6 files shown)


rwindegger:
@conan-io/barbarians Please check why the pipelines are not running.

rwindegger:
@conan-io/barbarians The pipeline is stuck since multiple days is there anything I can do?

ErniGH:
Hello @rwindegger ,

First of all, thank you very much for your PR and for your interest in adding your library to Conan.

The main issue with CI is that, in Conan, we usually don’t allow the execution of unit tests in our pipelines. The tests we typically include are focused on verifying that the linker worked correctly and that the symbols were properly linked. In this case, since it is a header-only library, we only need a small test that, by using one of its functionalities, ensures the package has been built correctly. Using GTest and building all unit tests would be excessive.

We would appreciate it if the change could be made directly upstream. CMake provides the `BUILD_TESTING` variable for these cases—you can check this link for an example of how to implement it. https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html

Additionally, requirements are usually not added to `conandata.yml`. While this approach is sometimes used in recipes with many dependencies, in this case, it is much simpler to add it directly in requirements using `self.requires("gtest/1.15.0")`. 😄

rwindegger:
Hello @ErniGH,

thanks for your kind words. I hope I've adjusted upstream in the right way. I've also changed that the dependencies are now directly in the conanfile.py.

Thanks for your support.

rwindegger:
Hello @ErniGH,
I'm not really sure why the CI is still failing. It's not finding symbols that are available in C++23, but the standard is set accordingly:
`-- Warning: Standard CMAKE_CXX_STANDARD value defined in conan_toolchain.cmake to 17 has been modified to 23 by /home/conan/workspace/cci_prod_PR-26640/conan-home/p/b/msgpa866cbd09df20c/b/CMakeLists.txt`.
Also it seems the introduced option is not honored, do I need to adjust something else in the conanfile.py?

ErniGH:
I attach here the build log in Mac:
[mac-log.txt](https://github.com/user-attachments/files/19344290/mac-log.txt)

ErniGH (approved):
LGTM Thanks for this new library 😄

Issue #26646, opened 2025-02-19
[package] mimalloc/*: Unexpected override of MI_WIN_REDIRECT


NickGaertner:
### Description

The conan recipe silently turns off the redirect mechanism on Windows by setting `MI_WIN_REDIRECT = "OFF"`. Unfortunately, the mechanism is required on Windows for mimalloc to properly work with third party code, see their [description ](https://github.com/microsoft/mimalloc?tab=readme-ov-file#windows-override) for details.

The override is unexpected, since it 
- contradicts the upstream CMake default value, 
- isn't communicated to the package consumer and 
- results in mysteriously broken builds, which seem to work until third party libraries are involved.

The PR (#23709) that introduced that line only did it so to fix intermediate versions 1.8.4 and 2.1.4. The final versions added were 1.8.7 and 2.8.7 instead, which would've built fine without setting `MI_WIN_REDIRECT = "OFF"` (verified locally).

I propose to add a **recipe option** for **MI_WIN_REDIRECT** to add transparency for users but **disable it by default** to avoid breaking builds.

### Package and Environment Details

* Package Name/Version: **mimalloc/***
* Operating System+version: **Windows***

### Conan profile

\# The profile I used for tests
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194

### Steps to reproduce

N/A

### Logs

N/A

NickGaertner:
Resolved by [#26647](https://github.com/conan-io/conan-center-index/pull/26647)

PR #26647, opened 2025-02-19
mimalloc: Added recipe option for MI_WIN_REDIRECT

NickGaertner:
### Summary
Changes to recipe:  **mimalloc/***

proposes fix for #26646
#### Motivation
Adding an option for MI_WIN_REDIRECT adds transparency and avoids hard to find build errors. See  #26646

---

### DIFF
--- a/recipes/mimalloc/all/conanfile.py
+++ b/recipes/mimalloc/all/conanfile.py
@@ -45,6 +45,8 @@ class MimallocConan(ConanFile):
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
+        else:
+            del self.options.win_redirect
 
         # single_object and inject are options
         # only when overriding on Unix-like platforms:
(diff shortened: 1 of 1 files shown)


RazielXYZ:
I agree that redirect should be configurable in the conan package. However, I think redirect only makes sense for shared override builds under windows with a msvc dynamic runtime. 
Perhaps checks should be added to that effect, and the option deleted otherwise, similar to the other checks in config_options, configure and validate?

NickGaertner:
Yeah, agreed. I didn't remove the option in `configure` since it could result in silently misconfigured builds again. Please let me know if there's a good reason to remove the option nonetheless; I'm not that familiar with conan.

perseoGI (approved):
Looks good to me!
More info at https://github.com/microsoft/mimalloc?tab=readme-ov-file#windows-override

I'm approving this PR as it will improve user experience and will solve windows arm64 issues.

PR #26667, opened 2025-02-21
OpenFX: add version 1.5.0

garyo:
* Move 1.4.0 files to 1.4/
* Add new 1.5/ directory with modern Conan 2.x recipe
* Update config.yml to point to both directories
* Require newer Conan version

### Summary
Changes to recipe:  **OpenFX/1.5.0**

#### Motivation
Add new release of OpenFX

#### Details
* Move 1.4.0 files to 1.4/
* Add new 1.5/ directory with modern Conan 2.x recipe
* Update config.yml to point to both directories
* Require newer Conan version

Note: I am a primary maintainer of OpenFX. Please ask me any questions!

---


garyo:
Any news on this? Do I need to do something to enable the CI build?

gmeeker:
I'd love to see this merged soon!

luzpaz:
soft bump

AbrilRBS:
Hi @garyo thanks a lot for your PR, sorry that this took so long for us to review this.

To simplify the process of getting this PR merged, I took the time to make some changes to the recipe:
* I've removed the old 1.4 version. Note that this does not remove the version from the `conancenter` remote, it just means we'll stop publishing new changes for that version
* I've simplified the recipe code, and added a patch so that we can control the installation from CMake, which is easier to maintain.
* I've kept your code around the components, but simplified their approach a bit
* I have disabled examples and such, which simplify the recipe and its dependencies by a lot, while also fixing missing requirement traits for its only non-examples dependency.
* I have simplified the test package code, it's now simpler and just checks that the packaging was done properly. If you have any suggestions on those, feel free to share if something more _needs_ to be checked :)

With the requirement changes, a question arose that I'd like to understand before moving forward: The plugins are controlled by `BUILD_EXAMPLE_PLUGINS` which is set to off by default, and that I've kept that way, but your old test package was using some of those. My question is then if the openfx users would find them useful/necessary to properly use openfx, or if for Conan Center index, we can keep those to off?

Note that recipes from Conan Center usually are not identical to the ones used for development of the packages in the repositories themselves, and that's why these recipes usually don't handle building examples unless strictly necessary

garyo:
Thank you so much for the detailed review, @AbrilRBS ! I will check this in full detail today, but reading your notes that all makes sense. Users of this recipe will not typically need the examples, just the headers and libs. The patch for CMake control makes sense too. I'm not a conan expert so I appreciate the expert guidance!

garyo:
I gave this a review and it looks perfect. Ready to merge.

recipes/openfx/all/conanfile.py
```diff
@@ -1,112 +1,90 @@
...
 
     def generate(self):
+        deps = CMakeDeps(self)
+        deps.generate()
+
         tc = CMakeToolchain(self)
-        tc.generate()
-        tc = CMakeDeps(self)
+        if self.settings.os == "Windows":
+            tc.preprocessor_definitions["WINDOWS"] = 1
```
AbrilRBS, line 59: @garyo one last concern we have for this recipe are this preprocesor definitions, here and in the package info. We would like to understand if they are necessary to: build the package (haven't seen anything in your upstream code outside of the conanfile), and to consume it (which is much more rare, and defining those should not be done unless necessary) :)

garyo: You're certainly right, we should absolutely not set any preprocessor definitions for consumers of this package! Really we should remove that WINDOWS definition even in our examples, and use `_WIN32` (predefined on Windows) everywhere. Currently that definition is probably needed for proper building of the package on Windows though. 
I'm not a conan expert, but is it sufficient to remove the definitions in package_info? (At least til I can make the above change)

```patch
diff --git i/recipes/openfx/all/conanfile.py w/recipes/openfx/all/conanfile.py
index 4cde6060d..63d163b84 100644
--- i/recipes/openfx/all/conanfile.py
+++ w/recipes/openfx/all/conanfile.py
@@ -81,10 +81,5 @@ class Openfx(ConanFile):
         self.cpp_info.components["Support"].includedirs = [os.path.join("include", "Support")]
         self.cpp_info.components["Support"].requires = ["Api"]
 
-            win_defines = ["WINDOWS", "NOMINMAX"]
-            self.cpp_info.components["Api"].defines = win_defines
-            self.cpp_info.components["HostSupport"].defines = win_defines
-            self.cpp_info.components["Support"].defines = win_defines
-        elif self.settings.os in ["Linux", "FreeBSD"]:
             self.cpp_info.components["HostSupport"].system_libs = ["dl"]
```

Also, when we do a new release (i.e. fixing the _WIN32 issue), do I just start another PR here? At least it should be simpler now that we have something to base it on. :-)

AbrilRBS: Yes! Thanks a lot for the clarification, I'll remove it from the package info :)

Re new version: Yes, once this PR is merged, newer versions should be way quicker to get merged, as they won't need much (if any) recipe changes

garyo (commented):
LGTM!

garyo:
Ah, I see from the CI failure that, sadly, at the moment WINDOWS is needed to be defined even for consumers, because it's used in one of our public headers. This means I should change it in our source and I'll have to push another release (which should be done anyway as I mentioned above). Stay tuned.

PR #26668, opened 2025-02-21
caches: add author field

JoelLefkowitz:
### Summary
Changes to recipe:  **caches/0.4.0**

#### Details
Adds the author field to caches/0.4.0

---

### DIFF
--- a/recipes/caches/all/conanfile.py
+++ b/recipes/caches/all/conanfile.py
@@ -1,5 +1,4 @@
 import os
-
 from conan import ConanFile
 from conan.tools.files import copy, get
 from conan.tools.layout import basic_layout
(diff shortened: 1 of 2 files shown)


jcar87:
Thanks @JoelLefkowitz - the author attribute is conventionally used to indicate the maintainer of the recipe (https://docs.conan.io/2/reference/conanfile/attributes.html#author) - which in this case is the Conan Center community, so by convention we don't define this attriburte in Conan Center recipes.

What we do to reflect the author of the _library_, is use the `homepage`attribute, which I believe is correct in this case.  Apologies if the documentation is not clear in this regard.

On the other hand, as a general note: please avoid reformatting unrelated lines on code in PRs, thanks!

JoelLefkowitz:
@jcar87 thanks for reviewing this. I'll drop the author field from the other open PRs I have. 

Maybe we can add something to the docs to make this a bit clearer for other users too?

The reformatting was to make the recipe a little more consistent with the others in the repo and to reduce the complexity of the test file. I was wrong to think to include it in this PR too since it adds confusion but do you think it's worth putting it in another PR?

jcar87:
Don't think it's worth it - the recipe as it is, is consistent with the templates we have at https://github.com/conan-io/conan-center-index/tree/master/docs  (not that formatting is enforced)
Don't think it's worth CI or reviewer time for trivial changes like that.
If the test package is only calling a few constructors in a header only library - calling 1 or 4 does not make much different either, but 🤷  - given the backlog, we are focusing on functional changes

PR #26694, opened 2025-02-25
opentrackio-cpp: add version 1.0.1

jamesmosys:
### Summary
Changes to recipe:  **opentrackio-cpp/1.0.0**

#### Motivation
Replacing with the correct version 1.0.0 to match the outputs of [CamDKit](https://github.com/SMPTE/ris-osvp-metadata-camdkit)

#### Details
Fix link and SHA to v1.0.0

---

### DIFF
--- a/recipes/opentrackio-cpp/all/conandata.yml
+++ b/recipes/opentrackio-cpp/all/conandata.yml
@@ -1,5 +1,5 @@
 sources:
-  "1.0.0":
+  "1.0.1":
     url:
-    - "https://github.com/mosys/opentrackio-cpp/archive/refs/tags/1.0.0.tar.gz"
-    sha256: "c3dd23723a3b5bdc5a210434cbcd0542c8cd993b9b89cabfc6d3c9592caeae14"
+    - "https://github.com/mosys/opentrackio-cpp/archive/refs/tags/1.0.1.tar.gz"
+    sha256: "079b3998e9e1e12e5b7a3c72f2f177bbc48f7da5220cd0f6dfd556056ba5663f"
--- a/recipes/opentrackio-cpp/all/conandata.yml
+++ b/recipes/opentrackio-cpp/all/conandata.yml
@@ -1,5 +1,5 @@
 sources:
   "1.0.0":
     url:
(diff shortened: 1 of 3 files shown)


recipes/opentrackio-cpp/all/conandata.yml
```diff
@@ -1,5 +1,5 @@
 sources:
   "1.0.0":
     url:
-    - "https://github.com/mosys/opentrackio-cpp/archive/refs/tags/v1.0.0.tar.gz"
-    sha256: "89bd489774a8efac9b18ddcff9a2219d57d46a73d53ea0a5fd71a94c27fc1257"
+    - "https://github.com/mosys/opentrackio-cpp/archive/refs/tags/1.0.0.tar.gz"
```
jcar87, line 4: Thanks @jamesmosys  - keep in mind that there is already a Conan package with version 1.0.0 that has already been published, and we very much avoid republishing existing versions with different contents -  I suspect that the "previous" version 1.0.0 was not actually so? 

The concerns is for existing consumers - if they update the recipe revision to the latest, would it still work for _consumers_? Some changes would cause breakages that are disruptive, like source incompatibilities (changes in APIs), ABI changes (conan won't rebuild consumers in this instance, so existing consumers may work at runtime) - etc. 

If this is safe to do, and you are involved with the maintainers of opentrackio-cpp, any change a 1.0.1 or similar could be published?

jamesmosys: Thanks @jcar87 you are correct the previous version should not have been 1.0.0! I have arranged a bump to opentrackio-cpp to bring it up to 1.0.1 as you suggest.

jcar87: Awesome, thanks!

jamesmosys:
Hi @jcar87, is there anything else preventing this being merged now? What is required to get this over the line? It's been a few weeks and I am really keen to get this moving as it is blocking a range of other open source projects moving forward.

jcar87 (approved):
Thanks for sorting out the 1.0.1 release!

PR #26705, opened 2025-02-26
assimp: fix bug where using draco compression in gltf2 file errors

### LINKED ISSUE #26695: [package] assimp: Draco compression not working with GLTF2
### Description

After reinstalling `assimp` via Conan after a while, loading of a GLTF2 file errors out with an error that Draco compression is not supported. This used to work for the same version (though previously we installed it under Conan < 2).

I think with the simplification of the patching in #22047 the part that sets the correct preprocessor definition for draco support got removed (in the unvendor 5.2.x patch):
Image
Without the `-DASSIMP_ENABLE_DRACO`, the loader will throw the above error.

There is no equivalent code after the patching cleanup, unless I'm missing something. Looking at the build commands for the package in the conan folders, there is also no `ASSIMP_ENABLE_DRACO` preprocessor symbol defined.

* Package Name/Version: **assimp/5.228**
* Operating System+version: **WSL running Linux Ubuntu 22.04**
* Compiler+version: **GCC 13**
* Conan version: **conan 2.8.1**
* Python version: **Python 3.12.3**

### Conan profile

[settings]
arch=x86_64
build_type=RelWithDebInfo
compiler=gcc
compiler.cppstd=20
compiler.version=13
compiler.libcxx=libstdc++11
os=Linux

[conf]

### Steps to reproduce

Install the assimp package.
Using the assimp package, try loading a gltf2 file with draco compression.

### Logs

N/A

bacusters:
Ensure -DASSIMP_ENABLE_DRACO preprocessor symbol is defined when draco is used in assimp. Add test file to verify that the importer works with draco compression.

### Summary
Changes to recipe:  **assimp/***

#### Motivation
Fixes #26695. After a refactor, the draco support was no longer enabled in assimp when using GLTF2, even though the dependency was being built. A preprocessor definition was missing to make Assimp aware that draco is available.

#### Details
Change adds a extra call in the injected `conan_deps.cmake` to define the preprocessor symbol `ASSIMP_ENABLE_DRACO`. This ensures that the GLTF2 importer of Assimp will support and use draco compression instead of throwing an error. Added another small test file that uses GLTF2 with compression and added a run in the test_package to verify that it works after the fix.

---


recipes/assimp/5.x/test_package/conanfile.py
```diff
@@ -29,3 +29,8 @@ def test(self):
 
             bin_c_path = os.path.join(self.cpp.build.bindirs[0], "test_package_c")
             self.run(f"{bin_c_path} {obj_path}", env="conanrun")
+
+            # Check that a GLTF2 file with draco compression can be loaded without error
```
jcar87, line 33: Hi @bacusters  - we no longer allow checking binary files into the repository. Please note that the intended purpose of the test package is to check that `find_package` works in CMake, that the compiler can resolve includes from the package, and that the linker can correctly resolve the library and symbols from the binary files , and *not* a functionality check. Apologies as some legacy test packages still don't reflect this.

AbrilRBS (approved):
Confirmed that this fix is needed to reflect upstream's CMakeLists, and that a gltf2 file won't load unless this is set, thanks a lot! :)

bacusters:
@AbrilRBS is there a way to retrigger the CI?

jcar87:
> @AbrilRBS is there a way to retrigger the CI?

I think there's an issue with our validator if there's a binary in the history of the branch. I've just forced pushed the branch to move this PR along, whilist @AbrilRBS  is kindly fixing this issue in the backend so that it doesn't happen in the future - thanks so much for fixing this and for your patience

AbrilRBS (approved):

PR #26709, opened 2025-02-27
boost: add 1.87.0

TibiIius:
### Summary
Changes to recipe:  **boost/1.87.0**

#### Motivation
Provide most recent version of boost. Continuation of #26539.

#### Details
I was not able to build with `without_context=True`. The boost build logs indicate that `boost_context` is not being built, but the recipe errors out with:

> These libraries were built, but were not used in any boost module: {boost_context}.

(line 2018 of the Conanfile), and the library was indeed built and present in the final package. I have no idea why that is the case.

---

### DIFF
--- a/recipes/boost/all/conandata.yml
+++ b/recipes/boost/all/conandata.yml
@@ -1,4 +1,9 @@
 sources:
+  "1.87.0":
+    url:
+      - "https://archives.boost.io/release/1.87.0/source/boost_1_87_0.tar.bz2"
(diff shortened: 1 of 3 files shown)


Nekto89:
FYI there is also PR https://github.com/conan-io/conan-center-index/pull/26185

jcar87:
thanks @TibiIius !
we expect to move through the "boost" PR backlog between tomorrow and early next week, including the new version. Will take all open PRs and existing comments into account for this, thanks!

PR #26724, opened 2025-02-28
ada: add version 3.2.1

toge:
### Summary
Changes to recipe:  **ada/3.2.1**

#### Motivation
Major version up.

#### Details
https://github.com/ada-url/ada/compare/v2.9.2...v3.2.1
https://www.yagiz.co/release-of-ada-v3/

---

### DIFF
--- a/recipes/ada/all/conanfile.py
+++ b/recipes/ada/all/conanfile.py
@@ -44,9 +44,14 @@ class AdaConan(ConanFile):
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires <charconv>, please use libc++ or upgrade compiler."
             )
-        if Version(self.version) >= "3.0.0" and self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "14.3":
+        if Version(self.version) >= "3.0.0" and \
+            ( \
+                (self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "14.3") \
+                or (self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "12") \
+                or (self.settings.compiler == "clang" and Version(self.settings.compiler.version) < "14") \
+            ):
             raise ConanInvalidConfiguration(
-                f"{self.ref} requires <ranges>, which has been provided since apple-clang 14.3 or later."
+                f"{self.ref} doesn't support ${self.settings.compiler} ${self.settings.compiler.version}"
             )
 
     def build_requirements(self):
--- a/recipes/ada/all/conanfile.py
+++ b/recipes/ada/all/conanfile.py
@@ -4,6 +4,7 @@ from conan.tools.files import get, copy, rmdir, replace_in_file
 from conan.tools.build import check_min_cppstd
 from conan.tools.scm import Version
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.microsoft import is_msvc
 import os
 
 required_conan_version = ">=2.0"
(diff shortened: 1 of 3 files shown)


anonrig (approved):

recipes/ada/all/conanfile.py
```diff
@@ -52,24 +38,35 @@ def layout(self):
...
 
         if self.settings.compiler == "clang" and Version(self.settings.compiler.version) < "12.0.0" and self.settings.compiler.libcxx != "libc++":
             raise ConanInvalidConfiguration(
                 f"{self.ref} requires <charconv>, please use libc++ or upgrade compiler."
             )
+        if Version(self.version) >= "3.0.0" and \
+            ( \
+                (self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "14.3") \
+                or (self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "12") \
+                or (self.settings.compiler == "clang" and Version(self.settings.compiler.version) < "14") \
```
ErniGH, line 51: Where are this versions coming from?

recipes/ada/all/conanfile.py
```diff
@@ -52,24 +38,35 @@ def layout(self):
...
 
     def build_requirements(self):
         self.tool_requires("cmake/[>=3.16 <4]")
 
+    def _patch_sources(self):
+        # solve APPLE RELOCATABLE SHARED LIBS (KB-H077)
+        if Version(self.version) < "2.6.10":
+            replace_in_file(self, os.path.join(self.source_folder, "cmake", "ada-flags.cmake"), "set(CMAKE_MACOSX_RPATH OFF)", "")
+        if Version(self.version) >= "3.0.0":
+            replace_in_file(self, os.path.join(self.source_folder, "src", "CMakeLists.txt"), "-Wfatal-errors", "")
```
ErniGH, line 65: Why are we patching this at 3.0.0 if this was added at 1.0.0? Why is this an issue now?
https://github.com/ada-url/ada/blame/v1.0.0/src/CMakeLists.txt#L27C10-L27C10

recipes/ada/all/conanfile.py
toge: Due to supporting `std::ranges::any_of`.
https://github.com/ada-url/ada/blob/v3.2.1/src/url.cpp#L14

recipes/ada/all/conanfile.py
toge: Because ada>=3.0.0 begins to use C++20 with lots of changes, there are several warnings in some compilers. (especially in older compilers)
This patch try to solve it.

recipes/ada/all/conanfile.py
ErniGH: Please revert this

It works with older versions. You can check by yourself here 😄 
https://godbolt.org/z/hh4sfMbnP

recipes/ada/all/conanfile.py
ErniGH: Yes, it feels like an issue. But we can use something like `tc.extra_cxxflags = ["-Wno-fatal-errors"]` in the `generate` instead of using the `replace_in_file`.

recipes/ada/all/conanfile.py
toge: @ErniGH 
Thanks!
I relaxed compiler version checks.

recipes/ada/all/conanfile.py
toge: @ErniGH @anonrig 
Thanks for comments.
I removed one `replace_in_file` which I added.

ErniGH (dismissed):
LGTM, sorry for the delay 👍

AbrilRBS (approved):
Thanks @toge!

AbrilRBS (dismissed):
Oops, missed updating the `config.yml` file :)

ErniGH (dismissed):

Issue #26717, opened 2025-02-28
[question] Consider dropping swig patch for relative paths of swiglib


uilianries:
### What is your question?

Hello, when reviewing the PR #26582 and revisiting the motivation behind the patch [swig-linux-library-path.patch](https://github.com/conan-io/conan-center-index/pull/26582/files#diff-dd23082e275cbf3098d1e7eff1070a2af752dd0f53702b949f7da86d078c5525), some thoughts came up:

- This non-official patch is a feature that allows findind `swiglib` folder by using its relative path.
- The upstream only supports absolute path, and the same is hardcoded in the `swig` executable. The result is a non-portable path when providing it via CI or similar.
- This case was reported in the upstream already, and the actual state indicates to keep absolute path only: https://github.com/swig/swig/issues/253
- This patch was sent to the upstream, but not accepted as well: https://github.com/swig/swig/pull/2509. It broke some tests and authors are not confident about.
- The seems be known in CCI as well: https://github.com/conan-io/conan-center-index/issues/15324
- The Swig's [documentation](https://www.swig.org/Doc4.3/SWIGDocumentation.html#Library_nn2) points to use the environment variable `SWIG_LIB` in case needing to override the swiglib path.
- The variable `SWIG_LIB` is configured in the recipe already, which means, when running a `conan install` + `source conanbuild.sh`, it should be enough to keep the `swig -swiglib` working.

PR #26734, opened 2025-03-01
removed swig linux library path patch

tomas-krupa:
### Summary
Changes to recipe:  **swig/4.3.0**

#### Motivation
As @uilianries explained in https://github.com/conan-io/conan-center-index/pull/26582#pullrequestreview-2648197890 the patch can be dropped

#### Details
dropping unnecessary patch, more details in https://github.com/conan-io/conan-center-index/issues/26717

---


uilianries (approved):
LGTM. As I commented in the issue, this patch was not accepted by the upstream, and using the environment variable should be enough instead.

The only risk would be breaking a few users that are consuming it.

Waiting for @perseoGI as the second pair of eyes to find anything different.

perseoGI (approved):
LGTM patch seems to be not needed and the removal is justified

PR #26736, opened 2025-03-02
[boost] Reduce test package in order to avoid large builds

uilianries:
### Summary
Changes to recipe:  **boost/[*]**

#### Motivation

The Boost package uses several full examples for the test package. It could be reduced to not only simplify the maintenance but also reduce the building time. 

#### Details

Related to #26030. I moved changes from there but only associated with the test package, so the review should be smooth. 

---


recipes/boost/all/test_package/CMakeLists.txt
```diff
@@ -1,144 +1,71 @@
 cmake_minimum_required(VERSION 3.15)
 project(test_package LANGUAGES CXX)
 
-if(UNIX AND NOT APPLE AND NOT EMSCRIPTEN)
-    # use RPATH instead of RUNPATH so that
-    # transitive dependencies can be located
-    add_link_options("LINKER:--disable-new-dtags")
```
jcar87, line 7: I think removing this should be okay, since we are now invoking the test executables with the correct Conan env

uilianries: Good point. This test package is almost the same since Conan 1.x

PR #26739, opened 2025-03-02
opencppcoverage: add recipe

luizfeldmann:
### Summary
Changes to recipe:  **opencppcoverage/0.9.9.0**

#### Motivation
- **opencppcoverage** is a code coverage analysis tool for Windows.
- Over 900 stars on GitHub.
- Seems to not have received updates for a while, but we lack other options for Windows. Linux has other tools available.

#### Details
Few unusual points about the recipe:
- The project is a Visual Studio solution that imports dependencies via _Nuget_. The _Nuget_ commands are called from a _Powershell_ script.
- I'm keeping this behavior in the recipe as per the build instructions in the repo's docs.
- The dependencies DO exist on _conan center_, but changing that would required heavy patching of the original repo.
- The application does not provide a "--version" option, and the "--help" option returns a hardcoded constant, which overflows in 32-bit, so it gives a different number from the 64-bits version, causing the _test\_package_ to be somewhat convoluted.
- This is an application meant to be used as a *tool_requires*

---

### DIFF
--- a/recipes/opencppcoverage/all/conandata.yml
+++ b/recipes/opencppcoverage/all/conandata.yml
@@ -1,0 +1,5 @@
+sources:
+  "0.9.9.0":
+    url:
+    - "https://github.com/OpenCppCoverage/OpenCppCoverage/archive/refs/tags/release-0.9.9.0.tar.gz"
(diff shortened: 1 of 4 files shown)


lucaskdc:
pls don't forget this PR

it would be really useful to have a code coverage tool that support project that due old design decisions are only built for windows

PR #26754, opened 2025-03-04
Qt6: download sources only for enabled modules

valgur:
### Summary

#### Motivation
The Qt recipe always downloads the full set of sources, which is almost 1 GB compressed and several GB unzipped.
Downloading the sources can take minutes and unpacking them is not fast either.

In contrast, the default build with no optional modules enabled requires only a 50 MB source archive download for qtbase.

#### Details
I also refactored qtmodules.conf file parsing (which is just a .gitmodules file from https://github.com/qt/qt5/blob/dev/.gitmodules) for clarity. It's strictly not needed, though.

---


valgur:
@jcar87 You might find this PR useful for any testing / experiments you might have in mind for changes related to the Qt recipe. It speeds up `conan create` for the recipe significantly.

recipes/qt/6.x.x/mirrors.txt
valgur, line 1: I dropped some mirrors that were unresponsive.

fnadeau:
Generic question: wouldn't that break sources export? Generally, package can be exported with source and rebuild without access to internet as long has the source is either in cache or in backup source.

valgur:
> Generic question: wouldn't that break sources export? Generally, package can be exported with source and rebuild without access to internet as long has the source is either in cache or in backup source.

It still uses `get(self, ...)` which loads the source archive from sources cache, if it's available.

PR #26757, opened 2025-03-04
godot-cpp: add v4.x, switch from SCons to CMake

valgur:
### Summary
Changes to recipe:  **godot-cpp/[*]**

#### Motivation
Adds godot-cpp 4.x, which has been out for two years already.

The existing SCons-based build is limited in several ways:
- SCons is much less well supported by Conan.
- SCons is intentionally very rigid and basically predefines a set of build profiles / toolchains in the project files. This means that the compiler executables in the Conan profiles are currently effectively ignored, for example.
- Limited to a static build only on v3.
- Cross-compilation is not supported by the current recipe.

#### Details

---


valgur:
More info: https://github.com/godotengine/godot-cpp/blob/master/doc/cmake.rst

Issue #12947, opened 2022-09-14
[package] : can't build wayland/1.21.0 with clang on Linux


AndreyMlashkin:
### Package and Environment Details

* Package Name/Version: **wayland/1.21.0**
* Operating System+version: **Linux Ubuntu 22.04**
* Compiler+version: **clang 14**
* Conan version: **conan 1.52.0**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=clang
compiler.cppstd=17
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux
[options]
*:with_ffmpeg=False
cmake:with_openssl=False
opencv:with_gtk=False
spirv-tools:shared=False
vulkan-loader:shared=True
[conf]
[build_requires]
[env]

### Steps to reproduce

conan install wayland/1.21.0@ -r conancenter --build wayland --build missing

### Logs

<details><summary>Click to expand log</summary>

```
wayland/1.21.0: Applying build-requirement: meson/0.63.1
wayland/1.21.0: Applying build-requirement: pkgconf/1.7.4
wayland/1.21.0: Applying build-requirement: ninja/1.11.0
wayland/1.21.0: Configuring sources in /root/.conan/data/wayland/1.21.0/_/_/source/src
wayland/1.21.0: 
wayland/1.21.0: Copying sources to build folder
wayland/1.21.0: Building your package in /root/.conan/data/wayland/1.21.0/_/_/build/4eb3a093a955aef34043efc7c4b793862d5b6a66
wayland/1.21.0: Generator 'VirtualRunEnv' calling 'generate()'
wayland/1.21.0: Generator 'VirtualBuildEnv' calling 'generate()'
wayland/1.21.0: Generator 'PkgConfigDeps' calling 'generate()'
wayland/1.21.0: Calling generate()
wayland/1.21.0: WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
wayland/1.21.0: Aggregating env generators
wayland/1.21.0: Calling build()
[...]

PR #26799, opened 2025-03-05
xkbcommon: Enable building for Android

denniskempin:
### Summary
Changes to recipe:  xkbcommon (and its dependencies wayland, xkeyboard-config) to enable building for Android

X11 is not supported, neither are xkbcommon CLI tools. Both are disabled.

#### Motivation
We are experimenting with porting wayland applications to Android, and xkbcommon is a dependency.

#### Details
- Adds "Android" to the allowed configurations for the above packages 
- Makes `with_x11` an optional option that is removed for Android
- Disables `enable_tools` for Android builds

---

### DIFF
--- a/recipes/wayland/all/conanfile.py
+++ b/recipes/wayland/all/conanfile.py
@@ -133,7 +133,9 @@ class WaylandConan(ConanFile):
             self.cpp_info.components["wayland-server"].libs = ["wayland-server"]
             self.cpp_info.components["wayland-server"].set_property("pkg_config_name", "wayland-server")
             self.cpp_info.components["wayland-server"].requires = ["libffi::libffi"]
-            self.cpp_info.components["wayland-server"].system_libs = ["pthread", "m"]
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                self.cpp_info.components["wayland-server"].system_libs = ["pthread", "m"]
+
             self.cpp_info.components["wayland-server"].resdirs = ["res"]
             if self.version >= Version("1.21.0") and self.settings.os == "Linux":
                 self.cpp_info.components["wayland-server"].system_libs += ["rt"]
@@ -149,7 +151,8 @@ class WaylandConan(ConanFile):
             self.cpp_info.components["wayland-client"].libs = ["wayland-client"]
             self.cpp_info.components["wayland-client"].set_property("pkg_config_name", "wayland-client")
             self.cpp_info.components["wayland-client"].requires = ["libffi::libffi"]
-            self.cpp_info.components["wayland-client"].system_libs = ["pthread", "m"]
+            if self.settings.os in ["Linux", "FreeBSD"]:
+                self.cpp_info.components["wayland-client"].system_libs = ["pthread", "m"]
             self.cpp_info.components["wayland-client"].resdirs = ["res"]
             if self.version >= Version("1.21.0") and self.settings.os == "Linux":
                 self.cpp_info.components["wayland-client"].system_libs += ["rt"]
--- a/recipes/wayland/all/conanfile.py
+++ b/recipes/wayland/all/conanfile.py
@@ -54,8 +54,8 @@ class WaylandConan(ConanFile):
         self.requires("expat/[>=2.6.2 <3]")
 
     def validate(self):
-        if self.settings.os != "Linux":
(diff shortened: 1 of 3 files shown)


perseoGI:
Greetings @denniskempin,
First of all, thank you for your contribution and for trying to improve CCI recipes.

I've been trying to compile these recipes with your changes, but I've been unsuccessful. Could you please share which profile you are using to build with Android? Are you using the `android-ndk` profile? Which API version?

Thank you in advance! Also, in CCI, we do not compile on Android, so it is recommended to attach compilation traces to speed up the review process.

Also, `wayland-protocols` should also be updated to support Android.

Thank you again!

denniskempin:
Hi @perseoGI, thank you for having a look at this pull request! 

I have been testing this with the following profile to compile for X86 android
```
include(default)

[settings]
os=Android
os.api_level=34
arch=x86_64
compiler=clang
compiler.version=19
compiler.libcxx=c++_static
compiler.cppstd=17

[conf]
tools.android:ndk_path=/usr/local/android-ndk-r28
```

Then building it with `conan install . --remote fork --build=missing --profile=android-x86_64 --update` on my project with the following conanfile.txt

```
[requires]
xkbcommon/1.6.0
```

I can share the dev container Dockerfile I use if that's helpful. I tried capturing the compile output of the above command, but it refuses any attempt to redirect stdout/err, and --out-file will also just print to the terminal and leave an empty file. Is there any magic incantation for capturing the output that I am missing?

denniskempin:
FWIW, I just tried the same profile to compile for armv8, which seems to work for me as well. 

In case it helps, I uploaded the dev container that I use here: https://github.com/denniskempin/wayland_demo/tree/main/nativelib/src/main/rust/.devcontainer

It's a pretty basic ubuntu container with just a few extra packages, conan and the r28 android NDK

perseoGI:
Hi @denniskempin,
Sorry for the delay.
I've been testing these recipes with proper changes.

Finally, I managed to compile all of them. 
It is worth mentioning that `xkeyboard-config` can't be compiled from a macos system (cross compiling to android) due to missing X11 support on conan ecosystem. I've not added any validation to that scenario as it is a very strange one.

On the other side, I've done some modifications:

* `wayland` recipe: when cross compiling to Android, I've encountered a `pthread` linkage issue and that is because in `package_info` method, we were always linking to `system_libraries` (as this package was only supported in linux). Android runtime has its own implementation of `pthread` and does not need (actually the linker fails) to link against `libpthread`.
* `xkbcommon` recipe: updated validation error adding `Android` as a supported OS. Also, I've made some recipe conan v2 maintenance => **Remember these tasks are only allowed by maintainers and should not be performed by contributors**.


Overall, I've tested this `xkbcommon` with an Android `x86_64` profile and with an `armv8` profile.
These are the execution logs of both `wayland` and `xkbcommon` on an android `arm` emulator:

```sh
$ emu64a:/data/local/tmp $ ./test_package_wayland
error: XDG_RUNTIME_DIR is invalid or not set in the environment.
Can't connect to display
```

```sh
$ emu64a:/data/local/tmp $ ./test_package_xkbcommon
xkbcommon: ERROR: failed to add default include path /usr/share/X11/xkb
```

We are still intrigued which is exactly your use case.
Is it `wayland` actually supported on Android?

denniskempin:
Hi @perseoGI 

Thank you for trying the build and fixing the issues.

You are correct, wayland is not actually supported on Android. That's actually what I am experimenting with, implementing a Wayland compositor on Android APIs. The xkbcommon library would be a requirement for handling keymaps.

I understand that it's a very odd use case, and compiling any of these libraries/binaries for Android is probably going to be useless to just about anyone else. Please let me know if you think this is upstreamable, we can also consider working on a forked conan index for now!

kambala-decapitator:
@perseoGI looks like removing VirtualBuildEnv / VirtualRunEnv made #12947 reproducible again, see https://github.com/Laserlicht/vcmi-dependencies/actions/runs/22695222866/job/65799993116

uilianries:
@kambala-decapitator Hello! Could you please open a proper issue reporting the case? It'll be hard to track it in a close PR. We had a similar case recently with packaging GTK, and we opted for a safer approach: https://github.com/conan-io/conan-center-index/blob/master/recipes/gtk/all/conanfile.py#L131 Regards!

kambala-decapitator:
@uilianries thanks, opened #29737

Issue #23118, opened 2024-03-16
Not able to ompile soxr library for Apple Silicon M1 Pro


flak228:
### What is your question?

Hey!)
I want to mention, that I'm a newbee in programming, so I'm really sorry, if something there is sound dumb, I just trying to do my best with it!)

I need to compile soxr library to use it with my FFMPEG build for M1 Pro
If I understood corredtly, it doesn't support Apple Silicon for now, so I got to this brunch, and tried to use that fix!)
I download conandata.yml, conanfile.py and patches form the root folder, and t
hem put them in the main folder
Then I used command conan install .
And then I used command conan create . --version=0.1.3
And then, according to the instructions, I run ./conanbuild.sh
And after that I had new file, deactivate_conanbuild-env-release-armv8.sh
And nothing more(
And if I got it right, for my purposes I need to have file with .a to use it in the next steps!)
Can I ask oyu to help me to understand what I suppose to do with that?)

One more time wants to say big sorry for being not so good in it, and gonna be really thankful for any help!)
![image_conan](https://github.com/conan-io/conan-center-index/assets/22274250/d4a1a9bc-106c-4956-a205-a19ebcdc328e)

uilianries:
@flak228 Thank you for sharing your scenario!

> I need to compile soxr library to use it with my FFMPEG build for M1 Pro

Indeed `soxr` is not managed right now in the ffmpeg recipe, it would require a new PR adding a new option, configuring cpp_info and managing the option by its dependencies. The ffmpeg is not an easy Conan recipe to understand, as you mentioned be a newbie, you may have difficult to understand that recipe at first sight.

>  If I understood corredtly, it doesn't support Apple Silicon for now, so I got to this brunch, and tried to use that fix!)

Actually, we have the [soxr](https://conan.io/center/recipes/soxr?version=0.1.3) library packaged already in Conan Center and it has Mac M1 packages available (you can see on the Packages tab). Please, try that package first.

> I download conandata.yml, conanfile.py and patches form the root folder, and them put them in the main folder

I'm not sure which guide you are using, but I would recommend you just cloning conan-center-index and building what you need instead. I expect you are running Conan 2.6.0 or similar version:

```
git clone https://github.com/conan-io/conan-center-index.git
cd conan-center-index/recipes/soxr/
conan create all/ --version=0.1.3
```

Another way, is using the Conan install command, so you do not need to clone the repository. Conan client will download the recipe and build from source automatically:

[...]

PR #26811, opened 2025-03-07
ffmpeg: Add missing soxr support

AbrilRBS:
### Summary
Changes to recipe:  **ffmpeg/[*]**

#### Motivation

Closes https://github.com/conan-io/conan-center-index/issues/23118
Note that this issue, regardless of its title, intended to address soxr in ffmpeg as a final goal. I've confirmed that I can compile soxr in my Mac M2, so that stopped being an issue at some point in the past few months

#### Details

The additions are quite straightforward. All supported versions of ffmpeg support the option, and running `yq '.versions |keys' ../config.yml | cut -d ' '  -f 2 | xargs -n1 conan create . -b=missing -o="&:swresample=True" -o="&:with_soxr=True" --version` correctly compiles for all versions, noting that:

* If the library was not found, we would see `ERROR: libsoxr not found`
* If the option was unknown, we would see `Unknown option "--enable-libsoxr"`
* By default this is false in all versions upstream, so the package for consumers is not altered without explicit opt-in in the recipe either

Attaching a few of the above successful compilations, more can be provided if needed, but they look very similar:

4.4.4 static ``: 
[build_logs_4.4.4.txt](https://github.com/user-attachments/files/18085325/build_logs_4.4.4.txt)

7.0.1 shared ``:
[build_logs_7.0.1.txt](https://github.com/user-attachments/files/18085344/build_logs_7.0.1.txt)

### DIFF
--- a/recipes/ffmpeg/all/conanfile.py
+++ b/recipes/ffmpeg/all/conanfile.py
@@ -65,6 +65,7 @@ class FFMpegConan(ConanFile):
         "with_vdpau": [True, False],
         "with_vulkan": [True, False],
         "with_xcb": [True, False],
+        "with_soxr": [True, False],
         "with_appkit": [True, False],
         "with_avfoundation": [True, False],
         "with_coreimage": [True, False],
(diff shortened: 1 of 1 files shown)


AbrilRBS:
Don't know how I messed up the last rebase

perseoGI (approved):
LGTM

PR #26828, opened 2025-03-11
libssh2/1.10.0: Fix CVE-2020-22218

Marc-Pierre-Barbier:
https://nvd.nist.gov/vuln/detail/CVE-2020-22218

patch source: https://github.com/libssh2/libssh2/pull/476

### Summary
Changes to recipe:  **lib/1.10.0**
Changes to recipe:  **lib/1.9.0**
Changes to recipe:  **lib/1.8.2**
Changes to recipe:  **lib/1.8.0**

#### Motivation
I noticed CVE-2020-22218 with a severity of 7.5HIGH was not patched in Conan center, and it was fairly easy to fix.

#### Details
all .patch files differ only by the line address the code itself is from https://github.com/libssh2/libssh2/pull/476 which is cited by nist as the fix.

---

### DIFF
--- a/recipes/libssh2/all/conandata.yml
+++ b/recipes/libssh2/all/conandata.yml
@@ -18,11 +18,6 @@ sources:
     sha256: 39f34e2f6835f4b992cafe8625073a88e5a28ba78f83e8099610a7b3af4676d4
     url: https://github.com/libssh2/libssh2/releases/download/libssh2-1.8.0/libssh2-1.8.0.tar.gz
 patches:
-  "1.10.0":
-    - patch_file: "patches/libssh2-1.9.0-CVE2020-22218.patch"
-      patch_description: "Fix CVE2020-22218"
-      patch_type: "vulnerability"
-      patch_source: "https://github.com/libssh2/libssh2/pull/476"
   "1.9.0":
     - patch_file: "patches/libssh2-1.9.0-CVE2020-22218.patch"
       patch_description: "Fix CVE2020-22218"
--- a/recipes/libssh2/all/conandata.yml
+++ b/recipes/libssh2/all/conandata.yml
@@ -17,3 +17,25 @@ sources:
   "1.8.0":
     sha256: 39f34e2f6835f4b992cafe8625073a88e5a28ba78f83e8099610a7b3af4676d4
     url: https://github.com/libssh2/libssh2/releases/download/libssh2-1.8.0/libssh2-1.8.0.tar.gz
+patches:
+  "1.10.0":
+    - patch_file: "patches/libssh2-1.9.0-CVE2020-22218.patch"
+      patch_description: "Fix CVE2020-22218"
+      patch_type: "vulnerability"
+      patch_source: "https://github.com/libssh2/libssh2/pull/476"
+  "1.9.0":
+    - patch_file: "patches/libssh2-1.9.0-CVE2020-22218.patch"
+      patch_description: "Fix CVE2020-22218"
+      patch_type: "vulnerability"
+      patch_source: "https://github.com/libssh2/libssh2/pull/476"
+  "1.8.2":
+    - patch_file: "patches/libssh2-1.8.2-CVE2020-22218.patch"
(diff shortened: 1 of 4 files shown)


jcar87:
Hi @Marc-Pierre-Barbier  - thanks for your contribution! 

Please note that if there are versions published in Conan Center that contain the fix (e.g. the 1.11.x versions) we would very strongly recommend to use those instead - libraries are provided _as is_ (as provided by the upstream maintainers). We would consider backporting patches when a patched version has not been published. However this seems like an trivial thing to backport. Thanks for linking the CVE and patch sources - this is very useful and gives us the right context :)

Marc-Pierre-Barbier:
> Hi @Marc-Pierre-Barbier - thanks for your contribution!
> 
> Please note that if there are versions published in Conan Center that contain the fix (e.g. the 1.11.x versions) we would very strongly recommend to use those instead

I agree but procedures to update dependencies can be very slow in companies, at least with these patches I hope I can make some software safer.

jcar87:
> I agree but procedures to update dependencies can be very slow in companies, at least with these patches I hope I can make some software safer.

Absolutely, we understand this. However our responsibility here is to provide the version without the vulnerability. We appreciate companies do this - however, it's not fair that they discharge their responsibilities in the open source community - they have to be active participants in making software better, it starts with determining whether their software is vulnerable and at the very least, attempting to try the latest version. We've seen many cases where it is not trivial to backport patches to older versions and upstream authors will not provide them. We may consider exceptions if there are widely used versions where the new version has API breakages, or when there is no new release that is patched - and for grave exceptional circumstances (e.g. xz-utils) - we may remove versions from the remote altogether. 

As as said, this PR an exception (trivial patch across all versions) - but at the same time, libssh2 seems to have a stable API and should be trivial to use the newer version.

recipes/libssh2/all/conandata.yml
```diff
@@ -17,3 +17,25 @@ sources:
   "1.8.0":
     sha256: 39f34e2f6835f4b992cafe8625073a88e5a28ba78f83e8099610a7b3af4676d4
     url: https://github.com/libssh2/libssh2/releases/download/libssh2-1.8.0/libssh2-1.8.0.tar.gz
+patches:
+  "1.10.0":
+    - patch_file: "patches/libssh2-1.9.0-CVE2020-22218.patch"
```
jcar87, line 22: Screenshot 2025-03-11 at 11 33 44

The upstream commit is showing that this fix should already be part of 1.10. - is this not the case?

Marc-Pierre-Barbier: When i look at the code, it is not applied. And the NIST listing show 1.10 as vulnerable.

jcar87: 1.10.0 contains a follow-up patch that moves the check a few lines above: 
https://github.com/libssh2/libssh2/commit/0b44e558f311671f6e6d14c559bc1c9bda59b8df

which is how it remains implemented up to and including the most recent 1.11.0 release

it's confusing because the patch can actually be applied to all versions (including the most recent one), but not all are vulnerable.

As to why NIST shows 1.10.0 - I would suggest reaching out to NIST or the libssh2 maintainers to clarify this - but this does appear to have been addressed in 1.10

Debian sources for 1.10.0 already have the fix without a patch: https://sources.debian.org/src/libssh2/1.10.0-3/src/transport.c/ and Debian's 1.10 is not listed as vulnerable (https://security-tracker.debian.org/tracker/CVE-2020-22218 -> says fixed for 1.10), whereas 1.9.0 does have the patch: https://sources.debian.org/src/libssh2/1.9.0-2%2Bdeb11u1/debian/patches/CVE-2020-22218.patch/ 

When it comes to security we really do want to err on the side of caution - backporting patches to earlier versions can be risky, because a patch may apply "cleanly" and create a variant of the code that doesn't exist anywhere else. 

I'd drop this from 1.10, keep it for 1.8.0, 1.9.0 and 1.9.2 and reiterate that it is the consumers responsibility to use a released version that is not vulnerable

PR #26829, opened 2025-03-11
Align the minimum required CMake in most test_packages

perseoGI:
Fix using test_package in Conan Center with CMake 4.0, which would otherwise fail with the following error:

```
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

Notes:
* The CMake integrations in Conan 2 require CMake 3.15 at a minimum - it was always the intention to keep this at 3.15 when we started migrating recipes.
* We follow the recommendations in "Professional CMake: A Practical Guide" - we know we don't support any version older than 3.15, and the test_packages are not "included" as part of other projects - so there is no reason for the policy to be set any lower.


AbrilRBS (dismissed):
This looks good on my end, we just need to see if we want to be proactive and rebase all PRS from the users that are affected by this

jcar87 (approved):

PR #26836, opened 2025-03-12
add recipe openiap

skadefro:
### Summary
add recipe for:  **openiap/0.01**

#### Motivation
Want to make it easy for people to get started using our client libraries to work with OpenCore

#### Details
First attempt at pushing a recipe for conan.
Tested it locally from both linux, macos and windows

---
- [ x ] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
- [ x ] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)
- [ x ] Tested locally with at least one configuration using a recent version of Conan

### DIFF
--- a/recipes/openiap/all/conandata.yml
+++ b/recipes/openiap/all/conandata.yml
@@ -1,0 +1,16 @@
+sources:
+  "0.0.1":
+    - url: "https://github.com/openiap/rustapi/releases/download/0.0.27/libopeniap-linux-arm64.so"
+      sha256: 3c165f4179bad9d5142f412d2a0d4561c24b625b8a01c15cb010ee2d54c4d270
+    - url: "https://github.com/openiap/rustapi/releases/download/0.0.27/libopeniap-linux-x64.so"
+      sha256: 1dd4449e121037a755e5f349bb26b48002d57117848aa90e0bd3f90ab49ddbb1
+    - url: "https://github.com/openiap/rustapi/releases/download/0.0.27/libopeniap-macos-arm64.dylib"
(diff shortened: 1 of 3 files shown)


skadefro:
Is there any ETA on when this will get merged ?
It's been stuck now for 3 weeks

jcar87:
Apologies @skadefro for the delay. 

This PR is held up because save for some very few exceptions (in particular, for build tools) - we do not repackage binary libraries in Conan Center.

Not only for the security aspect of it (they would have to come for very trusted sources or well there is a well-established level of community trust), but also because we would need to properly model the actual binary compatibility. For example, Linux users with the musl runtime may have an expectation that if they build recipes from sources, it would be compatible with their systems. Which may not be the case for `.so` files that require the glibc runtime. As a side note, for Windows I'm missing a `.lib` link library to go with the DLL so that linking is possible.

Unfortunately it is very unlikely that this PR will be merged as is. If the binaries can be built from source with a C/C++ compiler, then I would advise that the recipe is refactored so that they are built from source. If rust is required to build the library from source, please open an issue about this - there are other cases that require this and it's a valid request that we can prioritise if there is sufficient interest in the community. 

Thanks!

skadefro:
You are correct, everything is built using Rust: [https://github.com/openiap/rustapi/](https://github.com/openiap/rustapi/)
However, not all systems have Cargo installed, so it felt better to offer pre-built binaries.
I completely agree that this could pose a security risk.
If I updated this to check for Cargo and use it when available, and if not, fall back to downloading pre-built binaries (yes, we have both musl and glibc versions now—this PR is a bit old), would that “pass”?

PR #26843, opened 2025-03-12
multiple recipes: increase cmake_minimum_required (2)

perseoGI:
For the following recipes:
- bigint
- cgltf
- cmp
- co
- cpp-optparse

These recipes have a `CMakeLists.txt` in the `conan-center-index` repository rather than upstream.
Change the minimum to 3.15 to allow building with the soon to be release CMake 4.0
Fixes the following error:

```
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

Notes:
* The CMake integrations in Conan 2 require CMake 3.15 at a minimum - 
* We follow the recommendations in "Professional CMake: A Practical Guide" - we know we don't support any version older than 3.15 (for Conan 2+), and these CMakeLists are not included as part of other projects

### DIFF
--- a/recipes/bigint/all/CMakeLists.txt
+++ b/recipes/bigint/all/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.15)
 project(bigint LANGUAGES CXX)
 
 include(GNUInstallDirs)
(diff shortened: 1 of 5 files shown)


jcar87 (approved):
LGTM -

`co` is not yet ported to Conan 2, but it's good to update it so that when that happens it has the correct value

Issue #19463, opened 2023-08-28
[package] <libpq>/<*>: Openssl > 1.1.1 dependency fails to build on Mac 


KiteAnton:
### Description

Our current conanfile.txt uses
```
[requires]
libpq/14.5
...
[options]
libpq*:with_openssl=True
```
and with the latest change, https://github.com/conan-io/conan-center-index/pull/19201, this now pulls in openssl 3.1.2 which fails to build on mac. 

`Undefined symbols for architecture arm64: "_deflate", referenced from: _zlib_stateful_compress_block in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_write in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_flush in libcrypto.a(libcrypto-lib-c_zlib.o) "_deflateEnd", referenced from: _zlib_stateful_finish in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_free in libcrypto.a(libcrypto-lib-c_zlib.o) "_deflateInit_", referenced from: _zlib_stateful_init in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_write in libcrypto.a(libcrypto-lib-c_zlib.o) "_inflate", referenced from: _zlib_stateful_expand_block in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_read in libcrypto.a(libcrypto-lib-c_zlib.o) "_inflateEnd", referenced from: _zlib_stateful_finish in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_free in libcrypto.a(libcrypto-lib-c_zlib.o) "_inflateInit_", referenced from: _zlib_stateful_init in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_read in libcrypto.a(libcrypto-lib-c_zlib.o) "_zError", referenced from: _bio_zlib_write in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_read in libcrypto.a(libcrypto-lib-c_zlib.o) _bio_zlib_flush in libcrypto.a(libcrypto-lib-c_zlib.o) ld: symbol(s) not found for architecture arm64 clang: error: linker command failed with exit code 1 (use -v to see invocation) make: *** [libpq.5.15.dylib] Error 1`

### Package and Environment Details

* Package Name/Version: **libpq/**
* Operating System+version: **Apple Mac OS 13.3 M1 **
[...]

PR #26863, opened 2025-03-13
libpq: Add v16.8

the-nic:
### Summary
Changes to recipe:  **libpq/16.8**

#### Motivation
New major version

#### Details
Only add v16.8, libm was required to avoid link errors in the test

---


recipes/libpq/all/conanfile.py
```diff
@@ -52,8 +54,11 @@ def export_sources(self):
 
     def config_options(self):
         if self.settings.os == "Windows":
-            self.options.rm_safe("fPIC")
-            self.options.rm_safe("disable_rpath")
+            del self.options.fPIC
+            del self.options.disable_rpath
+            del self.options.with_icu
+        if Version(self.version) < "16.0":
+            del self.options.with_icu
```
jcar87, line 61:
```
ERROR: libpq/15.5: Error in config_options() method, line 61
	del self.options.with_icu
	ConanException: option 'with_icu' doesn't exist
Possible options are ['shared', 'with_openssl']
```

On Windows this would be a double delete

jcar87, line 61:
```suggestion
            self.options.rm_safe("with_icu")
```

ErniGH (approved):
LGTM

I test the new `with_icu` option on MacOS and Linux, and it works 😄

uilianries:
Hello, I'm doing an extra review here.

The "new" patch is actually the same as used for libpq 15:


libpq-patch-diff

Plus, doing a local build using this new option, I can build without error (so far):

#### Linux
- [libpq-16.8-linux-shared-with-icu.log](https://github.com/user-attachments/files/19500639/libpq-16.8-linux-shared-with-icu.log)
- [libpq-16.8-linux-static-with-icu.log](https://github.com/user-attachments/files/19500640/libpq-16.8-linux-static-with-icu.log)

#### MacOS
- TODO
- TODO

#### Windows
- TODO
- TODO

the-nic:
@uilianries the screenshot actually shows that the patches are not identical, and in fact the patch from 15.x does not apply on 16.x

uilianries:
> @uilianries the screenshot actually shows that the patches are not identical, and in fact the patch from 15.x does not apply on 16.x

I meant, the are identical in terms of changes, not the position. Both are applying the very same changes.

the-nic:
> > @uilianries the screenshot actually shows that the patches are not identical, and in fact the patch from 15.x does not apply on 16.x
> 
> I meant, the are identical in terms of changes, not the position. Both are applying the very same changes.

I'm not getting what you want me to do here. Remove the patch? Because thats the only option as the 15.x patch still does not apply to 16.x. It is *not* only the position that changed, if you look carefully

jcar87:
> > > @uilianries the screenshot actually shows that the patches are not identical, and in fact the patch from 15.x does not apply on 16.x
> > 
> > 
> > I meant, the are identical in terms of changes, not the position. Both are applying the very same changes.
> 
> I'm not getting what you want me to do here. Remove the patch? Because thats the only option as the 15.x patch still does not apply to 16.x. It is _not_ only the position that changed, if you look carefully

I think there's confusion here - when patches are added in a PR and they seem to be "retrofitted" from the pre-existing patches from earlier versions, the review team will try to confirm that they are indeed the same (same goal, same behaviour) - we just want to minimise the possibility of something going wrong, or something being merged without review. 

If the patches in this PR behave the same as the pre-existing patches and are just adapted to the new version being added, there's nothing to do :) - we just want to make sure there's confirmation in the PR thread that this was reviewed (hence @uilianries's message) 

>  It is _not_ only the position that changed, if you look carefully

if, on the other hand, the behaviour of the patches is not the same as the pre-existing ones, please let us know what's different so that we can review it, thanks!

uilianries:
> I'm not getting what you want me to do here. Remove the patch? Because thats the only option as the 15.x patch still does not apply to 16.x. It is not only the position that changed, if you look carefully

@the-nic Sorry, I'm confirming the patch is good 😄 , that it does not apply different change from those expected from the previous one used in 15.x. This review is always needed when a new or re-used is present in the PR. Here I only expressed the diff by a screenshot to show there is no unexpected changed in that patch.

uilianries:
@the-nic Thank you for applying that fix!

I made a few changes over your PR to make ICU work with Windows, and reverted some changes that was working previously in the recipe.

The ICU option was introduced in the version 10.0: https://www.postgresql.org/docs/release/10.0/

> Allow the [ICU](https://www.postgresql.org/docs/10/install-procedure.html#CONFIGURE) library to optionally be used for collation support (Peter Eisentraut)
>
> The ICU library has versioning that allows detection of collation changes between versions. It is enabled via configure option --with-icu. The default still uses the operating system's native collation library.

However, since version 16.0, ICU is active by default: https://www.postgresql.org/docs/release/16.0/

> Build ICU support by default (Jeff Davis) [§](https://postgr.es/c/fcb21b3ac)
>
> This removes [build flag](https://www.postgresql.org/docs/16/installation.html) --with-icu and adds flag --without-icu.

The same reflects the behavior by other package managers, like Homebrew: https://github.com/Homebrew/homebrew-core/blob/35b3b0775f48c5c623564516d73b97a42e83c6ca/Formula/lib/libpq.rb#L28

Plus, I tested locally on Windows, using both option ICU and OpenSSL. Here are my build logs:

- [libpq-16.8-windows-static-openssl.log](https://github.com/user-attachments/files/19566515/libpq-16.8-windows-static-openssl.log)
- [libpq-14.9-windows-shared-icu.log](https://github.com/user-attachments/files/19566516/libpq-14.9-windows-shared-icu.log)
- [libpq-14.9-windows-shared-openssl.log](https://github.com/user-attachments/files/19566519/libpq-14.9-windows-shared-openssl.log)
- [libpq-14.9-windows-static-icu.log](https://github.com/user-attachments/files/19566520/libpq-14.9-windows-static-icu.log)
- [libpq-14.9-windows-static-openssl.log](https://github.com/user-attachments/files/19566521/libpq-14.9-windows-static-openssl.log)
- [libpq-16.8-windows-shared-icu.log](https://github.com/user-attachments/files/19566522/libpq-16.8-windows-shared-icu.log)
- [libpq-16.8-windows-shared-openssl.log](https://github.com/user-attachments/files/19566523/libpq-16.8-windows-shared-openssl.log)
- [libpq-16.8-windows-static-icu.log](https://github.com/user-attachments/files/19566524/libpq-16.8-windows-static-icu.log)

Both version 16.8 and 14.9 were tested. All results without errors. 

I'll be checking for Linux and MacOS now.

the-nic:
Thanks a lot for your improvements @uilianries, and I dont have a strong opinion on the defaults, but I think its also a little arbitrary, as the meson [build seems to behave a bit differently](https://github.com/postgres/postgres/blob/REL_16_8/meson_options.txt#L97), but that could be unintended.

Have a question though, related to the updates of dependent libs: Currently, most libs in this repo depend on libpq/15.4, whats the policy to bump them, at least to the latest minor version? Otherwise, I would have also removed the outdated versions, but I also didnt want to explode the scope of this PR.

uilianries:
> Have a question though, related to the updates of dependent libs: Currently, most libs in this repo depend on libpq/15.4, whats the policy to bump them, at least to the latest minor version? Otherwise, I would have also removed the outdated versions, but I also didnt want to explode the scope of this PR.

@the-nic Thank you for checking other packages using libpq. That's really important in terms of keeping the entire ecosystem stable. However, bump dependencies is a topic where we tend to be more conservative, we need evidences that new versions are really required and will not break users. Also, when we need to update a dependency version, it affects the package ID as well, resulting in an entire new build graph for those affected packages, which costs not only our effort, but CI time as well. 
Still, users can use the `replace_requires` Conan feature in case they want to use a different version for a specific package.

the-nic:
@uilianries  In that case, the newer version fixes a cve, see https://www.postgresql.org/support/security/CVE-2025-1094/, hence I think it is relatively important.

uilianries:
The previous build error on Mac (https://github.com/conan-io/conan-center-index/runs/39837995740) was because pkg-config was not installed in the system, so when running autoconf, libpq project could not find ICU .pc files. The commit [6083dd0](https://github.com/conan-io/conan-center-index/pull/26863/commits/6083dd023d8fa88f157dc8023187f59ef8777cf1) fixed it by adding pkg-config as tool requirement.

I made a few tests on my local Mac, where I can find ICU working with OpenSSL:

- [libpq-14.9-mac-static-icu.log](https://github.com/user-attachments/files/19582041/libpq-14.9-mac-static-icu.log)
- [libpq-16.8-mac-static-icu.log](https://github.com/user-attachments/files/19582046/libpq-16.8-mac-static-icu.log)
- [libpq-16.8-macos-shared-icu-openssl.log](https://github.com/user-attachments/files/19582050/libpq-16.8-macos-shared-icu-openssl.log)

However, I still can find the reported error on the issue #19463 when building with static OpneSSL, but this is a topic for a separated PR and should not block this one.

uilianries:
> @uilianries In that case, the newer version fixes a cve, see [postgresql.org/support/security/CVE-2025-1094](https://www.postgresql.org/support/security/CVE-2025-1094/), hence I think it is relatively important.

@the-nic In this case, please, open a separate issue to discuss it, as this PR is focused on bumping the project version only, but it's taking more much time by adding icu option. But thank you for informing the CVE, security is indeed an important topic.

uilianries:
After analysing the Windows build results https://github.com/conan-io/conan-center-index/runs/39898705126, I see it will be better to keep ICU option as disabled by default.

The libpq is a C project, which does not require the setting compiler.cppstd configured, however, its new dependency ICU requires C++17. 

The libpq is consumed by 16 other recipes in ConanCenterIndex (see below), which may affect them, by reducing the number of available packages with C++14, for instance, when ICU is enabled.

```
find . -name conanfile.py -exec grep 'libpq/'  {} + | cut -d/ -f2,3                                                                                                                                                                
soci/all
libpqxx/all
oatpp-postgresql/all
gdal/post_3.5.0
gdal/pre_3.5.0
apr-util/all
resiprocate/all
qt/5.x.x
qt/6.x.x
taocpp-taopq/all
quickfix/all
cyrus-sasl/all
drogon/all
wt/all
poco/all
yandex-ozo/all
```

uilianries (approved):
LGTM. 

The version 16.8 is working in Mac, Windows and Linux. The ICU option is working for all platforms as well.

ErniGH (approved):
LGTM 
Yes, it's better to leave 'with_icu' disabled by default.

Good job! 👍

PR #26870, opened 2025-03-13
qtkeychain: initial

### LINKED ISSUE #19046: [request] implement qtkeychain/0.14.0
### Package Name/Version

qtkeychain/0.14.0

### Webpage

https://github.com/frankosterfeld/qtkeychain

### Source code

https://github.com/frankosterfeld/qtkeychain

Platform-independent Qt API for storing passwords securely.

DeepDiver1975:
### Summary
Changes to recipe:  **lib/qtkeychain**

#### Motivation
fixes #19046 

#### Details
Introduce new recipie qtkeychain 

---


DeepDiver1975:
Update pushed to resolve linter warnings.

recipes/qtkeychain/all/conanfile.py
```diff
@@ -0,0 +1,59 @@
...
+
+required_conan_version = ">=2.0.9"
+class PackageConan(ConanFile):
+    name = "qtkeychain"
+    description = "Platform-independent Qt API for storing passwords securely."
+    topics = ("qt", "keychain")
+    license = "BSD-3-Clause"
+    homepage = "https://github.com/frankosterfeld/qtkeychain"
+    url = "https://github.com/conan-io/conan-center-index"
+    topics = ("topic1", "topic2", "topic3")
```
DeepDiver1975, line 15: :see_no_evil:

recipes/qtkeychain/all/conanfile.py
```diff
@@ -0,0 +1,59 @@
...
+        self.requires("qt/[>=6.7 <7]", options={"with_dbus": True}, transitive_headers=True, transitive_libs=True)
+        self.requires("dbus/1.15.8", transitive_headers=True, transitive_libs=True)
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        tc.variables["BUILD_WITH_QT6"] = True
+        # tc.cache_variables["PACKAGE_BUILD_TESTS"] = False
```
DeepDiver1975, line 42: ?

DeepDiver1975:
> * Tested locally with at least one configuration using a recent version of Conan

Local build is currently blocked by https://github.com/frankosterfeld/qtkeychain/issues/267 ... :shrug:

recipes/qtkeychain/all/conanfile.py
```diff
@@ -29,20 +31,33 @@ class PackageConan(ConanFile):
     def layout(self):
         cmake_layout(self, src_folder="src")
 
+    def validate(self):
+        check_min_cppstd(self, 17)
+        qt_options = self.dependencies["qt"].options
+        if self.settings.os in ("Linux", "FreeBSD") and not qt_options.with_dbus:
```
AbrilRBS, line 37: Im having some problems testing this configuration currently, will dig deeper on Monday

recipes/qtkeychain/all/conanfile.py
```diff
@@ -29,20 +31,33 @@ class PackageConan(ConanFile):
...
+
     def requirements(self):
-        self.requires("qt/[>=6.7 <7]", options={"with_dbus": True}, transitive_headers=True, transitive_libs=True)
-        self.requires("dbus/1.15.8", transitive_headers=True, transitive_libs=True)
+        self.requires("qt/[>=6.7 <7]", transitive_headers=True, transitive_libs=True)
+        # self.requires("dbus/1.15.8")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"),
```
AbrilRBS, line 46: The library needs C++17 to actually build anything QT related when it includes QT 6, removing this line lets the recipe compile when asking conan to use C++17 or greater

recipes/qtkeychain/all/conanfile.py
```diff
@@ -29,20 +31,33 @@ class PackageConan(ConanFile):
...
+                        "SET(CMAKE_CXX_STANDARD 11)", "")
 
     def generate(self):
         tc = CMakeToolchain(self)
-        tc.variables["BUILD_WITH_QT6"] = True
-        # tc.cache_variables["PACKAGE_BUILD_TESTS"] = False
+        tc.cache_variables["BUILD_WITH_QT6"] = True
+        # Settings this to True would need LinguistTools, which requires qttools, gui and widgets
+        # qttools is not built by default in qt package, so we would need to add a check in validate()
+        tc.cache_variables["BUILD_TRANSLATIONS"] = False
```
AbrilRBS, line 54: Turned this off by default because it would need some modules that are not currently built by default in Conan, but if support for translations is desired we can add them back, but no binaries will be publish in CCI for this recipe then

AbrilRBS (commented):
Thanks a lot for taking the time to contribute to CCI, we appreciate it :)

I did some cleanups, some work still needs to be done, thanks (Please do try to avoid force-pushing as it makes collaborations harder, thanks!)

DeepDiver1975:
> I did some cleanups, some work still needs to be done, thanks

Will be watching and learning ... ;-)

AbrilRBS (commented):
Weird that the Macos build fails here but is successful locally for me, will check on monday, thanks for your patience!

DeepDiver1975:
FYI: the build is now working properly on Linux :dancers:

DeepDiver1975:
@AbrilRBS are we missing the qt option 'with_doubleconversion'?

https://github.com/conan-io/conan-center-index/blob/62d2d8fa803b2c395e32ed78d1d4e2f9f02be53c/recipes/qt/6.x.x/conanfile.py#L353-L354

AbrilRBS:
> @AbrilRBS are we missing the qt option 'with_doubleconversion'?
> 
> https://github.com/conan-io/conan-center-index/blob/62d2d8fa803b2c395e32ed78d1d4e2f9f02be53c/recipes/qt/6.x.x/conanfile.py#L353-L354

Sorry for the delay. The `multiconfiguration` option is off by default so `double-conversion` won't be built by default, but I can't seem to find where `double-conversion` is required to be used in qtkeychain, can you point me to the codepath that uses it? Thanks!

DeepDiver1975:
> but I can't seem to find where `double-conversion` is required to be used in qtkeychain, can you point me to the codepath that uses it?

I could not find it either ... by proposal to fix was more derived from the compiler error than from qtkeychain inner needs ...

AbrilRBS:
> by proposal to fix was more derived from the compiler error than from qtkeychain inner needs 

I think we should find why this fails the way it does. If the proper solution is to have qt built with those extra options, then we'll merge with the check in the validate method checking for their values, but it would be preferrable if we understand if/how we need it, in order to be able to publish binaries for this :)

DeepDiver1975:
> I think we should find why this fails the way it does. If the proper solution is to have qt built with those extra options, then we'll merge with the check in the validate method checking for their values, but it would be preferrable if we understand if/how we need it, in order to be able to publish binaries for this :)

@AbrilRBS can we drag somebody in here with deeper understanding of the qt build? I am just a consumer of the libraries and currently lack deeper understanding. THX you!

DeepDiver1975:
Thank you @jcar87 for the fix

DeepDiver1975:
nobody interested -> close

Issue #26873, opened 2025-03-14
[request] Update sqlite dependencies


stevenwdv:
A lot of packages seem to use old pinned SQLite versions. Shouldn't these be either ranges (e.g. `[^3.45]`) or be updated to new versions?

PR #26875, opened 2025-03-14
Update sqlite_orm -> sqlite3 dependency to range

stevenwdv:
### Summary
Changes to recipe:  **sqlite_orm/1.9**

#### Motivation
See #26873

#### Details
Update `sqlite3` dependency to be a range. Note: sqlite_orm does not mention specific compatibility, just that [you should install sqlite3](https://github.com/fnc12/sqlite_orm?tab=readme-ov-file#requirements). What I take away from this is that they intend to support any (modern?) sqlite3 version.

The version could also just be bumped without turning it into a range, but this seems fine by me?

---

### DIFF
--- a/recipes/sqlite_orm/all/conanfile.py
+++ b/recipes/sqlite_orm/all/conanfile.py
@@ -40,7 +40,7 @@ class SqliteOrmConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("sqlite3/3.45.0", transitive_headers=True, transitive_libs=True)
+        self.requires("sqlite3/[^3.45]", transitive_headers=True, transitive_libs=True)
 
     def package_id(self):
         self.info.clear()


recipes/sqlite_orm/all/conanfile.py
```diff
@@ -40,7 +40,7 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("sqlite3/3.45.0", transitive_headers=True, transitive_libs=True)
+        self.requires("sqlite3/[^3.45]", transitive_headers=True, transitive_libs=True)
```
jcar87, line 43:
```suggestion
        self.requires("sqlite3/[>=3.45.0 <4]", transitive_headers=True, transitive_libs=True)
```

(so that it matches the range as we have it elsewhere)

jcar87 (approved):
Thanks @stevenwdv !

stevenwdv:
@jcar87 Thanks for merging! Sorry to bother you, but since you took a look at this, can you maybe take a look at #26523 as well?

Issue #26878, opened 2025-03-14
Building recipes with CMake 4.0


jcar87:
CMake 4.0 removes compatibility with versions of CMake older than 3.5.
This means that projects that have a call to `cmake_minimum_required()`
where the version (or version range) does *not* cover 3.5 or higher,
will fail with the following error:

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

This will impact users that install CMake 4.0 once it is released.

The following use cases in Conan Center may be affected:

- Building some recipes from source where the upstream CMakeLists.txt
  has not been updated.
- Building some recipes from source where the `CMakeLists.txt` exists
  in the `conan-center-index` repository 
- Building the `test_package` project when building recipes a cloned
    copy of the `conan-center-index` repository (e.g. `conan create` or `conan test`)

**NOT** affected:
- Downloading existing binary packages (whether built locally or from the 
  Conan Center remote) - irrespective of the above.
- Building recipes from source where the upstream project assumes CMake 3.5 
  or higher (this is the **majority** of recipes)
- Building recipes from source that a `tool_requires` that specifies a version of CMake lower than 4.0
  This recipes will use CMake from Conan, irrespective of what is installed on the system.

# Workaround

If you have CMake 4.0 installed on your system and encounter errors building recipes from source,
add the following to your Conan profile(s):

```
[tool_requires]
!cmake/*: cmake/[>=3 <4]
```

This will cause Conan to fetch the most recently available CMake 3.x from the Conan Center remote,
and use that to build the recipe.

### Remedial actions in Conan Center index

The team is currently working on amending affected recipes to ensure they continue building
with CMake 4.0. This covers:

- Recipes where the upstream `CMakeLists.txt` needs to be fixed
- Recipes where the `CMakeLists.txt` exists in the `conan-center-index` repository
- All the `CMakeLists.txt` found in the `test_package` folder in the `conan-center-index` repository

Please do not open PRs addressing CMake 4.0 support, and instead comment on this issue, thanks.

jcar87:
List of recipes affected by CMake 4 broken by status:

[...]

PR #26879, opened 2025-03-14
CMake: add version 4.0.0-rc4

perseoGI:
### Summary
Changes to recipe:  **cmake/4.0.0-rc4**

#### Motivation
Following https://github.com/conan-io/conan-center-index/issues/26878
Adding CMake latest release candidate will facilitate testing recipes.
Also, this is a good opportunity to distinguish from other package managers which does not yet support any v4 CMake version.

Checkout SHAs here: https://cmake.org/files/ReleaseCandidate/cmake-4.0.0-rc4-SHA-256.txt

### DIFF
--- a/recipes/cmake/binary/conandata.yml
+++ b/recipes/cmake/binary/conandata.yml
@@ -1,4 +1,23 @@
 sources:
+  "4.0.0-rc4":
+    Linux:
+      armv8:
+        url: "https://cmake.org/files/ReleaseCandidate/cmake-4.0.0-rc4-linux-aarch64.tar.gz"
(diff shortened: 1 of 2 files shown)


jcar87 (approved):

AbrilRBS:
For everyone wanting to use this with recipes that have cmake as a tool_requires using version ranges like `[>=3.5 <4]`, note that although 4-rc4 < 4 semverwise, this version won't be picked up even if you set the `core.version_ranges:resolve_prereleases` conf to `True`.

You'll want to add an entry to your `[replace_tool_requires]` section in your profile with something like 
```
[replace_tool_requires]
cmake/*: cmake/[>=3.5 <5]
```
or whatever the new limit you desire that includes the the CMake 4 releases (But noting that the prerelease versions won't yet be picked up without the conf mentioned above)

PR #26880, opened 2025-03-14
openssl: clarify maintained versions

jcar87:
OpenSSL 3.x can take too long to build on CI - and there's diminshing returns in terms of simultaneously maintaining too many versions from the same recipe, considering that as per OpenSSL's policy, OpenSSL 3.x is to remain backwards compatible https://openssl-library.org/policies/releasestrat/index.html)

- Other recipes in Conan Center depend on the version range `openssl/[>=1.1 <4]` (or similar) - which means that they will require the most recent version as soon as it's published
- Align with the supported releases by OpenSSL maintainers

This does *not* mean that those other versions are "deleted" - once a version is published in the Conan Center remote, it remains there, and if it works for a user, it will continue to work.
What this means is that we will not be maintaining the recipe (bugfixes/features) for versions other than the ones as per this policy.

### DIFF
--- a/recipes/openssl/3.x.x/conandata.yml
+++ b/recipes/openssl/3.x.x/conandata.yml
@@ -1,28 +1,25 @@
+# Versions to keep:
+#  - The last patch version of the supported releases, as per https://openssl-library.org/source/
+#  - The last patch version of the current LTS release
+#  - The most recent FIPS validated version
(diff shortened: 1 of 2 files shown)


franramirez688 (approved):

franramirez688 (approved):

uilianries (commented):
LGTM

uilianries (approved):

PR #26901, opened 2025-03-18
pcre 8.45: support cmake 4

jcar87:
pcre 8.45: fixes to support CMake 4
* increase cmake minimum required to 3.5, fixes build error when using CMake 4.0
* use `$<TARGET_FILE:dftables>` instead of inspecting the `LOCATION` target property
* remove lines that are a no-op in Conan 2
* remove older versions: this is the last release of pcre


perseoGI (approved):
Perfect!

PR #26905, opened 2025-03-19
Disable Control Flow Guard on OpenSSL asm build

zhihaoy:
### Summary
Changes to recipe:  **openssl/3.4.1**

#### Motivation
Allow end-users to build openssl from the Conan center when their global profile add CFG.

#### Details
See also: openssl/openssl#22554

---

(Tested by applying the patch in a private fork running Conan v1)

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -439,6 +439,15 @@ class OpenSSLConan(ConanFile):
         defines = 'defines => add("%s"),' % defines if defines else ""
         targets = "my %targets"
 
+        # OpenSSL 3's use of assembly is incompatible with Control Flow Guard
+        if is_msvc(self) and not self.options.no_asm:
+            def not_guard_cf(flag):
+                return re.match(r'[-/](?i)(guard:cf)$', flag) is None
+
+            cflags = list(filter(not_guard_cf, cflags))
+            cxxflags = list(filter(not_guard_cf, cxxflags))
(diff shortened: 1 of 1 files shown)


jcar87:
HI @zhihaoy  - thanks for your contribution



> Allow end-users to build openssl from the Conan center when their global profile add CFG.

This PR is an usual approach of overriding the user's intention in a way that it does.. the opposite of what's requested.

If the flags are set in the profile, I would strongly advise to just exclude openssl instead:
```
[conf]
!openssl/*:tools.build:cflags
!openssl/*tools.build:cxxflags
```

zhihaoy:
> If the flags are set in the profile, I would strongly advise to just exclude openssl instead:
> 
> ```
> [conf]
> !openssl/*:tools.build:cflags=
> !openssl/*:tools.build:cxxflags=
> ```

Correct me if I don't get it - Let's say this profile adds multiple cflags to all packages, like this
```
[conf]
tools.build:cflags=/Qspectre /diagnostics:caret
tools.build:cxxflags=/EHsc /Qspectre /diagnostics:caret
```
Now I want to add `/guard:cf` to all packages except openssl, I can
```
[conf]
tools.build:cflags=/Qspectre /diagnostics:caret
tools.build:cxxflags=/EHsc /Qspectre /diagnostics:caret
!openssl/*:tools.build:cflags+=/guard:cf
!openssl/*:tools.build:cxxflags+=/guard:cf
```
?

jcar87:
That's how it's intended to work, yes :)

PR #26911, opened 2025-03-20
azure-sdk-for-cpp: allow macos and clang in conanfile

dascandy:
### Summary
Changes to recipe:  **azure-sdk-for-cpp/all**

#### Motivation
We are using this package with AppleClang-16 on MacOS and Clang-17 on Ubuntu, and this is upstreaming our local patch.

There is a single warning that Clang finds for which we don't have a patch that we are confident in, and as such we're disabling warnings-as-errors while building this package as a dependency.

#### Details
We loosened the checks in the conanfile.py to allow building with clang and on macos. The check for clang-10 is to give an early error for very old versions; we have only tested with 16 and 17 but want to avoid a sequence of "actually this works on 15", "actually ..." PRs after this.

---

### DIFF
--- a/recipes/azure-sdk-for-cpp/all/conanfile.py
+++ b/recipes/azure-sdk-for-cpp/all/conanfile.py
@@ -51,16 +51,18 @@ class AzureSDKForCppConan(ConanFile):
             check_min_cppstd(self, 14)
 
         # Open to contributions for windows and apple
-        if self.settings.os != "Linux":
+        if self.settings.os != "Linux" and self.settings.os != "Macos":
             raise ConanInvalidConfiguration(
                 f"{self.ref} Conan recipe in ConanCenter still does not support {self.settings.os}, contributions to the recipe welcome.")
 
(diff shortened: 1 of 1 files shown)


dascandy:
force-push was for rebase; I had apparently not updated in the past 8 days.

jcar87 (approved):
LGTM!!

jcar87:
> but want to avoid a sequence of "actually this works on 15", "actually ..." PRs after this.

Greatly appreciate this, thanks!

dascandy:
@jcar87 when pulling azure-sdk-for-cpp from conan-center with its most recent version (1.11.3) I still get the old version. Is there a way to force it to go to the new version, or do I have to wait for a newer release?

jcar87:
> @jcar87 when pulling azure-sdk-for-cpp from conan-center with its most recent version (1.11.3) I still get the old version. Is there a way to force it to go to the new version, or do I have to wait for a newer release?


```
conan remote update conancenter --url https://center2.conan.io
```

PR #26921, opened 2025-03-24
oatpp-swagger: Added version 1.3.0.latest

### LINKED ISSUE #22205: [package] oatpp-swagger/1.3.0: Test package does not find oatpp includes
### Description

The test package of oatpp-swagger fails to build because it does not find `oatpp/core/data/mapping/type/Type.hpp`, a header file of oatpp (tested on Linux and Windows).

I'm not sure what's the good way of fixing this:
1. In the recipe of oatpp-swagger, use
   ```python
   self.requires(f"oatpp/{self.version}", transitive_headers=True)
   ``` 
2. Add a dependency to oatpp in the test package recipe of oatpp-swagger (and CMakeLists.txt).

There are other packages in a similar situation:
- oatpp-libressl: has the same problem
- oatpp-openssl: (other problem: fails to include `openssl/ssl.h`)
- oatpp-postgresql: uses solution 1 (`transitive_headers=True`)
- oatpp-sqlite: has the same problem
- oatpp-websocket: has the same problem

Shall I propose a PR that applies solution 1 to these packages?

* Package Name/Version: **oatpp-swagger/1.3.0**
* Operating System+version: **Linux Ubuntu 20.04 / Windows 10**
* Compiler+version: **GCC 11 + MSVC**
* Docker image: **N/A**
* Conan version: **conan 2.0.14**
* Python version: **Python 3.9.6**

### Conan profile

[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=20
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=193
os=Windows
[conf]
tools.microsoft.msbuild:vs_version=17
tools.microsoft.msbuild:max_cpu_count=8

### Steps to reproduce

conan export --version 1.3.0 recipes/oatpp/all
conan export --version 1.3.0 recipes/oatpp-swagger/all
conan create -b missing --version 1.3.0 recipes/oatpp-swagger/all

### Logs

```
...
======== Testing the package: Building ========
oatpp-swagger/1.3.0 (test package): Calling build()
oatpp-swagger/1.3.0 (test package): Running CMake.configure()
oatpp-swagger/1.3.0 (test package): RUN: cmake -G "Ninja" -DCMAKE_TOOLCHAIN_FILE="/builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package/build/gcc-11-x86_64-release/generators/conan_toolchain.cmake" -DCMAKE_INSTALL_PREFIX="/builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package" -DCMAKE_POLICY_DEFAULT_CMP0091="NEW" -DCMAKE_BUILD_TYPE="Release" "/builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package"
-- Using Conan toolchain: /builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package/build/gcc-11-x86_64-release/generators/conan_toolchain.cmake
-- The CXX compiler identification is GNU 11.4.0
-- Conan: Component target declared 'oatpp::oatpp-swagger'
-- Conan: Component target declared 'oatpp::oatpp'
-- Conan: Component target declared 'oatpp-test::oatpp-test'
-- Conan: Target declared 'oatpp::oatpp-test'
-- Configuring done
-- Generating done
-- Build files have been written to: /builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package/build/gcc-11-x86_64-release
oatpp-swagger/1.3.0 (test package): Running CMake.build()
oatpp-swagger/1.3.0 (test package): RUN: cmake --build "/builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package/build/gcc-11-x86_64-release" -- -j32
[1/2] Building CXX object CMakeFiles/test_package.dir/test_package.cpp.o
FAILED: CMakeFiles/test_package.dir/test_package.cpp.o 
/usr/bin/c++  -isystem /root/.conan2/p/b/oatppd91237ff96d15/p/include/oatpp-1.3.0/oatpp-swagger -m64 -O3 -DNDEBUG -MD -MT CMakeFiles/test_package.dir/test_package.cpp.o -MF CMakeFiles/test_package.dir/test_package.cpp.o.d -o CMakeFiles/test_package.dir/test_package.cpp.o -c /builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package/test_package.cpp
In file included from /builds/cc/devops/conan-center-index/recipes/oatpp-swagger/all/test_package/test_package.cpp:2:
/root/.conan2/p/b/oatppd91237ff96d15/p/include/oatpp-1.3.0/oatpp-swagger/oatpp-swagger/Types.hpp:28:10: fatal error: oatpp/core/data/mapping/type/Type.hpp: No such file or directory
   28 | #include "oatpp/core/data/mapping/type/Type.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
Recipe 'oatpp-swagger/1.3.0 (test package)' cannot build its binary
It is possible that this recipe is not Conan 2.0 ready
If the recipe comes from ConanCenter, report it at https://github.com/conan-io/conan-center-index/issues
If it is your recipe, check if it is updated to 2.0
ERROR: oatpp-swagger/1.3.0 (test package): Error in build() method, line 21
	cmake.build()
```

</details>

### LINKED ISSUE #26353: [request] oatpp-swagger/1.3.0.latest
### Package Name/Version

oatpp-swagger/1.3.0.latest

### Changelog

https://github.com/oatpp/oatpp-swagger/releases/tag/1.3.0-latest

The above-mentioned version is newly released by the upstream project and not yet available as a recipe.
Please add this version.

DannyLihard:
Fixes #26353.
Fixes #22205.

### Summary
Changes to recipe:  **oatpp-swagger/1.3.0.latest**

#### Motivation
1.3.0-latest is available.

#### Details
- As in the `oatpp` package, the `_version` property was added because of the `.latest` suffix of the Conan recipe version.
- Fixed missing `oatpp` transitive headers.

---

### DIFF
--- a/recipes/oatpp-swagger/all/conanfile.py
+++ b/recipes/oatpp-swagger/all/conanfile.py
@@ -47,7 +47,7 @@ class OatppSwaggerConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires(f"oatpp/{self.version}", transitive_headers=True)
+        self.requires(f"oatpp/{self.version}", transitive_headers=True, transitive_libs=True)
 
     def validate(self):
         if self.info.settings.compiler.get_safe("cppstd"):


DannyLihard:
Hello, I prefer not to use `1.3.1` for the following reasons:
- Consistency with the equivalent [PR](https://github.com/conan-io/conan-center-index/pull/23696) for the `oatpp` recipe, which does *not* exist in version 1.3.1. The `1.3.0.latest` recipe version was picked instead of the upstream `1.3.0-latest` Git tag to solve the pre-release order issue.
Because all Oat++ Conan recipes are bound to use the same version, this would force us to create the `oatpp/1.3.1` recipe before `oatpp-xxx/1.3.1` recipes.
- Oat++ tagged `1.3.1` the very same commit already tagged `1.3.0-latest`. i.e., everything in the source code still is '1.3.0', which is not great IMHO, and could create some unexpected problems.
- Not all Oat++ extensions have the `1.3.1` tag (e.g. `oatpp-openssl`). For the same reason as above, everytime someone wants to create `1.3.1` recipe PR for one of the concerned extensions, they must first ask Oat++ to create the upstream tag.

DannyLihard:
Could someone trigger the CI run please ? I think I blocked the thing because I force pushed with a correct commiter/author email address to be CLA-compliant.

jcar87:
> Could someone trigger the CI run please ? I think I blocked the thing because I force pushed with a correct commiter/author email address to be CLA-compliant.

Done! :) 
CI build is unrelated to CLA - for new contributors we do a visual inspection of the diff before running the PR on our CI machines (depending on the recipe/PR, a single PR can have quite a significant resource utilisation, and it also gives us the change to filter out PRs opened by bots, etc)

recipes/oatpp-swagger/all/conanfile.py
```diff
@@ -20,38 +20,33 @@ class OatppSwaggerConan(ConanFile):
...
+        if version.endswith(".latest"):
+            version = version[:-len(".latest")]
+        return version
+
     def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
 
-    def configure(self):
-        if self.options.shared:
```
jcar87, line 36: any context as to why this is being removed?

DannyLihard: As the `configure()` method was defined only to do some actions in the case where `shared=True` was passed, it does not make sense anymore because the `shared` option was simply removed, since the library can be built in its static form only.

jcar87: I don't think that is the case - when `self.options.shared`, the `fPIC` option is removed, not the shared option.
The `shared` option is marked as "invalid" on Windows - but not on the other platforms. The upstream project does seem to support this being toggeable via `BUILD_SHARED_LIBS` (https://github.com/oatpp/oatpp-swagger/blob/8cb460b546e84b06939a12897c85a6d863c1c53e/CMakeLists.txt#L20) 

Unless I'm missing something, this change would introduce a regression since the library could be previously built as shared on supported platforms

DannyLihard: Sorry I wasn't clear, I meant I (not the `configure()` method) completely removed the `shared` option in the last commit, because even though the corresponding CMake option exists in the upstream project, it is said [here](https://oatpp.io/docs/installation/unix-linux/#installation-cmake-options) not to use it. Although this statement is for the Oat++ base module, I think we can extend it to Oat++ extensions. Indeed, it does not make sense to build multiple Oat++ extensions if each one of them is built on top of `liboatpp.a`.
An application that uses several of these extensions has no reason to link against them if all of them are shared but they are all linked against the same static library.

Moreover, I don't think we're introducing a regression here. I tried building `oatpp-swagger/1.3.0` with `*:shared=True` and it does not work, even before this PR.
We can see in the the [build logs](https://c3i.jfrog.io/c3i/misc/summary.html?json=https://c3i.jfrog.io/c3i/misc/logs/pr/8524/4/summary.json) of the CCI PR for `oatpp-swagger/1.3.0` that `liboatpp-swagger.so` was built with linking against `liboatpp.a`, that is because `oatpp-swagger:shared=True` was passed.
However, in the CI run of this PR, we can see `*:shared=True` is passed, so the difference lies in how the CI configuration was modified since the PR for `oatpp-swagger/1.3.0`, not in the current PR itself.

DannyLihard:
Hello, could someone merge this please ? See explanations in the review above.

recipes/oatpp-swagger/all/conanfile.py
jcar87: Thanks for adding more context. I would ask to please remove these changes and open an issue so that we can analyse this better and take appropriate action, and keep the changes in this PR to the minimal required by the new version.

Please sympathise with the fact that we have to review a high number of pull requests and we can't guess the reason for a change like this - this can only delay PR reviews.

> Moreover, I don't think we're introducing a regression here. I tried building oatpp-swagger/1.3.0 with *:shared=True and it does not work, even before the current PR.

The `*:shared=True` does work, in the sense that it produces a package with the correct file. What doesn't work is the test_package, which is fixed by either having it link against `oatpp::oatpp`(which is reasonable, since these are extension modules), or by adding `transitive_libs=True` to the requirement on this recipe (which also reflects the usage - since including the headers cause these symbol requirements). 

So out of caution, given that it is possible to have consumers using `*:shared=True` even with the current recipe, we would not remove this (and we would not do it as part of an unrelated PR). 

Please note that as per our [contribution guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md#pull-request-scope), we ask that pull requests have a single clear objective (in this case, I'm guessing adding a new version).

DannyLihard:
Hello, I eventually kept changes to a minimum. Can this PR be approved please?

DannyLihard:
@franramirez688 Hello, could you review this please? It's been open for a long time and there are little changes.

recipes/oatpp-swagger/all/conanfile.py
```diff
@@ -28,6 +28,13 @@ class OatppSwaggerConan(ConanFile):
         "fPIC": True,
     }
 
+    @property
+    def _version(self):
+        version = str(self.version)
+        if version.endswith(".latest"):
+            version = version[:-len(".latest")]
+        return version
```
franramirez688: `self.version` is always a string, so we can just split it:

```suggestion
        return self.version.split(".latest")[0]
```

recipes/oatpp-swagger/all/conanfile.py
```diff
@@ -43,7 +50,7 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires(f"oatpp/{self.version}")
+        self.requires(f"oatpp/{self.version}", transitive_headers=True, transitive_libs=True)
```
franramirez688, line 50: I think that we only need the `transitive_headers` param?

```suggestion
        self.requires(f"oatpp/{self.version}", transitive_headers=True)
```

recipes/oatpp-swagger/all/conanfile.py
```diff
@@ -43,7 +47,7 @@ def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires(f"oatpp/{self.version}")
+        self.requires(f"oatpp/{self.version}", transitive_headers=True)
```
franramirez688: I just saw the errors in the CI. Reverting the previous change and keeping it as it was:

```suggestion
        self.requires(f"oatpp/{self.version}", transitive_headers=True, transitive_libs=True)
```

ErniGH (approved):

PR #26932, opened 2025-03-25
capnproto: CMake 4 support

perseoGI:
capnproto: fixes to support CMake 4

* Increase CMake minimum required to 3.5, fixing build error when using CMake 4

Upstream v2 [CMakeLists.txt](https://github.com/capnproto/capnproto/blob/v2/CMakeLists.txt#L1) already supports CMake 4.

### DIFF
--- a/recipes/capnproto/all/conanfile.py
+++ b/recipes/capnproto/all/conanfile.py
@@ -1,5 +1,5 @@
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration, ConanException
+from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import fix_apple_shared_install_name
 from conan.tools.build import check_min_cppstd, cross_building
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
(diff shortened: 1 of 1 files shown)


recipes/capnproto/all/conanfile.py
```diff
@@ -116,6 +116,9 @@ def generate(self):
             tc.variables["EXTERNAL_CAPNP"] = False
             tc.variables["CAPNP_LITE"] = False
             tc.variables["WITH_OPENSSL"] = self.options.with_openssl
+            tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5" # CMake 4 support (v2 branch does not need this)
+            if Version(self.version) > "1.1.0": # pylint: disable=conan-unreachable-upper-version
```
jcar87, line 120: if we know an upcoming version 2 already has this fixed, maybe only error if a version is above 1.1.0 but below 2.0?

perseoGI: That is a good point, but we don't really know how are they going to version those new versions, right now they have a `v2` branch. 
Maybe when the new version is released some one decides that we should add the new recipe in a `2.x` folder. We don't really know. Anyways, I will add the check to `< 2` as I think capnproto is not going to make big changes on minor/patches releases.

jcar87 (approved):

Issue #26906, opened 2025-03-19
[request] systemc/3.0.1


ankenman:
### Package Name/Version

systemc/3.0.1

### Changelog

https://github.com/accellera-official/systemc/releases/tag/3.0.1

### Context about the new update

This is a new major release. Included is support for macOS and ARM.

PR #26937, opened 2025-03-25
Add: new version systemc/3.0.1

ErniGH:
### Summary
Changes to recipe:  **systemc/3.0.1**

https://github.com/accellera-official/systemc/compare/2.3.3...3.0.1

We need to disable the sanitizer on macOS:
https://github.com/accellera-official/systemc/blob/3.0.1/src/CMakeLists.txt#L95

Windows with shared=True needs `SC_WIN_DLL`:
https://github.com/accellera-official/systemc/blob/main/INSTALL.md#33-building-against-a-systemc-dll

Close: https://github.com/conan-io/conan-center-index/issues/26906

### DIFF
--- a/recipes/systemc/all/conanfile.py
+++ b/recipes/systemc/all/conanfile.py
@@ -1,11 +1,10 @@
-import textwrap
-
 from conan import ConanFile
 from conan.errors import ConanInvalidConfiguration
 from conan.tools.apple import is_apple_os
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
-from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir, replace_in_file
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
 from conan.tools.microsoft import is_msvc
+from conan.tools.build import check_min_cppstd
 import os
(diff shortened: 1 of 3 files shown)


recipes/systemc/all/conanfile.py
```diff
@@ -71,6 +75,22 @@ def validate(self):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
+        if Version(self.version) >= "3.0.1":
+            replace_in_file(self, os.path.join(self.source_folder, "src/sysc/kernel/sc_cor_qt.cpp"),
```
jcar87, line 80: where does this patch come from and what error is it attempting to address?

AbrilRBS (dismissed):

uilianries (approved):
LGTM https://github.com/accellera-official/systemc/compare/2.3.3...3.0.1

The build log presents examples to ON, but no example is really built. There more changes related CMake, but most to examples and testing. The sanitizer flags are not present on Linux. 

Plus, we have to pass a definition to consume/build DLL always: https://github.com/accellera-official/systemc/blob/3.0.1/src/CMakeLists.txt#L437

PR #26941, opened 2025-03-26
cctz: CMake 4 support

perseoGI:
cctz: fixes to support CMake 4

* Increase CMake minimum required to 3.5, fixing build error when using CMake 4
* Removed conan v1 deprecated code

Upstream master [CMakeLists.txt](https://github.com/google/cctz/blob/master/CMakeLists.txt#L3) already supports CMake 4.

### DIFF
--- a/recipes/cctz/all/conanfile.py
+++ b/recipes/cctz/all/conanfile.py
@@ -3,9 +3,11 @@ from conan.tools.apple import is_apple_os
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
 from conan.tools.microsoft import check_min_vs
+from conan.tools.scm import Version
+from conan.errors import ConanException
 import os
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=2.1"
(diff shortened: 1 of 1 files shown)


recipes/cctz/all/conanfile.py
```diff
@@ -56,8 +58,9 @@ def generate(self):
         tc.variables["BUILD_BENCHMARK"] = False
         # For shared msvc
         tc.variables["CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS"] = True
-        # Relocatable shared lib on Macos
-        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
+        tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5" # CMake 4 support (master branch already has 3.16)
+        if Version(self.version) > "2.4": # pylint: disable=conan-unreachable-upper-version
```
jcar87, line 62: if upsteram already has fixed this and we have a realistic expectation that the newer version will be cut from their main branch (the change was added 8 months ago!) - then I wouldn't error out for a new version, just apply this to all versions 2.4 or lower

jcar87 (approved):

PR #26975, opened 2025-03-27
easyhttpcpp: CMake 4 support

perseoGI:
easyhttpcpp: fixes to support CMake 4

* Fix `any` generator warning reported by linter

### DIFF
--- a/recipes/easyhttpcpp/all/conanfile.py
+++ b/recipes/easyhttpcpp/all/conanfile.py
@@ -1,11 +1,12 @@
 from conan import ConanFile
-from conan.errors import ConanInvalidConfiguration
+from conan.errors import ConanInvalidConfiguration, ConanException
 from conan.tools.build import check_min_cppstd, valid_min_cppstd
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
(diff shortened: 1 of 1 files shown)


recipes/easyhttpcpp/all/conanfile.py
```diff
@@ -49,6 +50,7 @@ def layout(self):
 
     def requirements(self):
         self.requires("poco/1.12.4", transitive_headers=True, transitive_libs=True)
+        self.requires("openssl/[>=1.1 <4]")
```
jcar87, line 53: Added this because the library is making _direct calls_ to openssl code:
https://github.com/sony/easyhttpcpp/blob/ef7dca78f76877d51b329464c322e87eef54195f/src/unix/SslContextCreatorImpl.cpp#L39-L40

Otherwise can fail with
```
[100%] Linking CXX shared library lib/libeasyhttp.dylib
  "_SSL_CTX_load_verify_locations", referenced from:
      easyhttpcpp::SslContextCreatorImpl::createContext(Poco::AutoPtr<easyhttpcpp::EasyHttpContext>) in SslContextCreatorImpl.cpp.o
clang: error: linker command failed with exit code 1 (us
```

jcar87 (approved):

PR #26976, opened 2025-03-27
armadillo: fix source method and cmake 4 support

marlamb:
Clearly a typo.


jcar87:
Thanks @marlamb  - I'm changing this to use the pattern in all other recipes, as there is no reason to change the filename that I can see (the default behaviour is to derive the filename from the URL) . It also wasn't broken as long as the filesystem allows using `{` and `}` in a filename.

pinging @perseoGI  so that we can retrofit the CMake 4 fixes where needed

PR #26988, opened 2025-03-28
folly: CMake 4 support

perseoGI:
folly: fixes to support CMake 4

* Removed conan v1 deprecated code
* Removed `is_msvc` check in `_cppstd_flag_value` as it is already checked in L169 and linter was reporting `possibly using variable 'prefix' before assignment (possibly-used-before-assignment)`

### DIFF
--- a/recipes/folly/all/conanfile.py
+++ b/recipes/folly/all/conanfile.py
@@ -131,13 +131,6 @@ class FollyConan(ConanFile):
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=False)
 
-    def _cppstd_flag_value(self, cppstd):
-        prefix = "c"
-        year = str(cppstd)
-        if year > "17":
-            year = "latest"
-        return f"{prefix}++{year}"
-
     def generate(self):
         env = VirtualBuildEnv(self)
         env.generate()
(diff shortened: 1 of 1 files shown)


recipes/folly/all/conanfile.py
```diff
@@ -132,11 +132,10 @@ def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=False)
 
     def _cppstd_flag_value(self, cppstd):
-        if is_msvc(self):
-            prefix = "c"
-            year = str(cppstd)
-            if year > "17":
-                year = "latest"
+        prefix = "c"
```
jcar87, line 135: I wouldn't refactor anything because of a linter warning that does not apply (it doesn't apply because `_cppstd_flag_value` is only used already guarded by a conditional `is_msvc`) - so it's not "wrong", regardless of what the linter says.

Having `c++latest

If we really want to refactor this: turn this into a variable used here instead - that way its localised (otherwise it looks strange to have a `_cppstd_flag_value` that would have `c++latest` as a value, which is only accepted by msvc.

https://github.com/conan-io/conan-center-index/blob/19e213670100fa4a3cfd0abdf83b31d600039fe7/recipes/folly/all/conanfile.py#L171-L172

jcar87 (changes requested):

jcar87 (approved):

PR #26993, opened 2025-03-28
cgal: bump dependencies

lrineau:
### Summary

Changes to recipe:  **cgal/6.0.1**

#### Motivation

Fixes a bug reported by a user on Windows about Boost version 1.83.0:

> Apparently, this version of Boost has a small bug that causes an annoying warning to be generated when compiled using MSVC:
>
> ```
> C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Tools\MSVC\14.41.34120\include\stdfloat(12): warning STL4038: The contents of <stdfloat> are available only with C++23 or later.
> ```

#### Details

I switch from requirements with explicit versions to version ranges.

---

### DIFF
--- a/recipes/cgal/all/test_package/conanfile.py
+++ b/recipes/cgal/all/test_package/conanfile.py
@@ -9,8 +9,8 @@ class TestPackageConan(ConanFile):
     generators = "CMakeDeps", "CMakeToolchain"
 
     def requirements(self):
-        self.requires("eigen/3.4.0")
         self.requires(self.tested_reference_str)
+        self.requires("eigen/[*]")
 
     def build(self):
         cmake = CMake(self)


lrineau:
@AbrilRBS @uilianries: you approved several of my last pull-requests about CGAL. Do you know why this one is stalled?

uilianries:
@lrineau Hello :wave: Thank you for putting your time on this PR! :smile: 

Please, try to refer the origin of that reported error, so we can reproduce it first. It's really important to track and understand this new change.

Please, avoid using version range largely, we only use it for some packages in CCI: https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges. We know very well about their abi/api compatibility, so using version range should not break users for newer versions. From my experience, Boost is not abi compatible, as some libraries change the required C++ standard from one minor version to another. Still, I only remember Boost Cobalt needing C++20, but no library using C++23. 

And please, only change the specific package that's required. We are more conservative in terms of bumping versions in Conan Center Index, as past actions resulted in compatibility problems, missing packages, and other issues. 

Cheers!

recipes/cgal/all/conanfile.py
```diff
@@ -47,10 +47,10 @@ def layout(self):
...
 
     def requirements(self):
-        self.requires("boost/1.83.0")
-        self.requires("eigen/3.4.0")
-        self.requires("mpfr/4.2.1")
-        self.requires("gmp/6.3.0")
+        self.requires("boost/[>=1.84.0]")
+        self.requires("eigen/[>=3.4.0 <4.0.0]")
+        self.requires("mpfr/[>=4.2.1 <5.0.0]")
+        self.requires("gmp/[>=6.3.0 <7.0.0]")
```
jcar87:
```suggestion
        self.requires("boost/[>=1.83.0 <=1.89.0]")
        self.requires("eigen/3.4.0")
        self.requires("mpfr/[>=4.2.1 <5]")
        self.requires("gmp/[>6.3.0 <7]")
```

expanding boost to cover 1.83.0 or greater, up to and including the most recent release (the test package seems to pass just fine with 1.89.0)

Still leaving 1.83.0 as that was deemed to be compatible and the issue motivating this PR is only a warning.

Apologies this PR took too long. For future reference, the reason this was *not* prioritised:
- It was fixing a warning, while we presume the code was still compiling
- We do not generally merge PRs with unrelated changes, and when it comes to version ranges, we require some assurance that the versions in the range are all source compatible, which is not always guaranteed (it is certainly not the case for boost)

lrineau:
Unfortunately, the upstream CGAL has an issue with boost/1.89.0. See https://github.com/boostorg/bimap/issues/49.

jcar87:
> Unfortunately, the upstream CGAL has an issue with boost/1.89.0. See https://github.com/boostorg/bimap/issues/49.

Thanks for checking this.

Given that the original motivation behind this PR is to fix a warning, and that it works with the versions of the dependencies listed in the recipe - I would be inclined to close this PR - we can update once there's a stronger need.

sloriot:
> Unfortunately, the upstream CGAL has an issue with boost/1.89.0. See [boostorg/bimap#49](https://github.com/boostorg/bimap/issues/49).

The issue is actually with boost HEAD, not any released version.

PR #27002, opened 2025-03-28
Gdal 3.10.3

vilasfe:
### Summary
Changes to recipe:  **gdal/3.10.3**

Fixes: https://github.com/conan-io/conan-center-index/issues/27003

#### Motivation
Update gdal to 3.10.3. One key change in the library is including cstdint to make it compile on more recent gcc versions

#### Details
- Added 3.10.3
- Updated patches from older versions to work with the significant cmake changes that gdal has made
- Updated dependencies (especially arrow) to take advantage of new gdal features that needed newer version
- Check that the conanfile works for 1.66.0 and 2.x

---


vilasfe:
@conan-io/barbarians It looks like the Pre-Checks step hung for most of the weekend. I did a resync with master this morning and the build is still stuck in the Pre-Check step. Any ideas?

vilasfe:
I managed to fix the build. It was due to setting options in the `conanfile.py` in the incorrect spot. For some reason that was hanging the build rather than reporting an error.

recipes/gdal/post_3.5.0/conanfile.py
```diff
@@ -169,12 +169,13 @@ def config_options(self):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+        self.options["arrow"].filesystem_layer = True
```
ErniGH, line 172:
```suggestion
```
Adding this option by default will break the dependency graph. You can use this option adding `-o arrow/*:filesystem_layer=True`.

recipes/gdal/post_3.5.0/conanfile.py
```diff
@@ -169,12 +169,13 @@ def config_options(self):
...
 
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("json-c/0.17")
+        self.requires("json-c/0.18")
```
ErniGH, line 178: Could you please let me know why you're bumping this dependency? 
We have the policy of bump dependencies when it is fully necessary

recipes/gdal/post_3.5.0/conanfile.py
```diff
@@ -184,7 +185,7 @@ def requirements(self):
         if self.options.with_armadillo:
             self.requires("armadillo/12.6.4")
         if self.options.with_arrow:
-            self.requires("arrow/14.0.2")
+            self.requires("arrow/16.1.0")
```
ErniGH, line 188: Same here ☝️

recipes/gdal/post_3.5.0/conanfile.py
```diff
@@ -221,7 +222,7 @@ def requirements(self):
         if self.options.with_heif:
             self.requires("libheif/1.16.2")
         if self.options.with_jpeg == "libjpeg":
-            self.requires("libjpeg/[>=9e]")
+            self.requires("libjpeg/9e")
```
ErniGH, line 225: Please keep the version range

recipes/gdal/post_3.5.0/conanfile.py
```diff
@@ -634,10 +635,10 @@ def _patch_sources(self):
                         "JXL_THREADS", "JXL", strict=False)
         # Workaround for Parquet and ArrowDataset being provided by Arrow on CCI.
         replace_in_file(self, os.path.join(self.source_folder, "cmake", "helpers", "CheckDependentLibraries.cmake"),
-                        "gdal_check_package(Parquet", "# gdal_check_package(Parquet")
+                        "gdal_check_package(Parquet", "# gdal_check_package(Parquet", strict=False)
         if Version(self.version) >= "3.6.0":
             replace_in_file(self, os.path.join(self.source_folder, "cmake", "helpers", "CheckDependentLibraries.cmake"),
-                            "gdal_check_package(ArrowDataset", "# gdal_check_package(ArrowDataset")
+                            "gdal_check_package(ArrowDataset", "# gdal_check_package(ArrowDataset", strict=False)
```
ErniGH, line 641: Please revert this. If you have any issues, please explain your situation.

It is not expected to use `strict=False` because it skips applying patches without any warning.

ErniGH (commented):
Thanks for your effort, all the patches are nice, but there are some issues with the conanfile 😄

recipes/gdal/post_3.5.0/conanfile.py
vilasfe: reverted

recipes/gdal/post_3.5.0/conanfile.py
vilasfe: reverted.

This was from when I was building on conan 1.66, which I'm stuck with until gitlab updates, and I forgot to revert it before pushing to conan center that is on conan 2.x.

recipes/gdal/post_3.5.0/conanfile.py
vilasfe: Reverted

These patches are unneeded in newer gdal versions, so a version check made more sense than `strict=False` and is safer for the reasons you gave.

recipes/gdal/post_3.5.0/conanfile.py
vilasfe: For this one, I tried a version check as well. It turns out that gdal will not build without that setting on updated arrow for it to use the new features.

If we're worried about the graph, is it a maybe a better answer for a PR to the arrow recipe that adds `filesystem_layer` as a default option, and then it would propagate to the gdal recipe? Open to other thoughts here too.

ErniGH:
The best approach I see for these cases is that, since this PR is about adding a new version, we should focus on making the minimal changes necessary to support it. If there's no need to bump the Arrow version, let's leave it at `14.0.2`.

If you'd like to upgrade Arrow to use all the new `file_system` features in GDAL, it would be better to do that in a separate PR where we can discuss those changes more thoroughly. It's better to have two simple PRs than one that's a bit more complex.

vilasfe:
I reverted to `arrow/14.0.2` though it seems to be having a build issue on my box.  Can we try the build system to see if it is an issue on my machine or if gdal-3.10.2 does actually require both the `file_system` change and a version bump for `arrow`?

That should help answer if the update is required. If it is, then we can do a different PR against the `arrow` recipe and put this one on hold until that one closes, or take a different approach if you have a better one.

vilasfe:
After reviewing the build logs from the weekend, it seems that leaving `arrow/14.0.2` should work, as long as we enable the `filesystem_layer = True` option for any `gdal/[>=3.10.0]` so I'll push a fix to enable that while still leaving the older version number.

WDYT @ErniGH ?

ErniGH (approved):
Sorry for the delay. Arrow has the `filesystem_layer` option set to `false` by default in `conan-center`, so using it as `true` in this recipe could cause issues. Since the version in the requirements is fixed and the package is compiled in the remote with that option, it doesn’t throw an error.

I’ve been running some tests, and Glad requires this option to work. Also, I think we should set Arrow’s `filesystem_layer` option to `true` by default in future recipe updates. It’s the default behavior of Arrow, and I believe keeping it as `false` could lead to confusion.

I believe there’s no risk in merging this new version. 👍 Thanks @vilasfe

franramirez688 (commented):
Hi @vilasfe! Thanks a lot for all the efforts contributing to this PR! It worked fine on my local machine 👏 

Last question before approving, can we update this version to 3.10.3? That is the latest patch. I just checked it locally, and it works fine. The diff is:

 ```diff
diff --git a/recipes/gdal/config.yml b/recipes/gdal/config.yml
index db5e1c5ea..95d7d2800 100644
--- a/recipes/gdal/config.yml
+++ b/recipes/gdal/config.yml
@@ -1,5 +1,5 @@
 versions:
-  "3.10.2":
+  "3.10.3":
     folder: "post_3.5.0"
   "3.8.3":
     folder: "post_3.5.0"
diff --git a/recipes/gdal/post_3.5.0/conandata.yml b/recipes/gdal/post_3.5.0/conandata.yml
index 8b7f90d82..5ddfe71dd 100644
--- a/recipes/gdal/post_3.5.0/conandata.yml
+++ b/recipes/gdal/post_3.5.0/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "3.10.2":
-    url: "https://github.com/OSGeo/gdal/releases/download/v3.10.2/gdal-3.10.2.tar.gz"
-    sha256: "ca710aab81eb4d638f5dbd4f03d4d4b902aeb6ee73a3d4a8c5e966b6b648b0da"
+  "3.10.3":
+    url: "https://github.com/OSGeo/gdal/releases/download/v3.10.3/gdal-3.10.3.tar.gz"
+    sha256: "e4bf7f104acbcb3e2d16c97fd1af2b92b28d0ba59d17d976e3ef08b794f4153b"
   "3.8.3":
     url: "https://github.com/OSGeo/gdal/releases/download/v3.8.3/gdal-3.8.3.tar.gz"
     sha256: "f7a30387a8239e9da26200f787a02136df2ee6473e86b36d05ad682761a049ea"
@@ -12,7 +12,7 @@ sources:
     url: "https://github.com/OSGeo/gdal/releases/download/v3.5.3/gdal-3.5.3.tar.gz"
     sha256: "a9ea0300d17e35bab71df4f16e62bb2fb8081caf994ab3ee0502ce4cf0d4e593"
 patches:
-  "3.10.2":
+  "3.10.3":
     - patch_file: "patches/3.10.2/0-replace-find-package.patch"
       patch_description: "Use custom version of find_package() for Conan deps"
       patch_type: "conan"
```

franramirez688 (approved):
👏

vilasfe:
@franramirez688 looks like you got to the fix before I could. Thanks for taking care of that. :-)

uilianries (commented):
Looking good, but let's wait for #27431 first, so we can mitigate the missing dependency error in the results: https://github.com/conan-io/conan-center-index/pull/27002/checks?check_run_id=42215308767

ErniGH (approved):
Thanks all! good job 👍

Issue #27005, opened 2025-03-29
[request] resultlib/1.0.0


guillermocalvo:
### Package Name/Version

resultlib/1.0.0

### Webpage

https://result.guillermo.dev/

### Source code

https://github.com/guillermocalvo/resultlib

### Description of the library/tool

A header-only C library that provides a monadic container to encapsulate operations that may succeed or fail in a type-safe way.

PR #27006, opened 2025-03-29
resultlib: new recipe

guillermocalvo:
### Summary
Add new recipe:  **resultlib/1.0.0**

#### Motivation
Close: #27005

#### Details
https://result.guillermo.dev/

---

### DIFF
--- a/recipes/resultlib/all/conandata.yml
+++ b/recipes/resultlib/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "1.0.0":
+    url: "https://github.com/guillermocalvo/resultlib/archive/refs/tags/1.0.0.tar.gz"
+    sha256: "aed74ce86f2b7b62a6484fd2637b388e23a832fc290f4e7f14f953cc633a6eca"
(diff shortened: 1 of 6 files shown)


AbrilRBS (approved):
Thanks for taking the time to submit your library to Conan Center Index, we appreciate it.

I've taken the liberty to clean the recipe up a little bit, but it was mostly great already :)

Do note that by default Conan is generating a `resultlib::resultlib` cmake target if another name is not specified, so your users will `target_link_libraries` to that target name - but I've seen that you don't define one/have `install()` instructions in your upstream `CMakeLists.txt`

If you plan to add one in the future and would prefer a different target name, do let me know so we can change it before merging, otherswise lgtm :)

guillermocalvo:
Thanks a lot for your review and the cleanup @AbrilRBS

Since this was a header-only library, I didn't think `install()` instructions were really necessary. Other than that, I wanted to keep things as simple as possible, so the default cmake target looked good enough to me.

However, I'd love to follow best practices, so I'm open to changing the target name if you think it makes more sense.

PR #27009, opened 2025-03-30
boost: add process_use_std_fs option

melak47:
### Summary
Changes to recipe:  **boost/\***

#### Motivation
Boost.Process (V2) has a configuration option to use `std::filesystem` instead of `boost::filesystem`. Exposing this option allows a code base using both `std::filesystem` and `boost::process::v2` to avoid having to convert paths back and forth between the two libraries.

#### Details
Exposes a new option `process_use_std_fs` (defaults to `False`) analogous to `filesystem_use_std_fs`, which defines the `BOOST_PROCESS_USE_STD_FS` build option and macro.

---
  - Tested on windows with msvc and conan 2.14.0

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -83,6 +83,7 @@ class BoostConan(ConanFile):
         "filesystem_no_deprecated": [True, False],
         "filesystem_use_std_fs": [True, False],
         "filesystem_version": [None, "3", "4"],
+        "process_use_std_fs": [True, False],
         "layout": ["system", "versioned", "tagged"],
         "magic_autolink": [True, False],  # enables BOOST_ALL_NO_LIB
         "diagnostic_definitions": [True, False],  # enables BOOST_LIB_DIAGNOSTIC
@@ -122,6 +123,7 @@ class BoostConan(ConanFile):
         "filesystem_no_deprecated": False,
         "filesystem_use_std_fs": False,
         "filesystem_version": None,
+        "process_use_std_fs": False,
         "layout": "system",
         "magic_autolink": False,
         "diagnostic_definitions": False,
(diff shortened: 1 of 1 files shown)


melak47:
> conan install --requires=boost/1.87.0 -o '&:extra_b2_flags=["boost.process.fs=std"]' --build=missing

It seems like the recipe expects this to be a string, not an array (shlex.split is used to turn it into an array of options here: https://github.com/conan-io/conan-center-index/blob/master/recipes/boost/all/conanfile.py#L1447).

Trying it with `&:extra_b2_flags="boost.process.fs=std"'`, it kind of works -- but then the user code needs to manually `#define BOOST_PROCESS_USE_STD_FS` before including the headers :/

uilianries:
@melak47 Thank you for checking it and fixing it. I made a mistake with Conan conf x recipe options, indeed it expects a string instead. :sweat: 

> but then the user code needs to manually #define BOOST_PROCESS_USE_STD_FS before including the headers :/

Yes, that's expected, same for other Boost header-only modules (around 123 are header-only in Boost). So you can imagine the mess would be that recipe in case adapting each define for those hundred modules and their particularities :fearful: 

Still, you can add a profile to track this Boost feature. The advantage will be being able to share it via Conan [config](https://docs.conan.io/2/reference/commands/config.html) command:

```ini
# ~/.conan2/profiles/boost

include(default)

[settings]
compiler.cppstd=17

[options]
boost/*:extra_b2_flags="boost.process.fs=std"

[conf]
&:tools.build:defines=['BOOST_PROCESS_USE_STD_FS=1']
```

The [profile](https://docs.conan.io/2/reference/config_files/profiles.html) above re-uses the content of your default profile, and only for Boost package, it passes that b2 option. In order to pass the definition to the package that you are building, you can use the [config](https://docs.conan.io/2/reference/config_files/global_conf.html) for extra defines. Also, because  [std::filesystem](https://en.cppreference.com/w/cpp/filesystem) requires C++17, the compiler cppstd version is set to 17 as well. 

Then, you can build it without any extra effort: 

    conan install -r conancenter --requires=boost/1.87.0 -pr boost --build=missing

I hope this solution fits your case. Regards!

amerry:
As of Boost 1.87.0, Boost.Process no longer uses the filesystem::path type in the compiled library, so `BOOST_PROCESS_USE_STD_FS` only affects the headers (meaning it can be used in client code regardless of how Boost was built). Therefore this option would only be useful for Boost 1.86 and earlier.

PR #27017, opened 2025-03-31
json-c: CMake 4 support

perseoGI:
json-c: fixes to support CMake 4

* Disable compilation of tests and, in newer versions, the application compilation (`json_parse`) which is a demo application never installed by cmake.

### DIFF
--- a/recipes/json-c/all/conanfile.py
+++ b/recipes/json-c/all/conanfile.py
@@ -47,8 +47,13 @@ class JSONCConan(ConanFile):
         if Version(self.version) >= "0.15":
             tc.variables["BUILD_STATIC_LIBS"] = not self.options.shared
             tc.variables["DISABLE_STATIC_FPIC"] = not self.options.get_safe("fPIC", True)
-        # To install relocatable shared libs on Macos
-        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
+        tc.cache_variables["BUILD_TESTING"] = False
+        if Version(self.version) < "0.17":
+            # To install relocatable shared libs on Macos
+            tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
+            tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5" # CMake 4 support
+        else:
+            tc.cache_variables["BUILD_APPS"] = False
         tc.generate()
 
     def build(self):


recipes/json-c/all/conanfile.py
```diff
@@ -47,8 +47,13 @@ def generate(self):
...
             tc.variables["DISABLE_STATIC_FPIC"] = not self.options.get_safe("fPIC", True)
-        # To install relocatable shared libs on Macos
-        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
+        tc.cache_variables["BUILD_TESTING"] = False
+        if Version(self.version) < "0.17":
+            # To install relocatable shared libs on Macos
+            tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0042"] = "NEW"
+            tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5" # CMake 4 support
+        else:
+            tc.cache_variables["BUILD_APPS"] = False
```
jcar87, line 56: Please do not include this line. If a recipe was already building apps, let's avoid causing regressions to users that are relying on this.

jcar87 (changes requested):

recipes/json-c/all/conanfile.py
perseoGI: Yes, I've taken that into account!

`cmake install` was never installing the generated application:
```sh
tree  <package-folder>                                                                                                                                                    
.
├── conaninfo.txt
├── conanmanifest.txt
├── include
│   └── json-c
│       ├── arraylist.h
│       ├── debug.h
│       ├── json.h
│       ├── json_c_version.h
│       ├── json_config.h
│       ├── json_inttypes.h
│       ├── json_object.h
│       ├── json_object_iterator.h
│       ├── json_patch.h
│       ├── json_pointer.h
│       ├── json_tokener.h
│       ├── json_types.h
│       ├── json_util.h
│       ├── json_visit.h
│       ├── linkhash.h
│       └── printbuf.h
├── lib
│   └── libjson-c.a
└── licenses
    └── COPYING
```

```sh
ls build/Release/apps/json_parse                                                                                                                           
build/Release/apps/json_parse
```

Therefore no consumer should have been able to use the `json_parse`. What do you think?

jcar87 (approved):

jcar87 (approved):

PR #27021, opened 2025-03-31
Add xtensor 0.26.0 version

flip1995:
### Summary
Add version:  **xtensor/0.26.0**

#### Motivation

This version adds support for clang 19 and above.

#### Details

The version 0.26.0 required some changes to the conanfile.py. Starting from this version a cppstd >= 17 is required. This also means, that newer compiler versions will be required. I took those exact versions from the cppreference and looked at other packages that use cppstd >= 17.

Further, code was restructured between this and the last version and with that include paths changed. The changes to the test_package make sure that the test still works with all versions.

Requires:
- https://github.com/conan-io/conan-center-index/pull/27016
- https://github.com/conan-io/conan-center-index/pull/27019

---

### DIFF
--- a/recipes/xtensor/all/conanfile.py
+++ b/recipes/xtensor/all/conanfile.py
@@ -45,6 +45,8 @@ class XtensorConan(ConanFile):
         if self.options.xsimd:
             if Version(self.version) < "0.24.0":
                 self.requires("xsimd/7.5.0")
+            elif Version(self.version) < "0.26.0":
+                self.requires("xsimd/13.0.0")
             else:
                 self.requires("xsimd/13.2.0")
         if self.options.tbb:
--- a/recipes/xtensor/all/conanfile.py
+++ b/recipes/xtensor/all/conanfile.py
@@ -32,17 +32,27 @@ class XtensorConan(ConanFile):
 
     @property
     def _min_cppstd(self):
-        return "14"
+        if Version(self.version) < "0.26.0":
+            return "14"
+        return "17"
 
     @property
     def _compilers_minimum_version(self):
-        # https://github.com/xtensor-stack/xtensor/blob/master/README.md
         return {
-            "Visual Studio": "14",
-            "msvc": "190",
-            "gcc": "4.9",
-            "clang": "4",
-        }
+            "14": {
+                "Visual Studio": "14",
+                "msvc": "190",
+                "gcc": "4.9",
+                "clang": "4",
+            },
+            # https://en.cppreference.com/w/cpp/compiler_support/17
+            "17": {
+                "Visual Studio": "15",
+                "msvc": "191",
+                "gcc": "7",
+                "clang": "4",
+            },
+        }.get(self._min_cppstd, {})
 
     def export_sources(self):
         export_conandata_patches(self)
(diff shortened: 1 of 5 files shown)


flip1995:
@AbrilRBS @uilianries  This is the final PR, then you won't have to deal with me anymore in the near future :sweat_smile: 

I rebased the PR on top of the `xlt` and `xsimd` bump PRs and applied the same cleanups to the `conanfile.py` that you folks did in the `xlt` PR.

As always: Let me know if I still need to address anything!

recipes/xtensor/all/conanfile.py
```diff
@@ -30,34 +29,24 @@ class XtensorConan(ConanFile):
...
+            self.requires("xtl/0.7.5")
+        else:
+            self.requires("xtl/0.8.0")
         self.requires("nlohmann_json/3.11.3")
         if self.options.xsimd:
             if Version(self.version) < "0.24.0":
                 self.requires("xsimd/7.5.0")
             else:
-                self.requires("xsimd/13.0.0")
+                self.requires("xsimd/13.2.0")
```
uilianries, line 49: The upstream says xsimd/11.x for xtensor 0.25 (supported version in CCI). Is there any compatibility limitation by using 13.x directly?

recipes/xtensor/all/test_package/CMakeLists.txt
```diff
@@ -7,11 +7,15 @@ find_package(xtensor REQUIRED CONFIG)
 
 add_executable(${PROJECT_NAME} test_package.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE xtensor)
-target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_14)
+target_compile_features(${PROJECT_NAME} PRIVATE cxx_std_17)
```
uilianries, line 10:
```suggestion
```

Conan 2.x will have settings.compiler.cppstd always, so the C++ standard passed to test package will be passed automatically. This target_compile_features is more a Conan 1.x feature, because in Conan 1.x we didn't have cppstd, so we needed to enforce the C++ standard in the test package.

uilianries (commented):
@flip1995 I only have a small question related to the compatibility range. Otherwise, it's looking good.

recipes/xtensor/all/conanfile.py
flip1995: Website said 11.x, but the conanfile already required 13.0 for it. I didn't want to risk any breaking and assumed that minor version bumps don't contain breaking changes.

I can insert an `elif Version(self.version) < "0.26.0"` branch that keeps it pinned to `13.0.0`.

flip1995: aab3bdf3b96b478176b610f5d63a474db05cef9d

Adds this additional check. I can drop it again, if you want.

recipes/xtensor/all/test_package/CMakeLists.txt
flip1995: 115409b5fca88b76d143038dee3d160a61bc4e3d

uilianries (approved):
LGTM.

The changes are following the upstream recommendation for their dependencies: https://github.com/xtensor-stack/xtensor?tab=readme-ov-file#dependencies

Also, C++17 is required since 0.26.0, as documented in the README: https://github.com/xtensor-stack/xtensor#introduction

uilianries:
@flip1995 Thank you for your PR, it's looking good. Now, we just need to wait for further review. 😄

AbrilRBS (approved):

Issue #27020, opened 2025-03-31
Detours - Only Release and static runtime allowed


BastianGrahm:
As the detours conanfile states
```python
def validate(self):
    if self.settings.os != "Windows":
        raise ConanInvalidConfiguration("Only os=Windows is supported")
    if is_msvc(self) and not is_msvc_static_runtime(self):
        # Debug and/or dynamic runtime is undesired for a hooking library
        raise ConanInvalidConfiguration("Only static runtime is supported (MT)")
    if self.settings.build_type != "Release":
        raise ConanInvalidConfiguration("Detours only supports the Release build type")
```
only static runtime is allowed as well as only the Release build type.

As explained shortly [in my comment on the original PR](https://github.com/conan-io/conan-center-index/pull/11480#discussion_r2018748497) the vcpkg package manager does not have these restrictions and in the upstream repository I couldn't find any hints that this should be forbidden.

As I'm not very deep into hooking libraries I asked on the original PR where the original contributor confirmed, that this is **_probably_** removable.
We used the library with Debug symbols before without having any issues.

I propose removing the restrictions and maybe adding a mere warning, that Detours **_might_** behave unexpected when not build with static runtime and Release mode, and will setup a pull request for it (if I figure out how to do that properly).

P.S.: I wasn't really sure which category this issue falls in, since it's not really an upstream update, but also not really a bug.

PR #27026, opened 2025-03-31
Detours - Replace restriction by warning

BastianGrahm:
Replaces restrictions on build type and runtime by warnings.

### Summary
Changes to recipe:  **detours** (all versions)

#### Motivation

For further explanation, see issue #27020

#### Details

The unneeded restrictions are replaced by mere warnings, as even though there is no evidence building with dynamic runtime or with debug symbols does any harm it is not particularly testable.
One could imagine the debug build enabling certain optimizations that mess with the address calculation Detours has to do. Or maybe it has an impact on the binary section layout.

We patched the recipe locally for our use cases and did not (yet) run into any issues. Also as elaborated in #27020, our previous package manager did not have these restrictions and we used it successfully for quite a while.

---

### DIFF
--- a/recipes/detours/all/conanfile.py
+++ b/recipes/detours/all/conanfile.py
@@ -40,10 +40,9 @@ class DetoursConan(ConanFile):
         if self.settings.os != "Windows":
             raise ConanInvalidConfiguration("Only os=Windows is supported")
         if is_msvc(self) and not is_msvc_static_runtime(self):
-            # Debug and/or dynamic runtime is undesired for a hooking library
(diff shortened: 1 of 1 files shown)


jcar87 (approved):
Thanks @BastianGrahm , and thanks for investigating the original PR that introduced that warning.

in the past year (and some), we have become more strict with these sort of invalidation in the recipe, we make every effort to make sure the specific reason and evidence behind the invalidation is in the PR precisely so that we don't have unexplained things like this in the future

PR #27046, opened 2025-03-31
Libwebsockets/cmake4

perseoGI:
libwebsockets: fixes to support CMake 4

### DIFF
--- a/recipes/libwebsockets/all/conanfile.py
+++ b/recipes/libwebsockets/all/conanfile.py
@@ -7,7 +7,7 @@ from conan.tools.scm import Version
 import os
 import textwrap
 
-required_conan_version = ">=1.56.0"
+required_conan_version = ">=2.1"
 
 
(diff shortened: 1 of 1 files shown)


jcar87 (approved):

recipes/libwebsockets/all/conanfile.py
```diff
@@ -422,6 +422,7 @@ def generate(self):
         tc.variables["LWS_WITH_SYS_SMD"] = self.settings.os != "Windows"
         tc.variables["DISABLE_WERROR"] = True
 
+        tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5" # CMake 4 support
```
jcar87, line 425: it's interesting that this was fixed in master upstream ~2 years ago
https://github.com/warmcat/libwebsockets/blob/main/CMakeLists.txt

but none of the recent releases have it

jcar87 (approved):

PR #27083, opened 2025-04-01
add extra_flags option to OpenSSL

Enhex:
### Summary
Changes to recipe:  **openssl/3.4.1**

#### Motivation
the recipe is missing some build flags as options, so having extra_flags enables working around that.

#### Details
looking at https://git.symas.net/symas-public/openssl/-/blob/openssl-3.4/INSTALL.md?ref_type=heads#enable-and-disable-features
some flags are missing, for example: no-http, no-scrypt, no-siphash, ...

haven't tested it locally.

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -89,12 +89,14 @@ class OpenSSLConan(ConanFile):
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
         "tls_security_level": [None, 0, 1, 2, 3, 4, 5],
+        "extra_flags": [None, "ANY"],
     }
     default_options = {key: False for key in options.keys()}
     default_options["fPIC"] = True
     default_options["no_md2"] = True
     default_options["openssldir"] = None
     default_options["tls_security_level"] = None
+    default_options["extra_flags"] = None
 
     @property
     def _is_clang_cl(self):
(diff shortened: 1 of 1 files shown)


uilianries:
@Enhex Hello! Thank you for your suggestion of expanding OpenSSL configure arguments. Please, consider the new Conan configuration for extra configure variables added in Conan 1.17.0: https://github.com/conan-io/conan/pull/18333 -

> tools.gnu:extra_configure_args: List of extra arguments to pass to configure when using AutotoolsToolchain and GnuToolchain

Please, read more about Conan configuration in the documentation: https://docs.conan.io/2/reference/config_files/global_conf.html

In this case, you can add new arguments to OpenSSL configure without touching the recipe. For instance:

```
conan install -r conancenter --requires=openssl/3.4.1 --build=openssl/3.4.1 -c tools.gnu:extra_configure_args='["no-http", "no-scrypt"]'
```

Or, you can add it to your profile as well:

```
include(default)

[conf]
openssl/* :tools.gnu:extra_configure_args=["no-http", "no-scrypt"]
```

This will be active for OpenSSL only, for any version.

However, the package ID will not be affected, which means it will override the same package with a different build. But you can change the package ID in order to follow the configuration by using a second configuration:

`-c tools.info.package_id:confs='["tools.gnu:extra_configure_args"]'`

The same can be added to your profile as well.


I commented about using this alternative because the OpenSSL recipe in CCI has +60 options available and most of them are not tested, which implies a fragile recipe with several issues. Accepting one more option like yours, will open another non-tested option by the CI, which is really hard to maintain. 

Please, feel free to share your thoughts about using the Conan configuration instead. Regards!

Enhex:
@uilianries is it possible to set `extra_configure_args` in a python recipe file?

uilianries:
> @uilianries is it possible to set `extra_configure_args` in a python recipe file?

Yes, it's possible, but I don't see much advantage as you are touching the recipe already, you can change the configure arguments directly in the recipe without needing the configuration.

Usually, Conan confs are changed in `cpp_info` but only to forward its values when consuming the recipe, for instance: https://github.com/conan-io/conan-center-index/blob/master/recipes/android-ndk/all/conanfile.py#L300

What scenario are you trying to achieve? Maybe we can explore a solution.

PR #27085, opened 2025-04-02
zlib 1.2.x: fix cmake 4 compatibility

jcar87:
Newer versions are already working


perseoGI (approved):
Tested locally, working fine with CMake 4

PR #27087, opened 2025-04-02
podofo: CMake 4 support

perseoGI:
podofo: fixes to support CMake 4

    * Removed `test_v1_package`
    * Removed this conan v1 specific code. This will never be executed as `check_min_cppstd` will throw first and it is confusing.
    ```py
     if Version(self.version) >= "0.9.7" and not valid_min_cppstd(self, 11):
         tc.cache_variables["CMAKE_CXX_STANDARD"] = 11
    ```
* Removed unreachable code reported by linter: 
```py
if conan_version.major >= 2 and Version(self.version) < "0.9.7":
```
* Removed unused patch `0001-fix-cmake-0.9.6.patch` from previously deleted version `0.9.6`
* Added missing `package_type="library"` as this recipe only packages a library (no binaries)

CMake 4 will be supported from `0.10.0`: https://github.com/podofo/podofo/blob/0.10.0/CMakeLists.txt#L1 (version not yet in CCI)


recipes/podofo/all/conanfile.py
```diff
@@ -80,15 +80,7 @@ def requirements(self):
...
-        if self.info.settings.compiler.get_safe("cppstd") and Version(self.version) >= "0.9.7":
-            check_min_cppstd(self, 11)
-        if conan_version.major >= 2 and Version(self.version) < "0.9.7":
-            # FIXME: linter complains, but function is there
-            # https://docs.conan.io/2.0/reference/tools/build.html?highlight=check_min_cppstd#conan-tools-build-check-max-cppstd
-            import sys
-            check_max_cppstd = getattr(sys.modules['conan.tools.build'], 'check_max_cppstd')
-            # INFO: error: no template named 'auto_ptr' in namespace 'std'. Removed in C++17.
-            check_max_cppstd(self, 14)
+        check_min_cppstd(self, 11)
```
jcar87, line 83: Do we not need to keep the `check_max_cppstd(self, 14)` for versiones older than `0.9.7`?

perseoGI: Not needed! The only supported version in CCI right now is `0.9.7`:
https://github.com/conan-io/conan-center-index/blob/master/recipes/podofo/config.yml?plain=1#L1

jcar87 (approved):

PR #27113, opened 2025-04-03
tinyxml2: CMake 4 support

perseoGI:
tinyxml2: fixes to support CMake 4 on older versions: `7.1.0` and `8.0.0`

Updated existing patch, which was just disabling hardcoded fPIC, and added needed changes to support CMake 4.

This changes consist on:
* Increasing `cmake_minimum_required` to 3.5
* Removing `CMP0063` OLD policy behavior (not supported by CMake 4)
* Adding the following lines which emulate the OLD behavior.

```cmake
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
set(CMAKE_VISIBILITY_INLINES_HIDDEN YES)
```

This changes have been partially brought from upstream commit:

https://github.com/leethomason/tinyxml2/commit/7df6b3ff4ab3e3cec7a242fa5f8749decd129e13


recipes/tinyxml2/all/conandata.yml
```diff
@@ -16,6 +16,12 @@ sources:
     sha256: "68ebd396a4220d5a9b5a621c6e9c66349c5cfdf5efaea3f16e3bb92e45f4e2a3"
 patches:
   "8.0.0":
-    - patch_file: "patches/0001-cmake-pic.patch"
+    - patch_file: "patches/0001-cmake4-support.patch"
+      patch_description: "Remove hardcoded fPIC, increase CMake minimum to 3.5, substitute CMP0063 OLD with new setup"
+      patch_type: "portability"
+      patch_source: "https://github.com/leethomason/tinyxml2/commit/7df6b3ff4ab3e3cec7a242fa5f8749decd129e13#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR19"
```
jcar87, line 22: Can't see much correlation between the linked patch source and the one being added in the PR?

perseoGI: Yes, you are right, the main changes extracted from the upstream patch was bumping the `cmake_minimum_required` to 3.5, removing the OLD policy and adding:
```cmake
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
set(CMAKE_VISIBILITY_INLINES_HIDDEN YES)
```

Maybe we can delete the `patch_source` tag to avoid further confusion.

jcar87, line 22:
```suggestion
```

recipes/tinyxml2/all/conandata.yml
```diff
@@ -16,6 +16,11 @@ sources:
...
-    - patch_file: "patches/0001-cmake-pic.patch"
+    - patch_file: "patches/0001-cmake4-support.patch"
+      patch_description: "Remove hardcoded fPIC, increase CMake minimum to 3.5, substitute CMP0063 OLD with new setup"
+      patch_type: "portability"
   "7.1.0":
-    - patch_file: "patches/0001-cmake-pic.patch"
+    - patch_file: "patches/0001-cmake4-support.patch"
+      patch_description: "Remove hardcoded fPIC, increase CMake minimum to 3.5, substitute CMP0063 OLD with new setup"
+      patch_type: "portability"
+      patch_source: "https://github.com/leethomason/tinyxml2/commit/7df6b3ff4ab3e3cec7a242fa5f8749decd129e13#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR19"
```
jcar87, line 26:
```suggestion
```

recipes/tinyxml2/all/patches/0001-cmake4-support.patch
```diff
@@ -0,0 +1,32 @@
...
+ ENDIF(BIICODE)
+-cmake_minimum_required(VERSION 2.6 FATAL_ERROR)
++cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+ cmake_policy(VERSION 2.6)
+ if(POLICY CMP0063)
+-	cmake_policy(SET CMP0063 OLD)
++	# cmake_policy(SET CMP0063 OLD)
+ endif()
+ 
++set(CMAKE_CXX_VISIBILITY_PRESET hidden)
```
jcar87, line 17: I may be missing something, why are these two lines being "moved" here, when they are already present in the CMakeLists? (they are the last two lines of this very patch) - I can't find any justification for moving them, as they should have the same in terms of the produced binaries? (perhaps it influences the compiler checks done by `project()`, if anything)

Given that this is versions 8.0.0 and 7.0.0 from 5+ years ago, if the newer versions already support CMake 4, I would be inclined to just not patch them at all.

jcar87 (approved):

PR #27133, opened 2025-04-03
mariadb-connector-cpp: CMake 4 support on non apple systems

perseoGI:
spirv-tools: fixes to support CMake 4

Increase CMake minimum required to 3.23 on non apple systems, as upstream already support CMake 4 for `APPLE` targets:
See https://github.com/mariadb-corporation/mariadb-connector-cpp/blob/master/CMakeLists.txt#L22


jcar87:
This recipe is not affected:

```
    def build_requirements(self):
        self.tool_requires("cmake/[>=3.23 <4]")
```

PR #27158, opened 2025-04-07
lua: CMake 4 support

perseoGI:
lua: fixes to support CMake 4

### DIFF
--- a/recipes/lua/all/CMakeLists.txt
+++ b/recipes/lua/all/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.15)
 project(lua)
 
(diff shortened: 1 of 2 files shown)


AbrilRBS (approved):

recipes/lua/all/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.5)
```
jcar87, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15)
```

if the CMakeLists is in _our_ repo, let's fix this to 3.15 (the minimum supported by Conan 2)

jcar87 (approved):

Issue #27152, opened 2025-04-05
[package] libcurl/8.12.1: cannot build with IDN support on Windows (MSVC)


zenden2k:
### Description

Building the libcurl package with IDN (International Domain Names) support is not possible when using MSVC.
The `with_libidn` option does not work.
It produces an error.
```
-- Could NOT find Libidn2 (missing: LIBIDN2_INCLUDE_DIR LIBIDN2_LIBRARY)
```
Additionally, there is no option to enable WinIDN (which corresponds to the USE_WIN32_IDN setting in libcurl's CMakeLists.txt).

### Package and Environment Details

* Package Name/Version: **libcurl/8.12.1**
* Operating System+version: **Windows 11 24H2**
* Compiler+version: **Visual Studio 2019** (C/C++ Compiler 19.29.30159)
* Conan version: **conan 2.11.0**
* Python version: **Python 3.12.4**

### Conan profile

[settings]
os=Windows
arch=x86
compiler=msvc
compiler.version=192
compiler.cppstd=17
compiler.runtime=static
build_type=Debug

### Steps to reproduce

conan install --build=missing

`conanfile.py`:

```
from conan import ConanFile

class TestRecipe(ConanFile):
    settings = "os", "compiler", "build_type", "arch"
    generators = "CMakeToolchain", "CMakeDeps"
    default_options = {
        "libcurl/*:with_libidn": True
    }
    def requirements(self):
        self.requires("libcurl/8.12.1")
```

### Logs

<details><summary>Click to expand log</summary>

```

======== Input profiles ========
Profile host:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows


======== Computing dependency graph ========
Graph root
    conanfile.py: D:\Test\testpackage\conanfile.py
Requirements
    libcurl/8.12.1#722cbde4f18cb5824bdc4d4f827b0ec6 - Cache
    openssl/3.3.2#9f9f130d58e7c13e76bb8a559f0a6a8b - Cache
    zlib/1.3.1#b8bc2603263cf7eccbd6e17e66b0ed76 - Cache
Build requirements
[...]

PR #27162, opened 2025-04-07
Libcurl: libidn fixes

AbrilRBS:
Fixes all libidn usages of libcurl, it's now consumed properly

This option is on by default, but in the spirit of not breaking potential users that expect this to be off by default, I've left the default in the recipe untouched.

I'll submit compilation reports in a bit

Closes https://github.com/conan-io/conan-center-index/issues/27152

### DIFF
--- a/recipes/libcurl/all/conanfile.py
+++ b/recipes/libcurl/all/conanfile.py
@@ -129,12 +129,8 @@ class LibcurlConan(ConanFile):
         return self.settings.os == "Windows" and self.settings.compiler == "gcc"
 
     @property
-    def _settings_build(self):
-        return getattr(self, "settings_build", self.settings)
-
-    @property
     def _is_win_x_android(self):
-        return self.settings.os == "Android" and self._settings_build.os == "Windows"
+        return self.settings.os == "Android" and self.settings_build.os == "Windows"
 
     @property
     def _is_using_cmake_build(self):
@@ -192,6 +188,9 @@ class LibcurlConan(ConanFile):
             self.requires("c-ares/[>=1.27 <2]")
         if self.options.get_safe("with_libpsl"):
             self.requires("libpsl/0.21.1")
+        if self.options.with_libidn:
+            self.requires("libidn2/2.3.0")
+
 
     def validate(self):
         if self.options.with_ssl == "schannel" and self.settings.os != "Windows":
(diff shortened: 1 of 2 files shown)


recipes/libcurl/all/conanfile.py
```diff
@@ -620,6 +625,7 @@ def _generate_with_cmake(self):
         tc.variables["ENABLE_UNICODE"] = True
         tc.variables["BUILD_TESTING"] = False
         tc.variables["BUILD_CURL_EXE"] = False
+        tc.cache_variables["ENABLE_CURL_MANUAL"] = False
```
AbrilRBS, line 628: This was looking for Perl on my system

recipes/libcurl/all/conanfile.py
```diff
@@ -714,16 +724,14 @@ def package_info(self):
         else:
             self.cpp_info.components["curl"].libs = ["curl"]
             if self.settings.os in ["Linux", "FreeBSD"]:
-                if self.options.with_libidn:
```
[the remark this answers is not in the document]
AbrilRBS: When building as True in Linux:

```
CMake Error at build/gcc-13-x86_64-gnu17-release/generators/cmakedeps_macros.cmake:66 (message):
  Library 'idn' not found in package.  If 'idn' is a system library, declare
  it with 'cpp_info.system_libs' property
  build/gcc-13-x86_64-gnu17-release/generators/CURL-Target-release.cmake:23 (conan_package_library_targets)
  build/gcc-13-x86_64-gnu17-release/generators/CURLTargets.cmake:24 (include)
  build/gcc-13-x86_64-gnu17-release/generators/CURLConfig.cmake:16 (include)
  CMakeLists.txt:4 (find_package)
```

because in fact, as we can see:

```
# ls -l /root/.conan2/p/b/libcu843b2b6692264/p/lib
total 1580
-rw-r--r-- 1 root root 1615910 Apr  7 16:08 libcurl.a
```

So removing this is the right approach (And probably the other option too, but it's left as a follow-up PR)

recipes/libcurl/all/test_package/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.5)
```
jcar87, line 1:
```suggestion
cmake_minimum_required(VERSION 3.15)
```

jcar87 (approved):

zenden2k:
Unfortunately, libidn2 doesn't compile with MSVC :(

```
.\limits.h(36): fatal error C1083: Cannot open include file: '': No such file or directory
```

I'll have to create a new issue, this time for the libidn2 package...

PR #27164, opened 2025-04-07
hictk: add v2.1.0

robomics:
### Summary
Changes to recipe:  **hictk/2.1.0**

#### Details

---

### DIFF
--- a/recipes/hictk/all/conanfile.py
+++ b/recipes/hictk/all/conanfile.py
@@ -100,6 +100,7 @@ class HictkConan(ConanFile):
         tc.variables["HICTK_BUILD_EXAMPLES"] = "OFF"
         tc.variables["HICTK_BUILD_TOOLS"] = "OFF"
         tc.variables["HICTK_ENABLE_GIT_VERSION_TRACKING"] = "OFF"
+        tc.variables["HICTK_ENABLE_TELEMETRY"] = "OFF"
         tc.variables["HICTK_ENABLE_TESTING"] = "OFF"
         tc.variables["HICTK_ENABLE_FUZZY_TESTING"] = "OFF"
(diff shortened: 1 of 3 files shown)


recipes/hictk/all/conanfile.py
```diff
@@ -100,6 +100,7 @@ def generate(self):
         tc.variables["HICTK_BUILD_EXAMPLES"] = "OFF"
         tc.variables["HICTK_BUILD_TOOLS"] = "OFF"
         tc.variables["HICTK_ENABLE_GIT_VERSION_TRACKING"] = "OFF"
+        tc.variables["HICTK_ENABLE_TELEMETRY"] = "OFF"
```
jcar87, line 103: https://github.com/paulsengroup/hictk/blob/53f6a6b1907b9e0d684fa6041645a3bd160b3c76/CMakeLists.txt#L175-L181

This variable is only used if `HICTK_BUILD_TOOLS` is True, but it is unconditionally set to off two lines above, so this wouldn't be necessary (and the documentation specifically mentions that it would be the executable, not the library, that has telemetry)

robomics: You are correct.

There are no plans to enable telemetry at the library level for the forseeable future, however I thought to explicitly define `HICTK_ENABLE_TELEMETRY` just to avoid problems in the future in case we change our mind (or in case we decide to update the Conan package to build the executable as well).

Let me know if you still prefer to leave the `HICTK_ENABLE_TELEMETRY` undefined in the Conan recipe.

robomics:
Converting to draft for now, as later today I will tag a new release to include this bugfix: https://github.com/paulsengroup/hictk/pull/399.

robomics:
@jcar87 can I get a review please :)

PR #27173, opened 2025-04-08
gsoap: add v2.8.139

NelDav:
### Summary
Changes to recipe:  **gsoap/2.8.138**

#### Motivation

New upstream version

#### Details

---

### DIFF
--- a/recipes/gsoap/all/conandata.yml
+++ b/recipes/gsoap/all/conandata.yml
@@ -7,6 +7,11 @@
 # At some point the Conan client might come pre-configured with this backup remote by default, but don't delete the links then
 # some users might still rely on the old revisions/not have it set to point to our CCI backup remote
 sources:
+  "2.8.138":
+    url:
+      - "https://sourceforge.net/projects/gsoap2/files/gsoap_2.8.138.zip/download"
+      - "https://c3i.jfrog.io/artifactory/conan-center-backup-sources/5ddd63eebd6a08e1838e93400a843975c65ae77e09a95276f9adf166d3ca6f74"
+    sha256: "5ddd63eebd6a08e1838e93400a843975c65ae77e09a95276f9adf166d3ca6f74"
   "2.8.134":
     url:
       - "https://sourceforge.net/projects/gsoap2/files/gsoap_2.8.134.zip/download"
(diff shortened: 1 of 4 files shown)


recipes/gsoap/all/conandata.yml
```diff
@@ -7,6 +7,11 @@
 # At some point the Conan client might come pre-configured with this backup remote by default, but don't delete the links then
 # some users might still rely on the old revisions/not have it set to point to our CCI backup remote
 sources:
+  "2.8.138":
+    url:
+      - "https://sourceforge.net/projects/gsoap2/files/gsoap_2.8.138.zip/download"
+      - "https://c3i.jfrog.io/artifactory/conan-center-backup-sources/5ddd63eebd6a08e1838e93400a843975c65ae77e09a95276f9adf166d3ca6f74"
```
jcar87, line 13: Please remove this as a mirror - it's not an official mirror, and the documented way to use it is transparently without modifying the recipes, thanks!

NelDav: Thanks for you review. I just followed the instruction at the top of the conandata file:

> ```
> # The sources for this project get deleted from time to time.
> # We use the backup sources from Conan Center as mirrors to let users compile in the future
> # When adding new versions, directly add the URL following previous versions schema
> # https://c3i.jfrog.io/artifactory/conan-center-backup-sources/ + sha256 of the sources
> # The file will be automatically uploaded after the first successful compilation of the new version
> # And will be usable as a mirror from then on.
> # At some point the Conan client might come pre-configured with this backup remote by default, but don't delete the links then
> # some users might still rely on the old revisions/not have it set to point to our CCI backup remote
> ```

I do not know enough about conan to decide if it's a good idea to do that or not. However I think it should be consistent within the file. Meaning if we remove the `conan-center-backup-sources` url from one version, it should be removed from all versions. For now I prefer following the instructions and let somebody else (in annother PR) decide if all those `conan-center-backup-source` url's should be removed or not.

jcar87: Apologies for the confusion @NelDav  - we'll handle this (the backups remote is not intended to be used like this, there was a miscommunication within the team)

jcar87:
Thanks!

NelDav:
Thank you guys for taking over and finishing this PR 😊

PR #27185, opened 2025-04-08
[boost] math: 1.87.0 no longer builds long double variants on Emscripten

rafzi:
### Summary
Changes to recipe:  **boost/1.87.0+**

#### Motivation
Boost fails to build on Conan since 1.87.0 on Emscripten with:

    boost/1.87.0: WARN: Boost component 'math_c99l' is missing libraries. Try building boost with '-o boost:without_math_c99l'. (Option is not guaranteed to exist)
    boost/1.87.0: WARN: Boost component 'math_tr1l' is missing libraries. Try building boost with '-o boost:without_math_tr1l'. (Option is not guaranteed to exist)
    ERROR: boost/1.87.0: Error in package_info() method, line 2011
     	raise ConanException(f"These libraries were expected to be built, but were not built: {non_built}")
     	ConanException: These libraries were expected to be built, but were not built: {'boost_math_c99l', 'boost_math_tr1l'}

The [boost math patch notes](https://github.com/boostorg/math/releases/tag/boost-1.87.0) mention: "fix: define BOOST_MATH_NO_LONG_DOUBLE_MATH_FUNCTIONS for emscripten": https://github.com/boostorg/math/pull/1216 . So it makes sense that these libraries are no longer produced.

#### Details
The fix is to no longer expect the long double variants as we already do for PPC architectures.

A minor follow-up fix was to avoid passing "--disable-new-dtags" to the linker, because that flag is not recognized by the Emscripten linker.

---

### DIFF
--- a/recipes/boost/all/conanfile.py
+++ b/recipes/boost/all/conanfile.py
@@ -1929,7 +1929,7 @@ class BoostConan(ConanFile):
                 for name in names:
                     if name in ("boost_stacktrace_windbg", "boost_stacktrace_windbg_cached") and self.settings.os != "Windows":
                         continue
-                    if name in ("boost_math_c99l", "boost_math_tr1l") and str(self.settings.arch).startswith("ppc"):
+                    if name in ("boost_math_c99l", "boost_math_tr1l") and (str(self.settings.arch).startswith("ppc") or (Version(self.version) >= "1.87.0" and self.settings.os == "Emscripten")):
                         continue
                     if name in ("boost_stacktrace_addr2line", "boost_stacktrace_backtrace", "boost_stacktrace_basic") and self.settings.os == "Windows":
                         continue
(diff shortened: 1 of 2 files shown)


uilianries (approved):
@rafzi Hello :wave: Thank you for your PR! :smile: 

I'm able to reproduce your reported case, here is my build log using the master branch with emscripten: [boost-1.87.0-emscripten-math-error.log](https://github.com/user-attachments/files/20099199/boost-1.87.0-emscripten-math-error.log)

Indeed there are good evidences about your case:

- https://github.com/boostorg/math/blob/boost-1.87.0/doc/html/math_toolkit/compilers_overview.html#L330
- https://github.com/boostorg/math/blob/boost-1.87.0/include/boost/math/tr1.hpp#L77

After using your patch, the error is no longer present: [boost-1.87.0-emscripten-math-patched.log](https://github.com/user-attachments/files/20099224/boost-1.87.0-emscripten-math-patched.log)

LGTM

PR #27186, opened 2025-04-08
wasm-micro-runtime: add `disable_hw_bound_check` and `disable_write_gs_base`

### LINKED ISSUE #27260: wasm-micro-runtime/2.2.0
My project requires certain flags to be set during CMake configuration that are not options available through conan. Specifically, I need to set `WAMR_DISABLE_HW_BOUND_CHECK` and `WAMR_DISABLE_WRITE_GS_BASE`.

After speaking to some kind folks over in cppslack, I am using the following workaround:
``` 
[conf]
tools.cmake.cmaketoolchain:extra_variables={"WAMR_DISABLE_HW_BOUND_CHECK": "1", "WAMR_DISABLE_WRITE_GS_BASE": "1"}
```

Can we add the ability to set these flags through `conanfile.txt`?

sgerbino:
Resolves #27260.

### Summary
Changes to recipe:  **wasm-micro-runtime**
Adds the ability to set CMake options `WAMR_DISABLE_HW_BOUND_CHECK` and `WAMR_DISABLE_WRITE_GS_BASE` using Conan options. These variables are normally determined at configuration time by the project. This PR allows the user to specify them explicitly.

#### Motivation
There are reasons a consumer of this library would want to explicitly set these build options. For example, performance benefits or compatibility with other tools like `valgrind`.

#### Details
When the options are unset, the build should behave as normal and allow the project to determine the flag setting. When the library consumer sets them, it should be passed to the CMake configuration stage.

---

### DIFF
--- a/recipes/wasm-micro-runtime/all/conanfile.py
+++ b/recipes/wasm-micro-runtime/all/conanfile.py
@@ -32,6 +32,8 @@ class WasmMicroRuntimeConan(ConanFile):
         "with_wasi_threads": [True, False],
         "with_tail_call": [True, False],
         "with_simd": [True, False],
+        "disable_hw_bound_check": [None, True, False],
+        "disable_write_gs_base": [None, True, False],
     }
     default_options = {
         "shared": False,
@@ -47,6 +49,8 @@ class WasmMicroRuntimeConan(ConanFile):
         "with_wasi_threads": False,
         "with_tail_call": False,
         "with_simd": True,
+        "disable_hw_bound_check": None,
+        "disable_write_gs_base": None,
     }
 
     def export_sources(self):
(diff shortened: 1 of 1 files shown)


recipes/wasm-micro-runtime/all/conanfile.py
```diff
@@ -47,6 +49,8 @@ class WasmMicroRuntimeConan(ConanFile):
         "with_wasi_threads": False,
         "with_tail_call": False,
         "with_simd": True,
+        "disable_hw_bound_check": None,
```
jcar87, line 52: Thanks! This is actually one of the few examples that I can see where a `None` option is properly implemented to mean "unless explicitly defined, let the underlying build system decide" 

However, looking at the upstream build scripts, it looks like hw bounds check are by default only for 32-bit platforms? if it is that clear cut, we could have `True` and `False` as the valid option values, _not_ give it a value in `default_options`, and then conditionally set it it to true or False in  `config_options()` based on the `self.settings.arch` values... what do you think?

franramirez688:
Closing this due to inactivity.

Please, do not hesitate to reopen this if you want to resume this PR 😁

PR #27198, opened 2025-04-09
Fix termcap settings

rbu9fe:
### Summary
Changes to recipe:  **termcap/1.3.1**

#### Motivation
This PR makes the recipe comply with the termcap's intended default behaviour:
1. By default install the termcap database into `/etc/termcap`.
2. By default don't `#define TERMCAP_FILE` so that the termcap database is searched under its default location `/etc/termcap`.

#### Details
This PR modifies package settings where by default the termcap database is installed and searched for so that it complied to the default settings that the termcap source code is using.

---

### DIFF
--- a/recipes/termcap/all/CMakeLists.txt
+++ b/recipes/termcap/all/CMakeLists.txt
@@ -21,7 +21,6 @@ endforeach()
 
 add_definitions(-DSTDC_HEADERS)
 add_definitions(-DTERMCAP_BUILDING)
-add_definitions("-DTERMCAP_FILE=\"${CMAKE_INSTALL_SYSCONFDIR}/termcap\"")
 if(BUILD_SHARED_LIBS)
     add_definitions(-DTERMCAP_SHARED)
 endif()
--- a/recipes/termcap/all/conanfile.py
+++ b/recipes/termcap/all/conanfile.py
@@ -86,7 +86,7 @@ class TermcapConan(ConanFile):
 
     @property
     def _termcap_path(self):
-        return os.path.join(self.package_folder, "bin", "etc", "termcap")
+        return os.path.join(self.package_folder, "etc", "termcap")
 
     def package_info(self):
         self.cpp_info.libs = ["termcap"]
--- a/recipes/termcap/all/conanfile.py
+++ b/recipes/termcap/all/conanfile.py
@@ -61,7 +61,6 @@ class TermcapConan(ConanFile):
         tc.cache_variables["TERMCAP_HEADERS"] = to_cmake_paths(headers)
         tc.cache_variables["TERMCAP_INC_OPTS"] = to_cmake_paths(optional_headers)
         tc.cache_variables["TERMCAP_CAP_FILE"] = os.path.join(self.source_folder, "termcap.src").replace("\\", "/")
-        tc.cache_variables["CMAKE_INSTALL_SYSCONFDIR"] = os.path.join(self.package_folder, "bin", "etc").replace("\\", "/")
         tc.generate()
 
     def _patch_sources(self):


jcar87:
Hi @rbu9fe  - thanks for your PR.

However I'm not entirely sure about the consequences of these changes

> By default install the termcap database into /etc/termcap.

Conan packages do not (and should not) write to system locations - and from what I can see in your PR, what you are changing is the directory _inside_ the package.. which is fine (arbitrary, but okay)

If your goal is to default the database to be look at `/etc/termcap` - I guess that's fine, but that means the one in the package would be ignored completely (and thus, it being part of the package folder is irrelevant, may as well not be there) - on the other hand, that is already the case when the package is built on one machine and used in another

Given that termcap is.. self-declared as obsolete, I'm wondering what is the usage motivation behind this change? so that we can understand why termcap is a recipe in Conan Center to begin with...

rbu9fe:
Thanks for your quick review. First of all I'm using termcap not directly but as a dependency of the [readline](https://conan.io/center/recipes/readline?version=8.2) library. I'm not aware that it's considered obsolete but since I'm using readline I'm also stuck with termcap (unless that can be configured to exclude it, I didn't check that).

With the existing recipe, when the package is installed to the final system, then (unless overwritten by a post-install hook) the termcap database is installed into `/bin/etc/termcap`. However, the major issue is the definition of
``` cmake
add_definitions("-DTERMCAP_FILE=\"${CMAKE_INSTALL_SYSCONFDIR}/termcap\"")
```
in the [cmake file](https://github.com/conan-io/conan-center-index/pull/27198/files#diff-ed18f6501ed1e674106688d5d731b5ab1119bfeb242ced8b677880f6c5773915L24). This expands to some path inside the Conan package since that's how `CMAKE_INSTALL_SYSCONFDIR` is defined in the [conanfile](https://github.com/conan-io/conan-center-index/pull/27198/files#diff-44b6fa186551f9bc1685d5673021fe6eb38caed16cb3b5d9f06612bbdc957a91L64):
```
tc.cache_variables["CMAKE_INSTALL_SYSCONFDIR"] = os.path.join(self.package_folder, "bin", "etc").replace("\\", "/")
```
The definition of `TERMCAP_FILE` indicates the default path where the termcap library is looking for the database. However, when the package (including the library and the database) is installed into a system, then obviously the defined `TERMCAP_FILE` isn't valid at runtime since it points to some package folder of the build machine inside the conan cache that doesn't exist on the target system. So either the definition of `TERMCAP_FILE` must be removed to rely on its default definition of
``` C
#ifndef TERMCAP_FILE
#define TERMCAP_FILE "/etc/termcap"
#endif
```
in the termcap library API, or it must at least point to some valid system path (preferably the same where the database is installed by default).

Additionally, when the package is installed in the final system, I think it's a good idea to install the database to the location where the library expects it by default. Therefore, I removed overwriting `CMAKE_INSTALL_SYSCONFDIR` [here](https://github.com/conan-io/conan-center-index/pull/27198/files#diff-44b6fa186551f9bc1685d5673021fe6eb38caed16cb3b5d9f06612bbdc957a91L64) and relay on its default value of `/etc` instead, and adjusted `_termcap_path` accordingly.

So to summarise:
1. The definition of `TERMCAP_FILE` _MUST_
    - either be adjusted to exclude the Conan cache package path so that it points to a potentially valid path in the system, or
    - be removed entirely so that the library looks for the database under its default location `/etc/termcap`.
 2. The location where the database is installed when the package is installed on the final system _SHOULD_ be `/etc/termcap` rather than `/bin/etc/termcap`.

rbu9fe:
Update: readline can indeed be configured to use ncurses instead of termcap, however, the ncurses recipe currently doesn't support cross-compilation to/from arm, which I need. Therefore, I'm stuck with termcap.

uilianries:
@rbu9fe Hello!

As an alternative, you still can override `CMAKE_INSTALL_SYSCONFDIR` value and install termcap whatever you need:

```
conan create all --version=1.3.1 -c tools.cmake.cmaketoolchain:extra_variables='{"CMAKE_INSTALL_SYSCONFDIR": "/home/user/.config/tercap"}' 
```

This configuration can be set to `/etc/` as well, but as you should know, it will require permission as super user. 

Still, only this configuration alone will not affect the package ID of your built package. In order to change its package ID based on `extra_variables`, you will need to use the configuration:

```
tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]
```

Regards!

rbu9fe:
Of course I can configure `CMAKE_INSTALL_SYSCONFDIR` accordingly but then it doesn't match the path that's reported by `_termcap_path`. Isn't that a problem?

uilianries:
Hello! After revising this PR and doing better research about the case, I think it would be better to update the `readline` recipe instead:

- Termcap is indeed obsolete and replaced by Terminfo: https://blog.netbsd.org/tnf/entry/terminfo_has_replaced_termcap. Also, marked it as obsolete in 2017.
- Other packages managers like AUR, Conda, Homebrew are using only curses as dependency, but no option for termcap.

PR #27200, opened 2025-04-10
openvino: add version 2025.1.0

culhatsker:
### Summary
Changes to recipe:  **openvino/2025.1.0**

#### Motivation
New release

#### Details
Backport patch https://github.com/openvinotoolkit/openvino/pull/30157

---


recipes/openvino/all/conanfile.py
```diff
@@ -162,6 +165,7 @@ def build_requirements(self):
     def requirements(self):
         self.requires("onetbb/2021.10.0")
         self.requires("pugixml/1.14")
+        self.requires("nlohmann_json/3.11.3")
```
ErniGH, line 168: Why is `nlohmann_json` needed now? Will it break with previous versions?

culhatsker: With new `openvino` release one of the components needs it now. 
I selected `v3.11.3` as it is the version we use, it may work with earlier `nlohmann_json` versions, but I can't tell for sure it will work with earlier ones.
Having this requirement when building `openvino<=2025.1` doesn't break anything so I decided to keep it simpler.

recipes/openvino/all/patches/2025.1.0/include_paths.patch
ErniGH, line 1: What is the origin of this patch? Is it related to any upstream commit, or what is it needed for?

culhatsker: I created a PR to the original repository.
There's some piece of code in cmake configuration that was supposed to fix something in vcpkg build but also broke building with conan. It was corrupting some template expressions that conan left in include directories.

culhatsker: @ErniGH is there something you're waiting for? Should I explain in details this patch?

ErniGH: It would be better to wait to see this patch merged by upstream before merging this PR.

culhatsker: @ErniGH that PR has been accepted.

AbrilRBS (approved):
Thanks! Let's remember to remove this patch for 2025.2 :)

ErniGH (approved):
Thanks! LGTM

PR #27207, opened 2025-04-10
grpc: Add frameworks only when on apple os

AbrilRBS:
This is needed as the new CMakeConfigDeps generator does not expect frameworks in non-apple OS,

as per @memsharded:

> The generator assumes that dependencies do not add frameworks information for other systems different than Apple, in the same way that dependencies shouldn't add any libraries, flags, pre-processor definitions, etc that don't work for other systems.

Closes https://github.com/conan-io/conan-center-index/issues/27206

### DIFF
--- a/recipes/grpc/all/conanfile.py
+++ b/recipes/grpc/all/conanfile.py
@@ -143,7 +143,7 @@ class GrpcConan(ConanFile):
 
     def build_requirements(self):
         # cmake >=3.25 required to use `cmake -E env --modify` below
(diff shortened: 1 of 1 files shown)


memsharded (approved):
@franramirez688 and I did a bit of research in ConanCenter, and it looked like that all ``cpp_info.frameworks`` definition were conditioned to ``is_apple_os()``. It is possible that some packages that rely on an external definition ``values``, won't have the values to account for it.

franramirez688 (approved):
👍

PR #27209, opened 2025-04-10
cnats: add version 3.10.1

wooffie:
### Summary
Changes to recipe:  **cnats/3.10.1**

#### Motivation
Nats.c introduced release 3.10.0, but there was problem about cmake definition, now it is fixed (https://github.com/nats-io/nats.c/pull/860) and releases in 3.10.1. 

#### Details
I just add option for experimental API and turn off it by default. I think one CMAKE option and C definition for older versions no make sense. Of course, I think you can advice more interesting solutions (I thought about using self.version for disabling definition for older packages version, but I think it is should make recipe more readable).

P.S. add trailing comma for default opts =)

I read about Experimental - it will bail off if TLS if off, but this flag for all unstable API, and I believe we shouldn't link it with TLS option

---


ErniGH (approved):
LGTM

PR #27210, opened 2025-04-10
ohnet: Update to 1.38

DoomHammer:
### Summary
Changes to recipe:  **ohnet/1.38**

#### Details

---

### DIFF
--- a/recipes/ohnet/all/conanfile.py
+++ b/recipes/ohnet/all/conanfile.py
@@ -79,7 +79,7 @@ class OhNetConan(ConanFile):
             args = self._fill_openhome_architecture(args)
             args.append("rsync=no")
             if str(self.settings.compiler.libcxx) == "libc++":
-                args.extend(["CPPFLAGS=-stdlib=libc++", "LDFLAGS=-stdlib=libc++"])
+                args.extend(["CPPFLAGS=-stdlib=libc++", "LDFLAGS=-stdlib=libc++", "PYTHON=python3"])
         return args
 
     def build(self):
--- a/recipes/ohnet/all/patches/0001-makefile-1.38.patch
+++ b/recipes/ohnet/all/patches/0001-makefile-1.38.patch
@@ -1,0 +1,491 @@
+diff --git a/Common.mak b/Common.mak
+index baae7b17..73eb1829 100644
+--- a/Common.mak
++++ b/Common.mak
+@@ -251,23 +251,23 @@ $(objdir)AsyncPrivate.$(objext) : OpenHome/Net/ControlPoint/AsyncPrivate.cpp $(h
+ 	$(compiler)AsyncPrivate.$(objext) -c $(cppflags) $(includes) OpenHome/Net/ControlPoint/AsyncPrivate.cpp
+ $(objdir)Bonjour.$(objext) : OpenHome/Net/Device/Bonjour/Bonjour.cpp $(headers)
+ 	$(compiler)Bonjour.$(objext) -c $(cppflags) $(includes) OpenHome/Net/Device/Bonjour/Bonjour.cpp
+ $(objdir)Buffer.$(objext) : OpenHome/Buffer.cpp $(headers)
+ 	$(compiler)Buffer.$(objext) -c $(cppflags) $(includes) OpenHome/Buffer.cpp
+ $(objdir)Converter.$(objext) : OpenHome/Converter.cpp $(headers)
+ 	$(compiler)Converter.$(objext) -c $(cppflags) $(includes) OpenHome/Converter.cpp
+ $(objdir)Crc32.$(objext) : OpenHome/Crc32.cpp $(headers)
+ 	$(compiler)Crc32.$(objext) -c $(cppflags) $(includes) OpenHome/Crc32.cpp
+ $(objdir)cencode.$(objext) : thirdparty/libb64/cencode.c $(headers)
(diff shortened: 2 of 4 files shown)


recipes/ohpipeline/all/conanfile.py
```diff
@@ -48,7 +48,7 @@ def build_requirements(self):
         self.tool_requires("cmake/[>3.23 <4]")
 
     def requirements(self):
-        self.requires("ohnet/1.37.5454", transitive_headers=True, transitive_libs=True)
```
jcar87, line 51: Please limit the PR to one recipe, thanks!

recipes/ohnet/all/patches/0001-makefile-1.38.patch
```diff
@@ -0,0 +1,491 @@
...
+ 	$(compiler)Http.$(objext) -c $(cppflags) $(includes) OpenHome/Http.cpp
+ $(objdir)anonymous.$(objext) : $(mDNSdir)/anonymous.c $(headers)
+-	$(compiler)anonymous.$(objext) -c $(cflags_third_party) $(includes) $(mDNSdir)/anonymous.c
++	$(compiler_cc)anonymous.$(objext) -c $(cflags_third_party) $(includes) $(mDNSdir)/anonymous.c
+ $(objdir)CryptoAlg.$(objext) : $(mDNSdir)/CryptoAlg.c $(headers)
+-	$(compiler)CryptoAlg.$(objext) -c $(cflags_third_party) $(includes) $(mDNSdir)/CryptoAlg.c
++	$(compiler_cc)CryptoAlg.$(objext) -c $(cflags_third_party) $(includes) $(mDNSdir)/CryptoAlg.c
+ $(objdir)DNSCommon.$(objext) : $(mDNSdir)/DNSCommon.c $(headers)
+-	$(compiler)DNSCommon.$(objext) -c $(cflags_third_party) $(includes) $(mDNSdir)/DNSCommon.c
++	$(compiler_cc)DNSCommon.$(objext) -c $(cflags_third_party) $(includes) $(mDNSdir)/DNSCommon.c
```
ErniGH, line 49: I need help to understand all the new changes in this patch. Could you please help me to understand the source of this? Is there any issue related to this in upstream?

All the changes on `Common.mak` are new

recipes/ohnet/all/conanfile.py
```diff
@@ -79,7 +79,7 @@ def _fill_ohnet_args(self, args):
             args = self._fill_openhome_architecture(args)
             args.append("rsync=no")
             if str(self.settings.compiler.libcxx) == "libc++":
-                args.extend(["CPPFLAGS=-stdlib=libc++", "LDFLAGS=-stdlib=libc++"])
+                args.extend(["CPPFLAGS=-stdlib=libc++", "LDFLAGS=-stdlib=libc++", "PYTHON=python3"])
```
ErniGH, line 82: Why does this flag now need it?

ErniGH (commented):
Here is the diff between versions
https://github.com/openhome/ohNet/compare/ohNet_1.37.5454...ohNet_1.38.5707

recipes/ohnet/all/patches/0001-makefile-1.38.patch
DoomHammer: Upstream does not respond to any issues so it is useless to file them there, unfortunately.

recipes/ohnet/all/conanfile.py
DoomHammer: The original `make` script calls `python` which on many systems mean `python2` as contrasted to `python3`. As you might imagine, I'm not a fan of installing Python 2 on people's machines.

recipes/ohnet/all/patches/0001-makefile-1.38.patch
ErniGH: The changes introduced by the patch should be backed by merged upstream PRs that address specific issues or be strictly necessary modifications to ensure compatibility with Conan, in which case, it would also be helpful to provide a clear justification.

DoomHammer: This is strictly necessary to build this piece of software at all. And upstream is not responding. We've been through this before with previous versions.

jcar87:
> This is strictly necessary to build this piece of software at all. And upstream is not responding. We've been through this before with previous versions.

Unfortunately the scope of the proposed patching in this recipe are outside of what we're comfortable reviewing. Our overarching goal with patches is that they are not needed in the future, and if there is some defect in the upstream code that could be remedied, we very much do require this is reported upstream such that in the future, in a new version, things can be easier to maintain. If that is not the case, and the upstream authors do not respond to issues, then perhaps Conan Center is not the best place for this recipe to exist (and from what I can see, it is not published in any other repository).

The patches proposed in this PR make non-trivial changes to low-level makefiles - I'd only be comfortable merging this PR if we had multiple reviewers *really* sure that the changes are sound and do not introduce bugs or security vulnerabilities (the `xz-utils` attack happened partly because changes to autotools/makefiles are not particularly easy to review). 

Given that every new version seems to require this level of invasive patching, and that this is not packaged anywhere else except in Conan Center (as per repology), and that there doesn't seem to be any prospects of having the issues reported and fixed upstream, I would consider this recipe no longer maintained in Conan Center. This is one of those cases where if it was a new recipe proposal, it would probably not be included at all.

PR #27213, opened 2025-04-11
emsdk: recipe maintenance and latest version

### LINKED ISSUE #14350: [package] emsdk/3.23.0: Allow usage without bundled nodejs
### Description

emsdk hard requires nodejs so it doesn't use the system one.

https://github.com/conan-io/conan-center-index/blob/7154dbe7cf2618faeff1804730978c99651ee145/recipes/emsdk/all/conanfile.py#L31-L34

@werto87 @jgsogo 
Should there be an option to allow the system nodejs version?

See my depending issue, why I ask this: https://github.com/conan-io/conan-center-index/issues/14349

### LINKED ISSUE #26067: [bug] Conflict between profile tool_require and recipe require
### Describe the bug

When using a profile with a tool_requires, I am getting a conflict between some dependency of the tool and a dependency of the recipe

conan install command
```
 conan install . --profile:build=conan/profiles/x86_64-gcc-10 --profile:host=conan/profiles/wasm-clang-16 --settings:host="*:build_type=Release" --settings:host="&:build_type=Debug" --build="*"
```

throws 

```
ERROR: Version conflict: Conflict between nodejs/20.16.0 and nodejs/16.3.0 in the graph.
Conflict originates from boost/1.82.0
```

According to discussion in Slack channel, the dependency in the profile should override others, but in this case still fails. 

conanfile.py
```
from conan import ConanFile
from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout, CMakeDeps
from conan.tools.env import VirtualRunEnv

required_conan_version = ">=1.59"


class ProfileConflictConan(ConanFile):
    name = "profile_conflict"

    # Binary configuration
    options = {"shared": [True, False], "fPIC": [True, False]}
    default_options = {"shared": False, "fPIC": True}

    def requirements(self):
        self.requires("boost/1.82.0")

    def config_options(self):
        if self.options.shared:
            del self.options.fPIC

        if self.settings.os == "Emscripten":
            self.options["boost"].header_only = True 

    def layout(self):
        cmake_layout(self)

    def generate(self):
        cmake = CMakeDeps(self)
        cmake.generate()

        toolchain = CMakeToolchain(self)
        toolchain.generate()
```

`x86_64-gcc-10` profile
```
[settings]
os=Linux
arch=x86_64
compiler=gcc
compiler.version=10
compiler.libcxx=libstdc++11
build_type=Release

[conf]
tools.cmake.cmake_layout:build_folder_vars=['settings.arch', 'settings.compiler', 'settings.compiler.version']
```

`wasm-clang-16` profile
```
[settings]
os=Emscripten
arch=wasm
compiler=clang
compiler.version=16
build_type=Release

[tool_requires]
emsdk/3.1.50
nodejs/20.16.0
```

### How to reproduce it

_No response_

### LINKED ISSUE #26509: [request] emsdk/4.0.2
### Package Name/Version

emsdk/4.0.2

### Changelog

https://github.com/emscripten-core/emsdk/releases/tag/4.0.2

The above-mentioned version is newly released by the upstream project and not yet available as a recipe.
Please add this version.

### LINKED ISSUE #26778: [request] emsdk/4.0.4
### Package Name/Version

emsdk/4.0.4

### Changelog

https://github.com/emscripten-core/emsdk/pull/1534

I’d like an update to the emsdk version, as it appears that it already supports SDL3 and other new features.

Thank you!

croes:
### Summary
Changes to recipe:  **emsdk**

#### Motivation
Add new version for emsdk

#### Details

Full explanation of changes in the following comment: https://github.com/conan-io/conan-center-index/pull/27213#issuecomment-2940444816

Closes #26778
Closes #26509
Closes #14350
Closes #26067

---

### DIFF
--- a/recipes/emsdk/all/conanfile.py
+++ b/recipes/emsdk/all/conanfile.py
@@ -1,12 +1,12 @@
 from conan import ConanFile
 from conan.tools.build import cross_building
 from conan.tools.env import Environment, VirtualBuildEnv
-from conan.tools.files import chdir, copy, get, replace_in_file
+from conan.tools.files import chdir, copy, get
 from conan.tools.layout import basic_layout
 from pathlib import Path
 import os
 
-required_conan_version = ">=2.0.9"
+required_conan_version = ">=2.1"
 
 
 class EmSDKConan(ConanFile):
@@ -14,19 +14,15 @@ class EmSDKConan(ConanFile):
     description = "Emscripten SDK. Emscripten is an Open Source LLVM to JavaScript compiler"
     url = "https://github.com/conan-io/conan-center-index"
     homepage = "https://github.com/kripken/emscripten"
-    topics = ("emsdk", "emscripten", "sdk")
+    topics = ("emsdk", "emscripten", "sdk", "emcc", "em++", "nodejs")
     license = "MIT"
     package_type = "application"
-    settings = "os", "arch", "compiler", "build_type"
+    settings = "os", "arch"
     upload_policy = "skip"
     build_policy = "missing"
 
     def layout(self):
         basic_layout(self, src_folder="src")
-
-    def package_id(self):
-        del self.info.settings.compiler
-        del self.info.settings.build_type
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version],
@@ -89,23 +85,9 @@ class EmSDKConan(ConanFile):
     def package(self):
         copy(self, "LICENSE", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
         copy(self, "*", src=self.source_folder, dst=os.path.join(self.package_folder, "bin"))
-        emscripten = os.path.join(self.package_folder, "bin", "upstream", "emscripten")
-        toolchain = os.path.join(emscripten, "cmake", "Modules", "Platform", "Emscripten.cmake")
-        # FIXME: conan should add the root of conan package requirements to CMAKE_PREFIX_PATH (LIBRARY/INCLUDE -> ONLY; PROGRAM -> NEVER)
-        # allow to find conan libraries
-        replace_in_file(self, toolchain,
-                              "set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)",
-                              "set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY BOTH)")
-        replace_in_file(self, toolchain,
-                              "set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)",
-                              "set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE BOTH)")
-        replace_in_file(self, toolchain,
-                              "set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)",
-                              "set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE BOTH)")
         if not cross_building(self):
             self.run("embuilder build MINIMAL", env=["conanemsdk", "conanrun"]) # force cache population
-            # the line below forces emscripten to accept the cache as-is, even after re-location
-            # https://github.com/emscripten-core/emscripten/issues/15053#issuecomment-920950710
+            # Avoid cache failures in case this package is uploaded as paths in sanity.txt are absolute
             os.remove(os.path.join(self._em_cache, "sanity.txt"))
 
     def _define_tool_var(self, value):
--- a/recipes/emsdk/all/test_package/CMakeLists.txt
+++ b/recipes/emsdk/all/test_package/CMakeLists.txt
@@ -1,4 +1,0 @@
-cmake_minimum_required(VERSION 3.1)
-project(test_package LANGUAGES CXX)
-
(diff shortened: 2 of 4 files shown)


recipes/emsdk/all/conanfile.py
```diff
@@ -29,7 +29,7 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("nodejs/16.3.0")
+        self.requires("nodejs/[>=20.16.0]")
```
jcar87, line 32: please explain the motivation behind this change, thanks!

croes: With the older nodeJS version, I got syntax errors when running emcc. After some additional testing, I actually removed the dependency on conan nodejs completely (cf. https://github.com/conan-io/conan-center-index/pull/27213/commits/c39441526d20e106ccbef5ece942c4de34e1e1b0).

Emsdk manages a nodejs installation by itself (4.0.6 even uses 20.18.0, but that wasn't on conan center yet). By removing this dependency, it makes it easier to update the emsdk recipe, and avoids potential clashes between the conan nodejs and the one managed by emsdk.

AbrilRBS:
We'll be looking into the proposed approach to have emdsdk fetch its node version by itself - this would solve some issues we've seen with transitive dependency conflicts in some rare scenarios such as https://github.com/conan-io/conan-center-index/issues/26067 (cc @jjcasmar)

perseoGI:
Greetings @croes,
The idea of removing the `nodejs` dependency from this recipe looks good to me. 
`emsdk` already vendorizes plenty of dependencies, and it is better to delegate to the `emscripten` team which node version to use.
Also, didn't make much sense to have only one of the enormous number of dependencies vendorized and not the rest.

In my latest commit, I've:
- Removed old `test_v1_package` and conan v1 specific code
- Fix some linter issues
- Removed numerous outdated versions that were no longer in use in CCI. There were 35 versions! I've limited it to 6
- Added latest release `4.0.9`
- Added `node` binaries path to the `bindirs` in `cpp_info`. This will avoid any breaking change with previous recipe revisions, as any user depending only on `emsdk` will still have `node`, `npm`, `npx` (and `corepack`) available on their `conanbuild` environment -> this detail is **crucial**. And not only to avoid breaking changes, but because it is correct to add it as `emsdk` is a SDK which packages all the tools needed for cross-compiling and running JS/asm.js/WASM applications.

**Motivation of adding both `4.0.6` and `4.0.9`**: version `4.0.6` was the latest version after an important breaking change was introduced in `4.0.7`, see https://github.com/emscripten-core/emscripten/blob/main/ChangeLog.md#407---041525.

This PR will add both the latest version of `emsdk` and the latest version before the breaking change. 

**Important**: Version `3.1.50` has not been removed as it is currently being used in CCI by `libqasm`.

----------

### EDIT

- Replaced version `4.0.9` for `4.0.10`.
- Added cross compilation capabilities by defining `EMSDK_ARCH`
- Removed old versions which didn't detect the `EMSDK_ARCH` environment
- Fixed `test_package` to run as a standard requirement and not as a `tool_requirement` due to C3I capabilities
- Marked this package with `upload_policy = "skip"`. Conan Center Index won't store emsdk binaries as this recipe only packages a python script which will download pre compiled binaries from emscripten cloud.

croes:
@perseoGI Thank you for taking the time to not just consider removing the node dependency, but also taking it to the next level! The great explainer was the cherry on the cake :cake:

perseoGI:
Thank you @croes, 
I've been researching WASM recently, and we will hopefully add Emscripten documentation and some examples to the Conan documentation.
As you seem familiar with Emscripten, would you mind checking our current WIP PRs?
https://github.com/conan-io/docs/pull/4115
Any feedback from Emscriptem users will be appreciated! 🐸

recipes/emsdk/all/conanfile.py
```diff
@@ -76,6 +72,13 @@ def generate(self):
         env.define_path("EM_CACHE", self._em_cache)
         env.vars(self, scope="emsdk").save_script("emsdk_env_file")
 
+        if cross_building(self):
```
jcar87: Strictly speaking, this doesn't need to be guarded in a conditional - quite the opposite, I'd say it's better to always set `EMSDK_ARCH` to match `self.settings.arch` - the architecture returned by the python interpreter may not always be `self.settings.arch` on systems that can run the interpreter on multiple architectures

fpelliccioni:
@AbrilRBS @jcar87 guys, could you review this PR? please 🙏

perseoGI:
Greetings @fpelliccioni,

This PR is probably not going to be merged in CCI. 
We have created a new repository, `conan-toolchains`.
Our plan is to move the toolchains in CCI to this new repository, and we have started with this one.

Check out the opened PR https://github.com/conan-io/conan-toolchains/pull/4
We expect merging this PR as soon as possible!

Thank you for your patience and happy coding 🐸

skhaz:
Any news on this? I really want to use emsdk 4.X.X

fpelliccioni:
@AbrilRBS @jcar87 hi again guys! Any news on this?

Issue #18323, opened 2023-07-05
[request] podofo/0.10.1 (change from sourceforge to github)


luzpaz:
### Package Name/Version

podofo/0.10.1

### Changelog

https://github.com/podofo/podofo/releases/tag/0.10.1

### Context about the new update

Per the podofo sourceforge URL (https://sourceforge.net/projects/podofo/)  

> Development of PoDoFo has been moved to GitHub: https://github.com/podofo/podofo

Currently the podofo package source in conan is listed as:

https://github.com/conan-io/conan-center-index/blob/ca5db98dbaba198110e72b86d6797a770c377ea2/recipes/podofo/all/conandata.yml#L1-L16

Current podofo stable is at 0.10.1  (https://github.com/podofo/podofo/releases/tag/0.10.1)
Current podofo stable listed in conan is 0.9.7

Malacath-92:
I hope this can be closed with the merging of #27214 or is there a need for exactly 0.10.1?

PR #27214, opened 2025-04-11
podofo: Add version 0.10.4

Malacath-92:
### Summary
Adds recipe for `podofo/0.10.4`

#### Motivation
There was a big API shift from `podofo/0.9.*` to `podofo/0.10.*` so these are very distinct versions. I would argue both should be available.
Also see https://github.com/podofo/podofo/issues/87 and https://github.com/conan-io/conan-center-index/issues/18323

#### Details
Changes in `conandata.yml`:
- Add version 0.10.4, using the correct github location of source

Changes in `conanfile.py`:
- Automated formatting (I can revert if it's unwanted, it was not my intention)
- Add `with_lib_only` option, which is a shorthand for removing tools, tests, and playground
- Fix for podofo/0.9.7 when `with_openssl` is `False`
- Forced some options based on version
- Remove `PODOFO_BUILD_SHARED` with the new version, will cause configure error
- Adjust libs and defines to correctly link statically on new version

Both versions still build and test successfully.

---
- [ x ] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
- [ x ] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)
- [ x ] Tested locally with at least one configuration using a recent version of Conan

### DIFF
--- a/recipes/podofo/all/conanfile.py
+++ b/recipes/podofo/all/conanfile.py
@@ -90,6 +90,10 @@ class PodofoConan(ConanFile):
         if self.options.with_unistring:
             self.requires("libunistring/0.9.10")
 
+    def build_requirements(self):
+        if Version(self.version) >= "0.10.4":  # pylint: disable=conan-condition-evals-to-constant
+            self.tool_requires("cmake/[>=3.15.7]")
+
     def validate(self):
         check_min_cppstd(self, 11)
 
--- a/recipes/podofo/all/conanfile.py
+++ b/recipes/podofo/all/conanfile.py
@@ -55,8 +55,18 @@ class PodofoConan(ConanFile):
         export_conandata_patches(self)
 
     def config_options(self):
+        if (
+            Version(self.version) < "0.10.4"
+        ):  # pylint: disable=conan-condition-evals-to-constant
+            # Not available in older versions
+            self.options.with_lib_only = False
+        else:
+            # Required for this version
+            self.options.with_openssl = True
+
         if self.settings.os == "Windows":
             del self.options.fPIC
+
         if is_msvc(self):
             # libunistring recipe raises for Visual Studio
             # TODO: Enable again when fixed?
@@ -105,7 +115,9 @@ class PodofoConan(ConanFile):
         tc = CMakeToolchain(self)
         tc.variables["PODOFO_BUILD_TOOLS"] = self.options.with_tools
         tc.variables["PODOFO_BUILD_LIB_ONLY"] = self.options.with_lib_only
-        if Version(self.version) < "0.10.0":
+        if (
+            podofo_version < "0.10.0"
+        ):  # pylint: disable=conan-condition-evals-to-constant
             tc.variables["PODOFO_BUILD_SHARED"] = self.options.shared
         tc.variables["PODOFO_BUILD_STATIC"] = not self.options.shared
         if not self.options.threadsafe:
--- a/recipes/podofo/all/conanfile.py
+++ b/recipes/podofo/all/conanfile.py
@@ -73,9 +73,7 @@ class PodofoConan(ConanFile):
     def requirements(self):
         self.requires("freetype/2.13.2")
         self.requires("zlib/[>=1.2.11 <2]")
-        if (
-            Version(self.version) >= "0.10.4"
-        ):  # pylint: disable=conan-condition-evals-to-constant
+        if Version(self.version) >= "0.10.4":  # pylint: disable=conan-condition-evals-to-constant
             self.requires("libxml2/[>2 <3]")
         if self.settings.os != "Windows":
             self.requires("fontconfig/2.15.0")
--- a/recipes/podofo/all/conanfile.py
+++ b/recipes/podofo/all/conanfile.py
@@ -82,7 +82,10 @@ class PodofoConan(ConanFile):
     def requirements(self):
         self.requires("freetype/2.13.2")
         self.requires("zlib/[>=1.2.11 <2]")
-        self.requires("libxml2/[>2 <3]")
+        if (
+            Version(self.version) >= "0.10.4"
+        ):  # pylint: disable=conan-condition-evals-to-constant
+            self.requires("libxml2/[>2 <3]")
         if self.settings.os != "Windows":
             self.requires("fontconfig/2.15.0")
         if self.options.with_openssl:
--- a/recipes/podofo/all/conanfile.py
+++ b/recipes/podofo/all/conanfile.py
@@ -1,7 +1,13 @@
 from conan import ConanFile
 from conan.tools.build import check_min_cppstd
 from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
-from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get, rmdir
+from conan.tools.files import (
+    apply_conandata_patches,
+    copy,
+    export_conandata_patches,
+    get,
+    rmdir,
+)
 from conan.tools.microsoft import is_msvc
 from conan.tools.scm import Version
 import os
@@ -29,6 +35,7 @@ class PodofoConan(ConanFile):
         "with_png": [True, False],
         "with_unistring": [True, False],
         "with_tools": [True, False],
+        "with_lib_only": [True, False],
     }
     default_options = {
         "shared": False,
@@ -41,6 +48,7 @@ class PodofoConan(ConanFile):
         "with_png": True,
         "with_unistring": True,
         "with_tools": False,
+        "with_lib_only": True,
     }
 
     def export_sources(self):
@@ -64,6 +72,7 @@ class PodofoConan(ConanFile):
     def requirements(self):
         self.requires("freetype/2.13.2")
         self.requires("zlib/[>=1.2.11 <2]")
+        self.requires("libxml2/[>2 <3]")
         if self.settings.os != "Windows":
             self.requires("fontconfig/2.15.0")
         if self.options.with_openssl:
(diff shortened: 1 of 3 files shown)


jcar87 (changes requested):
thanks for your PR!

Please disable the reformatting of the file, and limit the PR to the changes required by it, thanks!

Malacath-92:
All the non-functional changes have been reverted.

Malacath-92:
I'm not aware how build requirements work honestly. The build on Linux states `cmake/3.31.6` yet the build fails complaining about using a bad version of cmake
```
  CMake 3.16 or higher is required.  You are running version 3.15.7
```
I'll do some reading, but if it's an easy fix I'm happy to listen.

recipes/podofo/all/conanfile.py
```diff
@@ -79,17 +90,27 @@ def requirements(self):
...
             self.requires("libunistring/0.9.10")
 
+    def build_requirements(self):
+        if Version(self.version) >= "0.10.4":  # pylint: disable=conan-condition-evals-to-constant
+            self.tool_requires("cmake/[>=3.15.7]")
+
     def validate(self):
         check_min_cppstd(self, 11)
+        if Version(self.version) >= "0.10.4":
+            check_min_cppstd(self, "17")
```
ErniGH, line 100: The new version uses `std::string_view`

https://github.com/podofo/podofo/compare/0.9.7...0.10.4#diff-5801755f138bd6410f147d9ade2e7f1d1509ca9881c890e1664d09a9911b1a52R384

recipes/podofo/all/conanfile.py
```diff
@@ -41,12 +42,20 @@ class PodofoConan(ConanFile):
...
+        "with_lib_only": True,
     }
 
     def export_sources(self):
         export_conandata_patches(self)
 
     def config_options(self):
+        if Version(self.version) < "0.10.4":  # pylint: disable=conan-condition-evals-to-constant
+            # Not available in older versions
+            self.options.with_lib_only = False
```
ErniGH, line 54: In these cases, we usually delete the option

recipes/podofo/all/conanfile.py
```diff
@@ -41,12 +42,20 @@ class PodofoConan(ConanFile):
...
+        else:
+            # Required for this version
+            self.options.with_openssl = True
```
ErniGH, line 57: Instead of forcing it to `True`, it is a better approach to put it on `validate()` and raise an error in case it is not a valid option

recipes/podofo/all/conanfile.py
```diff
@@ -64,6 +73,8 @@ def layout(self):
     def requirements(self):
         self.requires("freetype/2.13.2")
         self.requires("zlib/[>=1.2.11 <2]")
+        if Version(self.version) >= "0.10.4":  # pylint: disable=conan-condition-evals-to-constant
+            self.requires("libxml2/[>2 <3]")
```
ErniGH, line 77: Here is the diff to justify this change 😄 
https://github.com/podofo/podofo/compare/0.9.7...0.10.4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR195

recipes/podofo/all/conanfile.py
```diff
@@ -79,17 +90,27 @@ def requirements(self):
         if self.options.with_unistring:
             self.requires("libunistring/0.9.10")
 
+    def build_requirements(self):
+        if Version(self.version) >= "0.10.4":  # pylint: disable=conan-condition-evals-to-constant
+            self.tool_requires("cmake/[>=3.15.7]")
```
ErniGH, line 95: CMake needs the upper bound for the version range
```suggestion
            self.tool_requires(""cmake/[>=3.15.7 <4]"")
```

recipes/podofo/all/conanfile.py
```diff
@@ -104,11 +125,15 @@ def generate(self):
...
         tc.variables["PODOFO_WITH_UNISTRING"] = self.options.with_unistring
-        tc.variables["PODOFO_HAVE_OPENSSL_1_1"] = Version(self.dependencies["openssl"].ref.version) >= "1.1"
-        if self.options.with_openssl and ("no_rc4" in self.dependencies["openssl"].options):
-            tc.variables["PODOFO_HAVE_OPENSSL_NO_RC4"] = self.dependencies["openssl"].options.no_rc4
-        if Version(self.version) < "0.10.0": # pylint: disable=conan-condition-evals-to-constant
+
+        if self.options.with_openssl:
+            tc.variables["PODOFO_HAVE_OPENSSL_1_1"] = Version(self.dependencies["openssl"].ref.version) >= "1.1"
+            if self.options.with_openssl and ("no_rc4" in self.dependencies["openssl"].options):
+                tc.variables["PODOFO_HAVE_OPENSSL_NO_RC4"] = self.dependencies["openssl"].options.no_rc4
```
ErniGH, line 132: Could you help me understand this change? I suppose this is related to the description `Fix for podofo/0.9.7 when with_openssl is False`, but there is some issue describing this bug?

recipes/podofo/all/conanfile.py
```diff
@@ -128,10 +153,18 @@ def package(self):
...
+            self.cpp_info.libs = ["podofo"]
+            if self.settings.os == "Windows" and self.options.shared:
+                self.cpp_info.defines.append("USING_SHARED_PODOFO")
+        else:
+            if not self.options.shared:
+                self.cpp_info.libs = ["podofo", "podofo_private"]
+                self.cpp_info.defines.append("PODOFO_STATIC")
+            else:
+                self.cpp_info.libs = ["podofo"]
+                self.cpp_info.defines.append("PODOFO_SHARED")
```
ErniGH, line 167: Here is the diff to justify this change
https://github.com/podofo/podofo/compare/0.9.7...0.10.4#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR76-R81

recipes/podofo/all/conanfile.py
```diff
@@ -79,17 +90,27 @@ def requirements(self):
...
                   destination=self.source_folder, strip_root=True)
 
     def generate(self):
+        podofo_version = Version(self.version)
+
         tc = CMakeToolchain(self)
         tc.variables["PODOFO_BUILD_TOOLS"] = self.options.with_tools
-        tc.variables["PODOFO_BUILD_SHARED"] = self.options.shared
+        tc.variables["PODOFO_BUILD_LIB_ONLY"] = self.options.with_lib_only
+        if podofo_version < "0.10.0":  # pylint: disable=conan-condition-evals-to-constant
```
ErniGH, line 112: I found comments like `# pylint: disable=conan-condition-evals-to-constant` in the new changes u added. I do not understand the purpose of those comments. All the ifs u added are completely valid and should not be constant. 
Could you please help me to understand them 🤔 ?

ErniGH (commented):
Thanks for your PR some comments are just to make it easier to track the changes with the diff between versions

https://github.com/podofo/podofo/compare/0.9.7...0.10.4

recipes/podofo/all/conanfile.py
Malacath-92: If `self.options.with_openssl` is set to `False` by the user then this recipe would not introduce the `openssl` dependency. Thereby accessing `self.dependencies["openssl"]` wouldn't work as expected. Hence all the code accessing this value is now avoided when the option is set to `False`

Malacath-92: There was no previous issue for this as far as I am aware. It was a very small fix so I did it as part of this PR.

recipes/podofo/all/conanfile.py
Malacath-92: I don't know enough about conan to know whether this is needed. I followed the code as it was seen in this recipe before my changes: https://github.com/conan-io/conan-center-index/pull/27214/files#diff-08a143f8891ecaeb9045930f6ddd684800e9dc0af45fa09f9f8e01307604011aL110

Perhaps that is not needed, perhaps it is. I did not run pylint locally as I didn't find any requirements to do so and did not find any instructions how to. Altough I didn't search very hard.

recipes/podofo/all/conanfile.py
Malacath-92: I addressed this, let me know if it is satisfactory.

recipes/podofo/all/conanfile.py
Malacath-92: I addressed this, let me know if it is satisfactory.

recipes/podofo/all/conanfile.py
Malacath-92: Please let me know if I should delete those pylint comments, or alternatively you can delete them yourself in this PR. Makes no difference to me.

jcar87: Please do, yes - they're not needed here

Malacath-92: Comments are removed.

ErniGH (dismissed):
LGTM

luzpaz:
Just a heads up: v0.10.5[^1] and v1.0.0[^2] have been released upstream  

[^1]: https://github.com/podofo/podofo/releases/tag/0.10.5
[^2]: https://github.com/podofo/podofo/releases/tag/1.0.0

Malacath-92:
@luzpaz and I want to fix recipes for those, but until this is merged I don't want to open a new PR

@jcar87 @ErniGH 
How do we proceed with this? The person who reviewed and approved this unfortunately committed a fix before approving so this can't be merged. Should I just update the PR for version 0.10.5 and we have that reviewed? Likely that is basically no work at all.

PR #27217, opened 2025-04-13
Update quant lib to v1.37, fix bug that conan always use static runtime ignoring profile settings

Feng-Zihao:
### Summary
Changes to recipe:  **QuantLib/[v1.37]**

#### Motivation
1. Upgrade QuantLib to version v1.37. 
2. Use the official version without patches.
3. Fix bug that the original conanfile.py didn't pass CMAKE_MSVC_RUNTIME_LIBRARY to cmake, then QuantLib will always build QuantLib-x64-mt-s.lib even if conan profiles set c runtime to dynamic. The original version can not be used under any project that use dynamic msvc c runtime.

#### Details
1. Add v1.37 version related
2. Modify conanfile.py to set CMAKE_MSVC_RUNTIME_LIBRARY according to compiler.runtime under MSVC
3. No patches is made, directly use official QuantLib-v1.37 source code

#### Local Build Test results.
1. Linux
```
quantlib/1.37 (test package): Running test()
quantlib/1.37 (test package): RUN: ./test_package

--------------
Exact: 2.6303
Quad: 2.6303
Grid: 2.6303

```
2. Windows MSVC
```
conanvcvars.bat: Activating environment Visual Studio 17 - amd64 - winsdk_version=None - vcvars_ver=14.4
[vcvarsall.bat] Environment initialized for: 'x64'
[2/2] Linking CXX executable test_package.exe


quantlib/1.37 (test package): Running test()
quantlib/1.37 (test package): RUN: .\test_package

--------------
Exact: 2.6303
Quad: 2.6303
Grid: 2.6303

```

--------
[ x ] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
[ x ] Checked that this PR is not a duplicate
[ x ] Tested locally with at least one configuration using a recent version of Conan

### DIFF
--- a/recipes/quantlib/all/conandata.yml
+++ b/recipes/quantlib/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.37":
+    url: "https://github.com/lballabio/QuantLib/releases/download/v1.37/QuantLib-1.37.tar.gz"
+    sha256: "b284e54ef2133c9eccccc23e210ae5b41b2f37c797b3d7257492a391e436ad24"
   "1.30":
     url: "https://github.com/lballabio/QuantLib/releases/download/QuantLib-v1.30/QuantLib-1.30.tar.gz"
     sha256: "10159054b68cb9a39480d9000b87851f49e6f42474a4cb9367e934bece2363f3"
@@ -6,6 +9,8 @@ sources:
     url: "https://github.com/lballabio/QuantLib/releases/download/QuantLib-v1.22/QuantLib-1.22.tar.gz"
     sha256: "85c81816f689f458596dd7073e4da8fd7f596c1e4c8ada81a6300389a39588af"
 patches:
+  "1.37":
+    - patch_file: "patches/1.37-fix-cmake.patch"
   "1.30":
     - patch_file: "patches/1.30-backport-pr-1644.patch"
     - patch_file: "patches/1.30-fix-cmake.patch"
--- a/recipes/quantlib/all/conandata.yml
+++ b/recipes/quantlib/all/conandata.yml
@@ -9,8 +9,6 @@ sources:
     url: "https://github.com/lballabio/QuantLib/releases/download/QuantLib-v1.22/QuantLib-1.22.tar.gz"
     sha256: "85c81816f689f458596dd7073e4da8fd7f596c1e4c8ada81a6300389a39588af"
 patches:
-  "1.37":
-    - patch_file: "patches/1.37-fix-cmake.patch"
   "1.30":
     - patch_file: "patches/1.30-backport-pr-1644.patch"
     - patch_file: "patches/1.30-fix-cmake.patch"
--- a/recipes/quantlib/all/conanfile.py
+++ b/recipes/quantlib/all/conanfile.py
@@ -43,7 +43,7 @@ class QuantlibConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("boost/1.80.0", transitive_headers=True)
+        self.requires("boost/[>=1.80.0]", transitive_headers=True)
 
     def validate(self):
         if self.info.settings.compiler.get_safe("cppstd"):
--- a/recipes/quantlib/all/conanfile.py
+++ b/recipes/quantlib/all/conanfile.py
@@ -36,6 +36,16 @@ class QuantlibConan(ConanFile):
             del self.options.fPIC
 
     def configure(self):
+        if is_msvc(self):
+            # From no later than 1.30, CMAKE_MSVC_RUNTIME_LIBRARY determine target name postfix
+            # CMAKE_MSVC_RUNTIME_LIBRARY will set to MultiThreaded$<$<CONFIG:Debug>:Debug> if not explicited set
+            # after testing, only this way can take effect
+            # (see https://github.com/lballabio/QuantLib/blob/QuantLib-v1.30/cmake/Platform.cmake#L7
+            #  and https://github.com/quantlib/QuantLib/blob/QuantLib-v1.30/CMakeLists.txt#L232)
+            if self.settings.compiler.runtime == "dynamic":
+                self.conf.define("tools.cmake.cmaketoolchain:extra_variables", {"CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL"})
+            else: # static runtime
+                self.conf.define("tools.cmake.cmaketoolchain:extra_variables", {"CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreaded$<$<CONFIG:Debug>:Debug>"})
         if self.options.shared:
             self.options.rm_safe("fPIC")
 
@@ -43,6 +53,7 @@ class QuantlibConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
+        # Loss check is preferred
         self.requires("boost/[>=1.80.0]", transitive_headers=True)
 
     def validate(self):
--- a/recipes/quantlib/config.yml
+++ b/recipes/quantlib/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "1.37":
+    folder: "all"
   "1.30":
     folder: "all"
   "1.22":


Feng-Zihao:
@AbrilRBS Could you review please? Sorry if I at the wrong guy.

uilianries:
@Feng-Zihao Thank you for sharing your case. Could you please also share a build log that provides evidence of your case? Only the test package result reveals almost nothing for us to compare the scenarios.

The Conan generator CMakeToolchain should be configuring the CMAKE_MSVC_RUNTIME_LIBRARY itself, so you should not need to enforce, because quantlib only enforces it in case not configured already: https://github.com/lballabio/QuantLib/blob/v1.37/cmake/Platform.cmake#L5

Feng-Zihao:
@uilianries Hi. Here's the end of output when not specifying 'CMAKE_MSVC_RUNTIME_LIBRARY '

profile
```
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=17
compiler.runtime=dynamic
compiler.version=194
os=Windows

[conf]
```

command to run (just download and build, not specifying 'CMAKE_MSVC_RUNTIME_LIBRARY ' in conanfile.py)
```
C:\Users\panzi\Documents\Project\conan-center-index\recipes\quantlib\all>conan create . --version 1.37
```

end of output 
```
-- Conan: Target declared 'boost::boost'
-- Conan: Target declared 'ZLIB::ZLIB'
-- Conan: Target declared 'BZip2::BZip2'
-- Conan: Including build module from 'C:/Users/panzi/.conan2/p/bzip25d2dbaf142a52/p/lib/cmake/conan-official-bzip2-variables.cmake'
-- Configuring done (2.9s)
-- Generating done (0.1s)
-- Build files have been written to: C:/Users/panzi/Documents/Project/conan-center-index/recipes/quantlib/all/test_package/build/msvc-194-x86_64-17-release

quantlib/1.37 (test package): Running CMake.build()
quantlib/1.37 (test package): RUN: cmake --build "C:\Users\panzi\Documents\Project\conan-center-index\recipes\quantlib\all\test_package\build\msvc-194-x86_64-17-release" -- -j12
conanvcvars.bat: Activating environment Visual Studio 17 - amd64 - winsdk_version=None - vcvars_ver=14.4
[vcvarsall.bat] Environment initialized for: 'x64'
[2/2] Linking CXX executable test_package.exe
FAILED: test_package.exe
C:\Windows\system32\cmd.exe /C "cd . && "C:\Program Files\CMake\bin\cmake.exe" -E vs_link_exe --msvc-ver=1944 --intdir=CMakeFiles\test_package.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100203~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100203~1.0\x64\mt.exe --manifests  -- C:\PROGRA~1\MICROS~2\2022\Preview\VC\Tools\MSVC\1444~1.349\bin\Hostx64\x64\link.exe /nologo CMakeFiles\test_package.dir\test_package.cpp.obj  /out:test_package.exe /implib:test_package.lib /pdb:test_package.pdb /version:0.0 /machine:x64 /INCREMENTAL:NO /subsystem:console -LIBPATH:C:\Users\panzi\.conan2\p\b\quant013b7aaf85cb5\p\lib   -LIBPATH:C:\Users\panzi\.conan2\p\boost4b2b6018aea99\p\lib C:\Users\panzi\.conan2\p\b\quant013b7aaf85cb5\p\lib\QuantLib-x64-mt-s.lib  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib && cd ."
LINK: command "C:\PROGRA~1\MICROS~2\2022\Preview\VC\Tools\MSVC\1444~1.349\bin\Hostx64\x64\link.exe /nologo CMakeFiles\test_package.dir\test_package.cpp.obj /out:test_package.exe /implib:test_package.lib /pdb:test_package.pdb /version:0.0 /machine:x64 /INCREMENTAL:NO /subsystem:console -LIBPATH:C:\Users\panzi\.conan2\p\b\quant013b7aaf85cb5\p\lib -LIBPATH:C:\Users\panzi\.conan2\p\boost4b2b6018aea99\p\lib C:\Users\panzi\.conan2\p\b\quant013b7aaf85cb5\p\lib\QuantLib-x64-mt-s.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST:EMBED,ID=1" failed (exit code 1104) with the following output:
LINK : fatal error LNK1104: 无法打开文件“QuantLib-x64-mt.lib”

ERROR: quantlib/1.37 (test package): Error in build() method, line 21
        cmake.build()
```

I'm using Chinese, **LINK : fatal error LNK1104: 无法打开文件“QuantLib-x64-mt.lib” ** means "can not open file QuanLib-x64-mt.lib".

It builds 'QuantLib-x64-mt-s.lib' under the above profile and command.

However, when building the test_package.exe, linker failed to find '''QuantLib-x64-mt.lib", which relies on 'dynamic runtime'.

I also found [CMake Document for CMAKE_MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html). It says

```
If this variable is not set then the [MSVC_RUNTIME_LIBRARY](https://cmake.org/cmake/help/latest/prop_tgt/MSVC_RUNTIME_LIBRARY.html#prop_tgt:MSVC_RUNTIME_LIBRARY) target property will not be set automatically. If that property is not set then CMake uses the default value MultiThreaded$<$<CONFIG:Debug>:Debug>DLL to select a MSVC runtime library.
```

There's a conflict default behavior between CMake defualts and Quantlib Defaults, as https://github.com/lballabio/QuantLib/blob/v1.37/cmake/Platform.cmake#L5 described.

Feng-Zihao:
https://github.com/conan-io/conan-center-index/blob/master/recipes/quantlib/all/patches/1.30-fix-cmake.patch#L62

OK. I found that in 1.30, this problem is found but fixed by deleting the related lines through 1.30-fix-cmake.patch.

I consider my fix is better because it leave everything 'as is' the official version. This will make later upgrade more easily.

jcar87:
Hi @Feng-Zihao - thanks for providing more details.

As @uilianries  mentions, Conan already sets the `CMAKE_MSVC_RUNTIME_LIBRARY` value following the profile settings - this is even printed in the CMake configuration log:

```
-- Using Conan toolchain: C:/j/b/quante671ac2a4b87e/b/build/generators/conan_toolchain.cmake
-- Conan toolchain: CMAKE_GENERATOR_TOOLSET=v143
-- Conan toolchain: Setting CMAKE_MSVC_RUNTIME_LIBRARY=$<$<CONFIG:Release>:MultiThreadedDLL>
-- Conan toolchain: C++ Standard 17 with extensions OFF
```

This should be supported by the logic upstream already, since by the time the variable is checked here https://github.com/lballabio/QuantLib/blob/v1.37/cmake/Platform.cmake#L5-L8, it is already defined - and thus.

> I'm using Chinese, LINK : fatal error LNK1104: 无法打开文件“QuantLib-x64-mt.lib” means "can not open file QuanLib-x64-mt.lib".
> 
> It builds 'QuantLib-x64-mt-s.lib' under the above profile and command.

I dont think this error means that the library was built with the static runtime (disregarding the Conan profile) - the error we would expect to see in that case is something like this:

```
error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MT_StaticRelease' doesn't match value 'MD_DynamicRelease' in file.obj. 
```
(from 
https://stackoverflow.com/questions/28887001/lnk2038-mismatch-detected-for-runtimelibrary-value-mt-staticrelease-doesn)


The issue is something else altogether

The `auto_link.hpp` file inside the package has this logic:

```
#ifndef quantlib_autolink_hpp
#define quantlib_autolink_hpp

#include <ql/version.hpp>
#include <boost/config.hpp>

#ifdef _M_X64
#  define QL_LIB_PLATFORM "-x64"
#else
#  define QL_LIB_PLATFORM
#endif

/*** libraries to be linked ***/

// select thread opt:
#ifdef _MT
#  define QL_LIB_THREAD_OPT "-mt"
#else
#  define QL_LIB_THREAD_OPT
#endif

// select linkage opt:
#ifdef _DLL
#  if defined(_DEBUG)
#    define QL_LIB_RT_OPT "-gd"
#  else
#    define QL_LIB_RT_OPT
#  endif
#else
#  if defined(_DEBUG)
#    define QL_LIB_RT_OPT "-sgd"
#  else
#    define QL_LIB_RT_OPT "-s"
#  endif
#endif

#define QL_LIB_NAME "QuantLib" QL_LIB_PLATFORM QL_LIB_THREAD_OPT QL_LIB_RT_OPT ".lib"

#pragma comment(lib, QL_LIB_NAME)
#ifdef BOOST_LIB_DIAGNOSTIC
#  pragma message("Will (need to) link to lib file: " QL_LIB_NAME)
#endif

#endif

```

What is happening here is that `#pragma comment` is instructing the msvc linker to link `QL_LIB_NAME`, which derives a name that is different to the one that was produced by the build system - so it fails.

This is probably because the logic in the build scripts is also wrong - you'll see that the patch has this - the way the regex is constructed upstream means that even when it is built with the dynamic runtime (the "DLL"), it is adding the `-mt` suffix, which is wrong. It is very fragile to perform a configure-time check against a generation-time expression

```
-    if (CMAKE_MSVC_RUNTIME_LIBRARY MATCHES ".*DLL$")
+    if (CMAKE_MSVC_RUNTIME_LIBRARY MATCHES "DLL")
```

My advice:
* Set `tc.cache_variables["QL_TAGGED_LAYOUT"] = False` to avoid the renaming (rather than patching it) - because it's broken otherwise - I would strongly advise to report this upstream - it may be possible to set the naming also with a generator expression to be more robust.
* Patch the `#pragma comment(lib, QL_LIB_NAME)` in that header file so that it is not used at all - it's not needed when consuming via CMake config, pkg-config files, or Conan.

jcar87:
Edit: nevermind, the problem is actually the test_package
```
#include <ql/qldefines.hpp>
#ifdef BOOST_MSVC
#include <ql/auto_link.hpp>
#endif
```

it shouldn't include `auto_link.hpp` unconditionally like that (The quantlib own headers don't do this) - that's why it's failing

Feng-Zihao:
@jcar87 Before discussing about the source code of test_package, I want to point out several points.

1. You see in CI build **-- Conan toolchain: Setting CMAKE_MSVC_RUNTIME_LIBRARY=$<$<CONFIG:Release>:MultiThreadedDLL>** is because of the modification of **self.conf.define("tools.cmake.cmaketoolchain:extra_variables", {"CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL"})**

2. Without this modification, it build **QuantLib-x64-mt-s.lib** actually. The detailed is posted above https://github.com/conan-io/conan-center-index/pull/27217#issuecomment-2800809048

3. I checked the recipes of 1.30, the patches delete the "NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY" test (see https://github.com/conan-io/conan-center-index/blob/19d3fc8e3f0e2f25acf5fd7fca6e4be4891f8fc7/recipes/quantlib/all/patches/1.30-fix-cmake.patch#L64). It seems that the commitor has already locate the problem and the effect of 1.30 patches.


I think the key problem is  "the ambiguous exactl value of **(NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY)**. **Not defined** may not the same of *${CMAKE_MSVC_RUNTIME_LIBRARY}".

jcar87:
> You see in CI build -- Conan toolchain: Setting CMAKE_MSVC_RUNTIME_LIBRARY=$<$CONFIG:Release:MultiThreadedDLL> is because of the modification of self.conf.define("tools.cmake.cmaketoolchain:extra_variables", {"CMAKE_MSVC_RUNTIME_LIBRARY": "MultiThreaded$<$CONFIG:Debug:Debug>DLL"})

Sorry, I forgot to clarify - but no, I tested version 1.37 without your changes (just adding the version). Conan *will* pass `-- Conan toolchain: Setting CMAKE_MSVC_RUNTIME_LIBRARY=$<$<CONFIG:Release>:MultiThreadedDLL>` without your changes

> Without this modification, it build QuantLib-x64-mt-s.lib

It _creates_ a file with that filename - but that library is linked with the dynamic runtime (or whatever the Conan profile is saying). That is, the build system is generating the wrong filename, but the correct library. This is because this block:
https://github.com/lballabio/QuantLib/blob/v1.37/CMakeLists.txt#L231-L236
and in particular
```
    set(DEBUG_POSTFIX ${DEBUG_POSTFIX}-mt)
    set(RELEASE_POSTFIX ${RELEASE_POSTFIX}-mt)
    if (CMAKE_MSVC_RUNTIME_LIBRARY MATCHES ".*DLL$")
        set(DEBUG_POSTFIX ${DEBUG_POSTFIX}-gd)
    else()
```

This is only setting the filename, not the runtime. The issue is that  `CMAKE_MSVC_RUNTIME_LIBRARY` is a cmake generator expression, but this logic expects a value that _ends_ with `DLL` when the passed value ends with `DLL>`. This is the bit that is currently patched for 1.30 in the recipe.


> I checked the recipes of 1.30, the patches delete the "NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY" test (see
> 
> [conan-center-index/recipes/quantlib/all/patches/1.30-fix-cmake.patch](https://github.com/conan-io/conan-center-index/blob/19d3fc8e3f0e2f25acf5fd7fca6e4be4891f8fc7/recipes/quantlib/all/patches/1.30-fix-cmake.patch#L64)
> 
> Line 64 in [19d3fc8](https://github.com/conan-io/conan-center-index/commit/19d3fc8e3f0e2f25acf5fd7fca6e4be4891f8fc7)
> 
>  -    if (NOT DEFINED CMAKE_MSVC_RUNTIME_LIBRARY) 
> ). It seems that the commitor has already locate the problem and the effect of 1.30 patches.

It's unclear why that was deleted - the Conan toolchain defines that value and thus deleting it or keeping it is a no-op - There's no reason to remove that logic, nor obvious benefit if one assumes Conan 2 conventions.


I've done a local build of 1.37 without the other changes, and the two issues are:
* The flawed logic for the filename, for which I suggest setting `QL_TAGGED_LAYOUT` to False (this only controls the filename)
* The test package, which is unconditionally including that header file

jcar87:
If you launch a Visual C++ command prompt, you can run

```
dumpbin /directives QuantLib-x64-mt-s.lib
```

and verify that it was built with `/MD` (the dynamic runtime), by checking for the presence of `/DEFAULTLIB:MSVCRT` 

if it was actually MT, it would say `/DEFAULTLIB:LIBCMT`, as per https://stackoverflow.com/questions/19487465/how-to-tell-if-a-lib-was-compiled-with-mt-or-md

Feng-Zihao:
OK, you're convinsible !

1. conan-toolchain did set the CMAKE_MSVC_RUNTIME_LIBRARY, usually in the generator file
```
cmake_policy(GET CMP0091 POLICY_CMP0091)
if(NOT "${POLICY_CMP0091}" STREQUAL NEW)
    message(FATAL_ERROR "The CMake policy CMP0091 must be NEW, but is '${POLICY_CMP0091}'")
endif()
message(STATUS "Conan toolchain: Setting CMAKE_MSVC_RUNTIME_LIBRARY=$<$<CONFIG:Release>:MultiThreadedDLL>")
set(CMAKE_MSVC_RUNTIME_LIBRARY "$<$<CONFIG:Release>:MultiThreadedDLL>")
```

2. I think the author of quantlib want to distribute multiple runtime/debug version in one folder in Boost style, so he **TAGGED** the filename, but but script control is wrong.

I'll report to upstream.

By the way, conan is great! I think it much faster and flexible than vcpkg. It's a great tool!

Feng-Zihao:
confirm upstream error, close this. 
if fixed by upstream, request pull again.

PR #27222, opened 2025-04-14
openssl: add `fips_module_version` option

gegles:
This follows up https://github.com/conan-io/conan-center-index/pull/24709 and the latest review guidance in this PR.

### Summary
Changes to recipe: **openssl/***

#### Motivation
Allow the OpenSSL 3 recipe to package the FIPS module from another OpenSSL recipe version when a consumer explicitly requests that module version.

The recipe does not decide whether a module version is externally validated. Users remain responsible for choosing the module version they need and running the required deployment/self-validation procedure in the environment where OpenSSL is used.

#### Details
* Add the `fips_module_version` option. It defaults to `None`.
* If `fips_module_version` is unset, or equals the current OpenSSL version, keep the existing recipe behavior.
* If `fips_module_version` is set to an earlier version, require `openssl/<fips_module_version>` privately and copy that package's FIPS module into the current package.
* Reject `fips_module_version` values greater than the current OpenSSL version.
* Reject combining `fips_module_version` with `no_fips=True`.
* Keep the test package narrow: when `fips_module_version` is set, it checks that the packaged FIPS module binary contains the requested module version without loading the provider or running `fipsinstall`.

#### Local validation
* `python3 -m py_compile recipes/openssl/3.x.x/conanfile.py recipes/openssl/3.x.x/test_package/conanfile.py`
* `git diff --check`
* Conan 2.30.0 graph checks for:
  * `openssl/3.6.3` with `fips_module_version=3.1.2`
  * `fips_module_version` equal to the current version, confirming no self-dependency is added
  * `fips_module_version` with `no_fips=True`, confirming the package is marked invalid

---

### DIFF
--- a/recipes/openssl/3.x.x/conanfile.py
+++ b/recipes/openssl/3.x.x/conanfile.py
@@ -89,12 +89,14 @@ class OpenSSLConan(ConanFile):
         "no_zlib": [True, False],
         "openssldir": [None, "ANY"],
         "tls_security_level": [None, 0, 1, 2, 3, 4, 5],
+        "fips_module_version": [None, "ANY"],
     }
     default_options = {key: False for key in options.keys()}
     default_options["fPIC"] = True
     default_options["no_md2"] = True
     default_options["openssldir"] = None
     default_options["tls_security_level"] = None
+    default_options["fips_module_version"] = None
 
     @property
     def _is_clang_cl(self):
@@ -109,6 +111,14 @@ class OpenSSLConan(ConanFile):
     def _use_nmake(self):
         return self._is_clang_cl or is_msvc(self)
 
+    @property
+    def _fips_module_version(self):
+        return self.options.get_safe("fips_module_version")
+
+    @property
+    def _uses_external_fips_module(self):
+        return self._fips_module_version not in (None, self.version)
+
     def config_options(self):
         if self.settings.os != "Windows":
             self.options.rm_safe("capieng_dialog")
@@ -121,6 +131,8 @@ class OpenSSLConan(ConanFile):
             self.options.rm_safe("fPIC")
         self.settings.rm_safe("compiler.libcxx")
         self.settings.rm_safe("compiler.cppstd")
+        if self._fips_module_version in (None, self.version):
+            self.options.rm_safe("fips_module_version")
 
     def layout(self):
         basic_layout(self, src_folder="src")
@@ -128,10 +140,24 @@ class OpenSSLConan(ConanFile):
     def requirements(self):
         if not self.options.no_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
+        if self._fips_module_version:
+            self.output.info(f"Using FIPS module from openssl/{self._fips_module_version}")
+            self.requires(
+                f"openssl/{self._fips_module_version}",
+                visible=False,
+                libs=False,
+                headers=False,
+                run=False,
+                options={"no_fips": False, "no_zlib": True},
+            )
 
     def validate(self):
         if self.settings.os == "iOS" and self.options.shared:
             raise ConanInvalidConfiguration("OpenSSL 3 does not support building shared libraries for iOS")
+        if self._fips_module_version and Version(self._fips_module_version) > Version(self.version):
+            raise ConanInvalidConfiguration("fips_module_version cannot be greater than the current OpenSSL version")
+        if self._fips_module_version and self.options.no_fips:
+            raise ConanInvalidConfiguration("fips_module_version requires no_fips to be set to False")
 
     def build_requirements(self):
         if self.settings_build.os == "Windows":
@@ -366,7 +392,8 @@ class OpenSSLConan(ConanFile):
         else:
             args.append("-fPIC" if self.options.get_safe("fPIC", True) else "no-pic")
 
-        args.append("no-fips" if self.options.get_safe("no_fips", True) else "enable-fips")
+        no_fips = self.options.get_safe("no_fips", True) or self._uses_external_fips_module
+        args.append("no-fips" if no_fips else "enable-fips")
         args.append("no-md2" if self.options.get_safe("no_md2", True) else "enable-md2")
         if str(self.options.tls_security_level) != "None":
             args.append(f"-DOPENSSL_TLS_SECURITY_LEVEL={self.options.tls_security_level}")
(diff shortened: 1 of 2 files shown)


GregDomjan:
I had a go with this recipe and found it a bit challenging to apply similar options to the direct dependency and also the transitive dependency.

While it was possible to set all the options in a profile, maintaining those options outside the recipe is something I'm still wrapping my head around.

Rather than specify only no_fips `options={'no_fips': False}` would it be reasonable to clone the options to apply to the same options to the dependency?

``` Python
                self.output.info(f"Using validated FIPS module from openssl/{self._fips_version}")
                suboptions = self.options.copy_conaninfo_options()
                suboptions.no_fips = False
                self.requires(f"openssl/{self._fips_version}", visible=False, libs=False, headers=False, run=False, options=suboptions)
```

gegles:
@jcar87 @uilianries and others, it's getting a bit old to keep maintaining this forward... Could we please merge this work? It may not be perfect, but it works really well.

LMK. Thanks!

gegles:
@jcar87 @uilianries and others, it's getting a bit old to keep maintaining this forward... Could we please merge this work? It may not be perfect, but it works really well.

LMK. Thanks!

GregDomjan:
@gegles  Noticing that the automated test builds are failing to find openssl runtime env, it seemed to me that running the "test" process in the packaging step didn't seem appropriate as the runtime for the package under build didn't seem to be part of the run env at that point.

> /home/conan/workspace/cci_prod_PR-27222/conan-home/p/b/opensf3983e0a96cb2/p/bin/openssl: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory

Perhaps this can help?
 - move the 'copy fips lib' to build step, and move the test process to the test conan project.
[0001-Move-fips-test-to-test_package.patch](https://github.com/user-attachments/files/22989553/0001-Move-fips-test-to-test_package.patch)

gegles:
> @gegles Noticing that the automated test builds are failing to find openssl runtime env, it seemed to me that running the "test" process in the packaging step didn't seem appropriate as the runtime for the package under build didn't seem to be part of the run env at that point.
> 
> > /home/conan/workspace/cci_prod_PR-27222/conan-home/p/b/opensf3983e0a96cb2/p/bin/openssl: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory
> 
> Perhaps this can help?
> 
> * move the 'copy fips lib' to build step, and move the test process to the test conan project.
>   [0001-Move-fips-test-to-test_package.patch](https://github.com/user-attachments/files/22989553/0001-Move-fips-test-to-test_package.patch)

Thanks @GregDomjan!

The FIPS verification previously executed in package() has been relocated into the test package,
    and the provider copy happens during the build/package phase. The new test logic runs with the
    generated run environment, so the openssl binary can load libssl.so.3, preventing the CI failure
    you flagged.

gegles:
FYI @jcar87 @uilianries, with my latest changes, all the tests now pass, hopefully we can work towards merging this soon.

I was also curious/wondering if and how we should add dedicated tests in the CI for the `use_validated_fips` option?

LMK. thx!

gegles:
> FYI @jcar87 @uilianries, with my latest changes, all the tests now pass, hopefully we can work towards merging this soon.
> 
> I was also curious/wondering if and how we should add dedicated tests in the CI for the `use_validated_fips` option?
> 
> LMK. thx!

@jcar87 @uilianries I know you must be extremely busy, but anything that could be done here to move this forward?

gegles:
@jcar87 @uilianries @ErniGH, @memsharded what do you guys want to do about this one? It's been 6+ months now....

Appreciate any guidance, thx!

jcar87:
Still working on a final review - please dont close it! :) 
(need to pick it up from https://github.com/conan-io/conan-center-index/pull/24709 - as some questions still prevailed)

gegles:
> Still working on a final review - please dont close it! :) (need to pick it up from #24709 - as some questions still prevailed)

Thanks @jcar87, don't worry, I was not gonna close, I am just really happy to hear that you or someone is actively looking into it, so thank you!

gegles:
@jcar87, any status update on this one? Thx!

gegles:
@jcar87, any status update on this one? Thx!

GregDomjan:
OpenSSL 4 updates have been merged blocking this MR yet again
"Another PR modifying the same folder as this PR was merged to master. Please rebase to sync with the most recent changes and push to relaunch the CI checks"

jcar87:
Hi @gegles  - finally have time to review this further.

I think the changes are good and work conceptually - although I have a couple of concerns. My gut feeling is that it is outside the scope of the recipe to take a stance as to whether or not the fips module is "validated" - in line with my earlier comments in https://github.com/conan-io/conan-center-index/pull/24709#discussion_r1995962848

Whether the FIPS module of a given version of OpenSSL is fips-validated is something that happens outside of the scope of the recipe - a version that is not validated today, may be validated in the future. Having the recipe maintain a list of "validated" versions means that it can be out of date to external factors - bear in mind that we do simultaneously maintain multiple versions for the recipe at any given time, so updating to add a new version to the "validated" list does come with a rather heavy CI build.

My second concern (also from https://github.com/conan-io/conan-center-index/pull/24709#discussion_r1995962848) is that there is a detailed procedure in OpenSSL documentations when it comes to the self-validation checks to actually deem the module valid - those fall in the realm of where the library is used, not where it is built, so I'm not fully comfortable having a "validated_fips" in the option name -  because it suggests the recipe is providing something that may not actually pass the self tests on some system (for whatever reason). 

To address both my points above, I would propose something like this: https://github.com/jcar87/conan-center-index/pull/new/lcc/proof-of-concept/openssl-fips-module-option - that is, simply have a `fips_module_version` option, and if a version of openssl earlier than the current is specified, then it will  require that earlier version and copy the fips module (so you end up with: newer openssl, fips module from older version) - with a removal of the option if `fips_module_version` equals `self.version` to prevent recursion. 

This means that the recipe does not need to be updated if a new version becomes fips certified, and discharges all responsibility to the user to perform the self checks wherever it is installed - and makes the recipe logic a lot easier. To accommodate the test package, I'd also go for the simplest version - and just test that the fips module reports the earlier version that was requested, rather than the current version - so the test package should be simplified as well (in line with https://github.com/conan-io/conan-center-index/pull/24709#discussion_r1693453515 - although `strings` doesnt work on windows)

gegles:
Thanks @jcar87, I reworked this along the lines you suggested.

The PR now uses a neutral `fips_module_version` option instead of `use_validated_fips`, does not add or maintain a list of validated versions, and requires/copies the requested module from the selected earlier `openssl/<version>` package.

I also simplified the test package so it only checks that the packaged FIPS module binary contains the requested module version, without loading the provider or running `fipsinstall`. The PR description has been updated with the new behavior and local graph checks.

jcar87 (approved):
LGTM, thanks so much @gegles for your patience!

PR #27234, opened 2025-04-15
Removed test_v1_package from several recipes

perseoGI:
### Summary

Core recipe maintainance: removed deprecated `test_v1_package` and `test_v1_package_module` from all recipes which does not have any openened PR to avoid conflicts.


jcar87:
Will address separately -

PR #27251, opened 2025-04-16
podofo: add libjpeg options "libjpeg", "libjpeg-turbo", "mozjpeg"

geraldgg:
[### Summary

#### Details
](https://github.com/conan-io/conan-center-index/discussions/24240)

---

### DIFF
--- a/recipes/podofo/all/conanfile.py
+++ b/recipes/podofo/all/conanfile.py
@@ -24,7 +24,7 @@ class PodofoConan(ConanFile):
         "threadsafe": [True, False],
         "with_openssl": [True, False],
         "with_libidn": [True, False],
-        "with_jpeg": [True, False],
+        "with_libjpeg": [False, "libjpeg", "libjpeg-turbo", "mozjpeg"],
         "with_tiff": [True, False],
         "with_png": [True, False],
         "with_unistring": [True, False],
@@ -36,7 +36,7 @@ class PodofoConan(ConanFile):
         "threadsafe": True,
         "with_openssl": True,
         "with_libidn": True,
-        "with_jpeg": True,
+        "with_libjpeg": "libjpeg",
         "with_tiff": True,
         "with_png": True,
         "with_unistring": True,
@@ -70,8 +70,12 @@ class PodofoConan(ConanFile):
             self.requires("openssl/[>=1.1 <4]")
         if self.options.with_libidn:
             self.requires("libidn/1.36")
-        if self.options.with_jpeg:
+        if self.options.with_libjpeg == "libjpeg":
             self.requires("libjpeg/9e")
+        elif self.options.with_libjpeg == "libjpeg-turbo":
+            self.requires("libjpeg-turbo/3.0.2")
+        elif self.options.with_libjpeg == "mozjpeg":
+            self.requires("mozjpeg/4.1.5")			
         if self.options.with_tiff:
             self.requires("libtiff/4.6.0")
         if self.options.with_png:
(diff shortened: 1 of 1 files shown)


recipes/podofo/all/conanfile.py
```diff
@@ -131,6 +135,27 @@ def package_info(self):
...
+        if self.options.with_png:
+            self.cpp_info.requires.append("libpng::libpng")
+        if self.options.with_unistring:
+            self.cpp_info.requires.append("libunistring::libunistring")
+        if self.options.with_libjpeg == "libjpeg":
+            self.cpp_info.requires.append("libjpeg::libjpeg")
+        elif self.options.with_libjpeg == "libjpeg-turbo":
+            self.cpp_info.requires.append("libjpeg-turbo::libjpeg-turbo")
+        elif self.options.with_libjpeg == "mozjpeg":
+            self.cpp_info.requires.append("mozjpeg::mozjpeg")
```
AbrilRBS, line 158: This should not be necessary, if no requires are explicitly set, Conan will automatically deduce which ones to use based on the present requirements, which simplifies this method a lot :)

recipes/podofo/all/conanfile.py
```diff
@@ -36,7 +36,7 @@ class PodofoConan(ConanFile):
         "threadsafe": True,
         "with_openssl": True,
         "with_libidn": True,
-        "with_jpeg": True,
```
AbrilRBS, line 39: This would be a breaking change for users already setting `&:with_jpeg=True/False`, as the option won't exist anymore.

The proper way to move forward is to keep the old option name,
and have its possible values be `[True, False, "libjpeg", "libjpeg-turbo", "mozjpeg"]`,
where being `True` would be an alias for `libjpeg`, something like:

```py
@property
def _with_jpeg(self):
    return {True: "libjpeg"}.get(self.options.with_jpeg, self.options.with_jpeg)
```

(Or a variation of it as you see fit!)

AbrilRBS (commented):
Thanks, some comments on how to better approach this, but otherwise is a good start :)

I do have one request, could you also pot succesful compilation logs for each of the jpeg options for future traceability? Thanks!

recipes/podofo/all/conanfile.py
geraldgg: I remember that conan was complaining about these missing info here. All these cpp.info.requires can be removed?
I'll give it a try.

geraldgg:
Here are the build logs with 3 different jpeg libs
[
[podofo_with_libjpeg-mozjpeg.log](https://github.com/user-attachments/files/19978763/podofo_with_libjpeg-mozjpeg.log)
[podofo_with_libjpeg-True.log](https://github.com/user-attachments/files/19978764/podofo_with_libjpeg-True.log)
[podofo_with_libjpeg-turbo.log](https://github.com/user-attachments/files/19978765/podofo_with_libjpeg-turbo.log)
](url)

recipes/podofo/all/conanfile.py
```diff
@@ -131,7 +135,4 @@ def package_info(self):
         pkg_config_name = f"libpodofo-{podofo_version.major}" if podofo_version < "0.9.7" else "libpodofo"
         self.cpp_info.set_property("pkg_config_name", pkg_config_name)
         self.cpp_info.libs = ["podofo"]
-        if self.settings.os == "Windows" and self.options.shared:
```
AbrilRBS, line 134: This seems to be necessary still, no need to remove it

AbrilRBS:
Thanks for providing the logs of the compilations, it's quite useful :)

perseoGI:
Greetings @geraldgg,
I've been working on Podofo's recipe. See #29047.

We have added the option you are proposing with multiple values of `jpeg` (`libjpeg`, `libjpeg-turbo`, `mozjpeg`). 

In the meantime, you can achieve the same behavior with a `replace_requires` in your profile:

`podofo-jpeg.profile`
```
[replace_requires]
libjpeg/[>=9e]:libjpeg-turbo/[>=3.0.2 <4]

# or
# libjpeg/[>=9e]:mozjpeg/[>=4.1.5 <5]
```
This will tell Conan to replace any `libjpeg` with `libjpeg-turbo` or `mozjpeg` without having to change the recipe.

**TLDR**: Can be achieved with a profile:
```sh
$ conan create recipes/podofo/all --version 1.0.3 -b missing -o 'libtiff/*:jpeg=libjpeg-turbo' -pr default -pr misc/podofo-jpeg.profile`

geraldgg:
> Greetings @geraldgg, I've been working on Podofo's recipe. See #29047.

yeah, I saw that, thanks !

> In the meantime, you can achieve the same behavior with a `replace_requires` in your profile:

oh, great! I didn't know about that one. That will be very useful.

PR #27256, opened 2025-04-17
kickcat: add version v2.3-rc3

trns1997:
### Summary
Changes to recipe:  **kickcat/ v2.3-rc3**

---

### DIFF
--- a/recipes/kickcat/all/conandata.yml
+++ b/recipes/kickcat/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "v2.3-rc3":
+    url: "https://github.com/leducp/KickCAT/archive/refs/tags/v2.3-rc3.zip"
+    sha256: "8b529a39ace4565fd2d26baeb11cf56add3949a56969608c1c106f03ac316968"
   "v2.0-rc1":
     url: "https://github.com/leducp/KickCAT/archive/refs/tags/v2.0-rc1.zip"
     sha256: "3feec08fd2718d286a041cb83266479840f7739c1203b4450375ea1f59f234c8"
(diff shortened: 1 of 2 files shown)


leducp (approved):

jcar87:
Out of curiosity - what is the release strategy for this library?
We don't tend to add pre-releases - find it a bit odd that every version after 1.0.0 seems to be a release candidate - are there actual releases?

leducp:
Well, it depends on the POV, but this library is currently used by our medical device (exoskeleton) and is as such under really long verification and validation process. 
We see releases as final version that follow all our internal criteria for critical softwares and it can takes months/years, and we do not 'release' intermediate rc if we cannot prove them under this processus. However, we do use release candidate as 'release with less proof' so yes, it is a release :)

jcar87:
Does that mean that no version since October 2021 has actually passed this rigorous validation process?
Just trying to understand - when we include things in Conan Center, we assume they have wide use or there is wider interest in the community, and we tend to avoid including pre-release software.

leducp:
I think it is just a naming convention that differ: we put in release candidate 'production ready code' and we put them in release when they are 'safety critical production ready'. 
I understand that this naming scheme is not really communicative outside of the company... I can push an evolution in naming for our open source projects if you think it worth it, maybe something like x.y.z and x.y.z-safety? If you have insight for the naming they are welcomed as we don't have many outside perspective.

jcar87:
HI @trns1997  - would it make sense to publish v2.3 instead (which I think is the same as v2.3-rc6) - since that one is marked as passing the safety tests?

trns1997:
@jcar87, yup we're on it. With @leducp we plan to create a new release with no rc version, I will modify this PR for the new release. I will leave it as draft for now :)

trns1997:
Closing, i will create a new PR with the new release in due time

PR #27267, opened 2025-04-20
msgpack23: add version 3.0

rwindegger:
### Summary
Changes to recipe:  **msgpack23/3.0**

#### Motivation
Update the version to 3.0

#### Details
There where some API breaking changes mainly to use an `std::output_iterator` instead of a `std::vector` for serialization

---


recipes/msgpack23/v3.0/conanfile.py
```diff
@@ -0,0 +1,61 @@
+from conan import ConanFile
```
AbrilRBS, line 1: Why is this new version on a different recipe? Seems like the old one is identical to this (As far as I can see!) so we could reuse it just fine

rwindegger: The sample project is different. As the old version one will not compile. And I figured that's the way to make the change without loosing v2.1.

AbrilRBS: I've made the necessary changes to use the same recipe with different code for the test package, thanks for the info!

AbrilRBS (dismissed):

rwindegger:
Thanks for the changes @AbrilRBS.

uilianries (approved):
@rwindegger Hello! Thank you for your PR!

#### Reviewer Notes
- Version 3.1 is the latest: https://github.com/rwindegger/msgpack23/releases
- Version 3.x comes with BUILD_EXAMPLE=ON (default): https://github.com/rwindegger/msgpack23/blob/v3.1/CMakeLists.txt#L81
- Test package follows the new API

Tested locally on Linux (GCC15 + C++23): [msgpack23-3.1-linux-amd64-gcc15-cppstd23-release.log](https://github.com/user-attachments/files/28502863/msgpack23-3.1-linux-amd64-gcc15-cppstd23-release.log)

#### Maintainer Changes
- Replaced version 3.0 by 3.1 - Both version 2.1 and 3.0 don't work work gcc15 due missing utility header, tested locally: https://gist.github.com/uilianries/d9e0963c379ff9c6e30158dd0b4360b1. Fixed later by 3.1: https://github.com/rwindegger/msgpack23/commit/94bc2242cc6a5423b1b3f19b6156c01e0355bc78
- Fix bad URL attribute in the recipe to follow CCI url instead
- Added `BUILD_EXAMPLES` to avoid building examples by default
- Simplified test package to multiple files: Only validate header inclusion, not a real example.

Regards!

PR #27272, opened 2025-04-22
[ripper37-libbase] Add new recipe (v1.1.1)

RippeR37:
### Summary

This commit adds a new recipe for `libbase` library. To avoid name clashes and match name with `vcpkg` I've chosen to use `ripper37-libbase` recipe name.

Tested locally on Linux and Windows with different modules combinations.

#### Motivation

Publish new package for easier importing into projects for its users.

#### Details

I'm the author of `libbase` project that is a reimplementation of `//base` module from Chromium. I've recently updated the project and currently aim to make it easier to use for everyone by improving CMake build/installation and publishing it to vcpkg and Conan. The library is modular and allows its users to select which parts of it they want to reduce needed dependencies.

----

  - Tested on Windows and Linux platforms with all applicable configurations:
  - `&:module_net=[True|False]` x `&:module_win=[True/False]` x `&:module_wx=[True/False]`

### DIFF
--- a/recipes/ripper37-libbase/all/conandata.yml
+++ b/recipes/ripper37-libbase/all/conandata.yml
@@ -3,3 +3,9 @@ sources:
     url: "https://github.com/RippeR37/libbase/archive/refs/tags/v1.1.1.tar.gz"
     sha256: "d3e7ab81ad79b98b29087fe0bfb0a3c6ac2aa72cfaa06778276445a8b76f2f83"
     strip_root: true
+patches:
+  "1.1.1":
+    - patch_file: "patches/1.1.1-0001-glog-0.7.1-msvc-2019-compile-fix.patch"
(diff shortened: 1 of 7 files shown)


RippeR37:
Updated the PR with a fixup that adds a patch to recipe. The patch is backport of a commit that I applied just after release v1.1.1. It turns out that Conan CI uses older MSVC that isn't/wasn't strictly supported by the library but the issue was in `glog` headers so changing 2 warnings back to "warning" status was enough to make it work there.

This is already landed in the upstream so in the next version the patch would be removed, but for now I don't want to (re-)release new version just for this small fix so I kept it as a patch here. Tested different module combinations and they work fine (albeit on MSVC2022*).

Please let me know if I can do anything more here or if its ready to be tested and verified :)

PR #27279, opened 2025-04-22
nasm: use GitHub as download location after nasm.us being offline

DeepDiver1975:
### Summary
Changes to recipe:  **application/nasm**

#### Motivation
nasm.us is down ... get the sources from GitHub

#### Details
Change URL and hashes

---

### DIFF
--- a/recipes/nasm/all/conandata.yml
+++ b/recipes/nasm/all/conandata.yml
@@ -1,19 +1,22 @@
 sources:
-  "2.16.01":
-    sha256: "c77745f4802375efeee2ec5c0ad6b7f037ea9c87c92b149a9637ff099f162558"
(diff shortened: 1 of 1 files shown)


jcar87:
Closed via https://github.com/conan-io/conan-center-index/pull/27295

nasm.dev seems to be up, providing the same hashes, while they restore the original website

Please note that we won't normally accept PRs that change checksums for already published versions unless there's a detailed report that explains the checksum change (by comparing the contents..)

DeepDiver1975:
> nasm.dev seems to be up, providing the same hashes, while they restore the original website

now both are down .... :shrug: 



> Please note that we won't normally accept PRs that change checksums for already published versions unless there's a detailed report that explains the checksum change (by comparing the contents..)

The upstream project is not publishing the same source packages in GitHub as they do on their website. Also the build has to change when using the code from GitHub ....

PR #27293, opened 2025-04-23
bump up spdlog-1.15.2 release

pgeler:
### Summary
Changes to recipe:  spdlog/1.15.2

#### Motivation
bump up to latest release

#### Details
replaced 1.15.1 to 1.15.2 as it not been used in other packages

---

### DIFF
--- a/recipes/spdlog/all/conandata.yml
+++ b/recipes/spdlog/all/conandata.yml
@@ -1,7 +1,7 @@
 sources:
-  "1.15.1":
-    url: "https://github.com/gabime/spdlog/archive/v1.15.1.tar.gz"
-    sha256: "25c843860f039a1600f232c6eb9e01e6627f7d030a2ae5e232bdd3c9205d26cc"
+  "1.15.2":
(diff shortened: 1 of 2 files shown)


chausner:
I think removing versions that have already been published on Conan Center is strongly discouraged (except maybe for very old versions that cannot be supported anymore). Even if there are no other packages in Conan Center that depend on this package version, users could still have a direct dependency on it, and removing it from Conan Center would break builds for those users.

pgeler:
> I think removing versions that have already been published on Conan Center is strongly discouraged (except maybe for very old versions that cannot be supported anymore). Even if there are no other packages in Conan Center that depend on this package version, users could still have a direct dependency on it, and removing it from Conan Center would break builds for those users.

links are welcome for such discouragement. I've seen different opinions around this topic as well as conan-center do not just delete packages, and users are able to continue consuming them, while at the same time cleanup actually saves time for rebuild and elimination of the minor fixes

oleksandr-prykhodko-quickbase (approved):

jcar87 (approved):
LGTM, thanks!

PR #27299, opened 2025-04-24
Add vulkan memory allocator 3.2.1

elvisdukaj:
### Summary
Changes to recipe:  **vulkan-memory-allocator/3.2.1**

#### Motivation

Adding a new version 3.2.1, which also supports Vulkan 1.4 (from 3.2.0).

#### Details

`vulkan-memory-allocator` shouldn't be locked to a specific version of Vulkan. [See here](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/blob/master/include/vk_mem_alloc.h#L135-L147):

The supported Vulkan version are from Vulkan 1.0. Note that Vulkan always preserve backward compatibility, so version 1.4 is compatible with 1.3.

```C++
#if !defined(VMA_VULKAN_VERSION)
    #if defined(VK_VERSION_1_4)
        #define VMA_VULKAN_VERSION 1004000
    #elif defined(VK_VERSION_1_3)
        #define VMA_VULKAN_VERSION 1003000
    #elif defined(VK_VERSION_1_2)
        #define VMA_VULKAN_VERSION 1002000
    #elif defined(VK_VERSION_1_1)
        #define VMA_VULKAN_VERSION 1001000
    #else
        #define VMA_VULKAN_VERSION 1000000
    #endif
#endif
```

I tested locally with clang-20 on Ubuntu 24.04.

---
- [X[ Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)

### DIFF
--- a/recipes/vulkan-memory-allocator/all/conandata.yml
+++ b/recipes/vulkan-memory-allocator/all/conandata.yml
@@ -2,6 +2,9 @@ sources:
   "cci.20231120":
     url: "https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/archive/5e43c795daf43dd09398d8307212e85025215052.tar.gz"
     sha256: "1ee9922fb059bc3b1dc5bbf71020e7a590b6ceb27fc3025d746e15be53fe31b7"
+  "3.2.1":
+    url: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/archive/refs/tags/v3.2.1.zip
+    sha256: "46d5f8c529b9102f2d3dd933beba38f3873a59896b249a90dcad009ecca4bfff"
   "3.0.1":
     url: "https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/archive/refs/tags/v3.0.1.tar.gz"
     sha256: "2a84762b2d10bf540b9dc1802a198aca8ad1f3d795a4ae144212c595696a360c"
--- a/recipes/vulkan-memory-allocator/all/conanfile.py
+++ b/recipes/vulkan-memory-allocator/all/conanfile.py
@@ -29,7 +29,7 @@ class VulkanMemoryAllocatorConan(ConanFile):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("vulkan-headers/1.3.243.0")
+        self.requires("vulkan-headers/[>=1, <2]")
 
     def package_id(self):
         self.info.clear()


recipes/vulkan-memory-allocator/all/conanfile.py
```diff
@@ -29,7 +29,7 @@ def layout(self):
         basic_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("vulkan-headers/1.3.243.0")
+        self.requires("vulkan-headers/[>=1, <2]")
```
amitiunin, line 32: Incorrect syntax, it seems: according to [Conan docs](https://docs.conan.io/2/tutorial/versioning/version_ranges.html), there shouldn't be a comma.

The range you specified can also be replaced with a tilde operator:

```suggestion
        self.requires("vulkan-headers/[~1]")
```

elvisdukaj: Yes the comma is needed, I forgot to push the lady commit after I saw a green build 🙈

amitiunin:
You'll also need to add a record in the file `config.yml`; take a look at other PRs that bump a package's version.

elvisdukaj:
Thank you, I missed that out! I'll fix it!

recipes/vulkan-memory-allocator/all/conandata.yml
```diff
@@ -2,6 +2,9 @@ sources:
   "cci.20231120":
```
elvisdukaj, line 2: Most probably this cci.* version should be removed.

amitiunin (approved):

uilianries (approved):
@elvisdukaj Thank you for your PR!

I made a change related to CMake Target in your PR. In this new version that you are including in ConanCenterIndex, the upstream is explicit about the target config file and targets provided in its [CMakeLists.txt](https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/compare/v3.0.1...v3.2.1#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a)

It's pretty clear Vulkan headers 1.x will work for 3.x as commented in the features list: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator#features. 

The VMA 2.x only supports 1.0 and 1.1, not newer in the vk_mem_alloc.h and README, but I didn't see any error when testing.

jcar87 (approved):
Thanks! Apologies for the delay

Combined with https://github.com/conan-io/conan-center-index/pull/27603 to add the most recent version

PR #27305, opened 2025-04-25
[libidn2] Fix cross-building on Windows to x86

### LINKED ISSUE #27192: libidn2: build error on windows
> Unfortunately, libidn2 doesn't compile with MSVC :(
> 
> ```
> .\limits.h(36): fatal error C1083: Cannot open include file: '': No such file or directory
> ```
> I'll have to create a new issue, this time for the libidn2 package... 

 _Originally posted by @zenden2k in [#27152](https://github.com/conan-io/conan-center-index/issues/27152#issuecomment-2788291140)_

uilianries:
### Summary
Changes to recipe:  **libidn2/2.3.0**

#### Motivation

I'm capable to reproduce the reported issues as well, only when cross-building from x86_64 to x86 on Windows: [libidn2-8.2.0-windows-error.log](https://github.com/user-attachments/files/19906266/libidn2-8.2.0-windows-error.log)

The error does not occur when building to x86_64: [libidn2-8.2.0-windows-x86_64.log](https://github.com/user-attachments/files/19906272/libidn2-8.2.0-windows-x86_64.log)

fixes #27192

#### Details

The upstream recommends and uses `i686-w64-mingw32` when cross-building:

- https://gitlab.com/libidn/libidn2/-/blob/v2.3.3/.gitlab-ci.yml?ref_type=tags#L440
- https://gitlab.com/libidn/libidn2/-/tree/2.3.0?ref_type=tags#cross-compiling

Locally, I can build to x86 without errors when using this PR, you can check my build log:

[libidn2-2.3.0-windows-cross.log](https://github.com/user-attachments/files/19906242/libidn2-2.3.0-windows-cross.log)

Then, I checked the produced library using dumpbin to make sure it's x86: [libidn2-headers.log](https://github.com/user-attachments/files/19906246/libidn2-headers.log)

----

Still, when building multiple versions, I found since the version 2.3.5, there is no need of using --host, it works automatically. Doing a diff with 2.3.4 I found the follow diff in the configure:

diff

This change is not enough to fix 2.30, but starts to clarify why `i686-unknown-windows` works for 2.3.5 but not for older versions. The triplet `i686-unknown-windows` is defined by AutotoolsToolchain when cross-building to x86.

A second option would be add version 2.3.8 that works for cross-building without workarounds. 

---

### DIFF
--- a/recipes/libidn2/all/conanfile.py
+++ b/recipes/libidn2/all/conanfile.py
@@ -87,6 +87,22 @@ class LibIdn(ConanFile):
             "--disable-nls",
             "--disable-rpath",
         ]
+
+        if cross_building(self) and is_msvc(self):
+            # INFO: AutotoolsToolchain uses i686-unknown-windows for x86, causing error when building:
+            #   fatal error C1083: Cannot open include file: '': No such file or directory
+            # The upstram recommends using i686-w64-mingw32 when cross-compiling for x86:
+            #   https://gitlab.com/libidn/libidn2/-/blob/master/CONTRIBUTING.md#cross-compiling
+            triplet_arch_windows = {"x86_64": "x86_64", "x86": "i686", "armv8": "aarch64"}
+            host_arch = triplet_arch_windows.get(str(self.settings.arch))
+            build_arch = triplet_arch_windows.get(str(self.settings_build.arch))
+            host = self.conf.get("tools.gnu:host_triplet", f"{host_arch}-w64-mingw32")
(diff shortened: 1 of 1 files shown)


uilianries:
> A second option would be add version 2.3.8 that works for cross-building without workarounds.

It will be better bumping the project version instead of keeping --host and --build with pre-defined values, as it may have side-effects and break users that are cross-building already. 

I'll open a new PR with a bump version.

PR #27311, opened 2025-04-27
arrow: add version 20.0.0

assignUser:
### Summary
Changes to recipe:  **arrow/20.0.0**

#### Motivation
Arrow 20.0.0 was released 2025/04/26. This PR updates conan's index with the new release.

#### Details
We fully script these changes and they're minimal so this should be easy to review.

---

### DIFF
--- a/recipes/arrow/all/conandata.yml
+++ b/recipes/arrow/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "20.0.0":
+    url: "https://www.apache.org/dyn/closer.lua/arrow/arrow-20.0.0/apache-arrow-20.0.0.tar.gz?action=download"
+    sha256: "89efbbf852f5a1f79e9c99ab4c217e2eb7f991837c005cba2d4a2fbd35fad212"
   "19.0.1":
     url: "https://www.apache.org/dyn/closer.lua/arrow/arrow-19.0.1/apache-arrow-19.0.1.tar.gz?action=download"
     sha256: "acb76266e8b0c2fbb7eb15d542fbb462a73b3fd1e32b80fad6c2fafd95a51160"
--- a/recipes/arrow/config.yml
+++ b/recipes/arrow/config.yml
@@ -1,4 +1,6 @@
 versions:
+  "20.0.0":
+    folder: all
   "19.0.1":
     folder: all
   "18.1.0":


assignUser:
```
  CMake Error at C:/workspace/cci_prod_PR-27311/conan-home/p/b/arrow8a209707c80a8/b/build/generators/conan_toolchain.cmake:43 (message):
    The CMake policy CMP0091 must be NEW, but is ''
```
We set  that policy to new in our CMake so I am not sure where this comes from.

assignUser:
Thanks for the patches @AbrilRBS I will make sure to upstream those once this is merged!

I only see the CMP error that happened before. I don't think that's on our end.

pitrou:
> I only see the CMP error that happened before. I don't think that's on our end.

It happens while building mimalloc, right? Or am I misreading?

assignUser:
@pitrou yeah that's right, thanks for the pointer. In 19.0.1 a patch was added for it.

raulcd:
@assignUser I think the patch requires to be regenerated for the 20.0.0 version?
seems to fail to apply probaly due to line changes as  it seems the starting line isn't correct:
```
arrow/20.0.0: Uncompressing apache-arrow-20.0.0.tar.gz to .
arrow/20.0.0: Apply patch (conan): use cci package
arrow/20.0.0: WARN: /home/conan/workspace/cci_prod_PR-27311/conan-home/p/arrow0b605cf25c1ed/s/patches/20.0.0-0001-fix-downloaded-mimalloc.patch: file 1/1:	 b'cpp/cmake_modules/ThirdpartyToolchain.cmake'
arrow/20.0.0: WARN: /home/conan/workspace/cci_prod_PR-27311/conan-home/p/arrow0b605cf25c1ed/s/patches/20.0.0-0001-fix-downloaded-mimalloc.patch:  hunk no.1 doesn't match source file at line 2259
arrow/20.0.0: WARN: /home/conan/workspace/cci_prod_PR-27311/conan-home/p/arrow0b605cf25c1ed/s/patches/20.0.0-0001-fix-downloaded-mimalloc.patch:   expected: b'# mimalloc - Cross-platform high-performance allocator, from Microsoft'
arrow/20.0.0: WARN: /home/conan/workspace/cci_prod_PR-27311/conan-home/p/arrow0b605cf25c1ed/s/patches/20.0.0-0001-fix-downloaded-mimalloc.patch:   actual  : b''
arrow/20.0.0: /home/conan/workspace/cci_prod_PR-27311/conan-home/p/arrow0b605cf25c1ed/s/patches/20.0.0-0001-fix-downloaded-mimalloc.patch: source file is different - b'cpp/cmake_modules/ThirdpartyToolchain.cmake'
ERROR: arrow/20.0.0: Error in source() method, line 256
	self._patch_sources()
while calling '_patch_sources', line 380
	apply_conandata_patches(self)
	ConanException: Failed to apply patch: /home/conan/workspace/cci_prod_PR-27311/conan-home/p/arrow0b605cf25c1ed/s/patches/20.0.0-0001-fix-downloaded-mimalloc.patch
```
Regenerating from the apache-arrow-20.0.0 tag gives me this new patch:
```
diff --git a/cpp/cmake_modules/ThirdpartyToolchain.cmake b/cpp/cmake_modules/ThirdpartyToolchain.cmake
index 63bdd4ab76..9744f01b1e 100644
--- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
+++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
@@ -2268,6 +2268,10 @@ endif()
 # mimalloc - Cross-platform high-performance allocator, from Microsoft
 
 if(ARROW_MIMALLOC)
+ find_package(mimalloc REQUIRED CONFIG)
+endif()
+
+if(0)
   if(NOT ARROW_ENABLE_THREADING)
     message(FATAL_ERROR "Can't use mimalloc with ARROW_ENABLE_THREADING=OFF")
   endif()
```

recipes/arrow/all/conanfile.py
```diff
@@ -152,7 +152,7 @@ def requirements(self):
         if self.options.with_jemalloc:
             self.requires("jemalloc/5.3.0")
         if self.options.with_mimalloc:
-            self.requires("mimalloc/1.7.6")
+            self.requires("mimalloc/2.0.6")
```
franramirez688: Do we really need to bump this version? Is it related to adding this new arrow version?

franramirez688 (commented):
Thanks @assignUser for this PR! 🙏  

LGTM! Only a question before approving

recipes/arrow/all/conanfile.py
```diff
@@ -246,12 +245,15 @@ def validate(self):
             raise ConanInvalidConfiguration("arrow:parquet requires arrow:with_thrift")
 
     def build_requirements(self):
-        if Version(self.version) >= "13.0.0":
+        if Version(self.version) >= "20.0.0":
+            self.tool_requires("cmake/[>=3.25 <4]")
+        elif Version(self.version) >= "13.0.0":
```
franramirez688:
```suggestion
        else:
```

recipes/arrow/all/conanfile.py
assignUser: I just bumped it to what's currently in use as it's a major version difference. But that version has been in use for years, so we should be save reverting this.

pitrou:
Why keep an old version though?

Le 13 juin 2025 17:01:06 GMT+02:00, Jacob Wujciak-Jens ***@***.***> a écrit :

Issue #27321, opened 2025-04-29
[libcurl]  libcurl/8.12.1 fails to build with "autoreconf: command not found"


datalogics-robb:
### Description

When building libcurl (on MacOS with Xcode 15 in this case) - the recipe tool_requires libtool, but can not find the path to `autoreconf` because that is in the `autoconf` package.

Conan 1 leaks indirect dependencies down to clients, polluting build environments with unnecessary include paths, etc.  Looks like it did the same with build tools, but this is the first example I have encountered.  i.e. This was not a problem in conan 1 as a result of bugs in conan 1.

### Package and Environment Details

* Package Name/Version: **libcurl/8.12.1**
* Operating System+version: **MacOS 14.x**
* Compiler+version: **apple-clang 15**
* Conan version: **conan 2.15.1**
* Python version: **Python 3.11**

### Conan profile

[Profile host:

[settings]

arch=x86_64

build_type=Release

compiler=apple-clang

compiler.cppstd=17

compiler.libcxx=libc++

compiler.version=15.0

os=Macos

os.sdk=macosx

os.sdk_version=14.0

os.subsystem=None

os.version=10.15

[options]

boost/*:debug_level=1

boost/*:i18n_backend_iconv=off

boost/*:i18n_backend_icu=True

libtiff/*:cxx=True

libtiff/*:jbig=False

libtiff/*:jpeg=libjpeg

libtiff/*:libdeflate=False

libtiff/*:lzma=False

libtiff/*:webp=False

libtiff/*:zlib=True

libtiff/*:zstd=False

[buildenv]

MACOSX_DEPLOYMENT_TARGET=10.15

Profile build:

[settings]

arch=x86_64

build_type=Release

compiler=apple-clang

compiler.cppstd=17

compiler.libcxx=libc++

compiler.version=15.0

os=Macos

os.sdk=macosx

os.sdk_version=14.0

os.subsystem=None

os.version=10.15

[options]

boost/*:debug_level=1

boost/*:i18n_backend_iconv=off

boost/*:i18n_backend_icu=True

libtiff/*:cxx=True

libtiff/*:jbig=False

libtiff/*:jpeg=libjpeg

libtiff/*:libdeflate=False

libtiff/*:lzma=False

libtiff/*:webp=False

libtiff/*:zlib=True

libtiff/*:zstd=False

[buildenv]

MACOSX_DEPLOYMENT_TARGET=10.15

### Steps to reproduce

[...]

PR #27322, opened 2025-04-29
Fix issue#27321 building libcurl

datalogics-robb:
Add a tool_requires call for autoconf - builds are failing to find `autoreconf` which is provided by the autoconf package.

https://github.com/conan-io/conan-center-index/issues/27321

### Summary
Changes to recipe:  **libcurl/8.12.1**

#### Motivation

Libcurl's build needs autoconf executables to work - add tool_require for Autoconf provides them

#### Details

Conan 1.x seemed to leak the indirect dependencies down - which is why in Conan 1, tool_requires for libtool would have provided the paths for Autoconf - in Conan 2  this does not (and should not) happen, so it should be explicitly added.

---

### DIFF
--- a/recipes/libcurl/all/conanfile.py
+++ b/recipes/libcurl/all/conanfile.py
@@ -209,6 +209,7 @@ class LibcurlConan(ConanFile):
             if self._is_win_x_android:
                 self.tool_requires("ninja/[>=1.10.2 <2]")
         else:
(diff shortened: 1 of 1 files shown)


datalogics-robb:
This did work around an issue we were seeing, but I'm not convinced this is the right thing to do - I'd love to hear the maintainers' comments though.

PR #27331, opened 2025-04-30
SDL: package java sources in Android builds

kambala-decapitator:
### Summary
Changes to recipe:  **sdl/all**

#### Motivation

For Android, SDL also provides java platform code that must be built by an app using it. If you copy this code to your project, this means you must always synchronise it with SDL version used, which for Conan means that version range can't be used. By providing the java sources as part of Conan package the synchronisation is achieved automatically and you can easily use version range, then you can inject this path to Gradle via a property.

Btw Qt uses a similar approach, but it creates respective Gradle property automatically when using `androiddeployqt` tool.

#### Details

The new `src-java` directory will contain java files in `org/libsdl/app` directory.

Copying the directory structure properly via `mkdir`+`copy` tools would be way more verbose, so I used the `shutil.copytree` shortcut.

---

### DIFF
--- a/recipes/sdl/3.x/conanfile.py
+++ b/recipes/sdl/3.x/conanfile.py
@@ -7,6 +7,7 @@ from conan.tools.microsoft import is_msvc
 from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout, CMakeDeps
 
 import os
+import shutil
 
 required_conan_version = ">=2"
 
@@ -337,6 +338,12 @@ class SDLConan(ConanFile):
 
     def package(self):
         copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        if self.settings.os == "Android":
+            # too much hassle via mkdir+copy due to amount of subdirectories
+            shutil.copytree(
+                os.path.join(self.source_folder, "android-project", "app", "src", "main", "java"),
+                os.path.join(self.package_folder, "src-java"))
+
         cmake = CMake(self)
         cmake.install()
 
--- a/recipes/sdl/all/conanfile.py
+++ b/recipes/sdl/all/conanfile.py
@@ -8,6 +8,7 @@ from conan.tools.cmake import CMake, CMakeToolchain, cmake_layout
 from conan.tools.env import Environment
 
 import os
+import shutil
 
 required_conan_version = ">=1.55.0"
 
@@ -353,6 +354,11 @@ class SDLConan(ConanFile):
         rmdir(self, os.path.join(self.package_folder, "lib", "pkgconfig"))
         rmdir(self, os.path.join(self.package_folder, "libdata"))
         rmdir(self, os.path.join(self.package_folder, "share"))
+        if self.settings.os == "Android":
+            # too much hassle via mkdir+copy due to amount of subdirectories
+            shutil.copytree(
+                os.path.join(self.source_folder, "android-project", "app", "src", "main", "java"),
+                os.path.join(self.package_folder, "src-java"))
 
     def package_info(self):
         self.cpp_info.set_property("cmake_file_name", "SDL2")


AbrilRBS (commented):
Hi @kambala-decapitator thanks for your PR

A few things for this PR:
* We might want to restrict this change to SDL 3.x if possible
* How did you test this? Could we get some logs for the 3.x version?
* Using Conan's `copy` should be enough using patterns and `keep_path` argument

Thanks!

kambala-decapitator:
Hi @AbrilRBS 

> We might want to restrict this change to SDL 3.x if possible

currently our project uses 2.x and can't easily transition to 3.x. Would be great to have it for 2.x as well. Besides, this change doesn't actually require binary package rebuilds as the CI doesn't build Android binaries.

> How did you test this? Could we get some logs for the 3.x version?

Only built SDL and verified that the java sources were copied to the expected directory. Would you like to see a usage example with and android app?

[build_log.txt](https://github.com/user-attachments/files/20291203/build_log.txt)

```
> ll -R /Users/kambala/.conan2/p/b/sdl90322e2cce616/p/src-java
total 0
drwxr-xr-x  3 kambala  staff    96B May 19 14:56 org/

/Users/kambala/.conan2/p/b/sdl90322e2cce616/p/src-java/org:
total 0
drwxr-xr-x  3 kambala  staff    96B May 19 14:56 libsdl/

/Users/kambala/.conan2/p/b/sdl90322e2cce616/p/src-java/org/libsdl:
total 0
drwxr-xr-x  13 kambala  staff   416B May 19 14:56 app/

/Users/kambala/.conan2/p/b/sdl90322e2cce616/p/src-java/org/libsdl/app:
total 440
-rw-rw-r--  1 kambala  staff   597B May 26  2023 HIDDevice.java
-rw-rw-r--  1 kambala  staff    23K Dec 23 23:55 HIDDeviceBLESteamController.java
-rw-rw-r--  1 kambala  staff    27K Dec 23 23:55 HIDDeviceManager.java
-rw-rw-r--  1 kambala  staff   9.3K Nov  7  2024 HIDDeviceUSB.java
-rw-rw-r--  1 kambala  staff   3.4K Jan 11 02:27 SDL.java
-rw-rw-r--  1 kambala  staff    79K Mar  2 05:30 SDLActivity.java
-rw-rw-r--  1 kambala  staff   4.6K Jul 29  2024 SDLAudioManager.java
-rw-rw-r--  1 kambala  staff    30K Jan 22 08:33 SDLControllerManager.java
-rw-rw-r--  1 kambala  staff   2.3K Jan 11 02:27 SDLDummyEdit.java
-rw-rw-r--  1 kambala  staff   4.8K Jan 15 06:05 SDLInputConnection.java
-rw-rw-r--  1 kambala  staff    14K Feb 14 22:50 SDLSurface.java
```

> Using Conan's copy should be enough using patterns and keep_path argument

You're right! Initially I've used `copy` incorrectly: have been passing the relative path to `pattern`, but should've passed it to `src` instead. Pushed the fix.

kambala-decapitator:
> You could expose an env var in the package_info method

@AbrilRBS would this be acceptable?

```py
    java_src_dir = os.path.join(self.package_folder, "src-java")
    self.buildenv_info.define("ORG_GRADLE_PROJECT_SDL_JAVA_SRC", java_src_dir)
    self.cpp_info.set_property("src_java", java_src_dir)
```

when gradle finds an env var named `ORG_GRADLE_PROJECT_<something>`, it creates project property named `<something>`.

P.S. See the comment above for an example of building android app using SDL Java sources from the Conan package.

perseoGI:
Subject: Request for Assistance with Android Builds

Hello @kambala-decapitator,

I apologize for the delay in my response; I have been quite busy. 

I am not very familiar with Android builds, so it would be extremely helpful for our team to have a working basic example. This will also help you and other Conan users understand the benefits of the changes introduced in your PR.

The maintainers of the Conan Center Index prioritize security, backward compatibility, and ensuring that existing behavior is not disrupted. This caution is why we are hesitant to introduce changes that we do not fully understand, especially in areas where our knowledge is limited.

Thank you very much for your understanding, and I look forward to hearing from you soon! 🐸

kambala-decapitator:
Hi @perseoGI 

I've created [repo](https://github.com/kambala-decapitator/sdl-conan-android) with demo project based on SDL's [primitives example](https://github.com/libsdl-org/SDL/tree/main/examples/renderer/02-primitives). Let me explain how things work and how to repeat all the steps (also check repo commits).

1. I started with raw CMake project and ensured that it works on desktop (macOS).
2. Then I created new project in Android Studio - "No activity" template and only removed everything related to tests.
3. Now it's time to adapt CMake project to install libraries to the location Gradle expects. By default shared libraries must be installed to `src/main/jniLibs/<abi>` (but can be changed in the Gradle project), so we make sure that everything is installed in there: our app (which for android must be shared library), all the dependencies (just SDL in this case) and libc++ (since we selected shared version in the profile).
4. Android project also needs to be adapted to consume SDL Java sources from the external source - Conan package, what this PR is for. I have also applied [the aforementioned change](https://github.com/conan-io/conan-center-index/pull/27331#issuecomment-2907845986) locally to avoid passing `SDL_JAVA_SRC` property to gradle manually, but it could also be fetched using `self.dependencies.host["sdl"].cpp_info.get_property("src_java", check_type=str)` if necessary. Besides, the manifest file should have an Activity (otherwise app won't appear on desktop), and I define it to be an SDL class.
5. Now all's ready to build a working Android app.

First, I'll show all the necessary commands to be executed manually. This was tested on [one before the "last" commit](https://github.com/kambala-decapitator/sdl-conan-android/commit/b49fa0ea5bbc5fac31d75baeec8f7f618cd5d903). CWD is the repo root. We're building for aarch64 (see the profile).

```sh
# prepare Conan dependencies
conan create <PATH TO CCI>/recipes/sdl/3.x -pr profile-android64 -pr profile-options --version 3.2.6 -b missing --test-folder= -c user.sdl:android=True
conan install . -pr profile-android64 -pr profile-options -of conan-android -b never -d runtime_deploy --deployer-folder android-app/app/src/main/jniLibs/arm64-v8a

# produce shared library
cmake -S . -B build-android -G Ninja -D CMAKE_BUILD_TYPE=Release --toolchain conan-android/conan_toolchain.cmake
cmake --build build-android
cmake --install build-android --prefix android-app/app/src/main/jniLibs

# build Android project
. conan-android/conanbuild.sh
cd android-app
./gradlew assembleDebug

# install apk to the connected device (assumes that adb is in PATH)
adb install -r app/build/outputs/apk/debug/app-debug.apk
```

In the ["last" commit](https://github.com/kambala-decapitator/sdl-conan-android/commit/1d1add31d1ef8845751830ef78a6bf833b45af43) I've implemented building the project on Conan level (you only need to run `conan create` like above to obtain package from this PR):

```
conan build . -pr profile-android64 -pr profile-options -b missing -d runtime_deploy --deployer-folder android-app/app/src/main/jniLibs/arm64-v8a
```

And, like I said in the PR description, now updating to a newer SDL version becomes trivial: one simply consumes new Conan package without any extra work (unless something needs to be adjusted on the C/C++ side, of course) - no need to worry about copying new Java sources to your project.

> The maintainers of the Conan Center Index prioritize security, backward compatibility, and ensuring that existing behavior is not disrupted

to be honest I'm not sure how copying a few extra files to the package to a separate directory can disrupt something :)

perseoGI:
Greetings @kambala-decapitator, thank you very much for your detailed response!
To correctly replicate your steps I'll like to have access to your conan profiles.

You have added `conan-*` in your .gitignore and any conan profile has been uploaded!
Thank you in advance!

kambala-decapitator:
Hi @perseoGI 

> You have added conan-* in your .gitignore and any conan profile has been uploaded!
> Thank you in advance!

oops, I'm very sorry, that was quite unfortunate :) Just added them in the last commit and adjusted my above comment.

perseoGI:
Greetings @kambala-decapitator!

I’m excited to share that I successfully built the application and ran it locally on an Android emulator!



Thank you very much for taking the time to explain how to integrate a dependency into a native Android application. I will discuss your proposed changes with my colleagues, and I would like to enhance the existing Android documentation in [conan docs](https://docs.conan.io/2/examples/cross_build/android/ndk.html#examples-cross-build-android-ndk). I was the one who updated that documentation some time ago, but I didn’t have enough expertise in the field at the time.

Please feel free to open any pull requests on our documentation or provide feedback if you’d like. I believe many developers could benefit from your knowledge!

Let’s stay in touch. Thank you!

perseoGI:
Greetings @kambala-decapitator,

@AbrilRBS and I have been examining this PR and found that the upstream CMakeLists.txt already implements something similar natively; see https://github.com/libsdl-org/SDL/blob/release-3.2.6/CMakeLists.txt#L1436-L1444.

When compiling with an Android profile (`os=Android`), CMake generates a `.jar` file located at:
```
<package_folder>/share/java/SDL3
```
This jar includes the following files:
```
proguard.txt
SDL3-3.2.6-sources.jar
SDL3-3.2.6.jar
```
Upon uncompressing the jar, you can observe that the actual `.java` files are bundled as follows:
```
SDL3-3.2.6-sources
├── META-INF
│   └── MANIFEST.MF
└── org
    └── libsdl
        └── app
            ├── HIDDevice.java
            ├── HIDDeviceBLESteamController.java
            ├── HIDDeviceManager.java
            ├── HIDDeviceUSB.java
            ├── SDL.java
            ├── SDLActivity.java
            ├── SDLAudioManager.java
            ├── SDLControllerManager.java
            ├── SDLDummyEdit.java
            ├── SDLInputConnection.java
            └── SDLSurface.java
```

In Conan Center Index, we prefer to use upstream solutions (install methods) over custom tweaks, as we delegate the maintainability to the upstream library.

Do you think these generated `.jar` files could fit into your workflow? As I mentioned earlier, my Android knowledge is limited, but I believe a `.jar` should suffice.
If so, please let me know, and we can try adding your updated `buildenv` definition into the recipe to facilitate the consumer side of this recipe.

kambala-decapitator:
Hi @perseoGI 

That's a nice find! Didn't know that SDL3 already builds a jar file (as our project is still on v2). Yes, that's definitely the way to go for v3. And I also see that the jar files are already packaged in Conan to `share/java/SDL3`.

Would it be ok to proceed in the following way?

1. for v3, define variables for `buildenv_info` and `cpp_info` pointing to the aforementioned directory or jar file (dunno yet what'd be the proper way for android project) like in https://github.com/conan-io/conan-center-index/pull/27331#issuecomment-2907845986 Can also update the demo project to use this approach.
2. for v2, keep the current solution, as upstream doesn't build the jar file

perseoGI:
We are glad to see that the `.jar` file works well. How would you use that `.jar` from the consumer side? I've noticed that one way is to define a `gradle` environment variable named `ORG_GRADLE_PROJECT_sdlJarPath` and use it in Gradle like this:
```gradle
dependencies {
    implementation files(project.property('sdlJarPath'))
}
```
I haven't tested it yet. It would be helpful if you could update your project example to use SDL3 this way!

On the other hand, after discussing with the team, even though it would be more straightforward to use this library from Gradle, we shouldn't add those extra build environment variables or properties to the SDL recipe, as it can be easily managed from the consumer side and will keep the recipe cleaner.

Usage example:

```python
def generate(self):
    env = VirtualBuildEnv(self)
    venv = env.environment()
    # SDL 2.x usage
    venv.define("ORG_GRADLE_PROJECT_SDL_JAVA_SRC", os.path.join(self.dependencies.host["sdl"].package_folder, "src-java"))
    # SDL 3.x usage
    venv.define("ORG_GRADLE_PROJECT_sdlJarPath", os.path.join(self.dependencies.host["sdl"].package_folder, "share", "java", "SDL3", f'SDL3-{self.dependencies.host["sdl"].ref.version}.jar'))

    env.generate()
```

#### Conclusion

- Leave the 3.x recipe without modifications as cmake already creates `.jar` files and can be consumed with the previous example
- Update `all` recipe to copy the `.java` files to a path similar to `3.x`, something like `share/java/SDL2/`

Optional (would help a lot for us to test and maybe update our docs with your knowledge): update your git example to use the jar files!

Thank you very much!

kambala-decapitator:
Hi @perseoGI 

Adjusted PR as you suggested and also updated the demo project. Thank you for the help and suggestions!

perseoGI (approved):
Okay @kambala-decapitator, thank you very much!

It looks good to me now!
Thank you once again for your patience and kindness in creating the repository.

PR #27338, opened 2025-05-01
package(mpg123): Update to 1.33.0 to fix compilation with GCC 15

Sil3ntStorm:
### Summary
Changes to recipe:  **mpg123/1.33.0**

#### Motivation
Fixes compilation when using gcc 15.
Update to latest version, fixing several issues in the library

#### Details
Add version 1.33.0
Skip 1.32.10 since that fails to link on Windows due to using strtok_r
Remove ancient 1.26.4 since it fails on Windows using current MSVC

---

### DIFF
--- a/recipes/mpg123/all/conandata.yml
+++ b/recipes/mpg123/all/conandata.yml
@@ -14,16 +14,9 @@ sources:
       - "https://sourceforge.net/projects/mpg123/files/mpg123/1.29.3/mpg123-1.29.3.tar.bz2"
       - "https://www.mpg123.de/download/mpg123-1.29.3.tar.bz2"
     sha256: "963885d8cc77262f28b77187c7d189e32195e64244de2530b798ddf32183e847"
-  "1.26.4":
-    url:
-      - "https://sourceforge.net/projects/mpg123/files/mpg123/1.26.4/mpg123-1.26.4.tar.bz2"
-      - "https://www.mpg123.de/download/mpg123-1.26.4.tar.bz2"
-    sha256: "081991540df7a666b29049ad870f293cfa28863b36488ab4d58ceaa7b5846454"
 
 patches:
   "1.31.2":
-    - patch_file: "patches/0001-msvc-export-symbols.patch"
-  "1.26.4":
     - patch_file: "patches/0001-msvc-export-symbols.patch"
   "1.29.3":
     - patch_file: "patches/0001-msvc-export-symbols.patch"
--- a/recipes/mpg123/all/conandata.yml
+++ b/recipes/mpg123/all/conandata.yml
@@ -1,9 +1,9 @@
 sources:
-  "1.32.10":
+  "1.33.0":
     url:
-      - "https://sourceforge.net/projects/mpg123/files/mpg123/1.32.10/mpg123-1.32.10.tar.bz2"
-      - "https://www.mpg123.de/download/mpg123-1.32.10.tar.bz2"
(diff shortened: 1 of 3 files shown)


n-morales:
I would be very interested to see this merged -- the mpg123 bug is breaking all sorts of other packages like SDL. It looks like the error is in the Windows build; unfortunately I'm not very familiar with Windows but I would be happy to help out if there is a way I can.

Sil3ntStorm:
Can someone tackle the CI servers into obedience? They appear to be stuck.

uilianries:
I just restarted the CI build now.

As the project is distributed by Source Forge, I made a local diff. It's looking good so far, I can see some changes related to the build, but only new checks, nothing affecting the result.

Here is my diff file from version 1.31.2 vs 1.33.0: [mpg123-1.31.2-1.33.0-diff.txt](https://github.com/user-attachments/files/20905961/mpg123-1.31.2-1.33.0-diff.txt)

uilianries:
The current error https://c3i.jfrog.io/artifactory/cci-build-logs/cci/prod/PR-27338/3/package_build_logs/build_log_mpg123_1_33_0_32cc5118049a944165a7aab1cd94e311_a52442f090b162815ed1720fe8c2c75b6030e7f1.txt is because configure.ac changed that specific line to `if $CCAS $CCASFLAGS -c -o conftest.o conftest.s 1>/dev/null 2>&1; then`. 

It will require an update directly in the recipe.

uilianries:
@Sil3ntStorm I can not push directly to your PR, so I would recommend adding a condition to not patch the CFLAGS for version 1.33.0:

```patch
diff --git a/recipes/mpg123/all/conanfile.py b/recipes/mpg123/all/conanfile.py
index cad85a437..2ba8bc2c8 100644
--- a/recipes/mpg123/all/conanfile.py
+++ b/recipes/mpg123/all/conanfile.py
@@ -8,6 +8,7 @@ from conan.tools.files import get, copy, export_conandata_patches, apply_conanda
 from conan.tools.env import VirtualBuildEnv, VirtualRunEnv
 from conan.tools.build import cross_building
 import os
 
@@ -163,7 +164,7 @@ class Mpg123Conan(ConanFile):
             cmake.configure(build_script_folder=os.path.join(self.source_folder, "ports", "cmake"))
             cmake.build()
         else:
-            if self.settings.compiler == "apple-clang" and cross_building(self):
+            if self.settings.compiler == "apple-clang" and cross_building(self) and Version(self.version) < "1.33.0":
                 # when testing if the assembler supports avx - propagate cflags (CFLAGS) to the assembler
                 # (which should contain "-arch x86_64" when crossbuilding with appleclang)
                 replace_in_file(self, os.path.join(self.source_folder, "configure"),
```

The version 1.33.0 has `CCASFLAGS` which incorporates the cflags injected by Conan. 

Regards!

Sil3ntStorm:
@uilianries I saw that as well, but I was unable to access the system for the past few days. It's pushed now.

uilianries (approved):
@Sil3ntStorm Thank you for your PR! It look good to me.

It follows my suggestion here: https://github.com/conan-io/conan-center-index/issues/27499#issuecomment-2910014029

And removing the version 1.26.4 should not affect ConanCenterIndex as well:

```
find recipes/ -name conanfile.py -exec grep mpg123/  {} +                                                                                      
recipes//libsndfile/all/conanfile.py:            self.requires("mpg123/1.31.2")
recipes//sdl_mixer/cmake/conanfile.py:            self.requires("mpg123/1.31.2")
recipes//sdl_mixer/all/conanfile.py:            self.requires("mpg123/1.31.2")
```

PR #27340, opened 2025-05-01
qt-msbuild: add new package 3.3.1

chausner:
### Summary
Changes to recipe:  **qt-msbuild/3.3.1**

#### Motivation
We already have the `qt` package which contains the main Qt libraries. If one does not need the moc, uic, rcc compilers, then the `qt` package is sufficient and can be used as-is. However, if the moc/uic/rcc compilers are needed, then one needs additional build system integration to invoke the compilers and pass the generated output to the compiler. For MSBuild, there is Qt/MSBuild which consists of a bunch of .props and .targets files which does exactly that.

Traditionally, Qt/MSBuild was only available through the [Qt Visual Studio Tools](https://github.com/qt-labs/vstools) extension but since a while, Qt/MSBuild is also made available standalone, e.g.: https://download.qt.io/official_releases/vsaddin/3.3.1/qt-vsaddin-msbuild-3.3.1.zip

This PR packages the standalone Qt/MSBuild as a Conan package. This allows users who already use the `qt` package and want to invoke the moc, uic, rcc compilers through MSBuild to use this package. The Qt VS Tools extension is then no longer needed.

#### Details
The proposed `qt-msbuild` package registers an environment variable `QtMsBuild` with the path to the Qt/MSBuild distribution. This is exactly what the Qt VS Tools extension does: Qt/MSBuild is shipped with the extension and the extension registers an environment variable that points to this directory.

Therefore, no modifications to .vcxproj files are needed to migrate from Qt VS Tools to this package. The only requirement is that users use this package with the `VirtualBuildEnv` generator and activate the build environment via `conanenv.bat` before running devenv/MSBuild.

---

### DIFF
--- a/recipes/qt-msbuild/all/conandata.yml
+++ b/recipes/qt-msbuild/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "3.3.1":
+    url: "https://download.qt.io/official_releases/vsaddin/3.3.1/qt-vsaddin-msbuild-3.3.1.zip"
+    sha256: "fb3c9396dadba1bfcf69a5c0001935049c9a62f6a468e75e87844497bb358338"
--- a/recipes/qt-msbuild/all/conanfile.py
+++ b/recipes/qt-msbuild/all/conanfile.py
@@ -1,0 +1,39 @@
+import os
+
+from conan import ConanFile
+from conan.tools.files import copy, get
+from conan.tools.layout import basic_layout
+
+required_conan_version = ">=1.53.0"
+
+class QtMSBuildConan(ConanFile):
+    name = "qt-msbuild"
+    description = "Qt/MSBuild MSBuild rules and targets from Qt VS Tools"
+    license = "LicenseRef-Qt-Commercial OR GPL-3.0-only WITH Qt-GPL-exception-1.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://www.qt.io"
+    topics = ("qt", "msbuild")
+
+    package_type = "build-scripts"
+    settings = "os", "arch", "compiler", "build_type"
+    no_copy_source = True
+
+    def layout(self):
+        basic_layout(self, src_folder="src")
+
+    def package_id(self):
+        self.info.clear()
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version])
+
+    def package(self):
+        copy(self, "*", src=self.source_folder, dst=os.path.join(self.package_folder, "res"))
+
+    def package_info(self):
+        self.cpp_info.bindirs = []
+        self.cpp_info.includedirs = []
+        self.cpp_info.libdirs = []
+        self.cpp_info.resdirs = ["res"]
+
+        self.buildenv_info.define_path("QtMsBuild", os.path.join(self.package_folder, "res"))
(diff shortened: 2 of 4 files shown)


recipes/qt-msbuild/all/conandata.yml
```diff
@@ -0,0 +1,4 @@
+sources:
+  "3.3.1":
+    url: "https://download.qt.io/official_releases/vsaddin/3.3.1/qt-vsaddin-msbuild-3.3.1.zip"
```
chausner, line 3: Alternatively, we could get Qt/MSBuild from: https://github.com/qt-labs/vstools/tree/v3.3.1/QtMSBuild/QtMsBuild. I think they are identical to the ZIP distribution, although I haven't checked. But I think it's easier to just use the standalone download links.

recipes/qt-msbuild/all/conanfile.py
```diff
@@ -0,0 +1,39 @@
...
+class QtMSBuildConan(ConanFile):
+    name = "qt-msbuild"
+    description = "Qt/MSBuild MSBuild rules and targets from Qt VS Tools"
+    license = "LicenseRef-Qt-Commercial OR GPL-3.0-only WITH Qt-GPL-exception-1.0"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://www.qt.io"
+    topics = ("qt", "msbuild")
+
+    package_type = "build-scripts"
+    settings = "os", "arch", "compiler", "build_type"
```
chausner, line 18: I've noticed that other similar recipes always specify settings "os", "arch", "compiler", "build_type" even though I don't understand why.

Since the packaged contents are build scripts that are independent of any os/arch/build type, I think those settings could be removed?

recipes/qt-msbuild/all/conanfile.py
```diff
@@ -0,0 +1,39 @@
...
+from conan import ConanFile
+from conan.tools.files import copy, get
+from conan.tools.layout import basic_layout
+
+required_conan_version = ">=1.53.0"
+
+class QtMSBuildConan(ConanFile):
+    name = "qt-msbuild"
+    description = "Qt/MSBuild MSBuild rules and targets from Qt VS Tools"
+    license = "LicenseRef-Qt-Commercial OR GPL-3.0-only WITH Qt-GPL-exception-1.0"
```
chausner, line 12: You can find this license string in any of the packaged files, e.g.: https://github.com/qt-labs/vstools/blob/v3.3.1/QtMSBuild/QtMsBuild/Qt.props#L5

recipes/qt-msbuild/all/conanfile.py
```diff
@@ -0,0 +1,39 @@
+import os
+
+from conan import ConanFile
+from conan.tools.files import copy, get
+from conan.tools.layout import basic_layout
+
+required_conan_version = ">=1.53.0"
```
chausner, line 7: I haven't tested if the recipe works with Conan 1.x, I only tested with the latest 2.x version.

I've seen that other similar recipes also do
```python
self.env_info.QtMsBuild = os.path.join(self.package_folder, "res")
```
which seems to be for backwards-compatibility with Conan 1.x. Should I add that or should I rather specify
```python
required_conan_version = ">=2.0"
```
?

PR #27346, opened 2025-05-02
easyloggingpp: various fixes

chausner:
### Summary
Changes to recipe:  **easyloggingpp/9.97.1**

#### Motivation
This fixes a few bugs in the current recipe.

#### Details
* Add check for C++11 or higher (the [README](https://github.com/abumq/easyloggingpp) mentions that the library requires at least C++11)
* Fix the broken option `disable_trace_logs` (this previously set the wrong define)
* Fix the broken option `enable_default_logfile` (this previously behaved in the opposite way as expected)
* Add pthread to system libs on Linux/FreeBSD if option `enable_thread_safe` is enabled (the README mentions that linking with pthread is required in this case)

This PR has been split up from https://github.com/conan-io/conan-center-index/pull/27261.

---

### DIFF
--- a/recipes/easyloggingpp/all/conanfile.py
+++ b/recipes/easyloggingpp/all/conanfile.py
@@ -91,7 +91,7 @@ class EasyloggingppConan(ConanFile):
         if self.options.disable_verbose_logs:
             defines.append("ELPP_DISABLE_VERBOSE_LOGS")
         if self.options.disable_trace_logs:
-            defines.append("lib_utc_datetime")
+            defines.append("ELPP_DISABLE_TRACE_LOGS")
         if self.options.lib_utc_datetime:
             defines.append("ELPP_UTC_DATETIME")
         return defines
--- a/recipes/easyloggingpp/all/conanfile.py
+++ b/recipes/easyloggingpp/all/conanfile.py
@@ -122,3 +122,6 @@ class EasyloggingppConan(ConanFile):
     def package_info(self):
         self.cpp_info.libs = ["easyloggingpp"]
         self.cpp_info.defines = self._public_defines
+
+        if self.options.enable_thread_safe and self.settings.os in ["Linux", "FreeBSD"]:
+            self.cpp_info.system_libs.append("pthread")
--- a/recipes/easyloggingpp/all/conanfile.py
+++ b/recipes/easyloggingpp/all/conanfile.py
@@ -1,6 +1,7 @@
 import os
 
 from conan import ConanFile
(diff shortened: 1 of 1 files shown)


recipes/easyloggingpp/all/conanfile.py
```diff
@@ -54,6 +55,10 @@ def config_options(self):
         if self.settings.os == "Windows":
             del self.options.fPIC
 
+    def validate(self):
+        if self.settings.compiler.cppstd:
```
AbrilRBS, line 59: Conan 2 always defines a cppstd, so this check can be removed an always call `check_min_cppstd`

AbrilRBS (commented):
Thanks a lot! Only a minor change, otherwise looks good :)

recipes/easyloggingpp/all/conanfile.py
chausner: I removed the check and bumped `required_conan_version` to 2.0: 9ace18fbe58144bea5a94f0a826ca1e4e77a9063

uilianries (approved):
Hello! Thank you for your PR! I did a review and it's looking good. Here are my comments: 

- The upstream indicates C++11 for >=8.91 https://github.com/abumq/easyloggingpp#overview. Currently, we only support 9.97.z in CCI
- The define ELPP_NO_DEFAULT_LOG_FILE looks correct now: https://github.com/abumq/easyloggingpp/blob/v9.97.1/src/easylogging%2B%2B.cc
- Same for ELPP_DISABLE_TRACE_LOGS: https://github.com/abumq/easyloggingpp/blob/v9.97.1/src/easylogging%2B%2B.h#L333
- For pthread, it's documented in the readme: https://github.com/abumq/easyloggingpp#configuration-macros

LGTM.

-----

IMO, this recipe could be released without all those options, and use the Conan configuration `tools.build:defines` instead, when needed. The project itself has more possibilities of options, based on extra defines, so I would be inclined not to add extra options, but to simplify instead. Nothing related to this PR, just a comment, as we needed a fix for two options.

chausner:
> IMO, this recipe could be released without all those options, and use the Conan configuration tools.build:defines instead, when needed. The project itself has more possibilities of options, based on extra defines, so I would be inclined not to add extra options, but to simplify instead. Nothing related to this PR, just a comment, as we needed a fix for two options.

Thanks for the hint about `tools.build:defines`. There is actually a second MR (https://github.com/conan-io/conan-center-index/pull/27349) in which I propose to add a number of additional options. Let's continue the discussion there.

PR #27350, opened 2025-05-02
tree-sitter-glsl: Add new package

jan-kelemen:
### Summary
Changes to recipe:  **tree-sitter-glsl/0.2.0** https://github.com/tree-sitter-grammars/tree-sitter-glsl

#### Motivation
I wanted to use it but it wasn't available on Conan.

#### Details
Mostly based on already existing Conan package for tree-sitter-c https://github.com/conan-io/conan-center-index/tree/master/recipes/tree-sitter-c

---

### DIFF
--- a/recipes/tree-sitter-glsl/all/conanfile.py
+++ b/recipes/tree-sitter-glsl/all/conanfile.py
@@ -1,6 +1,6 @@
 from conan import ConanFile
-from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
-from conan.tools.files import get, copy 
+from conan.tools.cmake import CMake, cmake_layout
+from conan.tools.files import get, copy
 
 import os
 
@@ -25,28 +25,15 @@ class TreeSitterCConan(ConanFile):
         "shared": False,
         "fPIC": True,
     }
+    implements = ["auto_shared_fpic"]
+    languages = ["C"]
+    generators = "CMakeDeps", "CMakeToolchain"
 
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def export_sources(self):
         copy(self, "CMakeLists.txt", src=self.recipe_folder, dst=os.path.join(self.export_sources_folder, "src"))
-
-    def generate(self):
-        tc = CMakeToolchain(self)
-        tc.generate()
-        deps = CMakeDeps(self)
-        deps.generate()
-
-    def config_options(self):
-        if self.settings.os == "Windows":
-            del self.options.fPIC
-
-    def configure(self):
-        if self.options.shared:
-            self.options.rm_safe("fPIC")
-        self.settings.rm_safe("compiler.cppstd")
-        self.settings.rm_safe("compiler.libcxx")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
--- a/recipes/tree-sitter-glsl/all/conanfile.py
+++ b/recipes/tree-sitter-glsl/all/conanfile.py
@@ -1,0 +1,73 @@
+from conan import ConanFile
+from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
+from conan.tools.files import get, copy 
+
+import os
+
+required_conan_version = ">2.0"
+
+
+class TreeSitterCConan(ConanFile):
+    name = "tree-sitter-glsl"
+    description = "GLSL grammar for tree-sitter."
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/tree-sitter-grammars/tree-sitter-glsl"
+    topics = ("parser", "grammar", "tree", "glsl", "ide")
+    settings = "os", "arch", "compiler", "build_type"
+    package_type = "library"
+
+    options = {
+        "shared": [True, False],
+        "fPIC": [True, False],
+    }
+    default_options = {
+        "shared": False,
+        "fPIC": True,
+    }
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def export_sources(self):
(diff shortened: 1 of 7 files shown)


perseoGI (dismissed):
LGTM thank you!

perseoGI (dismissed):

recipes/tree-sitter-glsl/all/conanfile.py
```diff
@@ -0,0 +1,58 @@
...
+    generators = "CMakeDeps", "CMakeToolchain"
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def requirements(self):
+        self.requires("tree-sitter/0.25.1", transitive_headers=True, transitive_libs=True)
```
perseoGI, line 40: Downgrade `tree-sitter` dependency to be `0.24.3`, the one used along all other grammars in Conan Center Index. This will avoid possible version conflicts in users depending on multiple grammars. 

```suggestion
        self.requires("tree-sitter/0.24.3", transitive_headers=True, transitive_libs=True)
```

perseoGI (dismissed):

AbrilRBS (approved):
Thanks!

perseoGI (approved):

jcar87 (changes requested):
Thanks for your contribution!

Always a bit wary when we have to add a CMakeLists ourselves, though I can appreciate the need given that they only seem to have a makefile.

However, looking at the makefile upstream
* Windows is not supported: https://github.com/tree-sitter-grammars/tree-sitter-glsl/blob/master/Makefile#L37-L39 - out of caution the recipe shouldn't either, unless we clarify with the maintainers whether that comment is only relevant to the Makefile, or the entire library.
* The upstream makefile is setting soname (both when creating the file, and when giving it the filename). This recipe is not
  - https://github.com/tree-sitter-grammars/tree-sitter-glsl/blob/master/Makefile#L43
  - https://github.com/tree-sitter-grammars/tree-sitter-glsl/blob/24a6c8ef698e4480fecf8340d771fbcb5de8fbb4/Makefile#L57

This divergence is relevant - as the soname is embedded in downstream shared library consumers.

I would suggest that we either make sure we generate _exactly_ the same files as the upstream makefile, or just use the upstream makefile (and disable windows anyway).

jan-kelemen:
I would say the Windows not being supported in the Makefile isn't concerning, these thing's are generated anyway by the tree-sitter CLI.

Examples that are already in the CCI:
- https://github.com/tree-sitter/tree-sitter-c/blob/master/Makefile#L1
- https://github.com/shotover/tree-sitter-cql/blob/main/Makefile#L40

Depends when they were generated, but it's present in other grammars too:
- https://github.com/search?q=org%3Atree-sitter+"Windows+is+not+supported"&type=code

perseoGI:
Greetings @jan-kelemen,
We are waiting for the merging of [this upstream PR](https://github.com/tree-sitter-grammars/tree-sitter-glsl/pull/79#discussion_r2104103025).
It is crucial to fix the SONAME version, as tree-sitter incompatibilities may arise later.

The Windows target is supported in other grammars and recipes in the Conan Center Index. However, until the above PR is merged (which includes the new CMakeLists.txt that does not accurately reflect Windows systems), we may need to wait.

Additionally, the only distribution packaging this grammar is `openSUSE`: https://repology.org/project/tree-sitter-glsl/versions 
which appears not to address the SO_VERSION issue at all.

It would be very helpful if you could also express some interest and advocate for the upstream PR to see if they can move it forward.

Thank you in advance!

jan-kelemen:
Yup, it's my opinion too that it makes sense to wait for that PR and use the same CMakeLists.txt as generated by the current tree-sitter CLI bindings.

This isn't an issue for me as the SONAME difference isn't relevant for my current needs, but for CCI I understand it might be.

perseoGI:
Thank you for your understanding and your patience!

danimtb:
Hello @jan-kelemen!

Thank you for your time and contribution. Sorry it took us so long to get back to this PR.

I regret to say that I’m conservative about packaging it in Conan Center Index. Here are some of the reasons:

* The latest release, **v0.2.0**, is from March 2025, with no new tagged release since then:  
  https://github.com/tree-sitter-grammars/tree-sitter-glsl/releases

* Review already flagged that this recipe ships its own CMakeLists and does not match upstream SONAME / shared-library naming. We agreed to wait for a proper upstream CMake/SONAME story instead of carrying CCI-only build logic.

* That work is still open after more than a year, so there is no released install path we can follow without diverging from upstream:  
  https://github.com/tree-sitter-grammars/tree-sitter-glsl/pull/79

Accepting this now would risk this recipe becoming a refuge for community patches to “fix” problems not solved upstream first. That is not what Conan Center is intended for: we want to package artifacts as close as possible to the upstream implementation.

Any new bug found while packaging or consuming this recipe would also be hard to address: we can open an upstream issue, but there is no clear path to a release that includes the CMake/SONAME support this recipe needs.

This recipe would end up on the list of [Abandoned projects in ConanCenterIndex](https://github.com/conan-io/conan-center-index/issues/29097) that may be marked as deprecated later.

Unfortunately, I’m not inclined to accept this PR at this time, unless you can show that upstream has landed (and released) a maintained CMake/SONAME-compatible build we can follow without a CCI-owned CMakeLists.

That does not stop you from distributing the recipe yourself:

* Consumers can use a [local recipes index](https://docs.conan.io/2/tutorial/conan_repositories/setup_local_recipes_index.html) and point your GitHub fork as a remote.
* Use [Artifactory](https://docs.conan.io/2/tutorial/conan_repositories/setting_up_conan_remotes/artifactory/artifactory_ce_cpp.html) to publish pre-built packages for a team.

This is not a blocker for future contributions. Feel free to open new issues or a new PR once upstream ships a consumable C/CMake install.

We welcome future PRs and appreciate your involvement.
Hope that it makes sense.

Thank you

jan-kelemen:
Completely understandable, I've got no issues with closing the PR.

PR #27351, opened 2025-05-02
chore: Add CV_TRACE option to opencv recipe

clarkholdham:
### Summary
Changes to recipe:  **opencv/4.11.0**

#### Motivation
The opencv recipe did not expose an option to toggle the CV_TRACE compile flag. This is needed to work around a known memory leak in opencv: https://github.com/opencv/opencv/issues/19727

#### Details
Option: `cv_trace` defaulted to `True` (as this seems to be the default state from opencv CMakeLists.txt https://github.com/opencv/opencv/blob/4ad4bd5dc09b343b2be6c8137b563f143d68584d/CMakeLists.txt#L541). Will toggle `CV_TRACE` flag to `False` if `cv_trace` option is set to `False`.

---


uilianries (approved):
Thank you for your suggestion! The memory leak issue in OpenCV is still open today, so managing CV Trace looks fine. 

However, instead of adding a new option, let's just disable it by default. It's much simpler and can be overridden by the Conan configuration `tools.cmake.cmaketoolchain:extra_variables`, like:

    conan install ... -c "tools.cmake.cmaketoolchain:extra_variables={'CV_TRACE': 'ON'}"

When adding a new option, the package ID is changed and may result in a missing package for some consumers. It does not mean new options are not added, but we prefer adding those that are marked as valuable or important by several consumers. 

Doing a quick search, both Gentoo and Fedora are managing CV_TRACE as well:
* https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/opencv/opencv-4.12.0-r2.ebuild#n863
* https://src.fedoraproject.org/rpms/opencv/blob/rawhide/f/opencv.spec#_439

Regards!

PR #27355, opened 2025-05-03
bitserializer: add version 0.80

PavelKisliak:
### Summary
Changes to recipe:  **bitserializer/0.80**

#### Motivation
New version of BitSerializer 0.80 has been released ([Release notes](https://github.com/PavelKisliak/BitSerializer/releases)).

#### Details
- Add new version of library to recipe
- Remove too old versions (allows to simplify the recipe a bit)
- Remove support of "cmake_find_package*" generators

---

### DIFF
--- a/recipes/bitserializer/all/conanfile.py
+++ b/recipes/bitserializer/all/conanfile.py
@@ -7,15 +7,15 @@ from conan.tools.layout import basic_layout
 from conan.tools.scm import Version
 import os
 
-required_conan_version = ">=1.51.1"
+required_conan_version = ">=2.4"
 
 
 class BitserializerConan(ConanFile):
     name = "bitserializer"
-    description = "C++ 17 library for serialization to multiple output formats (JSON, XML, YAML, CSV, MsgPack)"
+    description = "Multi-format serialization library (JSON, XML, YAML, CSV, MsgPack)"
     topics = ("serialization", "json", "xml", "yaml", "csv", "msgpack")
     url = "https://github.com/conan-io/conan-center-index"
-    homepage = "https://bitbucket.org/Pavel_Kisliak/bitserializer"
+    homepage = "https://github.com/PavelKisliak/BitSerializer"
     license = "MIT"
     package_type = "header-library"
     settings = "os", "arch", "compiler", "build_type"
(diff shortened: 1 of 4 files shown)


recipes/bitserializer/all/conanfile.py
```diff
@@ -94,14 +80,7 @@ def package_id(self):
...
     def validate(self):
-        if self.settings.compiler.get_safe("cppstd"):
-            check_min_cppstd(self, self._min_cppstd)
-
-        minimum_version = self._compilers_minimum_version.get(str(self.settings.compiler), False)
-        if minimum_version and Version(self.settings.compiler.version) < minimum_version:
-            raise ConanInvalidConfiguration(
-                f"{self.ref} requires C++{self._min_cppstd}, which your compiler does not support.",
-            )
+        check_min_cppstd(self, 17)
```
PavelKisliak, line 83: This is definitely simple, but compilers which claim to support C++ 17 may not have all features which requires the library. I would prefer to have more granular check.

AbrilRBS: Thanks for the feedback - while it's true that some compilers might not actually have the necessary C++17 support, we're trying to move away from blanket checks for all compiler versions like this.

Of course, we will accept specific compiler version checks for those that -std=c++17 does not mean it would compile on those, but in most cases we've found that unless a proper message is given, the compiler version checks end up being a bit too confusing

So the question here would be if the old checks for those compilers accepted C++ 17 but failed otherwise, or it any of them were a "C++17" was introduced here?

Thanks!

PavelKisliak: The old checks were required for check versions of compilers which supports the library (requirements from the official [README.md](https://github.com/PavelKisliak/BitSerializer/blob/master/README.md#requirements)). The TestPackage from recipe may compile with older versions, but the library will not work on the end user side.

The library's CMake script has all necessary checks, but it's not used when enabled only header-only components.
So, I think that these checks of compiler version are still necessary for early inform user that he has wrong setup.

PavelKisliak: Hello! Can we move forward with this PR?
Please let me know if there is anything else I can do.
Are there any objections if I bring back compilers checking?

AbrilRBS (dismissed):
Sorry for the delay @PavelKisliak - I've brought back the compiler checks but updating the message, and cleaned up the last remaining issues, thanks for your patience and feedback! :)

AbrilRBS (approved):

PR #27361, opened 2025-05-04
25765: Have assimp use minizip-ng to avoid conflict with openimageio

EstebanDugueperoux2:
### Summary

#### Details

---

### DIFF
--- a/recipes/assimp/5.x/conanfile.py
+++ b/recipes/assimp/5.x/conanfile.py
@@ -32,7 +32,6 @@ class AssimpConan(ConanFile):
         "shared": False,
         "fPIC": True,
         "double_precision": False,
-        "minizip-ng/*:mz_compatibility": True,
     }
 
     _format_option_map = {
@@ -177,7 +176,7 @@ class AssimpConan(ConanFile):
     def requirements(self):
         # TODO: unvendor others libs:
         # - Open3DGC
-        self.requires("minizip-ng/4.0.7")
+        self.requires("minizip/1.2.13")
         self.requires("pugixml/1.14")
         self.requires("utfcpp/4.0.1")
         self.requires("zlib/[>=1.2.11 <2]")
--- a/recipes/assimp/5.x/conanfile.py
+++ b/recipes/assimp/5.x/conanfile.py
@@ -32,6 +32,7 @@ class AssimpConan(ConanFile):
         "shared": False,
         "fPIC": True,
         "double_precision": False,
+        "minizip-ng/*:mz_compatibility": True,
     }
 
     _format_option_map = {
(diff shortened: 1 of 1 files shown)


uilianries:
@EstebanDugueperoux2 Hello :wave: Thank you for taking the time to address this issue.

Unfortunately, we don't have `minizip-ng:mz_compatibility=True` available in ConanCenter, it's **False** by default: https://github.com/conan-io/conan-center-index/blob/master/recipes/minizip-ng/all/conanfile.py#L39. The result will be a missing package error when building it in the CI, and users should expect the same when installing Assimp. 

As a workaround, you can use [replace-requires](https://docs.conan.io/2/reference/config_files/profiles.html#replace-requires) in your Conan profile:

```ini
[replace_requires]
minizip/*: minizip-ng/4.0.7

[options]
minizip-ng/*:mz_compatibility=True
```

The CMake target name is a bit more complicated, you will need to keep that change in the Assimp recipe, as it affects what is expected by that project. I just checked locally if you could use a conanfile.py with CMakeDeps.set_property to override the cmake_target_name without changing the Assimpt recipe, but it did not work.

EstebanDugueperoux2:
Hi @uilianries,

Thanks for the tip but without "cd.set_property("minizip-ng", "cmake_target_name", "minizip::minizip")" I have following error:

`
CMake Error at code/CMakeLists.txt:1330 (ADD_LIBRARY):
  Target "assimp" links to:

    minizip::minizip

    * An ALIAS target is missing.

`

when building assimp.

It is for that I want to fix recipe.

Regards.

uilianries:
@EstebanDugueperoux2 Correct, as I commented before, you will need to keep the CMake target name change in the recipe:

```patch
diff --git a/recipes/assimp/5.x/conanfile.py b/recipes/assimp/5.x/conanfile.py
index f06d14969..b39c6988a 100644
--- a/recipes/assimp/5.x/conanfile.py
+++ b/recipes/assimp/5.x/conanfile.py
@@ -264,6 +264,7 @@ class AssimpConan(ConanFile):
         cd = CMakeDeps(self)
         cd.set_property("rapidjson", "cmake_target_name", "rapidjson::rapidjson")
         cd.set_property("utfcpp", "cmake_target_name", "utf8cpp::utf8cpp")
+        cd.set_property("minizip-ng", "cmake_target_name", "minizip::minizip")
         cd.generate()
 
         venv = VirtualBuildEnv(self)
```

Another alternative would be injecting an extra cmake file to override the target, but it's too much IMO.

EstebanDugueperoux2:
Hi @uilianries,

ok sorry I have not read correctly.
I have updated the PR.

Regards.

uilianries:
@EstebanDugueperoux2 Hello, as Assimp does not require minizip directly and it's a particular case, because you need both OpenImageIO and Assimp, I'm not inclined to merge your PR. Plus, this scenario is not covered by the CI in Conan Center Index, which will be hard to track and validate for every change.

As in this particular case, it could be used for other PR's to add support for different requirements not tested in Conan Center Index.

Still, you can keep your fork as repository, by using [Local Recipes Index](https://docs.conan.io/2/tutorial/conan_repositories/setup_local_recipes_index.html) for example.

PR #27370, opened 2025-05-06
libxslt: Add version 1.1.43 (CVE-2025-24855, CVE-2024-55549)

gsantner:
### Summary
* Changes to recipe:  **libxslt: Add version 1.1.43**

#### Motivation
New upstream releases, known security vulnerabilities

* libxslt
  * https://gitlab.gnome.org/GNOME/libxslt/-/tags
  * https://gitlab.gnome.org/GNOME/libxslt/-/commits/master
  * https://download.gnome.org/sources/libxslt/1.1/

#### Details
Version bumps

---

### DIFF
--- a/recipes/libxslt/all/conandata.yml
+++ b/recipes/libxslt/all/conandata.yml
@@ -1,4 +1,7 @@
 sources:
+  "1.1.43":
+    sha256: "5a3d6b383ca5afc235b171118e90f5ff6aa27e9fea3303065231a6d403f0183a"
+    url: "https://download.gnome.org/sources/libxslt/1.1/libxslt-1.1.43.tar.xz"
   "1.1.42":
     sha256: "85ca62cac0d41fc77d3f6033da9df6fd73d20ea2fc18b0a3609ffb4110e1baeb"
     url: "https://download.gnome.org/sources/libxslt/1.1/libxslt-1.1.42.tar.xz"
--- a/recipes/libxslt/all/conanfile.py
+++ b/recipes/libxslt/all/conanfile.py
@@ -179,7 +179,7 @@ class LibxsltConan(ConanFile):
             autotools.make()
 
     def package(self):
-        copy(self, "COPYING", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+        copy(self, "Copyright", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
         if is_msvc(self):
             copy(self, "*.h", src=os.path.join(self.source_folder, "libxslt"),
                               dst=os.path.join(self.package_folder, "include", "libxslt"))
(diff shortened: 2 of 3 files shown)


jcar87:
Thanks @gsantner  - please split them into separate PRs, thanks!

gsantner:
Is there no bot that could automate version bump commits and open such PRs automated?

A core library like libxml and openssl shouldn't went unnoticed/outdated for months after release and require manual intervention

gsantner:
Is there no bot that could automate (trying) version bump commits and open such PRs automated?

A core library like libxml and openssl shouldn't went unnoticed/outdated for months after release and require manual comparison between conan and upstream version control.

gsantner:
@jcar87 
I separated them.

Also it seems there are indeed bots scanning for versions, would be good if the information is being used to create PRs automatically.
Otherwise each and every open source project has to implement their own update scanning and library recipes, as the package manage provides by default outdated and vulnerable versions. 
Of course requires manual review and PR builds may fail some time, but for many cases it's a matter of bumping the version to newer patch version.

gsantner:
@jcar87 
Thanks, I separated them.

Also it seems there are indeed bots scanning for versions, would be good if the information is being used to create PRs automatically.
Otherwise each and every open source project has to implement their own update scanning (http request to each library project) and library recipes, as the package manage provides often outdated and vulnerable versions at latest option. 
Of course requires manual review and PR builds may fail some time, but for many cases it's a matter of bumping the version to newer patch version.

jcar87:
> @jcar87 Thanks, I separated them.
> 
> Also it seems there are indeed bots scanning for versions, would be good if the information is being used to create PRs automatically. Otherwise each and every open source project has to implement their own update scanning (http request to each library project) and library recipes, as the package manage provides often outdated and vulnerable versions at latest option. Of course requires manual review and possibly manual changes, PR builds may fail, but for many cases it can be a matter of bumping the version to newer patch version.

Conan Center provides libraries "as-is" - whether the consumer applications are affected by a CVE is a determination consumers have to make for themselves - this is not only scanning for whether a specific version has a CVE report, but also whether it's actually applicable in the consumer code. For example, one of the CVEs you listed for libxml2 was for the python bindings that the recipe doesn't build by default, and I would say its not used by any other recipe in Conan Center. There are a lot of cases of "optional" dependencies where the consumer applications actually don't use the library in their code paths, and the best course of action is to tweak options to disable the dependency altogether - again, these are determinations that only a security-conscious consumer can make, and not us. 

However, we should offer newer versions unaffected by CVEs once they are released, and we will prioritise PRs in this direction. We also have recently released the conan audit functionality (more about it [here](https://blog.conan.io/introducing-conan-audit-command/)) - to help users make that determination.

Having an automated try-bot is in our longer term plans - provided it does not result in wasteful use of resources (which tends to inevitably happen with such a blanket approach)

jcar87 (approved):
Thanks!

jcar87 (approved):

Issue #7249, opened 2021-09-11
[package] abseil/*: consider pinning the ABI at install time


coryan:
### Package and Environment Details (include every applicable attribute)

  * Package Name/Version: **abseil/all versions**
  * Operating System+version: any
  * Compiler+version: any

### Steps to reproduce (Include if Applicable)

Try to use `Abseil/*` with C++17 enabled. A good example can be found in #5766 .

By default, Abseil assumes that the whole application (including Abseil) is compiled with the same compiler and flags.  A good discussion can be found in abseil/abseil-cpp#696.  If Abseil is compiled with C++11 and used from an application compiled in C++17 this does not work, as the ABI (and arguably the API) changes depending on the C++ flags in effect when the Abseil headers are *used*.

However, it is possible to "pin" Abseil's API to whatever was used when it was compiled and installed.

Specifically one would need to set the pre-processor macros in this file to `0` if compiled with C++ < 17, and to `2` if compiled with C++ >= 17:

https://github.com/abseil/abseil-cpp/blob/master/absl/base/options.h

We use this to compile Abseil with C++11 and then use it from both C++11 and C++17:

https://github.com/googleapis/google-cloud-cpp/blob/7ee1d91d4fb6915dc50617c01ebc04777e183ebd/ci/cloudbuild/dockerfiles/fedora-34.Dockerfile#L60

coryan:
I will be sending a PR for this.

coryan:
@uilianries I can try to fix this bug, but first I would like to check if this is a bad idea.

My proposal would be to pin the ABI of the Abseil package to the ABI used when it was compiled.  One does this by setting these defines:

https://github.com/abseil/abseil-cpp/blob/1ce4ceca2b2931bc4d7e470228c2dbb2f3dfea0f/absl/base/options.h#L130
https://github.com/abseil/abseil-cpp/blob/1ce4ceca2b2931bc4d7e470228c2dbb2f3dfea0f/absl/base/options.h#L157
https://github.com/abseil/abseil-cpp/blob/1ce4ceca2b2931bc4d7e470228c2dbb2f3dfea0f/absl/base/options.h#L183

The values must be `0` if Abseil was compiled with C++ < 17.  The values must be `1` if Abseil was compiled with C++ >= 17, and then one can only use the installed library with C++ >= 17.

It seems to me one could change the `options.h` header in the `package()` step.  Does that sound feasible?

uilianries:
[...]

Issue #26920, opened 2025-03-24
Potentially unsupported open62541 combination


AiyionPrime:
First of all: Thanks for the effort of packaging opne62541pp, I highly appreciate having it in conan!

One thing I stumble across was the chosen open62541 version, though.
There exists an open62541pp Release [0.17.0](https://github.com/open62541pp/open62541pp/releases/tag/v0.17.0), which has a dependency on `open62541` (mind the mssing pp) in version 1.3.15:
https://github.com/open62541pp/open62541pp/tree/master/3rdparty

The conan package `open62541pp` https://conan.io/center/recipes/open62541pp on the other hand has a dependency on open62541 in version 1.4.6 (as conan package)
https://conan.io/center/recipes/open62541?version=1.4.6

https://github.com/conan-io/conan-center-index/blob/1f8a9d826fdf0c0e3cedf798b44a10dadb3223fe/recipes/open62541pp/all/conanfile.py#L38C24-L38C28

This makes migrating from e.g. submodules a little troublesome, as both a change in the build system as well as a change in versions is bound to happen at the same time.

Would it be an option to stick with the supported versions the open62541pp maintainers chose instead?

Anyway, nice to have this at all, thanks again!

RudolfRendier:
I believe you can work around this with 'replace_requires'
https://blog.conan.io/2024/02/20/Conan-2-graph-features.html

Unfortunately not every version is available in conan-center.
It would be better to have versions to choose from, both in 1.3.x and 1.4.x

AiyionPrime:
My bad, 1.3.15 does indeed appear to be missing.
https://conan.io/center/recipes/open62541 (versions tab)

@mynameislame In order to follow up eventually.

PR #27372, opened 2025-05-07
[WIP] open62541: Add version 1.4.11.1 and option for NodeSetLoader

fischjo3SICKAG:
# Summary
This pull request adds support for version 1.4.11.1 of open62541, introducing a new Conan option for enabling the NodesetLoader feature. This feature allows users to optionally include the NodesetLoader functionality.

# Motivation
The addition of the NodesetLoader option allows Conan users to utilize the NodesetLoader feature.

# Details
- **conanfile.py**:
    - Add a boolean option `nodeset_loader` to enable or disable the NodesetLoader feature.
    - Conditionally add libxml2 as a dependency when `nodeset_loader` is enabled.
- **CMakeLists.txt**:
    - Update to use the libxml2 dependency provided by Conan when the NodesetLoader option is enabled.
- **submoduledata.yml**:
    - Add NodesetLoader sources to support the new feature.

-------------------------------------------------

 - [x] Read the [CONTRIBUTING.md](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
       ⚠️ Found [PR #26457](https://github.com/conan-io/conan-center-index/pull/26457): This PR also upgrades open62541 to version 1.4.11.1 but introduces a different Conan option (QNX support).

# Open Questions
- **Which version of NodesetLoader should be used?**: I raised this question in the upstream open62541 repository ([open62541 issue #7249](https://github.com/open62541/open62541/issues/7249)).
- **How to proceed with the partially duplicate PR?**: I suggest waiting for PR #26457 to be merged before proceeding with this pull request.

### DIFF
--- a/recipes/open62541/all/conanfile.py
+++ b/recipes/open62541/all/conanfile.py
@@ -109,6 +109,8 @@ class Open62541Conan(ConanFile):
         "cpp_compatible": [True, False],
         # UA_ENABLE_STATUSCODE_DESCRIPTIONS=readable_statuscodes
         "readable_statuscodes": [True, False],
+        # UA_ENABLE_NODESETLOADER=nodeset_loader
+        "nodeset_loader": [True, False],
     }
     default_options = {
         "fPIC": True,
@@ -137,6 +139,7 @@ class Open62541Conan(ConanFile):
         "hardening": True,
         "cpp_compatible": False,
         "readable_statuscodes": True,
+        "nodeset_loader": False,
     }
 
     exports = "submoduledata.yml"
@@ -152,6 +155,10 @@ class Open62541Conan(ConanFile):
 
         if Version(self.version) >= "1.3.1":
             del self.options.embedded_profile
+
+        # NodesetLoader has only rudimentary Windows support --> disabling for now. This might change in the future.
+        if Version(self.version) < "1.4.8" or self.settings.os != "Linux": 
+            del self.options.nodeset_loader
 
     def configure(self):
         if self.options.shared:
@@ -170,6 +177,11 @@ class Open62541Conan(ConanFile):
         if self.options.web_socket:
             self.options["libwebsockets"].with_ssl = self.options.encryption
 
+        if self.options.nodeset_loader:
+            self.options["libxml2"].iconv = False
+            self.options["libxml2"].shared = False
+            self.options["libxml2"].with_subunit = False
+
     def layout(self):
         cmake_layout(self, src_folder="src")
 
@@ -182,6 +194,8 @@ class Open62541Conan(ConanFile):
             self.requires("libwebsockets/4.3.2")
         if self.options.discovery == "With Multicast" or "multicast" in str(self.options.discovery):
             self.requires("pro-mdnsd/0.8.4")
+        if self.options.nodeset_loader:
+            self.requires("libxml2/2.13.4")
 
     def validate(self):
         if not self.options.subscription:
@@ -248,7 +262,8 @@ class Open62541Conan(ConanFile):
                     destination=path,
                     filename=archive_name,
                     strip_root=True)
-
+        self._patch_sources()
+        
     def _get_log_level(self):
         return {
             "Fatal": "600",
@@ -356,6 +371,8 @@ class Open62541Conan(ConanFile):
         # Honor BUILD_SHARED_LIBS from conan_toolchain (see https://github.com/conan-io/conan/issues/11840)
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
 
+        if Version(self.version) >= "1.4.8":
+            tc.variables["UA_ENABLE_NODESETLOADER"]=self.options.nodeset_loader
         tc.generate()
         tc = CMakeDeps(self)
         tc.set_property("mbedtls", "cmake_additional_variables_prefixes", ["MBEDTLS"])
@@ -363,11 +380,10 @@ class Open62541Conan(ConanFile):
 
     def _patch_sources(self):
         apply_conandata_patches(self)
-        if Version(self.version) >= "1.3.1":
+        if Version(self.version) >= "1.3.1" and Version(self.version) <="1.4.7":
             os.unlink(os.path.join(self.source_folder, "tools", "cmake", "FindPython3.cmake"))
 
     def build(self):
-        self._patch_sources()
         cmake = CMake(self)
         cmake.configure()
         cmake.build()
--- a/recipes/open62541/all/patches/0001-disable-sanitizers-1_4_11_1.patch
+++ b/recipes/open62541/all/patches/0001-disable-sanitizers-1_4_11_1.patch
@@ -1,0 +1,35 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index bc3f94b54..4253efd3b 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -663,18 +663,18 @@ if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_C_COMPILER_ID STREQUAL "Clang")
+     set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS "") # cmake sets -rdynamic by default
+ 
+     # Debug
+-    if(UA_ENABLE_DEBUG_SANITIZER AND BUILD_TYPE_LOWER_CASE STREQUAL "debug" AND UNIX AND NOT UA_BUILD_OSS_FUZZ AND
+-       CMAKE_C_COMPILER_ID STREQUAL "Clang" AND NOT UA_ENABLE_UNIT_TESTS_MEMCHECK)
+-        # Add default sanitizer settings when using clang and Debug build.
+-        # This allows e.g. CLion to find memory locations for SegFaults
+-        message(STATUS "Sanitizer enabled")
+-        set(SANITIZER_FLAGS "-g -fno-omit-frame-pointer -gline-tables-only -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=leak -fsanitize=undefined")
+-        if(CMAKE_CXX_COMPILER_VERSION AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.0)
+-            set(SANITIZER_FLAGS "${SANITIZER_FLAGS} -fsanitize-coverage=trace-pc-guard")
+-        endif()
+-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${SANITIZER_FLAGS}")
+-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SANITIZER_FLAGS}")
+-    endif()
++    # if(UA_ENABLE_DEBUG_SANITIZER AND BUILD_TYPE_LOWER_CASE STREQUAL "debug" AND UNIX AND NOT UA_BUILD_OSS_FUZZ AND
++    #    CMAKE_C_COMPILER_ID STREQUAL "Clang" AND NOT UA_ENABLE_UNIT_TESTS_MEMCHECK)
++    #     # Add default sanitizer settings when using clang and Debug build.
++    #     # This allows e.g. CLion to find memory locations for SegFaults
++    #     message(STATUS "Sanitizer enabled")
++    #     set(SANITIZER_FLAGS "-g -fno-omit-frame-pointer -gline-tables-only -fsanitize=address -fsanitize-address-use-after-scope -fsanitize=leak -fsanitize=undefined")
++    #     if(CMAKE_CXX_COMPILER_VERSION AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.0)
++    #         set(SANITIZER_FLAGS "${SANITIZER_FLAGS} -fsanitize-coverage=trace-pc-guard")
++    #     endif()
++    #     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${SANITIZER_FLAGS}")
++    #     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${SANITIZER_FLAGS}")
++    # endif()
+ 
+     if(NOT MINGW AND UA_ENABLE_HARDENING AND ((CMAKE_BUILD_TYPE STREQUAL "Release") OR (CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")))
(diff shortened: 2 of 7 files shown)


ErikTempelaarVO:
Following this because I have issues with v1.4.6 and this recipe on MinGW, see [here](https://github.com/conan-io/conan-center-index/pull/25598#issuecomment-2854526856)

Also was working on a recipe that includes 1.3.15, the version that's currently included in open62541pp.
(as discussed [here](https://github.com/conan-io/conan-center-index/issues/26920))

```
  "1.3.15":
    url: "https://github.com/open62541/open62541/archive/v1.3.15.tar.gz"
    sha256: "1f71c1a57eb98b2cf3c8f92ee9efcc7e3139932a97546a2bc8dffcfb7f1b0fd9"
```

```
  "1.3.15":
    - patch_file: "patches/0001-disable-sanitizers-1_3_1x.patch"
      patch_description: "Disable static code analysis"
      patch_type: "conan"
```

```
  "1.3.15":
    deps/ua-nodeset:
      url: "https://github.com/OPCFoundation/UA-Nodeset/archive/Glass=1.0.0-2022-01-01.zip"
      sha256: "032c93e3f7c335bbb2d2d699804a804c9cdce2a2ecc3fe7cde552f2130982c26"
      archive_pattern: "UA-Nodeset-{version}"
```
     
 Not sure if there's a policy on adding "older" versions, 1.3.15 is newer than 1.4.6, but older than 1.4.11.1
 Maybe it can't be done if this PR has been merged.

recipes/open62541/all/patches/0005-use-external-libxml2.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -1092,7 +1092,8 @@ if(UA_ENABLE_NODESETLOADER)
+     list(APPEND lib_sources ${NODESETLOADER_SOURCES})
+     list(APPEND plugin_headers ${PROJECT_SOURCE_DIR}/plugins/include/open62541/plugin/nodesetloader.h)
+     list(APPEND plugin_sources ${PROJECT_SOURCE_DIR}/plugins/ua_nodesetloader.c)
+-    list(APPEND open62541_LIBRARIES ${NODESETLOADER_DEPS_LIBS})
++    find_package(LibXml2 REQUIRED)
```
AbrilRBS, line 9: Conan's file name is `libxml2`
```suggestion
+    find_package(libxml2 REQUIRED)
```

recipes/open62541/all/conanfile.py
```diff
@@ -182,6 +194,8 @@ def requirements(self):
             self.requires("libwebsockets/4.3.2")
         if self.options.discovery == "With Multicast" or "multicast" in str(self.options.discovery):
             self.requires("pro-mdnsd/0.8.4")
+        if self.options.nodeset_loader:
+            self.requires("libxml2/2.13.4")
```
AbrilRBS, line 198: Libxml2 should be required using version ranges
```suggestion
            self.requires("libxml2/[>=2.12.5 <3]")
```

recipes/open62541/all/conanfile.py
```diff
@@ -170,6 +177,11 @@ def configure(self):
         if self.options.web_socket:
             self.options["libwebsockets"].with_ssl = self.options.encryption
 
+        if self.options.nodeset_loader:
```
AbrilRBS, line 180: As this option is removed in Windows, accessing it should be done thru a safe getter method
```suggestion
        if self.options.get_safe("nodeset_loader"):
```

Note that this is the case for every instance of option access in the PR, they should also be udpated!

recipes/open62541/all/patches/0001-disable-sanitizers-1_4_11_1.patch
AbrilRBS, line 1: Instead of this patch, why can't we just set `UA_ENABLE_DEBUG_SANITIZER` to `False` to avoid taking this branch?

AbrilRBS:
@ErikTempelaarVO let's move the discussion about the possibility of adding an old version in https://github.com/conan-io/conan-center-index/issues/27373 :)

recipes/open62541/all/conanfile.py
fischjo3SICKAG: Thank you for pointing this out. I have applied your suggestion in [commit 18ab99c4](https://github.com/conan-io/conan-center-index/pull/27372/commits/18ab99c497f2f477ab163f76b4cc53e9dfcba1dd). Additionally, I successfully built it locally both with the option removed and with the option enabled.

recipes/open62541/all/patches/0001-disable-sanitizers-1_4_11_1.patch
fischjo3SICKAG: I created a version that builds without the patch file [here](https://github.com/conan-io/conan-center-index/pull/27372/commits/c34e281259b8bd13d614f48b2d212c91d4b48882). It builds successfully on my machine.

ErikTempelaarVO: Looks like that option is available for all 1.4.x

```
$ git log -S UA_ENABLE_DEBUG_SANITIZER
commit a820093aa97d08334147ecb85d8602f917d4f511
Author: theirix <theirix@gmail.com>
Date:   Sun Jun 19 14:06:57 2022 +0300

    feat(build): add option UA_ENABLE_DEBUG_SANITIZER
```

```
$ git tag --contains a820093aa97d08334147ecb85d8602f917d4f511
prev-v1.4.2
v1.4.0
v1.4.0-rc1
v1.4.0-rc2
v1.4.1
v1.4.10
v1.4.11
v1.4.11.1
v1.4.2
v1.4.3
v1.4.4
v1.4.5
v1.4.6
v1.4.7
v1.4.8
v1.4.9
```

fischjo3SICKAG: There is a [patch file for 1.3.x](https://github.com/conan-io/conan-center-index/blob/master/recipes/open62541/all/patches/0001-disable-sanitizers-1_3_x.patch). The 1.4.11.1 patch file from my sided was initially created to replicate the patching behavior of earlier open62541 recipes.

fischjo3SICKAG:
@AbrilRBS I've tried to address all issues, do you think this is ok?

recipes/open62541/all/conanfile.py
```diff
@@ -170,6 +177,11 @@ def configure(self):
         if self.options.web_socket:
             self.options["libwebsockets"].with_ssl = self.options.encryption
 
+        if self.options.get_safe("nodeset_loader"):
+            self.options["libxml2"].iconv = False
+            self.options["libxml2"].shared = False
+            self.options["libxml2"].with_subunit = False
```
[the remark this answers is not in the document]
fischjo3SICKAG: Thank you for pointing this out. I've reviewed these options and found the following: 

I searched for usage of the `with_subunit` option, but couldn't find any current usage and don't recall why I added it in the first place, so I will remove it. 

Setting `iconv` and `shared` to `False` is not strictly required. However, since open62541 aims to be as self-contained and minimal as possible, I assumed setting these options to `False` would be the recommended settings by the open62541 maintainers. If we do not enforce these settings in `validate()`, users can still override them. My proposed solution is to set these options to `False`, but not enforce them in `validate`.

These changes are now reflected in the latest commit.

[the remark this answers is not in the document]
jcar87: Thanks for taking the time to provide more insights.

There tends to be a difference between what the developers/maintainers or a library do "in isolation", and what what package/repository maintainers do so that things "work together".

Depending on different factors, this could cause the shared variant of `open62541` to contain the symbols of `libxml2` - and they could also end up being linked from libxml2 standalone. The same symbols would end up defined in multiple libraries, leading to potential issues.  This is problematic and we very explicitly want to avoid that. 

> I assumed setting these options to False would be the recommended settings by the open62541 maintainers. 

Unless there is explicit documentation from the open62541 maintainers, or prior art in other repositories, please remove these changes- in general, we are asking contributors to limit the scope of PRs https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md#pull-request-scope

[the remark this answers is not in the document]
fischjo3SICKAG: @jcar87 When I submitted this pull request, https://github.com/conan-io/conan-center-index/pull/26457 was not yet complete. Now that it is finished, I have rebased my changes on the latest open62541 recipe and limited the scope of this pull request to 'Add NodesetLoader Option' only.

For the libxml2 options: I tested building without setting any options for libxml2, and it worked. So I have removed these options entirely.

fischjo3SICKAG:
Received a [reply](https://github.com/open62541/open62541/issues/7249#issuecomment-2909444119) from the open62541 maintainers. They recommend using the NodesetLoader version from the release tag. I have updated the submodule to the commit referenced in the open62541 release.

recipes/open62541/all/patches/0005-use-external-libxml2.patch
jcar87: This was not wrong to begin with, but both work.
The legacy canonical name is `LibXml2`(a built-in find module in cmake), and the libxml2 project uses `libxml2` all lowercase. CMake looks for `<lowercase>-config.cmake` _regardless_ of the name, so `LibXml2` (which is what is more widely in use), will happily match and be found, with the added advatange that if the build scripts evaluate `LibXml2_FOUND`, it will also be defined. Not the case here: both `find_package(LibXml2 REQUIRED)` and  `find_package(libxml2 REQUIRED)` are behaving exactly the same way - or should (once the libxml2 and/or cmakeconfigdeps stop providing find modules)

recipes/open62541/all/conanfile.py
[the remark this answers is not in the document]
fischjo3SICKAG: @jcar87 Can we resolve this conversation?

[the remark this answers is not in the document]
jcar87: Thanks for the follow up work! :)

recipes/open62541/all/conanfile.py
```diff
@@ -329,6 +347,8 @@ def generate(self):
         # Honor BUILD_SHARED_LIBS from conan_toolchain (see https://github.com/conan-io/conan/issues/11840)
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
 
+        if version >= "1.4.11.1":
+            tc.cache_variables["UA_ENABLE_NODESETLOADER"] = self.options.get_safe("nodeset_loader", False)
```
franramirez688: Lines above, it's removing the option for every OS but Linux:

```python
        # NodesetLoader has only rudimentary Windows support --> disabling for now. This might change in the future.
        if Version(self.version) < "1.4.11.1" or self.settings.os != "Linux":
            del self.options.nodeset_loader
```
So it explains why the CI macOS machines were failing (cc @AbrilRBS) and the purpose of this line is to avoid getting `None` instead of a boolean one:

```python
tc.cache_variables["UA_ENABLE_NODESETLOADER"] = self.options.get_safe("nodeset_loader", False)
```
Now, my question is, should we change the lines above or these into

```suggestion
        if version >= "1.4.11.1" and self.settings.os == "Linux":
            tc.cache_variables["UA_ENABLE_NODESETLOADER"] = self.options.get_safe("nodeset_loader")
```
?

fischjo3SICKAG: @franramirez688:

You're absolutely right - there's a logical inconsistency in the current code. The option is deleted for non-Linux platforms but then accessed with `get_safe()`.

I agree with your analysis. However, rather than using your suggested fix (which only sets the CMake variable for Linux), I prefer to explicitly set `UA_ENABLE_NODESETLOADER=False` for non-Linux platforms:

```python
if version >= "1.4.11.1":
    if self.settings.os == "Linux":
        tc.cache_variables["UA_ENABLE_NODESETLOADER"] = self.options.get_safe("nodeset_loader", False)
    else:
        tc.cache_variables["UA_ENABLE_NODESETLOADER"] = False
```

Although the current implementation of nodesetloader only supports Linux (see [nodesetloader README](https://github.com/open62541/open62541-nodeset-loader/blob/fbedde3f81c6c517ce0656661c13e46503b19342/README.md?plain=1#L10)), I expect future versions will support additional platforms. This approach has two advantages:

1. **Explicit behavior:** The CMake variable is always in a known state across all platforms, making the build configuration more predictable
2. **Future compatibility:** When nodesetloader gains support for additional platforms, we only need to update the OS condition rather than restructuring the logic

franramirez688: Thanks a lot for your comment 👏 
Ok, then I think that we can keep the code as it is.

franramirez688 (approved):
LGTM! 👏

fischjo3SICKAG:
Incorporated feedback from @franramirez688 and @AbrilRBS in 43c54a7. Dismissing stale reviews to reflect the latest changes. Thanks for your input!

recipes/open62541/all/conanfile.py
```diff
@@ -329,6 +347,12 @@ def generate(self):
         # Honor BUILD_SHARED_LIBS from conan_toolchain (see https://github.com/conan-io/conan/issues/11840)
         tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0077"] = "NEW"
 
+        if version >= "1.4.11.1":
+            if self.settings.os == "Linux":
+                tc.cache_variables["UA_ENABLE_NODESETLOADER"] = self.options.get_safe("nodeset_loader", False)
+            else:
+                tc.cache_variables["UA_ENABLE_NODESETLOADER"] = False
```
franramirez688: Pythonic and safer I guess 😉 
```suggestion
            tc.cache_variables["UA_ENABLE_NODESETLOADER"] = self.options.nodeset_loader if self.settings.os == "Linux" else False
```

fischjo3SICKAG:
Hi @abrilRBS,

Could you please take a look at this PR when you have a moment? All CI tests are passing and we've addressed the previous feedback. Would be great to get this merged once you approve the changes.

Thanks! 🙂

AbrilRBS (approved):

fischjo3SICKAG:
Thank you @abrilRBS, @jcar87, @ErikTempelaarVO, and @franramirez688 for the thorough reviews and guidance throughout this pull request. This was a productive and educational process that improved both the recipe and my understanding of Conan packaging best practices.

jcar87:
Thanks for the kind words, and apologies for the time it took to get this past the finish line!

Issue #27366, opened 2025-05-05
[package] sdl/3.2.6: Conflict between libalsa/1.2.12 and libalsa/1.2.10 in the graph


skhaz:
### Description

Only happened on Ubuntu, see here https://github.com/willtobyte/carimbo/actions/runs/14839224678/job/41657801338

### Package and Environment Details

```python
from conan import ConanFile
from conan.tools.cmake import CMakeDeps
from conan.tools.cmake import CMakeToolchain
from pathlib import Path


class Carimbo(ConanFile):
    settings = "os", "arch", "compiler", "build_type"

    def _not_webassembly(self):
        return str(self.settings.os).lower() not in {"emscripten"}

    def requirements(self):
        self.requires("fmt/11.1.4")
        self.requires("libspng/0.7.4")
        self.requires("nlohmann_json/3.12.0")
        self.requires("ogg/1.3.5")
        self.requires("openal-soft/1.23.1")
        self.requires("physfs/3.2.0")
        self.requires("sdl/3.2.6")
        self.requires("sol2/3.5.0")
        self.requires("vorbis/1.3.7")

        # self.requires("libalsa/1.2.12", override=True)

        if self._not_webassembly():
            self.requires("boost/1.87.0")
            self.requires("luajit/2.1.0-beta3")
            self.requires("openssl/3.4.1")

    def configure(self):
        self.options["boost"].header_only = True

        self.options["physfs"].sevenzip = True
        self.options["physfs"].zip = False
        self.options["physfs"].grp = False
        self.options["physfs"].wad = False
        self.options["physfs"].hog = False
        self.options["physfs"].mvl = False
        self.options["physfs"].qpak = False
        self.options["physfs"].slb = False
        self.options["physfs"].iso9660 = False
        self.options["physfs"].vdf = False

        if self._not_webassembly():
            self.options["sol2"].with_lua = "luajit"

    def generate(self):
        path = Path(self.build_folder) / "LICENSES"
        with path.open("w", encoding="utf-8") as out:
            for dependecy in self.dependencies.values():
                if dependecy.is_build_context or dependecy.package_folder is None:
                    continue
                package_id = f"{dependecy.ref.name}/{dependecy.ref.version}"
                for path in Path(dependecy.package_folder).rglob("*"):
                    if not path.is_file():
                        continue
                    name = path.name.lower()
                    if name.startswith(("license", "copying", "copyright")):
[...]

PR #27374, opened 2025-05-07
sdl3: Add version range for libalsa

AbrilRBS:
After checking with the team, we're moving to version ranges for libalsa allowing different patch versions - please do not open further PRs to other recipes to update this dependency there unless you are facing a version conflict (do report that, thanks!), we'll do it progressively.

Closes https://github.com/conan-io/conan-center-index/issues/27366

### DIFF
--- a/recipes/sdl/3.x/conanfile.py
+++ b/recipes/sdl/3.x/conanfile.py
@@ -227,7 +227,7 @@ class SDLConan(ConanFile):
         if self.options.get_safe("pulseaudio"):
             self.requires("pulseaudio/17.0")
         if self.options.get_safe("alsa"):
-            self.requires("libalsa/1.2.12")
+            self.requires("libalsa/[>=1.2 <1.3]")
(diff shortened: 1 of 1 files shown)


jcar87 (approved):

jcar87:
(before merging, worth looking at the hook warnings - Iooks like they're valid and should be addressed)

recipes/sdl/3.x/conanfile.py
```diff
@@ -339,6 +339,9 @@ def package(self):
         copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
         cmake = CMake(self)
         cmake.install()
+        rmdir(self, os.path.join(self.package_folder, "lib", "cmake"))
```
jcar87, line 342: on windows these seem to be in `cmake` rather than `lib/cmake`?

jcar87 (approved):

PR #27389, opened 2025-05-09
Add support for cmake 4 in reported libraries

AbrilRBS:
For the reports in https://github.com/conan-io/conan-center-index/issues/26878 for the last few weeks

PR #27391, opened 2025-05-09
wxWidgets: Add version 3.2.8

irieger:
### Summary
Changes to recipe:  **wxwidgets/3.2.8**

#### Motivation
Previous version of wxWidgets don't build as it uses a function that is no longer available with macOS 15.0 and later. Due to that I currently can't build an app I'm working on here.

#### Details
https://github.com/wxWidgets/wxWidgets/releases/tag/v3.2.8

Error I get with version 3.2.6 which was solved by 3.2.7 or 3.2.8:
```
/Users/ingmar/.conan2/p/b/wxwidbb69dd0fdc132/b/src/src/osx/carbon/dcscreen.cpp:73:13: error: 'CGDisplayCreateImage' is unavailable: obsoleted in macOS 15.0 - Please use ScreenCaptureKit instead.
   73 |     image = CGDisplayCreateImage(kCGDirectMainDisplay);
      |             ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.4.sdk/System/Library/Frameworks/CoreGraphics.framework/Headers/CGDirectDisplay.h:383:33: note: 'CGDisplayCreateImage' has been explicitly marked unavailable here
  383 | CG_EXTERN CGImageRef __nullable CGDisplayCreateImage(CGDirectDisplayID displayID)
      |                                 ^
1 error generated.
```

---


recipes/wxwidgets/all/conanfile.py
```diff
@@ -177,7 +177,7 @@ def _patch_sources(self):
                         "CMAKE_OSX_DEPLOYMENT_TARGET",
                         "CMAKE_OSX_DEPLOYMENT_TARGET_IGNORED")
         # Fix for strcpy_s (fix upstream?)
-        if is_apple_os(self):
```
irieger, line 180: Can't fully judge this but it build with disabling this option. Otherwise it needs to be adapted as the current version uses cmake_version `3.0...3.31` now.

(This from my understanding only influences the policy and newer CMake versions should still work, right?)

irieger:
The change triggers a rebuild of old packages which now fail due Cmake 4.0 incompatiblity. Wonder if I should remove the older versions?

AbrilRBS:
I'll retrigger after https://github.com/conan-io/conan-center-index/pull/27389 is merged, thanks for providing a version that compiles in newer macos :)

irieger:
> I'll retrigger after #27389 is merged, thanks for providing a version that compiles in newer macos :)

Thanks. Could you also have a look into the error on Linux which comes from a missing system dependency that the original package requires?

recipes/wxwidgets/all/conanfile.py
```diff
@@ -180,6 +180,8 @@ def _patch_sources(self):
         # Fix for strcpy_s (fix upstream?)
         if is_apple_os(self):
```
AbrilRBS, line 181: This fix is present in all available versions of the recipe, we might want to remove it

AbrilRBS (approved):

irieger:
@AbrilRBS Looks like that patch isn't unnecessary at all. Locally on macOS 15 it built but not on the older version in the CI. (This is why I at first excluded it for the current version, as I thought it is unnecessary. But it only worked locally, not in the CI afair.)

AbrilRBS:
@irieger seems like you're right, feel free to revert, sorry :)

jcar87 (approved):
Thanks!

PR #27397, opened 2025-05-11
[botan] add versions 3.7.1 and 3.8.1

Phatcat:
Title says it all.

### Summary
Changes to recipe:  **botan/[3.8.1]**

#### Motivation

Version bump.

#### Details

It's just adding a new version, not changing any existing code.

---
This is not a duplicate but someone PR'd adding 3.7.0 earlier and I suggest just skipping it and adding 3.8.1 and 3.7.1 (not 3.8.0 as it's failing on newer clang)
(this is needed for a buildrunner, but I verified that it indeed exists and verified that hash with several tools)


jcar87:
Thanks @Phatcat  - please limit this to only adding the most recent version, unless there's a strong reason why a user may not be able to move up to 3.8.1, thanks!

PR #27421, opened 2025-05-14
qt: Add Qt:: namespace alias for every module.

cadmiy:
### Summary
Changes to recipe:  **qt/5.x.x**, **qt/6.x.x**

#### Motivation
Qt6 supports `Qt::` _namespace_ for cmake targets. This namespace has also been backported to Qt5.
However, the namespace is not available in conan recipes. The problem arose when we were trying to switch from Qt5 to Qt6.

#### Details
Inside `Qt5{module}Config.cmake` files for every module there is an alias for Qt5/Qt6 namespace to Qt. For example, for `Qt5CoreConfig.cmake` it looks like this:
```
# Add a versionless target, for compatibility with Qt6.
if(NOT "${QT_NO_CREATE_VERSIONLESS_TARGETS}" AND TARGET Qt5::Core AND NOT TARGET Qt::Core)
    add_library(Qt::Core INTERFACE IMPORTED)
    set_target_properties(Qt::Core PROPERTIES
        INTERFACE_LINK_LIBRARIES "Qt5::Core"
    )
endif()
```
However, this alias is not available in conan recipes. The suggested changes fix that.

---

### DIFF
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -1058,6 +1058,7 @@ Prefix = ..""")
             componentname = f"qt{module}"
             assert componentname not in self.cpp_info.components, f"Module {module} already present in self.cpp_info.components"
             self.cpp_info.components[componentname].set_property("cmake_target_name", f"Qt5::{module}")
+            self.cpp_info.components[componentname].set_property("cmake_target_aliases", [f"Qt::{module}"])
             self.cpp_info.components[componentname].set_property("pkg_config_name", f"Qt5{module}")
             self.cpp_info.components[componentname].names["cmake_find_package"] = module
             self.cpp_info.components[componentname].names["cmake_find_package_multi"] = module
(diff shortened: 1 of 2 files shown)


jcar87:
Closing in favour of https://github.com/conan-io/conan-center-index/pull/25198 - which was opened earlier and was more complete.

Please note that Qt themselves [recommend](https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#recommended-practices) using the _versioned_ CMake targets - rather than the compatibility ones.

This always had a workaround, also documented by Qt here: https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#supporting-qt-5-versions-older-than-5-15

cadmiy:
Thanks for merging!

> Please note that Qt themselves [recommend](https://doc.qt.io/qt-6/cmake-qt5-and-qt6-compatibility.html#recommended-practices) using the _versioned_ CMake targets - rather than the compatibility ones.

Qt says, `Use the versioned targets unless you have to support Qt 5 and Qt 6 in the same project.`, which is exactly my case.

PR #27426, opened 2025-05-14
 Add sqlitemap header-only library

bw-hro:
### Summary
Changes to recipe:  **sqlitemap/[1.1.0]**

#### Motivation
Add [sqlitemap](https://github.com/bw-hro/sqlitemap)

#### Details
sqlitemap is a lightweight C++ wrapper around SQLite(only dependency) that provides a simple, map-like interface.
It’s designed to make working with key-value storage easy and convenient. The library is implemented as a single-header file and is licensed under the MIT License.

I have tested it successfully on a local Ubuntu with gcc 13 and on a local Windows with MSVC 19.

This is the first time I created a recipe, so hope It match all requirements. In case anything else is needed I'm open for feedback : ) 

---

### DIFF
--- a/recipes/sqlitemap/all/conanfile.py
+++ b/recipes/sqlitemap/all/conanfile.py
@@ -1,0 +1,52 @@
+from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.build import check_min_cppstd
+from conan.tools.files import apply_conandata_patches, copy, export_conandata_patches, get
+from conan.tools.layout import basic_layout
+import os
+
+required_conan_version = ">=2.0"
+
+
+class PackageConan(ConanFile):
+    name = "sqlitemap"
+    description = "sqlitemap - Persistent Map Backed by SQLite"
+    license = "MIT"
+    url = "https://github.com/conan-io/conan-center-index"
+    homepage = "https://github.com/bw-hro/sqlitemap"
+    topics = ("sqlite", "database", "hash map", "header-only")
+    package_type = "header-library"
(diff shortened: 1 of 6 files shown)


recipes/sqlitemap/all/conanfile.py
```diff
@@ -0,0 +1,45 @@
...
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+
+    def package(self):
+        copy(self, "LICENSE.txt", self.source_folder, os.path.join(self.package_folder, "licenses"))
+        copy(self, "*", os.path.join(self.source_folder, "include"), os.path.join(self.package_folder, "include"))
+
+    def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "sqlitemap")
+        self.cpp_info.set_property("cmake_target_name", "sqlitemap::sqlitemap")
```
AbrilRBS, line 42: I'm trusting you as the maintainer for it to be right, as I couldn't find anywhere in the documentation what the expected target name would be :)

AbrilRBS (commented):
Thanks! I've cleaned up some minor things, but the changes were great :)

AbrilRBS (approved):
Actually approve it

recipes/sqlitemap/all/conanfile.py
bw-hro: Hi @AbrilRBS, thanks for you review and adjustments. Yes this is the intended target.

perseoGI (approved):
Thank you, very clean recipe!

PR #27437, opened 2025-05-16
geographiclib: do not build tools when option tools=False, fixes #27433

### LINKED ISSUE #27433: [package] geographiclib/2.4: fails to build for iOS
### Description

The recipe has a boolean option `tools`, which is obviously meant to enable/disable the building of GeographicLib tools along with the library. This is implemented in such a way that the tools are first built anyway, and then deleted before being installed.

This doesn't work when compiling for iOS, because executables have to be signed, need a bundle-identifier, yadda yadda...

* Package Name/Version: **geographiclib/2.4**
* Operating System+version: **build: Macos, host: iOS**
* Compiler+version: **apple-clang / any version**
* Conan version: **2.13.0**

### Conan profile

Profile host:
[settings]
arch=armv8
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu20
compiler.libcxx=libc++
compiler.version=16
os=iOS
os.sdk=iphoneos
os.version=16.0
[options]
[conf]
tools.apple:sdk_path=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk
tools.build:cxxflags=['-DBOOST_NO_CXX98_FUNCTION_BASE', '-D_HAS_AUTO_PTR_ETC=0', '-Wno-error=enum-constexpr-conversion']
tools.build:sysroot=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk
tools.cmake.cmaketoolchain:generator=Xcode

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=apple-clang
compiler.cppstd=gnu17
compiler.libcxx=libc++
compiler.version=16
os=Macos
[conf]
tools.build:cxxflags=['-DBOOST_NO_CXX98_FUNCTION_BASE', '-D_HAS_AUTO_PTR_ETC=0', '-Wno-error=enum-constexpr-conversion']
tools.cmake.cmaketoolchain:generator=Xcode

### Steps to reproduce

conan install -r conancenter --requires=geographiclib/2.4 -pr:b=default -pr:h=ios -b geographiclib/2.4

### Logs

```
GeographicLib.xcodeproj: error: Bundle identifier is missing. CartConvert doesn't have a bundle identifier. Add a value for PRODUCT_BUNDLE_IDENTIFIER in the build settings editor. (in target 'CartConvert' from project 'GeographicLib')
...
... same error for every other included tool ...
...
```

</details>

maysl:
### Summary
Changes to recipe:  **geographiclib/\***

#### Motivation
Fix #27433

#### Details
When the boolean option `tools` is `False`, do not build the tools. Uses a mostly undocumented cmake mechanism, maintainer comment here: https://github.com/geographiclib/geographiclib/pull/39#issuecomment-2885315450

---

### DIFF
--- a/recipes/geographiclib/all/conanfile.py
+++ b/recipes/geographiclib/all/conanfile.py
@@ -118,6 +118,9 @@ class GeographiclibConan(ConanFile):
         tc = CMakeToolchain(self)
         tc.variables["GEOGRAPHICLIB_LIB_TYPE"] = "SHARED" if self.options.shared else "STATIC"
         tc.variables["GEOGRAPHICLIB_PRECISION"] = self._cmake_option_precision
+        if (not self.options.tools) and (Version(self.version) >= "2.3"):
+            # https://github.com/geographiclib/geographiclib/pull/39#issuecomment-2885315450
+            tc.variables["BINDIR"] = "OFF"
         tc.generate()
 
         VirtualBuildEnv(self).generate()
@@ -152,7 +155,7 @@ class GeographiclibConan(ConanFile):
         ]:
             rmdir(self, os.path.join(os.path.join(self.package_folder, folder)))
         rm(self, "*.pdb", os.path.join(self.package_folder, "bin"))
-        if not self.options.tools:
+        if (not self.options.tools) and (Version(self.version) < "2.3"):
             rmdir(self, os.path.join(self.package_folder, "sbin"))
             bin_files = [it for it in os.listdir(os.path.join(self.package_folder, "bin")) if not it.endswith(".dll")]
             for it in bin_files:


maysl:
bump

maysl:
@conan-io/barbarians please approve the CI run for this PR, so that the pipeline can progress. Thank you!

uilianries (commented):
@maysl Hello! Sorry for delaying your pull request review. 

Thank you for checking with the upstream about building the tools and pointing it in this PR.

As this Geographiclib is built always with `tools=True` in CI, we can not validate it without tools, as your change targets. So I built locally every version to be able to compare:

- [geographiclib-1.50.2-no-tools.log](https://github.com/user-attachments/files/22539881/geographiclib-1.50.2-no-tools.log)
- [geographiclib-1.51-no-tools.log](https://github.com/user-attachments/files/22539879/geographiclib-1.51-no-tools.log)
- [geographiclib-1.52-no-tools.log](https://github.com/user-attachments/files/22539880/geographiclib-1.52-no-tools.log)
- [geographiclib-2.3-no-tools.log](https://github.com/user-attachments/files/22539882/geographiclib-2.3-no-tools.log)
- [geographiclib-2.4-no-tools.log](https://github.com/user-attachments/files/22539883/geographiclib-2.4-no-tools.log)

The versions 1.5x.z always built tools; they ignore `BINDIR`. It result: It will need to revert your change in `package()` to keep the package without tools.

Version 2.3 always builds tools, but does not package them when `BINDIR` is disabled; 

Version 2.4 is the big win: It does not build tools at all when that option is disabled.

maysl:
@uilianries thanks for the catch, I had completely ignored the fact that older versions are still supposed to be supported.

I have pushed a variant that works for older versions too.

`2.3` is still somewhat problematic on iOS, because while the tools are now correctly skipped, this version also builds test executables. I think this should be fixed separately though.

uilianries:
@maysl Hello, Thank you for your follow-up. Please, let's simplify it without needing patches. I understand old version are not flexible and they will build tools always, but I don't think it worth trying to fix it as it will diverge from the upstream. 

As I commented before:

> The versions 1.5x.z always built tools; they ignore BINDIR. It result: It will need to revert your change in package() to keep the package without tools.

> Version 2.3 always builds tools, but does not package them when BINDIR is disabled;

> Version 2.4 is the big win: It does not build tools at all when that option is disabled.

For version 1.5 you just can keep it as the original: Remove binaries only from the package folder.
For 2.3 you can disable BINDIR. 

Regards;

maysl:
@uilianries I am generally fine with doing it that way, but the result would be that only `2.4` is usable on iOS (iOS being the motivation for the fix in the first place). In other words, the fix would only affect the newest version of geographiclib. All older versions would remain "unfixed".

I think the concern about diverging from upstream is not too big for versions that are no longer maintained there.

uilianries:
> In other words, the fix would only affect the newest version of geographiclib. All older versions would remain "unfixed".

@maysl Thank you for clarifying. No problem, any person who consumes old versions from the upstream would have the same problem as well.

maysl:
@uilianries point taken, pushed a simplified new version

Issue #27428, opened 2025-05-15
[package] ktx/4.3.2:Does not compile with Clang 20


n0F4x:
### Description

_ktx_ has an outdated version of _fmt_ in its source, which causes the issue.

### Package and Environment Details

* Package Name/Version: **ktx/4.3.2**
* Operating System+version: **Windows MSYS2**
* Compiler+version: **Clang 20**
* Conan version: **conan 2.16.1**
* Python version: **Python 3.13.1**

### Conan profile

[settings]
arch=x86_64
build_type=Debug
compiler=clang
compiler.cppstd=26
compiler.libcxx=libc++
compiler.version=20
os=Windows
[conf]
tools.cmake.cmaketoolchain:generator=Ninja
tools.info.package_id:confs=['tools.build:cflags', 'tools.build:cxxflags', 'tools.build:exelinkflags', 'tools.build:sharedlinkflags']
[buildenv]
PATH+=(path)C:/dev/MSYS2/clang64/bin
PATH+=(path)C:/Users/Gabor/AppData/Local/Programs/CLion 2/bin/cmake/win/x64/bin
PATH+=(path)C:/Users/Gabor/AppData/Local/Programs/CLion 2/bin/ninja/win/x64
[runenv]
PATH+=(path)C:/dev/MSYS2/clang64/bin

### Steps to reproduce

```py
def requirements(self):
    self.requires("ktx/4.3.2")
```
`conan install .`

### Logs

<details><summary>Click to expand log</summary>

```
/.conan2/p/b/ktxe2d82b5ecdfa0/b/src/external/fmt/include/fmt/format-inl.h:61:19: error: call to consteval function 'fmt::basic_format_string<char, fmt::basic_string_view<char> &, const char (&)[3]>::basic_format_string<FMT_COMPILE_STRING, 0>' is not a constant expression
   61 |     format_to(it, FMT_STRING("{}{}"), message, SEP);
      |                   ^
...
```

</details>

AbrilRBS:
Hi @n0F4x thanks a lot for the report

It seems like ktx is vendorizing fmt in their source code, but this is only used when compiling the tools, so for now the easiest way to avoid this is to compile with the `tools=False` option set.

We'll look into providing a PR (contributions welcome!) that unvendorizes this for a library coming from CCI, but this issue is something that the people at ktx might want to know about so they can update their vendorized version of fmt for some next release if you are able to report this to them too :)

PR #27448, opened 2025-05-17
ktx: Unvendorize fmt

AbrilRBS:
See reasoning and discussion in https://github.com/conan-io/conan-center-index/issues/27428

### DIFF
--- a/recipes/ktx/all/conanfile.py
+++ b/recipes/ktx/all/conanfile.py
@@ -80,7 +80,7 @@ class KtxConan(ConanFile):
             self.requires("lodepng/cci.20230410")
         self.requires("zstd/1.5.5")
         if self.options.tools:
-            self.requires("fmt/10.2.1")
+            self.requires("fmt/10.2.1", transitive_libs=False)
 
     def validate(self):
         if self.settings.compiler.get_safe("cppstd"):
--- a/recipes/ktx/all/conanfile.py
+++ b/recipes/ktx/all/conanfile.py
@@ -79,6 +79,8 @@ class KtxConan(ConanFile):
             # https://github.com/KhronosGroup/KTX-Software/blob/v4.2.1/tools/imageio/png.imageio/lodepng.h#L26-L32
             self.requires("lodepng/cci.20230410")
         self.requires("zstd/1.5.5")
+        if self.options.tools:
+            self.requires("fmt/10.2.1")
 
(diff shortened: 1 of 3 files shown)


perseoGI (approved):
Tested locally, the unvendorize worked well:
```

ktx/4.3.2 (test package): Running test()
ktx/4.3.2 (test package): RUN: ./test_package
Test: File open failed. (expected)
```

recipes/ktx/all/conanfile.py
```diff
@@ -159,16 +161,9 @@ def package_info(self):
...
-        self.cpp_info.names["cmake_find_package_multi"] = "KTX"
-        self.cpp_info.components["libktx"].names["cmake_find_package"] = "ktx"
-        self.cpp_info.components["libktx"].names["cmake_find_package_multi"] = "ktx"
         self.cpp_info.components["libktx"].set_property("cmake_target_name", "KTX::ktx")
         self.cpp_info.components["libktx"].requires = ["zstd::zstd"]
         if Version(self.version) < "4.2.0":
             self.cpp_info.components["libktx"].requires.append("lodepng::lodepng")
         if self.options.tools:
-            self.env_info.PATH.append(os.path.join(self.package_folder, "bin"))
+            self.cpp_info.components["libktx"].requires.append("fmt::fmt")
```
jcar87, line 169: given

```
       if self.options.tools:
            self.requires("fmt/10.2.1")
```

libktx does not depend on fmt - this is a dependency that shouldn't propagate linkage  - what am I missing?

uilianries: @AbrilRBS @perseoGI Hello! @jcar87 is right, there is no linkage of fmt here: [ktx-4.3.2-macos-clang-unverdor-fmt.log](https://github.com/user-attachments/files/20576857/ktx-4.3.2-macos-clang-unverdor-fmt.log)


You can find fmt usage in tools/xtx, which builds ktxtools executable, but nothing related to the library.

jcar87: This is needed for

```
ERROR: ktx/4.3.2: package_info(): The direct dependency 'fmt' is not used by any '(cpp_info/components).requires'.
```

I've added `transitive_libs=False` to mitigate/prevent the propagation of linkage, however this probably needs to be addressed properly in the client side

PR #17030, opened 2023-04-14
qt 5.x: several fixes for conan v2

SpaceIm:
- inject `VirtualRunEnv` in `build` scope (in case of native build) since path to shared libs of several requirements must be visible when executables built during qt build process are executed (because these executables link to some requirements).
- Also fix `DYLD_LIBRARY_PATH` workaround on macOS, it was broken (it's the same usage than previous point, but as usual SIP defeats everything, leading to this workaround).
- as a consequence, also remove some build requirements erroneously added by https://github.com/conan-io/conan-center-index/pull/16404 (see https://github.com/conan-io/conan-center-index/pull/16404#discussion_r1148160492). They are requirements actually, but their shared libs must be visible in PATH/LD_LIBRARY_PATH/DYLD_LIBRARY_PATH during build process as described in previous point.
- don't rely on `is_msvc()` in `package_id()` method, it's forbidden in conan v2.
- lowecase system libs if Windows (`uxtheme` instead of `UxTheme`)
- use modern generators & helpers in test package, and add test_v1_package
- use version range for openssl
- bump several dependencies
- prefer `openal-soft` in requirements over deprecated `openal` recipe

---

- [ ] I've read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md).
- [ ] I've used a [recent](https://github.com/conan-io/conan/releases/latest) Conan client version close to the [currently deployed](https://github.com/conan-io/conan-center-index/blob/master/.c3i/config_v1.yml#L6).
- [ ] I've tried at least one configuration locally with the [conan-center hook](https://github.com/conan-io/hooks.git) activated.

PR #27449, opened 2025-05-17
Qt5: fix cross-compilation on Windows

### LINKED ISSUE #15403: [question] How to build Qt with its dependencies as shared libraries
### What is your question?

As an example, I would like to build `qt/5.15.6` with `zlib:shared=True`.

This is something that I have tried, and it failed on Windows (Linux build worked, probably because `zlib` is installed in the docker image).
Example of a MinGW build:
```shell
moc C:\.conan\dd64af\1\qt5\qtbase\src\corelib\animation\qabstractanimation.h 
mingw32-make[4]: *** [Makefile.Release:10149: .moc/release/moc_qabstractanimation.cpp] Error -1073741515
mingw32-make[3]: *** [Makefile:45: release] Error 2
```
Note that `Error -1073741515` means `0xFFFFFFFFC0000135`: missing dll.

After investigating: `moc` (and also `rcc`) depends on `zlib1.dll`.
Building with `zlib:shared=False` works.

The question: should something be changed in the recipes to be able to build Qt with its dependencies built as shared libraries ?

I have tried to update the `qt/5.x.x` recipe on my fork to use [VirtualBuildEnv](https://docs.conan.io/en/latest/reference/conanfile/tools/env/virtualbuildenv.html):
```python
...
from conan.tools.env import VirtualBuildEnv
...
generators = "pkg_config", "VirtualBuildEnv"
```
(Here is the [diff](https://github.com/scandyna/conan-center-index/commit/3ff968d49eeb697ab6e7447cfca3a52869a9fd85)).

This does not solve the problem.

Should the `zlib` recipe also be updated to bring its shared libraries to the `PATH` ?
As an example for Windows only:
```python
def package_info(self):
  self.buildenv_info.prepend_path( "PATH", os.path.join(self.package_folder, "bin") )
```

If the above is true, should every dependency be updated ?
Or is there another mechanism for that ?

Some discussions that seem to be similar:
- #11118
- https://github.com/conan-io/conan-center-index/pull/11703

Note: I only mention `zlib` here, but building other dependencies, like `icu` , as shared libraries is also expected.

kambala-decapitator:
### Summary
Changes to recipe:  **qt/5.x.x**

#### Motivation

- fixes cross-compilation on Windows, e.g. x64 -> x86 with msvc
- fixes accessing an unavailable property in the recipe

Closes #15403 with the final fix

#### Details

I started using Conan in Windows builds and noticed that x64 build succeeds, but x64 -> x86 fails with a rather cryptic error, an example can be seen [here](https://github.com/vcmi/vcmi-dependencies/actions/runs/15064116693):

> rcc C:\Users\runneradmin\.conan2\p\qtc879bf41d37a5\s\qt5\qtbase\src\gui\painting\qpdf.qrc 
> jom: C:\Users\runneradmin\.conan2\p\b\qt239c4430fe06b\b\build_folder\qtbase\src\gui\Makefile.Release [.rcc\release\qrc_qpdf.cpp] Error 2

After some time, I figured out that `rcc.exe` links to `zlib` and I'm building zlib as shared library. #17030 had handled this case when not cross-compiling, but Windows is rather special here.

Also, when testing my fix, I noticed an exception regarding accessing `settings.compiler.libcxx`, so fixed it as well. The exception was:

```
ERROR: qt/5.15.16: Error in validate() method, line 357
        if not self.options.with_doubleconversion and str(self.settings.compiler.libcxx) != "libc++":
        ConanException: 'settings.compiler.libcxx' doesn't exist for 'msvc'
'settings.compiler' possible configurations are ['cppstd', 'runtime', 'runtime_type', 'toolset', 'update', 'version']
```

Successfully built 5.15.16 locally in the following configuration:

```
Profile host:
[settings]
arch=x86
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows
[options]
qt/*:openssl=False
qt/*:shared=True
qt/*:widgets=False
qt/*:with_atspi=False
qt/*:with_dbus=False
qt/*:with_freetype=False
qt/*:with_glib=False
qt/*:with_gstreamer=False
qt/*:with_harfbuzz=False
qt/*:with_libjpeg=False
qt/*:with_libpng=False
qt/*:with_md4c=False
qt/*:with_mysql=False
qt/*:with_odbc=False
qt/*:with_openal=False
qt/*:with_pcre2=False
qt/*:with_pq=False
qt/*:with_pulseaudio=False
qt/*:with_sqlite3=False
qt/*:with_vulkan=False
qt/*:with_zstd=False
zlib/*:shared=True
[platform_tool_requires]
cmake/3.31.0
jom/1.1.3
ninja/1.12.1
[conf]

Profile build:
[settings]
arch=x86_64
build_type=Release
compiler=msvc
compiler.cppstd=14
compiler.runtime=dynamic
compiler.runtime_type=Release
compiler.version=194
os=Windows
[platform_tool_requires]
cmake/3.31.0
jom/1.1.3
ninja/1.12.1
[conf]
```

P.S. In Qt 6 `rcc.exe` links to `zstd` instead, but may need a similar fix (haven't checked the current state).

---

### DIFF
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -354,7 +354,7 @@ class QtConan(ConanFile):
         if self.options.get_safe("with_fontconfig", False) and not self.options.get_safe("with_freetype", False):
             raise ConanInvalidConfiguration("with_fontconfig cannot be enabled if with_freetype is disabled.")
 
-        if not self.options.with_doubleconversion and str(self.settings.compiler.libcxx) != "libc++":
+        if not self.options.with_doubleconversion and self.settings.get_safe("compiler.libcxx") != "libc++":
             raise ConanInvalidConfiguration("Qt without libc++ needs qt:with_doubleconversion. "
                                             "Either enable qt:with_doubleconversion or switch to libc++")
 
--- a/recipes/qt/5.x.x/conanfile.py
+++ b/recipes/qt/5.x.x/conanfile.py
@@ -527,7 +527,7 @@ class QtConan(ConanFile):
         ms.generate()
         vbe = VirtualBuildEnv(self)
         vbe.generate()
-        if not cross_building(self):
+        if not cross_building(self) or self.settings.os == "Windows":
             vre = VirtualRunEnv(self)
             vre.generate(scope="build")
         env = Environment()


recipes/qt/5.x.x/conanfile.py
```diff
@@ -527,7 +527,7 @@ def generate(self):
         ms.generate()
         vbe = VirtualBuildEnv(self)
         vbe.generate()
-        if not cross_building(self):
+        if not cross_building(self) or self.settings.os == "Windows":
```
kambala-decapitator, line 530: do you think it's better to check for "both build and host OS are Windows"?

franramirez688: I think that I'd try something like this:

```suggestion
        if not cross_building(self,  skip_x64_x86=self.settings.os == "Windows"):
```

franramirez688 (commented):
Thanks for this PR @kambala-decapitator 👏 

LGTM! Small suggestion related to the fix before approving 😁

franramirez688 (approved):
LGTM! 👏

PR #27469, opened 2025-05-19
msys2: clear package cache

ericLemanissier:
### Summary
Changes to recipe:  **msys2/cci.latest**

#### Motivation
it removes the content of bin\msys64\var\cache\pacman\pkg saving more than 100MB

#### Details

---

### DIFF
--- a/recipes/msys2/all/conanfile.py
+++ b/recipes/msys2/all/conanfile.py
@@ -144,6 +144,7 @@ class MSYS2Conan(ConanFile):
             for package in ['pkgconf']:
                 if self.run(f'bash -l -c "pacman -Qq {package}"', ignore_errors=True, quiet=True) == 0:
                     self.run(f'bash -l -c "pacman -Rs -d -d {package} --noconfirm"')
+            self.run(f'bash -l -c "pacman -Scc --noconfirm"')
 
         self._kill_pacman()
 


ericLemanissier:
We can see this worked by the absence of .zst files in the package. Previously there as `msys2/cci.latest: package(): Packaged 105 '.zst' files`

ericLemanissier:
quoting https://wiki.archlinux.org/title/Pacman#Cleaning_the_package_cache:

> To remove all files from the cache, use the clean switch twice, this is the most aggressive approach and will leave nothing in the cache directory:
> 
> `# pacman -Scc`
> Warning: One should avoid deleting from the cache all past versions of installed packages and all uninstalled packages unless one desperately needs to free some disk space. This will prevent downgrading or reinstalling packages without downloading them again.

As conan binaries are supposed to be immutable, user should never downgrade or reinstall any pacman package  (stored in conan msys2 binary). So the warning is not relevant in this context. If a user still wants to downgrade or reinstall any pacman package, they can still re-download it, so this change is not a limitation.

AbrilRBS (approved):
Thanks a lot for the PR and the explanation/context, this will save some bandwidth :)

uilianries (approved):
LGTM. That command is legitimate: https://wiki.archlinux.org/title/Pacman#Cleaning_the_package_cache.

BTW, I use Arch.

PR #27470, opened 2025-05-19
Add a new recipe : cpp-smtpclient-library

jeremydumais:
### Summary
Add a new recipe: cpp-smtpclient-library/1.1.10

#### Motivation
I want my C++ SMTP client library to be easily accessible in users projects.

#### Details
This is really a new recipe for me and my first contribution to the Conan Center Index so please let me know if improvement in my recipe is required.

The CPP-SMTPClient-Library is a simple SMTP client library built in C++ that support authentication and secure connections (Opportunistic SSL/TLS and Forced SSL encryption). Works on Windows, Linux and macOS.

---


perseoGI (approved):
Greetings @jeremydumais 
I've cleaned up a little bit your recipe.

Some changes:
- Recipes in CCI can't have a fixed `version` attribute as it is retrieved from `conandata.yml`. 
- In the `source()` method recipes must use URL information from `conandata.yml` file.
- `conan` minimum requirement is CMake 3.15 so any recipe which does not have a minimum requirement of CMake greater than 3.15 doesn't need to add a CMake build requirement. See https://github.com/conan-io/conan-center-index/blob/889c38ef689596787f409603d873585c5d25d9c7/docs/adding_packages/dependencies.md#version-ranges: 

> Note that CCI recipes assume 3.15 is installed in the system, so add this version range only when a requirement for a newer version is needed.
- Always use `os.path.join` when declaring paths as Windows users could be affected.

The rest looks great!
Thank you for taking the time to add your library to CCI! 🐸

jeremydumais:
Great, thanks for the info! I will fix all that and my next submission will be clean :)

perseoGI:
@jeremydumais no need to fix it, I've already pushed to your branch 😄

AbrilRBS (approved):
Thanks!

PR #27166, opened 2025-04-07
sdl_ttf: Add version 3.2.2

AbrilRBS:
This PR adds support for version 3.2.2 of sdl_ttf after the releae of sdl3

Note that this PR contains 2 recipes:

* Plutosvg, a new dependency of sdl_ttf since 3.x, which depends on the available plutovg
* the new sdl_ttf 3.x recipe, as that is different enough from the 2.x version

Closes https://github.com/conan-io/conan-center-index/issues/26832

PR #27474, opened 2025-05-19
Plutosvg: Add new recipe

AbrilRBS:
Necessary for sdl_ttf v3, split from https://github.com/conan-io/conan-center-index/pull/27166

`plutovg`: fix include directory expected by consumers

### DIFF
--- a/recipes/plutosvg/all/conanfile.py
+++ b/recipes/plutosvg/all/conanfile.py
@@ -6,7 +6,7 @@ from conan.tools.layout import basic_layout
 from conan.tools.meson import Meson, MesonToolchain
 import os
 
-required_conan_version = ">=2"
+required_conan_version = ">=2.4"
 
 class PlutoSVGConan(ConanFile):
     name = "plutosvg"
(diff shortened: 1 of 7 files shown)


perseoGI (approved):
LGTM 🐸

perseoGI (dismissed):

recipes/plutosvg/all/conanfile.py
```diff
@@ -0,0 +1,89 @@
...
+        self.tool_requires("meson/[>=1.2.3 <2]")
+        if not self.conf.get("tools.gnu:pkg_config", default=False, check_type=str):
+            self.tool_requires("pkgconf/[>=2.2 <3]")
+
+    def source(self):
+        get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        # Seems like upstream places the header in a different folder from where the installed header is
+        replace_in_file(self, os.path.join(self.source_folder, "source", "plutosvg.h"),
+                        "#include <plutovg.h>",
+                        "#include <plutovg/plutovg.h>")
```
jcar87, line 53: This is indicative of an issue with the `plutovg` package. 
Their example uses https://github.com/sammycage/plutovg/blob/main/examples/smiley.c#L1

The recipe is using meson, but plutovg supports both meson and cmake
https://github.com/sammycage/plutovg?tab=readme-ov-file#installation

I wouldn't be surprised if the meson scripts had a different behaviour than CMake. We _really_, really need to stop using untested meson build scripts if that is the case

perseoGI (approved):

Issue #26916, opened 2025-03-21
[package] protobuf/5.27.0: duplicate symbols from abseil dependency


lia-viam:
### Description

As of a week or so ago our conan CI started failing with the following error:
https://github.com/viamrobotics/viam-cpp-sdk/actions/runs/13990621624/job/39173566652
in particular:
```
...
[ 99%] Building CXX object CMakeFiles/libprotoc.dir/upb_generator/mangle.cc.o
[100%] Linking CXX static library libprotoc.a
[100%] Built target libprotoc
[100%] Building CXX object CMakeFiles/protoc.dir/src/google/protobuf/compiler/main.cc.o
[100%] Linking CXX executable protoc
/usr/bin/ld: /github/home/.conan2/p/b/absei022e04c585f4b/p/lib/libabsl_status.a(status.cc.o):(.rodata+0xe0): multiple definition of `absl::lts_20240116::Status::kMovedFromString'; libprotoc.a(command_line_interface.cc.o):(.rodata._ZN4absl12lts_202401166Status16kMovedFromStringE[_ZN4absl12lts_202401166Status16kMovedFromStringE]+0x0): first defined here
```

Namely, building protobuf fails because it is getting duplicate symbols from the upstream abseil dependency. I've reduced it to the following failing conanfile.py:
```python
from conan import ConanFile
from conan.tools.cmake import CMake, CMakeDeps, CMakeToolchain, cmake_layout
from conan.tools.files import load
from conan.tools.apple import is_apple_os
import os
import re

class MweRecipe(ConanFile):
    name = "mwe"

    version = "0.1.0"

    package_type = "library"

    settings = "os", "compiler", "build_type", "arch"

    options = {
        "shared": [True, False]
    }

    default_options = {
        "shared": True
    }

    exports_sources = "CMakeLists.txt", "LICENSE", "src/*"

    def configure(self):
        if self.options.shared:
            for lib in ["grpc", "protobuf"]:
                self.options[lib].shared = True

    def requirements(self):
        self.requires('grpc/[>=1.48.4]')
        self.requires('protobuf/[>=3.17.1]')

    def build_requirements(self):
        self.tool_requires('grpc/[>=1.48.4]')
        self.tool_requires('protobuf/[>=3.17.1]')

    def generate(self):
        tc = CMakeToolchain(self)

        tc.generate()

        CMakeDeps(self).generate()

    def build(self):
        cmake = CMake(self)
        cmake.configure()
        cmake.build()

[...]

Issue #27196, opened 2025-04-09
[package] Protobuf/6.30.1: Crossbuild from Windows to arm with Ninja


alexandrkolesnikov1:
### Description

Hello everyone! I'm trying to make a cross compilation from Windows to Arm and it turned out that the project requires protobuf. However, the protobuf itself does not compile due to the fact that the wrong path is written in the Releas/generator/conan_tolchain.cmake.

Conanfile.py:
```
from conan import ConanFile
from conan.tools.cmake import CMakeToolchain, CMake, cmake_layout

class OrangeTest(ConanFile):
   settings = "os", "compiler", "build_type", "arch"
   generators = "CMakeDeps"

   def requirements(self):
      self.requires("boost/1.86.0")
      self.requires("protobuf/6.30.1")

   def build_requirements(self):
      self.tool_requires ("cmake/3.31.5")

   def generate(self):
      tc = CMakeToolchain(self, generator="Ninja")
      tc.generate()
```

### Package and Environment Details

* Package Name/Version: **protobuf/6.30.1**
* Operating System+version: **Windows 10**
* Compiler+version: **arm-gnu-toolchain-12.2.rel1-mingw-w64-i686-aarch64-none-linux-gnu**
* Conan version: **conan 2.9.2**
* Python version: **Python 3.10.3**

### Conan profile

build profile:
```
[settings] 
arch=x86_64
compiler=msvc
compiler.cppstd=20
compiler.version=194
os=Windows
build_type=Release

[tool_requires]
ninja-xrlab/1.12.1
```

host profile:
```
[settings]
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=14
arch=armv8
os=Linux

[tool_requires]
arm-toolchain/12.2
ninja-xrlab/1.12.1

[conf]
tools.cmake.cmaketoolchain:generator=Ninja
```

### Steps to reproduce

`conan install . --output-folder=build/Debug --build=missing --profile:build=conan_profiles/windows.txt -pr:h conan_profiles/linux.txt -s build_type=Debug`

### Logs

<details><summary>Click to expand log</summary>

```
[...]

PR #27476, opened 2025-05-19
abseil: set compatibility_cppstd to False

jcar87:
### Summary
- Set `compatibility_cppstd` to false to prevent the compatibility plugin from bringing in an abseil binary built with a _different_ `compiler.cppstd` setting than being requested
- Remove all logic pertaining to "abi trick" - no longer needed given the changes of this PR
- Stop publishing new revisions for the older versions of abseil
- Linter fixes

### Background info
- The abseil maintainers **do not** recommend using prebuilt binaries of abseil in a way that could cause mismatches between the `-std=` flag between abseil and consumers. This PR enforces this by preventing the compatibility plugin from mixing this - now it will only bring whatever is requested.
- The `options.h` mechanism https://abseil.io/docs/cpp/guides/options - which causes certain macros to be different and match how abseil was built "so that users are forced to compile their program with the same set of options that were used to build the pre-compiled library" - While the idea behind this is to prevent ABI incompatibilities, it *only* works if the consumers actually evaluated those macros. This is not the case with grpc >=1.70 which requires c++17: it assumes std::variant is available, and compilation spectacularly fails if abseil assumes it is not available - there are compiler errors related to templates (it looks as though some classes expect absl::variant and others std::variant).
- We have seen different issues ehre it is clear we can't mix/match abseil around cppstd - This PR would have mitigated or prevented the following:
    - https://github.com/conan-io/conan-center-index/issues/27196#issuecomment-2792035842
    - https://github.com/conan-io/conan-center-index/issues/26916#issuecomment-2745280052

Close https://github.com/conan-io/conan-center-index/issues/26916


AbrilRBS (dismissed):
So much simpler now too

recipes/abseil/all/conanfile.py
```diff
@@ -84,47 +61,18 @@ def layout(self):
...
         tc = CMakeToolchain(self)
-        tc.variables["ABSL_ENABLE_INSTALL"] = True
-        tc.variables["ABSL_PROPAGATE_CXX_STD"] = True
-        tc.variables["BUILD_TESTING"] = False
-        # We force CMP0067 policy to NEW for our abi trick in _patch_sources()
-        tc.cache_variables["CMAKE_POLICY_DEFAULT_CMP0067"] = "NEW"
-        if is_msvc(self):
-            # see https://github.com/abseil/abseil-cpp/issues/649
-            tc.preprocessor_definitions["_HAS_DEPRECATED_RESULT_OF"] = 1
-            tc.cache_variables["ABSL_MSVC_STATIC_RUNTIME"] = is_msvc_static_runtime(self)
```
franramirez688, line 98: Is this related to that issue mentioned?

franramirez688 (approved):

AbrilRBS (approved):

PR #27478, opened 2025-05-19
LibRaw: Add 0.21.4 and snapshot

irieger:
### Summary
Changes to recipe:  **libraw/0.21.4** and **libraw/0.22.0 snapshot**

#### Motivation
Add the latest official release and a snapshot "release" from february. As this library is adding many things that land in the development branch to support new camera files & features, it makes sense to use the snapshot provided every few months. This latest snapshot contains support for larger files by setting a specific define which is exposed for newer versions. Tested locally not only by building but also by using it and testing with a very large file (the default size supportet is 2 GB, I tested with a 512 GB file).

The local test in my environment was done with this branch: https://github.com/irieger/conan-center-index/commits/libraw/snapshot.20250516/, but the result should be the same.

#### Details
Release notes for 0.21.4: https://github.com/LibRaw/LibRaw/releases/tag/0.21.4

Snapshot from Snapshot-Commit: https://github.com/LibRaw/LibRaw/commit/8afe44cd0e96611ba3cb73779b83ad05e945634c
(LibRaw is only officially tagging a version every ~1.5 to 3 years and published (untagged) snapshots every now and then. See section "Update policy" https://www.libraw.org/)

---

### DIFF
--- a/recipes/libraw/all/CMakeLists.txt
+++ b/recipes/libraw/all/CMakeLists.txt
@@ -59,6 +59,10 @@ if (LIBRAW_WITH_JASPER)
     target_link_libraries(raw PUBLIC Jasper::Jasper)
 endif ()
 
+if (LIBRAW_MAX_CR3_RAW_FILE_SIZE)
+    target_compile_definitions(raw PUBLIC "-DLIBRAW_MAX_CR3_RAW_FILE_SIZE=${LIBRAW_MAX_CR3_RAW_FILE_SIZE}LL")
+endif ()
+
 include(GNUInstallDirs)
 if(LIBRAW_BUILD_THREAD_SAFE)
     add_library(raw_r ${libraw_LIB_SRCS})
--- a/recipes/libraw/all/conandata.yml
+++ b/recipes/libraw/all/conandata.yml
@@ -1,16 +1,13 @@
 sources:
+  "0.21.4":
+    url: "https://github.com/LibRaw/LibRaw/archive/0.21.4.tar.gz"
+    sha256: "8baeb5253c746441fadad62e9c5c43ff4e414e41b0c45d6dcabccb542b2dff4b"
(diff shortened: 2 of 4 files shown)


perseoGI:
Greetings @irieger,

Thank you very much for your contribution and for suggesting the inclusion of the LibRaw snapshot version in Conan Center Index.

After internal discussions, we have carefully evaluated this proposal. While we truly appreciate the effort and understand the motivation behind including newer camera support via snapshot releases, we’ve decided not to accept this version into CCI.

According to the official [LibRaw website](https://www.libraw.org/):

> These versions contain support for new cameras that was added after the previous major release. The API/ABI of public snapshots is not frozen, and may change.

Conan Center Index maintains a strict policy regarding ABI compatibility across package versions. This is crucial to ensure consistent and predictable behavior for consumers who rely on binary compatibility between versions. Snapshot or prerelease versions **specially those that explicitly state their ABI/API is not stabilized** pose a high risk of breaking that guarantee, and therefore cannot be accepted into the central repository.

Additionally, as part of our standard review process, we often consider how other package ecosystems handle similar cases. A quick check on [Repology](https://repology.org/project/libraw/versions) shows that no other major package manager is currently distributing these snapshot versions, which reinforces our conclusion that they are not intended for wide consumption in binary package form.

For users who need access to snapshot versions or want to experiment with pre-release builds, we recommend maintaining a custom fork of Conan Center Index using a [Local Recipe Index](https://docs.conan.io/2/devops/devops_local_recipes_index.html). This mechanism allows full flexibility to package and use versions that fall outside the compatibility policies of the central index.

Once again, thank you for your understanding and for contributing to Conan Center. We hope to see more proposals from you in the future!

Happy coding 🐸

perseoGI (approved):
Thank you very much for your understanding!
LGTM now!

Issue #20552, opened 2023-10-13
[request] imgui_test_engine


yliharsilamikko:
### Package Name/Version

imgui_test_engine

### Webpage

https://github.com/ocornut/imgui_test_engine

### Source code

https://github.com/ocornut/imgui_test_engine

### Description of the library/tool

Dear ImGui Test Engine + Test Suite

Designed to automate and test Dear ImGui applications.
Create screenshots & videos for documentation.

ImGui library has Conan recipe (https://github.com/conan-io/conan-center-index/tree/master/recipes/imgui)
It would be also nice to have recipe for Test Engine.

Related issue in imgui_test_engine repository
https://github.com/ocornut/imgui_test_engine/issues/36

PR #27395, opened 2025-05-11
imgui: Alive Martin Valgur PR

EstebanDugueperoux2:
This PR to alive @valgur https://github.com/conan-io/conan-center-index/pull/25325

PR #27485, opened 2025-05-21
ImGui: enable ImGui TestEngine integration

geertbleyen:
### Summary
Changes to recipe:  imgui/all

#### Motivation
We want to use ImGui's testengine for UI test automation and automatic screenshotting for documentation of our application. For this to work, ImGui must be built with the IMGUI_ENABLE_TEST_ENGINE preprocessor defined. This causes a couple of macros to resolve to extern declared functions, which are provided by the imgui_test_engine repo. Integration of the imgui_test_engine is supposed to be in-source according to the docs (https://github.com/ocornut/imgui_test_engine/wiki/Setting-Up#building)

This change should also solve #20552 

#### Details
Add logic to download the sources of the matching imgui_test_engine repo to the recipe and propagate the needed cmake defines and preprocessor definitions to imgui's cmake.

---


geertbleyen:
@AbrilRBS, @ErniGH any chance either of you can look at his PR (also #27395)? I've seen you being involved in previous imgui recipe updates. I'm holding off my own internal developments for these changes as I'd prefer not to spin my own conan recipes.

franramirez688 (dismissed):
Thanks for this PR @geertbleyen Sorry for the delay on the final reviews 🙏 

This looks great! I have tested it locally and it worked fine 👏

recipes/imgui/all/conanfile.py
```diff
@@ -21,11 +21,20 @@ class IMGUIConan(ConanFile):
...
         "fPIC": True,
+        "enable_test_engine": False
     }
+    
+    def _base_version(self):
+        match = re.match(r'cci\.\d{8}\+(?P<version>\d+\.\d+(?:\.\d+))\.docking', str(self.version)) or re.match(r'(?P<version>\d+\.\d+(.\d+)?)-docking', str(self.version))
+        return Version(match.group('version')) if match else self.version
+    
+    def _is_docking_branch(self):
+        return re.match(r'.+(\.|-)docking', str(self.version))
```
jcar87:
```suggestion
    def _base_version(self):
        return self.version.split('-')[0]
    
    def _is_docking_branch(self):
        return "docking" in self.version
```

If this is the underlying goal, I would go for simplicity

recipes/imgui/all/conanfile.py
```diff
@@ -90,7 +110,8 @@ def package(self):
         cmake.install()
 
     def package_info(self):
-        self.conf_info.define("user.imgui:with_docking", bool(self._match_docking_branch()))
+        self.conf_info.define("user.imgui:with_docking", bool(self._is_docking_branch()))
+        self.conf_info.define("user.imgui:enable_test_engine", bool(self.options.get_safe("enable_test_engine")))
```
jcar87: This is an usual one (ignoring the docking one) - consumers can already check downstream with `self.dependencies['imgui'].options.enable_test_engine`
is there a scenario where one may want to build imgui with `enable_test_engine=True` but _also_ define `user.imgui:enable_test_engine` to False? (as that can be defined outside the recipe, independently)

recipes/imgui/all/conanfile.py
```diff
@@ -43,10 +56,21 @@ def layout(self):
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        base_version = self._base_version()
+        try :
+            get(self, **self.conan_data["sources"][f"{base_version}-testengine"], strip_root=True, destination="test_engine")
+        except KeyError:
+            self.output.warning("No test engine found for this version, skipping download")
```
jcar87: I would probably use the same check as in the `config_options` , e.g
```
if (f"{self._base_version()}-testengine") not in self.conan_data["sources"]:
```

recipes/imgui/all/conanfile.py
geertbleyen: You're right, packages that used the cci versioning style aren't even in conandata.yml anymore. So no point in checking it with a complex regex. Similar for _is_docking_branch, although I would opt to go for .endswith("-docking"), but proposed solution also is fine for me. I'll push the needed changes asap.

recipes/imgui/all/conanfile.py
geertbleyen: I don't think that'd be a realworld scenario, so I'll remove that user.imgui:enable_test_engine define there.
I'll make the adjustments and update the test project's recipe to use the dependency's options.

geertbleyen:
@jcar87 I think I've made the necessary changes based on the PR comments..

franramirez688 (dismissed):
👏

recipes/imgui/all/test_package/conanfile.py
```diff
@@ -18,8 +18,10 @@ def layout(self):
 
     def generate(self):
         with_docking = self.dependencies[self.tested_reference_str].conf_info.get("user.imgui:with_docking", False)
+        with_test_engine = self.dependencies[self.tested_reference_str].options.enable_test_engine
```
franramirez688:
```suggestion
        with_test_engine = self.dependencies[self.tested_reference_str].options.get_safe("enable_test_engine", False)
```

franramirez688 (approved):

ericLemanissier:
@jcar87 @franramirez688 I could be mistaken, but I think there is problem with the way sources are handled. conadata.yml file reduction seems to remove all the `-testengine` keys before uploading to the remote. As a consequence the command `conan install --requires imgui/1.91.8 -o imgui/*:enable_test_engine=True` gets this conandata.yml in cache:
```
sources:
  1.91.8:
    sha256: db3a2e02bfd6c269adf0968950573053d002f40bdfb9ef2e4a90bce804b0f286
    url: https://github.com/ocornut/imgui/archive/v1.91.8.tar.gz
```
And it says `imgui/1.91.8: WARN: No test engine found for this version, removing test engine option`

Shouldn't the test engine sources be added as subkeys to the main version ?

geertbleyen:
> @jcar87 @franramirez688 I could be mistaken, but I think there is problem with the way sources are handled. conadata.yml file reduction seems to remove all the `-testengine` keys before uploading to the remote. As a consequence the command `conan install --requires imgui/1.91.8 -o imgui/*:enable_test_engine=True` gets this conandata.yml in cache:
> 
> ```
> sources:
>   1.91.8:
>     sha256: db3a2e02bfd6c269adf0968950573053d002f40bdfb9ef2e4a90bce804b0f286
>     url: https://github.com/ocornut/imgui/archive/v1.91.8.tar.gz
> ```
> 
> And it says `imgui/1.91.8: WARN: No test engine found for this version, removing test engine option`
> 
> Shouldn't the test engine sources be added as subkeys to the main version ?

Not sure about this or what the default behavior of this should be. We're using Artifactory for hosting conan repos and I don't see this behavior where sources are stripped from the package when being uploaded. My uploaded package contains 1-1 conandata.yml file in artifactory.

franramirez688:
Hi @ericLemanissier @geertbleyen 

It's already fixed by https://github.com/conan-io/conan-center-index/pull/27776

Remember to add `--update` to download the latest revision:

```bash
conan install --requires imgui/1.91.8 -o "imgui/*:enable_test_engine=True" --update --build=missing
```

geertbleyen:
@franramirez688 looks nice, thanks!!

PR #27487, opened 2025-05-21
curlpp: Cmake 4 support

AbrilRBS:
For https://github.com/conan-io/conan-center-index/issues/26878#issuecomment-2897286987

This library is a bit weird, as most of the times it does not call cmake_mi

### DIFF
--- a/recipes/curlpp/all/conanfile.py
+++ b/recipes/curlpp/all/conanfile.py
@@ -7,7 +7,7 @@ from conan.tools.scm import Version
 from conan.tools.microsoft import is_msvc
 import os
 
-required_conan_version = ">=1.53.0"
+required_conan_version = ">=2"
 
 class CurlppConan(ConanFile):
     name = "curlpp"
@@ -27,20 +27,10 @@ class CurlppConan(ConanFile):
         "fPIC": True,
     }
 
-    @property
-    def _min_cppstd(self):
-        return 11
+    implements = ["auto_shared_fpic"]
 
     def export_sources(self):
         export_conandata_patches(self)
-
-    def config_options(self):
-        if self.settings.os == "Windows":
-            del self.options.fPIC
-
-    def configure(self):
-        if self.options.shared:
-            self.options.rm_safe("fPIC")
 
     def layout(self):
         cmake_layout(self, src_folder="src")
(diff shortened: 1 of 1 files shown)


recipes/curlpp/all/conanfile.py
```diff
@@ -50,26 +40,23 @@ def requirements(self):
...
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
+        apply_conandata_patches(self)
 
     def generate(self):
         tc = CMakeToolchain(self)
-        tc.variables["CURLPP_BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["CURLPP_BUILD_SHARED_LIBS"] = self.options.shared
+        tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5"  # CMake 4 support
```
perseoGI, line 54: Maybe we should add this check, equivalent to other CMake 4 PRs, as in the future this library could add support, being this `cache_variable` definition incorrect. 
(If added, don't forget importing proper `ConanException`)

```suggestion
        tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5"  # CMake 4 support
        if Version(self.version) > "0.8.1.cci.20240530": # pylint: disable=conan-unreachable-upper-version
             raise ConanException("CMAKE_POLICY_VERSION_MINIMUM hardcoded to 3.5, check if new version supports CMake 4")

```

recipes/curlpp/all/conanfile.py
```diff
@@ -50,26 +40,23 @@ def requirements(self):
...
-        if self.settings.compiler.get_safe("cppstd"):
-            check_min_cppstd(self, self._min_cppstd)
-        else:
-            if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "6":
-                raise ConanInvalidConfiguration("${self.ref} requires C++11. Please set 'compiler.cppstd=11'.")
-            if self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "14":
-                raise ConanInvalidConfiguration("${self.ref} requires C++11. Please set 'compiler.cppstd=11'.")
+        check_min_cppstd(self, 11)
+        if self.settings.compiler == "gcc" and Version(self.settings.compiler.version) < "6":
+            raise ConanInvalidConfiguration("${self.ref} requires C++11. Please set 'compiler.cppstd=11'.")
```
jcar87, line 45: the message in this error does not match the condition - this will raise on gcc <6 regardless of `compiler.cppstd`

perseoGI (approved):

PR #27495, opened 2025-05-22
vulkan-hpp: version 1.4.313 (a new recipe) (outdated macOS SDK)

gav2xlin:
### Summary
Add a new recipe:  **vulkan-hpp/[1.4.313]**

#### Motivation

Add a new recipe for [https://github.com/gav2xlin/Vulkan-Hpp/](https://github.com/gav2xlin/Vulkan-Hpp/)

#### Details

recipes/vulkan-hpp/all/conandata.yml
recipes/vulkan-hpp/all/conanfile.py
recipes/vulkan-hpp/all/test_package/conanfile.py
recipes/vulkan-hpp/config.yml

---

conan create . --version 1.4.313

This pull request depends on
vulkan-headers [https://github.com/conan-io/conan-center-index/pull/27455](https://github.com/conan-io/conan-center-index/pull/27455)
glslang [https://github.com/conan-io/conan-center-index/pull/27462](https://github.com/conan-io/conan-center-index/pull/27462)

### DIFF
--- a/recipes/vulkan-hpp/all/conandata.yml
+++ b/recipes/vulkan-hpp/all/conandata.yml
@@ -1,0 +1,8 @@
+sources:
+  "1.4.313.0":
+    source:
+      url: "https://github.com/KhronosGroup/Vulkan-Hpp/archive/refs/tags/v1.4.313.tar.gz"
+      sha256: "66777b4397a46e6d1fe8beb79a127d6101277be56f7d2aa183df9f3d0edbc3cc"
+    tinyxml2:
+      url: "https://github.com/leethomason/tinyxml2/archive/refs/tags/11.0.0.tar.gz"
+      sha256: "5556deb5081fb246ee92afae73efd943c889cef0cafea92b0b82422d6a18f289"
--- a/recipes/vulkan-hpp/all/conanfile.py
+++ b/recipes/vulkan-hpp/all/conanfile.py
@@ -1,0 +1,66 @@
+from conan import ConanFile, Version
+from conan.errors import ConanInvalidConfiguration
+from conan.tools.cmake import cmake_layout, CMakeDeps, CMakeToolchain, CMake
+from conan.tools.files import copy, get
+import os
+
+required_conan_version = ">=2"
+
+
+class VulkanHppConan(ConanFile):
+    name = "vulkan-hpp"
+    description = "Open-Source Vulkan C++ API"
+    license = "Apache-2.0"
+    topics = ("vulkan-hpp", "vulkan", "gpu")
+    homepage = "https://github.com/KhronosGroup/Vulkan-Headers"
+    url = "https://github.com/KhronosGroup/Vulkan-Hpp/"
+    package_type = "application"
+    settings = "os", "arch", "compiler", "build_type"
+    no_copy_source = True
+
+    def layout(self):
+        cmake_layout(self, src_folder="src")
+
+    def requirements(self):
+        self.requires(f"vulkan-headers/{self.version}", transitive_headers=True)
+
+    def validate(self):
+        if self.settings.compiler == "msvc" and Version(self.settings.compiler.version) < "193":
+            raise ConanInvalidConfiguration("Visual Studio 2022 (MSVC 1930 or newer) is required by qt >= 6.8.3")
+
+        if self.settings.compiler == "apple-clang" and Version(self.settings.compiler.version) < "14":
+            raise ConanInvalidConfiguration("apple-clang >= 14 is required by qt >= 6.8.3")
+
+    def source(self):
+        data = self.conan_data["sources"][self.version]
+        get(self, **data["source"], strip_root=True)
+        get(self, **data["tinyxml2"], strip_root=True, destination="tinyxml2")
+
+    def generate(self):
+        tc = CMakeToolchain(self)
+        #tc.variables["VULKAN_HPP_GENERATOR_BUILD"] = False
+        tc.generate()
+        deps = CMakeDeps(self)
+        deps.generate()
+
+    def build(self):
+        cmake = CMake(self)
+        cmake.configure()
+        cmake.build()
+
+    def package(self):
+        cmake = CMake(self)
+        cmake.install()
+        copy(self, "LICENSE.txt", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
+
+        bin_path = os.path.join(self.package_folder, "bin")
+        copy(self, "VideoHppGenerator*", src=self.cpp.build.bindir, dst=bin_path)
+        copy(self, "VulkanHppGenerator*", src=self.cpp.build.bindir, dst=bin_path)
+
+    def package_info(self):
+        bin_path = os.path.join(self.package_folder, "bin")
+        self.output.info(f"Appending PATH environment variable: {bin_path}")
+        self.env_info.path.append(bin_path)
+
+        self.cpp_info.includedirs = []
+        self.cpp_info.libdirs = []
--- a/recipes/vulkan-hpp/all/test_package/conanfile.py
+++ b/recipes/vulkan-hpp/all/test_package/conanfile.py
@@ -1,0 +1,15 @@
+from conan import ConanFile
+from conan.tools.build import can_run
+
+
+class TestPackageConan(ConanFile):
+    settings = "os", "arch", "compiler", "build_type"
+    generators = "VirtualBuildEnv"
+    test_type = "explicit"
+
+    def build_requirements(self):
+        self.tool_requires(self.tested_reference_str)
+
+    def test(self):
+        if can_run(self):
+            self.run("VulkanHppGenerator")
--- a/recipes/vulkan-hpp/config.yml
+++ b/recipes/vulkan-hpp/config.yml
@@ -1,0 +1,3 @@
+versions:
+  "1.4.313.0":
+    folder: all


recipes/vulkan-hpp/all/conandata.yml
```diff
@@ -0,0 +1,20 @@
+sources:
+  "1.4.313":
+    source:
+      url: "https://github.com/KhronosGroup/Vulkan-Hpp/archive/refs/tags/v1.4.313.tar.gz"
+      sha256: "66777b4397a46e6d1fe8beb79a127d6101277be56f7d2aa183df9f3d0edbc3cc"
+    glslang:
+      url: "https://github.com/KhronosGroup/glslang/archive/refs/tags/vulkan-sdk-1.4.313.0.tar.gz"
+      sha256: "555ac780ccceca926fa25775834639ce6ffc744120bfb68fb8657dd4032d21ee"
```
planetmarshall: #27278 updates glslang

recipes/vulkan-hpp/all/conandata.yml
planetmarshall, line 1: I think you'll want to add dependencies on the respective conan packages rather than downloading the sources directly, eg
```.py
def requirements(self):
  self.requires("glm/1.0.1")
  self.requires("tinyxml2/11.0.0")
  ... etc
```

recipes/vulkan-hpp/all/conandata.yml
```diff
@@ -0,0 +1,20 @@
...
+    glfw:
+      url: "https://github.com/glfw/glfw/releases/download/3.4/glfw-3.4.zip"
+      sha256: "b5ec004b2712fd08e8861dc271428f048775200a2df719ccf575143ba749a3e9"
```
planetmarshall: I think glfw is only required to build the samples

recipes/vulkan-hpp/all/conanfile.py
```diff
@@ -0,0 +1,140 @@
...
+        "no_setters": [True, False],
+        "no_smart_handle": [True, False],
+        "no_spaceship_operator": [True, False],
+        "no_to_string": [True, False],
+        "no_win32_prototypes": [True, False],
+        "raii_no_exceptions": [True, False],
+        "smart_handle_implicit_cast": [True, False],
+        "typesafe_conversion": [True, False],
+        "use_reflect": [True, False],
+        "install": [True, False]
```
planetmarshall: I think a lot of these options are only relevant if you're building the samples and tests - otherwise it's a header only library - in which case the recipe can probably be a lot simpler.

recipes/vulkan-hpp/all/conandata.yml
gav2xlin: It will not work because tinyxml2, glfw, glm, glslang and even vulkan-headers are submodules of vulkan-hpp 🤷 

[https://github.com/KhronosGroup/Vulkan-Hpp](https://github.com/KhronosGroup/Vulkan-Hpp)

Vulkan-Headers @ 75ad707
Update Vulkan-Headers to v1.4.315 (#2161)

glfw @ 8e6c8d7
Update submodules glfw and glm (#1772)

glm @ b06b775
Update submodules glfw and glm (#1772)

glslang @ fc9889c
Update submodule glslang to version 15.3.0 (#2156)

tinyxml2 @ ff61650
Update tinyxml2

[https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/CMakeLists.txt](https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/CMakeLists.txt)

Only: find_package( Vulkan )

I found that some Conan packages install dependencies the same way in this case

recipes/vulkan-hpp/all/conandata.yml
gav2xlin: removed

[https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/CMakeLists.txt](https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/CMakeLists.txt)

if( VULKAN_HPP_SAMPLES_BUILD OR VULKAN_HPP_TESTS_BUILD )
	# external libraries
	add_subdirectory( glm )
	add_subdirectory( glfw )
	add_subdirectory( glslang )
	...
endif()

recipes/vulkan-hpp/all/conanfile.py
gav2xlin: I kept several options

recipes/vulkan-hpp/all/conandata.yml
```diff
@@ -0,0 +1,11 @@
+sources:
+  "1.4.313":
+    source:
+      url: "https://github.com/KhronosGroup/Vulkan-Hpp/archive/refs/tags/v1.4.313.tar.gz"
+      sha256: "66777b4397a46e6d1fe8beb79a127d6101277be56f7d2aa183df9f3d0edbc3cc"
+    tinyxml2:
+      url: "https://github.com/leethomason/tinyxml2/archive/refs/tags/11.0.0.tar.gz"
+      sha256: "5556deb5081fb246ee92afae73efd943c889cef0cafea92b0b82422d6a18f289"
+    vulkan-headers:
```
planetmarshall: use `self.requires(f"vulkan-headers/{self.version}.0"` (or something similar) in `def requirements(self)`

recipes/vulkan-hpp/all/conandata.yml
```diff
@@ -0,0 +1,11 @@
+sources:
+  "1.4.313":
+    source:
+      url: "https://github.com/KhronosGroup/Vulkan-Hpp/archive/refs/tags/v1.4.313.tar.gz"
+      sha256: "66777b4397a46e6d1fe8beb79a127d6101277be56f7d2aa183df9f3d0edbc3cc"
+    tinyxml2:
```
planetmarshall, line 6: You don't need `tinyxml2` unless you're going to build the generator

recipes/vulkan-hpp/all/conandata.yml
gav2xlin: Vulkan-Headers [https://github.com/KhronosGroup/Vulkan-Hpp](https://github.com/KhronosGroup/Vulkan-Hpp)
This is a submodule of Vulkan-Hpp, but not an external library 🤔

....
vulkan-hpp/1.4.313 (test package): Running test()
vulkan-hpp/1.4.313 (test package): RUN: **VulkanHppGenerator**
clang-format version 19.1.5
VulkanHppGenerator: Found VulkanHppGenerator: Loading /home/user/.conan2/p/vulka12dc84093f168/s/src/Vulkan-Headers/registry/vk.xml
VulkanHppGenerator: failed to load file /home/agavrilin/.conan2/p/vulka12dc84093f168/s/src/Vulkan-Headers/registry/vk.xml with error <XML_ERROR_FILE_NOT_FOUND>

ERROR: vulkan-hpp/1.4.313 (test package): Error in test() method, line 15
        self.run("VulkanHppGenerator")
        ConanException: Error 255 while executing

recipes/vulkan-hpp/all/conandata.yml
gav2xlin: tinyxml2 is a submodule of Vulkan-Hpp

I have no idea why the authors decided to link the library this way🤷

...
CMake Error at CMakeLists.txt:113 (add_executable):
  Cannot find source file:

    /home/user/.conan2/p/vulkaca4289208f6ce/s/src/tinyxml2/tinyxml2.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm
  .ccm .cxxm .c++m .h .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90
  .f95 .f03 .hip .ispc


CMake Error at CMakeLists.txt:113 (add_executable):
  No SOURCES given to target: VulkanHppGenerator


CMake Error at CMakeLists.txt:123 (add_executable):
  No SOURCES given to target: VideoHppGenerator

gav2xlin: [https://github.com/conan-io/conan-center-index/tree/master/recipes/sbp](https://github.com/conan-io/conan-center-index/tree/master/recipes/sbp)

sources:
  "5.0.4":
    cmake:
      url: "https://github.com/swift-nav/cmake/archive/954e6cebe4c902d07d33beef166df89073a7b4f2.zip"
      sha256: "910e14e35d02fd766bc885cfc9b99626006c12ef26d6446cfbf728862d7cebab"
    source:
      url: "https://github.com/swift-nav/libsbp/archive/refs/tags/v5.0.4.tar.gz"
      sha256: "0b51707780dec08f24fac091b935e041d0d61abf2093562df829e85b9510d72b"

    def source(self):
        data = self.conan_data["sources"][self.version]
        get(self, **data["source"], strip_root=True)
        get(self, **data["cmake"], strip_root=True, destination=os.path.join("c", "cmake", "common"))

I used the same approach

planetmarshall: Yes, but in the case of Swift those are just extra cmake files - it's not a dependency that is already available in the conan index, or one that other libraries in the dependency graph could also be depending on.

I'm just a voluntary contributor and have no say over what gets approved but I think the recipe is unlikely to be approved if you consume the dependencies this way.

> This is a submodule of Vulkan-Hpp, but not an external library

It's the same thing - The Vulkan-Hpp maintainers have just decided to make their dependencies available as submodules to make it easier to consume the project. Conan recipes achieve the same thing by using `requires`.

See [paho-mqtt-cpp](https://github.com/conan-io/conan-center-index/blob/master/recipes/paho-mqtt-cpp/all/conanfile.py) as an example of another recipe where a dependency on a submodule (the paho-mqtt-c library) is provided via `requires`

gav2xlin: [https://github.com/conan-io/conan-center-index/blob/master/recipes/paho-mqtt-cpp/all/conandata.yml](https://github.com/conan-io/conan-center-index/blob/master/recipes/paho-mqtt-cpp/all/conandata.yml)

[https://github.com/conan-io/conan-center-index/blob/master/recipes/paho-mqtt-cpp/all/conanfile.py](https://github.com/conan-io/conan-center-index/blob/master/recipes/paho-mqtt-cpp/all/conanfile.py)

def requirements(self):
        # Headers are exposed https://github.com/conan-io/conan-center-index/pull/16760#issuecomment-1502420549
        # Symbols are exposed   "_MQTTProperties_free", referenced from: mqtt::connect_options::~connect_options() in test_package.cpp.o
        self.requires("paho-mqtt-c/1.3.13", transitive_headers=True, transitive_libs=True)

It works because [paho.mqtt.cpp](https://github.com/eclipse-paho/paho.mqtt.cpp/) supports packages

[https://github.com/eclipse-paho/paho.mqtt.cpp/blob/master/CMakeLists.txt](https://github.com/eclipse-paho/paho.mqtt.cpp/blob/master/CMakeLists.txt)

# PAHO_WITH_MQTT_C is disabled by default

```
option(PAHO_WITH_MQTT_C "Build Paho C from the internal GIT submodule." FALSE)

if(PAHO_WITH_MQTT_C)
    message(STATUS "Paho C: Bundled")

    ...
else()
    **find_package(eclipse-paho-mqtt-c REQUIRED)**
endif()
```

vulkan-hpp does not find CMake packages at all (except Vulkan)

[https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/CMakeLists.txt](https://github.com/KhronosGroup/Vulkan-Hpp/blob/main/CMakeLists.txt)

find_package( Vulkan REQUIRED ) (conditionally)

gav2xlin: if( DEFINED VULKAN_HPP_VULKAN_HEADERS_SRC_DIR )
if( NOT DEFINED VULKAN_HPP_TINYXML2_SRC_DIR )

Is there a way to set these variables from conanfile.py ?

def requirements(self):
   self.requires("vulkan-headers/...")
   self.requires("tinyxml2/...")

and what further 🤔

planetmarshall: > vulkan-hpp does not find CMake packages at all (except Vulkan)

The conan `vulkan-headers` package creates a target called `Vulkan::Headers` which will be included by the `vulkan-hpp` cmake configuration when it calls `find_package(Vulkan)`. So you don't need to (and shouldn't) download the vulkan headers explicitly - just use `self.requires('vulkan-headers...')`

For `tinyxml` you will probably need to patch `CMakeLists.txt` to use `find_package`. I think it's unlikely that the recipe will be approved if you try to include the sources directly - see [handling-internal-dependencies](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#handling-internal-dependencies) - but the maintainers will have the final word on this.

Including the dependency on `tinyxml` should probably also be optional as it is not required if `vulkan-hpp` is consumed as a header-only library (ie you are not building the generator).

planetmarshall: You might also consider making the generator build `False` by default, as this is how `Vulkan-Hpp` is configured when it's consumed as a subproject (which I think is typically how most devs would be using the conan package).

```.cmake
option( VULKAN_HPP_GENERATOR_BUILD "Build the HPP generator" ${PROJECT_IS_TOP_LEVEL} )
```

Probably one for the maintainers to weigh in on, though.

PR #27504, opened 2025-05-23
[nanobind] New recipe

cqc-alec:
Add new recipe:  **nanobind/2.7.0**

---

### DIFF
--- a/recipes/nanobind/all/conandata.yml
+++ b/recipes/nanobind/all/conandata.yml
@@ -1,0 +1,4 @@
+sources:
+  "2.7.0":
+    url: "https://github.com/wjakob/nanobind/archive/refs/tags/v2.7.0.tar.gz"
+    sha256: "6c8c6bf0435b9d8da9312801686affcf34b6dbba142db60feec8d8e220830499"
--- a/recipes/nanobind/all/conanfile.py
+++ b/recipes/nanobind/all/conanfile.py
@@ -1,0 +1,80 @@
+import os
+
+from conan import ConanFile
+from conan.tools.build import check_min_cppstd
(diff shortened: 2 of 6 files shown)


cqc-alec:
It seems that the requirement for `cpython` breaks this. When run locally, and also on github actions after running `actions/setup_python`, this recipe works for me _without_ adding that requirement, but on the CI here the cmake build of the test fails at `find_package(Python 3.8 REQUIRED COMPONENTS Development)`. I am not sure how to fix this.

perseoGI:
Greetings @cqc-alec,

Thank you for your contribution! 🚀 

After discussing your pull request (PR) with @AbrilRBS, I made some slight modifications to your test package. I removed the actual execution of the Python module import. It's important that `test_package`s in the Conan Center Index remain as simple as possible and can run on any platform. Relying on the results of the `test_package` to invoke Python (which may not be in every user's PATH) is risky and not recommended.

I tested your package locally, and I'm happy to report that it works:
```python
>>> import test_module
>>> print(test_module.add(1, 2))
3
```

I also pushed some changes to the `test()` function. These changes simply check whether the shared object is being created, rather than attempting to execute it. This execution should be handled by the upstream CI tests.

Thank you again for your contributions!

Happy coding 🐸

perseoGI:
Okey, there is another issue in the compilation, we will check with the team.
Stay tuned!

Todiq:
Thanks to https://github.com/conan-io/conan/issues/18222, we might be able to do:
```python
def package_info(self):
    self.cpp_info.set_property("cmake_additional_dependencies", ["Python3"])
```

Which would be way easier than having to depend on `cpython`. What do you reckon @perseoGI?

PR #18009, opened 2023-06-21
Allow to set ENABLE_HARU and WT_RASTERIMAGE_IMPLEMENTATION

ViktarHasiul231862:
Right now original Wt recipe doesn't allow to set two CMake flags, that I need to set in order to properly build Wt for my project.
WT_RASTERIMAGE_IMPLEMENTATION can not be set at all and ENABLE_HARU is always false: 
tc.variables["ENABLE_HARU"] = False

PR #27509, opened 2025-05-23
allow to set pango

ViktarHasiul231862:
allow to set pango flag. Similar to my previous pull request: https://github.com/conan-io/conan-center-index/pull/18009

### DIFF
--- a/recipes/wt/all/conanfile.py
+++ b/recipes/wt/all/conanfile.py
@@ -33,6 +33,7 @@ class WtConan(ConanFile):
         "with_opengl": [True, False],
         "with_unwind": [True, False],
         "with_haru": [True, False],
+        "with_pango": [True, False],
         "raster_image": ["none", "Direct2D", "GraphicsMagick"],
         "no_std_locale": [True, False],
         "no_std_wstring": [True, False],
@@ -54,6 +55,7 @@ class WtConan(ConanFile):
         "with_opengl": False,
         "with_unwind": True,
         "with_haru": False,
+        "with_pango": False,
         "raster_image": "none",
         "no_std_locale": False,
         "no_std_wstring": False,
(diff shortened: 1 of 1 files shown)


ViktarHasiul231862:
bump

uilianries:
@ViktarHasiul231862 Hello! As you may have noted, Wt has several options already, which makes it hard to cover all combinations, and possibly reach bugged scenarios and switching a few of them. Based on that, we avoid adding new options unless they are seen as really useful or are asked for by multiple users. 

In case of adding the pango option, the CI will not validate when having it as True. As a result, if there is a bug, we will only know much later, as we would need to validate each combination locally first.

Still, you can use that CMake definition with your build without touching the recipe. You can use the Conan configuration `tools.cmake.cmaketoolchain:extra_variables`. You will need to add it to your [Conan profile](https://docs.conan.io/2/reference/config_files/profiles.html#conf):

```ini
[conf]
wt/*:tools.cmake.cmaketoolchain:extra_variables={"ENABLE_PANGO": "ON"}
wt/*:tools.info.package_id:confs=["tools.cmake.cmaketoolchain:extra_variables"]
```

Using this configuration, Conan will inject `ENABLE_PANGO=ON` in the generator used to build Wt only. The second configuration `tools.info.package_id:confs` ensures the package ID produced by this new configuration will be affected by the values present in `extra_variables`. So it will not collide when building a regular package. 

And you can use the very same configuration on your command line in case you don't want to update your Conan profile:

```
conan install -r conancenter --requires=wt/4.11.2 --build=wt -c tools.cmake.cmaketoolchain:extra_variables='{"ENABLE_PANGO": "ON"}' -c tools.info.package_id:confs='["tools.cmake.cmaketoolchain:extra_variables"]'
```

Regards.

ViktarHasiul231862:
@uilianries Got it, thank you for detailed explanation.

Issue #27497, opened 2025-05-22
[question] How to fix cross-compilation issue for drogon - can't find openssl dep target (it used to work)


lkotsonis:
### What is your question?

Hi, I am trying to understand why a cross-compilation for `drogon` is failing. 

I'm using a yocto cross-compilation SDK to cross compile from x64 to arm (which I have used with drogon before, among many other libraries).

Now there is an issue with finding the `openssl::openssl` target that is declared in `trantor`. 

I am providing the log from running the conan install command inside the docker image that wraps the yocto SDK (you can also see the profile reported by conan in the beginning).

When I build with `-pr:b x64 -pr:h x64` it builds fine.

Any help is appreciated.

<details open>
<summary>Log</summary>
mycompany@66f3a255257c:/tmp/crater$ conan install --requires=drogon/1.9.10 --build missing -pr:b x64 -pr:h arm.jinja -s:a compiler.cppstd=20 -o:a "drogon/*:with_boost=False" -r=conancenter

======== Input profiles ========
Profile host:
[settings]
arch=armv7hf
build_type=Release
compiler=gcc
compiler.cppstd=20
compiler.libcxx=libstdc++11
compiler.version=14
os=Linux
[options]
drogon/*:with_boost=False
[conf]
tools.build:cflags=['-mthumb', '-mfpu=neon', '-mfloat-abi=hard', '-mcpu=cortex-a9']
tools.build:compiler_executables={'asm': 'arm-mycompany-linux-gnueabi-as', 'c': 'arm-mycompany-linux-gnueabi-gcc', 'cpp': 'arm-mycompany-linux-gnueabi-g++'}
tools.build:cxxflags=['-mthumb', '-mfpu=neon', '-mfloat-abi=hard', '-mcpu=cortex-a9']
tools.build:exelinkflags=['-mthumb', '-mfpu=neon', '-mfloat-abi=hard', '-mcpu=cortex-a9']
tools.build:sharedlinkflags=['-mthumb', '-mfpu=neon', '-mfloat-abi=hard', '-mcpu=cortex-a9']
tools.build:sysroot=/usr/local/oecore-x86_64/sysroots/cortexa9t2hf-neon-mycompany-linux-gnueabi
tools.gnu:make_program=make
[buildenv]
CHOST=arm-mycompany-linux-gnueabi
[...]

PR #27510, opened 2025-05-23
trantor: use requires explicitly to avoid issues with openssl

lkotsonis:
### Summary
Changes to recipe:  **trantor/[version]**

#### Motivation
In some cases (e.g. cross-compilation), if openSSL is not present, the cmake target is incorrectly communicated as having the alias openssl::openssl. This change makes sure that the correct openSSL names are propagated to the consumers. I had already posted an issue #27497 .

#### Details
When cross-compiling as a consumer of the library there can be issues with openSSL being correctly detected through its targets. When I tried cross compiling drogon that depends on trantor, I got an error on CMake which basically said that `openssl::openssl` is not a declared target.

When not cross-compiling that worked fine. That's most likely because my native (x64) configuration had openSSL installed while the cross-compiled host config did not. 

The change explicitly sets the names of the dependencies so that the conan-generated `openssl::openssl` is avoided. This of course means that all the requirements must be explicitly set in `package_info()`.

---

### DIFF
--- a/recipes/trantor/all/conanfile.py
+++ b/recipes/trantor/all/conanfile.py
@@ -126,6 +126,13 @@ class TrantorConan(ConanFile):
         self.cpp_info.set_property("cmake_target_name", "Trantor::Trantor")
         self.cpp_info.libs = ["trantor"]
 
+        self.cpp_info.requires = ["openssl::ssl", "openssl::crypto"]
+
+        if self.options.with_c_ares:
+            self.cpp_info.requires.append("c-ares::cares")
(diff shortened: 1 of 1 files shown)


jcar87:
Closing this as the underlying cause is probably what we describe in: https://github.com/conan-io/conan-center-index/issues/27497#issuecomment-3627316572

Merging this PR *might* result in a successful build if CMake finds `OpenSSL` from the sysroot and not from Conan, so it's more of a workaround than an actual solution - the actual solution is to use OpenSSL from Conan, otherwise you may run into linker/runtime errors.

Issue #27505, opened 2025-05-23
[package] libpq/16.8: problem with headers


pgeler:
### Description

in generated package PostgreSQL-relwithdebinfo-x86_64-data.cmake file  libpq_INCLUDE_DIRS_RELWITHDEBINFO is not populated as a result, find_package/target_link_libraries could not locate headers. However in conan_toolchain.cmake at CMAKE_INCLUDE_PATH folder points to the right location.

so far can reproduce on both MSVC and Linux builds.

### Package and Environment Details

* Package Name/Version: **libpq/16.8**
* Operating System+version: **Windows/Linux Ubuntu 24.04**
* Compiler+version: **Clang 18/MSVC**
* Docker image: **conanio/gcc8**
* Conan version: **conan 2.16.1**
* Python version: **Python 3.9.x**

### Conan profile

[settings]
arch=x86_64
os=Windows
compiler=msvc
compiler.version=194
compiler.runtime=dynamic
compiler.cppstd=20
build_type=Debug
[conf]
tools.cmake.cmaketoolchain:generator=Ninja
tools.build:compiler_executables={"c":"C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe","cpp":"C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe","rc":"C:/Program Files (x86)/Windows Kits/10/bin/10.0.26100.0/x64/rc.exe"}

### Steps to reproduce

cmake-conan...
cmake -G Ninja ${SOURCE_DIR}
cmake --build .

### Logs

<details><summary>Click to expand log</summary>

```
Put your log output here
```

</details>

pgeler:
> clarify
Yes, will close it after PR is merged, was able to consume it as is, by supplying `requires("libpq/16.8", force=True)`

uilianries:
@pgeler Thank you for your answer! Please, avoid using `force`, it will override versions and may cause api/abi compatibility errors. I wrote some better possibilities in that PR.

PR #27513, opened 2025-05-23
add plibpq transitive_headers into soci build

pgeler:
### Summary
Changes to recipe:  **soci/4.0.3**

#### Motivation
CMake cache does not provide INCLUDE_DIRS, while libpq is used in soci headers

#### Details

this is primarily to address #27505, however still not clear how to consume recipe without this if dependent library required in consuming project.

---

### DIFF
--- a/recipes/soci/all/conanfile.py
+++ b/recipes/soci/all/conanfile.py
@@ -68,7 +68,7 @@ class SociConan(ConanFile):
         if self.options.with_mysql:
             self.requires("libmysqlclient/8.1.0")
         if self.options.with_postgresql:
-            self.requires("libpq/15.4")
+            self.requires("libpq/15.4", transitive_headers=True)
         if self.options.with_boost:
             self.requires("boost/1.83.0")
 


uilianries:
So far, so good in Mac when building with all options equal True: [soci-4.0.3-macos-clang17-opt-all.log](https://github.com/user-attachments/files/20454859/soci-4.0.3-macos-clang17-opt-all.log) 🎉

uilianries (approved):
LGTM, the version 4.0.3 has external dependencies listed in public headers. It changed in 4.1.z, so it should be changed when bumping the recipe version. 

Thank you @pgeler @AbrilRBS for fixing this recipe! 😄

AbrilRBS (approved):

recipes/soci/all/conanfile.py
```diff
@@ -24,10 +24,7 @@ class SociConan(ConanFile):
         "fPIC":             [True, False],
         "empty":            [True, False],
         "with_sqlite3":     [True, False],
-        "with_db2":         [True, False],
```
jcar87, line 27: @AbrilRBS  @uilianries  - I can't find anywhere in the descriptions or the comments as to why these lines, which are unrelated to the PR goal/description, are here? Maybe I'm missing something

pgeler: @jcar87 IMHO the point was that those flags not implemented anyway, I can add `and cleanup` to make this PR to move

jcar87: Makes sense - this was confusing when reviewing as was not mentioned anywhere

rjahanbakhshi:
Before this, sqlite3 backend was compiling fine but now I get the error: fatal error: sqlite3.h: No such file or directory.
Additionally, the change is done on an existing recipe version which is breaking already working projects.

uilianries:
@rjahanbakhshi Sorry to hear about your error after this update. 

The PRs in ConanCenterIndex are well reviewed and tested by maintainers, but unfortunately, they're not infallible. To avoid breaking your local copies of recipes, there are documented ways to protect your environment from breaking changes:  https://github.com/conan-io/conan-center-index/blob/master/docs/consuming_recipes.md

Also, it would be great to understand your scenario and what change made it happen for your environment, so we can work on a fix. Could you please open an [issue](https://github.com/conan-io/conan-center-index/issues/) with all the asked details there? As the CI did not fail for the regular build, and locally I can build with sqlite3 without errors, we will need more information to reproduce your case. 

Here is my build log using the latest Soci revision with sqlite3 and working: [soci-4.0.3-macos-clang17-sqlite3.log](https://github.com/user-attachments/files/20585769/soci-4.0.3-macos-clang17-sqlite3.log)

I'll be waiting for your report, regards!

jcar87:
Thanks for reporting @rjahanbakhshi 

Fix proposed here: https://github.com/conan-io/conan-center-index/pull/27599

We strongly encourage the use of lockfiles to ensure reproducibility and guard yourself from these sort of (hopefully infrequent) regressions: https://docs.conan.io/2/devops/using_conancenter.html#repeatability-and-reproducibility

recipes/soci/all/conanfile.py
```diff
@@ -37,106 +34,69 @@ class SociConan(ConanFile):
...
+        tc.cache_variables["WITH_BOOST"] = self.options.with_boost
         if Version(self.version) < "4.1.0": # pylint: disable=conan-condition-evals-to-constant
             tc.cache_variables["CMAKE_POLICY_VERSION_MINIMUM"] = "3.5" # CMake 4 support
         tc.generate()
 
         deps = CMakeDeps(self)
         deps.set_property("mysql", "cmake_file_name", "MYSQL")
         deps.set_property("libpq", "cmake_file_name", "POSTGRESQL")
-        deps.set_property("sqlite3", "cmake_file_name", "SQLITE3")
+        deps.set_property("sqlite3", "cmake_file_name", "SQLite3")
```
jcar87, line 96: @AbrilRBS

rjahanbakhshi:
@uilianries, thanks for the quick response and helpful advice.
@jcar87, really appreciate the fast patch — all our pipeline jobs are green again.
Also, thanks for sharing the reproducibility guides. This was actually the first time we encountered this issue in the past year or two. I’ll make sure to apply the guidelines going forward to avoid any potential regressions.

PR #27517, opened 2025-05-24
gsl-lite: Modernize recipe, add versions 0.43.0 and 1.0.1

### LINKED ISSUE #5521: [package] gsl-lite/0.38.1: Contract violation usability issue

mbeutel:
### Summary
Changes to recipe:  **gsl-lite/0.43.0**, **gsl-lite/1.0.1**

#### Motivation
This PR modernizes the recipe and adds new package versions 0.43.0 and 1.0.1.
The configuration option `on_contract_violation` is removed. Contract violation behavior should be defined through a build flag, not at package manager level.

Closes #5521. Renders #25157 obsolete.

#### Details
See release notes for *gsl-lite* [v0.43.0](https://github.com/gsl-lite/gsl-lite/releases/tag/v0.43.0), [v1.0.0](https://github.com/gsl-lite/gsl-lite/releases/tag/v1.0.0) and [v1.0.1](https://github.com/gsl-lite/gsl-lite/releases/tag/v1.0.1).

---

### DIFF
--- a/recipes/gsl-lite/1.x/conanfile.py
+++ b/recipes/gsl-lite/1.x/conanfile.py
@@ -22,9 +22,9 @@ class GslLiteConan(ConanFile):
 
     def package_id(self):
         self.info.clear()
-    
+
     def build_requirements(self):
-        self.tool_requires("cmake/[>=3.16 <4]")
+        self.tool_requires("cmake/[>=3.16]")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)


AbrilRBS (approved):
Hi @mbeutel thanks a lot for taking the time nto add these new versions and cleaning the recipe up

I've made some more changes and simplifications, and I'm approving the new 0.43 patch too as this is the last 0.x version, as the new major version has some breaking changes and users might want to update first to 0.43 before moving to the 1.x major

AbrilRBS (changes requested):
More changes needed, sorry for the noise

recipes/gsl-lite/all/conanfile.py
```diff
@@ -40,34 +40,42 @@ def _contract_map(self):
...
+        cmake_layout(self, src_folder="src")
 
     def package_id(self):
         self.info.clear()
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
 
+    def generate(self):
+        tc = CMakeToolchain(self)
```
AbrilRBS, line 57: Checking if this generates the same files our copy did

mbeutel:
> > The configuration option on_contract_violation is removed. Contract violation behavior should be defined through a build flag, not at package manager level.
> 
> While I might agree that for header-only libraries define flags are a bit harder to properly manage, I think we need to keep the option for the old releases, for which users might expect it to exist and removing it would be a breaking change for them

Fair enough. Removing the option for v1.0 and newer only seems reasonable.

recipes/gsl-lite/all/conanfile.py
mbeutel: Not quite sure what the question is, sorry.

The old recipe just copied some files from the source directory. The updated recipe now just uses the CMake install process, which I think makes more sense.

AbrilRBS: Sorry, not a question but a note for future reviewers that this needs checking, sorry about the noise :)

AbrilRBS (approved):

mbeutel:
Hi @AbrilRBS, could you please rerun the CI for the package? I hope my update from last week fixes the CMake version requirement issue.

recipes/gsl-lite/1.x/conanfile.py
```diff
@@ -16,13 +16,15 @@ class GslLiteConan(ConanFile):
...
-    tool_requires = "cmake/[>=3.16]"
 
     def layout(self):
         cmake_layout(self, src_folder="src")
 
     def package_id(self):
         self.info.clear()
+    
+    def build_requirements(self):
+        self.tool_requires("cmake/[>=3.16 <4]")
```
mbeutel: Out of curiosity: why require "<4"? The package works fine with CMake v4.

mbeutel:
Is there anything I can do to help getting this PR merged?

recipes/gsl-lite/1.x/conanfile.py
franramirez688: Indeed, let's remove that:

```suggestion
        self.tool_requires("cmake/[>=3.16]")
```

recipes/gsl-lite/config.yml
```diff
@@ -1,15 +1,19 @@
...
+    folder: old
   "0.38.1":
-    folder: all
+    folder: old
   "0.38.0":
-    folder: all
+    folder: old
   "0.37.0":
-    folder: all
+    folder: old
```
franramirez688, line 19: We could also remove all those versions (`0.42.0` to `0.37.0`) from here to stop publishing new revisions for them.

recipes/gsl-lite/old/conandata.yml
```diff
@@ -1,4 +1,7 @@
 sources:
+  "0.43.0":
+    url: "https://github.com/gsl-lite/gsl-lite/archive/v0.43.0.tar.gz"
+    sha256: "e48c3138648156d2b85905b1d280d661fad61524c5c0ca10d3857036ca3dd519"
```
franramirez688, line 4: Related to my previous comment, we could remove the `0.42.0` to `0.37.0` from this

franramirez688 (commented):
Hi @mbeutel

Thanks for this PR, and sorry for the delay 🙏 

LGTM! It'd be great if you could solve the conflict with `master` branch. Also, a couple of minor suggestions 😁

recipes/gsl-lite/config.yml
franramirez688: UPDATED: Sorry, I just saw that `0.40.0` and `0.41.0` are used by other recipes. Let's clean this up in another PR to avoid the noise

AbrilRBS (approved):

mbeutel:
Excellent, thanks for merging.

PR #27519, opened 2025-05-24
b2: Fix compiler errors in the Intel Linux toolset

MarcinOstrowskiB3D:
### Summary
Changes to recipe:  **b2/5.3.2**

#### Motivation
When building Boost with the Intel LLVM compiler (icpx) on Ubuntu via Conan, the default b2 recipe (version 5.3.2) emits unsupported PCH flags (-pch-create) and mis-invokes the frontend, leading to errors like:

```
icpx: error: unsupported argument 'h-create' to option '-pc'
icpx: error: no such file or directory: '…/intl-lnx-2025./rls/…'
```

This contradicts Intel’s own “Building Boost with oneAPI” guidelines (https://www.intel.com/content/www/us/en/developer/articles/technical/building-boost-with-oneapi.html) and prevents successful boost compiling.

#### Details
- Conan recipe: in recipes/b2/portable/conanfile.py, call apply_conandata_patches(self) after get(...) so our patch is applied.
- Patch (patches/5.3.2-intel-linux-pch-fix.patch): in src/tools/intel-linux.jam
    - Replace -pch-create with Clang-style -Xclang -emit-pch -o for both C and C++ headers.

---

### DIFF
--- a/recipes/b2/portable/conandata.yml
+++ b/recipes/b2/portable/conandata.yml
@@ -32,3 +32,8 @@ sources:
   "5.3.2":
     url: "https://github.com/bfgroup/b2/releases/download/5.3.2/b2-5.3.2.tar.bz2"
     sha256: "881827649356dfa59caa30e169931424993c72a1626a502abb7ea474484eac26"
+patches:  
+  "5.3.2":
+    - patch_file: "patches/5.3.2-intel-linux-pch-fix.patch"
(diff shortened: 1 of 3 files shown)


jcar87:
Hi @MarcinOstrowskiB3D - please let us know if there is an upstream issue that we should be checking. For patches like this, it's best if they've been pre-vetted by the maintainers of the relevant component

PR #27529, opened 2025-05-26
added enable_builtin option for libpsl versions >= 0.21.5

laci-aura:
### Summary
Changes to recipe:  **libpsl/0.21.5**

#### Motivation
Version 0.21.5 introduces enable_builtin / disable_builtin options but the current recipe only supports the deprecated behaviour which cannot enable built-in PSL data when the "with_dna" option is not set:
https://github.com/rockdaboot/libpsl/blob/490bd6f98a2addcade55028ea60c36cce07e21e4/configure.ac#L174

#### Details
I've added direct control over the enable_builtin / disable_builtin options as supported by libpsl/0.21.5

---

### DIFF
--- a/recipes/libpsl/all/conanfile.py
+++ b/recipes/libpsl/all/conanfile.py
@@ -24,11 +24,13 @@ class LibPslConan(ConanFile):
         "shared": [True, False],
         "fPIC": [True, False],
         "with_idna": [False, "icu", "libidn", "libidn2"],
+        "enable_builtin": [True, False],
     }
     default_options = {
         "shared": False,
         "fPIC": True,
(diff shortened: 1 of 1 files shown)


uilianries:
@laci-aura Hello! Thank you for your submission and intention to improve this recipe. 

I see the builtin option was actually changed on version [0.21.2](https://github.com/rockdaboot/libpsl/commit/aa4909766c24c17bd1d9000ca419f6dc6b32c238). I still think your point is valid because it was previously based on idna option value. 

I would like to understand first if you need builtin enabled? If it's the case, I would be more inclined to not add that new option, but keep the default behavior that's enabled.

PR #26107, opened 2024-12-04
abseil: Fix build on macOS when using LLVM clang

AbrilRBS:
### Summary
Changes to recipe:  **abseil/[*]**

#### Motivation

There is a single file in the abseil build that is built with both x86_64 and armv8. When using LLVM clang (this is *not* an issue on Apple-Clang), some versions fail with this error:

```
clang++: error: unsupported option '-msse4.1' for target 'arm64-apple-darwin23.6.0'
make[2]: *** [absl/random/CMakeFiles/random_internal_randen_hwaes_impl.dir/internal/randen_hwaes.cc.o] Error 1
make[1]: *** [absl/random/CMakeFiles/random_internal_randen_hwaes_impl.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```

#### Details
Backport the upstream fix - this is only a fix in the build system to work as intended.

Close https://github.com/conan-io/conan-center-index/issues/25992
Close https://github.com/conan-io/conan-center-index/issues/25747

PR #27531, opened 2025-05-26
fix: make re2 use abseil version 20250127.0

Taepper:
### Summary
Changes to recipe:  **re2/***

#### Motivation

On recent Mac OS versions, Abseil did not compile anymore. See also these issues:
https://github.com/GenSpectrum/LAPIS-SILO/issues/774
https://github.com/zeldamods/oead/pull/41

However, as re2 fixes the abseil version, this could not be fixed without changing the recipe

#### Details

This updates the Abseil version which is required by the re2 package.

---

### DIFF
--- a/recipes/re2/all/conanfile.py
+++ b/recipes/re2/all/conanfile.py
@@ -62,7 +62,7 @@ class Re2Conan(ConanFile):
         if self.options.get_safe("with_icu"):
             self.requires("icu/73.2")
         if Version(self.version) >= "20230601":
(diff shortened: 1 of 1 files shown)


jcar87:
Hi @Taepper - thanks for this PR.

since https://github.com/conan-io/conan-center-index/pull/26107, our abseil recipes should be patched to avoid the error you are describing on macOS. Can you confirm that you are using an updated version of the recipe (passing `--update` to your Conan command may do the trick) - and if it is still failing, could you provide us with error logs so that we can have a look? thanks!

PR #27561, opened 2025-05-29
openmvg: add missing transitivity for requirements

maksim-petukhou:
### Summary
Changes to recipe:  **openmvg/\***

#### Motivation and Details

Problem: can't compile code if you do `#include <openmvg/graph/graph.hpp>` and link **openmvg** targets in CMake, because it also requires **coin-lemon** headers which are not available.

PR adds headers and libs transitivity to **coin-lemon** and **ceres-solver** (also missing).
**ceres-solver** headers are available through `openmvg/sfm/sfm_data_BA_ceres_camera_functor.hpp` public header
and **coin-lemon** headers are available through `openmvg/graph/graph.hpp` and some other public headers from `openmvg/graph`.
As both libraries are not header only, we have to add libs transitivity too.

---

### DIFF
--- a/recipes/openmvg/all/conanfile.py
+++ b/recipes/openmvg/all/conanfile.py
@@ -61,7 +61,7 @@ class Openmvgconan(ConanFile):
 
     def requirements(self):
         self.requires("cereal/1.3.2", transitive_headers=True)
-        self.requires("ceres-solver/2.1.0", transitive_headers=True, transitive_libs=True)
+        self.requires("ceres-solver/2.1.0")
         self.requires("coin-clp/1.17.7")
         self.requires("coin-lemon/1.3.1", transitive_headers=True, transitive_libs=True)
         self.requires("coin-osi/0.108.7")


SpaceIm:
Ok for coin-lemon, indeed it's a public dependency.
Regarding ceres-solver, I don't think so. It's marked as private in their CMakeLists. It's indeed included by `openmvg/sfm/sfm_data_BA_ceres_camera_functor.hpp` but this header is actually private I think (even though it's installed by their CMakeLists, which is doing a brutal GLOB *.h *.hpp). This header is not included by `openmvg/sfm/sfm.hpp`, only by a .cpp file (`openmvg/sfm/sfm_data_BA_ceres.cpp`).

maksim-petukhou:
@SpaceIm I think you're right. I've removed transitivity for ceres-solver

maksim-petukhou:
@SpaceIm Please let me know if anything else is required from my side

PR #27583, opened 2025-06-02
Nvtx: fix recipe and add v3.3.0

phwissmann:
### Summary
Changes to recipe:  **nvtx/***

#### Motivation
Add missing packaging of header "nvtx3/nvtx3.hpp" which is the main cpp entrypoint as described in 
https://nvidia.github.io/NVTX/doxygen-cpp/index.html

Add version 3.3.0
Fix export targets

---

### DIFF
--- a/recipes/nvtx/all/conandata.yml
+++ b/recipes/nvtx/all/conandata.yml
@@ -2,12 +2,6 @@ sources:
   "3.3.0":
     url: "https://github.com/NVIDIA/NVTX/archive/v3.3.0.tar.gz"
     sha256: "67d0cda2f9d19a89684592dab40c0bf2c2b13d5d588e51392076c0890a64b6c0"
-  "3.2.2":
-    url: "https://github.com/NVIDIA/NVTX/archive/v3.2.2.tar.gz"
-    sha256: "76b875507d2501c9431d79e272ae5b6d4131bd0ffa7a7d3b79ac261d35609f7c"
-  "3.2.1":
-    url: "https://github.com/NVIDIA/NVTX/archive/v3.2.1.tar.gz"
-    sha256: "737c3035f0e43a2252e7cd94c3f26e11e169f624236efe31794f044ce44a70af"
   "3.1.0":
     url: "https://github.com/NVIDIA/NVTX/archive/v3.1.0.tar.gz"
     sha256: "dc4e4a227d04d3da46ad920dfee5f7599ac8d6b2ee1809c9067110fb1cc71ced"
--- a/recipes/nvtx/config.yml
+++ b/recipes/nvtx/config.yml
@@ -1,10 +1,6 @@
 versions:
   "3.3.0":
     folder: all
-  "3.2.2":
-    folder: all
-  "3.2.1":
-    folder: all
   "3.1.0":
     folder: all
   "3.0.1":


SoShiny:
In case you want to add a test for the `.hpp` header being copied:

```diff
diff --git a/recipes/nvtx/all/test_package/CMakeLists.txt b/recipes/nvtx/all/test_package/CMakeLists.txt
index f9d181212..85a03bfa3 100644
--- a/recipes/nvtx/all/test_package/CMakeLists.txt
+++ b/recipes/nvtx/all/test_package/CMakeLists.txt
@@ -1,8 +1,11 @@
 cmake_minimum_required(VERSION 3.15)
-project(test_package LANGUAGES C)
+project(test_package LANGUAGES C CXX)

 find_package(nvtx REQUIRED CONFIG)

 add_executable(${PROJECT_NAME} test_package.c)
 target_link_libraries(${PROJECT_NAME} PRIVATE nvtx::nvtx)
 target_compile_features(${PROJECT_NAME} PRIVATE c_std_99)
+
+add_executable(${PROJECT_NAME}_cxx test_package.cpp)
+target_link_libraries(${PROJECT_NAME}_cxx PRIVATE nvtx::nvtx)
diff --git a/recipes/nvtx/all/test_package/conanfile.py b/recipes/nvtx/all/test_package/conanfile.py
index 0a6bc6871..c9d8782b7 100644
--- a/recipes/nvtx/all/test_package/conanfile.py
+++ b/recipes/nvtx/all/test_package/conanfile.py
@@ -24,3 +24,5 @@ class TestPackageConan(ConanFile):
         if can_run(self):
             bin_path = os.path.join(self.cpp.build.bindirs[0], "test_package")
             self.run(bin_path, env="conanrun")
+            bin_path = os.path.join(self.cpp.build.bindirs[0], "test_package_cxx")
+            self.run(bin_path, env="conanrun")
diff --git a/recipes/nvtx/all/test_package/test_package.cpp b/recipes/nvtx/all/test_package/test_package.cpp
new file mode 100644
index 000000000..bdfbec516
--- /dev/null
+++ b/recipes/nvtx/all/test_package/test_package.cpp
@@ -0,0 +1,10 @@
+#include <stdlib.h>
+#include <nvtx3/nvtx3.hpp>
+
+int main()
+{
+    nvtxInitialize(NULL);
+    nvtxRangeId_t main_range = nvtxRangeStartA("main");
+    nvtxRangeEnd(main_range);
+    return EXIT_SUCCESS;
+}
```

phwissmann:
Added. Makes sense. Thanks.

phwissmann:
@conan-io/barbarians could you approve the CI run of this PR?

recipes/nvtx/all/conanfile.py
```diff
@@ -35,8 +35,12 @@ def build(self):
     def package(self):
         copy(self, "LICENSE.txt", src=self.source_folder, dst=os.path.join(self.package_folder, "licenses"))
         copy(self, "*.h", src=os.path.join(self.source_folder, "c", "include"), dst=os.path.join(self.package_folder, "include"))
+        copy(self, "*.hpp", src=os.path.join(self.source_folder, "c", "include"), dst=os.path.join(self.package_folder, "include"))
 
     def package_info(self):
+        self.cpp_info.set_property("cmake_file_name", "nvtx3")
+        self.cpp_info.set_property("cmake_target_name", "nvtx3::nvtx3-c")
+        self.cpp_info.set_property("cmake_target_aliases", ["nvtx3::nvtx3-cpp", "CUDA::nvtx3"])
```
AbrilRBS: The issue with these changes it that they would break existing consumers for the old versions too, just as you needed to update the existing test_package CMakeLists.

While I see that the file names are in fact wrong, I think we need to keep the actual `nvtx` one until at least the next major release, where breaking changes like this can be more acceptable.

What we can fix for now is the missing target names, by adding aliases to all of the missing ones, so keeping this only for now would make sense (as then current users keep their existing targets working too)


```suggestion
        self.cpp_info.set_property("cmake_target_aliases", ["nvtx3::nvtx3-cpp", "nvtx3::nvtx3-c"])
```

Note that I have not added the CUDA one, because I can't see where upstream is defining such a target, so I'd be happy to get some pointers there, thanks :)

phwissmann: The cuda one I just copy&pasted from here after being point to it: https://github.com/conan-kiln/kiln/blob/main/recipes/nvtx/all/conanfile.py#L37-L39

Ok, so I added the `nvtx` file and target name for backwards compatibility

AbrilRBS (approved):

franramirez688 (approved):

PR #27606, opened 2025-06-04
rocksdb: add folly support, add version 10.2.1

RazielXYZ:
### Summary
Changes to recipe:  **rocksdb**

#### Motivation
Support for folly and coroutines can further speed up rocksdb, and we like speed.

For version ranges:

- Using version ranges for common dependencies (snappy, lz4, zlib, zstd) allows us to avoid issues when using rocksdb in combination with one of the many other packages that also depend on any of those.
- Using a version range for folly allows us to update folly later (as the only version on CC is getting relatively old now) without having to update this again; I imagine we would want to switch many of folly's dependencies to version ranges as well.
- gflags has not gotten a new release in a long time, so I did not think it necessary to switch to a version range.
- jemalloc has been recently archived for unknown reasons, and the latest release is also a long time ago, so ditto.
- TBB I didn't test so I dared not touch it, but I would assume newer versions would work just fine, assuming the current integration works at all.

The two patches:

- fix-folly fixes rocksdb's cmakelists check to see if it found folly to use the conan-provided folly_LIBRARIES.
- fix-msvc-coroutines avoids adding gcc flags to msvc which would make it error out, as well as avoids overriding the CMAKE_CXX_STANDARD provided by conan

#### Details
Added support and options for coroutines and folly, and patches as needed 
Condition some older options that were removed to not apply to newer versions
Switch some dependencies to version ranges
Add the latest release

---

### DIFF
--- a/recipes/rocksdb/all/conanfile.py
+++ b/recipes/rocksdb/all/conanfile.py
@@ -35,6 +35,8 @@ class RocksDBConan(ConanFile):
         "with_jemalloc": [True, False],
         "enable_sse": [False, "sse42", "avx2"],
         "use_rtti": [True, False],
+        "use_coroutines": [True, False],
+        "use_folly": [True, False],
     }
     default_options = {
         "shared": False,
@@ -49,11 +51,13 @@ class RocksDBConan(ConanFile):
         "with_jemalloc": False,
         "enable_sse": False,
         "use_rtti": False,
+        "use_coroutines": False,
+        "use_folly": False,
     }
 
     @property
     def _min_cppstd(self):
-        return "11" if Version(self.version) < "8.8.1" else "17"
+        return "11" if Version(self.version) < "8.8.1" else "20" if self.options.use_coroutines else "17"
 
     def export_sources(self):
         export_conandata_patches(self)
@@ -69,6 +73,8 @@ class RocksDBConan(ConanFile):
     def configure(self):
         if self.options.shared:
             self.options.rm_safe("fPIC")
+        if Version(self.version) >= "8.0.0":
+            self.options.rm_safe("lite") # RocksDB Lite was removed in 8.0.0
 
     def layout(self):
         cmake_layout(self, src_folder="src")
@@ -77,17 +83,19 @@ class RocksDBConan(ConanFile):
         if self.options.with_gflags:
             self.requires("gflags/2.2.2")
         if self.options.with_snappy:
-            self.requires("snappy/1.1.10")
+            self.requires("snappy/[>=1.1.10 <2]")
         if self.options.with_lz4:
-            self.requires("lz4/1.9.4")
+            self.requires("lz4/[>=1.10.0 <2]")
         if self.options.with_zlib:
             self.requires("zlib/[>=1.2.11 <2]")
         if self.options.with_zstd:
-            self.requires("zstd/1.5.5")
+            self.requires("zstd/[>=1.5.5 <2]")
         if self.options.get_safe("with_tbb"):
             self.requires("onetbb/2021.10.0")
         if self.options.with_jemalloc:
             self.requires("jemalloc/5.3.0")
+        if self.options.use_coroutines or self.options.use_folly:
+            self.requires("folly/[>=2024.08.12.00]")
 
     def validate(self):
         check_min_cppstd(self, self._min_cppstd)
@@ -97,6 +105,13 @@ class RocksDBConan(ConanFile):
 
         if is_msvc(self) and Version(self.settings.compiler.version) < "191":
             raise ConanInvalidConfiguration("Rocksdb requires MSVC version >= 191")
+        
+        folly_options = [bool(self.options.use_coroutines), bool(self.options.use_folly)]
+        if sum(folly_options) > 1:
+            raise ConanInvalidConfiguration("Rocksdb requires at most one of use_coroutines or use_folly")
+        
+        if self.options.shared and (self.options.use_coroutines or self.options.use_folly):
+            raise ConanInvalidConfiguration("Rocksdb does not support a shared build with folly or coroutines")
 
     def source(self):
         get(self, **self.conan_data["sources"][self.version], strip_root=True)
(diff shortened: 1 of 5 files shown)


perseoGI:
Greetings @RazielXYZ,

Thanks a lot for taking the time to improve this recipe 🙌.

I tried pushing some changes to your branch, but it looks like maintainers don’t have permission to update it. Could you please enable that option so we can help streamline the process?

Regarding the dependencies: according to our [documentation](https://github.com/conan-io/conan-center-index/blob/master/docs/adding_packages/dependencies.md#version-ranges), only a specific set of libraries should use version ranges. At the moment, snappy, lz4, and folly are not part of that list.

For us to allow a dependency as a version range, we need to be confident that the upstream project follows strict semantic versioning, ideally with an explicit statement in their documentation. Without that guarantee, we need to pin the dependency to an exact version instead.

Thanks again for your contribution!

RazielXYZ:
@perseoGI thanks for looking this over! I was not aware there was a specific list of libraries that are allowed as version ranges in requirements.
As for permission to update the branch, as far as I can see here it is allowed. Not sure where/how I would check if there's an issue with it.

uilianries:
Hi! I made a few changes under this PR, including:

- Dropped coroutines option: That option is not exposed by other package managers and requires C++20 at least, and still requires an extra patch to work on Windows. In case needing it, I would recommend using the Conan configuration `tools.cmake.cmaketoolchain:extra_variables` instead. For the Windows part, I would recommend discussing with the upstream first to have a proper fix.
- Renamed Folly option to `with_folly` to keep aligned with others
- Folly is only available for newer versions; otherwise, it will consume a vendorized code
- The option Lite is no longer available for newer versions
- Adjusted ranged versions to what is actually used in CCI:
  - Folly always breaks when we need to release a new version in CCI, so the version range will be a problem
  - lz4 in most cases is used 1.9.4, but folly uses 1.10.0. The upstream documents the compatibility, so it seems okay to keep the version range.
  - Snappy has a similar case: most packages are using 1.1.10, but Folly requires 1.2.1.
- Dropped both new patches:
  - It's possible to manipulate CMake variables for dependencies without patching
  - Fix for an upstream bug related to coroutines on Windows - Discuss with the author first
  
----

As this new option is `False` by default, I built locally on Linux and Windows using True:
- Windows MSVC 194: [rocksdb-10.5.1-windows-msvc194-folly.log](https://github.com/user-attachments/files/21917560/rocksdb-10.5.1-windows-msvc194-folly.log)
- Linux GCC 11: [rocksdb-10.5.1-linux-gcc11-folly.log](https://github.com/user-attachments/files/21917558/rocksdb-10.5.1-linux-gcc11-folly.log)

RazielXYZ:
@uilianries that looks good. I'll open an issue or PR upstream when able related to the fixes to their CMake file for windows/coroutines

uilianries:
On Mac it's not possible to build using Folly due to the following bug: https://github.com/facebook/rocksdb/issues/13895

Added that rule in the validate()

perseoGI (approved):
LGTM thank you @RazielXYZ and @uilianries

recipes/rocksdb/all/conanfile.py
```diff
@@ -97,6 +116,14 @@ def validate(self):
...
         if is_msvc(self) and Version(self.settings.compiler.version) < "191":
             raise ConanInvalidConfiguration("Rocksdb requires MSVC version >= 191")
+        
+        if self.options.shared and self.options.get_safe("with_folly"):
+            # https://github.com/facebook/rocksdb/blob/v10.5.1/CMakeLists.txt#L603
+            raise ConanInvalidConfiguration(f"{self.ref} does not support a shared build with folly")
+
+        if self.settings.compiler in ["apple-clang", "clang"] and self.options.get_safe("with_folly"):
+            # https://github.com/facebook/rocksdb/issues/13895
+            raise ConanInvalidConfiguration(f"{self.ref} does not support folly with {self.settings.compiler}.")
```
franramirez688: If that's the problem, why not replace that flag?

```suggestion
```
And let's add this piece of code:

```python
    def build(self):
        apply_conandata_patches(self)
        if self.settings.compiler in ["apple-clang", "clang"] and self.options.get_safe("with_folly"):
            replace_in_file(self, os.path.join(self.source_folder, "CMakeLists.txt"), "-Wl,--copy-dt-needed-entries", "-Wl")
```
This worked fine for me. Wdyt?

franramirez688 (commented):
@RazielXYZ thanks a lot for this PR 🙏 

LGTM! Only a small comment about folly + clang before approving/merging this

recipes/rocksdb/all/conanfile.py
franramirez688: cc @uilianries

RazielXYZ: I would prefer this over not having clang at all as well. It doesn't seem like facebook is very responsive to issues, I have no idea who we'd have to poke for upstream fixes to actually happen.

uilianries: Agreed, the `--copy-dt-needed-entries` is only for `ld.bfd`. I suggested something different in the issue, but ignoring it sounds good to me as well. Let's keep Clang supported, then.

uilianries: Done, changes made on commit a318fab

uilianries:
Validated using the patch provided by the commit [a318fab](https://github.com/conan-io/conan-center-index/pull/27606/commits/a318fabbcc9cd346fc0ff3c7b84a53b63e056f18):

[rocksdb-10.5.1-apple-clang17-patched.log](https://github.com/user-attachments/files/22404485/rocksdb-10.5.1-apple-clang17-patched.log)

The patch requires version 7.7.2 or later: https://github.com/facebook/rocksdb/commit/be09943fb58a2dd3f70e6e30781ebfa3fcbcb8fa

Without the patch, the error persists on Mac: [rocksdb-10.5.1-apple-clang17-no-patched.log](https://github.com/user-attachments/files/22404495/rocksdb-10.5.1-apple-clang17-no-patched.log)

recipes/rocksdb/all/conanfile.py
```diff
@@ -143,10 +172,20 @@ def generate(self):
...
         if self.options.with_zstd:
             deps.set_property("zstd", "cmake_target_name", "zstd::zstd")
+        if self.options.get_safe("with_folly"):
+            deps.set_property("folly", "cmake_additional_variables_prefixes", ["FOLLY",])
         deps.generate()
 
+    def _patch_sources(self):
+        # INFO: --copy-dt-needed-entries is only needed for ld.bfd and breaks other linkers like ld.gold and lld
+        # https://github.com/facebook/rocksdb/issues/13895
+        if Version(self.version) >= "7.7.2":
```
franramirez688:
```suggestion
        if self._has_folly:
```

PR #27608, opened 2025-06-05
wolfssl: add version 5.8.0 and option enable_asio

wshon:
### Summary
Changes to recipe:  **wolfssl**

#### Details

1. Add new version 5.7.4 5.7.6 5.8.0
2. Add new option with_asio to --enable-asio, see https://www.wolfssl.com/wolfssl-support-asio-boost-asio-c-libraries/

---

### DIFF
--- a/recipes/wolfssl/all/conandata.yml
+++ b/recipes/wolfssl/all/conandata.yml
@@ -1,4 +1,13 @@
 sources:
+  "5.8.0":
+    url: "https://github.com/wolfSSL/wolfssl/archive/v5.8.0-stable.tar.gz"
+    sha256: "f90f18c7f12913a0b351b1f4305e768697ea1380794df1f2984b9452ab0aeeaf"
+  "5.7.6":
+    url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.6-stable.tar.gz"
+    sha256: "52b1e439e30d1ed8162a16308a8525a862183b67aa30373b11166ecbab000d63"
+  "5.7.4":
+    url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.4-stable.tar.gz"
+    sha256: "6c70f6fab34e2f963a255c3637c0c6f7250df2d1ef7c34eebca4f8e8d7323399"
   "5.7.2":
     url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.2-stable.tar.gz"
     sha256: "0f2ed82e345b833242705bbc4b08a2a2037a33f7bf9c610efae6464f6b10e305"
--- a/recipes/wolfssl/all/conandata.yml
+++ b/recipes/wolfssl/all/conandata.yml
@@ -2,12 +2,6 @@ sources:
   "5.8.0":
     url: "https://github.com/wolfSSL/wolfssl/archive/v5.8.0-stable.tar.gz"
     sha256: "f90f18c7f12913a0b351b1f4305e768697ea1380794df1f2984b9452ab0aeeaf"
-  "5.7.6":
-    url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.6-stable.tar.gz"
-    sha256: "52b1e439e30d1ed8162a16308a8525a862183b67aa30373b11166ecbab000d63"
-  "5.7.4":
-    url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.4-stable.tar.gz"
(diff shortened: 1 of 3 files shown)


jcar87:
Hi @wshon  - thanks for your contribution. Please amend the PR to only add the most recent version, unless there's strong reasons where users may not be able to use the latest.

Also I think that the asio option that's beeind added may need a dependency on boost or asio as well

moskupols:
Hi @jcar87 , I'm also interested in this PR as I'd like to try wolfssl 5.8.0.
As far as I understand from build script of wolfssl, --enable-asio merely enables a set of feature flags required for asio to rely on wolfssl. wolfssl built with --enable-asio does not require user to link asio.
Asio on the other hand has to be explicitly configured to depend on wolfssl.

recipes/wolfssl/all/conandata.yml
```diff
@@ -1,4 +1,13 @@
 sources:
+  "5.8.0":
+    url: "https://github.com/wolfSSL/wolfssl/archive/v5.8.0-stable.tar.gz"
+    sha256: "f90f18c7f12913a0b351b1f4305e768697ea1380794df1f2984b9452ab0aeeaf"
+  "5.7.6":
+    url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.6-stable.tar.gz"
+    sha256: "52b1e439e30d1ed8162a16308a8525a862183b67aa30373b11166ecbab000d63"
+  "5.7.4":
+    url: "https://github.com/wolfSSL/wolfssl/archive/v5.7.4-stable.tar.gz"
+    sha256: "6c70f6fab34e2f963a255c3637c0c6f7250df2d1ef7c34eebca4f8e8d7323399"
```
jcar87: Please only add a single version - unless there's a strong reason why a user on an already published version (e.g. 5.7.2) may not be able to move to `5.8.0` (e.g. major API breakages, known issues, LTS and so on).

recipes/wolfssl/all/conanfile.py
```diff
@@ -63,6 +64,7 @@ class WolfSSLConan(ConanFile):
         "with_quic": False,
         "with_experimental": False,
         "with_rpk": False,
+        "with_asio": False,
```
jcar87: I would suggest setting this to `enable_asio` as it is called upstream. `with_asio` was confusing initially because for us recipe maintainers it would imply this library is built "with" asio (that is, using asio in its implementation), which appears to not be the case - this option in the upstream buildscripts only enables opensslall/opensslextra and defines a couple of macros in the options header

jcar87 (changes requested):
just a couple minor changes!

jcar87:
> Hi @jcar87 , I'm also interested in this PR as I'd like to try wolfssl 5.8.0. As far as I understand from build script of wolfssl, --enable-asio merely enables a set of feature flags required for asio to rely on wolfssl. wolfssl built with --enable-asio does not require user to link asio. Asio on the other hand has to be explicitly configured to depend on wolfssl.

Thanks for providing additional details @moskupols  - I was initially confused because of the name of the option `with_asio`, with typically implies the implementation would rely on asio which needs to be discoverable.

As far as I can see, `--enable-asio` simply implies enabling these two options to `True`:
```
        "opensslextra": [True, False],
        "opensslall": [True, False],
```
as well presumably defining `ASIO_USE_WOLFSSL` in the wolfssl options.h header. Technically it's already usable if those two options are enabled and one defines `ASIO_USE_WOLFSSL` before including the asio headers (which is also documented as a valid option).

However, happy to keep the option, but would suggest renaming it to `enable_asio` instead (makes it a bit less confusing than `with_asio`).  Whilst technically we could add additional logic to reflect the opensslextra/opensslall being turned to "True", I dont think it's worth it - so just a name change would do.

recipes/wolfssl/all/conanfile.py
wshon: I completely agree. When I was writing the name, I also hesitated whether to use "with" or "enable". In the end, I decided to keep it consistent with the previous part. Of course, aligning it with the upstream content is also a good idea.

recipes/wolfssl/all/conandata.yml
wshon: There's nothing special here. I just included all the versions I came across, though it might be better to only include the latest one. The version I'm using is also 5.8.0.

jcar87:
Thanks @wshon and sorry for the wait. We have published the most recent version and added the option as `asio` in this PR: https://github.com/conan-io/conan-center-index/pull/30240

PR #27612, opened 2025-06-05
Add glm options

proceduralnoisy:
### Summary
Changes to recipe:  **glm/***

#### Motivation
Enable GLM consumers to customize GLM preprocessor options via Conan options. I exposed most of the options available (as documented in the [GLM manual](https://github.com/g-truc/glm/blob/master/manual.md))

#### Details
Should be a fairly straightforward review. I tested most of these new options manually (by enabling the new `force_messages` option to print the GLM configuration options, turning assorted options on, and verifying they appear in the configuration messages)

---

### DIFF
--- a/recipes/glm/all/conanfile.py
+++ b/recipes/glm/all/conanfile.py
@@ -1,4 +1,5 @@
 from conan import ConanFile
+from conan.errors import ConanInvalidConfiguration
 from conan.tools.files import copy, get
 from conan.tools.layout import basic_layout
 from conan.tools.scm import Version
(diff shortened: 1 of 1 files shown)


jcar87:
Thanks @proceduralnoisy for your PR. 

However, I'm afraid this is not a change we would merge for a header-only library.
The manual you have linked at https://github.com/g-truc/glm/blob/master/manual.md, specifies usage like this:

```
#define GLM_FORCE_MESSAGES // Or defined when building (e.g. -DGLM_FORCE_SWIZZLE)
#include <glm/glm.hpp>
```

If the expectation is for C++ consumers to define these macros ahead of including the headers - then that's how it should be handled, and not via the recipe.

A recipe for a header only library (as is the case here), only has a single package ID - even when it has options. We've had issues in the past with "convenience" options that pass those macros via `-D` defines depending on recipe options - some users require _different_ options for different translation units, and find that the Conan recipe will unconditionally define based on the options - so they lose flexibility that they would otherwise have.

We would only consider adding options to a header only library when the upstream build scripts generate _different_ configuration files (e.g `-config.cmake` files) depending on options specified at "build" or package time. That doesn't seem to be the case here: https://github.com/g-truc/glm/blob/master/CMakeLists.txt - thus, it looks like these macros should be defined by users in the consuming code, rather than the recipe.

proceduralnoisy:
Thanks @jcar87 for the consideration and detailed reply. I understand the philosophy, and will close the PR shortly.

If you have some extra time, it would be helpful if you have an alternative recommendation for achieving our goals for options like these (if you'd prefer, I can start a new question issue instead). 

Two use cases jump to mind:
(for context, we're using Conan in a corporate environment that will be juggling hundreds of in-house and external Conan packages)

**Use case 1: Intermediate utility package**
Let's say we have Conan package `CalcUtils` that uses GLM to perform computations for consumers. It relies on certain GLM preprocessor flags. Some computations are implemented in headers, so these flags must be set for downstream consumers as well.

We could require `CalcUtils` developers to explicitly set GLM flags in every header, but that feels a bit error-prone. An omitted flag is bound to make it through code review eventually, and the effects of these flags can be subtle and disastrous (e.g. `GLM_FORCE_EXPLICIT_CTOR` controls whether to default-initialize objects).

Conan options appeared to be a good solution here - `CalcUtils` could specify the GLM options and check them in `validate`. I suppose we could get a similar effect by setting the GLM flags explicitly in `package_info` (e.g. `self.cpp_info.defines.append("GLM_FORCE_CTOR_INIT")`. Or maybe by enforcing the use of common Conan profiles that include `tools.build:defines` statements.

**Use case 2: Global option enforcement**
Some library settings may not be permitted within an organization for security or legal reasons. If these settings are exposed as Conan options, they can be checked/validated via Conan APIs. 

Thanks!

jcar87:
Thanks for your message @proceduralnoisy.

I'm not familiar with the options in glm  - however, as I've mentioned before, enforcing them as options in the Conan package _removes_ the ability of consumers to use different settings in different translation units, as the library authors intend - we've had issues in this area in the past. So what may be suitable for a set of users like you, may not be suitable for everyone.

In situations like this, when in doubt, we'll ask ourselves the question: what is the documented use of the library? how would one deal with this if this library was used outside of Conan, with another package manager for instance? Hence the outcome.

> I suppose we could get a similar effect by setting the GLM flags explicitly in package_info (e.g. self.cpp_info.defines.append("GLM_FORCE_CTOR_INIT"). Or maybe by enforcing the use of common Conan profiles that include tools.build:defines statements.

These are certainly good options as well, that don't require modifying the recipe. We do recommend teams use sync'd up configuration (with `conan config install`) - so if you assume all your teams have the same profiles and most developers are not manually modifying them, this is not a bad option.

Additionally if you only require a single set of values, i.e., you need certain macros to be defined in a single specific value and you don't have multiple combinations of glm options - you can check this in some strategic header such and have a compiler error if the value does not match what you expect. 

> Let's say we have Conan package CalcUtils that uses GLM to perform computations for consumers. It relies on certain GLM preprocessor flags. Some computations are implemented in headers, so these flags must be set for downstream consumers as well.

The same headers that contain this implementation can have the `#define` to the value that works, right before including the public glm headers... which I'd say matches the glm manual.

> Conan options appeared to be a good solution here 

> If these settings are exposed as Conan options, they can be checked/validated via Conan APIs.

These are good solutions for your particular needs - but they seem to go beyond what we aim to maintain in Conan Center, which is provide the libraries as they are provided and intended to be used by their authors. What you are proposing may conflict with that other users expect. 

If it suits your needs, you can host your own version of the recipe, see docs for [hosting your own Conan remote](https://docs.conan.io/2/tutorial/conan_repositories/setting_up_conan_remotes.html), or consider the [local recipes index.](https://docs.conan.io/2/tutorial/conan_repositories/setup_local_recipes_index.html)

proceduralnoisy:
Thanks for your help @jcar87. I am closing this PR.

PR #27614, opened 2025-06-06
Update highs to 1.11.0

galabovaa:
### Summary
Changes to recipe:  **highs/1.11.0**

#### Motivation

Add new version of HiGHS 

#### Details

Version bump and an update of the source directory name in HiGHS 

---
- [ X] Read the [contributing guidelines](https://github.com/conan-io/conan-center-index/blob/master/CONTRIBUTING.md)
- [ X] Checked that this PR is not a duplicate: [list of PRs by recipe](https://github.com/conan-io/conan-center-index/discussions/24240)
- [ X] Tested locally with at least one configuration using a recent version of Conan

### DIFF
--- a/recipes/highs/all/conanfile.py
+++ b/recipes/highs/all/conanfile.py
@@ -34,7 +34,7 @@ class HiGHSConan(ConanFile):
             self.options.rm_safe("fPIC")
 
     def layout(self):
-        cmake_layout(self, src_folder="highs")
+        cmake_layout(self, src_folder="src")
 
     def requirements(self):
         self.requires("zlib/[>=1.2.11 <2]")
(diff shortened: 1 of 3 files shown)


recipes/highs/all/conanfile.py
```diff
@@ -34,7 +34,7 @@ def configure(self):
             self.options.rm_safe("fPIC")
 
     def layout(self):
-        cmake_layout(self, src_folder="src")
+        cmake_layout(self, src_folder="highs")
```
jcar87: thanks @galabovaa - the source directory is only used internally by the recipe, and for all recipes in Conan Center we use the `src` convention

jcar87 (changes requested):
please see comment

recipes/highs/all/conanfile.py
galabovaa: Thank you! I modified it back to "src"

PR #27626, opened 2025-06-09
[libtins] Add compability to CMake 4.x and fix missing defines

uilianries:
Changes to recipe:  **libtins/4.5.[version]**

#### Motivation

When updating the OusterSDK recipe in the PR https://github.com/conan-io/conan-center-index/pull/26108, it resulted in failure due to missing libtins in Conan Center. The last change in libtins recipes was in 2023, and when reviewing the recipe again, there were a few points that slipped out of my hands. This PR fixes those important points.

#### Details

- libtins' CMakeLists.txt there is no `cmake_minium_required` listed. We need to pass `CMAKE_POLICY_VERSION_MINIMUM=3.5` in order to make it compatible with CMake 4.x
- Use `cache_variables` instead of `CMakeToolchain.variables`. It's CMakeLists.txt has no `project()`
- From PR #17470 there is no evidence of needed transitive libs for pcap. Checking the public headers, only a few defines and `pcap_t` (an opaque struct) are used. 
- The public header [ack_tracker.h](https://github.com/mfontanini/libtins/blob/v4.5/include/tins/tcp_ip/ack_tracker.h#L35) checks the define `TINS_HAVE_ACK_TRACKER`
- The public header [stream.h](https://github.com/mfontanini/libtins/blob/v4.5/include/tins/tcp_ip/stream.h#L47) checks the define `TINS_HAVE_TCP_STREAM_CUSTOM_DATA` to use or not Boost.
- Bumped lipcap to 1.10.5 (patch version), so we can align with OusterSDK PR and both use the latest version.

---

### DIFF
--- a/recipes/libtins/all/test_package/test_package.cpp
+++ b/recipes/libtins/all/test_package/test_package.cpp
@@ -1,4 +1,6 @@
 #include <tins/tins.h>
+#include <tins/data_link_type.h>
+#include <tins/tcp_ip/ack_tracker.h>
 
 using namespace Tins;
 
--- a/recipes/libtins/all/conanfile.py
+++ b/recipes/libtins/all/conanfile.py
@@ -58,7 +58,8 @@ class LibTinsConan(ConanFile):
         cmake_layout(self, src_folder="src")
 
     def requirements(self):
-        self.requires("libpcap/1.10.4", transitive_headers=True, transitive_libs=True)
+        # pcap.h is largely imported in public headers: tins/packet_writer.h, tins/sniffer.h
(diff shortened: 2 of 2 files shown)


recipes/libtins/all/test_package/test_package.cpp
```diff
@@ -1,4 +1,6 @@
 #include <tins/tins.h>
+#include <tins/data_link_type.h>
```
uilianries, line 2: So we can validate transitive headers.

AbrilRBS (dismissed):

AbrilRBS (approved):

